PR #15481, opened 2023-08-19
fix: avoid problematic serde release 

matklad:
serde 1.0.172 and up rely on opaque non-reproducible binary blobs to
function, explicitly not providing a library-level opt-out.

This is problematic for two reasons:

- directly, unauditable binary blobs are a security issue.
- indirectly, it becomes much harder to predict future behaviors of the
  crate.

As such, I am willing to go on a limb here and forbid building
rust-analyzer with those versions of serde. Normally, my philosophy is
to defer the choice to the end user, but it's also a design constraint
of rust-analyzer that we don't run random binaries downloaded from the
internet without explicit user's concent.

Concretely, this upper-bounds serde for both rust-analyzer workspace, as
well as the lsp-server lib.

See https://github.com/serde-rs/serde/issues/2538 for wider context.

### DIFF
--- a/lib/lsp-server/Cargo.toml
+++ b/lib/lsp-server/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "lsp-server"
-version = "0.7.2"
+version = "0.7.3"
 description = "Generic LSP server scaffold."
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/rust-lang/rust-analyzer/tree/master/lib/lsp-server"
@@ -9,7 +9,8 @@ edition = "2021"
 [dependencies]
 log = "0.4.17"
 serde_json = "1.0.96"
-serde = { version = "1.0.156", features = ["derive"] }
+# See https://github.com/serde-rs/serde/issues/2538#issuecomment-1684517372 for why we pin serde
+serde = { version = "1.0.156, < 1.0.171", features = ["derive"] }
 crossbeam-channel = "0.5.6"
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


lib/lsp-server/Cargo.toml
```diff
@@ -9,7 +9,8 @@ edition = "2021"
 [dependencies]
 log = "0.4.17"
 serde_json = "1.0.96"
-serde = { version = "1.0.156", features = ["derive"] }
+# See https://github.com/serde-rs/serde/issues/2538#issuecomment-1684517372 for why we pin serde
+serde = { version = "1.0.156, < 1.0.171", features = ["derive"] }
```
ahlinc, line 13: I think there is a mistake and it should be `< 1.0.172` like on https://github.com/rust-lang/rust-analyzer/pull/15481/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R101

ahlinc: @matklad :point_up:

matklad:
reopened as https://github.com/rust-lang/rust-analyzer/pull/15482 --- github's having trouble with updateing the PR after a --force-with-lease?

lib/lsp-server/Cargo.toml
lnicola: Yeah, looks like 1.0.171 is the latest version without it.

ahlinc:
> reopened as #15482 --- github's having trouble with updateing the PR after a --force-with-lease?

May be because on top of this PR it complains on: `The head ref may contain hidden characters: "\u{1FA84}deblobify"`

PR #15482, opened 2023-08-19
fix: avoid problematic serde release

matklad:
serde 1.0.172 and up rely on opaque non-reproducible binary blobs to function, explicitly not providing a library-level opt-out.

This is problematic for two reasons:

- directly, unauditable binary blobs are a security issue.
- indirectly, it becomes much harder to predict future behaviors of the crate.

As such, I am willing to go on a limb here and forbid building rust-analyzer with those versions of serde. Normally, my philosophy is to defer the choice to the end user, but it's also a design constraint of rust-analyzer that we don't run random binaries downloaded from the internet without explicit user's concent.

Concretely, this upper-bounds serde for both rust-analyzer workspace, as well as the lsp-server lib.

See https://github.com/serde-rs/serde/issues/2538 for wider context.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1000,22 +1000,22 @@ checksum = "b06a4cde4c0f271a446782e3eff8de789548ce57dbc8eca9292c27f4a42004b4"
 [[package]]
 name = "lsp-server"
 version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "37ea9ae5a5082ca3b6ae824fc7666cd206b99168a4d4c769ad8fe9cc740df6a6"
 dependencies = [
  "crossbeam-channel",
  "log",
- "lsp-types",
  "serde",
  "serde_json",
 ]
 
 [[package]]
 name = "lsp-server"
-version = "0.7.2"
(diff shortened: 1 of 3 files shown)


lnicola (approved):
I don't know for how long we can keep it like this, but sure, let's see what happens.

Too bad, though 😞.

lnicola:
I didn't look closely, can we build our own version of the blob and run it instead?

Veykril:
I doubt pinning the version will cause us problems given most people seem to do the same currently.

@bors r+

matklad:
Practically, I would thing that just sticking to 171 should be fine, I don’t thing we need latest and greatest serde. 


for compile-time, the bigger problem for us is not that serde is slow to compile, but that it generates a loooot of code. 

Long term, my plan always was to remove serde in favor of directly generateing slow-but-fast—to-compile code from LSP meta model

matklad:
Learned about https://github.com/serde-rs/json/pull/745. On a quick look, that sort of thing could be a major win for us. Also, https://github.com/serde-rs/serde/pull/2250#pullrequestreview-1564212707 has a lot of valuable context.

lnicola:
CC https://github.com/serde-rs/serde/releases/tag/v1.0.184

PR #15483, opened 2023-08-19
internal: post-bump lsp-server version

matklad:


matklad:
let me just check that the permissions are set up correctly...


@bors r+

PR #15484, opened 2023-08-19
Bool to enum assist

### LINKED ISSUE #14779: assist idea bool to enum
An assist that triggers on local variables, fields, constants and statics of type `bool` that allows turning its type into a new enum with two variants `True` and `False` and replacing all usages appropriately. That is replacing assignment by the variants and usages with `== True`  where boolean values are assumed. 

That allow easily creating new type "booleans" and renaming the variants. This is an encouraged pattern and such an assist would make it a lot easier to employ!

rmehri01:
This adds the `bool_to_enum` assist, which converts the type of boolean local variables, fields, constants and statics to a new `enum` type, making it easier to distinguish the meaning of `true` and `false` by renaming the variants.

Closes #14779


crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -0,0 +1,724 @@
...
+            make::tail_only_block_expr(true_expr),
+            Some(ast::ElseBranch::Block(make::tail_only_block_expr(false_expr))),
+        )
+        .clone_for_update()
+    }
+}
+
+/// Replaces all usages of the target identifier, both when read and written to.
+fn replace_usages(edit: &mut SourceChangeBuilder, usages: &UsageSearchResult) {
+    for (_, references) in usages.iter() {
```
rmehri01, line 190: I think it would be nice to handle cross file and module references too, especially in the case of constants. I had a version before where I would just check to see if the `ctx` file is different than the usage file, then I would insert an import which seemed to work out alright for basic files but this doesn't handle modules within a file so I was wondering if there is another approach that would be nicer?

crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -0,0 +1,724 @@
...
+                    );
+                } else if let Some((record_field, initializer)) = find_record_expr_usage(name_ref) {
+                    cov_mark::hit!(replaces_record_expr);
+
+                    let record_field = edit.make_mut(record_field);
+                    let enum_expr = bool_expr_to_enum_expr(initializer);
+                    record_field.replace_expr(enum_expr);
+                } else if name_ref.syntax().ancestors().find_map(ast::Expr::cast).is_some() {
+                    // for any other usage in an expression, replace it with a check that it is the true variant
+                    edit.replace(range, format!("{} == Bool::True", name_ref.text()));
```
rmehri01, line 217: Is there a convention for when to use `edit.replace` vs `ted::replace`? I think it seems nicer to use `ted` instead of a formatted string but I got a bit tripped up with when to make things mutable

crates/ide-assists/src/handlers/bool_to_enum.rs
rmehri01: Maybe handling a test like this could be nice?

```rust
    #[test]
    fn const_in_module() {
        check_assist(
            bool_to_enum,
            r#"
mod foo {
    pub const $0FOO: bool = true;
}

fn main() {
    if foo::FOO {
        println!("foo");
    }
}
"#,
            r#"
mod foo {
    #[derive(PartialEq, Eq)]
    pub enum Bool {
        True,
        False,
    }

    pub const FOO: Bool = Bool::True;
}

fn main() {
    use foo::Bool;

    if foo::FOO == Bool::True {
        println!("foo");
    }
}
"#,
        )
    }
```

rmehri01: Two approaches may be:

1. Use `ImportScope::find_insert_use_container` to check if the definition and usage are in different containers and then do the appropriate `insert_use`.
2. Use `is_` methods for each of the variants, as done in `generate_enum_is_method` so that there are no imports needed.

crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -0,0 +1,728 @@
...
+    target_node: SyntaxNode,
+    name: ast::Name,
+    ty_annotation: Option<ast::Type>,
+    initializer: Option<ast::Expr>,
+    definition: Definition,
+}
+
+/// Attempts to find an appropriate node to apply the action to.
+fn find_bool_node(ctx: &AssistContext<'_>) -> Option<BoolNodeData> {
+    if let Some(let_stmt) = ctx.find_node_at_offset::<ast::LetStmt>() {
```
Veykril, line 88: We should look for `ctx.find_node_at_offset::<ast::Name>() ` first, and then check if that name is part of static, let, or const. Reason being we don't want the assist to be available when the cursor is in the initializer expressions of these constructs whi can be annoying. That is `let foo$0 = ...;` should have the assist be available, but not `let foo = bar::bar$0();`

crates/ide-assists/src/handlers/bool_to_enum.rs
Veykril: Handling it cross modules properly would be great yes, you might be able to check on `crates\ide-assists\src\handlers\extract_struct_from_enum_variant.rs` how things are done there. Iirc it needs to import stuff as well when rewriting other  files

crates/ide-assists/src/handlers/bool_to_enum.rs
Veykril: There isn't really a single proper way to editing unfortunately, either should be fine

rmehri01: Ah okay, no problem then thanks!

crates/ide-assists/src/handlers/bool_to_enum.rs
rmehri01: I see thanks for the pointer! I think I've got it to work in a similar way now using the import approach. The only problem is I can't seem to get the tabstop snippet to show up in the right place because of the added import, so I've just removed the tabstop for now.

crates/ide-assists/src/handlers/bool_to_enum.rs
rmehri01: Ah okay, should be fixed now!

crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -0,0 +1,1514 @@
...
+        }
+
+        Some(BoolNodeData {
+            target_node: let_stmt.syntax().clone(),
+            name,
+            ty_annotation: let_stmt.ty(),
+            initializer: let_stmt.initializer(),
+            definition: Definition::Local(def),
+        })
+    } else if let Some(const_) = name.syntax().ancestors().find_map(ast::Const::cast) {
```
Veykril, line 114:
```suggestion
    } else if let Some(const_) = name.syntax().parent().and_then(ast::Const::cast) {
```

Here we only need the parent, as the name is the direct child of the item.

crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -0,0 +1,1514 @@
...
+        }
+
+        Some(BoolNodeData {
+            target_node: const_.syntax().clone(),
+            name,
+            ty_annotation: const_.ty(),
+            initializer: const_.body(),
+            definition: Definition::Const(def),
+        })
+    } else if let Some(static_) = name.syntax().ancestors().find_map(ast::Static::cast) {
```
Veykril, line 128:
```suggestion
    } else if let Some(static_) = name.syntax().parent().and_then(ast::Static::cast) {
```

crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -0,0 +1,1514 @@
...
+
+        Some(BoolNodeData {
+            target_node: static_.syntax().clone(),
+            name,
+            ty_annotation: static_.ty(),
+            initializer: static_.body(),
+            definition: Definition::Static(def),
+        })
+    } else {
+        let field = name.syntax().ancestors().find_map(ast::RecordField::cast)?;
```
Veykril, line 143:
```suggestion
        let field = name.syntax().parent().and_then(ast::RecordField::cast)?;
```

Veykril (approved):

Veykril:
Thanks! Really looking forward to using this assist 🎉 
@bors r+

PR #15486, opened 2023-08-20
fix: Fix item tree lowering pub(self) to pub()

### LINKED ISSUE #15134: private visibility is shown in autocompletion
I have `pub(self) fn function() {}` in module named `becoding` , and autocomplete in outer files shows that function .
image
but the error is ok 
image

But when it comes to `pub(in crate::becoding)fn function(){}` , autocomplete hides that function .
image
image

petr-tik:
Prior to this, the item tree lowered `pub(self)` visibility to `pub()`
Fix #15134 - tested with a unit test and 
a manual end-to-end test of building rust-analyzer from my branch and opening the reproduction repository

### DIFF
--- a/crates/hir-def/src/nameres/path_resolution.rs
+++ b/crates/hir-def/src/nameres/path_resolution.rs
@@ -93,11 +93,15 @@ impl DefMap {
                 if remaining.is_some() {
                     return None;
                 }
-                let types = result.take_types()?;
-                match types {
-                    ModuleDefId::ModuleId(m) => Visibility::Module(m),
-                    _ => {
-                        // error: visibility needs to refer to module
+                let types = result.take_types();
+
+                match (types, path.kind) {
+                    (Some(ModuleDefId::ModuleId(m)), _) => Visibility::Module(m),
+                    // resolve_path doesn't find any values for a plan pathkind of a private function
+                    (None, PathKind::Plain | PathKind::Crate) => {
+                        Visibility::Module(self.module_id(original_module))
+                    }
+                    (_, _) => {
                         return None;
                     }
                 }
--- a/crates/hir-def/src/nameres/path_resolution.rs
+++ b/crates/hir-def/src/nameres/path_resolution.rs
@@ -93,15 +93,11 @@ impl DefMap {
                 if remaining.is_some() {
                     return None;
                 }
-                let types = result.take_types();
-
-                match (types, path.kind) {
-                    (Some(ModuleDefId::ModuleId(m)), _) => Visibility::Module(m),
-                    // resolve_path doesn't find any values for a plan pathkind of a private function
-                    (None, PathKind::Plain | PathKind::Crate) => {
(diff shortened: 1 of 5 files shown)


crates/hir-def/src/nameres/path_resolution.rs
```diff
@@ -93,11 +93,15 @@ impl DefMap {
...
-                    _ => {
-                        // error: visibility needs to refer to module
+                let types = result.take_types();
+
+                match (types, path.kind) {
+                    (Some(ModuleDefId::ModuleId(m)), _) => Visibility::Module(m),
+                    // resolve_path doesn't find any values for a plan pathkind of a private function
+                    (None, PathKind::Plain | PathKind::Crate) => {
+                        Visibility::Module(self.module_id(original_module))
+                    }
```
Veykril, line 103: This is wrong, the issue seems to be that for some reason we construct a plain path with 0 segments for `pub(self)`, while we should be creating a `Super(0)` path with 0 segments

crates/ide-completion/src/completions/expr.rs
```diff
@@ -9,6 +9,23 @@ use crate::{
...
+    ctx: &CompletionContext<'_>,
+    module: Option<&Module>,
+) -> bool {
+    match (def, module) {
+        (ScopeDef::GenericParam(hir::GenericParam::LifetimeParam(_)) | ScopeDef::Label(_), _) => {
+            false
+        }
+        (ScopeDef::ModuleDef(hir::ModuleDef::Macro(mac)), _) => mac.is_fn_like(ctx.db),
+        (ScopeDef::ModuleDef(hir::ModuleDef::Function(f)), Some(m)) => {
+            f.is_visible_from(ctx.db, *m)
```
Veykril, line 23: This should not be necessary, we already check visibility in `add_path_resolution`

crates/hir-def/src/nameres/path_resolution.rs
petr-tik: Thanks for confirming my suspicion that the root cause was in nameres

> Please let me know what's the best way to add a unit tests to nameres

for the self case as well as the `pub(in crate::krate)` case

petr-tik: is there a difference between `pub(self)` and `pub(self::)`, because they end up with different visibility values

Veykril: I think the problem happens in item tree construction, so a place to add a test case would be in `hir-def/src/item_tree/tests.rs` (ideally one that tests all kinds of visibility paths). `pub(self::)` is different in that this contains an incomplete path, I think they ideally should have the same visibility. Might be good to check what happens here in case of other paths (that is `super` vs `super::`, `crate` vs `crate::` etc_.

petr-tik:
Thanks for your review. Life has been a bit hectic, I will try to get back to this over this weekend

PR #15490, opened 2023-08-21
fix help text for rust-analyzer.check.invocation{Strategy,Location}

RalfJung:
I highly doubt that `check.invocationLocation` only has an effect if `cargo.buildScripts.overrideCommand` is set -- looks like a copy-paste mistake from `buildScripts.invocationLocation` to me.


Veykril:
Yep thats a copy paste mistake.
Regarding failing CI thats because you'll need to regenerate some files, but I'll go ahead an do that since that currently requires building the entirety of r-a.

Veykril:
I guess you beat me to it 😄

RalfJung:
Ah sorry... yeah I grepped and updated the other files by hand.^^ But it's possible I got that wrong, so if it fails again I will leave it to you. :)

Veykril:
Seems to have worked
@bors r+

PR #15492, opened 2023-08-21
extend check.overrideCommand and buildScripts.overrideCommand docs

RalfJung:
Extend check.overrideCommand and buildScripts.overrideCommand docs regarding invocation strategy and location.

However something still seems a bit odd -- the docs for `invocationStrategy`/`invocationLocation` talk about "workspaces", but the setting that controls which workspaces are considered is called `linkedProjects`. Is a project the same as a workspace here or is there some subtle difference?


Veykril:
They are the same in this regard. The terms are kind of confusing, projects, workspaces, cargo workspaces, vscode workspaces. `linkedProjects` is named as such because it was initially meant for pointing to the `rust-project.json` format, but you can also point it to cargo manifests.

RalfJung:
To make things worse, the "once" option is documented as

> The command will be executed in the project root.

So this uses "project" in a way very different from `linkedProjects`...

Veykril:
We should probably define some terms and use them appropriately in the docs, as sometimes we do want to talk about the entire opened directory opposed to contained cargo workspace within. I can put that on my homework list

PR #15494, opened 2023-08-21

lnicola:


lnicola:
Not sure what's happening here. The two tests (`resolve_proc_macro` and `out_dirs_check`) also fail for me on `master`, when running `CI=1 cargo +nightly test --features sysroot-abi`. CI passes because we don't run them most of the time.

lnicola:
Huh, it's now working after upgrading to the latest nightly.

Veykril:
It looks like `internal_features` was set to deny by default now, so we need to allow that on the test fixture's builds script part of the failing test?

Veykril:
This is the one part thats annoying about upstream CI not running r-a tests, because this is the one integration test we do want to run ...

lnicola:
I can reproduce the hang on `nightly`. I think we didn't spot this because we usually run `analysis-stats` on `stable` on CI.

CC @HKalbasi

.github/workflows/ci.yaml
```diff
@@ -86,6 +86,12 @@ jobs:
       - name: Test
         run: cargo test ${{ env.USE_SYSROOT_ABI }} -- --nocapture --quiet
 
+      - name: Switch to stable toolchain
```
lnicola, line 89: @Veykril heads-up, just so you're aware of this change.

PR #15515, opened 2023-08-25
Check if lhs is also a binexpr and use its rhs in flip binexpr assist

### LINKED ISSUE #15508: Flip binary expression and associativity
**rust-analyzer version**: rust-analyzer 1.73.0-nightly (180dffb 2023-08-14)

The *flip binary expression* assist behaves a bit unexpectedly with associative operators. If I flip the rightmost plus here:
```rs
1 + 2 + 3 + 4;
          ^
```

The 4 is moved all the way to the left.

```rs
4 + 1 + 2 + 3;
  ^
```

While not mathematically incorrect, it is a bit unexpected.

cardoso:
Closes #15508 

From the original PR, flip binexpr assist is not meant to preserve equivalence, so I went with the simplest solution here.

I can add some extra checks to keep equivalence, but I think they should go in different specific assists (eg. flip arith op / flip logic op / etc), otherwise this one will get out of hand pretty quickly.


Veykril:
r=me with CI green
@bors delegate+

PR #15517, opened 2023-08-26
fix: diagnostics for 'while let' loop with label in condition

### LINKED ISSUE #15516: Incorrect report of undeclared label
**rust-analyzer version**: 0.3.1631-standalone

**rustc version**: rustc 1.72.0 (5680fa18f 2023-08-23)

**relevant settings**: I don't believe I've set anything particular. I can share the rust-analyzer settings file if need be.
Hello, I'm having an issue with ra where it reports a ```use of undeclared label `'my_label` ``` (which I easily visualize with error lens on code).
Neither `cargo check` nor `clippy` report any warning (much less error as is the case with ra) regarding this when I run them in my terminal.
I'm not sure about a minimal example but here's one I quickly crafted from the actual snippet that causing me the issue.
```rs
struct PlaceHolder;

impl Iterator for PlaceHolder {
    type Item = Result<Option<()>, ()>;

    fn next(&mut self) -> Option<Self::Item> {
        Some(Ok(Some(())))
    }
}

fn main() {
    let mut placeholder = PlaceHolder;
    'my_label: while let Some(a) = match placeholder.next().unwrap() {
        Ok(_) => todo!(),
        Err(err) => {
            continue 'my_label; // use of undeclared label `'my_label`
        }
    } {
        a
    }
}
```

image

xffxff:
fix #15516

### DIFF
--- a/crates/hir-def/src/body/lower.rs
+++ b/crates/hir-def/src/body/lower.rs
@@ -744,7 +744,27 @@ impl ExprCollector<'_> {
     fn collect_while_loop(&mut self, syntax_ptr: AstPtr<ast::Expr>, e: ast::WhileExpr) -> ExprId {
         let label = e.label().map(|label| self.collect_label(label));
         let body = self.collect_labelled_block_opt(label, e.loop_body());
-        let condition = self.collect_expr_opt(e.condition());
+
+        // Labels can also be used in the condition expression, like this:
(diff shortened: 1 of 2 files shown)


lnicola (approved):

lnicola:
Thanks!

@bors r+

PR #15520, opened 2023-08-27
FIX: generic parameters handling by the `generate_delegate_trait` assist

ponyii:
A part of https://github.com/rust-lang/rust-analyzer/issues/15108
I'm going to write a separate PR to solve the problem completely; currently only `impl` bofy generation is fixed, but generation of the `impl ...` line itself is to be overhauled. 

The code might look somewhat complicated, but basically the changes are as follows:
1. the destribution of generic params between trait and type is memorized (see `item_tree.rs`);  
2. the path transformation source scope is updated to handle generic params correctly (see `generate_delegate_trait.rs`)

### DIFF
--- a/crates/hir-def/src/item_tree.rs
+++ b/crates/hir-def/src/item_tree.rs
@@ -696,9 +696,33 @@ pub struct TraitAlias {
     pub ast_id: FileAstId<ast::TraitAlias>,
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct GenericParamGroups {
+    // pub target_trait: Option<Vec<String>>,
+    pub self_ty: Option<Vec<String>>,
+}
+
+impl GenericParamGroups {
+    fn from_impl(impl_: &ast::Impl) -> Self {
+        Self {
+            // target_trait: Self::get_arg_list(impl_.trait_()),
+            self_ty: Self::get_arg_list(impl_.self_ty()),
+        }
+    }
+
+    fn get_arg_list(ty: Option<ast::Type>) -> Option<Vec<String>> {
+        let list = match ty? {
+            ast::Type::PathType(p) => p.path()?.segment()?.generic_arg_list(),
+            _ => None,
+        };
+        Some(list?.generic_args().map(|arg| arg.syntax().text().to_string()).collect())
+    }
+}
+
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Impl {
     pub generic_params: Interned<GenericParams>,
+    pub generic_param_groups: GenericParamGroups,
     pub target_trait: Option<Interned<TraitRef>>,
     pub self_ty: Interned<TypeRef>,
     pub is_negative: bool,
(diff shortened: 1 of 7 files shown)


crates/hir-def/src/item_tree.rs
```diff
@@ -696,9 +696,33 @@ pub struct TraitAlias {
...
+
+    fn get_arg_list(ty: Option<ast::Type>) -> Option<Vec<String>> {
+        let list = match ty? {
+            ast::Type::PathType(p) => p.path()?.segment()?.generic_arg_list(),
+            _ => None,
+        };
+        Some(list?.generic_args().map(|arg| arg.syntax().text().to_string()).collect())
+    }
+}
+
```
Veykril, line 721: This is the wrong way too approach this. The fix to the issue here should not involve touching item tree whatsoever.

crates/hir/src/lib.rs
```diff
@@ -3355,6 +3356,21 @@ impl Impl {
         let src = self.source(db)?;
         src.file_id.as_builtin_derive_attr_node(db.upcast())
     }
+
+    pub fn filter_self_ty_params(
```
Veykril, line 3360: The name isnt really descriptive, I'm not sure I understand what this is filtering

crates/hir-def/src/item_tree.rs
ponyii: I has to touch it, I'm afraid.

Let's have a look on few relevant lines from the test:
```
struct S(Foo<i32, bool>);
...
impl<A, B, C> Trait<A, B> for Foo<C, B> {
```
In this case the assist should turn `C` and `B` into `i32` and `bool` repectively. In general, it should get `Foo`-related generic arguments from the trait implementation and correlate them `Foo`'s parameters within the structure. Consequently, we have to memorize the "distribution" of the generic arguments in `impl` lines; that's exactly what happens in this file - the names of args pertaining to `self_ty` are being saved.

crates/hir/src/lib.rs
ponyii: It filters an `impl`'s generic arguments pertaining to `self_ty`, e.g. for `impl<A, B, C> Trait<A, B> for Foo<C, B>` it will return `C` and `B` only. I failed to find a better name, do you have any suggestions? May be I could just add a comment with this example?

crates/hir-def/src/item_tree.rs
Veykril: That data is already encoded in the `sekf_ty` `TypeRef` in the item tree. So you just need to fetch that from there instead, that is check if the `TypeRef` is a `TypeRef::Path` and take the generic args from that path instead

crates/hir/src/lib.rs
Veykril: Ye a, comment would be good for now

crates/hir-def/src/item_tree.rs
ponyii: Thank you! Now these data are fetched from `Type` instead of `TypeRef`, but anyway the item tree is unchanged

PR #15524, opened 2023-08-28
Bind unused parameter assistant

vsrs:
This PR introduces a new **Bind unused parameter assistant**.

While we do have a QuickFix from `rustc` (prefixing the parameter with an underscore), it's sometimes more convenient to suppress the warning using the following approach:
```rust
 fn some_function(unused: i32) {}
```
->
```rust
fn some_function(unused: i32) {
    let _ = unused;
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/bind_unused_param.rs
+++ b/crates/ide-assists/src/handlers/bind_unused_param.rs
@@ -27,10 +27,7 @@ use super::remove_unused_param::is_trait_impl;
 pub(crate) fn bind_unused_param(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
     let param: ast::Param = ctx.find_node_at_offset()?;
 
-    let ident_pat = match param.pat()? {
-        ast::Pat::IdentPat(it) => it,
-        _ => return None,
-    };
+    let Some(ast::Pat::IdentPat(ident_pat)) = param.pat() else { return None };
 
     let func = param.syntax().ancestors().find_map(ast::Fn::cast)?;
     if is_trait_impl(&func) {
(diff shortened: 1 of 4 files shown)


crates/ide-assists/src/handlers/bind_unused_param.rs
```diff
@@ -0,0 +1,159 @@
...
+//     let _ = x;
+// }
+// ```
+pub(crate) fn bind_unused_param(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
+    let param: ast::Param = ctx.find_node_at_offset()?;
+
+    let ident_pat = match param.pat()? {
+        ast::Pat::IdentPat(it) => it,
+        _ => return None,
+    };
```
Veykril, line 33: nit
```suggestion
    let Some(ast::Pat::IdentPat(ident_pat)) = param.pat() else { return None };
```

crates/ide-assists/src/handlers/bind_unused_param.rs
```diff
@@ -0,0 +1,159 @@
...
+    let stmt_list = func.body()?.stmt_list()?;
+    let l_curly_range = stmt_list.l_curly_token()?.text_range();
+    let r_curly_range = stmt_list.r_curly_token()?.text_range();
+    let left_line = line_index.line_col(l_curly_range.end()).line;
+    let right_line = line_index.line_col(r_curly_range.start()).line;
+
+    if left_line == right_line {
+        cov_mark::hit!(single_line);
+        text.push('\n');
+    }
```
Veykril, line 65: All of this can be moved into the `|builder|` closure, except for the `let stmt_list = func.body()?.stmt_list()?;`

Veykril (commented):
Would be nice to have this as a diagnostic quickfix oncew e have a diagnostic for unused vars

vsrs:
> Would be nice to have this as a diagnostic quickfix oncew e have a diagnostic for unused vars

`remove_unused_param` assist from https://github.com/rust-lang/rust-analyzer/blob/62268e474e9165de0cdb08d3794eec4b6ef1c6cd/crates/ide-assists/src/handlers/remove_unused_param.rs#L78 has `AssistKind::Refactor` kind; I belive it's worth to have the same (or change both to QuickFix)

Veykril:
That one I'd say should also ideally be a diagnostic quickfix

vsrs:
Just realized that this assist is different from 'remove_unused_param' one and can work for trait impl methods too. So, rollbacked all refactorings in `remove_unused_param` code.

Veykril (approved):

PR #15528, opened 2023-08-29
Enable `rust_analyzer` for cfgs when code is being analyzed by rust-analyzer

Veykril:
This allows one to have r-a skip analysis/replace macros that work not well with r-a at all by gating them behind this cfg (an example being the `quote` macro which r-a struggles with in terms of performance).


Veykril:
Actually, let me change things so that we only set these for the workspace to prevent people from relying on this in their libraries
@bors r-

PR #15532, opened 2023-08-29
On type format '(', by adding closing ')' automatically

SomeoneToIgnore:
If I understand right, `()` can surround pretty much the same `{}` can, so add another on type formatting pair for convenience: sometimes it's not that pleasant to write parenthesis in `Some(2).map(|i| (i, i+1))` cases and I would prefer r-a to do that for me.

One note: currently, https://github.com/rust-lang/rust-analyzer/blob/b06503b6ec98c9ed44698870cbf3302b8560b442/crates/rust-analyzer/src/handlers/request.rs#L357 fires always.
Should we remove the assertion entirely now, since apparently things work in release despite that check?

### DIFF
--- a/crates/ide/src/typing.rs
+++ b/crates/ide/src/typing.rs
@@ -32,7 +32,7 @@ use crate::SourceChange;
 pub(crate) use on_enter::on_enter;
 
 // Don't forget to add new trigger characters to `server_capabilities` in `caps.rs`.
-pub(crate) const TRIGGER_CHARS: &str = ".=<>{";
+pub(crate) const TRIGGER_CHARS: &str = ".=<>{(";
 
 struct ExtendedTextEdit {
     edit: TextEdit,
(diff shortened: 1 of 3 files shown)


DropDemBits:
> Should we remove the assertion entirely now, since apparently things work in release despite that check?

Ah, I was the one who added that assert in, and only realized now that this should be a `stdx::always` instead of a `stdx::never` 😅

<details><summary>More context on this assert</summary>
I originally added this assert since structured snippets add in placeholders & tabstops later on instead of embedding them directly, and I didn't think it was worth it right now to port the on-type formatting to structured snippets as it only uses one tabstop in a fixed location.
</details>

SomeoneToIgnore:
Thank you, works much better with the different assert 🙂

I've pushed the commit here, but feel free to create a separate PR, you have better context and can describe it better.

Veykril:
Now I wonder, we might as well add `[` to the mix here no?

Veykril:
Well we can do that in separate PR, but I think that should work in all positions there likewise.
@bors r+

SomeoneToIgnore:
> Now I wonder, we might as well add `[` to the mix here no?

I personally rarely index something nested with this, so was afraid to add it myself.
Also, was not sure if it's the same as `(` rule-wise to apply its counterpart?
Anyway, not sure if very requested feature, so I restrained from bloating the PR.

Ideally, I'd add `|` since typing closures is sometimes hard, but that needs a but different set of rules.

Veykril:
I was more thinking of arrays than indexing here, unsure whether its useful though.

PR #15552, opened 2023-09-03
Bump `rustc-perf` checkout for metrics, replace webrender, diesel and ripgrep with newer versions

Veykril:
The previous versions were removed hence the updates. The metric results for the changed ones are disconnected now as to not muddle things up.

I think it's worth it for us to occasionally bump the `rustc-perf` checkout, as it allows us to include more interesting new targets.

Notably, the newer diesel is 3 times as fast to analyze as the one of the current checkout.

### DIFF
--- a/.github/workflows/metrics.yaml
+++ b/.github/workflows/metrics.yaml
@@ -67,7 +67,7 @@ jobs:
   other_metrics:
     strategy:
       matrix:
-        names: [self, ripgrep, webrender, diesel]
+        names: [self, ripgrep-13.0.0, webrender-2022, diesel-1.4.8]
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


Veykril:
r? @lnicola (since you've seen it already :^)

Gonna check how much time we get back to spent on new targets after this has been merged

lnicola (approved):
Looks good, as far as I can tell by looking on my phone, but I didn't actually check the paths etc.

Veykril:
I added an assert that checks the paths (otherwise we do perf on the rustc-perf repo instead if the path is mistyped thanks to how our discovery works 😅), so they are good

@bors r=lnicola

Veykril:
Had to change CI to pick up the new files
@bors r+

PR #15559, opened 2023-09-05
Implement builtin#format_args, using rustc's format_args parser

Veykril:
`format_args!` now expands to `builtin#format_args(template, args...)`, the actual expansion now instead happens in lowering where we desugar this expression by using lang paths. 

As a bonus, we no longer need to evaluate `format_args` as an eager macro which means less macro expansions overall -> less cache thrashing!

Fixes https://github.com/rust-lang/rust-analyzer/issues/15082
cc https://github.com/rust-lang/rust/issues/110680


Veykril:
> exprs: 662115, ??ty: 30 (0%), ?ty: 116 (0%), !ty: 3
> pats: 155657, ??ty: 4 (0%), ?ty: 5 (0%), !ty: 0

This does increase analysis-stats mir memory consumption by ~80 mb though, definitely something we wanna take a look at again (as in to optimize if possible). That makes me wonder if there is some easy to write MIR opt that can optimize for MIR size?

crates/hir-expand/src/builtin_fn_macro.rs
```diff
@@ -247,160 +243,22 @@ fn format_args_expand_general(
     _db: &dyn ExpandDatabase,
     _id: MacroCallId,
     tt: &tt::Subtree,
-    end_string: &str,
+    // FIXME: Make use of this so that mir interpretation works properly
```
Veykril, line 246: Note, (I'm probably not fixing this in this PR) `format_args_nl!` will behave the same as `format_args!` that is no newline will be appended. This is fine for most hings but mir eval.

Veykril: I'm unsure how to solve this, we could give the construct an extra "parameter" or something that tells it to emit an additional newline, make a `builtin#format_args_nl` construct or try to fiddle with the template string which I really do not want to do (and I don't think we can reasonably do that).

Veykril:
Let's see how this goes
@bors r+

crates/hir-expand/src/builtin_fn_macro.rs
Veykril: I'm leaning towards adding a second builtin# variant, it seems simplest

Veykril:
Looks like debug render on hover is still broken though, so something fails in mir eval still

HKalbasi:
> That makes me wonder if there is some easy to write MIR opt that can optimize for MIR size

Yes, there are some easy optimization passes that we can implement. Generating a better MIR in first place is also something we can try, which may achieve somethings that easy optimizations can't.

 > Looks like debug render on hover is still broken though, so something fails in mir eval still

Yes, also failed mirs on self count is 2488, while I hoped it to be dropped to near zero after this PR. Will look at it.

PR #15560, opened 2023-09-05
project-model: when using `rust-project.json`, prefer the sysroot-defined rustc over discovery in `$PATH`

davidbarsky:
At the moment, rust-analyzer discovers `rustc` via the `$PATH` even if the `sysroot` field is defined in a `rust-project.json`. However, this does not work for users who do not have rustup installed, resulting in any `cfg`-based inference in rust-analzyer not working correctly. In my (decently naive!) opinion, it makes more sense to rely on the `sysroot` field in the `rust-project.json`.

One might ask "why not add `rustc` to the `$PATH`?" That is a reasonable question, but that doesn't work for my use case:
- The path to the sysroot in my employer's monorepo changes depending on which platform a user is on. For example, if they're on Linux, they'd want to use the sysroot defined at path `a`, whereas if they're on macOS, they'd want to use the sysroot at path `b` (I wrote the sysroot resolution functionality [here](https://github.com/facebook/buck2/blob/765da4ca1e0b97a0de1e82b2fb3af52766fd06f4/integrations/rust-project/src/sysroot.rs#L39), if you're curious).
- The location of the sysroot can (and does!) change, especially as people figure out how to make Rust run successfully on non-Linux platforms (e.g., iOS, Android, etc.) in a monorepo. Updating people's `$PATH` company-wide is hard while updating a config inside a CLI is pretty easy.

## Testing

I've created a `rust-project.json` using [rust-project](https://github.com/facebook/buck2/tree/main/integrations/rust-project) and was able to successfully load a project with and without the `sysroot`/`sysroot_src` fields—without those fields, rust-analyzer fell back to the `$PATH` based approach, as evidenced by `[DEBUG project_model::rustc_cfg] using rustc from env rustc="rustc"` showing up in the logs.

### DIFF
--- a/crates/project-model/src/rustc_cfg.rs
+++ b/crates/project-model/src/rustc_cfg.rs
@@ -102,6 +102,5 @@ fn get_rust_cfgs(
         cmd.args(["--target", target]);
     }
 
-    let out = utf8_stdout(cmd).context("Unable to run `rustc`")?;
-    Ok(out)
+    utf8_stdout(cmd).context("Unable to run `rustc`")
 }
--- a/crates/project-model/src/rustc_cfg.rs
+++ b/crates/project-model/src/rustc_cfg.rs
@@ -4,10 +4,11 @@ use std::process::Command;
 
 use rustc_hash::FxHashMap;
 
-use crate::{cfg_flag::CfgFlag, utf8_stdout, ManifestPath};
+use crate::{cfg_flag::CfgFlag, utf8_stdout, ManifestPath, Sysroot};
 
 pub(crate) fn get(
     cargo_toml: Option<&ManifestPath>,
+    sysroot: Option<&Sysroot>,
     target: Option<&str>,
     extra_env: &FxHashMap<String, String>,
 ) -> Vec<CfgFlag> {
@@ -25,7 +26,7 @@ pub(crate) fn get(
     // Add miri cfg, which is useful for mir eval in stdlib
     res.push(CfgFlag::Atom("miri".into()));
 
-    match get_rust_cfgs(cargo_toml, target, extra_env) {
+    match get_rust_cfgs(cargo_toml, sysroot, target, extra_env) {
         Ok(rustc_cfgs) => {
             tracing::debug!(
                 "rustc cfgs found: {:?}",
@@ -44,6 +45,7 @@ pub(crate) fn get(
 
 fn get_rust_cfgs(
     cargo_toml: Option<&ManifestPath>,
+    sysroot: Option<&Sysroot>,
     target: Option<&str>,
     extra_env: &FxHashMap<String, String>,
 ) -> anyhow::Result<String> {
@@ -62,8 +64,22 @@ fn get_rust_cfgs(
             Err(e) => tracing::debug!("{e:?}: falling back to querying rustc for cfgs"),
         }
     }
+
+    let rustc = match sysroot {
+        Some(sysroot) => {
+            let rustc = sysroot.discover_rustc()?.into();
+            tracing::debug!(?rustc, "using rustc from sysroot");
+            rustc
+        }
+        None => {
+            let rustc = toolchain::rustc();
+            tracing::debug!(?rustc, "using rustc from env");
+            rustc
+        }
+    };
+
     // using unstable cargo features failed, fall back to using plain rustc
-    let mut cmd = Command::new(toolchain::rustc());
+    let mut cmd = Command::new(rustc);
     cmd.envs(extra_env);
     cmd.args(["--print", "cfg", "-O"]);
     if let Some(target) = target {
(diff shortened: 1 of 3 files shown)


crates/project-model/src/rustc_cfg.rs
```diff
@@ -25,49 +32,68 @@ pub(crate) fn get(
...
+            let mut cmd = Command::new(toolchain::cargo());
+            cmd.envs(extra_env);
+            cmd.current_dir(cargo_toml.parent())
+                .args(["rustc", "-Z", "unstable-options", "--print", "cfg"])
+                .env("RUSTC_BOOTSTRAP", "1");
+            if let Some(target) = target {
+                cmd.args(["--target", target]);
+            }
+
+            return utf8_stdout(cmd).context("Unable to run `cargo rustc`");
```
davidbarsky, line 85: quick note for @Veykril: I removed the fallback to `rustc` here for the following two reasons: 
- I'm not entirely sure in what cases `RUSTC_BOOTSTRAP=1` would fail but invoking `rustc` directly would pass.
- If I'm reading the discussion on https://github.com/rust-lang/rust-analyzer/pull/8774 correctly, the fallback was meant for users who might not have a nightly installed and rust-analyzer would rely on a less accurate `rustc`. However, with the switch to `RUSTC_BOOTSTRAP=1`, it seems to me that the fallback was kept as an oversight, so I took the opportunity to remove that fallback.

crates/project-model/src/rustc_cfg.rs
```diff
@@ -2,14 +2,21 @@
 
 use std::process::Command;
 
+use anyhow::Context;
 use rustc_hash::FxHashMap;
 
-use crate::{cfg_flag::CfgFlag, utf8_stdout, ManifestPath};
+use crate::{cfg_flag::CfgFlag, utf8_stdout, ManifestPath, Sysroot};
+
+pub(crate) enum RustcCfgConfig<'a> {
```
davidbarsky, line 18: The name here is a little generic, but I'm struggling to think of anything better. In any case, I hope that the doc comment helps a little.

crates/project-model/src/rustc_cfg.rs
```diff
@@ -2,14 +2,21 @@
...
+    Cargo(&'a ManifestPath),
+    Explicit(&'a Sysroot),
+    Discover,
+}
 
 pub(crate) fn get(
-    cargo_toml: Option<&ManifestPath>,
     target: Option<&str>,
     extra_env: &FxHashMap<String, String>,
+    config: RustcCfgConfig<'_>,
```
davidbarsky, line 27: I've restructured the signature of this function to more closely align with how the rest of rust-analyzer is structured: it was getting a little unwieldly.

crates/project-model/src/sysroot.rs
```diff
@@ -115,10 +115,19 @@ impl Sysroot {
         Ok(Sysroot::load(sysroot_dir, src))
     }
 
-    pub fn discover_rustc(&self) -> Option<ManifestPath> {
+    pub fn discover_rustc_src(&self) -> Option<ManifestPath> {
```
davidbarsky, line 118: This name appeared to be a little inaccurate given its usage, but happy to undo.

crates/project-model/src/sysroot.rs
```diff
@@ -115,10 +115,19 @@ impl Sysroot {
...
         get_rustc_src(&self.root)
     }
 
+    pub fn discover_rustc(&self) -> Result<AbsPathBuf, std::io::Error> {
```
davidbarsky, line 122: The body of this function largely replicates some functionality that exists elsewhere, but if it's an issue, I'd be happy to refactor that.

davidbarsky (commented):
No rush on reviewing this, but I wanted to leave comments on the approach I took here since changes inside of `project-model` can be kinda tricky.

crates/project-model/src/rustc_cfg.rs
davidbarsky: `builtin_cfg` is a good option. While we're bikeshedding, what do you think something like `target_cfg` since the output consists of things like:

<details>
```
rust-analyzer on davidbarsky/use-sysroot-rustc-to-determine-cfgs [$] via 🦀 v1.71.0
❯ rustc --print cfg -O
panic="unwind"
target_arch="aarch64"
target_endian="little"
target_env=""
target_family="unix"
target_feature="aes"
target_feature="crc"
target_feature="dit"
target_feature="dotprod"
target_feature="dpb"
target_feature="dpb2"
target_feature="fcma"
target_feature="fhm"
target_feature="flagm"
target_feature="fp16"
target_feature="frintts"
target_feature="jsconv"
target_feature="lor"
target_feature="lse"
target_feature="neon"
target_feature="paca"
target_feature="pacg"
target_feature="pan"
target_feature="pmuv3"
target_feature="ras"
target_feature="rcpc"
target_feature="rcpc2"
target_feature="rdm"
target_feature="sb"
target_feature="sha2"
target_feature="sha3"
target_feature="ssbs"
target_feature="vh"
target_has_atomic="128"
target_has_atomic="16"
target_has_atomic="32"
target_has_atomic="64"
target_has_atomic="8"
target_has_atomic="ptr"
target_os="macos"
target_pointer_width="64"
target_vendor="apple"
unix
```
</details>

PR #15568, opened 2023-09-07
Replace format-args parser with upstream fork

Veykril:
Turns out we can't bump rustc_abi right now because it got its generics removed accidentally https://github.com/rust-lang/rust/pull/107163


Veykril:
well thats annoying, upstream changed things so much since thats its non trivial to update

PR #15587, opened 2023-09-09
Fix autoimport does nothing when importing trait that is as _ imports

### LINKED ISSUE #15128: autoimport does nothing when importing trait that is `as _` imports
```rs
mod foo {
    pub trait Foo {
        fn f(self) {}
    }
    impl Foo for () {}
}
use self::foo::Foo as _;
fn f<X: Foo$0>() {
    ().foo();    
}

```

dfireBird:
Potentially fixes #15128 

There are two cases of imports:
1. With simple path
2. With use tree list (or say complex path).

On deeper inspection, the [`recursive_merge`](https://github.com/rust-lang/rust-analyzer/blob/994df3d6a31d39f11600f30a6df0b744b13937c1/crates/ide-db/src/imports/merge_imports.rs#L87) function (called by [`try_merge_trees_mut`)](https://github.com/rust-lang/rust-analyzer/blob/994df3d6a31d39f11600f30a6df0b744b13937c1/crates/ide-db/src/imports/merge_imports.rs#L69) is meaningful only in the case of complex path (i.e when the UseTree contains a UseTreeList). 

The [`recursive_merge`](https://github.com/rust-lang/rust-analyzer/blob/994df3d6a31d39f11600f30a6df0b744b13937c1/crates/ide-db/src/imports/merge_imports.rs#L87) function has [match with `Ok` arm](https://github.com/rust-lang/rust-analyzer/blob/994df3d6a31d39f11600f30a6df0b744b13937c1/crates/ide-db/src/imports/merge_imports.rs#L106), that is only executed when both LHS and RHS has `PathSegment` with same `NameRef`. The removal of underscore is implemented in this arm in the case of complex path.

For simple paths, the underscore is removed by checking if both LHS and RHS are simple paths and if their `Path` is same (the check is done [here](https://github.com/rust-lang/rust-analyzer/blob/994df3d6a31d39f11600f30a6df0b744b13937c1/crates/ide-db/src/imports/merge_imports.rs#L74)) and remove the underscore if one is found (I made an assumption here that RHS will always be what rust-analyzer suggests to import, because at this point I'm not sure how to remove underscore with help of `ted::replace`).

### DIFF
--- a/crates/ide-db/src/imports/insert_use/tests.rs
+++ b/crates/ide-db/src/imports/insert_use/tests.rs
@@ -993,6 +993,46 @@ use foo::bar::qux;
     );
 }
 
+#[test]
+fn insert_with_renamed_import_simple_use() {
+    check_with_config(
+        "use self::foo::Foo",
+        r#"
+use self::foo::Foo as _;
+"#,
+        r#"
+use self::foo::Foo;
+"#,
+        &InsertUseConfig {
+            granularity: ImportGranularity::Crate,
+            prefix_kind: hir::PrefixKind::BySelf,
+            enforce_granularity: true,
+            group: true,
+            skip_glob_imports: true,
+        },
+    );
+}
+
+#[test]
+fn insert_with_renamed_import_complex_use() {
+    check_with_config(
+        "use self::foo::Foo;",
+        r#"
+use self::foo::{self, Foo as _, Bar};
+"#,
+        r#"
+use self::foo::{self, Foo, Bar};
+"#,
+        &InsertUseConfig {
+            granularity: ImportGranularity::Crate,
+            prefix_kind: hir::PrefixKind::BySelf,
+            enforce_granularity: true,
+            group: true,
+            skip_glob_imports: true,
+        },
+    );
+}
+
 fn check_with_config(
     path: &str,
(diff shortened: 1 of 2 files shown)


dfireBird:
++ @Veykril I described the logic as descriptive as possible. I hope it's understandable and you could relate it with the diff. Please give any feedback if you have any?

dfireBird:
Regarding tests, I forgot to mention. But this only happens when if the `rust-analyzer.imports.prefix` is set to "self", but for these tests the setting was set to "plain". I didn't see a way to override it for just single test or two. So I was not sure how to proceed with tests. 

P.S: I actually have the tests in my local tree with modified config such `rust-analyzer.imports.prefix` is set "self" instead of "plain". It breaks other tests in the module (not all but some) but works on my tests

dfireBird:
I was adding the tests in `tests` module of https://github.com/dfireBird/rust-analyzer/blob/fix-15128/crates/ide-assists/src/handlers/auto_import.rs . That's why I couldn't find the `check_with_config`. I didn't realize the tests should be added to the module you noted. Sorry, my bad.

Veykril (approved):

Veykril:
Thanks!
@bors r+

Issue #15398, opened 2023-08-05
De-`unwrap` all assists


lnicola:
Historically, the assists have been one of our major source of crashes, usually because they didn't handle invalid or incomplete code. Let's try to clean these up a bit.

Here are the ones I've found, though some of the usages there might actually be correct.

- [ ] generate_derive
- [ ] generate_delegate_trait #15397
- [ ] remove_unused_imports
- [ ] inline_call
- [ ] extract_function
- [ ] add_missing_match_arms
- [ ] wrap_return_type_in_result
- [ ] generate_delegate_methods
- [ ] desugar_doc_comment
- [ ] remove_dbg
- [ ] generate_default_from_new
- [ ] convert_to_guarded_return
- [ ] convert_comment_block
- [ ] convert_named_struct_to_tuple_struct
- [ ] apply_demorgan

lowr:
The [`unwrap()` in `remove_dbg`](https://github.com/rust-lang/rust-analyzer/blob/70fa270637193a4eb35fc7c07728c4e1f79a118b/crates/ide-assists/src/handlers/remove_dbg.rs#L49) is fine because `replacements` is checked to be non-empty right before it.

lnicola:
That's right, but we could still extract that into a local variable and check it properly. It's not worse, and we won't spend time figuring it out again the next 10 times we go through the assists.

lowr:
A comment would suffice I'd say. It's definitely bad to `unwrap()` without any explanation but for those as simple as this one I'd rather not write unnecessary control flow. I'll file a PR that adds a comment if that sounds reasonable.

Meanwhile, I've found that `unwrap()` in `generate_default_from_new` *is* problematic, sigh.

lnicola:
I mean, a `?` (or even my original suggestion) will let us remove the check above, so it's not more control flow. It's much easier to grep and not have the unwrap there than to read the comment and think 😅.

lnicola:
By the way, my idea here is to clean up as much of these as we can, then add a "ratchet"-style test to make sure we're not adding more by mistake.

Milo123459:
can this be closed because all the tasks have been completed?

lnicola:
Almost. I think there are a couple more in `generate_function` which we can turn into `expect`.

Milo123459:
can i potentially look into doing those?

lnicola:
Sure. Check out lines 407 and 686 in `generate_function.rs`.

[...]

PR #15594, opened 2023-09-11
Deunwrap add_missing_match_arms

alibektas:
Last subtask of #15398

### DIFF
--- a/crates/ide-assists/src/handlers/add_missing_match_arms.rs
+++ b/crates/ide-assists/src/handlers/add_missing_match_arms.rs
@@ -1944,4 +1944,35 @@ fn main() {
 "#,
         );
     }
+
+    /// See [`discussion`](https://github.com/rust-lang/rust-analyzer/pull/15594#discussion_r1322960614)
+    #[test]
+    fn missing_field_name() {
+        check_assist(
+            add_missing_match_arms,
+            r#"
+enum A {
+    A,
+    Missing { a: u32, : u32, c: u32 }
+}
+
+fn a() {
+    let b = A::A;
+    match b$0 {}
+}"#,
+            r#"
+enum A {
+    A,
+    Missing { a: u32, : u32, c: u32 }
+}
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/add_missing_match_arms.rs
```diff
@@ -446,21 +447,23 @@ fn build_pat(
...
                         iter::repeat(make::wildcard_pat().into()).take(field_list.fields().count());
                     make::tuple_struct_pat(path, pats).into()
                 }
                 ast::StructKind::Record(field_list) => {
-                    let pats = field_list
-                        .fields()
-                        .map(|f| make::ext::simple_ident_pat(f.name().unwrap()).into());
+                    let pats = field_list.fields().map(|f| {
+                        make::ext::simple_ident_pat(
+                            f.name().expect("Record field must have a name"),
```
DropDemBits, line 459: This feels like it could panic in the following situation, but I'm not exactly sure if it will:

```rust
enum A {
    A,
    Missing { a: u32, : u32, c: u32 }
}

fn a() {
    let b = A::A;
    match b$0 {}
}
```

alibektas: Thank you for mentioning this, I checked it and it doesn't. Syntax tree for the missing field name shows : 

```
[crates/ide-assists/src/handlers/add_missing_match_arms.rs:458] &f = RecordField {
    syntax: RECORD_FIELD@40..43
      NAME@40..43
        IDENT@40..43 "u32"
    ,
```
how little the piece is it is still being recognized as NAME and something like `{ a:32 , , c: u32}` completely skips seeing a record field between a and c. So we are good.

Veykril: Would be good to add this as a test case though, we might change recovery in the future which could change this behavior.

PR #15597, opened 2023-09-11
Field shorthand overwritten in promote local to const assist

rmehri01:
Currently, running `promote_local_to_const` on the following:

```rust
struct Foo {
    bar: usize,
}

fn main() {
    let $0bar = 0;
    let foo = Foo { bar };
}
```

Results in:

```rust
struct Foo {
    bar: usize,
}

fn main() {
    const BAR: usize = 0;
    let foo = Foo { BAR };
}
```

But instead should be something like:

```rust
struct Foo {
    bar: usize,
}

fn main() {
    const BAR: usize = 0;
    let foo = Foo { bar: BAR };
}
```


HKalbasi:
Thanks!
@bors r+

PR #15564, opened 2023-09-06
internal: use current folder's `rustfmt.toml` with all rustfmt configurations

davidbarsky:
## Introduction

Resolves https://github.com/rust-lang/rust-analyzer/issues/15540. I moved the `chdir` functionality outside of the `match` to ensure that this functionality wouldn’t fall through again. As part of this PR, I also changed `from_proto::file_range` to accept a `TextDocumentIdentifier` by reference instead of by value, but I can undo this change if desired.

## Testing

I added a `rustfmt.toml` will the contents below at `crates/rust-analyzer/rustfmt.toml`: 

```toml
reorder_modules = false
use_small_heuristics = "Max"
# this is the only difference from the `rustfmt.toml` at the root of the repo
tab_spaces = 8
```

In addition, I've also added `"rust-analyzer.rustfmt.overrideCommand": ["rustfmt"]` to my VS Code configuration.

With the above changes, saving `crates/rust-analyzer/src/handlers/request.rs` results in 8-space indentation. Meanwhile, saving `crates/toolchain/src/lib.rs` _does not_ result in any formatting changes.

PR #15600, opened 2023-09-12
fix: ensure `rustfmt` runs when configured with `./`

davidbarsky:
(Hopefully) resolves https://github.com/rust-lang/rust-analyzer/issues/15595. This change kinda approaches canonicalization—which I am not a fan of—but only in service of making `./`-configured commands run correctly.

Longer-term, I feel like this code should be removed once `rustfmt` supports recursive searches of configuration files or interpolation of values like `${workspace_folder}` lands in rust-analyzer.

## Testing

I cloned `rustc`, setup rust-analyzer as suggested in the [`rustc` dev guide](https://rustc-dev-guide.rust-lang.org/building/suggested.html#configuring-rust-analyzer-for-rustc), saved and formatted files in `src/tools/miri` and `compiler`, and saw `rustfmt` (seemingly) correctly.

### DIFF
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -1997,41 +1997,13 @@ fn run_rustfmt(
         }
         RustfmtConfig::CustomCommand { command, args } => {
             let cmd = PathBuf::from(&command);
-            let mut components = cmd.components();
-
-            // to support rustc's suggested, default configuration
-            let mut cmd = match components.next() {
-                Some(std::path::Component::CurDir) => {
-                    let rest = components.as_path();
-
-                    let roots = snap
-                        .workspaces
-                        .iter()
-                        .flat_map(|ws| ws.workspace_definition_path())
-                        .collect::<Vec<&AbsPath>>();
-
-                    let abs: Option<AbsPathBuf> = roots.into_iter().find_map(|base| {
-                        let abs = base.join(rest);
-                        std::fs::metadata(&abs).ok().map(|_| abs)
-                    });
-
-                    let command = match abs {
-                        Some(cmd) => cmd,
-                        None => {
-                            tracing::error!(
-                                rustfmt = ?command,
-                                "Unable to make the format command an absolute path"
-                            );
-                            anyhow::bail!(
-                                "Unable to make the format command an absolute path: {}",
-                                command
-                            );
-                        }
-                    };
-
-                    process::Command::new(&command.as_os_str())
+            let workspace = CargoTargetSpec::for_file(&snap, file_id)?;
+            let mut cmd = match workspace {
+                Some(spec) => {
+                    let cmd = spec.workspace_root.join(cmd);
+                    process::Command::new(cmd.as_os_str())
                 }
-                _ => process::Command::new(command),
+                None => process::Command::new(cmd),
             };
 
             cmd.envs(snap.config.extra_env());
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -2000,8 +2000,18 @@ fn run_rustfmt(
             let workspace = CargoTargetSpec::for_file(&snap, file_id)?;
             let mut cmd = match workspace {
                 Some(spec) => {
-                    let cmd = spec.workspace_root.join(cmd);
-                    process::Command::new(cmd.as_os_str())
+                    // approach: if the command name contains a path seperator, join it with the workspace root.
+                    // however, if the path is absolute, joining will result in the absolute path being preserved.
+                    // as a fallback, rely on $PATH-based discovery.
+                    let cmd_path =
+                        if cfg!(windows) && command.contains(&[std::path::MAIN_SEPARATOR, '/']) {
+                            spec.workspace_root.join(cmd).into()
+                        } else if command.contains(std::path::MAIN_SEPARATOR) {
+                            spec.workspace_root.join(cmd).into()
+                        } else {
+                            cmd
+                        };
+                    process::Command::new(cmd_path)
                 }
                 None => process::Command::new(cmd),
             };
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -4,6 +4,7 @@
 use std::{
     fs,
     io::Write as _,
+    path::PathBuf,
     process::{self, Stdio},
 };
 
@@ -1995,13 +1996,51 @@ fn run_rustfmt(
             cmd
         }
         RustfmtConfig::CustomCommand { command, args } => {
-            let mut cmd = process::Command::new(command);
+            let cmd = PathBuf::from(&command);
+            let mut components = cmd.components();
+
+            // to support rustc's suggested, default configuration
+            let mut cmd = match components.next() {
+                Some(std::path::Component::CurDir) => {
+                    let rest = components.as_path();
+
+                    let roots = snap
+                        .workspaces
+                        .iter()
+                        .flat_map(|ws| ws.workspace_definition_path())
+                        .collect::<Vec<&AbsPath>>();
+
+                    let abs: Option<AbsPathBuf> = roots.into_iter().find_map(|base| {
+                        let abs = base.join(rest);
+                        std::fs::metadata(&abs).ok().map(|_| abs)
+                    });
+
+                    let command = match abs {
+                        Some(cmd) => cmd,
+                        None => {
+                            tracing::error!(
+                                rustfmt = ?command,
+                                "Unable to make the format command an absolute path"
+                            );
+                            anyhow::bail!(
+                                "Unable to make the format command an absolute path: {}",
+                                command
+                            );
+                        }
+                    };
+
+                    process::Command::new(&command.as_os_str())
+                }
+                _ => process::Command::new(command),
+            };
 
             cmd.envs(snap.config.extra_env());
             cmd.args(args);
             cmd
         }
     };
+
+    tracing::debug!(?command, "created format command");
 
     // try to chdir to the file so we can respect `rustfmt.toml`
     // FIXME: use `rustfmt --config-path` once


davidbarsky:
@RalfJung Sorry for the delay: my initial, simple fix didn't work for simply specifying something like `rustfmt`. Please try this out, maybe?

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -1995,14 +1996,52 @@ fn run_rustfmt(
...
+                    let roots = snap
+                        .workspaces
+                        .iter()
+                        .flat_map(|ws| ws.workspace_definition_path())
+                        .collect::<Vec<&AbsPath>>();
+
+                    let abs: Option<AbsPathBuf> = roots.into_iter().find_map(|base| {
+                        let abs = base.join(rest);
+                        std::fs::metadata(&abs).ok().map(|_| abs)
+                    });
```
davidbarsky, line 2016: I'm not a fan of this approach, but I can't find a better way to determine what workspace a `TextDocument` is a part of.

RalfJung:
I'm afraid I won't be able to try this, I don't have the setup to build RA myself.

Why isn't this as simple as this?
```rust
let command = self.current_dir().unwrap().join(command);
```
Before https://github.com/rust-lang/rust-analyzer/pull/15564, `command` was implicitly resolved in the `current_dir`, now we just need to do that explicitly.

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -1995,14 +1996,52 @@ fn run_rustfmt(
             cmd
         }
         RustfmtConfig::CustomCommand { command, args } => {
-            let mut cmd = process::Command::new(command);
+            let cmd = PathBuf::from(&command);
+            let mut components = cmd.components();
+
+            // to support rustc's suggested, default configuration
+            let mut cmd = match components.next() {
+                Some(std::path::Component::CurDir) => {
```
RalfJung, line 2004: This should probably work with `build/host/rustfmt` as well; special-casing the leading `./` sounds very fragile.

crates/rust-analyzer/src/handlers/request.rs
Veykril: Take a look at `CargoTargetSpec::for_file`, that function has to solve the same problem of relating a file to its workspace(s)

Veykril:
> Longer-term, I feel like this code should be removed once `rustfmt` supports recursive searches of configuration files or interpolation of values like `${workspace_folder}` lands in rust-analyzer.

This should already be working for vscode fwiw, we should probably have the default vscode settings in rust-lang/rust make use of these now that you mention that.

And ralf brought up a good point here, should relative paths be relative to the servers working dir, or relative to the workspace root, I'm not sure.

crates/rust-analyzer/src/handlers/request.rs
Veykril: Ye, we can just [Path::join](https://doc.rust-lang.org/std/path/struct.Path.html#method.join) the rustfmt path onto the base here. If the path is absolute it'll replace the base, otherwise it"ll append which is exactly the semantics we want to have here.

davidbarsky:
Appreciate the feedback, y'all.

> And ralf brought up a good point here, should relative paths be relative to the servers working dir, or relative to the workspace root. I think workspace does make more sense though, as the working dir of the server could be anywhere in theory.

I'm going make to the path relative to the workspace root because I've seen the server be in _weird_ spots.

davidbarsky:
Sorry for the delay in responding/further elaborating, folks. Had a muscle spasm in my neck earlier and it's left me a little incapacitated. I'm substantially better now, though.

I'll address the comments in order and expand on details where I should've expanded earlier.

@RalfJung:
> Why isn't this as simple as this?
> 
> ```rust
> let command = self.current_dir().unwrap().join(command);
> ```
> Before https://github.com/rust-lang/rust-analyzer/pull/15564, `command` was implicitly resolved in the `current_dir`, now we just need to do that explicitly.

Prior to #15564, `current_dir` was whatever workspace the user has currently opened in their editor. The fact that `./` happened to work is, in my view, a happy and convenient accident.  The trouble that emerges is that rust-analyzer-the-server can have multiple interpretations of what "the current workspace" is:
- It can be whatever `ProjectWorkspace::workspace_definition_path` corresponds to, of which there can be multiple `ProjectWorkspace`s, so it might not be clear as to _what_ workspace `./` is referring to.
    - This is not something that I can reasonably tackle in this issue.
- The parent of `CargoTargetSpec::for_file` might not necessarily be the workspace root. That might refer to, in Miri's case, `src/tools/miri`, not the the repository root, which the existing, suggested configuration would prefer.
    - This becomes an issue when I saved `src/tools/miri/src/machine.rs` and used `CargoTargetSpec::for_file` to determine the owning workspace and joined `./build/host/rustfmt/bin/rustfmt`. In my case, I got something along the lines of `/Users/dbarsky/Developer/rust/src/tools/miri/Cargo.toml/./build/host/rustfmt/rustfmt`, which does not exist.

I bring up these complications not to filibuster a solution, but rather, outline my thoughts, which are not settled. I'd like to fix this before Monday.

I'd _personally_ prefer to avoid the solution that entails `joining` whatever custom command was passed with any of the paths, as that might foreclose an option to use `$PATH`-discovered commands in the future [^1]. In that case, I see several paths moving forward:
- Special-case the leading `./`, which is not ideal and fragile. This is the existing PR, but given the feedback, that doesn't appear to be desired.
- Ensure that `build/host/rustfmt` also works. I'm decently sure that I can make this work without removing the possibility of `$PATH`-discovery in the implementation by running `std::fs::metadata(path)` on the joined path, but it seems unfortunate to do that each time the file gets formatted. I suppose I can do that _once_ during the language server's initialization phase.
- Change the rustc dev guide to make use of `${workspacePath}` instead of `./`. Optionally, I can add errors to rust-analyzer informing users of the risks of `./`, but I don't know if users would find this to be bothersome.
- Add a `$PATH`-based interpolation key to the extension, but opting-out seems to be a bit trickier than _opting in_ with workspace-relative paths.

Anyways, I can do any of the above options, just let me know what you prefer.

[^1]: At work, I've setup rust-analyzer to point a `${userHome}`-rooted path in the monorepo (thanks for that fix, Lukas!), but I've recently learned that the more correct approach would be to delegate to [arcanist](https://secure.phabricator.com/book/phabricator/article/arcanist/) in the form of `arc format`. This would allow the team that manages the Rust toolchain at work to move things around not worry about breaking folks.

RalfJung:
> I'd personally prefer to avoid the solution that entails joining whatever custom command was passed with any of the paths, as that might foreclose an option to use $PATH-discovered commands in the future

Ah, that's fair. My goal was to bring the behavior back to what it was before https://github.com/rust-lang/rust-analyzer/pull/15564, but I assume PATH-discovered commands worked then.

Sadly it doesn't seem like `Command` supports resolving the binary in a different path than what it is executed in.

However, applying shell logic, what should work is:
- if the command name contains no path seperator, use it as-is and rely on PATH discovery
- otherwise, `join` it from the current working directory

I appreciate that the current working directory is arbitrary and might not match user intuition, but it matches the behavior before #15564, and the primary priority right now should be to fix the regression. Getting a proper idea of which path to base this on can wait, IMO.

Veykril:
Taking a look at this in a bit, but setting the path to `"${workspaceFolder}/build/host/rustfmt/bin/rustfmt"` should yield the previous behavior for the time being as well.

RalfJung:
Oh, that already works in currently released RA? I can try that.

Veykril:
Well, technically speaking this is a VSCode feature that we re-implemented in the extension for a very few selected ones, https://code.visualstudio.com/docs/editor/variables-reference
(so this only applies to VSCode)

Veykril:
> - The parent of `CargoTargetSpec::for_file` might not necessarily be the workspace root. That might refer to, in Miri's case, `src/tools/miri`, not the the repository root, which the existing, suggested configuration would prefer.
    - This becomes an issue when I saved `src/tools/miri/src/machine.rs` and used `CargoTargetSpec::for_file` to determine the owning workspace and joined `./build/host/rustfmt/bin/rustfmt`. In my case, I got something along the lines of `/Users/dbarsky/Developer/rust/src/tools/miri/Cargo.toml/./build/host/rustfmt/rustfmt`, which does not exist.

Huh that is odd, miri is a member of the main workspace so that should work here just fine? 

> However, applying shell logic, what should work is:
> - if the command name contains no path seperator, use it as-is and rely on PATH discovery
> - otherwise, `join` it from the current working directory

That sounds like a good idea to me (replacing working directory with workspace root).

davidbarsky:
> That sounds like a good idea to me (replacing working directory with workspace root).

Will take that approach, then!

davidbarsky:
Whoops, sorry: I thought I had pushed this a while ago, but I updated this PR according to feedback. It turns out that I was using `CargoTargetSpec::cargo_toml` instead of `CargoTargetSpec::workspace_root`, which _does_ find the `rustc` workspace root.

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -1995,14 +1996,24 @@ fn run_rustfmt(
...
-            let mut cmd = process::Command::new(command);
+            let cmd = PathBuf::from(&command);
+            let workspace = CargoTargetSpec::for_file(&snap, file_id)?;
+            let mut cmd = match workspace {
+                Some(spec) => {
+                    let cmd = spec.workspace_root.join(cmd);
+                    process::Command::new(cmd.as_os_str())
+                }
+                None => process::Command::new(cmd),
+            };
```
Veykril, line 2007: This is treating stuff like `rustfmt` as relative paths still if I see this right? Iiirc from the discussion in that case we want to treat it as a plain command though (such that `PATH` lookup works)

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -1995,14 +1996,31 @@ fn run_rustfmt(
...
+            let mut cmd = match workspace {
+                Some(spec) => {
+                    // approach: if the command name contains a path seperator, join it with the workspace root.
+                    // however, if the path is absolute, joining will result in the absolute path being preserved.
+                    // as a fallback, rely on $PATH-based discovery.
+                    let cmd_path = if command.contains(std::path::MAIN_SEPARATOR_STR) {
+                        spec.workspace_root.join(cmd).into()
+                    } else {
+                        cmd
+                    };
```
Veykril, line 2010:
```suggestion
                    let cmd_path = if cfg!(windows) && command.contains(&[std::path::MAIN_SEPARATOR_STR, '/']) {
                        spec.workspace_root.join(cmd).into()
                    } else if command.contains(std::path::MAIN_SEPARATOR_STR) {
                        spec.workspace_root.join(cmd).into()
                    } else {
                        cmd
                    };
```
Since windows can use either separators.

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -1995,14 +1996,33 @@ fn run_rustfmt(
...
+                    // approach: if the command name contains a path seperator, join it with the workspace root.
+                    // however, if the path is absolute, joining will result in the absolute path being preserved.
+                    // as a fallback, rely on $PATH-based discovery.
+                    let cmd_path = if cfg!(windows) && command.contains(&[std::path::MAIN_SEPARATOR_STR, '/']) {
+                        spec.workspace_root.join(cmd).into()
+                    } else if command.contains(std::path::MAIN_SEPARATOR_STR) {
+                        spec.workspace_root.join(cmd).into()
+                    } else {
+                        cmd
+                    };
```
Veykril, line 2012:
```suggestion
                    let cmd_path = if cfg!(windows) && command.contains(&[std::path::MAIN_SEPARATOR, '/']) {
                        spec.workspace_root.join(cmd).into()
                    } else if command.contains(std::path::MAIN_SEPARATOR) {
                        spec.workspace_root.join(cmd).into()
                    } else {
                        cmd
                    };
```

PR #15609, opened 2023-09-13
internal: Remove most of the duplication from `Semantics{,Impl}` via deref

WaffleLapkin:
See https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/Semantics.20vs.20SemanticsImpl/near/390795952


Veykril:
Thanks!
@bors r+

PR #15615, opened 2023-09-15
Refactor/fix clippy lints

shogo-nakano-desu:
As title says.

### DIFF
--- a/crates/flycheck/src/lib.rs
+++ b/crates/flycheck/src/lib.rs
@@ -488,7 +488,9 @@ impl CargoActor {
                     // Skip certain kinds of messages to only spend time on what's useful
                     JsonMessage::Cargo(message) => match message {
                         cargo_metadata::Message::CompilerArtifact(artifact) if !artifact.fresh => {
-                            self.sender.send(CargoMessage::CompilerArtifact(artifact)).unwrap();
+                            self.sender
+                                .send(CargoMessage::CompilerArtifact(Box::new(artifact)))
+                                .unwrap();
                         }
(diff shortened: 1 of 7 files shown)


crates/flycheck/src/lib.rs
```diff
@@ -533,7 +535,7 @@ impl CargoActor {
 }
 
 enum CargoMessage {
-    CompilerArtifact(cargo_metadata::Artifact),
+    CompilerArtifact(Box<cargo_metadata::Artifact>),
```
lnicola, line 538: Last time this came up (can't find the PR, though), it was pointed out that this is the most common variant, so boxing the contents will cause almost every instance to incur an allocation.

I don't know if this overhead will be larger than the speed-up obtained from the faster moves. Either way, I suspect it doesn't matter too much compared to the whole `cargo check` invocation.

lnicola (approved):

crates/flycheck/src/lib.rs
shogo-nakano-desu: Ah, I see. I will keep it in my mind. Thank you for your review.

shogo-nakano-desu: @lnicola I have removed Box in
[dd84306](https://github.com/rust-lang/rust-analyzer/pull/15615/commits/dd843060f9ebff3862dc48c033c3ddbabb3b6a84) . Thanks

Veykril:
Thanks!
@bors r+
On that note, I believe the lints section in Cargo.toml has been set for stabilization now, https://github.com/rust-lang/cargo/issues/12115 so we can soon enable clippy in general.

Issue #15569, opened 2023-09-07
Revive the librarification effort


Veykril:
Our fork of `rustc_abi` is quite outdated by now, and unfortunately upstream broke the non nightly build of it so we can't update it anymore without putting in significant effort into fixing that. This will most likely be a recurring thing so maybe it is time to revive the librarification effort (and potentially crate a working group for that) to lay out a proper plan for such crates.

Obviously not the best place to have this discussion right now, but putting it down here as a reminder (to myself at least).

HKalbasi:
First thing to do is adding the toggle mentioned in #14846 to switch to the workspace dependencies for rustc crate when r-a is being built inside of the rustc tree, so things like https://github.com/rust-lang/rust/pull/107163 will be detected in the rustc CI. For `rustc_lexer`, `rustc_parse_format`, `rustc_index`, ... it has no problem since rust is backward compatible, and for `rustc_abi` (which we can't toggle it right now anyway) it means we are in sync with latest nightly instead of latest stable, which is not ideal, but better than the current state where we are several stable release back.

I'm all for librarification, in any shape. It seems like out of tree librarification (like chalk and polonious) made things hard for rustc developers, and in tree librarification (like rustc_lexer) is the way to go, which that toggle can help with it.

About `rustc_abi`, ideal librarification of it is probably blocked on the librarification of type system, but it doesn't mean that we can't improve the current state.

Veykril:
With toggle you mean the `cfg` right?

Having them synced properly is orthogonal to this issue. It's mainly about the crates being usable on a stable context such other consumers (which r-a is a prime example of) can use them.
These would stay in tree I agree, there is not much point in having them be lifted out, likewise we don't need them to have a stable API as we can easily work with API breakage churn. My point being we should make it as easy as possible for the rustc maintainers here as this is a request coming from us and it will make their life a bit more annoying no matter what, so lets keep that to a minimum if possible.

Issue #14846, opened 2023-05-18
How to sync rustc_abi with the "current" stable release


Veykril:
We now have support for calculating layouts of types which is a great thing to have, but there is a problem here. Type layouts are not stable so ideally the latest r-a we offer should work with the latest stable's type layout while the rustup releases should work with the corresponding rustc_abi crate there.

Currently we just try to do the former and the rustup releases are bound to just be out of sync with the toolchains actual layout (unless nothing changed over a version).

This is an interesting problem to have, and an important one to fix as with const eval being able to make use of this via `mem::size_of`` and friends, users will run into diagnostics due to mismatches and more. How can we solve this?

A simple somewhat working approach would be to just have current master always work with the latest stable version of the layout crate and patching upstream whenever a beta branch is done to use the upstream crate there. That will at least give us correct stable version matches, but ideally we would handle all releases (nightly as well) properly. That is somewhat tricky to do though since the crate has no stable API. Maybe we can abstract things so that we can have a feature toggle whether to use the upstream crate or the crates.io clone?

HKalbasi:
> Maybe we can abstract things so that we can have a feature toggle whether to use the upstream crate or the crates.io clone?

If it was only size, it was easy. But we are currently using every small details of `Layout` in the mir interpreter to put the bytes in memory. So we will inevitably break if rustc decide to change layout radically (for example by adding a new encoding similar to niche encoding). But that kind of breakage is probably very low so maybe we can do that in a best effort basis (being compatible with both stable and nightly with a switch until a breakage happen, and then drop one of them until they become sync again).

An extreme solution is dropping our (really awesome) weekly release model, and making r-a a "normal" rust component, with normal nightly, beta and stable releases, shipped by rustup, ... . This will resolve the need for auto published crates, and saves some trouble in syncs with rustc repo.

Another extreme solution is declaring that `r-a` is a different rust compiler, and the layout of types is deferred to the implementation by the rust language, so whatever we show as the layout size is correct. We will try to keep it in sync with rustc stable, but users should add `#[cfg(not(ra))]` to static asserts that are sensitive to layout info. (or maybe `#[cfg(rustc)]`, to make code also compatible with `gccrs` and other future rust compilers).

flodiebold:
> But that kind of breakage is probably very low so maybe we can do that in a best effort basis (being compatible with both stable and nightly with a switch until a breakage happen, and then drop one of them until they become sync again).

[...]

PR #15616, opened 2023-09-15
Switch to in-tree rustc dependencies with a cfg flag

HKalbasi:
We can use this flag to detect and prevent breakages in rustc CI. (see #14846 and #15569)

~The `IN_RUSTC_REPOSITORY` is just a placeholder. Is there any existing cfg flag that rustc CI sets?~

### DIFF
--- a/crates/hir-def/src/lib.rs
+++ b/crates/hir-def/src/lib.rs
@@ -8,6 +8,7 @@
 //! actually true.
 
 #![warn(rust_2018_idioms, unused_lifetimes, semicolon_in_expressions_from_macros)]
+#![cfg_attr(feature = "in-rust-tree", feature(rustc_private))]
 
 #[allow(unused)]
 macro_rules! eprintln {
@@ -48,7 +49,7 @@ pub mod visibility;
 pub mod find_path;
 pub mod import_map;
 
-pub use rustc_abi as layout;
+pub use rustc_dependencies::abi as layout;
 use triomphe::Arc;
 
 #[cfg(test)]
--- a/crates/parser/src/lib.rs
+++ b/crates/parser/src/lib.rs
@@ -19,6 +19,7 @@
 
 #![warn(rust_2018_idioms, unused_lifetimes, semicolon_in_expressions_from_macros)]
 #![allow(rustdoc::private_intra_doc_links)]
+#![cfg_attr(feature = "in-rust-tree", feature(rustc_private))]
 
 mod lexed_str;
 mod token_set;
--- a/crates/rust-analyzer/Cargo.toml
+++ b/crates/rust-analyzer/Cargo.toml
@@ -57,6 +57,7 @@ flycheck.workspace = true
 hir-def.workspace = true
 hir-ty.workspace = true
 hir.workspace = true
+rustc-dependencies.workspace = true
 ide-db.workspace = true
 # This should only be used in CLI
 ide-ssr.workspace = true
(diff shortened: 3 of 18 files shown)


lnicola:
> Is there any existing cfg flag that rustc CI sets?

It enables the `in-rust-tree` feature, IIUC.

HKalbasi:
This will probably make syncs harder, but I think it does worth it. Feel free to revert if it ended up breaking something or caused too much trouble.

@bors r+

lnicola:
> This will probably make syncs harder, but I think it does worth it.

I hope not :grimacing:.

Veykril:
Not a fan of the implications this has for our dependency graph. This is probably fine right now but won't scale in the future.

crates/parser/src/lib.rs
```diff
@@ -19,6 +19,7 @@
 
 #![warn(rust_2018_idioms, unused_lifetimes, semicolon_in_expressions_from_macros)]
 #![allow(rustdoc::private_intra_doc_links)]
+#![cfg_attr(feature = "in-rust-tree", feature(rustc_private))]
```
Veykril, line 22: why is this required?

crates/rust-analyzer/Cargo.toml
```diff
@@ -89,4 +91,12 @@ mbe.workspace = true
...
-in-rust-tree = ["sysroot-abi", "ide/in-rust-tree", "syntax/in-rust-tree"]
+in-rust-tree = [
+    "sysroot-abi",
+    "ide/in-rust-tree",
+    "syntax/in-rust-tree",
+    "parser/in-rust-tree",
+    "rustc-dependencies/in-rust-tree",
+    "hir-def/in-rust-tree",
+    "hir-ty/in-rust-tree",
+]
```
Veykril, line 102: Not a fan of this either, this will ultimately end up with mentioning almost all of our crates at some point

crates/hir-def/src/lib.rs
```diff
@@ -8,6 +8,7 @@
 //! actually true.
 
 #![warn(rust_2018_idioms, unused_lifetimes, semicolon_in_expressions_from_macros)]
+#![cfg_attr(feature = "in-rust-tree", feature(rustc_private))]
```
Veykril, line 11: Why is this needed?

crates/syntax/src/lib.rs
```diff
@@ -19,6 +19,7 @@
 //! [RFC]: <https://github.com/rust-lang/rfcs/pull/2256>
 //! [Swift]: <https://github.com/apple/swift/blob/13d593df6f359d0cb2fc81cfaac273297c539455/lib/Syntax/README.md>
 
+#![cfg_attr(feature = "in-rust-tree", feature(rustc_private))]
```
Veykril, line 22: same here

Veykril (commented):
While this is probably a solution, I have to say that I really don't like these changes 😅

crates/hir-def/src/lib.rs
HKalbasi: This feature is needed even if you import the rustc private crates indirectly, which I think is unnecessarily too aggressive. Is there any way to depend on the crates in the rustc workspace without this feature? Clippy and Miri use this, but they are fine with nightly and feature flags.

crates/rust-analyzer/Cargo.toml
HKalbasi: Me too, maybe a non cargo cfg is a better fit here?

Veykril: Yes, I think it'd be better if we could make it so that this doesn't thread through all the cargo files if possible

crates/hir-def/src/lib.rs
Veykril: Ah right, since we might mention the items from those crates now. Well that's unfortunate but not much we can do in that case. Would be nice  to add a comment regarding that then.

PR #15618, opened 2023-09-16
internal: port anymap

shogo-nakano-desu:
## Description
- The anymap crate has been ported. During this process, unnecessary features for rust-analyzer have been removed.
- From the tests that were checking the existing licenses, the anymap license (`BlueOak-1.0.0 OR MIT OR Apache-2.0`) has been removed.

## Requests
- While porting the code this time, I have tried to respect the original author's intentions and have kept the comments/codes as much as possible. Please don't hesitate to tell me if you think the comments/codes also need to be appropriately modified.
- If there are any necessary changes regarding the licensing or anything else, please let me know so I can fix them.

## Issue
https://github.com/rust-lang/rust-analyzer/issues/15500


crates/anymap/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    #[inline]
+    fn finish(&self) -> u64 {
+        self.value
+    }
+}
+
+#[cfg(any(feature = "std", feature = "hashbrown"))]
+macro_rules! everything {
```
Veykril, line 67: We can simplify all of this a bit more, lets inline the macro call, that is just replace it with its expansion and remove the feature flags, basically enabling them unconditionally, as well

Veykril: There is probably a lot more we can strip out of this after that as well, at which point we might as well just move this in to the `stdx` crate as a standalone module instead of puttin this into a separate crate.

crates/anymap/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "anymap"
+version = "0.0.0"
+description = "This crate is a port of only the necessary features from https://github.com/chris-morgan/anymap for use within rust-analyzer. Copyright © 2014–2022 Chris Morgan. COPYING: https://github.com/chris-morgan/anymap/blob/master/COPYING"
```
lnicola, line 4: We should probably mention the version here, especially given the license change.

crates/anymap/src/lib.rs
shogo-nakano-desu: Thank you for your comment, OK I will work on them.

crates/anymap/Cargo.toml
shogo-nakano-desu: Just to confirm, as Veykril [commented](https://github.com/rust-lang/rust-analyzer/pull/15618#discussion_r1330541952), if we move the implementation of `anymap` to the `stdx` crate, we need to modify the Cargo.toml of `stdx` crate, and in the description, we need to mention the version of `anymap` we ported and the change of the license from `Blue Oak Model 1.0.0 or MIT or Apache-2.0` to `MIT OR Apache-2.0` as in rust-analyzer. Is my understanding correct?

shogo-nakano-desu: I changed my mind and I have added comments on a file in this commit. If it is better to move these comments to somowhere else, please let me know:)
[2b891ca](https://github.com/rust-lang/rust-analyzer/pull/15618/commits/2b891ca084f38aa88842cd8c568913adbf0e88a2)

crates/anymap/src/lib.rs
shogo-nakano-desu: I have moved the code into stdx crate in this commit. Can you review again please?
[f671b0b](https://github.com/rust-lang/rust-analyzer/pull/15618/commits/f671b0b8644aff682616025af5e1a1780ffab0c3)

Cargo.toml
```diff
@@ -103,5 +103,12 @@ serde_json = "1.0.96"
...
-    "inline-more",
+  "inline-more",
 ], default-features = false }
+
+rustc_lexer = { version = "0.10.0", package = "ra-ap-rustc_lexer" }
+rustc_parse_format = { version = "0.10.0", package = "ra-ap-rustc_parse_format", default-features = false }
+
+# Upstream broke this for us so we can't update it
+rustc_abi = { version = "0.0.20221221", package = "hkalbasi-rustc-ap-rustc_abi", default-features = false }
+rustc_index = { version = "0.0.20221221", package = "hkalbasi-rustc-ap-rustc_index", default-features = false }
```
Veykril, line 114: I think you messed up your rebase?

Veykril (approved):

Cargo.toml
shogo-nakano-desu: Oops thank you for pointing it out. It was removed in this [PR](https://github.com/rust-lang/rust-analyzer/commit/f4704bc8ae92d150b2801e391692d9503f0c6126#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L104-R107) so removed in [1e11a55](https://github.com/rust-lang/rust-analyzer/pull/15618/commits/1e11a55f98d9b3f1f7ac2a6779e10b060fe3c896).

shogo-nakano-desu:
I have solved all the comments. Thank you for your reviews.

Veykril (approved):

PR #15621, opened 2023-09-16
Give `unmerge_use` a label explaining what it will affect.

kpreid:
When I'm trying to clean up `use`s, I often feel uncertain about what exactly the effects of choosing an assist will be. This PR makes a small improvement to that by giving “Unmerge use” a label which names the root of the tree that it's going to move, when one exists.

There is no test because I didn't see, among the test helpers, a way to assert on the assist label (as opposed to filtering on it). However, I did test the change manually.

I looked into making a similar change to “Merge imports”, but that is considerably trickier.


Veykril:
The merge one works quite a bit differently indeed. Maybe we can just state there if we are merging the one the cursor is on with the one below. As in `Merge import with the following one` or something clearer.

Veykril (approved):

PR #15625, opened 2023-09-17
fix: Don't skip closure captures after let-else

### LINKED ISSUE #15623: False positive "variable does not need to be mutable"
In this code:

```rust
pub struct Foo;

impl Foo {
    fn needs_mut(&mut self) {}
}

pub fn foo(mut foo: Foo) {
    let mut call_me = || {
        let 0 = 1 else { return };
        foo.needs_mut();
    };
    call_me();
}
```

Rust analyzer shows 2 "variable deos not need to be mutable" warnings:

image

image

The same code compiles with rustc without any warnings or errors. Removing either `mut` makes rust analyzer stop reporting that warning and does not introduce any other diagnostics from ra, but then rustc reports that mut is needed there.

**rust-analyzer version**: 0.3.1657-standalone (326f37ef1 2023-09-10)

**rustc version**: rustc 1.71.0 (8ede3aae2 2023-07-12)

**relevant settings**:

```
"rust-analyzer.checkOnSave": false,
"rust-analyzer.completion.postfix.enable": false,
"rust-analyzer.inlayHints.chainingHints.enable": false,
"rust-analyzer.inlayHints.closingBraceHints.enable": false,
"rust-analyzer.inlayHints.parameterHints.enable": false,
"rust-analyzer.inlayHints.typeHints.enable": false,
"rust-analyzer.lens.enable": false,
"rust-analyzer.procMacro.enable": true,
"rust-analyzer.diagnostics.experimental.enable": true,
```

jDomantas:
As I understand that `return` was left there by accident. It caused capture analysis to skip the rest of the block after a let-else, and then missed captures caused incorrect results in borrowck, closure hints, layout calculation, etc.

Fixes #15623

I didn't understand why I using the example from #15623 as-is doesn't work - I don't get the warnings unless I remove the `call_me()` call, even on the same commit as my own RA version which does show those warnings.

### DIFF
--- a/crates/hir-ty/src/infer/closure.rs
+++ b/crates/hir-ty/src/infer/closure.rs
@@ -469,13 +469,13 @@ impl InferenceContext<'_> {
                         Statement::Let { pat, type_ref: _, initializer, else_branch } => {
                             if let Some(else_branch) = else_branch {
                                 self.consume_expr(*else_branch);
-                                if let Some(initializer) = initializer {
-                                    self.consume_expr(*initializer);
-                                }
-                                return;
                             }
                             if let Some(initializer) = initializer {
-                                self.walk_expr(*initializer);
+                                if else_branch.is_some() {
+                                    self.consume_expr(*initializer);
+                                } else {
+                                    self.walk_expr(*initializer);
+                                }
                                 if let Some(place) = self.place_of_expr(*initializer) {
                                     self.consume_with_pat(place, *pat);
                                 }
--- a/crates/hir-ty/src/layout/tests/closure.rs
+++ b/crates/hir-ty/src/layout/tests/closure.rs
@@ -255,3 +255,17 @@ fn ellipsis_pattern() {
         }
     }
 }
+
+#[test]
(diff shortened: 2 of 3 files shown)


HKalbasi:
Thanks! Can you also add a test for captures themselves? Something like this:
```Rust
let a = 2;
let b = 3;
let c = 5;
|| {
  let x = a else { return b; };
  let y = c;
};
```
The closure should capture all variables. You can (ab)use hover tests as IIRC there is no infrastructure for testing closure captures directly.

> I didn't understand why I using the example from https://github.com/rust-lang/rust-analyzer/issues/15623 as-is doesn't work - I don't get the warnings unless I remove the call_me() call, even on the same commit as my own RA version which does show those warnings.

You probably need to add minicore flags needed for `Fn` traits.

jDomantas:
I added an indirect test for captures via layout test, if that's not ok I can look into redoing it with a hover test a bit later.

HKalbasi:
That works as well. We probably need dedicated capture tests in future.
@bors r+

Issue #15384, opened 2023-08-03
Import suggestions within a function call should use function argument types to prioritize suggestions


Lokathor:
Say we've got some AtomicUsize, and we call `fetch_add` with `1` and `Ordering` (not yet imported), and then press Ctrl+Space at the `[HERE]` position

```rust
NEXT_STR_ID.fetch_add(1, Ordering[HERE])
```

Currently, it suggests `core::cmp::Ordering` as the top result, and then `core::sync::atomic::Ordering` after that. While the `core::cmp::Ordering` is perhaps used more commonly overall, or it just sorts earlier alphabetically, since the argument type of the second argument of `fetch_add` actually matches one of the suggestions already in the list (the atomic ordering type), that suggestion should get top priority.

Veykril:
I suppose we aren't actually prioritizing by type for importing completions yet

jmintb:
Can I work on this? :)

Veykril:
Go ahead!

lnicola:
The quick fix still suggests `std::cmp::Ordering`, do we leave this open or file a new one?

Veykril:
Does it still not prioritize the right one? Then we can leave this open

lnicola:
![image](https://github.com/rust-lang/rust-analyzer/assets/308347/b531d92e-a938-4066-bfa0-4c5e4cbc74f6)

Veykril:
Oh I missed that you meant the quickfix, not the completion. The quickfix should probably get a separate issue, as it currently doesn't do any priority checks I think?

PR #15627, opened 2023-09-18
Draft: Prioritize import suggestions based on the expected type

jmintb:
Hi, this is a draft PR to solve #15384. `Adt` types work and now I have a few questions :)

1. What other types make sense in this context? Looking at [ModuleDef](https://github.com/rust-lang/rust-analyzer/blob/05666441bafd6010787a4097a6bd44266ad21018/crates/hir/src/lib.rs#L275) I am thinking everything except Modules.
2. Is there an existing way of converting between `ModeuleDef` and `hir::Type` in the rustanalyzer code base?
3. Does this approach seem sound to you?

Ups: Upon writing this I just realised that the enum test is invalided as there are no enum variants and this no variant is passed as a function argument.


crates/ide-completion/src/completions/flyimport.rs
```diff
@@ -377,13 +396,29 @@ fn import_assets_for_path(
...
+
     if fuzzy_name_length < 3 {
         cov_mark::hit!(flyimport_exact_on_short_path);
         assets_for_path.path_fuzzy_name_to_exact(false);
     }
+
     Some(assets_for_path)
 }
 
+fn is_expected_type(ctx: &CompletionContext<'_>, located_import: &LocatedImport) -> bool {
```
Veykril, line 408: This is the wrong way to approach it, our `CompletionItem` structure has a `type_match` field which is what we need to populate here. According to that, we are later going to sort the completions already. The `render_resolution_with_import(RenderContext::new(ctx), path_ctx, import)` call already does this for locals being completed in `render_resolution_path`. Now, the reason this therefore doesn't work here in flyimport already is that we are completin importable things which are never locals! See
https://github.com/Veykril/rust-analyzer/blob/994df3d6a31d39f11600f30a6df0b744b13937c1/crates/ide-completion/src/render.rs#L337-L351

We should instead generalize that part I believe, making it work for more scopedefs, mainly
`ScopeDef::ImplSelfType`, `ScopeDef::AdtSelfType` and reklevant `ScopeDef::ModuleDef`. For `ModuleDef`, all 3 `Adt`s, `Function`, `Variant`, `Const`, `Static` and `BuiltinType` are relevant. (Some of those might not have a function allowing you to easily turn them into a hir version yet. It's fine to skip them in that case (or (ask about how) to implement them).

jmintb: Awesome that makes a lot more sense haha.

crates/ide-completion/src/render.rs
```diff
@@ -349,6 +349,23 @@ fn render_resolution_path(
 
         path_ref_match(completion, path_ctx, &ty, &mut item);
     };
+
+    if let ScopeDef::ModuleDef(ModuleDef::Adt(adt)) = resolution {
```
jmintb, line 353: @Veykril is this the sort of thing you had in mind?

Veykril: Yes, that should be what I had in mind

crates/ide-completion/src/completions/flyimport.rs
jmintb: I used `UPDATE_EXPECT=1` to see what the new code produces. It looks more or less correct to me. I am not sure why there are two completion items though. Is it because we can choose to either import or use the fully qualified name? Also should `type_match: `Some(CouldUnify)` actually be  `type_match: `Some(Exact)`? as this is the exact type we are expecting.
```rust   
// crates/ide-completion/src/render.rs
 #[test]
    fn sort_suggested_structs_by_types() {
        check(
            r#"
//- /lib.rs crate:dep

pub mod test_mod_b {
    pub struct Struct {}
}

//- /main.rs crate:main deps:dep

fn test(input: dep::test_mod_b::Struct) { }

fn main() {
    test(Struct$0);
}
"#,
            SymbolKind::Struct,
            expect![[r#"
                [
                    CompletionItem {
                        label: "Struct (use dep::test_mod_b::Struct)",
                        source_range: 67..73,
                        delete: 67..73,
                        insert: "Struct",
                        kind: SymbolKind(
                            Struct,
                        ),
                        lookup: "Struct",
                        detail: "Struct",
                        relevance: CompletionRelevance {
                            exact_name_match: false,
                            type_match: Some(
                                CouldUnify,
                            ),
                            is_local: false,
                            is_item_from_trait: false,
                            is_name_already_imported: false,
                            requires_import: false,
                            is_op_method: false,
                            is_private_editable: false,
                            postfix_match: None,
                            is_definite: false,
                        },
                    },
                    CompletionItem {
                        label: "dep::test_mod_b::Struct {…}",
                        source_range: 67..73,
                        delete: 67..73,
                        insert: "dep::test_mod_b::Struct {  }$0",
                        kind: SymbolKind(
                            Struct,
                        ),
                        lookup: "test_mod_b::Struct{}",
                        detail: "dep::test_mod_b::Struct {  }",
                        relevance: CompletionRelevance {
                            exact_name_match: false,
                            type_match: Some(
                                CouldUnify,
                            ),
                            is_local: false,
                            is_item_from_trait: false,
                            is_name_already_imported: false,
                            requires_import: false,
                            is_op_method: false,
                            is_private_editable: false,
                            postfix_match: None,
                            is_definite: false,
                        },
                        trigger_call_info: true,
                    },
                ]
            "#]],
        );
    }
```

Veykril: The qualified completion is from another code path, you can ignore that one. Exact doesn't work in this case, since imported things might have generics that we aren't filling out when constructing their type (since we don't know the concrete generics for them).

crates/ide-completion/src/render.rs
```diff
@@ -334,21 +336,45 @@ fn render_resolution_path(
...
+    let mut set_item_relevance = |is_local: bool, ty: Type| {
         if !ty.is_unknown() {
             item.detail(ty.display(db).to_string());
         }
 
         item.set_relevance(CompletionRelevance {
             type_match: compute_type_match(completion, &ty),
             exact_name_match: compute_exact_name_match(completion, &name),
-            is_local: true,
+            is_local,
```
jmintb, line 348: Should the rest of `CompletionRelevance` be configured here? I assume options like `is_name_already_imported` are important.

crates/ide-completion/src/render.rs
```diff
@@ -595,6 +621,206 @@ mod tests {
         }
     }
 
+    #[test]
+    fn set_struct_type_completion_info() {
```
jmintb, line 625: What do you think of the tests? I will add similar tests for the remaining cases as well if these look good.

crates/ide-completion/src/render.rs
```diff
@@ -334,21 +336,45 @@ fn render_resolution_path(
...
+        }
+        ScopeDef::ModuleDef(ModuleDef::Function(func)) => set_item_relevance(false, func.ty(db)),
+        ScopeDef::ModuleDef(ModuleDef::Variant(variant)) => {
+            set_item_relevance(false, variant.parent_enum(db).ty(db))
+        }
+        ScopeDef::ModuleDef(ModuleDef::Const(konst)) => set_item_relevance(false, konst.ty(db)),
+        ScopeDef::ModuleDef(ModuleDef::Static(stat)) => set_item_relevance(false, stat.ty(db)),
+        ScopeDef::ModuleDef(ModuleDef::BuiltinType(bt)) => set_item_relevance(false, bt.ty(db)),
+        ScopeDef::ImplSelfType(imp) => set_item_relevance(true, imp.self_ty(db)),
+        ScopeDef::GenericParam(_)
```
jmintb, line 368: I have made the  ignored cases explicit so we get an error if new cases arise. Not sure if this is the best option I just didn't like the idea of new cases being silently ignored.

crates/ide-completion/src/render.rs
Veykril: `is_name_already_imported ` is only used by `use` item completions, so we can ignore this here as its being set somewhere else already (in this case its overwriting the relevance calculated here)

Veykril: Also instead of passing `is_local` as a param to the closre we can put `is_local: matches!(resolution, ScopeDef::Local(_))` here. That simplifies things a bit. (I noticed you also set it to true for `ImplSelfType` but that property is specifically meant for local variables, the doc comment on that isn't clear on that.)

Veykril: We should set the `requires_import` field for it here as well if `import_to_add` (the param of the function we are in right now is `Some`). The rest we can ignore I think.

crates/ide-completion/src/render.rs
Veykril: Instead of doing mere `check` tests here, let's make use of `check_relevance` instead, as that is what we are mainly interested in checking here (those are used a bit further down in this module)

crates/ide-completion/src/render.rs
```diff
@@ -447,13 +487,54 @@ fn compute_type_match(
...
-    if completion_ty == expected_type {
-        Some(CompletionRelevanceTypeMatch::Exact)
-    } else if expected_type.could_unify_with(ctx.db, completion_ty) {
-        Some(CompletionRelevanceTypeMatch::CouldUnify)
-    } else {
-        None
+    match_types(ctx, expected_type, completion_ty)
+}
+
+fn compute_function_type_match(
```
jmintb, line 493: As far as I could tell `compute_type_match` does not work for functions as each function has a unique type. Instead  `compute_function_type_match` looks at the params and returns type attempting to unify them. There might already be a way to do this but I couldn't find it.  

` compute_function_type_match` function feels more like a hack/workaround to me but it gets the job done. Perhaps expanding the type unification logic to handle callables could make sense in a separate issue.

crates/ide-completion/src/render.rs
```diff
@@ -595,6 +676,335 @@ mod tests {
...
+                en Enum (use dep::test_mod_b::Enum) [type_could_unify+requires_import]
+                fn main() []
+                fn test(…) []
+                md dep []
+                en Enum (use dep::test_mod_a::Enum) [requires_import]
+            "#]],
+        );
+    }
+
+    // TODO: How dowe test ModuleDef::Variant(Variant?)
```
jmintb, line 787: I could not come up with a test that produced a `ModelDef::Variant` suggestion. Do you know what such a case would look like?

crates/ide-completion/src/render/function.rs
```diff
@@ -77,8 +80,9 @@ fn render(
         .as_assoc_item(ctx.db())
         .and_then(|trait_| trait_.containing_trait_or_trait_impl(ctx.db()))
         .map_or(false, |trait_| completion.is_ops_trait(trait_));
+
     item.set_relevance(CompletionRelevance {
-        type_match: compute_type_match(completion, &ret_type),
```
jmintb, line 81: Looks like functions were only being type matched based on the return type. I am not sure if there was a reason for this :thinking:

jmintb:
I ended up going down a small rabbit hole getting the function type matching to work, but it should be good now.

Let me know what you think :)

Veykril:
I went ahead and cleaned up the remaining things, I hope you don't mind, I wanna get our PR backlog down and I imagine you might not have the PR in your head at this point anymore.
Thanks a lot for the work on this!
@bors r+

Veykril:
> I ended up going down a small rabbit hole getting the function type matching to work, but it should be good now.

I removed the signature match function again, as that is the wrong way to approach it (but you are right we need to do this ideally!) I'll open an issue on the matter and fill that out with details later.

jmintb:
Awesome thanks :)

lnicola:
image

PR #15631, opened 2023-09-19
Update language-configuration.json

vaqxai:
Add automatic closing for closure arguments


vaqxai:
Realized that "|" is also an operator, maybe make this togglable like "<>" is? Or find a way to discriminate operator vs closure argument?

PR #15633, opened 2023-09-19
scip: Allow customizing cargo config.

emilio:
Re-use the LSP config json for simplicity.

### DIFF
--- a/crates/rust-analyzer/src/cli/scip.rs
+++ b/crates/rust-analyzer/src/cli/scip.rs
@@ -12,7 +12,6 @@ use ide::{
 };
 use ide_db::LineIndexDatabase;
 use load_cargo::{load_workspace_at, LoadCargoConfig, ProcMacroServerChoice};
-use project_model::{CargoConfig, RustLibSource};
 use scip::types as scip_types;
 
 use crate::{
@@ -24,8 +23,6 @@ impl flags::Scip {
     pub fn run(self) -> anyhow::Result<()> {
         eprintln!("Generating SCIP start...");
         let now = Instant::now();
-        let mut cargo_config = CargoConfig::default();
-        cargo_config.sysroot = Some(RustLibSource::Discover);
 
         let no_progress = &|s| (eprintln!("rust-analyzer: Loading {s}"));
         let load_cargo_config = LoadCargoConfig {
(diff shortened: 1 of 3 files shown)


emilio:
r? @lnicola 

This is useful for use-cases like https://github.com/mozsearch/mozsearch/pull/652 (but also for any custom workspace shenanigans for which the rust-analyzer configuration is useful).

crates/rust-analyzer/src/cli/scip.rs
```diff
@@ -34,6 +32,20 @@ impl flags::Scip {
...
+            /* workspace_roots = */ vec![],
+        );
+
+        if let Some(p) = self.config_path {
+            let mut file = std::io::BufReader::new(std::fs::File::open(p)?);
+            let json = serde_json::from_reader(&mut file)?;
+            config.update(json)?;
+        }
+        let mut cargo_config = config.cargo();
+        cargo_config.sysroot.get_or_insert(RustLibSource::Discover);
```
lnicola, line 48: Isn't this the default?

lnicola:
Regarding https://github.com/mozsearch/mozsearch/pull/652, wouldn't setting `CARGO_TARGET_DIR` when running `rust-analyzer scip` work just as well?

> also for any custom workspace shenanigans for which the rust-analyzer configuration is useful

Yeah, maybe that.

---

I'm just not sure we want to support this kind of configuration. But if we do, I think we want it as a global option. It could also be useful for `rust-analyzer diagnostics` or `analysis-stats`, for example.

crates/rust-analyzer/src/cli/scip.rs
emilio: Yeah I guess it is. I wanted to make sure I didn't change the no-config-path version of this, but now that we go through the more general config mechanism we should be getting the defaults properly.

emilio:
> Regarding [mozsearch/mozsearch#652](https://github.com/mozsearch/mozsearch/pull/652), wouldn't setting `CARGO_TARGET_DIR` when running `rust-analyzer scip` work just as well?

No, it's already set a few lines before calling `rust-analyzer scip`, but...

> > also for any custom workspace shenanigans for which the rust-analyzer configuration is useful
> 
> Yeah, maybe that.

Yeah, in practice, eventually I want to use this for Firefox, which has a much complex workspace / vendoring set-up and requires also `rust-analyzer.cargo.buildScripts.overrideCommand` / `rust-analyzer.check.overrideCommand`.

> I'm just not sure we want to support this kind of configuration. But if we do, I think we want it as a global option. It could also be useful for `rust-analyzer diagnostics` or `analysis-stats`, for example.

I think they could be useful but probably requires quite a bit more work to generalize this.

> And it should probably be marked as unstable, since we I don't want to commit to this approach to configuration yet.

Yeah, making them unstable is fine with me. My understanding is that these are effectively already unstable tho? `rust-analyzer --help` says:

> Subcommands and their flags do not provide any stability guarantees and may be removed or changed without notice. Top-level flags that are not are marked as [Unstable] provide backwards-compatibility and may be relied on.

I'm more than happy to adapt to changes in this area if the way of doing this changes in the future :)

lnicola:
> No, it's already set a few lines before calling rust-analyzer scip, but...

Then it should already work AFAICT. But we only use `rust-analyzer.server.extraEnv` in the VS Code extension. You might want `rust-analyzer.cargo.extraEnv` instead.

Veykril:
This will most likely be superceded by rust-analyzer.toml once we have that

emilio:
> All right, let's keep it only for `skip` for now. No need to mark it as unstable since it's not a top-level argument. But I'd still drop the `cargo_config.sysroot` part.

Thanks! Rebased and removed that sysroot bit since it is indeed redundant. Just confirmed that using scip without config file still gets `Discover` as the sysroot strategy.

@bors r=lnicola

Issue #12613, opened 2022-06-22
Hanging on "Roots scanned" in Neovim but not vscode


jrmoulton:
rust-analyzer ad6810e90 2022-06-08 dev

rustc 1.63.0-nightly (b2eed72a6 2022-05-22)

I'm running rust analyzer inside of neovim and I am getting an issue similar to #8161 where rust analyzer is hanging on "Roots Scanned". This happens inside of neovim but it doesn't happen when I open the project inside of VSCode. 

This also doesn't happen inside of most of my rust projects. It's happening using the [esp-idf-template](https://github.com/esp-rs/esp-idf-template) for an embedded project. 
I've gone through the troubleshooting guide but I haven't found any more useful information. Any other tips on how else can I debug what is going on here?

jrmoulton:
@andoriyu I am on an M1 mac

andoriyu:
> @andoriyu I am on an M1 mac

If you open RA process in activity monitor you will see which files it's scanning. It's not stuck, it's just very slow.


I had to change RA config to explicitly exclude certain directories for it to stop doing that by adding dirs to `rust-analyzer.files.excludeDirs`. 

Folder in my case was `.direnv` and was scanning entire `nixpkgs` repository.

jrmoulton:
Ahhh yeah that is whats happening. Ok I should be able to figure it out then. Thanks!

yanshay:
I face the same issues with Neovim and esp-idf projects. Roots scanned never ends. I couldn't figure out how to solve this. 
Any solutions?

SethFalco:
> I had to change RA config to explicitly exclude certain directories for it to stop doing that by adding dirs to `rust-analyzer.files.excludeDirs`.

I had a similar issue while trying to set up development for [GNOME Fractal](https://gitlab.gnome.org/GNOME/fractal) in Visual Studio Code.

Configuring `rust-analyzer.files.excludeDirs` was the solution for me, which I wrote the following config:

```json
{
  "files.watcherExclude": {
    "**/_build": true,
    "**/.classpath": true,
    "**/.dart_tool": true,
    "**/.factorypath": true,
    "**/.flatpak-builder": true,
    "**/.git/objects/**": true,
    "**/.git/subtree-cache/**": true,
    "**/.idea": true,
    "**/.project": true,
    "**/.scannerwork": true,
    "**/.settings": true,
    "**/.venv": true,
    "**/node_modules": true
  },
[...]

PR #15636, opened 2023-09-19
Show which roots are being scanned in progress messages

9999years:
This changes the `Roots Scanned` message to include the directory being scanned.

Before: `Roots Scanned 206/210 (98%)`
After: `Roots Scanned 206/210: .direnv (98%)`

This makes it a lot easier to tell that `rust-analyzer` isn't crashed, it's just trying to scan a huge directory.

See: #12613

### DIFF
--- a/crates/vfs-notify/src/lib.rs
+++ b/crates/vfs-notify/src/lib.rs
@@ -107,18 +107,12 @@ impl NotifyActor {
                             n_total,
                             n_done: 0,
                             config_version,
-                            file: None,
+                            dir: None,
                         });
 
                         self.watched_entries.clear();
 
                         for (i, entry) in config.load.into_iter().enumerate() {
-                            self.send(loader::Message::Progress {
-                                n_total,
-                                n_done: i,
-                                config_version,
-                                file: None,
-                            });
                             let watch = config.watch.contains(&i);
                             if watch {
                                 self.watched_entries.push(entry.clone());
@@ -127,7 +121,7 @@ impl NotifyActor {
                                 self.load_entry(entry, watch, |file| loader::Message::Progress {
                                     n_total,
                                     n_done: i,
-                                    file: Some(file),
+                                    dir: Some(file),
                                     config_version,
                                 });
                             self.send(loader::Message::Loaded { files });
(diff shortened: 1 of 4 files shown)


crates/vfs-notify/src/lib.rs
```diff
@@ -103,21 +103,39 @@ impl NotifyActor {
...
+                                    n_done: i,
+                                    file: Some(file),
+                                    config_version,
+                                });
                             self.send(loader::Message::Loaded { files });
                             self.send(loader::Message::Progress {
                                 n_total,
                                 n_done: i + 1,
                                 config_version,
+                                file: None,
```
Veykril, line 138: With this we now duplicate all loaded messages which seems unnecessary. Instead of having `load_entry` take a callback that triggers message sending we can just set the file field here no? That is, set it to the dir thats being loaded, and nothing if its a file set. From what I can tell, we don't set entries to files right now anyways (and probably won't use it much in the future either).

I worry especially about sending an update for every single file, that is very imperformant and we have already gotten complains before about r-a being too spammy with update notifications which this would make even worse.

9999years: This only sends updates for top-level directories when the `Directories` variant is used, not every single file. It does send notifications for each file in the `Files` variant, but given those are listed explicitly as a `Vec` I expect them to not be enormous. Am I missing some context here?

9999years: It'd probably be fine to remove the notifications for the `Files` variant if you'd like, though I don't love the inconsistency.

9999years:
PTAL?

crates/vfs-notify/src/lib.rs
Veykril: Sorry for the long wait, I'm still of the opinion that this is too spammy. There are clients (editors) out there that just can't deal with a bunch of messages, and as you've stated inthe PR description, 

> This makes it a lot easier to tell that `rust-analyzer` isn't crashed, it's just trying to scan a huge directory.

So I think only showing it for dirs makes the most sense

9999years: OK, changed it to only send notifications for directories.

Veykril: Your force push didn't seem to change anything

Veykril:
Thanks!
@bors r+

lnicola:
changelog internal (first contribution) show which roots are being scanned in progress messages

PR #15641, opened 2023-09-20
fix: preceding QualifiedPathType for into_to_from assist

### LINKED ISSUE #15598: The `into` to `from` assist outputs wrong syntax for some types
The "convert `into` to fully qualified `from`" assist breaks code for some output types, eg arrays:
```rust
fn tuple_to_array() -> [i32; 2] {
    (0,1).into()
}
```
produces `[i32; 2]::from((0,1))` instead of `<[i32; 2]>::from((0,1))` resulting in an "error: missing angle brackets in associated item path".

**rust-analyzer version**: 0.3.1657-standalone
**rustc version**: 1.72.0 (5680fa18f 2023-08-23)

alibektas:
fixes #15598

### DIFF
--- a/crates/ide-assists/src/handlers/into_to_qualified_from.rs
+++ b/crates/ide-assists/src/handlers/into_to_qualified_from.rs
@@ -66,10 +66,10 @@ pub(crate) fn into_to_qualified_from(acc: &mut Assists, ctx: &AssistContext<'_>)
             |edit| {
                 edit.replace(
                     method_call.syntax().text_range(),
-                    if sc.chars().find(|c| !c.is_alphanumeric() && c != &':').is_some() {
-                        format!("<{}>::from({})", sc, receiver)
-                    } else {
+                    if sc.chars().all(|c| c.is_alphanumeric() || c == ':') {
                         format!("{}::from({})", sc, receiver)
+                    } else {
+                        format!("<{}>::from({})", sc, receiver)
                     },
                 );
             },
--- a/crates/ide-assists/src/handlers/into_to_qualified_from.rs
+++ b/crates/ide-assists/src/handlers/into_to_qualified_from.rs
@@ -52,18 +52,24 @@ pub(crate) fn into_to_qualified_from(acc: &mut Assists, ctx: &AssistContext<'_>)
         == FamousDefs(sema, scope.krate()).core_convert_Into()?
     {
         let type_call = sema.type_of_expr(&method_call.clone().into())?;
-        let type_call_disp =
-            type_call.adjusted().display_source_code(db, scope.module().into(), true).ok()?;
+        let adjusted_tc = type_call.adjusted();
+
+        if adjusted_tc.is_unknown() && adjusted_tc.contains_unknown() {
+            return None;
+        }
+
+        let qualified_from = format!(
+            "<{}>::from({})",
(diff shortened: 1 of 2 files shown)


crates/ide-assists/src/handlers/into_to_qualified_from.rs
```diff
@@ -52,18 +52,24 @@ pub(crate) fn into_to_qualified_from(acc: &mut Assists, ctx: &AssistContext<'_>)
...
+
+        if adjusted_tc.is_unknown() && adjusted_tc.contains_unknown() {
+            return None;
+        }
+
+        let qualified_from = format!(
+            "<{}>::from({})",
+            adjusted_tc.display_source_code(db, scope.module().into(), true).ok()?,
+            receiver
+        );
```
Veykril, line 65: We can move everything but the `adjusted_tc.display_source_code(db, scope.module().into(), true).ok()?` line into the builder closure

crates/ide-assists/src/handlers/into_to_qualified_from.rs
```diff
@@ -52,18 +52,24 @@ pub(crate) fn into_to_qualified_from(acc: &mut Assists, ctx: &AssistContext<'_>)
         == FamousDefs(sema, scope.krate()).core_convert_Into()?
     {
         let type_call = sema.type_of_expr(&method_call.clone().into())?;
-        let type_call_disp =
-            type_call.adjusted().display_source_code(db, scope.module().into(), true).ok()?;
+        let adjusted_tc = type_call.adjusted();
+
+        if adjusted_tc.is_unknown() && adjusted_tc.contains_unknown() {
```
Veykril, line 57:
```suggestion
        if adjusted_tc.contains_unknown() {
```
should suffice

Veykril (commented):
While this is completely valid, it is a bit odd ending up with `<Foo>::thing` when `Foo::thing` is encouraged. We can probably iterate the rendered type string and check whether all characters are identifier chars and omit the `<>` then. (Ideally we could do that check on the hir::Type somehow but I don't think we have the infra for that and it sounds non trivial, not sure though).

crates/ide-assists/src/handlers/into_to_qualified_from.rs
```diff
@@ -52,17 +52,25 @@ pub(crate) fn into_to_qualified_from(acc: &mut Assists, ctx: &AssistContext<'_>)
...
+        let sc = adjusted_tc.display_source_code(db, scope.module().into(), true).ok()?;
         acc.add(
             AssistId("into_to_qualified_from", AssistKind::Generate),
             "Convert `into` to fully qualified `from`",
             nameref.syntax().text_range(),
             |edit| {
                 edit.replace(
                     method_call.syntax().text_range(),
-                    format!("{}::from({})", type_call_disp, receiver),
+                    if sc.chars().find(|c| !c.is_alphanumeric() && c != &':').is_some() {
```
Veykril, line 69:
```suggestion
                    if sc.chars().any(|&c| !c.is_alphanumeric() && c != ':') {
```
Alternatively (with the then and else branches flipped)
```rs
                    if sc.chars().all(|&c| c.is_alphanumeric() || c = ':') {
                        format!("{}::from({})", sc, receiver)
                    } else {
                        format!("<{}>::from({})", sc, receiver)
                    },
```
would be a bit more readable due to less negations

PR #15660, opened 2023-09-24
minor: hover_simple refactor

DaniPopes:
A minor refactor of `hover_simple` opportunity I noticed while skimming through the code.

- `if let`s -> `match`
- `iter::once(x).cycle` -> `iter::repeat`
- `classify_token` -> `classify_node`


HKalbasi:
Thanks!
@bors r+

PR #15668, opened 2023-09-26
feat: add backtick to surrounding and auto-closing pairs

connorskees:
Makes backticks always complete as a pair and also surround the current selection, similar to double quotes. This is useful primarily in the context of markdown doc comments, but is applied globally for simplicity.

Closes https://github.com/rust-lang/rust-analyzer/issues/11381


Veykril:
Makes sense to me :+1: 
@bors r+

Veykril:
Given we don't use any fancy functionality that seems fine to me

PR #15680, opened 2023-09-29
internal: re-generate lints.rs

DaniPopes:
Looks like this hasn't been run in a while


crates/ide-db/src/tests/sourcegen_lints.rs
```diff
@@ -1,7 +1,7 @@
 //! Generates descriptors structure for unstable feature from Unstable Book
-use std::{borrow::Cow, fs, path::Path};
 
 use itertools::Itertools;
+use std::{borrow::Cow, fs, path::Path};
```
Veykril, line 4: nit but could you undo this import movement? We keep std imports separate

Veykril:
Thanks!
@bors delegate+

PR #15681, opened 2023-09-29
Add config option to use `rust-analyzer` specific target dir

### LINKED ISSUE #6007: Feature request: Option to use `rust-analyzer`-specific target directory (relative to target directory)
This would be useful to implement one of the workarounds for [Rust analyzer "cargo check" blocks debug builds #4616](https://github.com/rust-analyzer/rust-analyzer/issues/4616) for people who use workspaces.

Normally, my current working directory is a specific crate, but I could also be in the workspace directory.

Because of that, if I add `["--target-dir", "/path/to/proect/target/check"]` to "rust-analyzer.checkOnSave.extraArgs" I can't use a relative path (the path would be "target/check" if I'm in the workspace, or "../target/check" if I'm in a crate directory).

Workarounds would be to use an absolute path (then I have to set this config option for every project) or to create a symbolic link in every crate to the target directory of the workspace.

A better solution, IMO, would be if there was an option for Rust-Analyzer to use a specific directory relative to the target directory.

Edit: [This comment](https://github.com/rust-lang/rust-analyzer/issues/6007#issuecomment-1379342831) contains a good workaround to switch Rust-Analyzer to a different Cargo profile, globally.

Tyrubias:
Adds a Rust Analyzer configuration option to set a custom target directory for builds. This is a workaround for Rust Analyzer blocking debug builds while running `cargo check`. This change should close #6007.

This is my first time contributing to this project, so any feedback regarding best practices that I'm not aware of are greatly appreciated! Thanks to all the maintainers for their hard work on this project and reviewing contributions.

### DIFF
--- a/crates/flycheck/src/lib.rs
+++ b/crates/flycheck/src/lib.rs
@@ -50,6 +50,7 @@ pub enum FlycheckConfig {
         extra_args: Vec<String>,
         extra_env: FxHashMap<String, String>,
         ansi_color_output: bool,
+        target_dir: Option<PathBuf>,
     },
     CustomCommand {
         command: String,
@@ -308,6 +309,7 @@ impl FlycheckActor {
                 features,
                 extra_env,
                 ansi_color_output,
+                target_dir,
             } => {
                 let mut cmd = Command::new(toolchain::cargo());
                 cmd.arg(command);
@@ -339,6 +341,9 @@ impl FlycheckActor {
                         cmd.arg("--features");
                         cmd.arg(features.join(" "));
                     }
+                }
+                if let Some(target_dir) = target_dir {
+                    cmd.arg("--target-dir").arg(target_dir);
                 }
                 cmd.envs(extra_env);
                 (cmd, extra_args)
--- a/crates/flycheck/src/lib.rs
+++ b/crates/flycheck/src/lib.rs
@@ -58,6 +58,7 @@ pub enum FlycheckConfig {
         extra_env: FxHashMap<String, String>,
         invocation_strategy: InvocationStrategy,
         invocation_location: InvocationLocation,
+        target_dir: Option<PathBuf>,
     },
 }
 
@@ -354,9 +355,14 @@ impl FlycheckActor {
                 extra_env,
                 invocation_strategy,
                 invocation_location,
+                target_dir,
             } => {
                 let mut cmd = Command::new(command);
                 cmd.envs(extra_env);
+
+                if let Some(target_dir) = target_dir {
+                    cmd.env("CARGO_TARGET_DIR", target_dir);
+                }
 
                 match invocation_location {
                     InvocationLocation::Workspace => {
--- a/crates/flycheck/src/lib.rs
+++ b/crates/flycheck/src/lib.rs
@@ -360,10 +360,6 @@ impl FlycheckActor {
                 let mut cmd = Command::new(command);
                 cmd.envs(extra_env);
 
-                if let Some(target_dir) = target_dir {
-                    cmd.env("CARGO_TARGET_DIR", target_dir);
-                }
-
                 match invocation_location {
                     InvocationLocation::Workspace => {
                         match invocation_strategy {
@@ -379,6 +375,10 @@ impl FlycheckActor {
                     InvocationLocation::Root(root) => {
                         cmd.current_dir(root);
                     }
+                }
+
+                if let Some(target_dir) = target_dir {
+                    cmd.arg("--target-dir").arg(target_dir);
                 }
 
                 (cmd, args)
--- a/crates/flycheck/src/lib.rs
+++ b/crates/flycheck/src/lib.rs
@@ -58,7 +58,6 @@ pub enum FlycheckConfig {
         extra_env: FxHashMap<String, String>,
         invocation_strategy: InvocationStrategy,
         invocation_location: InvocationLocation,
-        target_dir: Option<PathBuf>,
     },
 }
 
@@ -355,7 +354,6 @@ impl FlycheckActor {
                 extra_env,
                 invocation_strategy,
                 invocation_location,
-                target_dir,
             } => {
                 let mut cmd = Command::new(command);
                 cmd.envs(extra_env);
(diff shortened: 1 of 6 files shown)


crates/rust-analyzer/src/config.rs
```diff
@@ -2037,6 +2079,14 @@ pub enum MemoryLayoutHoverRenderKindDef {
     Both,
 }
 
+#[derive(Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
+#[serde(untagged)]
+pub enum TargetDirectory {
+    UseSubdirectory(bool),
+    Directory(String),
```
Veykril, line 2087:
```suggestion
    Directory(PathBuf),
```

crates/rust-analyzer/src/config.rs
```diff
@@ -1303,42 +1319,61 @@ impl Config {
...
-                    CargoFeaturesDef::All => vec![],
-                    CargoFeaturesDef::Selected(it) => it,
-                },
-                extra_args: self.check_extra_args(),
-                extra_env: self.check_extra_env(),
-                ansi_color_output: self.color_diagnostic_output(),
-            },
+            Some(_) | None => {
+                let mut extra_args = self.check_extra_args();
+                add_target_dir_to_args(&mut extra_args, target_directory);
```
Veykril, line 1324: We should add a dedicated field to `FlyCheckConfig` instead

crates/rust-analyzer/src/config.rs
```diff
@@ -1209,6 +1217,10 @@ impl Config {
         let sysroot_src =
             self.data.cargo_sysrootSrc.as_ref().map(|sysroot| self.root_path.join(sysroot));
 
+        let mut extra_args = self.data.cargo_extraArgs.clone();
+
+        add_target_dir_to_args(&mut extra_args, target_directory);
```
Veykril, line 1222: We should add a dedicated field for this to `CargoConfig` instead

Veykril (changes requested):

Tyrubias:
@Veykril I realized it might not make sense for us to pass `--target-dir` to a Flycheck custom command. Should we set `CARGO_TARGET_DIR` for Flycheck custom commands?

Veykril:
Why would the env var make more sense then passing the flag?

Tyrubias:
Is it not possible for the Flycheck custom command to invoke another command that invokes `cargo`? In that case adding `--target-dir` would cause an error, but then the environment variable might eventually be inherited by the `cargo` process. Please let me know if I'm mistaken.

Veykril:
That's fine, if you use a command override most of the settings no longer apply. In general we expect people to completely build up their command if they use an override.

Tyrubias:
That makes sense, I've changed it back to using `--target-dir` instead of `CARGO_TARGET_DIR`. Are there any more tests I should add for this behavior? Thanks.

crates/flycheck/src/lib.rs
```diff
@@ -50,13 +50,15 @@ pub enum FlycheckConfig {
...
         ansi_color_output: bool,
+        target_dir: Option<PathBuf>,
     },
     CustomCommand {
         command: String,
         args: Vec<String>,
         extra_env: FxHashMap<String, String>,
         invocation_strategy: InvocationStrategy,
         invocation_location: InvocationLocation,
+        target_dir: Option<PathBuf>,
```
Veykril, line 61:
```suggestion
```

If the user uses a custom command they should be specifying the target dir themselves in the args as we can't predict what format their command expect.

This is the bahvior with all cargo config args, we drop them all for the custom command

crates/rust-analyzer/src/config.rs
```diff
@@ -1335,10 +1345,25 @@ impl Config {
...
+            .rust_analyzerTargetDir
+            .as_ref()
+            .map(|target_dir| match target_dir {
+                TargetDirectory::UseSubdirectory(yes) if *yes => {
+                    Some(PathBuf::from("target/rust-analyzer"))
+                }
+                TargetDirectory::UseSubdirectory(_) => None,
+                TargetDirectory::Directory(dir) => Some(dir.clone()),
+            })
+            .flatten()
```
Veykril, line 1364:
```suggestion
            .and_then(|target_dir| match target_dir {
                TargetDirectory::UseSubdirectory(true) => {
                    Some(PathBuf::from("target/rust-analyzer"))
                }
                TargetDirectory::UseSubdirectory(false) => None,
                TargetDirectory::Directory(dir) => Some(dir.clone()),
            })
```

crates/rust-analyzer/src/config.rs
```diff
@@ -480,6 +480,13 @@ config_data! {
         /// tests or binaries. For example, it may be `--release`.
         runnables_extraArgs: Vec<String>   = "[]",
 
+        /// Optional path to a rust-analyzer specific target directory.
+        /// This is useful to prevent rust-analyzer's `cargo check` from blocking builds.
```
Veykril, line 484:
```suggestion
        /// This is prevents rust-analyzer's `cargo check` from locking the `Cargo.lock` at the expense of duplicating build artifacts.
```

Tyrubias:
Thank you so much for the mentorship! I've addressed your comments.

@rustbot ready

Veykril (approved):

Veykril:
Thanks!
@bors r+

crates/rust-analyzer/src/config.rs
```diff
@@ -480,6 +480,14 @@ config_data! {
         /// tests or binaries. For example, it may be `--release`.
         runnables_extraArgs: Vec<String>   = "[]",
 
+        /// Optional path to a rust-analyzer specific target directory.
+        /// This prevents rust-analyzer's `cargo check` from locking the `Cargo.lock`
+        /// at the expense of duplicating build artifacts.
+        ///
+        /// Set to `true` to use a subdirectory of the existing target directory or
+        /// set to a path relative to the workspace to use that path.
+        rust_analyzerTargetDir: Option<TargetDirectory> = "null",
```
lnicola, line 489: Would have made sense to call this `cargo_` instead.

matklad:
Brilliant idea!

PR #15686, opened 2023-09-29
Only run tests on linux in pull requests

Veykril:
There is little gain from running them on all platforms, they should almost never fail if the linux one pass and if they do, bors will catch it still.


Veykril:
We could probably skip building anything but linux as well, as again, its unlikely to happen and bors will catch it.

lnicola (approved):

PR #15688, opened 2023-09-29
Make rustc_layout_scalar_valid_range attributes work for non-decimal literals

Veykril:
Closes https://github.com/rust-lang/rust-analyzer/issues/15687

### DIFF
--- a/crates/hir-ty/src/layout/adt.rs
+++ b/crates/hir-ty/src/layout/adt.rs
@@ -119,7 +119,15 @@ fn layout_scalar_valid_range(db: &dyn HirDatabase, def: AdtId) -> (Bound<u128>, 
         let attr = attrs.by_key(name).tt_values();
         for tree in attr {
             if let Some(it) = tree.token_trees.first() {
-                if let Ok(it) = it.to_string().parse() {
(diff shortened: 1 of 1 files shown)


crates/hir-ty/src/layout/adt.rs
```diff
@@ -119,7 +119,15 @@ fn layout_scalar_valid_range(db: &dyn HirDatabase, def: AdtId) -> (Bound<u128>,
...
-                if let Ok(it) = it.to_string().parse() {
+                let text = it.to_string().replace('_', "");
+                let base = match text.as_bytes() {
+                    [b'0', b'x', ..] => 16,
+                    [b'0', b'o', ..] => 8,
+                    [b'0', b'b', ..] => 2,
+                    _ => 10,
+                };
+
+                if let Ok(it) = u128::from_str_radix(&text, base) {
```
lnicola, line 130: Are you sure this is working? `u128::from_str_radix("0xffffffff", 16)` fails because of the `0x` prefix.

Veykril: No I noticed it still not working, that would explain it

PR #15690, opened 2023-09-29
vscode: Fix line and col regexp for `rustc` problem matcher

Enselic:
When building the Rust compiler with `./x check` from within VS Code, the current `rustc` problem matcher thinks that the output from that command that looks like this:

    Build completed successfully in 0:00:26

is about a problem in a file named `0` on line 00, col 26. This wouldn't be so bad if it wasn't for that VS Code tends to get stuck on this problem because of problems with opening the file '0'.

The rust compiler will never output problems with a line or a column that starts with 0, so change the regexp to require lines and cols to begin with `[1-9]` to fix this problem.


Veykril (approved):

Veykril:
Sounds reasonable
@bors r+

PR #15696, opened 2023-10-01
feat: implement tuple return type to tuple struct assist

### LINKED ISSUE #14293: feature request: convert tuple return type to tuple struct.
It would be very nice to have an assist which performs the following:

```rust
fn bar() {
    let (a, b, c) = foo();
}

fn foo() -> ($0u32, u32, u32) {
    (1, 2, 3)
}
```
-> 
```rust
fn bar() {
    let FooResult(a, b, c) = foo();
}

struct FooResult(u32, u32,  32);

fn foo() -> FooResult {
    FooResult(1, 2, 3)
}
```

Many a time do I write a function that starts out with a simple return type, but eventually becomes more elaborate. I usually am lazy and make it a tuple to start with, but that becomes unwieldly and I perform the following refactor manually. 

The nice thing about this, is that you can also then use the "Convert to named struct" assist, to refactor this even further if the usage of a tuple struct becomes unwieldly.

rmehri01:
This PR implements the `convert_tuple_return_type_to_struct` assist, for converting the return type of a function or method from a tuple to a tuple struct. Additionally, it moves the `to_camel_case` and `char_has_case` functions from `case_conv` to `stdx` so that they can be used similar to `to_lower_snake_case`.

[tuple_return_type_to_tuple_struct.webm](https://github.com/rust-lang/rust-analyzer/assets/52933714/2803ff58-fde3-4144-9495-7c7c7e139075)

Currently, the assist puts the struct definition above the function, or above the nearest `impl` or `trait` if applicable and only rewrites literal tuples that are returned in the body of the function. Additionally, it only attempts to rewrite simple tuple pattern usages with the corresponding tuple struct pattern but does so across files and modules.

I think that this is sufficient for the majority of use cases but I could be wrong. One thing I'm still not sure how to approach is handling `Self` and generics/lifetimes in the tuple type to be extracted. I was thinking of either manually figuring out what lifetimes and generics are in scope and using them (sort of similar to the `generate_function` assist) or maybe using `ctx.sema.resolve_type` and `generic_params` on `hir::Type` but this seems to not deal with lifetimes.

Closes #14293

### DIFF
--- a/crates/ide-assists/src/handlers/convert_tuple_return_type_to_struct.rs
+++ b/crates/ide-assists/src/handlers/convert_tuple_return_type_to_struct.rs
@@ -1,0 +1,891 @@
+use either::Either;
+use hir::ModuleDef;
+use ide_db::{
+    assists::{AssistId, AssistKind},
+    defs::Definition,
+    helpers::mod_path_to_ast,
+    imports::insert_use::{insert_use, ImportScope},
+    search::{FileReference, UsageSearchResult},
+    source_change::SourceChangeBuilder,
+    syntax_helpers::node_ext::{for_each_tail_expr, walk_expr},
+    FxHashSet,
+};
+use syntax::{
+    ast::{self, edit::IndentLevel, edit_in_place::Indent, make, HasName},
+    match_ast, ted, AstNode, SyntaxNode,
+};
+
+use crate::assist_context::{AssistContext, Assists};
+
+// Assist: convert_tuple_return_type_to_struct
+//
+// This converts the return type of a function from a tuple type
+// into a tuple struct and updates the body accordingly.
+//
+// ```
+// fn bar() {
+//     let (a, b, c) = foo();
+// }
+//
+// fn foo() -> ($0u32, u32, u32) {
+//     (1, 2, 3)
+// }
+// ```
+// ->
+// ```
+// fn bar() {
+//     let FooResult(a, b, c) = foo();
+// }
+//
+// struct FooResult(u32, u32, u32);
+//
+// fn foo() -> FooResult {
+//     FooResult(1, 2, 3)
+// }
+// ```
+pub(crate) fn convert_tuple_return_type_to_struct(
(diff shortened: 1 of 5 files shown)


crates/ide-assists/src/handlers/convert_tuple_return_type_to_struct.rs
```diff
@@ -0,0 +1,891 @@
...
+
+                if let Some(fn_body) = fn_.body() {
+                    replace_body_return_values(ast::Expr::BlockExpr(fn_body), struct_name);
+                }
+            } else {
+                // replace tuple patterns
+                let pats = name
+                    .syntax()
+                    .ancestors()
+                    .nth(5)
```
Veykril, line 132: Could you add a comment about the magic 5 here?

Veykril (commented):
This disregards visibilities. Tuples and their fields are always public but the assist creates a struct

crates/ide-assists/src/handlers/convert_tuple_return_type_to_struct.rs
rmehri01: Oh yeah sorry, this is mostly a relic, I was basically trying to handle this case where we don't want to replace the outer tuple with the struct pattern. So, I was trying to get the direct parent of the call or method call expression but I've pushed a cleaner version now.

```rust
let ((bar1, bar2), foo) = (bar(), 3);
```

Also, in this case it would be nice to replace the corresponding pattern but I'm not sure if/how we should do that? Like:

```rust
let (BarResult(bar1, bar2), foo) = (bar(), 3);
```

Veykril: Ye that's tricky, we can only do heuristics here I think.

rmehri01: Yeah maybe we can leave it for now then, I think figuring out how to handle generics is more important.

PR #15697, opened 2023-10-01
Enable Syntax Highlighting for no_test Code Blocks

Indra-db:
Syntax highlighting is now enabled for code blocks annotated with `no_test`.

currently:
image

after:
image

might address https://github.com/rust-lang/rust/issues/63193 where there are comments about missing code syntax while also not being picked up by doc tests.

### DIFF
--- a/crates/ide-db/src/rust_doc.rs
+++ b/crates/ide-db/src/rust_doc.rs
@@ -15,7 +15,7 @@ pub fn is_rust_fence(s: &str) -> bool {
 
     for token in tokens {
         match token {
-            "should_panic" | "no_run" | "ignore" | "allow_fail" => {
+            "should_panic" | "no_run" | "ignore" | "allow_fail" | "no_test" => {
                 seen_rust_tags = !seen_other_tags
             }
             "rust" => seen_rust_tags = true,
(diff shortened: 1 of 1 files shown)


Veykril:
Alright I misunderstood you another time in the discord message actually. I assumed `no_test` was a proper rustdoc thing but it is not. So I'm reluctant to adding it here. There should definitely be support for a non compiling tag upstream though, so I am quite surprised that the team seems so against it ...

Indra-db:
@Veykril Alright, that's unfortunate, anything else I can do to help? 

Also out of curiosity, should a noncompiling tag actually be added upstream? What benefit would that give? It would essentially mean "do nothing", which is the same as having no direct support for it and just enables syntax highlighting on it 🤔.

Veykril:
But isn't such a tag exactly what you want here? You don't want it to be compiled, but still treated as rust for syntax highlighting. The same should therefor apply for rustdoc since it also does syntax highlighting so to me that sounds like you'd want such a tag to be added to rustdoc proper unless I'm missing something here

Indra-db:
ah I misunderstood what you meant! Yeah what you said is correct. 
Should I close this then if nothing is to be done?

Veykril:
Ye let's close this, we don't wanna diverge from rustdoc here

PR #15698, opened 2023-10-01
fix: allow more kinds of if let patterns in guarded return assist

### LINKED ISSUE #15695: Convert to guarded return does not work with many kinds of `if let`
The "Convert to guarded return" assist shows up for some instances of `if let`, like:
```rust
fn main() {
    if let Some(x) = Some(92) {
        foo(x);
    }
}
```
but does not show up for many others:
```rust
fn main() {
    if let None = Some(92) {
        foo();
    }
}
```
```rust
fn main() {
    if let [1, x] = [1, 92] {
        foo(x);
    }
}
```
```rust
fn main() {
    if let (Some(x), None) = (Some(92), None) {
        foo(x);
    }
}
```

I would also like to have "Convert to guarded continue" in loop bodies, as mentioned in the original suggestion #1782, but that should probably be a separate issue.

**rust-analyzer version**: 0.3.1673-standalone

**rustc version**: rustc 1.72.1 (d5c2e9c34 2023-09-13)

**relevant settings**: Nothing obviously relevant (that isn't set to default).

rmehri01:
Removes the checks that require the pattern to be a tuple struct with exactly 1 field that is unqualified and has an identifier pattern in it. I'm not sure if there should be more checks in place but they seem unnecessary now?

Closes #15695

### DIFF
--- a/crates/ide-assists/src/handlers/convert_to_guarded_return.rs
+++ b/crates/ide-assists/src/handlers/convert_to_guarded_return.rs
@@ -51,22 +51,7 @@ pub(crate) fn convert_to_guarded_return(acc: &mut Assists, ctx: &AssistContext<'
     // Check if there is an IfLet that we can handle.
     let (if_let_pat, cond_expr) = if is_pattern_cond(cond.clone()) {
         let let_ = single_let(cond)?;
-        match let_.pat() {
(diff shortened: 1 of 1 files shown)


Veykril (approved):
These are indeed unnecessary now that we are emitting `let ... else ...` here, good catch

PR #15705, opened 2023-10-03
fix: resolve Self type references in delegate method assist

### LINKED ISSUE #14485: Delegate method assist doesn't fix up `Self` references
```rs
struct Struct {
    field$0: u8,
}
```
produces
```rs
impl Struct {
    pub fn saturating_mul(self,rhs:Self) ->Self {
        self.field.saturating_mul(rhs)
    }
}
```
when it should produce
```rs
impl Struct {
    pub fn saturating_mul(self,rhs:u8) ->u8{
        self.field.saturating_mul(rhs)
    }
}
```

rmehri01:
This PR makes the delegate method assist resolve any `Self` type references in the parameters or return type. It also works across macros such as the `uint_impl!` macro used for `saturating_mul` in the issue example.

Closes #14485

### DIFF
--- a/crates/ide-assists/src/handlers/generate_delegate_methods.rs
+++ b/crates/ide-assists/src/handlers/generate_delegate_methods.rs
@@ -134,7 +134,7 @@ pub(crate) fn generate_delegate_methods(acc: &mut Assists, ctx: &AssistContext<'
                     vis,
                     fn_name,
                     type_params,
-                    None,
+                    method_source.where_clause(),
                     params,
                     body,
                     ret_type,
@@ -461,6 +461,53 @@ impl Person {
         self.age.age()
     }
 }"#,
+        );
+    }
+
+    #[test]
+    fn test_preserve_where_clause() {
+        check_assist(
+            generate_delegate_methods,
+            r#"
+struct Inner<T>(T);
+impl<T> Inner<T> {
+    fn get(&self) -> T
+    where
+        T: Copy,
+        T: PartialEq,
+    {
+        self.0
+    }
+}
+
+struct Struct<T> {
+    $0field: Inner<T>,
+}
+"#,
+            r#"
+struct Inner<T>(T);
+impl<T> Inner<T> {
+    fn get(&self) -> T
+    where
+        T: Copy,
+        T: PartialEq,
+    {
+        self.0
+    }
+}
+
+struct Struct<T> {
+    field: Inner<T>,
+}
+
+impl<T> Struct<T> {
+    $0fn get(&self) -> T where
+            T: Copy,
+            T: PartialEq, {
+        self.field.get()
+    }
+}
+"#,
         );
     }
 
--- a/crates/ide-assists/src/handlers/generate_delegate_methods.rs
+++ b/crates/ide-assists/src/handlers/generate_delegate_methods.rs
@@ -1,6 +1,7 @@
 use std::collections::HashSet;
 
 use hir::{self, HasCrate, HasSource, HasVisibility};
+use ide_db::path_transform::PathTransform;
 use syntax::{
     ast::{
         self, edit_in_place::Indent, make, AstNode, HasGenericParams, HasName, HasVisibility as _,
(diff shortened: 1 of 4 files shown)


Veykril:
Looking at the assist, I'm surprised we aren't using the `PathTransform` yet to fix up paths. We definitely should do that here instead (and then add support for substituting `Self` appropriately). Not sure how simple that is though.

rmehri01:
Yeah that sounds like a good idea, I could give it a try! `PathTransform` expects a `SemanticsScope` for the target scope though, so is there a way to get that with a macro? Like in this example, since `ctx.sema.scope` won't work:

```rust
macro_rules! test_method {
    () => {
        fn test(self) -> Self {
            self
        }
    };
}

struct Bar;

impl Bar {
    test_method!();
}

struct Foo {
    bar: $0Bar,
}
```

Veykril:
Why wouldn't it work there? We can take a scope from a macro expansion's node just fine

rmehri01:
Ah I was running into an issue where it said the lookup failed in this instance of `Semantics` but I added `sema.parse_or_expand`, which seems to update the cache and gets it to work! 

I added the change to `PathTransform` to change `Self` now but I'm not sure if its the right way to do it, seems to at least pass the tests I had before.

crates/ide-assists/src/handlers/generate_delegate_methods.rs
```diff
@@ -106,7 +107,10 @@ pub(crate) fn generate_delegate_methods(acc: &mut Assists, ctx: &AssistContext<'
             |edit| {
                 // Create the function
                 let method_source = match method.source(ctx.db()) {
```
Veykril, line 109:
```suggestion
                let method_source = match ctx.sema.source(method) {
```
We can do this instead of the extra `parse_or_expand`, this will do the caching for us then

crates/ide-assists/src/handlers/generate_delegate_methods.rs
```diff
@@ -183,6 +187,12 @@ pub(crate) fn generate_delegate_methods(acc: &mut Assists, ctx: &AssistContext<'
                 let assoc_items = impl_def.get_or_create_assoc_item_list();
                 assoc_items.add_item(f.clone().into());
 
+                PathTransform::generic_transformation(
+                    &ctx.sema.scope(strukt.syntax()).unwrap(),
+                    &ctx.sema.scope(method_source.syntax()).unwrap(),
+                )
+                .apply(f.syntax());
```
Veykril, line 194:
```suggestion
                if let Some((target, source)) =
                    ctx.sema.scope(strukt.syntax()).zip(ctx.sema.scope(method_source.syntax()))
                {
                    PathTransform::generic_transformation(&target, &source).apply(f.syntax());
                }
```
The unwraps shouldn't really be possible to hit here, but we are trying to not use unwraps in assist code if possible.

crates/ide-db/src/path_transform.rs
```diff
@@ -327,8 +327,36 @@ impl Ctx<'_> {
...
+
+                        if let Some(qual) = mod_path_to_ast(&found_path).qualifier() {
+                            let res = make::path_concat(qual, path_ty.path()?).clone_for_update();
+                            ted::replace(path.syntax(), res.syntax());
+                            return Some(());
+                        }
+                    }
+                }
+
+                ted::replace(path.syntax(), ast_ty.syntax());
```
Veykril, line 357: We probably wanna prefer keeping `Self` if it does not need to be replaced. So a check if the source and target scopes have the same `Self` (and in that case, skip this transformation here entirely). We can give `Ctx` a new bool field for that and set it accordingly in `build_ctx` by giving `SemanticsScope` a function like `fn has_same_self_type(&self, other: &SemanticsScope<'_>) -> bool` which just checks if the top `ImplDefScope`s of each scope's resolver has the same def in them.

crates/ide-assists/src/handlers/generate_delegate_methods.rs
rmehri01: Ah okay, sorry about that!

Veykril:
Thanks!
@bors r+

Veykril:
rebased
@bors r+

PR #15713, opened 2023-10-05
Do flyimport completions by prefix search for short paths

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/15711


Veykril:
I think this should be fine for single char queries, but if its not we could also special case it so we do exact searches for single char ones (practically useless), prefix search for 2 char queries and fuzzy for the rest

lnicola:
image

PR #15718, opened 2023-10-06
Fix typos exec_intrinsic

Toby222:
Fix names of intrinsics in error messages from exec_intrinsic
(moved from rust-lang/rust#116478)


lnicola:
Thanks!

@bors r+

PR #15725, opened 2023-10-07
internal: fix automatic rustc/rustdoc lint generation

DaniPopes:
Missed in #15680: the output of `-Whelp` changed since the last run so it generated some bad rustdoc lints entries.
Also preemptively fix a `-Whelp` breakage that might get merged in rust-lang/rust#116412


Veykril:
Thanks!
@bors r+

PR #15727, opened 2023-10-07
feat: generate descriptors for all unstable features

DaniPopes:
Most unstable features don't have their own chapter in the unstable book, so a rustc helper tool (`src/tools/unstable-book-gen`) generates shims to fill the gaps.
Run this tool to generate the full unstable-book source before parsing it.


HKalbasi:
Thanks!
@bors r+

PR #14662, opened 2023-04-26
 Provide links to locally built documentation for `experimental/externalDocs`

Ddystopia:
This pull request addresses issue #12867, which requested the ability to provide links to locally built documentation when using the "Open docs for symbol" feature. Previously, rust-analyzer always used docs.rs for this purpose. With these changes, the feature will provide both web (docs.rs) and local documentation links without verifying their existence.

Changes in this PR:

   - Added support for local documentation links alongside web documentation links.
   - Added `target_dir` path argument for external_docs and other related methods.
   - Added `sysroot` argument for external_docs.
   - Added `target_directory` path to `CargoWorkspace`.

API Changes:
   
   - Added an experimental client capability `{ "localDocs": boolean }`. If this capability is set, the `Open External Documentation` request returned from the server will include both web and local documentation links in the `ExternalDocsResponse` object.

Here's the `ExternalDocsResponse` interface:

```typescript
interface ExternalDocsResponse {
    web?: string;
    local?: string;
}
```

By providing links to both web-based and locally built documentation, this update improves the developer experience for those using different versions of crates, git dependencies, or local crates not available on docs.rs. Rust-analyzer will now provide both web (docs.rs) and local documentation links, leaving it to the client to open the desired link. Please note that this update does not perform any checks to ensure the validity of the provided links.

PR #15728, opened 2023-10-08
feat: vscode: Support opening local documentation if available

### LINKED ISSUE #12867: Feature request: Open docs for symbol should support locally built documentation
Rust-analyzer has a feature to open the docs for the symbol under the cursor, but it always uses docs.rs.
One could be using different versions of crates that aren't on docs.rs, like git dependencies, and also your local crate(s) aren't on docs.rs either.

Open docs should do a best-effort  detection of locally built documentation (usually $TARGET_DIR/doc), and open that if the symbol is available under that, before falling back to docs.rs.

EliasHolzmann:
This PR implements the VS code support for opening local documentation (server side support was already implemented in #14662).

[local_docs.webm](https://github.com/rust-lang/rust-analyzer/assets/9659253/715b84dd-4f14-4ba0-a904-749b847eb3d5)

Displaying local instead of web docs can have many benefits:
- the web version may have different features enabled than locally selected
- the standard library may be a different version than is available online
- the user may not be online and therefore cannot access the web documentation
- the documentation may not be available online at all, for example because it is for a new feature in a library the user is currently developing

If the documentation is not available locally, the extension still falls back to the web version.

Closes #12867.

-----

If my implementation isn't really idiomatic TypeScript: Sorry, I'm not much of a TypeScript developer. I am open to feedback, however.

### DIFF
--- a/editors/code/src/commands.ts
+++ b/editors/code/src/commands.ts
@@ -21,6 +21,7 @@ import type { LanguageClient } from "vscode-languageclient/node";
 import { LINKED_COMMANDS } from "./client";
 import type { DependencyId } from "./dependencies_provider";
 import { unwrapUndefinable } from "./undefinable";
+import { log } from "./util";
 
 export * from "./ast_inspector";
 export * from "./run";
@@ -947,7 +948,24 @@ export function openDocs(ctx: CtxInit): Cmd {
         const position = editor.selection.active;
         const textDocument = { uri: editor.document.uri.toString() };
 
-        const doclink = await client.sendRequest(ra.openDocs, { position, textDocument });
+        const doclinks = await client.sendRequest(ra.openDocs, { position, textDocument });
+
+        let fileType = vscode.FileType.Unknown;
+        if (typeof doclinks.local === "string") {
+            try {
+                fileType = (await vscode.workspace.fs.stat(vscode.Uri.parse(doclinks.local))).type;
+            } catch (e) {
+                log.debug("stat() threw error. Falling back to web version", e);
+            }
+        }
+
+        let doclink;
+        if (fileType & vscode.FileType.File) {
+            // file does exist locally
+            doclink = doclinks.local;
+        } else {
+            doclink = doclinks.web;
+        }
 
         if (doclink != null) {
             await vscode.env.openExternal(vscode.Uri.parse(doclink));
--- a/editors/code/src/commands.ts
+++ b/editors/code/src/commands.ts
@@ -950,7 +950,7 @@ export function openDocs(ctx: CtxInit): Cmd {
         const doclink = await client.sendRequest(ra.openDocs, { position, textDocument });
 
         if (doclink != null) {
-            await vscode.commands.executeCommand("vscode.open", vscode.Uri.parse(doclink));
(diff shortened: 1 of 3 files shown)


editors/code/src/commands.ts
```diff
@@ -947,10 +948,27 @@ export function openDocs(ctx: CtxInit): Cmd {
...
+                typeof doclinks.local === "string" &&
+                (await vscode.workspace.fs.stat(vscode.Uri.parse(doclinks.local)))
+            ) {
+                // file does exist locally
+                doclink = doclinks.local;
+            } else {
+                doclink = doclinks.web;
+            }
+        } catch (e) {
+            log.info("stat() threw error. Falling back to web version", e);
```
Veykril, line 965: This log isn't too useful (nor is the message in it), so we can drop this I think.

EliasHolzmann: My thougt process was it might be useful if `stat()` doesn't throw because of the documentation file being missing, but because of some different reason (corrupt filesystem, hardware fault, missing permissions, etc.). However, I do agree that this might be a bit chatty. I can remove it later today or tomorrow (don't have access to a computer at the moment).

Veykril: I'm fine with downgrading it `debug` in that case as well. (The chatty-ness was my concern here as well)

editors/code/src/commands.ts
```diff
@@ -947,10 +948,27 @@ export function openDocs(ctx: CtxInit): Cmd {
...
         const textDocument = { uri: editor.document.uri.toString() };
 
-        const doclink = await client.sendRequest(ra.openDocs, { position, textDocument });
+        const doclinks = await client.sendRequest(ra.openDocs, { position, textDocument });
+
+        let doclink;
+        try {
+            if (
+                typeof doclinks.local === "string" &&
+                (await vscode.workspace.fs.stat(vscode.Uri.parse(doclinks.local)))
```
Veykril, line 957: We should also check that the type here is a file (and not a folder)

EliasHolzmann: Done. Also refactored the code a little to remove one layer of indentation.

editors/code/src/commands.ts
EliasHolzmann: Done.

Veykril (approved):

Veykril:
Thanks!
@bors r+

EliasHolzmann:
@Ciel-MC: 
- For `std` (as well as `core` and all other libraries that are bundled with Rust), you need to install the documentation with `rustup component add rust-docs-<your target triple>` – so, for Linux, `rustup component add rust-docs-x86_64-unknown-linux-gnu`, for other systems, you may want to have a look at `rustup component list` (see [the Rustup book](https://rust-lang.github.io/rustup/concepts/components.html) for relevant documentation)
- For "normal" dependencies of your crate (and the crate itself), just run `cargo doc` to generate the documentation. You can rerun the command later as needed to update the existing documentation. (see [the Cargo book](https://doc.rust-lang.org/cargo/commands/cargo-doc.html) for relevant documentation – there are some interesting flags for `cargo doc`, for example `--document-private-items` that may be relevant to your use case)

PR #15730, opened 2023-10-08
Add postfix completion for let else

Alainx277:
Adds a postfix completion for let else syntax, similar to the if let postfix.

### DIFF
--- a/crates/ide-completion/src/completions/postfix.rs
+++ b/crates/ide-completion/src/completions/postfix.rs
@@ -75,11 +75,6 @@ pub(crate) fn complete_postfix(
 
     let try_enum = TryEnum::from_ty(&ctx.sema, &receiver_ty.strip_references());
     if let Some(try_enum) = &try_enum {
-        let in_loop = dot_receiver
-            .syntax()
-            .ancestors()
-            .any(|n| matches!(n.kind(), WHILE_EXPR | LOOP_EXPR | FOR_EXPR));
-
         match try_enum {
             TryEnum::Result => {
                 postfix_snippet(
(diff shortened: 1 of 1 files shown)


crates/ide-completion/src/completions/postfix.rs
```diff
@@ -84,6 +89,17 @@ pub(crate) fn complete_postfix(
                 )
                 .add_to(acc, ctx.db);
 
+                postfix_snippet(
+                    "lete",
+                    "let Ok else {}",
+                    &if in_loop {
+                        format!("let Ok($1) = {receiver_text} else {{\n    ${{2|continue,break,return|}};\n}};\n$0")
+                    } else {
+                        format!("let Ok($1) = {receiver_text} else {{\n    return;\n}};\n$0")
```
Veykril, line 98: We should make the `return` here also a snippet I think

Alainx277: Good idea

saiintbrisson (commented):
Super nice PR and auto completions. What do you think of adding the postfix for non-try types as well?

Alainx277:
> Super nice PR and auto completions. What do you think of adding the postfix for non-try types as well?

Thanks. I think it could be useful, but its also doesn't exist for `if let` yet. Maybe you could make a follow-up PR that adds it for both? It would be nice if the enum variants were placeholders (like "continue, break, return" in this PR) 👍

Alainx277:
@Veykril sorry for the ping, but I was wondering if this MR will get another review? I've addressed the last review over 2 months ago and would love for this small improvement to make it into RA.

Veykril:
Apologies I just don't have time on my hands to go through the PR backlog until january.

crates/ide-completion/src/completions/postfix.rs
```diff
@@ -84,6 +89,17 @@ pub(crate) fn complete_postfix(
...
                 .add_to(acc, ctx.db);
 
+                postfix_snippet(
+                    "lete",
+                    "let Ok else {}",
+                    &if in_loop {
+                        format!("let Ok($1) = {receiver_text} else {{\n    ${{2|continue,break,return|}};\n}};\n$0")
+                    } else {
+                        format!("let Ok($1) = {receiver_text} else {{\n    ${{2:return}};\n}};\n$0")
+                    },
```
Veykril, line 99:
```suggestion
                    &format!("let Ok($1) = {receiver_text} else {{\n    $0\n}};"),
```
Let's not prefill these, there are more options people could strife for. Same for the option one below

Alainx277: I'm a bit sad to lose the convenience, but I see how these could be annoying when you want to do more complex things.

Veykril:
Thanks!
@bors r+

PR #15743, opened 2023-10-11
add replace_is_some_with_if_let_some assist

Young-Flash:
This is a basic impl, if it is ok, `if my_result.is_ok()`-> `if let Ok(_x) = my_result ​`will be add next.

close https://github.com/rust-lang/rust-analyzer/issues/12977


HKalbasi:
Thanks!
@bors r+

Young-Flash:
@HKalbasi hi, I'd like to add `if my_result.is_ok()` -> `if let Ok(_x) = my_result` next, the logic is mostly same as this PR, should we make it in a new mod or make them in one mod (probably named as `replace_is_xxx_with_if_let_xxx`, uh, it's hard to give a proper name)

HKalbasi:
Do as you see fit. My weak preference is to have them in a single mod if they can share some code.

PR #15745, opened 2023-10-11
docs: Add Zed to IDE list.

osiewicz:


lnicola (approved):

lnicola:
Thanks. Does it also install it automatically when opening a _project_ (directory or manifest file)?

osiewicz:
Hey,
No, the LSP is installed only when one opens an `.rs` file.

PR #15744, opened 2023-10-11
fix: add diagnostics messages for chars and byte literal errors

pvalletbo:
This PR adds error messages for different invalid byte or character literals. 
Fixes #15395

PR #15746, opened 2023-10-11
String literals diagnose

pvalletbo:
Continues the work from #15744 to add diagnosis errors to Str, ByteStr, and CStr literal kinds. 

Also replaces `unescape_char` for `unescape_byte` to use the correct method for Byte literals.


pvalletbo:
Getting the text within the quotes for `RawStr`, `RawByteStr`, and `RawCStr` requires more work compared to the non raw Str kinds. I haven't had enough time to work on it yet, that's why they are not being handled in the PR yet. 

Happy to work on it as soon as possible.

Veykril (approved):

PR #15754, opened 2023-10-13
fix: Dedup duplicate crates with differing origins in CrateGraph construction

### LINKED ISSUE #15656: [Since 0.4.1658] Cannot rename a non-local definition.
# System Details

**rust-analyzer version**: 0.3.1665-standalone (first broken stable version)
**rustc version**: rustc 1.72.1 (d5c2e9c34 2023-09-13)

## Details

Since `v0.4.1658-standalone` it is not possible to rename any item inside one of my projects. More details in [reproduction](#reproduction) section.

## Regression

Broken since `v0.4.1658-standalone` (nightly). Earlier versions work as expected.

------

1658 (first broken ver.):
Actions Run:  https://github.com/rust-lang/rust-analyzer/actions/runs/6153193088
Commit: https://github.com/rust-lang/rust-analyzer/commit/cc6c8209cbaf7df55013977cf5cc8488d6b7ff1c

1655 (previous functional release): 
Commit: https://github.com/rust-lang/rust-analyzer/commit/994df3d6a31d39f11600f30a6df0b744b13937c1

------

I believe this was introduced in https://github.com/rust-lang/rust-analyzer/pull/15232 . 
However probably not directly caused by said commit (the commit seems to restrict needed functionality), but an unexpected edge case where a local item is believed to be non-local.

## Reproduction

I have tried to create a minimal reproduction, however did not succeed in creating a minimal repro (I even tried copying dependency structure and workspace code settings). 

So I am forced to instead link to one of my real [WIP] projects.

Reproduction Steps:
- Clone `https://github.com/Reloaded-Project/Reloaded.Hooks-rs.git`
- Checkout `main`
- Open main folder in VSCode (folder with README.MD)
  - Note: Projects are included via `.vscode/settings.json` -> `rust-analyzer.linkedProjects`.
- Open any file in `reloaded-hooks-portable` such as `projects/reloaded-hooks-portable/src/api/wrapper_instruction_generator.rs`.

It is not possible to rename any type or even local variable due to:

```
[Error - 03:04:27] Request textDocument/rename failed.
  Message: Cannot rename a non-local definition.
  Code: -32602
```

alibektas:
Partially fixes #15656 . Until now the condition for deduplication in crate graphs were the strict equality of two crates. One problem that arises from this is that in certain conditions when we see the same crate having different `CrateOrigin`s the first occurrence would be kept. This approach however results in some unwanted results such as making renaming forbidden as this has been recently only made available for local crates. The given example in #15656 can still not be resolved with this PR as that involves taking inconsistencies between dependencies into consideration. This will be addressed in a future PR.


crates/base-db/src/input.rs
```diff
@@ -572,25 +686,41 @@ impl CrateGraph {
...
-                        None
+            crate_data.dependencies.iter_mut().for_each(|dep| {
+                dep.crate_id = id_map[&dep.crate_id];
+            });
+            crate_data.dependencies.sort_by_key(|dep| dep.crate_id);
+            let res = self.arena.iter().find_map(|(id, data)| {
+                if data.almost_eq(crate_data) {
+                    if data.origin.is_lib() && crate_data.origin.is_local() {
+                        // See #15656 for a relevant example.
+                        return Some(ExtendStrategy::Replace(id));
```
Veykril, line 707: `Replace` is a strong word for us effectively wanting to deduplicate with an `origin` overwrite/preference. So instead of creating a new crate and removing the previous one, let's just update the duplicates origin below in the match.

crates/base-db/src/input.rs
```diff
@@ -324,6 +328,99 @@ pub struct CrateData {
     pub channel: Option<ReleaseChannel>,
 }
 
+impl CrateData {
+    /**
+    Check if [`other`] is almost equal to [`self`].
+    This method has some obscure bits. These are mostly there to be compliant with
+    some patches. References to the patches are given.
+    */
+    pub fn almost_eq(&self, other: &CrateData) -> bool {
```
Veykril, line 337: Let's call this `eq_ignoring_origin`, that's a bit more clear on what we ignore here in terms of equality

crates/base-db/src/input.rs
```diff
@@ -324,6 +328,97 @@ pub struct CrateData {
...
+        );
+
+        let mut cfgs = opts.into_iter();
+        if let Some(cfg) = cfgs.next() {
+            // Don't care if rust_analyzer CfgAtom is the only cfg in the difference set of self's and other's cfgs.
+            // https://github.com/rust-lang/rust-analyzer/blob/0840038f02daec6ba3238f05d8caa037d28701a0/crates/project-model/src/workspace.rs#L894
+            if cfgs.next().is_some() || cfg.to_string() != "rust_analyzer" {
+                return false;
+            }
+        }
```
Veykril, line 369: Deleaget https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.difference to `CfgOptions::difference` on its field

crates/base-db/src/input.rs
```diff
@@ -324,6 +328,97 @@ pub struct CrateData {
...
+
+        let mut itself = self.dependencies.iter();
```
Veykril, line 371:
```rs
self.dependencies.iter().filter(|it| it.kind != DependencyKind::Normal).eq(other.dependencies.iter().filter(|it| it.kind != DependencyKind::Normal))
```

crates/base-db/src/input.rs
```diff
@@ -324,6 +328,97 @@ pub struct CrateData {
     pub channel: Option<ReleaseChannel>,
 }
 
+impl CrateData {
+    /// Check if [`other`] is almost equal to [`self`] ignoring `CrateOrigin` value.
+    pub fn eq_ignoring_origin(&self, other: &CrateData) -> bool {
```
Veykril, line 333:
```suggestion
    pub fn eq_ignoring_origin_and_deps(&self, other: &CrateData) -> bool {
```

crates/base-db/src/input.rs
```diff
@@ -574,23 +686,33 @@ impl CrateGraph {
...
-
-            let res = self.arena.iter().find_map(
-                |(id, data)| {
-                    if data == crate_data {
-                        Some(id)
-                    } else {
-                        None
+            let res = self.arena.iter().find_map(|(id, data)| {
+                if data.eq_ignoring_origin(crate_data) {
+                    if data.origin.is_lib() && crate_data.origin.is_local() {
```
Veykril, line 696:
```rs
match (&data.origin, &crate_data.origin) {...}
```
When the origins are the same => simple equivalency check for dependencies
when one of them is a workspace member (is_local) and the other is not (is_lib) => equivalency check with the dev-deps ignored. In this case, if they are considered equivalent, keep the `is_local` one, as the deps may contain dev deps

crates/base-db/src/input.rs
alibektas: In my latest commit I changed `it.kind != DependencyKind::Normal` to `it.kind == DependencyKind::Normal`. I keep thinking about this and I think I did the right thing, didn't I :smile:

alibektas:
Forgot to move dev deps from lib crate to the local, so please wait for the next commit before reviewing

alibektas:
This is ready to be reviewed a third time

crates/base-db/src/input.rs
```diff
@@ -657,10 +766,34 @@ impl ops::Index<CrateId> for CrateGraph {
...
-    fn add_dep(&mut self, dep: Dependency) {
+    /// Add a dependency to `self` without checking if the dependency
+    // is existent among `self.dependencies`.
+    fn add_dep_unchecked(&mut self, dep: Dependency) {
         self.dependencies.push(dep)
     }
+
+    /// Add `deps` to `self` if the dependency is not already listed.
+    /// Finally returning an `Err` propagating the dependencies it couldn't add.
+    fn add_dep(&mut self, deps: Vec<Dependency>) -> Result<(), ExistingDepsError> {
```
alibektas, line 780: - Previously the same named function had been adding dependencies to self without checking if the dependency existed. That  function is now called `add_dep_unchecked`. 
- add_dep is not optimal for what it aims to do : I still do not know how well iterators optimize themselves so I just added the most naive linear search code hoping that you could enlighten me on how much/less I should be putting my trust on the compiler.
- Why add_dep return a `Result` : I don't know but I felt like we need to know if a dependency cannot be added.

crates/cfg/src/lib.rs
```diff
@@ -58,6 +58,10 @@ impl CfgOptions {
         self.enabled.insert(CfgAtom::KeyValue { key, value });
     }
 
+    pub fn diff<'a>(&'a self, other: &'a CfgOptions) -> HashSet<&CfgAtom> {
+        self.enabled.difference(&other.enabled).collect()
+    }
```
Veykril, line 63:
```suggestion
    pub fn difference<'a>(&'a self, other: &'a CfgOptions) -> impl Iterator<Item = &'a CfAtom> + 'a {
        self.enabled.difference(&other.enabled)
    }
```
we can use an opaque return type to save on the allocation here

crates/base-db/src/input.rs
```diff
@@ -324,6 +328,64 @@ pub struct CrateData {
...
+        }
+
+        if self.version != other.version {
+            return false;
+        }
+
+        let mut opts = self.cfg_options.diff(&other.cfg_options).into_iter();
+        match opts.len() {
+            0 => (),
+            1 => {
```
Veykril, line 359: (Given my change to `difference` below): Instead of then collecting the iterator here agin to get the length, we can just do two calls to `next()` and check i they are some / none and branch that way

crates/base-db/src/input.rs
Veykril: Yes and no actually. You are right that my comparison was wrong, but yours is also still wrong :D Yours ignores dev and build dependencies, we want to ignore only dev deps. so that shouldve been `it.kind != DependencyKind::Dev`

crates/base-db/src/input.rs
```diff
@@ -574,23 +653,51 @@ impl CrateGraph {
...
+                        if data.eq_ignoring_origin_and_deps(&crate_data, false) {
+                            return Some((id, false));
+                        }
                     }
-                },
-            );
-            if let Some(res) = res {
+                    (a @ CrateOrigin::Local { .. }, CrateOrigin::Library { .. })
+                    | (a @ CrateOrigin::Library { .. }, CrateOrigin::Local { .. }) => {
+                        // See #15656 for a relevant example.
```
Veykril, line 670:
```suggestion
                        // If the origins differ, check if the two crates are equal without considering the dev dependencies, if they are, they most likely are in different loaded workspaces which may cause issues. We keep the local version and discard the library one as the local version may have dev-dependencies that we want to keep resolving. See #15656 for more information.
```

crates/base-db/src/input.rs
```diff
@@ -574,23 +653,51 @@ impl CrateGraph {
...
+
+            if let Some((res, should_update_lib_to_local)) = res {
                 id_map.insert(topo, res);
+                if should_update_lib_to_local {
+                    let origin_old = self.arena[res].origin.clone();
+                    assert!(origin_old.is_lib());
+
+                    if let CrateOrigin::Library { repo, name } = origin_old {
+                        self.arena[res].origin = CrateOrigin::Local { repo, name: Some(name) };
+                    }
```
Veykril, line 689:
```suggestion
                    assert!(self.arena[res].origin.is_lib());
                    assert!(crate_data.origin.is_local());
                    self.arena[res].origin = crate_data.origin.clone()
```

crates/base-db/src/input.rs
```diff
@@ -574,23 +653,51 @@ impl CrateGraph {
...
+
+                    // Move local's dev dependencies into the newly-local-formerly-lib crate.
+                    let dev_deps = crate_data
+                        .dependencies
+                        .clone()
+                        .into_iter()
+                        .filter(|dep| dep.kind() == DependencyKind::Dev)
+                        .collect::<Vec<Dependency>>();
+
+                    self.arena[res].add_dep(dev_deps).unwrap_or_default();
```
Veykril, line 699:
```suggestion
                    self.arena[res].dependencies = crate_data.dependencies.clone();
```

crates/base-db/src/input.rs
Veykril: We can remove this again after my proposed changes, but as a note, you overcomplicated your implementation here. Given the invariant, if you insert a dep there should only ever be one duplicate at most as you are preventing duplicates from entering the deps here in the first place. So instead of a vec return value in the err case, a single element would've sufficed.

Veykril (approved):

Veykril:
That last commit contains nothing?

Veykril:
(also `retry` only works when the PR history has not been changed since the last `r+`)

lnicola:
Yeah, last commit is empty, but there was a force-push that did change some paths: https://github.com/rust-lang/rust-analyzer/compare/7dabf2718ef3724d12d9999c120d21df343253c6..32f382e4e8ea841da991ff3c3faade3e4d5052c5.

crates/base-db/src/input.rs
```diff
@@ -351,26 +411,43 @@ impl Env {
     }
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum DependencyKind {
+    Normal,
+    Dev,
+    Build,
+}
```
matklad, line 419: Please don't take this too seriously, as I have retreated to my Zig ivory tower in Lisbon some years ago and am mostly out of touch with the current goings of rust-analyzer, but I would say this is a significant architectural bug. 

By design, this layer of rust-analyzer knows nothing about cargo specific concepts. dev-vs-build-vs normal is 100% Cargo concept. `rustc` knows nothing about these words. As such, _any_ special handling of these concepts should happen in `workspace.rs`, not here. 

The motivation for this design is two-fold:

- practically, Cargo is _a_ build system, there might be others. In the core analysis parts, we shouldn't be prioritizing Cargo over other (current _and future_) build systems
- theoretically, it is important that the internal objects in rust-analyzer's semantic model reflect the physical reality of rustc. In that physical reality, cargo packages do not exists (and `-dev` as a concept applies to a package), there are only units of compilation.

cc @Veykril

Veykril: > Please don't take this too seriously, 

Not at all, you are completely right, this seems incorrect. This was a mistake on my part (I proposed this change somewhere I believe, might've been in DMs).

alibektas: In order to check for equality we needed to filter out dev dependencies and thus we ended up with the current state of things.  The alternatives were 

1. To filter them in `project_model` : I am looking into this since @matklad 's first comment here, I do not think it is going to be easy.
2. Adding `project-model` as a dependency of `base-db` which is ofc super wrong to do. ( As a matter of fact we already do have a `DependencyKind` under `project-model` named `DepKind` )

alibektas: But let me say that I also agree with matklad's concerns so I will look for a way to rectify this.

PR #15756, opened 2023-10-14
fix typo in mbe/lib.rs comment

harpsword:
fix typo in mbe/lib.rs comment


lnicola:
Thanks. But I think it's supposed to say "tests" 🙂.

@bors delegate+

PR #15760, opened 2023-10-14
make mir::ProjectionStore-impls pub-accessible

ChristianSchott:
When using RA as a crate the `mir::Place` `projection` is accessible, however there is no way to translate the `ProjectionId` to a `&[PlaceElem]`, as the `ProjectionId::lookup` is private.

Personally, I would only need the `ProjectionId::lookup`-fn to be `pub`, but I don't see any reason why the others should be kept private.. am I missing something @HKalbasi ?

Relates to: https://github.com/rust-lang/rust-analyzer/pull/15575

### DIFF
--- a/crates/hir-ty/src/mir.rs
+++ b/crates/hir-ty/src/mir.rs
@@ -243,16 +243,16 @@ impl Default for ProjectionStore {
 }
 
 impl ProjectionStore {
-    fn shrink_to_fit(&mut self) {
+    pub fn shrink_to_fit(&mut self) {
         self.id_to_proj.shrink_to_fit();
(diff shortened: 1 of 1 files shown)


crates/hir-ty/src/mir.rs
```diff
@@ -267,13 +267,13 @@ impl ProjectionStore {
 }
 
 impl ProjectionId {
-    const EMPTY: ProjectionId = ProjectionId(0);
+    pub const EMPTY: ProjectionId = ProjectionId(0);
```
ChristianSchott, line 270: not 100% sure about this, as it can be accessed indirectly via `From<LocalId>`..

PR #15762, opened 2023-10-15
fix: Fix metrics workflow to use deploy key

Veykril:
I think this should be correct


Veykril:
Well that didn't work

PR #15779, opened 2023-10-18
Add command for only opening external docs and attempt to fix vscode-remote issue

Veykril:
opening URI in a remote env causes vscode to ask the OS to handle `vscode-remote` URIs as there is no handler registered for such a scheme. This attempts to instruct vscode to handle those.

This is untested, as I can't figure out how to open a debug session on WSL rn.

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -249,6 +249,11 @@
                 "category": "rust-analyzer"
             },
             {
+                "command": "rust-analyzer.openExternalDocs",
+                "title": "Open External Docs",
+                "category": "rust-analyzer"
+            },
+            {
                 "command": "rust-analyzer.openCargoToml",
                 "title": "Open Cargo.toml",
                 "category": "rust-analyzer"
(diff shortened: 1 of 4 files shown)


gdziadkiewicz:
Hi @Veykril ,
Thanks for your punch aimed at solving this issue!

I encountered this `vscode-remote://` bug today while trying to open docs in WSL2. I managed to investigate (built and ran rust-analyzer in WSL2 with a debugger plus registered for `vscode-remote://` with Windows registry ) it and got:
1. The whole ULR passed to Windows: `vscode-remote://wsl+ubuntu/home/grzeg/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc/rust/html/std/io/stdio/struct.Stdin.html#method.read_to_string`. It's missing the `vscode-remote://` -> `vscode://vscode-remote/` replacement you implemented.
2. Reason why your fix does not work for my and possibly all such cases. When your workaround interacts with the URL it is in the form of `file:///home/grzeg/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc/rust/html/std/io/stdio/struct.Stdin.html#method.read_to_string`. The transformation to `vscode-remote://wsl+ubuntu/home/grzeg/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc/rust/html/std/io/stdio/struct.Stdin.html#method.read_to_string` is done by the vscode API call `vscode.env.openExternal`

In my opinion, it looks like a problem with the way `vscode.env.openExternal` acts. I will dig more and probably ask vscode for their input.

PR #15780, opened 2023-10-18
fix: import trait if needed for `unqualify_method_call` assist

Young-Flash:
before:

before

after:

after

follow up https://github.com/rust-lang/rust-analyzer/pull/13825

### DIFF
--- a/crates/ide-assists/src/handlers/unqualify_method_call.rs
+++ b/crates/ide-assists/src/handlers/unqualify_method_call.rs
@@ -89,31 +89,38 @@ fn add_import(
     ctx: &AssistContext<'_>,
     edit: &mut ide_db::source_change::SourceChangeBuilder,
 ) {
-    // for `<i32 as std::ops::Add>`
-    let path_type =
-        qualifier.segment().unwrap().syntax().children().filter_map(ast::PathType::cast).last();
-    let import = match path_type {
-        Some(it) => it.path().unwrap(),
-        None => qualifier,
-    };
-
-    // in case for `<_>`
-    if import.coloncolon_token().is_none() {
-        return;
-    }
-
-    let scope = ide_db::imports::insert_use::ImportScope::find_insert_use_container(
-        import.syntax(),
-        &ctx.sema,
-    );
-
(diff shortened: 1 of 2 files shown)


crates/ide-assists/src/handlers/unqualify_method_call.rs
```diff
@@ -76,10 +79,49 @@ pub(crate) fn unqualify_method_call(acc: &mut Assists, ctx: &AssistContext<'_>)
...
         },
     )
 }
 
+fn add_import(
+    path: &ast::Path,
+    ctx: &AssistContext<'_>,
+    edit: &mut ide_db::source_change::SourceChangeBuilder,
+) {
+    let qualifier = path.qualifier().unwrap();
```
lnicola, line 92: Let's try to avoid these `unwrap`s. I'm not sure if they can trigger, but it's probably fine to `return` in these cases.

crates/ide-assists/src/handlers/unqualify_method_call.rs
```diff
@@ -76,10 +79,49 @@ pub(crate) fn unqualify_method_call(acc: &mut Assists, ctx: &AssistContext<'_>)
...
+    // for `<i32 as std::ops::Add>`
+    let path_type =
+        qualifier.segment().unwrap().syntax().children().filter_map(ast::PathType::cast).last();
+    let import = match path_type {
+        Some(it) => it.path().unwrap(),
+        None => qualifier,
+    };
+
+    // in case for `<_>`
+    match import.coloncolon_token() {
```
lnicola, line 103:
```suggestion
    if import.coloncolon_token().is_none() {
        return;
    }
```

crates/ide-assists/src/handlers/unqualify_method_call.rs
Young-Flash: Actually it will not failed at this `unwrap`, casuse line 44 have (`_ = path.qualifier()?;`) verified. I will refactor it later.

Young-Flash:
ready for review

cc @lnicola

PR #15788, opened 2023-10-22
feat: import trait with alias

Young-Flash:
import_trait_with_alias

cc @Veykril

close https://github.com/rust-lang/rust-analyzer/issues/15684

### DIFF
--- a/crates/ide-assists/src/handlers/auto_import.rs
+++ b/crates/ide-assists/src/handlers/auto_import.rs
@@ -5,7 +5,7 @@ use ide_db::{
     helpers::mod_path_to_ast,
     imports::{
         import_assets::{ImportAssets, ImportCandidate, LocatedImport},
-        insert_use::{insert_use, ImportScope},
+        insert_use::{insert_use, insert_use_as_alias, ImportScope},
     },
 };
 use syntax::{ast, AstNode, NodeOrToken, SyntaxElement};
@@ -129,10 +129,12 @@ pub(crate) fn auto_import(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<
     for import in proposed_imports {
         let import_path = import.import_path;
 
+        let (assist_id, import_name) =
+            (AssistId("auto_import", AssistKind::QuickFix), import_path.display(ctx.db()));
         acc.add_group(
             &group_label,
-            AssistId("auto_import", AssistKind::QuickFix),
-            format!("Import `{}`", import_path.display(ctx.db())),
+            assist_id,
+            format!("Import `{}`", import_name),
             range,
             |builder| {
                 let scope = match scope.clone() {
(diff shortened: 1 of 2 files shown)


Young-Flash:
can anyone help review this?

thanks

crates/ide-db/src/imports/insert_use.rs
```diff
@@ -157,6 +157,33 @@ impl ImportScope {
 
 /// Insert an import path into the given file/node. A `merge` value of none indicates that no import merging is allowed to occur.
 pub fn insert_use(scope: &ImportScope, path: ast::Path, cfg: &InsertUseConfig) {
+    insert_use_with_alias_option(scope, path, cfg, None);
+}
+
+pub fn insert_use_as_alias(scope: &ImportScope, path: ast::Path, cfg: &InsertUseConfig) {
+    let text: &str = "use foo as _";
+    let parse = syntax::SourceFile::parse(text);
+    let node = match parse.tree().syntax().descendants().find_map(UseTree::cast) {
```
lnicola, line 166: Let's just `.expect` here, it's easier to read and it shouldn't fail anyway.

crates/ide-assists/src/handlers/auto_import.rs
```diff
@@ -723,7 +757,7 @@ fn main() {
             }
             ",
             r"
-            use test_mod::TestTrait;
+            use test_mod::TestTrait as _;
```
lnicola, line 760: We're still offering both options, right? What's causing us to pick the alias one in tests?

crates/ide-db/src/imports/insert_use.rs
```diff
@@ -175,8 +202,12 @@ pub fn insert_use(scope: &ImportScope, path: ast::Path, cfg: &InsertUseConfig) {
         };
     }
 
-    let use_item =
-        make::use_(None, make::use_tree(path.clone(), None, None, false)).clone_for_update();
+    let use_item = if alias.is_some() {
```
lnicola, line 205: Why not:

```rust
    let use_item =
        make::use_(None, make::use_tree(path.clone(), None, alias, false)).clone_for_update();
```

and is `clone_for_update` necessary there?

lnicola:
Sorry, I marked the wrong PR as draft by mistake.

Young-Flash:
> Was `check_assist` picking the last one? I didn't look.

`check_assist` doesn't specify `label`, when use `check_assist`, `assist_label` is `None`.

https://github.com/rust-lang/rust-analyzer/blob/a2e2881676b52f6d8bd4e4b38cd169e57790013f/crates/ide-assists/src/tests.rs#L55-L58
and it just get the `pop` out one
https://github.com/rust-lang/rust-analyzer/blob/a2e2881676b52f6d8bd4e4b38cd169e57790013f/crates/ide-assists/src/tests.rs#L184-L187

here should use `check_assist_by_label` instead.

PR #15809, opened 2023-10-27
feat: make extract_variable assist in place

Young-Flash:
extract_variable_without_select

close https://github.com/rust-lang/rust-analyzer/issues/15796


HKalbasi (approved):
Can you make the assist inapplicable if the expression has type `()`?

PR #15812, opened 2023-10-29
Fix some typos

cuishuang:


crates/ide-db/src/generated/lints.rs
```diff
@@ -10462,7 +10462,7 @@ if the `never_type` is enabled."##,
     },
     Lint {
         label: "clippy::empty_line_after_doc_comments",
-        description: r##"Checks for empty lines after documenation comments."##,
+        description: r##"Checks for empty lines after documentation comments."##,
```
lnicola, line 10465: These are or were misspelled in clippy, let's drop them from here.

cuishuang: Thanks. Done

lnicola: There's another one in the same file 😀

cuishuang: Sorry for my lack of rigor. 

Already edited

Issue #15648, opened 2023-09-21
Rust Analyzer doesn't work with large minimized files


TCROC:
**rust-analyzer version**:

rust-analyzer version: 0.3.1665-standalone

**rustc version**:

rustc 1.72.1 (d5c2e9c34 2023-09-13)

Example Project: https://github.com/TCROC/rust-analyzer-gdnative-issue-example

The Godot-Rust gdnative extension here: https://github.com/godot-rust/gdnative

Gdnative generates bindings for the Godot engine.  It does this in 1 of 3 ways:

1. Generating all of the bindings in a single minimized file.  This results in faster compile times when building against the bindings.  Rust analyzer does not work with this file.

2. Generating all of the bindings per class / struct in separate minimized files.  This seems to have similar build times.  Rust analyzer does work with this file.

3. Generating all of the bindings per class / struct in separate files and formatting those files.  Rust analyzer works on these files, but the compile times are very slow.

Fortunately I can use options 2 and 3 as workarounds for now.  But I feel that rust analyzer should work on a single minified file as well.  The linked repo contains examples for 1 (the issue) and 2 (the best workaround for now).

I created an example project here with a README explaining how to reproduce:

https://github.com/TCROC/rust-analyzer-gdnative-issue-example

Here is the README for convenient reading without clicking the above link:

# broken-example

This folder illustrates the issue.  If you open it in vs code: https://github.com/microsoft/vscode , you will see that the auto completions and comment hovers are not working.  Here is a screenshot to demonstrate:

![image](https://github.com/TCROC/rust-analyzer-gdnative-issue-example/assets/24307049/001b8cab-738b-4886-855b-ef5a202a6ee3)

Cargo.toml

```toml
[package]
name = "broken-example"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
gdnative = "0.11.3"
gdnative-bindings = "0.11.3"
```

# working-example

This folder illustrates a workaround that can be applied with the gdnative-bindings library.  If you enable the "one-class-one-file" feature:

```
gdnative-bindings = { version = "0.11.3", features = [ "one-class-one-file" ] }
```

Then the generated rust code will be broken up into multiple files instead of compressed in one file.  Now the rust-analyzer works properly:

[...]

PR #15819, opened 2023-11-01
feat: skip checking token tree count for include! macro call

### LINKED ISSUE #15335: Include based generated sources from time-tz fails autocomplete
In projects which uses build-scripts to generate rust source, rust-analyzer fails to provide any autocomplete or type information. Link to an example GitHub project that reproduces the issue: https://github.com/Yuri6037/time-tz.

**rust-analyzer version**: `0.4.1598-standalone (99718d0c8 2023-07-22)`

**rustc version**: `rustc 1.71.0 (8ede3aae2 2023-07-12)`

**relevant settings**: macOS Ventura, rustup install, no specific environment variable.

wasd96040501:
fix #15335 #15648

### DIFF
--- a/crates/hir-expand/src/db.rs
+++ b/crates/hir-expand/src/db.rs
@@ -614,9 +614,25 @@ fn macro_expand(db: &dyn ExpandDatabase, id: MacroCallId) -> ExpandResult<Arc<tt
         err = error.clone().or(err);
     }
 
-    // Set a hard limit for the expanded tt
-    if let Err(value) = check_tt_count(&tt) {
-        return value;
+    // Skip checking token tree limit for include! macro call
+    let skip_check_tt_count = match loc.kind {
+        MacroCallKind::FnLike { ast_id, expand_to: _ } => {
+            if let Some(name_ref) =
+                ast_id.to_node(db).path().and_then(|p| p.segment()).and_then(|s| s.name_ref())
+            {
+                name_ref.text() == "include"
+            } else {
+                false
+            }
+        }
+        _ => false,
+    };
+
+    if !skip_check_tt_count {
+        // Set a hard limit for the expanded tt
+        if let Err(value) = check_tt_count(&tt) {
+            return value;
+        }
     }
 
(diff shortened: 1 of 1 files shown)


crates/hir-expand/src/db.rs
```diff
@@ -614,9 +614,25 @@ fn macro_expand(db: &dyn ExpandDatabase, id: MacroCallId) -> ExpandResult<Arc<tt
...
-    // Set a hard limit for the expanded tt
-    if let Err(value) = check_tt_count(&tt) {
-        return value;
+    // Skip checking token tree limit for include! macro call
+    let skip_check_tt_count = match loc.kind {
+        MacroCallKind::FnLike { ast_id, expand_to: _ } => {
+            if let Some(name_ref) =
+                ast_id.to_node(db).path().and_then(|p| p.segment()).and_then(|s| s.name_ref())
+            {
+                name_ref.text() == "include"
```
lnicola, line 623: We could also do something like:

```rust
                name_ref.text().as_str()
                    == name![include].as_str().expect("include is not a tuple field")
```

(not sure if that's the best way to compare them)

lnicola: Or `name_ref.as_name() == name![include]`, I doubt this is performance-critical.

lnicola:
Thanks a lot, I was actually planning to look into this!

crates/hir-expand/src/db.rs
```diff
@@ -614,9 +618,25 @@ fn macro_expand(db: &dyn ExpandDatabase, id: MacroCallId) -> ExpandResult<Arc<tt
...
+            if let Some(name_ref) =
+                ast_id.to_node(db).path().and_then(|p| p.segment()).and_then(|s| s.name_ref())
+            {
+                name_ref.as_name() == name!(include)
+            } else {
+                false
+            }
+        }
+        _ => false,
+    };
```
Veykril, line 633: This should check that the `loc.def` is the builtin include macro instead of doing a name check. Otherwise any macro called `include` skips the token limit check

PR #15825, opened 2023-11-02
fix: better resolve assoc item with type bound'

### LINKED ISSUE #15772: `impl<T: Deref<Target = impl Trait>>`
```
rust-analyzer version: 0.4.1696-standalone (6572ec8d9 2023-10-15)
rustc 1.73.0 (cc66ad468 2023-10-03)
```

It seems any use of an `impl<T: Deref<Target = impl Trait>>` on some struct does not show methods even when using an appropriate `T`. Consider the following minimal repro:
```rust
use std::ops::Deref;

fn main() {
    let foo: Foo<&u8> = Foo::new(&42_u8);
    foo.foobar();
    foo.baz();
}

trait Bar {
    fn bar(&self);
}
impl Bar for u8 {
    fn bar(&self) {}
}

struct Foo<F> {
    foo: F,
}

impl<F> Foo<F> {
    fn new(foo: F) -> Foo<F> {
        Foo { foo }
    }
    fn baz(&self) {}
}

impl<F: Deref<Target = impl Bar>> Foo<F> {
    fn foobar(&self) {
        self.foo.deref().bar()
    }
}
```

This compiles [(playground link)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8796dd9df3fde5b673e6b17602ea1cda). However, the method list I see does not include `foobar`, despite the bounds being satisfied.
image

Austaras:
Closes #15772


lnicola:
Umm, r? @flodiebold maybe (no pressure).

crates/hir-ty/src/method_resolution.rs
```diff
@@ -1478,26 +1478,38 @@ fn is_valid_fn_candidate(
...
+
+                if let WhereClause::AliasEq(_) = p {
+                    alias.push(p);
+                } else {
+                    other_predicate.push(p);
+                }
+            }
+
+            for p in alias {
+                if !table.try_resolve_alias(p.cast(Interner)) {
```
flodiebold, line 1501: I think I would prefer if this applied the solution here, instead of hiding that inside `try_resolve_alias`. Also, is there a reason to only do this for aliases? What does rustc do here?

crates/ide-completion/src/completions/dot.rs
```diff
@@ -1095,4 +1095,81 @@ fn test(s: S<Unknown>) {
             "#]],
         );
     }
+
+    #[test]
+    fn assoc_impl_1() {
```
flodiebold, line 1100: it would be good to at least additionally have a test for this in `hir-ty` as well.

crates/hir-ty/src/method_resolution.rs
Austaras: > I think I would prefer if this applied the solution here

I'd also like to, but `apply_solution` is not public here.

> Also, is there a reason to only do this for aliases?

There are four kinds of `WhereClause`, and I believe only `AliasEq` would generate solution. But it does no harm to check everything for `Solution`.

flodiebold: I think making `apply_solution` `pub(crate)` would be fine.

> There are four kinds of WhereClause, and I believe only AliasEq would generate solution. But it does no harm to check everything for Solution.

Yeah, I think it's better to just do it for everything.

flodiebold:
@bors r+

thanks!

PR #15827, opened 2023-11-02
VSCode metadata. category:formatters

notpeter:
If you invoke cmd+p `Format Document` without rust-analyzer installed in the workspace, VSCode suggests you "install a formatter" which triggers a search for `category:formatters rust`. Sadly rust-analyzer extension is not suggested.

Tweak VSCode Extension Metadata so rust-analyzer shows up in 'category:formatters rust' search.

You can see the valid values for categories in the [extension manifest: field reference](https://code.visualstudio.com/api/references/extension-manifest#fields) docs.

search2
search_1

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -18,7 +18,6 @@
     ],
     "categories": [
         "Formatters",
-        "Linters",
         "Programming Languages"
     ],
     "capabilities": {
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -17,6 +17,8 @@
         "rust"
     ],
     "categories": [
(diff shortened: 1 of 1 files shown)


lnicola:
Thanks. Are we a linter too?

notpeter:
> Thanks. Are we a linter too?

I guess not. :D

lnicola:
It was an honest question. I know we're a formatter because "Format Document" works, is there a way to trigger an installed linter extension?

We _can_ be a linter if you set up Clippy as the `check` command. I don't know if that counts.

In the meanwhile,

@bors r+

PR #15830, opened 2023-11-03
code: expose workspaces to other extensions; remove `addProject` command

davidbarsky:
This (mostly red) PR does three things:
- Exposes two methods to companion extensions (`setWorkspaces` and `notifyRustAnalyzer`). 
    - `setWorkspaces` is needed to update `linkedProjects` _without_ writing workspace/global configuration.
    - `notifyRustAnalyzer` to get the server to pull the new configuration.
- Makes `Ctx` implement `RustAnalyzerExtensionApi` to prevent accidental regressions.
- Remove `rust-analyzer.addProject`, as that will live in a buck2 companion extension. No need for that to be in rust-analyzer!

I can see the utility of combining `notifyRustAnalyzer` and `setWorkspaces` into a single method (`updateWorkspacesAndNotify()`?), but I don't feel strongly about this. My feeling is that this API could be easily changed in the future.

### DIFF
--- a/editors/code/src/ctx.ts
+++ b/editors/code/src/ctx.ts
@@ -1,5 +1,5 @@
 import * as vscode from "vscode";
-import type * as lc from "vscode-languageclient/node";
+import * as lc from "vscode-languageclient/node";
 import * as ra from "./lsp_ext";
 
 import { Config, prepareVSCodeConfig } from "./config";
@@ -22,6 +22,7 @@ import {
 import { execRevealDependency } from "./commands";
 import { PersistentState } from "./persistent_state";
 import { bootstrap } from "./bootstrap";
+import type { RustAnalyzerExtensionApi } from "./main";
 
 // We only support local folders, not eg. Live Share (`vlsl:` scheme), so don't activate if
 // only those are in use. We use "Empty" to represent these scenarios
@@ -64,7 +65,7 @@ export type CtxInit = Ctx & {
     readonly client: lc.LanguageClient;
 };
 
-export class Ctx {
+export class Ctx implements RustAnalyzerExtensionApi {
     readonly statusBar: vscode.StatusBarItem;
     config: Config;
     readonly workspace: Workspace;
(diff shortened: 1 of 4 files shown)


davidbarsky:
cc: @Wilfred @P1n3appl3.

lnicola (approved):

davidbarsky:
@lnicola don't land this yet, sorry! I'm working on a (quick) change that modifies `discoverWorkspaceCommand` to accept a array of `vscode.Uri`. I think I'd prefer to reduce the churn on y'all by sneaking that in here!

(I should've commented sooner!)

lnicola:
Yeah, no worries. Veykril might also want to take a look when he's back.

editors/code/src/ctx.ts
```diff
@@ -190,7 +191,7 @@ export class Ctx {
                 const command = `${this.config.discoverProjectRunner}.discoverWorkspaceCommand`;
                 log.info(`running command: ${command}`);
                 const project: JsonProject = await vscode.commands.executeCommand(command);
-                this.addToDiscoveredWorkspaces([project]);
+                this.setWorkspaces([project]);
```
JakobDegen, line 194: Maybe I'm not reading this correctly (it's quite possible I am) - but is the implication of this that if I am currently working in `project1/foo.rs` and then open a new `.rs` file in `project2/bar.rs`, I will lose r-a support in project1?

davidbarsky: you will not lose IDE functionality. the code _here_, in particular, runs only once when rust-analyzer is starting up.

I have changes on a branch that make the companion extension fully control rust-analyzer's view of `linkedProjects`. Relatedly, rust-project (over in `buck2/integrations/rust-project`) will key/control each workspace (the generated `rust-project.json`) based on the target label or the target's membership in the workspace functionality you added.

davidbarsky: The relationship between the rust-analyzer extension and the companion extension is a little weird. At startup, the rust-analyzer extension will pull data from the companion extension, but in the steady state, the companion extension will _push_ changes to the rust-analyzer extension. The code here only runs at startup.

davidbarsky:
@lnicola depending on your interest level, i've updated this PR to include the changes I wanted to.

(let me know if I shouldn't @ you!)

editors/code/src/main.ts
```diff
@@ -11,6 +11,9 @@ const RUST_PROJECT_CONTEXT_NAME = "inRustProject";
 
 export interface RustAnalyzerExtensionApi {
```
Veykril, line 12: Can we put a comment on this that the API exposed here is not to be assumed stable or similar?

Veykril (approved):
lgtm with a the comment nit

Issue #15581, opened 2023-09-08
New assist: Generate `IndexMut` implementation from `Index` implementation


kpreid:
Given an existing `core::ops::Index` implementation, this assist would generate a copy of it where

* `Index` becomes `IndexMut`
* The method signature is updated
* `&` in the return value position in the function body (and perhaps other obviously-relevant positions) is replaced with `&mut`

For an example from the code I'm working on right now,

```rust
pub enum Axis { X = 0, Y = 1, Z = 2 }

impl<T> ops::Index<Axis> for [T; 3] {
    type Output = T;

    fn index(&self, index: Axis) -> &Self::Output {
        &self[index as usize]
    }
}
```

could be used to generate 

```rust
impl<T> ops::IndexMut<Axis> for [T; 3] {
    fn index_mut(&mut self, index: Axis) -> &mut Self::Output {
        &mut self[index as usize]
    }
}
```

with no manual changes instead of 6. There might be other ref/mut trait pairs that could be assisted similarly, too.

Veykril:
`Borrow`/`BorrowMut`, `AsRef`/`AsMut` come to mind here as well

Young-Flash:
I'd like to have a try

@rustbot claim

Young-Flash:
from 
```rust
impl<T> ops::Index<Axis> for [T; 3] {
    type Output = T;

    fn index(&self, index: Axis) -> &Self::Output {
        &self[index as usize]
    }
}
```
to generate
```rust
impl<T> ops::IndexMut<Axis> for [T; 3] {
    fn index_mut(&mut self, index: Axis) -> &mut Self::Output {
        &mut self[index as usize]
    }
}
```
this example is simple and easy to achieve, just replace `self` with `mut self`, but this 'replace' way isn't generic, given the following example
```rust
struct Map {
    data: HashMap<String, i32>   
}

impl Index<String> for Map {
    type Output = i32;

    fn index(&self, key: String) -> &Self::Output {
        self.data.get(&key).unwrap()
    }
}
```
there should generate
```rust
impl IndexMut<String> for Map {
    fn index_mut(&mut self, key: String) -> &mut Self::Output {
        self.data.get_mut(&key).unwrap()
    }
}
```
there problem here is that the `index_mut` method body may various, so there isn't a way to deal with it generally. how about just generate `index_mut` method body with a `todo!()` ? any suggestion?  @Veykril @kpreid 
```rust
impl IndexMut<String> for Map {
[...]

PR #15832, opened 2023-11-04
feat: add generate_mut_trait_impl assist

Young-Flash:
generate_mut_trait_impl

Generate proper `index_mut` method body refer to `index` method body may impossible due to the unpredicable case (#15581).

Here just leave the `index_mut` method body be same as `index` method body, user can modify it manually to meet their need.

### DIFF
--- a/crates/ide-assists/src/handlers/generate_mut_trait_impl.rs
+++ b/crates/ide-assists/src/handlers/generate_mut_trait_impl.rs
@@ -1,0 +1,195 @@
+use syntax::{
+    ast::{self, make},
+    ted, AstNode,
+};
+
+use crate::{AssistContext, AssistId, AssistKind, Assists};
+
+// Generate proper `index_mut` method body refer to `index` method body may impossible due to the unpredicable case [#15581].
+// Here just leave the `index_mut` method body be same as `index` method body, user can modify it manually to meet their need.
+
+// Assist: generate_mut_trait_impl
+//
+// Adds a IndexMut impl from the `Index` trait.
+//
+// ```
+// pub enum Axis { X = 0, Y = 1, Z = 2 }
+//
+// impl<T> Index$0<Axis> for [T; 3] {
+//     type Output = T;
+//
+//     fn index(&self, index: Axis) -> &Self::Output {
+//         &self[index as usize]
+//     }
+// }
+// ```
+// ->
+// ```
+// pub enum Axis { X = 0, Y = 1, Z = 2 }
+//
+// $0impl<T> IndexMut<Axis> for [T; 3] {
+//     fn index_mut(&mut self, index: Axis) -> &mut Self::Output {
+//         &self[index as usize]
+//     }
+// }
+//
+// impl<T> Index<Axis> for [T; 3] {
+//     type Output = T;
+//
+//     fn index(&self, index: Axis) -> &Self::Output {
+//         &self[index as usize]
+//     }
+// }
+// ```
+pub(crate) fn generate_mut_trait_impl(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
+    let impl_def = ctx.find_node_at_offset::<ast::Impl>()?.clone_for_update();
+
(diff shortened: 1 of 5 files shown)


crates/ide-assists/src/handlers/generate_mut_trait_impl.rs
```diff
@@ -0,0 +1,195 @@
...
+//     }
+// }
+// ```
+pub(crate) fn generate_mut_trait_impl(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
+    let impl_def = ctx.find_node_at_offset::<ast::Impl>()?.clone_for_update();
+
+    let trait_ = impl_def.trait_()?;
+    if let ast::Type::PathType(trait_type) = trait_.clone() {
+        let trait_name = trait_type.path()?.segment()?.name_ref()?.to_string();
+        if trait_name != "Index" {
```
lnicola, line 50: Let's use `name![Index]` like in https://github.com/rust-lang/rust-analyzer/pull/15819/files.

Young-Flash: Should we should introduce the `hir-expand` dependency in `crates/ide-assists/Cargo.toml`?

lnicola: Ah, maybe not.

Young-Flash: hahaha so let's keep it in this `.to_string()` way?

crates/ide-assists/src/handlers/generate_mut_trait_impl.rs
```diff
@@ -0,0 +1,195 @@
+use syntax::{
+    ast::{self, make},
+    ted, AstNode,
+};
+
+use crate::{AssistContext, AssistId, AssistKind, Assists};
+
+// Generate proper `index_mut` method body refer to `index` method body may impossible due to the unpredicable case [#15581].
```
alibektas, line 8: Adding a `FIXME` tag would be a good idea.

crates/ide-assists/src/handlers/generate_mut_trait_impl.rs
```diff
@@ -0,0 +1,195 @@
...
+    let trait_ = impl_def.trait_()?;
+    if let ast::Type::PathType(trait_type) = trait_.clone() {
+        let trait_name = trait_type.path()?.segment()?.name_ref()?.to_string();
+        if trait_name != "Index" {
+            return None;
+        }
+    }
+
+    // Index -> IndexMut
+    let index_trait = impl_def
```
alibektas, line 56: We need to check right after we figured out what the `ast::Type` is, whether `Index` actually comes from `std::ops::Index`. See [this](https://github.com/rust-lang/rust-analyzer/blob/69432287cbaa0cd69dd7219aa0c6c74e44dbe7e7/crates/ide-assists/src/handlers/into_to_qualified_from.rs#L50) as a reference.

crates/ide-assists/src/handlers/generate_mut_trait_impl.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    #[test]
+    fn test_generate_mut_trait_impl_not_applicable() {
+        check_assist_not_applicable(
+            generate_mut_trait_impl,
+            r#"
+impl<T> Add$0<i32> for [T; 3] {}
+"#,
+        );
+    }
```
alibektas, line 194: Adding a test where Index doesn't come from std::ops::Index would be nice.

alibektas (changes requested):

alibektas (approved):
I am happy with the current version of things, but let me say that I can't really be trusted in the review business as I am quite new to the project. So I leave the judgment to @lnicola

lnicola (approved):

lnicola:
@bors r+

Thanks, Young-Flash and alibektas!

PR #15838, opened 2023-11-05
fix:Diagnostics calculated from included files report them for the wrong file

### LINKED ISSUE #14968: Diagnostics calculated from `include`d files report them for the wrong file
Our diagnostics infra was written with the idea in mind that dianostics reported from macro expansions will only end up in the calling file and as such we never verify that the upmapped ranges belong to the file we calculate the diagnostics for. This assumption is incorrect, `include!` calls in fact have two macro inputs from differing files, as such, any diagnostic calculated from an included file right now will either report the diagnostic in some random location in the calling file or worse crash the server.

An example crash test case for `unresolved_macro_call` diagnostic:
```rs
    #[test]
    fn include_does_not_break_diagnostics() {
        check_diagnostics(
            r#"
//- minicore: include
//- /lib.rs crate:lib
include!("include-me.rs");
//- /include-me.rs
/// long doc that pushes the diagnostic range beyond the first file's text length
#[err]
mod prim_never {}
"#,
        );
    }
```

wasd96040501:
fix #14968

### DIFF
--- a/crates/hir-expand/src/lib.rs
+++ b/crates/hir-expand/src/lib.rs
@@ -713,6 +713,7 @@ impl ExpansionInfo {
         &self,
         db: &dyn db::ExpandDatabase,
         token: InFile<&SyntaxToken>,
+        ignore_included_file: bool,
     ) -> Option<(InFile<SyntaxToken>, Origin)> {
         assert_eq!(token.file_id, self.expanded.file_id.into());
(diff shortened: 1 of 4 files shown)


Veykril:
Oh sorry, I fixed the issue in https://github.com/rust-lang/rust-analyzer/pull/16035, thanks for working on this anyways though!

PR #15846, opened 2023-11-07
editor/code: add option to suppress internal error notifications

jprochazk:
Fixes https://github.com/rust-lang/rust-analyzer/issues/14193

- Added the `rust-analyzer.showRequestFailedErrorNotification` configuration option, which defaults to `true`
- If `rust-analyzer.showRequestFailedErrorNotification` is set to `true`, the current behavior is preserved.
- If `rust-analyzer.showRequestFailedErrorNotification` is set to `false`, no error toasts will be displayed for any of the failed requests caused by panics in r-a. This _only_ applies to events that are triggered "implicitly", such as `textDocument/hover`.

To test this, you can manually introduce a panic in one of the language server LSP handlers for non-command events. I added an explicit `panic!()` in the `textDocument/hover` event handler:

#### `rust-analyzer.showRequestFailedErrorNotification` set to `true` (default)

[2023-11-07 17-17-48.webm](https://github.com/rust-lang/rust-analyzer/assets/1665677/d0408ab8-79d1-42cf-a4e7-94e99d9783ec)

#### `rust-analyzer.showRequestFailedErrorNotification` set to `false`

[2023-11-07 17-16-49.webm](https://github.com/rust-lang/rust-analyzer/assets/1665677/0496d8d0-fb53-4bc6-a279-1a47f412dbdb)

### DIFF
--- a/editors/code/src/lang_client.ts
+++ b/editors/code/src/lang_client.ts
@@ -2,15 +2,25 @@ import * as lc from "vscode-languageclient/node";
 import * as vscode from "vscode";
 
 export class RaLanguageClient extends lc.LanguageClient {
-    override error(message: string, data?: any, showNotification?: boolean | "force"): void {
-        // ignore `Request TYPE failed.` errors
+    override handleFailedRequest<T>(
+        type: lc.MessageSignature,
+        token: vscode.CancellationToken | undefined,
+        error: any,
+        defaultValue: T,
+        showNotification?: boolean | undefined,
+    ): T {
         const showError = vscode.workspace
             .getConfiguration("rust-analyzer")
             .get("showRequestFailedErrorNotification");
-        if (!showError && message.startsWith("Request") && message.endsWith("failed.")) {
-            return;
+        if (
+            !showError &&
+            error instanceof lc.ResponseError &&
+            error.code === lc.ErrorCodes.InternalError
+        ) {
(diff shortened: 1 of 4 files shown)


editors/code/src/lang_client.ts
```diff
@@ -0,0 +1,16 @@
+import * as lc from "vscode-languageclient/node";
+import * as vscode from "vscode";
+
+export class RaLanguageClient extends lc.LanguageClient {
+    override error(message: string, data?: any, showNotification?: boolean | "force"): void {
```
Veykril, line 5: Do you know what data might contain? Does it contain the error code by any chance? Would be nice if we could match on that instead, because then we can filter out some specific errors that VSCode raises (on non panicking requests) unconditionally (and log them instead).

Veykril (commented):
Guess I was very much tunnel visioning when I looked into how to overwrite this given how simple the solution is after all. Thanks for looking into this!

editors/code/src/lang_client.ts
jprochazk: It seems like this is the `ResponseError` emitted by r-a, so it does contain the error code. I changed this to override `handleRequestFailed` instead of `error`, and set `showNotification` to `false` if we detect that it's an `InternalError`.

Veykril (approved):

PR #15857, opened 2023-11-09
fix: remove parenthesis should ensure space

Young-Flash:
close https://github.com/rust-lang/rust-analyzer/issues/15844

### DIFF
--- a/crates/ide-assists/src/handlers/remove_parentheses.rs
+++ b/crates/ide-assists/src/handlers/remove_parentheses.rs
@@ -1,4 +1,4 @@
-use syntax::{ast, AstNode};
+use syntax::{ast, AstNode, SyntaxKind, T};
 
 use crate::{AssistContext, AssistId, AssistKind, Assists};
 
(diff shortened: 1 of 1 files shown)


Young-Flash:
Can anyone help review this?

crates/ide-assists/src/handlers/remove_parentheses.rs
```diff
@@ -34,12 +34,27 @@ pub(crate) fn remove_parentheses(acc: &mut Assists, ctx: &AssistContext<'_>) ->
...
+        .token_at_offset(parens.syntax().text_range().start());
+    let need_to_add_ws = match token_at_offset {
+        syntax::TokenAtOffset::Between(before, _after) => {
+            // anyother `SyntaxKind` we missing here?
+            let tokens = vec![T![&], T![!], T!['('], T!['['], T!['{']];
+            before.kind() != SyntaxKind::WHITESPACE && !tokens.contains(&before.kind())
+        }
+        _ => false,
+    };
+    let expr = if need_to_add_ws { format!(" {}", expr) } else { expr.to_string() };
```
Veykril, line 50: Two things, first this "analysis" should go into the `|builder|` closure so its only executed if the user makes use of the assist (for perf reasons).

Second, we can do this simpler, we can just take the preceding token of the expression via `expr.syntax().first_token().and_then(|it| it.prev_token())` and check that for whitespace, if its not a whitespace token we should insert one.

Veykril: Ah seeing the commented out tests I just realized I was tunnel visioning on the `match(1+1)` thing, my approach doesnt generally work for other stuff I suppose hmm ...

Young-Flash: Sorry, I forgot to uncomment them after debug. Your approach works well over all the tests 👍

PR #15863, opened 2023-11-09
feature: Create `UnindexedProject` notification to be sent to the client

davidbarsky:
(Note that this branch contains commits from https://github.com/rust-lang/rust-analyzer/pull/15830, which I'll rebase atop of as needed.)

Based on the discussion in https://github.com/rust-lang/rust-analyzer/issues/15837, I've added a notification and off-by-default toggle to send that notification from `handle_did_open_text_document`. I'm happy to rename/tweak this as needed.

I've been using this for a little bit, and it does seem to cause a little bit more indexing/work in rust-analyzer, but it's something that I'll profile as needed, I think.

### DIFF
--- a/crates/rust-analyzer/src/global_state.rs
+++ b/crates/rust-analyzer/src/global_state.rs
@@ -33,7 +33,7 @@ use crate::{
     mem_docs::MemDocs,
     op_queue::OpQueue,
     reload,
-    task_pool::TaskPool,
+    task_pool::{TaskPool, TaskQueue},
 };
 
 // Enforces drop order
@@ -126,6 +126,17 @@ pub(crate) struct GlobalState {
         OpQueue<(), (Arc<Vec<ProjectWorkspace>>, Vec<anyhow::Result<WorkspaceBuildScripts>>)>,
     pub(crate) fetch_proc_macros_queue: OpQueue<Vec<ProcMacroPaths>, bool>,
     pub(crate) prime_caches_queue: OpQueue,
+
+    /// A deferred task queue.
+    ///
+    /// This queue is used for doing database-dependent work inside of sync
+    /// handlers, as accessing the database may block latency-sensitive
+    /// interactions and should be moved away from the main thread.
+    ///
+    /// For certain features, such as [`lsp_ext::UnindexedProjectParams`],
+    /// this queue should run only *after* [`GlobalState::process_changes`] has
+    /// been called.
+    pub(crate) deferred_task_queue: TaskQueue,
 }
 
 /// An immutable snapshot of the world's state at a point in time.
@@ -163,6 +174,11 @@ impl GlobalState {
             let (sender, receiver) = unbounded();
             let handle = TaskPool::new_with_threads(sender, 1);
             Handle { handle, receiver }
+        };
+
+        let task_queue = {
+            let (sender, receiver) = unbounded();
+            TaskQueue { sender, receiver }
         };
 
         let mut analysis_host = AnalysisHost::new(config.lru_parse_query_capacity());
@@ -208,6 +224,8 @@ impl GlobalState {
             fetch_proc_macros_queue: OpQueue::default(),
 
             prime_caches_queue: OpQueue::default(),
+
+            deferred_task_queue: task_queue,
         };
         // Apply any required database inputs from the config.
         this.update_configuration(config);
--- a/crates/rust-analyzer/src/handlers/notification.rs
+++ b/crates/rust-analyzer/src/handlers/notification.rs
@@ -70,7 +70,15 @@ pub(crate) fn handle_did_open_text_document(
         if already_exists {
             tracing::error!("duplicate DidOpenTextDocument: {}", path);
         }
+
         state.vfs.write().0.set_file_contents(path, Some(params.text_document.text.into_bytes()));
+        if state.config.notifications().unindexed_project {
+            tracing::debug!("queuing task");
+            let _ = state
+                .deferred_task_queue
+                .sender
+                .send(crate::main_loop::QueuedTask::CheckIfIndexed(params.text_document.uri));
+        }
     }
     Ok(())
 }
--- a/crates/rust-analyzer/src/main_loop.rs
+++ b/crates/rust-analyzer/src/main_loop.rs
@@ -1,5 +1,6 @@
 //! The main loop of `rust-analyzer` responsible for dispatching LSP
 //! requests/replies and notifications back to the client.
+use crate::lsp::ext;
 use std::{
     fmt,
     time::{Duration, Instant},
@@ -56,6 +57,7 @@ pub fn main_loop(config: Config, connection: Connection) -> anyhow::Result<()> {
 enum Event {
     Lsp(lsp_server::Message),
     Task(Task),
+    QueuedTask(QueuedTask),
     Vfs(vfs::loader::Message),
     Flycheck(flycheck::Message),
 }
@@ -67,13 +69,20 @@ impl fmt::Display for Event {
             Event::Task(_) => write!(f, "Event::Task"),
             Event::Vfs(_) => write!(f, "Event::Vfs"),
             Event::Flycheck(_) => write!(f, "Event::Flycheck"),
+            Event::QueuedTask(_) => write!(f, "Event::QueuedTask"),
         }
     }
 }
 
 #[derive(Debug)]
+pub(crate) enum QueuedTask {
+    CheckIfIndexed(lsp_types::Url),
+}
+
+#[derive(Debug)]
 pub(crate) enum Task {
     Response(lsp_server::Response),
+    ClientNotification(ext::UnindexedProjectParams),
     Retry(lsp_server::Request),
     Diagnostics(Vec<(FileId, Vec<lsp_types::Diagnostic>)>),
     PrimeCaches(PrimeCachesProgress),
@@ -115,6 +124,7 @@ impl fmt::Debug for Event {
         match self {
             Event::Lsp(it) => fmt::Debug::fmt(it, f),
             Event::Task(it) => fmt::Debug::fmt(it, f),
+            Event::QueuedTask(it) => fmt::Debug::fmt(it, f),
             Event::Vfs(it) => fmt::Debug::fmt(it, f),
             Event::Flycheck(it) => fmt::Debug::fmt(it, f),
         }
@@ -193,6 +203,9 @@ impl GlobalState {
             recv(self.task_pool.receiver) -> task =>
                 Some(Event::Task(task.unwrap())),
 
+            recv(self.deferred_task_queue.receiver) -> task =>
+                Some(Event::QueuedTask(task.unwrap())),
+
             recv(self.fmt_pool.receiver) -> task =>
                 Some(Event::Task(task.unwrap())),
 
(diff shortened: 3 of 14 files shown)


davidbarsky:
Did some profiling of this on my Mac, by the way:

davidbarsky:
CI failed because of a timeout in receiving a notification, which doesn't feel great.

crates/rust-analyzer/src/handlers/notification.rs
```diff
@@ -64,7 +64,14 @@ pub(crate) fn handle_did_open_text_document(
         if already_exists {
             tracing::error!("duplicate DidOpenTextDocument: {}", path);
         }
-        state.vfs.write().0.set_file_contents(path, Some(params.text_document.text.into_bytes()));
+        {
+            let vfs = &mut state.vfs.write().0;
+            vfs.set_file_contents(path.clone(), Some(params.text_document.text.into_bytes()));
+        }
```
Veykril, line 70: No need to do this (especially the path clone), I assume remnant from testing something

Veykril:
Blocking in a notification handler is gonna end in a bad experience as notifications are handled on the main thread which means blocking there blocks processing anything else aside from thread pool tasks. It would probably make more sense to add a new [`Task`](https://github.com/rust-lang/rust-analyzer/blob/e8c4007cfcbb606e167ee90bc4eca2753284f7fc/crates/rust-analyzer/src/main_loop.rs#L64) kind for this and spawn the special handler on the task pool.

crates/rust-analyzer/src/handlers/notification.rs
davidbarsky: I'll need to try removing these, but I'm *decently* confident that I needed to do this to drop the guard corresponding to `state.vfs.write()`.

It's probably going to be a moot point after I switch this over to the task-based approach.

Veykril: The temporary allocated for the write guard in `state.vfs.write().0.set_file_contents(path, Some(params.text_document.text.into_bytes()));` will be dropped immediately after the statement ends.

davidbarsky:
Moved over to a task-based approach. I'm not *fully* sure if this is the approach that you had in mind since `GlobalState::process_changes` isn't running on a different thread, but I guess no other caller of it really is either.

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -486,6 +488,27 @@ impl GlobalState {
...
+                        tracing::debug!(?uri, "rust-analyzer does not track this file");
+                        self.send_notification::<ext::UnindexedProject>(
+                            ext::UnindexedProjectParams {
+                                text_documents: vec![lsp_types::TextDocumentIdentifier { uri }],
+                            },
+                        );
+                    } else {
+                        tracing::warn!("was unable to get analysis for crate")
+                    }
+                }
```
Veykril, line 510: Now we are still processing it on the main thread :D All of this should be done in the spawned task on the thread pool, where as `Task::FileIndexState` should be used to return the result of the computation here (the `snap.analysis.crates_for(id)` result) and then have the task handler here just send the notification.

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -486,6 +488,27 @@ impl GlobalState {
             // Only retry requests that haven't been cancelled. Otherwise we do unnecessary work.
             Task::Retry(req) if !self.is_completed(&req) => self.on_request(req),
             Task::Retry(_) => (),
+            Task::FileIndexState(uri) => {
+                let _p = profile::span("run_unindexed_project");
+
+                tracing::debug!(changes = self.process_changes(), "processes changes");
```
lnicola, line 494: Drive-by remark: does this actually work with `features = ["max_level_info"]` etc.?

crates/rust-analyzer/src/handlers/notification.rs
```diff
@@ -64,7 +66,16 @@ pub(crate) fn handle_did_open_text_document(
         if already_exists {
             tracing::error!("duplicate DidOpenTextDocument: {}", path);
         }
-        state.vfs.write().0.set_file_contents(path, Some(params.text_document.text.into_bytes()));
+
+        let vfs = &mut state.vfs.write().0;
+        vfs.set_file_contents(path, Some(params.text_document.text.into_bytes()));
```
Veykril, line 71: Also undo this please, because this will actually widen the temporary scope of the guard compared to the one liner

crates/rust-analyzer/src/main_loop.rs
davidbarsky: ...you know, i don't think it will. i'll have to move it out.

crates/rust-analyzer/src/handlers/notification.rs
davidbarsky: can do.

crates/rust-analyzer/src/main_loop.rs
davidbarsky: removed it, since it was unnecessary.

crates/rust-analyzer/src/handlers/notification.rs
davidbarsky: undid this change.

crates/rust-analyzer/src/main_loop.rs
davidbarsky: After chatting about this over DMs, I'm _not_ going to take the suggested approach and make this noop task more substantial since there is no way to ensure that this code runs _after_ `GlobalState::process_changes()` has been called without doing one of the following:
- Add a retry mechanism for tasks, which would make this a de-facto spin loop. This might be okay!
- Creating a mechanism of enqueuing a task, but deferring spawning them until _after_ a turn of `GlobalState::handle_event`. This ensures that `GlobalState::process_changes()` has been called.

davidbarsky:
Right, I made two sets of changes that I would be completely happy with backing out on:
- I made a background task queue with a lightweight struct wrapper. Its only purpose is spawn a task _after_ `GlobalState::process_changes`.
- To avoid adding a *regular* `Task` whose only purpose is to send a notification to the client, I opted to make `lsp_server::ReqQueue` implement `Clone` and I wrapped in an `Arc` and call `Arc::make_mut` as appropriate. This was necessary because the spawned task would move `self` across threads.
    - As an alternative, I can inline all two lines of `GlobalState::send_notification` into `GlobalState::handle_queued_task` and undo the changes made to `lsp_server::ReqQueue` and any subsequent knock-on changes. This means that if a `QueuedTask` ever grows an enum variant that needs to send a _request_, they might need to take the approach that I currently took with `ReqQueue`, but that day may never come—after all, the current approach to concurrency in `rust-analyzer` worked well until now.

davidbarsky:
(Squash commits; rebased against master. No meaningful change.)

davidbarsky:
I've decided to back out of changes where I factored the `GlobalState.sender` and `GlobalState.req_queue` fields into a dedicated `Conn` struct alongside all the other, supporting changes (pervasive use of `Arc::make_mut`, cloneable fields in `lsp_server`). I've instead inlined the contents of `GlobalState::send_notification` into the event handler, which feels fine to me.

Anyways, I've had this PR deployed in my employer's [buck2](https://github.com/facebook/buck2)-based monorepo and it has driven down instance of rust-analyzer not running for folks _dramatically_.

editors/code/src/ctx.ts
```diff
@@ -233,6 +233,19 @@ export class Ctx implements RustAnalyzerExtensionApi {
                     this.outputChannel!.show();
                 }),
             );
+            this.pushClientCleanup(
+                this._client.onNotification(ra.unindexedProject, async (params) => {
+                    const command = `${this.config.discoverProjectRunner}.discoverWorkspaceCommand`;
```
Veykril, line 238: Does this work when `discoverProjectRunner` is `undefined` or will executing then throw?

Veykril (commented):
There is a similar thing I added for whether proc-macros have changed in https://github.com/rust-lang/rust-analyzer/pull/16247, we might be able to switch that to this approach as well. Will need to check

editors/code/src/ctx.ts
davidbarsky: You're right, this would absolutely panic/throw. I _assumed_ that if notifications are configured, a `discoverProjectRunner` would also be configured, but I don't think we should. I'll make this check if `discoverProjectRunner` is set, but arguably, one should imply the other.

Do you think it'd be a good idea to error out before initializing the server here? I don't think the client notifications are useful without a mechanism to _discover_ other projects.

davidbarsky:
> There is a similar thing I added for whether proc-macros have changed in https://github.com/rust-lang/rust-analyzer/pull/16247, we might be able to switch that to this approach as well. Will need to check

Having briefly looked over https://github.com/rust-lang/rust-analyzer/pull/16247, it's certainly a maybe. I considered an using the request queue system, but that only really is able to handle a _single_ event at a time. The task-based approach I did here can handle multiple events, which can happen with both `FileDidOpenNotifications` _and_ proc macros being changed.

editors/code/src/ctx.ts
Veykril: Seems fine, as long as we don't have the throwing behavior here

davidbarsky: Whoops, sorry—performance stuff was happening this past week. fixed.

crates/rust-analyzer/src/global_state.rs
```diff
@@ -126,6 +126,10 @@ pub(crate) struct GlobalState {
         OpQueue<(), (Arc<Vec<ProjectWorkspace>>, Vec<anyhow::Result<WorkspaceBuildScripts>>)>,
     pub(crate) fetch_proc_macros_queue: OpQueue<Vec<ProcMacroPaths>, bool>,
     pub(crate) prime_caches_queue: OpQueue,
+
+    /// a deferred task queue. this should only be used if the enqueued Task
+    /// can only run *after* [`GlobalState::process_changes`] has been called.
```
Veykril, line 131: A deferred task queue used for doing database dependent work inside sync handlers. Accessing the database may block and as such should be offloaded from the main thread.

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -625,6 +644,35 @@ impl GlobalState {
...
+                    let id = from_proto::file_id(&snap, &uri).expect("unable to get FileId");
+                    if let Ok(crates) = &snap.analysis.crates_for(id) {
+                        if crates.is_empty() {
+                            let not = lsp_server::Notification::new(
+                                ext::UnindexedProject::METHOD.to_string(),
+                                ext::UnindexedProjectParams {
+                                    text_documents: vec![lsp_types::TextDocumentIdentifier { uri }],
+                                },
+                            );
+                            sender.send(not.into()).expect("unable to send notification");
```
Veykril, line 665: Hm, instead of exposing the sender stuff here, I think I'd prefer sending back a `Task` (via the `_` parameter of the spawn closure, that's a task sender) and add a `Task::Notification` kind to be used for that. (then we can undo the `+ Send` changes and the `pub(crate)`)

davidbarsky: I wrote that initially (prior to publishing this PR), but opted for the current approach because I wasn’t sure if it’d be a premature architectural change. Will switch back!

crates/rust-analyzer/src/global_state.rs
davidbarsky: i’ll incorporate that. at least for this functionality, the fact that this occurs _after_ `GlobalState::process_changes` is load-bearing, so i’ll keep that detail.

(it’s also enforced by the slow test I added, but it doesn’t hurt to restate this detail.)

crates/rust-analyzer/src/main_loop.rs
davidbarsky: fixed!

crates/rust-analyzer/src/global_state.rs
davidbarsky: done!

Veykril (approved):
clippy is complaining on CI

PR #15864, opened 2023-11-10
fix: find `Self` reference

Young-Flash:
took a lot of time to debug to find the problem, here should compare the actual `Adt` type

close https://github.com/rust-lang/rust-analyzer/issues/12693

### DIFF
--- a/crates/ide-db/src/search.rs
+++ b/crates/ide-db/src/search.rs
@@ -584,7 +584,7 @@ impl<'a> FindUsages<'a> {
     ) -> bool {
         match NameRefClass::classify(self.sema, name_ref) {
             Some(NameRefClass::Definition(Definition::SelfType(impl_)))
-                if impl_.self_ty(self.sema.db) == *self_ty =>
+                if impl_.self_ty(self.sema.db).as_adt() == self_ty.as_adt() =>
             {
                 let FileRange { file_id, range } = self.sema.original_range(name_ref.syntax());
                 let reference = FileReference {
--- a/crates/ide/src/references.rs
+++ b/crates/ide/src/references.rs
@@ -684,6 +684,32 @@ enum Foo {
     }
 
     #[test]
+    fn test_self() {
+        check(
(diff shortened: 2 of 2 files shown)


lnicola (approved):

crates/ide-db/src/search.rs
```diff
@@ -584,7 +584,7 @@ impl<'a> FindUsages<'a> {
     ) -> bool {
         match NameRefClass::classify(self.sema, name_ref) {
             Some(NameRefClass::Definition(Definition::SelfType(impl_)))
-                if impl_.self_ty(self.sema.db) == *self_ty =>
+                if impl_.self_ty(self.sema.db).as_adt() == self_ty.as_adt() =>
```
Veykril, line 587: Hmm, this doesn't seem right to me, now we discard any other type that could appear as the self type in an impl. I'm curious as to why this equality fails in the given example in the first place. Sounds like we are losing some type info along the way somewhere?

Young-Flash: I am curious too. debug to that point and found the unequality. Is there anyother type for `Self` in an `impl` beside `Adt`? 

Actually I don't quite understand the whole implementation of find refenence, it‘s complicated for me :(

Veykril:
```rs
trait T {}
impl T for fn() {}
```
is an example. Don't think you can trigger `references` highlighting on that, but other code might require the `Self` resolution in search to work for that. I can take a look at that next week and see what's going on here

Veykril: And yes, the reference search is pretty messy looking and undocumented so I don't blame you for not being able to fully grasp the implementation there :)

Veykril: Ah the problem is that the `self_ty` here contains an `{error}` for the generic param where as the `impl.self_ty()` returns the struct with a var for the generic parameter, so the equality check fails. Hm

PR #15868, opened 2023-11-10
fix: failed to infer OUT_DIR when workspace root contains symlink

### LINKED ISSUE #15867: Cannot infer `OUT_DIR` when workspace root contains symlink
https://github.com/rust-lang/rust-analyzer/blob/11a87c917943dac5a568579f799c2d7458324103/crates/project-model/src/build_scripts.rs#L260-L269

In line 264, because package id from cargo metadata(used by workspace) is different from cargo check(used by build scripts) when root contains symlink. We will "ignore" these package id so that `OUT_DIR` will not be set.

And I will make a PR to fix this.

wasd96040501:
fix #15867


Veykril:
Block on discussion in https://github.com/rust-lang/rust-analyzer/issues/15867#issuecomment-1854723355

crates/rust-analyzer/tests/slow-tests/testdir.rs
```diff
@@ -10,7 +10,7 @@ pub(crate) struct TestDir {
 }
 
 impl TestDir {
-    pub(crate) fn new() -> TestDir {
+    pub(crate) fn new(symlink: bool) -> TestDir {
```
Veykril, line 13: nit, make this constructor private and expose two new ones without the bool param `new` and `new_symlinked` or similar that wrap the private one.

Veykril (approved):
Okay thanks to the explanation given in the issue I can see the point of this change now 👍

Veykril (approved):

Veykril:
Thanks!
@bors r+

wasd96040501:
It seems that [unit test](https://github.com/rust-lang/rust-analyzer/actions/runs/7490843636/job/20390652421) has failed on windows.
panic 1
```
thread 'root_contains_symlink_out_dirs_check' panicked at crates\rust-analyzer\tests\slow-tests\support.rs:109:56:
called `Result::unwrap()` on an `Err` value: Os { code: 183, kind: AlreadyExists, message: "Cannot create a file when that file already exists." }
```
panic 2
```
thread 'root_contains_symlink_out_dirs_check' panicked at crates\rust-analyzer\tests\slow-tests\testdir.rs:82:13:
failed to remove temporary directory C:\Users\RUNNER~1\AppData\Local\Temp\testdir\612_4\_symlink: The directory name is invalid. (os error 267)
```

I will attempt to fix it.

wasd96040501:
@Veykril 
CI fixed.
``` rust
#[cfg(target_os = "windows")]
std::os::windows::fs::symlink_file(path, &symlink_path).unwrap();
```
replaced to
``` rust
#[cfg(target_os = "windows")]
std::os::windows::fs::symlink_dir(path, &symlink_path).unwrap();
```

Issue #15856, opened 2023-11-09
Misleading command name `> rust-analyzer: Expand macro recursively`


loynoir:
### actual

As far as I found,

- The command name is misleading `> rust-analyzer: Expand macro recursively`

I understand it as expand full file.

### expected

- command name `> rust-analyzer: Expand macro recursively at cursor`

lnicola:
Expanding the full file would be called `Expand macros recursively`.

Veykril:
This command doesnt expand all macros in the current file, but only the one under the cursor (recursively), so this is labeled correctly

loynoir:
I know.

I suggest, add two words `at cursor` to command display name. 

Zero ambiguity.

Veykril:
Fair, I agree we can change it to `Expand macro at cursor recursively` that seems like an improvement.

PR #15870, opened 2023-11-11
minor: Make "Expand macro" command title more explicit

lnicola:
Closes [#15856](https://github.com/rust-lang/rust-analyzer/issues/15856).

I opted for "caret", since it's the better term (cursor is the mouse), but I'm not sure how popular it is these days.

### DIFF
--- a/crates/ide/src/expand_macro.rs
+++ b/crates/ide/src/expand_macro.rs
@@ -14,12 +14,12 @@ pub struct ExpandedMacro {
 
 // Feature: Expand Macro Recursively
 //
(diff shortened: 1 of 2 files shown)


crates/load-cargo/src/lib.rs
```diff
@@ -208,6 +208,7 @@ impl ProjectFolders {
             let entry = {
                 let mut dirs = vfs::loader::Directories::default();
                 dirs.extensions.push("rs".into());
+                dirs.extensions.push("toml".into());
```
edwin0cheng, line 211: ?

PR #15260, opened 2023-07-10
internal: Migrate assists to the structured snippet API, part 3

DropDemBits:
Continuing from #15231

Migrates the following assists:
- `add_missing_match_arms`
- `fix_visibility`
- `promote_local_to_const`

The `add_missing_match_arms` changes are best reviewed commit-by-commit since they're relatively big changes compared to the rest of the commits.

Issue #15710, opened 2023-10-04
Get rid of mutable syntax node usage


Veykril:
Somewhat meant as a means of discussion:

After having used and reviewed a bunch of PRs using the mutable syntax node API I am personally not too happy about it. It can be awkward to use at times and has some pitfalls that one has to be aware of. A major problem with that API is that it introduces iterator invalidation (you can iterate while mutating the tree), giving way to easily end up panicking if one does not delay the edits after iteration via collecting into a vec or similar. Likewise it makes it harder to replace our version of rowan to one that uses [proper slots](https://github.com/rust-lang/rust-analyzer/issues/6584) which is something I am in favor of, as the slot API is unlikely to have the same feature (unless we implement it back into rome's fork of rowan which would be the current slot based rowan solution).

The current benefit of it is that we have the `ted` api that allows automagically fixing up whitespace on edits, but that's really not necessary in the end game when we have [our own syntax formatter ](https://github.com/rust-lang/rust-analyzer/issues/1665) which would allow us to format edited nodes on the fly.

The main question then arises, how do we do edits if not via mutable syntax nodes? The current approach was using the handwritten `make` module to stitch nodes together (which is still somewhat used). But that approach is really annoying to use, inefficient and looks horrid as well. One nice looking approach would be having a quasi quote api as described in https://github.com/rust-lang/rust-analyzer/issues/2227, that could also do some static checks ideally (like checking well typedness of the tree at comp time maybe). And alternative to mutation is https://github.com/rust-lang/rust-analyzer/issues/9649 though I haven't looked too much into it.

In general, assists are all over the place currently, some do text based editing, some do make api + stringified text edits and finally we have some that use the mutable node api. Obviously not ideal if we mix and match however we like to.

Veykril:
cc @matklad feels like an issue you'd have an opinion about or two :)

matklad:
Yeah, I wouldn’t say that mutable syntax API is a success :-)

That being said, I think it barks at the right tree: the fundamental capability there is “tracking nodes across edits”, and I think it would be important to have that in one way or another. But I am not too confident here: maybe we don’t need cursor API at all?

One curious solution here is Roslyn’s SyntaxAnnotations:

https://joshvarty.com/2015/09/18/learn-roslyn-now-part-13-keeping-track-of-syntax-nodes-with-syntax-annotations/

my understanding is that this mechanism powers all “mutable” editing APIs, which allow you to incrementally mutate the tree.

[...]

PR #15874, opened 2023-11-12
internal: Migrate assists to the structured snippet API, part 4

DropDemBits:
Continuing from #15260

- `add_turbo_fish`
- `add_type_ascription`
- `destructure_tuple_binding`
- `destructure_tuple_binding_in_subpattern`

I did this a while ago, but forgot to make a PR for the changes until now. 😅

### DIFF
--- a/crates/ide-assists/src/handlers/destructure_tuple_binding.rs
+++ b/crates/ide-assists/src/handlers/destructure_tuple_binding.rs
@@ -197,15 +197,7 @@ impl AssignmentEdit {
     fn apply(self) {
         // with sub_pattern: keep original tuple and add subpattern: `tup @ (_0, _1)`
         if self.in_sub_pattern {
-            ted::insert_all_raw(
-                ted::Position::after(self.ident_pat.syntax()),
-                vec![
-                    make::tokens::single_space().into(),
-                    make::token(T![@]).into(),
-                    make::tokens::single_space().into(),
-                    self.tuple_pat.syntax().clone().into(),
-                ],
-            )
+            self.ident_pat.set_pat(Some(self.tuple_pat.into()))
         } else {
             ted::replace(self.ident_pat.syntax(), self.tuple_pat.syntax())
         }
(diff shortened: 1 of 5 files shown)


crates/ide-assists/src/handlers/destructure_tuple_binding.rs
```diff
@@ -142,98 +152,123 @@ enum RefType {
...
+        if self.in_sub_pattern {
+            ted::insert_all_raw(
+                ted::Position::after(self.ident_pat.syntax()),
+                vec![
+                    make::tokens::single_space().into(),
+                    make::token(T![@]).into(),
+                    make::tokens::single_space().into(),
+                    self.tuple_pat.syntax().clone().into(),
+                ],
+            )
```
DropDemBits, line 208: This doesn't feel great, as it makes the syntax tree not match with the parsed version (which may be a problem when automatically fixing the formatting of the syntax tree). I'll see if I can make this be a `ted::replace` instead.

Veykril:
Right, we were actively moving assists to use the mutable syntax node API which is at odds with my opinion of them nowadays https://github.com/rust-lang/rust-analyzer/issues/15710 😓 Wonder if we can figure out a solution for this API should we ever move away from them.
But that aside, moving this is fine for now
@bors r+

DropDemBits:
@Veykril

> Right, we were actively moving assists to use the mutable syntax node API which is at odds with my opinion of them nowadays #15710 😓

Haven't seen that issue until now (slowly getting back into contributing in general after a break), but hard agree on on the problems with the mutable syntax tree api. It is a step up from the text-based apis, but still not ideal 😔

Once I'm done with the structured snippet port + a few structured snippet related odds & ends (according to my personal todo list, maybe mid-December?), I do have an interest in tackling things like the syntax quasi-quote api and at the very least an auto-indenter (though we'll see how I feel by then).

> Wonder if we can figure out a solution for this API should we ever move away from them.

The neat thing about how structured snippets are implemented is that `SourceChangeBuilder` only cares about where nodes & tokens are, and then once the edit is finished, the snippet locations are converted into the corresponding `TextRange` & `TextSize` for the protocol conversion layer to handle, i.e. it does not depend on the mutable syntax tree api 😁

In fact, I think it'd be even better with a different node tracking api, as I've run into the issue of "losing" syntax nodes quite a few times (e.g. for `add_derive`, and in the future of supporting placeholder groups for `extract_variable`).

Veykril:
A related issue regarding node tracking https://github.com/rust-lang/rust-analyzer/issues/9649

PR #15875, opened 2023-11-12
fix `PathSegment` grammar

Young-Flash:
close https://github.com/rust-lang/rust-analyzer/issues/15778

### DIFF
--- a/crates/syntax/src/ast/node_ext.rs
+++ b/crates/syntax/src/ast/node_ext.rs
@@ -361,15 +361,12 @@ impl ast::Impl {
     }
 }
 
-// for `PathSegment` '<i32 as core::ops::Add>', call `first_path_type` will get `i32` and `last_path_type` will get `core::ops::Add`
-// for '<&i32 as core::ops::Add>', call `first_path_type` and `last_path_type` will both get `core::ops::Add` cause `&i32` is `Type(RefType)`
+// [#15778](https://github.com/rust-lang/rust-analyzer/issues/15778)
 impl ast::PathSegment {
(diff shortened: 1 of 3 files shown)


crates/syntax/src/ast/node_ext.rs
```diff
@@ -361,6 +361,18 @@ impl ast::Impl {
...
+impl ast::PathSegment {
+    pub fn first_path_type(&self) -> Option<ast::PathType> {
+        self.syntax().children().find_map(ast::PathType::cast)
+    }
+
+    pub fn last_path_type(&self) -> Option<ast::PathType> {
+        self.syntax().children().filter_map(ast::PathType::cast).last()
+    }
+}
+
```
Veykril, line 375: This should have the following api instead 
```rs
impl ast::PathSegment {
    pub fn qualifying_trait(&self) -> Option<ast::PathType> {
    	...
    }
}
```
we just want a function that gives us the relevant path type there (though we have no consumers of that API yet). Regarding the implementation, we basically need to peek, if the Type in `<Type (as PathType>`  is anything but a `PathType`, then the first `PathType` in the children we find will yield the correct path type. If `Type` is a `PathType` we need to make sure to pick the second occurrence, so we peek if there is a second one, if there is always pick that otherwise pick the first one.

<details>
  <summary>An implementation of what I wrote above, feel free to try implementing it yourself first</summary>
  
  ```rs
  let mut path_types = support::children(self.syntax());
  let first = path_types.next()?;
  path_types.next().or(Some(first))
  ```
  
</details>

Veykril (approved):

PR #15269, opened 2023-07-12
internal: Defer structured snippet rendering to allow escaping snippet bits

DropDemBits:
Since we know exactly where snippets are, we can transparently escape snippet bits to the exact text edits that need it, and not have to do it for anything other text edits.

Also will eventually fix #11006 once all assists are migrated. This comes as a side-effect of text edits that don't have snippets get marked as having no insert formatting at all.

PR #15876, opened 2023-11-12
minor: Allow multiple snippet edits in a `TextDocumentEdit`

DropDemBits:
Explicitly[^1] allow a single `TextDocumentEdit` to have multiple `SnippetTextEdit`s. This allows things like renaming extracted variables and functions without having to go through a separate rename step. For an example of what this looks like, see the video in [this comment](https://github.com/microsoft/vscode/issues/145374#issuecomment-1177341711).

The behavior described here lines up with [what VSCode does](https://github.com/microsoft/vscode/blob/bdc113ffe148a92d0e1a8ec34b12c44ea0b73f29/src/vscode-dts/vscode.d.ts#L3728-L3731), and presumably what the eventual LSP behavior will be.

[^1]: This was technically the case before #15269, a single `TextDocumentEdit` always had multiple edits which were `InsertTextFormat.Snippet` as all of the edits were marked as being snippets, even if there weren't any tab stops or placeholders.

### DIFF
--- a/docs/dev/lsp-extensions.md
+++ b/docs/dev/lsp-extensions.md
@@ -57,8 +57,9 @@ export interface TextDocumentEdit {
 }
 ```
 
-When applying such code action or text edit, the editor should insert snippet, with tab stops and placeholder.
-At the moment, rust-analyzer guarantees that only a single edit will have `InsertTextFormat.Snippet`.
+When applying such code action or text edit, the editor should insert snippet, with tab stops and placeholders.
+At the moment, rust-analyzer guarantees that only a single `TextDocumentEdit` will have edits which can be `InsertTextFormat.Snippet`.
+Any additional `TextDocumentEdit`s will only have edits which are `InsertTextFormat.PlainText`.
 
 ### Example
 


DropDemBits:
Wasn't sure on which changelog prefix to use since it's only a documentation change (would minor work for this?).

nemethf:
@DropDemBits, do I understand correctly that currently rust-analyzer does not have a feature that creates a single TextDocumentEdit with multiple SnippetTextEdits?  Once it has, can you, please, update [the example section](https://github.com/rust-lang/rust-analyzer/blob/1e4686865efbe5c3a305708d3aceacc181081be8/docs/dev/lsp-extensions.md#example)?  It would help me to implement this in a LSP client.

Thank you.

DropDemBits:
@nemethf That is correct. It's something that I'm planning to do too, and I can also ping when I make the PR.

nemethf:
>  It's something that I'm planning to do too, and I can also ping when I make the PR.

@DropDemBits, have you made any progress on this?  Thanks.

DropDemBits:
@nemethf Sorry, just started working on rust-analyzer again recently after a bit of a break. 

I still have this tracked in my backlog, though with college running up again it's hard to find free time to work on stuff again. 😅

PR #15879, opened 2023-11-12
Implement completion for the callable fields.

### LINKED ISSUE #14656: Complete callable fields when used in a method call position
```rs
struct S {
    field: fn()
}

S { field: || {} }.fi$0()
```
Here we should complete the field (as if it was a method call) and automatically wrap the receiver in a pair of parens to fix up the call:
```rs
struct S {
    field: fn()
}

(S { field: || {} }.field)($0)
```

dfireBird:
Fixes #14656 

PR is opened with basic changes. It could be improved by having a new `SymbolKind` for the callable fields and implementing a separate render function similar to the `render_method` for the new `SymbolKind`.
It could also be done without any changes to the `SymbolKind` of course, have the new function called based on the type of field. 
I prefer the former method.

Please give any thoughts or changes you think is appropriate for this method. I could start working on that in this same PR.

### DIFF
--- a/crates/ide-completion/src/completions/dot.rs
+++ b/crates/ide-completion/src/completions/dot.rs
@@ -37,9 +37,6 @@ pub(crate) fn complete_dot(
         is_field_access,
     );
 
-    if let DotAccessKind::Method { .. } = dot_access.kind {
-        cov_mark::hit!(test_no_struct_field_completion_for_method_call);
-    }
     complete_methods(ctx, receiver_ty, |func| acc.add_method(ctx, dot_access, func, None, None));
 }
 
@@ -259,7 +256,6 @@ impl A {
 
     #[test]
     fn test_no_struct_field_completion_for_method_call() {
-        cov_mark::check!(test_no_struct_field_completion_for_method_call);
         check(
             r#"
 struct A { the_field: u32 }
@@ -1202,7 +1198,7 @@ fn foo() { S { va_field: 0, fn_field: || {} }.fi$0() }
 "#,
             r#"
 struct S { va_field: u32, fn_field: fn() }
-fn foo() { (S { va_field: 0, fn_field: || {} }).fn_field() }
+fn foo() { (S { va_field: 0, fn_field: || {} }.fn_field)() }
 "#,
         );
     }
@@ -1235,7 +1231,7 @@ fn foo() {
 struct B(u32, fn())
 fn foo() {
    let b = B(0, || {});
-   (b).1()
+   (b.1)()
 }
 "#,
         )
--- a/crates/ide-completion/src/completions/dot.rs
+++ b/crates/ide-completion/src/completions/dot.rs
@@ -1206,4 +1206,17 @@ impl<B: Bar, F: core::ops::Deref<Target = B>> Foo<F> {
         "#]],
         );
     }
+
+    #[test]
+    fn test_struct_function_field_completion() {
+        check(
+            r#"
+struct S { field: fn() }
+fn foo() { S { field: || {} }.fi$0() }
+"#,
+            expect![[r#"
+                fd field fn()
+            "#]],
+        );
+    }
 }
--- a/crates/ide-completion/src/completions/dot.rs
+++ b/crates/ide-completion/src/completions/dot.rs
@@ -1219,4 +1219,20 @@ fn foo() { S { field: || {} }.fi$0() }
             "#]],
         );
     }
+
+    #[test]
+    fn test_tuple_function_field_completion() {
+        check(
+            r#"
+struct B(u32, fn())
+fn foo() {
+   let b = B(0, || {});
+   b.$0()
+}
+"#,
+            expect![[r#"
+                fd 1 fn()
+            "#]],
+        );
+    }
 }
--- a/crates/ide-completion/src/completions/dot.rs
+++ b/crates/ide-completion/src/completions/dot.rs
@@ -1186,12 +1186,24 @@ impl<B: Bar, F: core::ops::Deref<Target = B>> Foo<F> {
     fn test_struct_function_field_completion() {
         check(
             r#"
-struct S { field: fn() }
-fn foo() { S { field: || {} }.fi$0() }
+struct S { va_field: u32, fn_field: fn() }
+fn foo() { S { va_field: 0, fn_field: || {} }.fi$0() }
 "#,
             expect![[r#"
-                fd field fn()
+                fd fn_field fn()
             "#]],
+        );
+
+        check_edit(
+            "fn_field",
+            r#"
+struct S { va_field: u32, fn_field: fn() }
+fn foo() { S { va_field: 0, fn_field: || {} }.fi$0() }
+"#,
+            r#"
+struct S { va_field: u32, fn_field: fn() }
+fn foo() { (S { va_field: 0, fn_field: || {} }).fn_field() }
+"#,
         );
     }
 
@@ -1209,5 +1221,23 @@ fn foo() {
                 fd 1 fn()
             "#]],
         );
+
+        check_edit(
+            "1",
+            r#"
+struct B(u32, fn())
+fn foo() {
+   let b = B(0, || {});
+   b.$0()
+}
+"#,
+            r#"
+struct B(u32, fn())
+fn foo() {
+   let b = B(0, || {});
+   (b).1()
+}
+"#,
+        )
     }
 }
--- a/crates/ide-completion/src/completions/dot.rs
+++ b/crates/ide-completion/src/completions/dot.rs
@@ -28,6 +28,13 @@ pub(crate) fn complete_dot(
 
     if let DotAccessKind::Method { .. } = dot_access.kind {
         cov_mark::hit!(test_no_struct_field_completion_for_method_call);
+        complete_fn_fields(
+            acc,
+            ctx,
+            receiver_ty,
+            |acc, field, ty| acc.add_field(ctx, dot_access, None, field, &ty),
+            |acc, field, ty| acc.add_tuple_field(ctx, None, field, &ty),
+        );
     } else {
         complete_fields(
             acc,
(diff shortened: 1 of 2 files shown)


crates/ide-completion/src/render.rs
```diff
@@ -147,7 +148,40 @@ pub(crate) fn render_field(
...
+        let mut builder = TextEdit::builder();
+        // Using TextEdit, insert '(' before the struct name and ')' before the
+        // dot access, then comes the field name and optionally insert function
+        // call parens.
+
+        if let Some(receiver) = &dot_access.receiver {
+            let range = receiver.syntax().text_range();
+            builder.insert(range.start(), "(".to_string());
+            builder.insert(range.end(), ")".to_string());
+        }
```
DropDemBits, line 161: So that it works in macros too, you could use `ctx.completion.sema.original_ast_node()` like so:

```suggestion
        if let Some(receiver) = &dot_access.receiver {
            if let Some(receiver) = ctx.completion.sema.original_ast_node(receiver.clone()) {
                let range = receiver.syntax().text_range();
                builder.insert(range.start(), "(".to_string());
                builder.insert(range.end(), ")".to_string());
            }
        }
```

crates/ide-completion/src/render.rs
```diff
@@ -147,7 +148,40 @@ pub(crate) fn render_field(
...
+        builder.replace(
+            ctx.source_range(),
+            field_with_receiver(db, receiver.as_ref(), &escaped_name).into(),
+        );
+
+        let is_fn_expected =
+            ctx.completion.expected_type.as_ref().map_or(false, |ty| ty.is_fn() || ty.is_closure());
```
DropDemBits, line 168:
```suggestion
            ctx.completion.expected_type.as_ref().is_some_and(|ty| ty.is_fn() || ty.is_closure());
```

crates/ide-completion/src/render.rs
```diff
@@ -147,7 +148,40 @@ pub(crate) fn render_field(
...
+
+        // This could be refactored as method of DotAccessKind
+        let is_parens_needed = if let DotAccessKind::Method { has_parens } = dot_access.kind {
+            !has_parens
+        } else {
+            true
+        };
```
DropDemBits, line 175: I think it can be simplified to this `matches!` expression:

```suggestion
        let is_parens_needed = !matches!(dot_access.kind, DotAccessKind::Method { has_parens: true });
```

crates/ide-completion/src/render.rs
```diff
@@ -147,7 +148,40 @@ pub(crate) fn render_field(
...
+
+        if !is_fn_expected && is_parens_needed {
```
DropDemBits, line 177: Was confused on why the `!is_fn_expected` condition was needed, but realized that for the following code, we don't add parentheses since it's expecting a callable type:
```rs
struct S { field: fn() }

fn foo() {
    let a: fn() = S { field: || () }.fi$0;
}
```

`is_fn_expected` could probably be renamed to something like `expects_fn_type`, and if this is not already covered by an existing test, it'd be great to add a test for it too.

crates/ide-completion/src/completions/dot.rs
```diff
@@ -144,6 +151,33 @@ fn complete_methods(
...
+    mut tuple_index: impl FnMut(&mut Completions, usize, hir::Type),
+) {
+    let mut seen_names = FxHashSet::default();
+    for receiver in receiver.autoderef(ctx.db) {
+        for (field, ty) in receiver.fields(ctx.db) {
+            if seen_names.insert(field.name(ctx.db)) && (ty.is_fn() || ty.is_closure()) {
+                named_field(acc, field, ty);
+            }
+        }
+        for (i, ty) in receiver.tuple_fields(ctx.db).into_iter().enumerate() {
```
DropDemBits, line 168: Should probably also add tests for callable tuple fields.

DropDemBits (commented):
Some nits and notes, but overall LGTM.

crates/ide-completion/src/render.rs
dfireBird: I was actually looking for this macro but I couldn't find it, so I did what I did. Thanks for helping out.

DropDemBits: No problems!

dfireBird:
I've now added the tests for both the cases you mentioned and made the refactors as well.

crates/ide-completion/src/render.rs
```diff
@@ -1606,30 +1638,110 @@ fn foo(f: Foo) { let _: &u32 = f.b$0 }
...
+                        text_edit: TextEdit {
+                            indels: [
+                                Indel {
+                                    insert: "(",
+                                    delete: 57..57,
+                                },
+                                Indel {
+                                    insert: ")",
+                                    delete: 75..75,
+                                },
```
DropDemBits, line 1714: Ah whoops, I didn't notice that the parentheses wrapping code happens regardless of if it's expecting a function type or not. 😅
Should only be for when we're actually calling the function.

DropDemBits (approved):
Just one last nit that I missed in the first review, and otherwise LGTM.

crates/ide-completion/src/render.rs
dfireBird: Ah right. My bad I missed it too.
Can we consider it calls function when not expecting a function type or only when the `DotAcessKind` is `Method`?

DropDemBits: I think when it's not expecting a function type is when we should consider it calling a function.

Sorry for the delay, was tired for the past two days. 😅

dfireBird: Don't be sorry, you're not obligated to reply to me.

Anyways, thanks that's what I thought too implement as well.

Veykril (commented):
There is one interesting case that we should also have a test for which is `foo.bar::<>`, I think this should result in a `DotAccessKind::Method { has_parens: false }` (the only case where its false I believe as otherwise missing parens will result in a field access node).

crates/ide-completion/src/completions/dot.rs
```diff
@@ -1095,4 +1129,33 @@ fn test(s: S<Unknown>) {
             "#]],
         );
     }
+
+    #[test]
+    fn test_struct_function_field_completion() {
+        check(
```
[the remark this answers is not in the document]
dfireBird: I knew that test wasn't sufficient, but I just used the `check` function other tests were using. Thanks for guiding me.

crates/ide-completion/src/completions/dot.rs
```diff
@@ -144,6 +151,33 @@ fn complete_methods(
     );
 }
 
+fn complete_fn_fields(
```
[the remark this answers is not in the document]
dfireBird: I'm not sure why I created a separate function, probably to not break any API that relies on it. But it's quite bad of me to create a new function for that case.

dfireBird:
For the `foo.bar::<>` case, it should complete to `(foo).bar::<>()` right? 
In current implementation it completes to `(foo).baz()::<>`.

Veykril:
> For the `foo.bar::<>` case, it should complete to `(foo).bar::<>()` right? In current implementation it completes to `(foo).bar()::<>`.

I should complete to `foo.bar::<>()`, no reason to parenthesize the receiver

dfireBird:
Sorry I was busy with some thing and couldn't find time to fix the tests. I have fixed the tests.

I'll create a follow up PR for the `foo.baz::<>` case, once this PR is merged.

crates/ide-completion/src/completions/dot.rs
```diff
@@ -26,16 +26,19 @@ pub(crate) fn complete_dot(
...
+        acc,
+        ctx,
+        receiver_ty,
+        |acc, field, ty| acc.add_field(ctx, dot_access, None, field, &ty),
+        |acc, field, ty| acc.add_tuple_field(ctx, None, field, &ty),
+        is_field_access,
+    );
+
     if let DotAccessKind::Method { .. } = dot_access.kind {
         cov_mark::hit!(test_no_struct_field_completion_for_method_call);
```
Veykril, line 41: This cov mark makes little sense to be tested like this now (since the `else` branch is gone, which was what this mark wanted to assert is not being called)

crates/ide-completion/src/completions/dot.rs
```diff
@@ -1172,4 +1181,63 @@ impl<B: Bar, F: core::ops::Deref<Target = B>> Foo<F> {
...
+
+        check_edit(
+            "fn_field",
+            r#"
+struct S { va_field: u32, fn_field: fn() }
+fn foo() { S { va_field: 0, fn_field: || {} }.fi$0() }
+"#,
+            r#"
+struct S { va_field: u32, fn_field: fn() }
+fn foo() { (S { va_field: 0, fn_field: || {} }).fn_field() }
```
Veykril, line 1205: This is why I wanted the `check_edit` test :^) The output here is wrong, this should be
```suggestion
fn foo() { (S { va_field: 0, fn_field: || {} }.fn_field)() }
```
This is how you call fields in rust (to disambiguate from method calls)

crates/ide-completion/src/render.rs
```diff
@@ -147,7 +148,43 @@ pub(crate) fn render_field(
...
+
+        let expected_fn_type =
+            ctx.completion.expected_type.as_ref().is_some_and(|ty| ty.is_fn() || ty.is_closure());
+
+        if !expected_fn_type {
+            if let Some(receiver) = &dot_access.receiver {
+                if let Some(receiver) = ctx.completion.sema.original_ast_node(receiver.clone()) {
+                    let range = receiver.syntax().text_range();
+                    builder.insert(range.start(), "(".to_string());
+                    builder.insert(range.end(), ")".to_string());
```
Veykril, line 172: To fix the wrong edit we should probably just check the parent of the (original ast node) receiver here, if its a field add the closing parenthesis after the field etc.

Veykril (commented):
Sorry for all the review rounds and thanks for sticking with it :)

crates/ide-completion/src/completions/dot.rs
dfireBird: > Sorry for all the review rounds and thanks for sticking with it, this should be the last :)
No worries, this is just lesson for me to put good a PR next time.

I keep getting confused where to put the end parenthesis. I somehow finally decided this was correct thing, my bad.

crates/ide-completion/src/render.rs
dfireBird: Yeah, I should do something like this, I suppose storing the range is kinda useless now.
```rust
builder.insert(receiver.syntax().text_range().start(), "(".to_string());
builder.insert(ctx.source_range().end(), ")".to_string());
```

Veykril: Ye that should work I think

Veykril (approved):

Veykril:
Thanks!
@bors r+

lnicola:
image

PR #15885, opened 2023-11-13
Fix typo in highlight_related.rs

RunDevelopment:
While reading the user manual, I noticed a random tick, so I went ahead and fixed it. The `await` keyword should be properly rendered as inline code.

image


lnicola (approved):

lnicola:
Thanks!

@bors r+

PR #15891, opened 2023-11-14
feat: Diagnose some orphan trait impl cases

Veykril:


lnicola:
image

I think the diagnostic range is a bit too large here?

Veykril:
it covers the entire item ye, we should limit it to the signature only probably

PR #15896, opened 2023-11-14
Show placeholder while run command gets runnables from server

minestarks:
This PR fixes a UI annoyance in the VS Code extension when working in large codebases where rust-analyzer can take a few moments to interact with the server. Scenario:

1. Invoke "rust-analyzer: Run" from the command palette or hotkey
2. Quickly start typing to filter the list (or press Enter to accept the last runnable)

We often do this quickly from muscle memory without waiting to see the picker. The picker often takes several seconds to come up, causing us to type garbage into the currently open editor. 

Fix:

Show a placeholder item before we call out to the server. 

image

Selecting this item does nothing so if the user accidentally hits Enter nothing happens.

The list is populated and the placeholder dismissed when the actual runnables are retrieved. From here the behavior is the same as before.

image

### DIFF
--- a/editors/code/src/run.ts
+++ b/editors/code/src/run.ts
@@ -7,6 +7,8 @@ import type { CtxInit } from "./ctx";
 import { makeDebugConfig } from "./debug";
 import type { Config, RunnableEnvCfg, RunnableEnvCfgItem } from "./config";
 import { unwrapUndefinable } from "./undefinable";
+import type { LanguageClient } from "vscode-languageclient/node";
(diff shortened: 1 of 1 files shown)


minestarks:
Maintainers @Veykril @lnicola is there anything I still need to do to get eyes on this PR? This is my first contribution, I believe I've done due diligence and thank this will be a benefit to users. Thank you for your attention! 🙂

Veykril (approved):

minestarks:
Thank you for taking this 🧡

PR #15898, opened 2023-11-15
Update manual.adoc to add language server client - multilspy

LakshyAAAgrawal:


Veykril:
This does not seem to be a text editor but a library, so I don't really see a reason for us to list this.

PR #15899, opened 2023-11-15
internal: Sync from rust

lnicola:


lnicola:
Hold on, I'll try to redo this to get rid of most of the merge commits. For some reason, I end up with hundreds of conflicts when I do it in one go, but only a couple of them if I do it in smaller steps.

Veykril:
Is that because we merged things here while the other sync was going on? I don't remember when we had to pause changes and when not

lnicola:
Yes, but [yesterday's](https://github.com/rust-lang/rust/pull/117701) sync was based off the previous release (Nov 6) and it was such a pain I didn't know it would work in the end, so didn't start again from the one on Nov 13.

For the record, it looks like the best way to merge these back is to do it first up to the release (Nov 6), then merge everything un master. That avoids most conflicts.

crates/rust-analyzer/src/bin/main.rs
```diff
@@ -3,6 +3,10 @@
 //! Based on cli flags, either spawns an LSP server, or runs a batch analysis
 
 #![warn(rust_2018_idioms, unused_lifetimes, semicolon_in_expressions_from_macros)]
+#![cfg_attr(feature = "in-rust-tree", feature(rustc_private))]
+#[cfg(feature = "in-rust-tree")]
```
lnicola, line 7: This either expands to:

```rust
#[allow(unused_extern_crates)]
extern crate rustc_driver;
```

or to nothing, right? At least it still works without the feature enabled.

I _should_ know, but..

crates/ide/src/lib.rs
```diff
@@ -8,8 +8,9 @@
 //! in this crate.
 
 // For proving that RootDatabase is RefUnwindSafe.
-#![recursion_limit = "128"]
 #![warn(rust_2018_idioms, unused_lifetimes, semicolon_in_expressions_from_macros)]
+#![cfg_attr(feature = "in-rust-tree", feature(rustc_private))]
+#![recursion_limit = "128"]
```
lnicola, line 13: Guess we can drop the recursion limit here, 128 is the default.

crates/rust-analyzer/src/bin/main.rs
Veykril: yes, the `cfg` configures the whole annotated item, including its attributes

Veykril:
> Yes, but [yesterday's](https://github.com/rust-lang/rust/pull/117701) sync was based off the previous release (Nov 6) and it was such a pain I didn't know it would work in the end, so didn't start again from the one on Nov 13.

I've been following your struggles there 😅 I feel sorry for anyone working on rustc bootstrap honestly. Thank you so much for dealing with this part of r-a btw 🙇

lnicola:
Then I can

@bors r=Veykril

oxalica:
Is there a reason why to `extern crate rustc_driver` although it's unused? I guess there's some context outside this PR?

lnicola:
It fails to link without it.

PR #15901, opened 2023-11-15
fix: Diagnose everything in nested items, not just def diagnostics

Veykril:
Turns out we only calculated def diagnostics for these before (was wondering why I wasn't getting any type mismatches)


lnicola:
@bors r-

This just failed with some unused import warnings.

PR #15903, opened 2023-11-15
Fix builtin line! expansion

Veykril:
Annoyingly rustc is cheating here since they can just create syntax nodes whereas we have to reparse, so our suffix handling is kind of wrong, this should be a `0` with an implied suffix of u32 but we don't have such a thing.


Veykril:
And it fails, ugh. This is gonna be annoying

PR #15904, opened 2023-11-15
editor/code: bump minimum version of VS Code to 1.78

davidbarsky:
Undoes https://github.com/rust-lang/rust-analyzer/pull/15333.

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -27,7 +27,7 @@
         }
     },
     "engines": {
-        "vscode": "^1.75.0"
+        "vscode": "^1.78.0"
     },
     "enabledApiProposals": [],
     "scripts": {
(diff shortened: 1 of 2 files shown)


editors/code/package.json
```diff
@@ -2178,4 +2178,4 @@
             }
         ]
     }
-}
+}
```
lnicola, line 2181: Nit: can you leave this untouched?

lnicola:
I think  you need to do `npm update @types/vscode`, otherwise you end up updating everything.

editors/code/package.json
davidbarsky: whoops, sorry about that—i think my editor made this change.

davidbarsky:
> I think you need to do `npm update @types/vscode`, otherwise you end up updating everything, including `tsc` to a version that's more strict when parsing its config file.

Cool, yeah—I think this PR was so large because I was futzing around with `npm` before I realized it was stuck on some DNS stuff. Sorry for the churn.

lnicola:
Might be easier to revert the changes to `package-lock.json` and start over.

davidbarsky:
> Might be easier to revert the changes to `package-lock.json` and start over.

yeah, already on that. are y'all okay with force pushes?

PR #15906, opened 2023-11-15
feat: let-else postfix

saiintbrisson:
Creates a new `lete` postfix standing for the recent `let-else` addition.

The postfix follows the rules for `ifl/match` and completes with Ok/Some
if the expression is a try enum.

2023-11-15 19-46-40 (online-video-cutter com)


crates/ide-completion/src/completions/postfix.rs
```diff
@@ -186,6 +200,13 @@ pub(crate) fn complete_postfix(
                 &format!("match {receiver_text} {{\n    ${{1:_}} => {{$0}},\n}}"),
             )
             .add_to(acc, ctx.db);
+
+            postfix_snippet(
+                "lete",
+                "let pat = expr else {}",
+                &format!("let $1 = {receiver_text} else {{\n    $0\n}};"),
+            )
+            .add_to(acc, ctx.db);
```
Young-Flash, line 209: I think it's a irrefutable `let...else` pattern, like `let foo = 0 else { return; }`

saiintbrisson: One is able to match against any pattern they wish, like match, 
`let Enum::Variant(val)` = ... else ...`

Alainx277:
I have a pull request implementing this feature that has been stuck 😕 
https://github.com/rust-lang/rust-analyzer/pull/15730

Veykril:
Apologies for that, I've been on vacation for 3 weeks and only came back a week ago so I have to slowly catch up with the PRs again.

saiintbrisson:
@Alainx277 thank you for the heads-up. I don't know how I missed it!

Issue #15741, opened 2023-10-10
Meta feature: Language server support for `Cargo.toml` file


tgross35:
This issue is an attempt to put all information regarding potential `Cargo.toml` support in one place.

## Blockers

There is no TOML support in RA, and needs a format-preserving parser. [`toml-edit`](https://github.com/toml-rs/toml) is maintained along with `toml` and probably the best choice.

## Potential features

- [ ] Autocomplete or schema help (reference: https://doc.rust-lang.org/cargo/reference/manifest.html)
  - [ ] Autocomplete keys
  - [ ] Suggest valid values for fixed keys like `edition`, `rust-version`, `license`
  - [ ] Suggest in-workspace files for `license-file`, `readme`, `build`
  - [ ] Autocomplete dependencies
  - [ ] Autocomplete dependency features
 - [ ] Schema validation and warnings/errors, based on the above
- [ ] Quick actions & refactoring
  - [ ] "Move feature to `dev-dependencies`/`dependencies`/`build-dependencies`"
  - [ ] "Update crate to latest compatible version"
  - [ ] "Update crate to latest incompatible version"
  - [ ] Toggle `no-default-features`
  - [ ] Refactor simple dependencies to objects and back, like RustRover does
  ![image](https://github.com/rust-lang/rust-analyzer/assets/13724985/a15890b7-9bc6-4e4a-94bb-be35df636be2) 
- [ ] Link to crates like the `crates` extension: https://github.com/rust-lang/rust-analyzer/issues/13532
- [ ] Add crates to `Cargo.toml` from a use statement #10904
- [ ] Add syntax highlighting to TOML files #3565. I think the syntax highlighting is not much of a concern since `Even better TOML` does it well, but there are some good general suggestions in the issue

See the label https://github.com/rust-lang/rust-analyzer/labels/S-needs-toml

obsgolem:
My issue https://github.com/rust-lang/rust-analyzer/issues/10117 needs the needs-toml label.

tgross35:
> My issue #10117 needs the needs-toml label.

Thanks for the link, I added it to the list

alibektas:
[...]

Issue #13529, opened 2022-11-02
rust-analyzer.toml


Veykril:
This is a tracking issue/discussion for a potential `rust-analyzer.toml`. There have been a lot of requests for a per-project configuration in various issues, some clients like VSCode allow specifying settings per workspace, but usually these kinds of settings are not commited in a repo and are therefor usually not meant to necessarily configure r-a for the project specifically (like certain coding style settings, think of import merging etc.). Other related things might be requiring a special built procedure for a cargo project, prime example being https://github.com/rust-lang/rust which currently has a few suggested config lines in the dev guide which a user has to specify manually when they check out the repo locally, see https://rustc-dev-guide.rust-lang.org/building/suggested.html#configuring-rust-analyzer-for-rustc.

So there is clearly a need for a project config option like this, but here are few open questions.

- [x] We could make use of the `Cargo.toml` metadata section instead, but that would only solve the problem for cargo based projects and it feels wrong to have certain settings in there opposed to a `rust-analyzer.toml` I believe.
  - I believe a `rust-analyzer.toml` to be the correct solution here. 
- [ ] How does the priority work in regards to other mechanisms that set configurations? A `rust-analyzer.toml` should probably overwrite the client's configuration, as we cannot know if a client sents us the default config for something or if its explicitly set.
- [ ] How would r-a look for these, is nesting allowed, that is can we have hierarchies for these files? We should probably stick to what other tools like rustfmt do in this regard. Hierarchies could pose an implementation problem though, as r-a only has global configurations (global in the sense of project wide, not per workspace). I am unsure how complex it would be to change that, so this might entail a bigger architecture change.

Qix-:
Note: Highly suggest making this a dotfile (`.rust-analyzer.toml`).

emilio:
Yes please.

We need something like this to provide proper integration with our build system and right now we can only provide that for [vscode](https://searchfox.org/mozilla-central/rev/26790fecfcda622dab234b28859da721b80f3a35/python/mozbuild/mozbuild/backend/mach_commands.py#321-328) which is unfortunate.

>  How does the priority work in regards to other mechanisms that set configurations? Ideally we'd favor a client's settings over rust-analyzer.toml, but we cannot know if a client sends us the default config for something or if its explicitly set, so we might have to bite the bullet and prefer rust-analyzer.toml.

FWIW clangd has this feature ([`.clangd` files](https://clangd.llvm.org/config#files)) so may be worth checking out.

[...]

PR #15908, opened 2023-11-16
Taplo Integration Tracking

alibektas:
This PR is used for tracking how far we have gotten in terms of integrating [taplo](https://github.com/tamasfe/taplo), which we will use as our rowan based TOML parser. 

Related issues are #13529 and #15741

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "ahash"
+version = "0.7.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


flodiebold:
Two notes:
- for rust-analyzer.toml files, do we want to do the parsing in the HIR crates? I would see configuration loading and parsing as an ide-level concern and expect it to live in the ide or project-model crates (and any analysis-relevant settings would be passed from there, as happens now). It's also worth noting that at least I would not expect changes to rust-analyzer.toml to take effect immediately when you edit the file, rather only when you explicitly save, which is contrary to how the VFS and Salsa work. Also, when using RA crates as a library, one might want to control these settings in a different way.
- apart from that, for Cargo.toml, do we want the related code to live in `syntax` etc., or should they have their own crates?

flodiebold:
Actually thinking about it a bit more, we might later want to have rust-analyzer.toml in the VFS to provide IDE features while editing them, but should still only load the actual settings from disk :thinking:

Veykril:
I'll close this PR. A tracking PR doesn't make much sense, we should instead use a tracking issue for that. The changes in this PR also don't seem to be too useful right now.

PR #15911, opened 2023-11-16
fix: handle default constant values in `trait_impl_missing_assoc_item` diagnostic

Young-Flash:
A patch of https://github.com/rust-lang/rust-analyzer/pull/15895, close https://github.com/rust-lang/rust-analyzer/issues/15909

cc @Veykril

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -667,21 +667,21 @@ impl Module {
                 let items = &db.trait_data(trait_.into()).items;
                 let required_items = items.iter().filter(|&(_, assoc)| match *assoc {
                     AssocItemId::FunctionId(it) => !db.function_data(it).has_body(),
-                    AssocItemId::ConstId(_) => true,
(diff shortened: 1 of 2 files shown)


lnicola (approved):
Sorry, this should have been in today's release :pensive:.

Young-Flash:
> Sorry, this should have been in today's release 😔.

It's okay, there are relatively few maintainers while there are a lot of PR after all. Noticed that there are [a few people in R-A org](https://github.com/orgs/rust-analyzer/people), will R-A team consider expanding maintainers from contributors?

BTW, there is a downstream user affected by the underlying bug, see https://github.com/godot-rust/gdext/issues/490, maybe we should make a nigthly release in a few days?

lnicola:
> Noticed that there are [a few people in R-A org](https://github.com/orgs/rust-analyzer/people), will R-A team consider expanding maintainers from contributors?

Sure, but most (or almost all, if you could see the full list) of the people in there haven't been very active lately.

> maybe we should make a nigthly release in a few days

The nightlies are made nightly, so the fix is already out.

PR #15912, opened 2023-11-16
chore: remove unused `PhantomData`

Sarrus1:
This PR removes an unused `PhantomData` in `FileItemTreeId`.

*Note:* I am not sure how this should be implemented, maybe as a type instead of a wrapper struct? I'd be happy to do so if needed 👍


Sarrus1:
Bumping this just in case: is there anything I can add to this PR?

PR #15922, opened 2023-11-17
Updated instructions for installing rust-analyzer under Gentoo.

paul-jewell:
No need to install the guru overlay to install rust-analyzer. This is now installed based on use flag settings for dev-lang/rust and dev-lang/rust-bin. This pull request changes the instructions in the user manual.

Note: rust-analyzer is not available in the guru repository, so the old instructions no longer work.

### DIFF
--- a/docs/user/manual.adoc
+++ b/docs/user/manual.adoc
@@ -186,18 +186,7 @@ $ pacman -S rust-analyzer
 
 ==== Gentoo Linux
 
-`rust-analyzer` is available in the GURU repository:
-
-- https://gitweb.gentoo.org/repo/proj/guru.git/tree/dev-util/rust-analyzer?id=9895cea62602cfe599bd48e0fb02127411ca6e81[`dev-util/rust-analyzer`] builds from source
-- https://gitweb.gentoo.org/repo/proj/guru.git/tree/dev-util/rust-analyzer-bin?id=9895cea62602cfe599bd48e0fb02127411ca6e81[`dev-util/rust-analyzer-bin`] installs an official binary release
-
-If not already, GURU must be enabled (e.g. using `app-eselect/eselect-repository`) and sync'd before running `emerge`:
-
-[source,bash]
-----
-$ eselect repository enable guru && emaint sync -r guru
-$ emerge rust-analyzer-bin
-----
+`rust-analyzer` is installed when the `rust-analyzer` use flag is set for dev-lang/rust or dev-lang/rust-bin. You also need to set the `rust-src` use flag.
 
 ==== macOS
 


lnicola:
Thanks. Are you sure those are the same thing? The Rust toolchain also includes rust-analyzer, but it's usually pretty outdated. I would expect `dev-util/rust-analyzer-bin` to match this repo and `dev-lang/rust-bin` to match the official toolchain (i.e. what you'd get by using `rustup`).

paul-jewell:
Thanks Inicola. First of all, please accept my apologies - I was not familiar with the fact that the version of rust-analyzer distributed with rust is older. I am working on setting up the helix editor and it needs rust-analyzer, but the original instructions failed to provide it, hence my pull request. I can't find reference to dev-util/rust-analyzer-bin in the standard gentoo portage repository, or in the guru repository. Do you know if it is available in a different overlay?

lnicola:
The reference is in the paragraph you've deleted, I don't know if it still works.

lnicola:
That is, https://github.com/rust-lang/rust-analyzer/pull/15922/files#diff-acd52621eba10b506adcd0d993d53530ee7f70b2056f390f8e0cc4928def039fL192.

paul-jewell:
The guru overlay referenced in the original paragraph no longer includes dev-util/rust-analyzer - looking at this comment on their git repo, it was removed when the rust-analyzer flag was implemented on the dev-lang/rust and dev-lang/rust-bin packages: [here](https://gitweb.gentoo.org/repo/proj/guru.git/commit/?id=6d9e35ab39bfbb03a15b2fd21248db612880b9bc)

I installed the overlay to confirm that it is indeed no longer present. In the absence of a better solution, the instructions in my pull request will at least give gentoo users a working version of rust-analyzer, albeit an older version. The version reported by rust-analyzer is currently is 1.71.1, aligned with the rust compiler. The date on the rust compiler release is 2023-08-03, so I guess the rust-analyzer version comes from around the same period.

lnicola:
> so I guess the rust-analyzer version comes from around the same period

rust-analyzer gets synced with downstream once a week, but changes don't get ported to the beta branch. So for example, 1.77 will be released on March 21, gets branched on February 2, so only includes a RA version before that date, which is around January 29.

And that's in the ideal case. At time syncs were suspended for months at a time, usually because of technical issues.

docs/user/manual.adoc
```diff
@@ -186,18 +186,7 @@ $ pacman -S rust-analyzer
...
-- https://gitweb.gentoo.org/repo/proj/guru.git/tree/dev-util/rust-analyzer-bin?id=9895cea62602cfe599bd48e0fb02127411ca6e81[`dev-util/rust-analyzer-bin`] installs an official binary release
-
-If not already, GURU must be enabled (e.g. using `app-eselect/eselect-repository`) and sync'd before running `emerge`:
-
-[source,bash]
-----
-$ eselect repository enable guru && emaint sync -r guru
-$ emerge rust-analyzer-bin
-----
+`rust-analyzer` is installed when the `rust-analyzer` use flag is set for dev-lang/rust or dev-lang/rust-bin. You also need to set the `rust-src` use flag.
```
lnicola, line 189:
```suggestion
`rust-analyzer` is installed when the `rust-analyzer` use flag is set for `dev-lang/rust` or `dev-lang/rust-bin`.
You also need to set the `rust-src` use flag.

This is similar to using the `rustup` component, so it will install a version that lags behind the GitHub releases.
```

paul-jewell: I'm happy with your suggested changes @lnicola  - thanks. If a gentoo user wants to use the latest rust-analyzer version, they will need to make their own ebuild (in a local overlay) to install the latest release. I think the removal from the guru repository was perhaps not the best decision, given the different version situation, so I will ask whether it is worth re-instating the ebuild there, or even in the main gentoo repository.

Veykril:
Whats the status of this PR (I haven no clue about gentoo)

paul-jewell:
If you use rust in Gentoo with the system installed tools, following the instructions in my pull request will ensure you have the version of rust-analyzer installed which came with the compiler release. The alternative is to use rustup at a user level, and install rust-analyzer that way - something I didn't document in the pull request. The instructions to use the guru overlay are not valid at present, since rust-analyzer has been removed as a package from that overlay. I recommend merging the pull request, and if you want me to document the rustup approach as well, then let me know and I will make another pull request to implement.

Veykril:
If the guru version was removed than this seems to be better indeed already, adding rustup wouldn't hurt either here though but it doesn't matter  too much.
@bors r+

lnicola:
@paul-jewell did you end up asking the guru maintainers about this?

I'll file a follow-up PR with my proposed addition in https://github.com/rust-lang/rust-analyzer/pull/15922#discussion_r1398224723.

PR #15925, opened 2023-11-17
internal: Remove debugging code in path resolution

### LINKED ISSUE #15924: Many unnecessary allocations in path resolution
https://github.com/rust-lang/rust-analyzer/blob/bc978217b32cd10abd140bec1724a218f43629fb/crates/hir-def/src/nameres/path_resolution.rs#L187-L193

This does a bunch of string formatting on the happy path. According to my dhat analysis (opening rust-lang/rust with RA running in dhat and then waiting for a few minutes (it is very slow)), this accounts for 13% of the allocated blocks inside rust-analyzer. I don't think it uses much memory, but it's probably quite slow. I don't think my analysis is necessarily representative of normal usage, as it only really captured startup, but it still seems pretty relevant.

image

**rust-analyzer version**: current master bc978217b32cd10abd140bec1724a218f43629fb

**rustc version**: 1.76.0-nightly, doesn't matter

**relevant settings**: the recommended settings for rust-lang/rust, provided there in `src/etc/rust_analyzer_settings.json`

lnicola:
Closes #15924

### DIFF
--- a/crates/hir-def/src/nameres/path_resolution.rs
+++ b/crates/hir-def/src/nameres/path_resolution.rs
@@ -183,15 +183,6 @@ impl DefMap {
         shadow: BuiltinShadowMode,
         expected_macro_subns: Option<MacroSubNs>,
     ) -> ResolvePathResult {
-        let graph = db.crate_graph();
-        let _cx = stdx::panic_context::enter(format!(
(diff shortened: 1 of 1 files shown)


lnicola:
This might have sped up `analysis-stats` by about 3 seconds out of 71.

PR #15932, opened 2023-11-19
internal: store indeterminate imports alongside unresolved ones

rosefromthedead:
Seems to give some speedups, especially in diesel inference: 298s -> 272s on my machine.

### DIFF
--- a/crates/hir-def/src/nameres/collector.rs
+++ b/crates/hir-def/src/nameres/collector.rs
@@ -104,7 +104,6 @@ pub(super) fn collect_defs(db: &dyn DefDatabase, def_map: DefMap, tree_id: TreeI
         deps,
         glob_imports: FxHashMap::default(),
         unresolved_imports: Vec::new(),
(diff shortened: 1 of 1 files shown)


rosefromthedead:
Err, maybe it doesn't make sense for this to affect inference. I'm new to the perf infrastructure around here, is there a recommended way of measuring the difference?

lnicola:
`analysis-stats` is the way to measure the difference. I didn't test on `diesel`, but on `self` it doesn't seem to have any effect on the performance _or_ the accuracy of type inference (which is a good thing).

rosefromthedead:
Yeah, my measurements had a lot more noise than I thought. I'm not sure this PR makes a difference to anything now.

Veykril:
I also don't see where this would really save anything, we are getting rid of the `extend` in the `collect` but that's not a hot path. Though the `retain_mut` is probably a bit better in terms of allocation saving

Veykril:
Closing as there seems to be no benefit from doing this

PR #15933, opened 2023-11-19
feat: resolve inherent and implemented associated items in docs

### LINKED ISSUE #9694: Support doc links to associated functions on other structs
```rust
struct F;
impl F {
    // expect to go here but it fails instead
    fn linkme() {}
}

/// See [`F::linkme`]
//           ^-- go to definition here
fn f1() {}
```

If I do the above go to definition, it doesn't find a target. `rustdoc` appears to compile it fine.

The example from #9617 linking to other things on the same top level item *does* work, it seems to just be linking to things from other top level items that's a trouble here.

Version: 5983d3745ac8490c1c2798fdc963aa9fc803dd03

71:
This partially fixes #9694.

Supported:
- Trait methods and constants.
  * Due to resolution differences pointed out during the review of the PR, trait associated types are _not_ supported.
- Inherent methods, constants and associated types.
  * Inherent associated types are a [nightly feature](https://github.com/rust-lang/rust/issues/8995), and are supported with no additional work in this PR.

Screenshot of VS Code running with the change:

image

You can see that the items are resolved (excl. trait associated types) since they are semantically highlighted in the doc comment.

### DIFF
--- a/crates/hir/src/attrs.rs
+++ b/crates/hir/src/attrs.rs
@@ -1,5 +1,7 @@
 //! Attributes & documentation for hir types.
 
+use std::ops::ControlFlow;
+
 use base_db::FileId;
 use hir_def::{
     attr::AttrsWithOwner,
@@ -13,13 +15,13 @@ use hir_expand::{
     name::Name,
     span_map::{RealSpanMap, SpanMapRef},
 };
-use hir_ty::db::HirDatabase;
+use hir_ty::{db::HirDatabase, method_resolution};
 use syntax::{ast, AstNode};
 
 use crate::{
     Adt, AsAssocItem, AssocItem, BuiltinType, Const, ConstParam, DocLinkDef, Enum, ExternCrateDecl,
-    Field, Function, GenericParam, Impl, LifetimeParam, Macro, Module, ModuleDef, Static, Struct,
-    Trait, TraitAlias, TypeAlias, TypeParam, Union, Variant, VariantDef,
+    Field, Function, GenericParam, HasCrate, Impl, LifetimeParam, Macro, Module, ModuleDef, Static,
+    Struct, Trait, TraitAlias, Type, TypeAlias, TypeParam, Union, Variant, VariantDef,
 };
 
 pub trait HasAttrs {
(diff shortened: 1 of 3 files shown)


crates/ide-completion/src/completions/pattern.rs
```diff
@@ -154,7 +154,6 @@ pub(crate) fn complete_pattern_path(
                     }
 
                     ctx.iterate_path_candidates(&ty, |item| match item {
-                        AssocItem::TypeAlias(ta) => acc.add_type_alias(ctx, ta),
```
Veykril, line 157: Why is this gone now?

71: Without removing this, a couple of snapshot tests fail because completions for type aliases are emitted several times / duplicated. I figured this code was a workaround the fact that `method_resolution.rs` did not consider `TypeAlias`es, which changed in this PR, so I removed these explicit iterations of `TypeAlias`es.

Same for all your other comments for changes in `src/completions/*`. This is the main thing I'm uncertain about in this PR; it looks like adding support for `TypeAlias`es had impacts in other parts of the code (which led me to make these changes, hence the question "should be only handle `TypeAlias`es when resolving doc comments" above.

With that said, it looks like handling `TypeAlias`es in a generic way in `method_resolution.rs` allows us to explicitly handle them in other parts of the codebase, so this may be a good thing. I am not aware of _why_ it was handled differently before, though; it's possible that there were strong reasons for handling it differently which this PR breaks (subtly enough that additional tests do not fail).

71: For instance if I undo this particular diff, I get the following test failure:

```
---- tests::pattern::enum_qualified stdout ----


error: expect test failed
   --> crates/ide-completion/src/tests/pattern.rs:314:9

You can update all `expect!` tests by running:

    env UPDATE_EXPECT=1 cargo test

To update a single test, place the cursor on `expect` token and use `run` feature of rust-analyzer.

Expect:
----
ct ASSOC_CONST const ASSOC_CONST: ()
bn RecordV {…} RecordV { field$1 }$0
bn TupleV(…)   TupleV($1)$0
bn UnitV       UnitV$0

----

Actual:
----
ct ASSOC_CONST const ASSOC_CONST: ()
ta AssocType   type AssocType = ()
bn RecordV {…} RecordV { field$1 }$0
bn TupleV(…)   TupleV($1)$0
bn UnitV       UnitV$0

----

Diff:
----
ct ASSOC_CONST const ASSOC_CONST: ()
ta AssocType   type AssocType = ()    (NOTE: THIS LINE WAS ADDED)
bn RecordV {…} RecordV { field$1 }$0
bn TupleV(…)   TupleV($1)$0
bn UnitV       UnitV$0

----
```

For this particular case it looks like the problem is that the `AssocType` is wrongly suggested, but I do remember that for other changes there were duplicate completions.

Veykril (changes requested):
So (finally) revisiting this, it seems weird that method probing / resolution now also supports type aliases, as their resolution works a bit differently. So this seems to be the wrong way to approach this. We might just need to somehow expose a similar API but for type alias resolution that can be used for the doc resolution, I am unsure in what way though.

Veykril:
Can you squash the last two commits into one? (Or all three up to you), the last commit will noise up git blame unnecessarily.

71:
I removed support for type aliases and tried reusing existing features more than before, which significantly simplified the PR.

Veykril (approved):

Veykril:
Thanks! I've wanted this for a long time.
@bors r+

PR #15938, opened 2023-11-20
feat: add hover display for trait assoc items

Young-Flash:
This PR enable preview assoc items when hover on `trait`

image

inspired by https://github.com/rust-lang/rust-analyzer/pull/15847

### DIFF
--- a/crates/hir/src/display.rs
+++ b/crates/hir/src/display.rs
@@ -598,7 +598,7 @@ impl HirDisplay for Trait {
 
         let assoc_items = self.items(f.db);
         let assoc_items_size = assoc_items.len();
-        let max_display_size = f.max_size.unwrap_or(assoc_items_size);
+        let limited_size = f.limited_size.unwrap_or(assoc_items_size);
         if assoc_items.is_empty() {
             f.write_str(" {}")?;
         } else {
@@ -617,7 +617,7 @@ impl HirDisplay for Trait {
                     }
                 };
                 f.write_str(",\n")?;
-                if index + 1 == max_display_size && index + 1 != assoc_items_size {
+                if index + 1 == limited_size && index + 1 != assoc_items_size {
                     f.write_str("    ...\n")?;
                     break;
                 }
--- a/crates/hir/src/display.rs
+++ b/crates/hir/src/display.rs
@@ -17,10 +17,7 @@ use hir_ty::{
 };
 
 use crate::{
-    Adt, AsAssocItem, AssocItemContainer, Const, ConstParam, Enum, ExternCrateDecl, Field,
-    Function, GenericParam, HasCrate, HasVisibility, LifetimeParam, Macro, Module, SelfParam,
-    Static, Struct, Trait, TraitAlias, TupleField, TyBuilder, Type, TypeAlias, TypeOrConstParam,
-    TypeParam, Union, Variant,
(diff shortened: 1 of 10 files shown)


flodiebold:
How does this look when hovering `Iterator`?

lnicola:
image

image

lnicola:
Looks like we don't display `Where T: Tr,` properly, and we should probably respect `#[unstable]` and `#[doc(hidden)]`.

flodiebold:
IMO hiding the documentation below this is not worth it, at least not by default. Or we should at least limit it (like rustdoc collapses the long list of methods).

Veykril:
Agreed, if we do this we should at least have a config for limiting the number of items / disabling it fully, similarly how we should still do such a config for struct fields hover -> https://github.com/rust-lang/rust-analyzer/pull/15847#issuecomment-1805777597

Young-Flash:
> How does this look when hovering `Iterator`?

kind of ugly haha

Veykril:
@Young-Flash what is the status on this PR?

Young-Flash:
will continue working on this after my vacation

Young-Flash:
add `"rust-analyzer.traitItemDisplayNum"` (default 7) config item, now it looks like the following

image

Veykril:
Yes let's have a config for those as well

Young-Flash:
> Yes let's have a config for those as well

will make it at next PR

Young-Flash:
how about merge this for now (anything else I can do here?)

crates/rust-analyzer/src/config.rs
```diff
@@ -591,6 +590,9 @@ config_data! {
         /// Show documentation.
         signatureInfo_documentation_enable: bool                       = "true",
 
+        /// How many trait item display on hover.
+        traitItemDisplayNum: Option<usize> = "7",
```
Veykril, line 594: Something like 
```suggestion
        /// How many associated items of a trait to display when hovering a trait.
        hover_show_traitAssocItems: Option<usize> = "null",
```
follows our naming scheme better

crates/ide-db/src/defs.rs
```diff
@@ -24,6 +24,12 @@ use crate::documentation::{Documentation, HasDocs};
 use crate::famous_defs::FamousDefs;
 use crate::RootDatabase;
 
+#[derive(Default)]
+pub struct HoverDisplayConfig {
+    pub trait_item_display_num: Option<usize>,
+    // todo: add config for struct & enum
+}
```
Veykril, line 31: Let's put this into the hover module, and instead of modifying `Definition::label` we check if the `Definition` is a trait before calling the label function explicitly in the hover module.

crates/hir/src/display.rs
```diff
@@ -595,6 +595,36 @@ impl HirDisplay for Trait {
         let def_id = GenericDefId::TraitId(self.id);
         write_generic_params(def_id, f)?;
         write_where_clause(def_id, f)?;
+
+        let assoc_items = self.items(f.db);
+        let assoc_items_size = assoc_items.len();
+        let max_display_size = f.max_size.unwrap_or(assoc_items_size);
```
Veykril, line 601: `max_size` is meant to limit the overall character count of a display, but the config here is supposed to limit by the number of assoc items. So we'll need to add a specific extra field to the formatter unfortunately that describes this limit and use that appropriately here.

Veykril:
Thanks! (will do a minor adjustment as a follow up)
@bors r+

PR #15939, opened 2023-11-20
Update perf-event to 0.4.8

gigaroby:


lnicola:
Thanks, any particular reason for this?

gigaroby:
Just convenience for our internal process, it's easier to import the latest version of dependencies. If this kind of version bump is not appreciated, I'm happy to undo it.

lnicola:
No, it's fine, but I think we lag on others too, so I was wondering if there's anything in particular about this one.

Veykril:
Looks like we can't bump it? (given CI)

PR #15944, opened 2023-11-21
Update Arch Linux package URL in manual.adoc

brunobell:
The old URL returns 404 now.


lnicola:
Did we get promoted to `extra`? Nice!

@bors r+

PR #14775, opened 2023-05-10
feat: add #[doc(alias(..))]-based method completions

hecatia-elegua:
Screenshot showing a completion when you type a doc alias instead of a real name

PR #15956, opened 2023-11-24
Improve completion label details display

ahlinc:
This PR improves completion label details display by separating trait and alias info from the `label` LSP field to an additional existing `label_detail` field. Changes look like the following:

### Before

Screenshot from 2023-11-24 12-03-08

### After

Screenshot from 2023-11-24 12-21-57

_All existing tests are passed without any changes in test themselves logic._

### DIFF
--- a/crates/ide-completion/src/item.rs
+++ b/crates/ide-completion/src/item.rs
@@ -26,6 +26,10 @@ use crate::{
 pub struct CompletionItem {
     /// Label in the completion pop up which identifies completion.
     pub label: SmolStr,
+    /// Addition label details in the completion pop up that are
+    /// displayed and aligned on the right side after the label.
+    pub label_detail: Option<SmolStr>,
+
     /// Range of identifier that is being completed.
     ///
     /// It should be used primarily for UI, but we also use this to convert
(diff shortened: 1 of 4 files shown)


lnicola (approved):

crates/ide-completion/src/tests.rs
```diff
@@ -150,16 +150,29 @@ fn render_completion_list(completions: Vec<CompletionItem>) -> String {
     fn monospace_width(s: &str) -> usize {
         s.chars().count()
     }
-    let label_width =
-        completions.iter().map(|it| monospace_width(&it.label)).max().unwrap_or_default().min(22);
+    let label_width = completions
+        .iter()
+        .map(|it| {
+            monospace_width(&it.label)
+                + monospace_width(it.label_detail.as_deref().unwrap_or_default())
```
lnicola, line 157: Should we add 1 here for the space?

crates/ide-completion/src/item.rs
```diff
@@ -454,10 +456,17 @@ impl Builder {
         if let [import_edit] = &*self.imports_to_add {
             // snippets can have multiple imports, but normal completions only have up to one
             if let Some(original_path) = import_edit.original_path.as_ref() {
-                label = SmolStr::from(format!("{label} (use {})", original_path.display(db)));
+                label_detail = Some(SmolStr::from(format!(
```
lnicola, line 459: This is a bit hard to follow, but I don't have any suggestions :shrug:.

ahlinc: @lnicola I rewrote logic to avoid reassignments and force pushed it the PR branch.

ahlinc: But I don't know would @bors see my latest changes and take them into account.

ahlinc: @lnicola As I understand @bors will complete this PR as it saw it on the https://github.com/rust-lang/rust-analyzer/commit/f8bf1dc15d770642b252ed6e35bcfb84ef074fda commit.
Can I open an another PR with a small fix that force pushed here?

lnicola:
I think it's using the new version now: https://github.com/rust-lang/rust-analyzer/commit/87337283235bb4056fb8f8e06127d7e44e58322c.

lnicola:
image

hecatia-elegua:
~I think this might have broken #14775.~
~For structs it still seems to work though.~

edit: okay no, that feature only works if the alias starts with an uppercase letter for some reason.

lnicola:
Consider filing a new issue.

PR #15960, opened 2023-11-24
Replace `option.map(cond) == Some(true)` with `option.is_some_and(cond)`

dtolnay:
Extracted from https://github.com/rust-lang/rust/pull/118253.


lnicola:
Thanks!

@bors r+

PR #15962, opened 2023-11-25
fix: add fallback for completion label details

ahlinc:
This PR adds a fallback to a previous implementation in a case when the label detail field isn't supported by LSP client and the support isn't reported by the LSP initialize request. In this case additional info about trait and aliases would be merged into the label field as it was before the #15956 PR.


ahlinc:
@lnicola could you review and merge this small fix addition for my [previous](https://github.com/rust-lang/rust-analyzer/pull/15956) PR?

lnicola (approved):

lnicola:
Saw it before, but couldn't look. Thanks!

@bors r+

PR #15970, opened 2023-11-26
fix variant resolve for type alias

### LINKED ISSUE #15943: Generic type alias fails to resolve fields
**rust-analyzer version**: 0.3.1740-standalone
**rustc version**:  1.73.0

```rust
type Wrap<T> = T;

enum X {
    A { cool: u32, stuff: u32 },
    B,
}

fn main() {
    let wrapped = Wrap::<X>::A {
        // autocomplete for fields breaks here
        cool: 100,
        stuff: 100,
    };

    // autocomplete for fields breaks here
    // RA fails to resolve `cool`'s type (lists as unknown)
    if let Wrap::<X>::A { cool, .. } = &wrapped {}
}
```

And a screenshot of the failed type hint:
image

## Context

I'm [building a DSL for Bevy Scenes](https://github.com/bevyengine/bevy/discussions/9538) with the goal of supporting as many RA niceties as possible (ex: autocomplete, go to definition, etc). I [created a discussion here a few months ago](https://github.com/rust-lang/rust-analyzer/discussions/15452) and got very helpful advice, which has generally worked!

However when building support for enums in my DSL, I encountered a limitation in the Rust type system that required the use of [this type alias wrapper workaround](https://github.com/rust-lang/rust/issues/86935) (or an experimental feature ... which isn't an option for us).

The actual proc_macro code in my impl looks like this:

```rust
 if let Wrap::<<#path as Schematic>::Props>::#variant { #(#fields,)* .. } = &mut props {
 }
```

If I replace `Wrap::<<#path as Schematic>::Props>` with the "actual" type, RA field autocomplete does work for the enum in my DSL. Sadly, I cannot name the type in the macro (as it is defined outside of the macro), so I need to rely on the type system to resolve this.

Given that RA fails to resolve the fields in the simplified `if let Wrap::<X>::A { cool, .. } = &wrapped {}`, I'm assuming that if RA can handle autocomplete for this case, it can probably handle autocompletion in the Bevy DSL proc_macro as well.

Austaras:
Closes #15943

### DIFF
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -1152,20 +1152,15 @@ impl<'a> InferenceContext<'a> {
                 (ty, variant)
             }
             TypeNs::TypeAliasId(it) => {
-                let container = it.lookup(self.db.upcast()).container;
-                let parent_subst = match container {
-                    ItemContainerId::TraitId(id) => {
-                        let subst = TyBuilder::subst_for_def(self.db, id, None)
-                            .fill_with_inference_vars(&mut self.table)
-                            .build();
-                        Some(subst)
-                    }
-                    // Type aliases do not exist in impls.
-                    _ => None,
+                let resolved_seg = match unresolved {
+                    None => path.segments().last().unwrap(),
+                    Some(n) => path.segments().get(path.segments().len() - n - 1).unwrap(),
                 };
-                let ty = TyBuilder::def_ty(self.db, it.into(), parent_subst)
-                    .fill_with_inference_vars(&mut self.table)
-                    .build();
+                let substs =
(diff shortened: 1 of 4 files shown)


crates/hir-ty/src/infer.rs
```diff
@@ -1152,20 +1152,14 @@ impl<'a> InferenceContext<'a> {
                 (ty, variant)
             }
             TypeNs::TypeAliasId(it) => {
-                let container = it.lookup(self.db.upcast()).container;
-                let parent_subst = match container {
-                    ItemContainerId::TraitId(id) => {
```
Austaras, line 1157: It doesn't seem possible to encounter this -- trait assoc type would either results in Self or Trait.

cart:
Thanks for the fix!

PR #15971, opened 2023-11-26
fix: don't make `MissingMatchArms` diagnostic for empty match body

Young-Flash:
before 
before

after
after

close https://github.com/rust-lang/rust-analyzer/issues/15954

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -1914,17 +1914,20 @@ impl DefWithBody {
                             if let ast::Expr::MatchExpr(match_expr) =
                                 &source_ptr.value.to_node(&root)
                             {
-                                if let Some(scrut_expr) = match_expr.expr() {
-                                    acc.push(
-                                        MissingMatchArms {
-                                            scrutinee_expr: InFile::new(
-                                                source_ptr.file_id,
-                                                AstPtr::new(&scrut_expr),
-                                            ),
-                                            uncovered_patterns,
-                                        }
-                                        .into(),
-                                    );
+                                match match_expr.expr() {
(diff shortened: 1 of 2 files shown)


lnicola:
Looks like the test is failing because of the diagnostic range.

Young-Flash:
> Looks like the test is failing because of the diagnostic range.

yeah, are there any mark that indicate diagnostic range with the same begin and end like `23..23`?

Veykril:
You can just make a variant of `check_diagnostics` that also discards syntax errors for this one otherwise

lnicola:
It would be nice to make this more like `check_diagnostics`, that's quite a bit of duplicated code.

And making filtering syntax errors in `check_diagnostics_with_config` doesn't cause any test to fail.

crates/ide-diagnostics/src/tests.rs
```diff
@@ -154,6 +154,65 @@ pub(crate) fn check_diagnostics_with_config(config: DiagnosticsConfig, ra_fixtur
...
+    config.disabled.insert("inactive-code".to_string());
+    for file_id in files {
+        let line_index = db.line_index(file_id);
+        let diagnostics = super::diagnostics(&db, &config, &AssistResolveStrategy::All, file_id);
+
+        let expected = extract_annotations(&db.file_text(file_id));
+        let mut actual = diagnostics
+            .into_iter()
+            .filter_map(|d| {
+                if d.message.contains("Syntax Error: ") {
```
lnicola, line 170: I think we can check `d.code` instead.

lnicola:
Wow, that actually works?

@bors r+

Young-Flash:
> Wow, that actually works?

exactly 😄

PR #15983, opened 2023-11-28
In rustc workspace, use local rustc_private crates

eggyal:
Prior to this PR, when working in the rustc workspace (with `rust-analyzer.rustc.source` set to the workspace root) and upon encountering packages with `package.metadata.rust-analyzer.rustc_private = true`, transitive dependencies of `rustc_driver` are added to the crate graph with `CrateOrigin::Rustc` and conflict with those same crates added to the crate graph with `CrateOrigin::Local` from the workspace.  This conflict is causing some breakage, e.g. errors due to failed proc-macro resolution.

This PR looks up and uses instead the local crate, if available.

### DIFF
--- a/crates/project-model/src/workspace.rs
+++ b/crates/project-model/src/workspace.rs
@@ -2,7 +2,13 @@
 //! metadata` or `rust-project.json`) into representation stored in the salsa
 //! database -- `CrateGraph`.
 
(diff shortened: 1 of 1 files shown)


eggyal:
This isn't the fix that I thought it was.

PR #15990, opened 2023-11-29
feat: add trait_impl_reduntant_assoc_item diagnostic

Young-Flash:
part of https://github.com/rust-lang/rust-analyzer/issues/15958, will try to add quickfix for the diagnostic if this PR is ok with you guys


crates/hir/src/diagnostics.rs
```diff
@@ -310,3 +311,10 @@ pub struct TraitImplMissingAssocItems {
     pub impl_: AstPtr<ast::Impl>,
     pub missing: Vec<(Name, AssocItem)>,
 }
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct TraitImplReduntantAssocItems {
```
Veykril, line 316:
```suggestion
pub struct TraitImplRedundantAssocItems {
```
and where else you wrote retuntant

crates/hir/src/lib.rs
```diff
@@ -685,6 +685,26 @@ impl Module {
...
 
+                let reduntant: Vec<_> = impl_assoc_items_scratch
+                    .iter()
+                    .filter(|(id, name)| {
+                        !items.iter().any(|(impl_name, impl_item)| {
+                            discriminant(impl_item) == discriminant(id) && impl_name == name
+                        })
+                    })
+                    .map(|(item, name)| (name.clone(), AssocItem::from(*item)))
+                    .collect();
```
Veykril, line 696:
```suggestion
                    .map(|(item, name)| (name.clone(), AssocItem::from(*item)));
```
No need to collect here

crates/ide-diagnostics/src/handlers/trait_impl_reduntant_assoc_item.rs
```diff
@@ -0,0 +1,88 @@
...
+    let db = ctx.sema.db;
+
+    let root = ctx.sema.db.parse_or_expand(d.file_id);
+    let range = root.text_range();
+    let impl_ = d.impl_.to_node(&root);
+    let trait_name = if let Some(trait_name) = trait_name(ctx, impl_) {
+        trait_name
+    } else {
+        String::from("{unkown}")
+    };
```
Veykril, line 24: We have the `hir::Trait` when constructing the diagnostic already, so we might as well just pass that into the `TraitImplRedundantAssocItems` struct so we can fetch its name through the hir here directly instead of going through the AST

crates/ide-diagnostics/src/handlers/trait_impl_reduntant_assoc_item.rs
```diff
@@ -0,0 +1,88 @@
...
+                .map(|it| it.syntax().value.text_range())
+                .unwrap_or(range),
+        ),
+    };
+
+    Diagnostic::new(
+        DiagnosticCode::RustcHardError("E0407"),
+        format!("{reduntant_item_name} is not a member of trait `{trait_name}`"),
+        diagnostic_range,
+    )
```
Veykril, line 48: Take a look at the `trait_impl_missing_assoc_item` assist here, you should be able to structure this similarly to that which is a bit more ocncise (that is the `diagnostic_range` fetching mainly)

Young-Flash: for `trait_impl_missing_assoc_item` diagnostic, the `diagnostic_range` just cover the trait name like the following:

https://github.com/rust-lang/rust-analyzer/blob/c9d189d1375e59a6c9b4d62fdede94ade001f6ee/crates/ide-diagnostics/src/handlers/trait_impl_missing_assoc_item.rs#L67-L71

while for `trait_impl_redundant_assoc_item` diagnostic, we want to cover every redundant assoc item inside the impl body, so the `diagnostic_range` fetching way of `trait_impl_missing_assoc_item` maynot fit here

Young-Flash:
@Veykril @lnicola could you help review this when you are free?

Veykril:
Small nit but otherwise lgtm. We should definitely look into adding some more utility functions for creating better diagnostic text ranges for things (currently most assists still create the ranges in some random ways).

@bors delegate+

lnicola:
image

saiintbrisson:
Hi! This feature introduced a bug when unwrapping this `file_id`:
https://github.com/rust-lang/rust-analyzer/blob/932d85b52946d917deab2c23ead552f7f713b828/crates/ide-diagnostics/src/handlers/trait_impl_redundant_assoc_item.rs#L47

The function returns `None` if the file is a macro file:
https://github.com/rust-lang/rust-analyzer/blob/932d85b52946d917deab2c23ead552f7f713b828/crates/span/src/lib.rs#L253-L259

PR #15991, opened 2023-11-29
Use build scripts from rustc source workspace

eggyal:
The status quo ante (since #14348)  is that, if a workspace has one or more packages with `package.metadata.rust-analyzer.rustc_private = true`, build scripts for those `rustc_private` crates are located either from building the workspace itself (if the workspace is the rustc source root) or else from the toolchain's `target-libdir` (as determined from invoking `rustc --print target-libdir`).

The problem with this approach is that, when working on rustc, there are such workspaces that are not themselves the rustc source root—and the toolchain `target-libdir` in such situations is not the appropriate one for use by the stage0 proc-macro server.  For example, `rustc_codegen_cranelift` is opened as a workspace, it is not the rustc source root, and (owing to its `rust-toolchain` override file) its `target-libdir` is that of a specific nightly compiler; attempting to use the proc-macro dylibs therein gives rise to an ABI mismatch error.

The correct build scripts are instead located from building the opened rustc source root workspace, but finding these when building another workspace's crate graph requires the rustc source root workspace already to have been identified (and built).  This PR therefore ensures that any such workspace is processed first, and that a reference thereto is passed into the crate graph construction functions for each workspace.

Incidentally, while working on debugging this issue, I noticed that such arising `CrateOrigin::Rustc` crates end up as duplicates of the `CrateOrigin::Local` crates from the opened rustc source root workspace—I already have a little deduplication tweak to the `CrateGraph::extend` function which I will open as a separate PR (and this pretty much halves the number of crates indexed by RA when working on rustc).

### DIFF
--- a/crates/load-cargo/src/lib.rs
+++ b/crates/load-cargo/src/lib.rs
@@ -40,7 +40,7 @@ pub fn load_workspace_at(
 ) -> anyhow::Result<(AnalysisHost, vfs::Vfs, Option<ProcMacroServer>)> {
     let root = AbsPathBuf::assert(std::env::current_dir()?.join(root));
     let root = ProjectManifest::discover_single(&root)?;
-    let mut workspace = ProjectWorkspace::load(root, cargo_config, progress)?;
+    let mut workspace = ProjectWorkspace::load(root, cargo_config, progress, false)?;
 
     if load_config.load_out_dirs_from_check {
         let build_scripts = workspace.run_build_scripts(cargo_config, progress)?;
@@ -81,6 +81,7 @@ pub fn load_workspace(
             vfs.file_id(&path)
         },
         extra_env,
+        None,
     );
     let proc_macros = {
         let proc_macro_server = match &proc_macro_server {
--- a/crates/project-model/src/lib.rs
+++ b/crates/project-model/src/lib.rs
@@ -50,7 +50,7 @@ pub use crate::{
     manifest_path::ManifestPath,
     project_json::{ProjectJson, ProjectJsonData},
     sysroot::Sysroot,
-    workspace::{CfgOverrides, PackageRoot, ProjectWorkspace},
+    workspace::{CfgOverrides, PackageRoot, ProjectWorkspace, RustcWorkspace},
 };
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
--- a/crates/project-model/src/tests.rs
+++ b/crates/project-model/src/tests.rs
@@ -11,8 +11,8 @@ use rustc_hash::FxHashMap;
 use serde::de::DeserializeOwned;
 
 use crate::{
-    CargoWorkspace, CfgOverrides, ProjectJson, ProjectJsonData, ProjectWorkspace, Sysroot,
-    WorkspaceBuildScripts,
+    CargoWorkspace, CfgOverrides, ProjectJson, ProjectJsonData, ProjectWorkspace, RustcWorkspace,
+    Sysroot, WorkspaceBuildScripts,
 };
 
 fn load_cargo(file: &str) -> (CrateGraph, ProcMacroPaths) {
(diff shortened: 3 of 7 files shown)


Zalathar:
Crate duplication also causes other problems (e.g. #15656), so I’m curious to see whether your upcoming PR has the side-effect of fixing that (for rustc) too.

eggyal:
On the crate deduplication front, I've eliminated a bunch of cases when merging sysroot (`CrateOrigin::Lang`) crates from different workspaces by ignoring the `SysrootPublicDeps` that rust-analyzer synthetically adds to them if they're in a loaded Cargo workspace (which also is the origin of various cyclic dependency errors that one sees).  This change alone has a huge impact because pretty much everything depends upon them and the failure to deduplicate them causes dependency comparisons in downstream deduplications to fail, thus causing cascading failures/duplicates throughout the entire crate graph.

I haven't looked into the issue you've linked to see whether that alone would be sufficient, but the reduction in duplicates is very marked.  However, I've been holding off opening the PR because there remain some further interesting duplicates that I think should also be eliminated... but perhaps it makes more sense to do this in small steps.  I'll probably open a PR tomorrow either way.

Veykril:
I wonder, wouldn't it be enough to just check if the projects workspace root is a child of the rustc one in this case? That is change the equivalence check from #14348 to path child check. The rustc dir itself should be correct for all already since the recommended config has it set to a path (opposed to leaving it to "discover")

Veykril:
And yes, it does not surprise me that there are duplicated crates here. Iirc if you look into the library crates there are a bunch of type mismatches because all the library crates are duplicated and their dependencies are all over the place due to incorrect/missing deduplication there.

eggyal:
> wouldn't it be enough to just check if the projects workspace root is a child of the rustc one in this case?

In practice, probably yes.  But conceptually there's conceptually nothing that *requires* all workspaces that are opened with a rustc project to be children of the rustc source route.

eggyal:
Also, if by "just check" you mean simply to change `if rustc_workspace.workspace_root() == cargo.workspace_root()` to `if cargo.workspace_root().starts_with(rustc_workspace.workspace_root())` or similar then no that *alone* wouldn't work, as  `rustc_workspace` would still be the toolchain's `target-libdir`.

Veykril:
> Also, if by "just check" you mean simply to change `if rustc_workspace.workspace_root() == cargo.workspace_root()` to `if cargo.workspace_root().starts_with(rustc_workspace.workspace_root())` or similar then no that _alone_ wouldn't work, as `rustc_workspace` would still be the toolchain's `target-libdir`.

Huh? Doesn't the recommended vscode settings json in the rustc repo set the rustc-dir explicitly (via `rustc.source` config)? Then that should always point to the repo itself, not the toolchain. Now I'm confused here.

> In practice, probably yes. But conceptually there's nothing that requires all workspaces that are opened with a rustc project to be children of the rustc source route.

And this I am happy to just not support. It's bad enough that we are having to hack things into r-a just to support the rustc repo's workspace as is. That OTOH is an even more exotic setup that I don't see value in trying to make work. I'd rather just pursue `x.py` emitting a `rust-project.json` to fix all of this in the end.

Veykril:
I'll close this, as mentioned I think we (and with that I mean someone other than me) should instead investigate adding a command to x.py / having x.py generate a `rust-project.json` that contains a setup for the relevant profile of the developer. Then we can also finally remove some of the rustc workspace hacks in r-a.

I'm happy to help out with how to do that for anyone interested but I won't be working on that myself.

eggyal:
@Veykril I'm once again running up against this problem, but have some time to try and implement the fix that you were proposing.  However, having looked over the manual section on [Non-Cargo Based Projects](https://rust-analyzer.github.io/manual.html#non-cargo-based-projects) (which describes the `rust-project.json` format), I can't see how any of the existing settings can address this problem.

Were you proposing that the format be modified if need be?  If so, then I'm not entirely sure what we would need to represent in the file that isn't already available to/known by RA elsewhere?

Veykril:
There is a PR open that implements rust-project.json as a bootstrap step here https://github.com/rust-lang/rust/pull/120611. I don't know the current status of the PR though

Issue #15785, opened 2023-10-20
Feature request: Option to show inlay hints for implicit `Drop` calls


tjf801:
I feel like having an option to explicitly show any potentially implicit calls to `Drop()` would be a nice option to have.

For example, in a function like
```rust
fn foo(x: std::sync::Mutex<bool>) {
    let xx = x.lock().unwrap();
    *xx = !*xx;
    // add an inlay hint here showing that xx gets dropped
}
```
it might be nice to be able to hover over something to see the `Drop` implementation for `MutexGuard`, and make it explicit (at least in the editor) where certain values are being dropped. (Which could also be helpful for rust beginners.) This is probably something that should only be done when the type has a non-trivial `Drop` impl, however.

Veykril:
That is something I want for r-a to have as well at some point (don't remember if I made an issue about this). The main problem is that inlay hints can't insert new lines, so it is tricky to figure out a nice way to render this.

Young-Flash:
could it be added at the end of line instead of a new line?

Veykril:
Yes, that would be the simplest option. Note that we are missing the semantic analysis regarding drop scopes for this feature still, so it's not easy to implement right now.

HKalbasi:
MIR has some support for the drop scopes, not sure how easy is to map it back to a span in the original code.

HKalbasi:
#16000 Adds a basic implementation for this. Next steps are:
1. Restrict it to the `MutexGuard` and other types with significant drop impls. Currently, it is too noisy to be usable. We can have a config to list the desired types, have an attribute on those types, or something else.
2. Currently, it emits hint even if the type is moved out. Although it is technically correct (deciding what should be dropped is done at runtime, using the drop flag) it is pretty confusing and limits the usefulness of the hint. A post process on mir to remove drops that are definitely no-op can fix this problem, plus reducing the memory usage.
3. The drop implementation in mir lowering is not very accurate and can be improved.

PR #16000, opened 2023-12-01
Initial support for implicit drop inlay hint

HKalbasi:
cc #15785


lnicola:
image

rben01:
Would it be possible for these to go above the statement they precede instead of to the left? Shifting a line right to make room for the `drop` inlay hint doesn't look so nice. For me, it'd be preferable to have the `drop` visible when hovering over the `return` or the `}`.

Veykril:
Unlikely, this is a limitation by the language server protocol. And VSCode doesnt support ghost text between lines either (only annotations but those dont work that well for this purpose)

PR #16009, opened 2023-12-04
fix: bug in extract_function.rs

werifu:
There is a little bug in extract_function: It appends `use path::to::ControlFlow;` if the function created contains string "ControlFlow".

 A case below (also in the test named `does_not_import_control_flow` which will fail in the original code)

image
image

Now I have changed the condition determining whether adding import statement. Only when the new function body contains ControlFlow::Break or ControlFlow::Continue can the import statement be added.

Last related PR: https://github.com/rust-lang/rust-analyzer/pull/10309


Veykril (approved):

Veykril:
Thanks!
@bors r+

Issue #15033, opened 2023-06-12
Add a setting to rebuild proc-macros/build script on save


Veykril:
Basically an analogue to `checkOnSave` where we rerun proc-macro building/build-script running when proc-macro or build-script sources change and are saved.

Working on proc-macros can be annoying in r-a, given one has to manually trigger a rebuilt or reload, we might still want this setting to be disabled by default just in case. `cargo.buildScripts.rebuildOnSave` or similar would be a good name probably (we might wanna rename buildScripts and such into something that reflects build artifacts to get rid of the annoying proc-macro and build-script setting interdependency)

Veykril:
(reserving for euro-rust impl room)
Similar to https://github.com/rust-lang/rust-analyzer/blob/8a2331450a2fa269aa60818b224a2beaef7d7198/crates/rust-analyzer/src/config.rs#L71-L73 we'd add a `cargo_buildScripts_rebuildOnSave` entry here (must be inserted in sorted order).

Then when the setting is set, instead of toggling the `GlobalState` field here https://github.com/veykril/rust-analyzer/blob/8a2331450a2fa269aa60818b224a2beaef7d7198/crates/rust-analyzer/src/global_state.rs#L347-L353 we should request a build script operation via `self.fetch_build_data_queue.request_op`.

PR #16011, opened 2023-12-04
feat: add proc-macro rebuild on save option

ClSlaid:
Related: #15033

I need some advice on how to test it.

### DIFF
--- a/crates/rust-analyzer/src/global_state.rs
+++ b/crates/rust-analyzer/src/global_state.rs
@@ -351,10 +351,6 @@ impl GlobalState {
 
                     crates.iter().any(|&krate| crate_graph[krate].is_proc_macro)
                 });
-            if self.proc_macro_changed && self.config.script_rebuild_on_save() {
-                self.fetch_build_data_queue
-                    .request_op(format!("proc-macro or build script source changed"), ())
-            }
         }
 
         true
--- a/crates/rust-analyzer/src/global_state.rs
+++ b/crates/rust-analyzer/src/global_state.rs
@@ -351,6 +351,10 @@ impl GlobalState {
 
                     crates.iter().any(|&krate| crate_graph[krate].is_proc_macro)
                 });
+            if self.proc_macro_changed && self.config.script_rebuild_on_save() {
+                self.fetch_build_data_queue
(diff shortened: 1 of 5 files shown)


crates/rust-analyzer/src/global_state.rs
```diff
@@ -351,6 +351,10 @@ impl GlobalState {
 
                     crates.iter().any(|&krate| crate_graph[krate].is_proc_macro)
                 });
+            if self.proc_macro_changed && self.config.script_rebuild_on_save() {
+                self.fetch_build_data_queue
+                    .request_op(format!("proc-macro or build script source changed"), ())
+            }
```
Veykril, line 357: Hmm, so there is a problem with this for us here. Currently we set the `proc_macro_changed` flag whenever a change comes in for one of the sources, but those changes are not necessarily commited yet (saved), so with the current setup here we would ask for server rebuilds on every keystroke! We should instead move the op request to the `DidSaveTextDocument` handler (and then also reset the `proc_macro_changed` flag

Veykril (commented):
Regarding testing, we don't really have the infra for that yet. So manual testing that it works suffices. https://github.com/rust-lang/rust-analyzer/blob/0840038f02daec6ba3238f05d8caa037d28701a0/docs/dev/debugging.md

You will also need to run the `rust-analyzer` crate tests to regenerate the config,json

crates/rust-analyzer/src/handlers/notification.rs
```diff
@@ -130,6 +130,13 @@ pub(crate) fn handle_did_save_text_document(
     state: &mut GlobalState,
     params: DidSaveTextDocumentParams,
 ) -> anyhow::Result<()> {
+    if state.config.script_rebuild_on_save() && state.proc_macro_changed {
+        // reset the flag
+        state.proc_macro_changed = false;
+        // rebuild the proc macros
+        state.fetch_build_data_queue.request_op(format!("ScriptRebuildOnSave"), ());
```
Veykril, line 137:
```suggestion
        state.fetch_build_data_queue.request_op("ScriptRebuildOnSave".to_owned(), ());
```

PR #16022, opened 2023-12-05
internal: Don't explicitly warn against `semicolon_in_expressions_from_macros`

Xanewok:
This has been warn-by-default for two years now and has already been added to the future-incompat lints in 1.68.

See https://github.com/rust-lang/rust/issues/79813 for the tracking issue.


lnicola:
Thanks!

@bors r+

PR #16027, opened 2023-12-05
feat: Implicit format args support

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/11260
Fixes https://github.com/rust-lang/rust-analyzer/issues/11296

image
Too lazy to make a gif of this right now (would probably be good to show renaming)


lnicola:
https://github.com/rust-lang/rust-analyzer/assets/308347/20bcb7f6-cfa2-4550-a2ba-a524ab52cba3

nyurik:
so now we can re-enable clippy lint for this? :)  Assuming it will only work by default in the "all params can be included" context.  If its a mixed case, that would stay as pedantic.

Veykril:
Yes you should be able to re-enable it now.

PR #16031, opened 2023-12-05
feature: Add `Replace if let else with Option::map_or_else` assist

Emilgardis:
Implements the [`option_if_let_else`](https://rust-lang.github.io/rust-clippy/master/index.html#/option_if_let_else) clippy lint.

[](https://github.com/rust-lang/rust-analyzer/assets/1502855/a56624e2-d964-4daa-ac35-6f44fbd71958)

### DIFF
--- a/crates/ide-assists/src/handlers/replace_if_let_else_option_with_map_or_else.rs
+++ b/crates/ide-assists/src/handlers/replace_if_let_else_option_with_map_or_else.rs
@@ -1,0 +1,101 @@
+use ide_db::{syntax_helpers::node_ext::single_let, ty_filter::TryEnum};
+use syntax::{
+    ast::{self, make},
+    AstNode, TextRange,
+};
+
+use crate::{utils::does_pat_match_variant, AssistContext, AssistId, AssistKind, Assists};
+
+// Assist: replace_if_let_else_option_with_map_or_else
+//
+// Replaces a `if let Some(_) = ... { ... } else { ... }` expression with `Option::map_or_else`.
+//
+// ```
+// fn do_complicated_function() -> i32 { 1 }
+//
+// let optional = Some(1);
+//
+// let _ = $0if let Some(foo) = optional {
+//     foo
+// } else {
+//     let y = do_complicated_function();
+//     y*y
+// };
+// ```
+// ->
+// ```
+// fn do_complicated_function() -> i32 { 1 }
+//
+// let optional = Some(1);
+//
+// let _ = optional.map_or_else(|| {
+//     let y = do_complicated_function();
+//     y*y
+// }, |foo| {
+//     foo
+// });
+// ```
+pub(crate) fn replace_if_let_else_option_with_map_or_else(
+    acc: &mut Assists,
+    ctx: &AssistContext<'_>,
+) -> Option<()> {
+    let if_expr: ast::IfExpr = ctx.find_node_at_offset()?;
+    let available_range = TextRange::new(
+        if_expr.syntax().text_range().start(),
+        if_expr.then_branch()?.syntax().text_range().start(),
+    );
+    let cursor_in_range = available_range.contains_range(ctx.selection_trimmed());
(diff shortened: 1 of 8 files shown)


crates/ide-assists/src/handlers/replace_if_let_else_option_with_map_or_else.rs
```diff
@@ -0,0 +1,101 @@
...
+            if !does_pat_match_variant(&pat_, &TryEnum::Option.happy_pattern_wildcard()) {
+                return None;
+            }
+            let ast::Pat::TupleStructPat(tuple_pat) = pat_ else {
+                return None;
+            };
+
+            if scrutinee_to_be_expr.syntax().text() != expr.syntax().text() {
+                return None;
+            }
```
Emilgardis, line 76: I don't know if this is needed, stolen from https://github.com/rust-lang/rust-analyzer/blob/b2f6fd4f961fc7e4fbfdb80cae2e6065f8436f15/crates/ide-assists/src/handlers/replace_if_let_with_match.rs#L82-L88

Emilgardis (commented):
This works fine, however we don't cheack for control flow like `break`, `return` or `continue`. I think that's fine.

I want to also implement the reverse but I had troub le implementing it.

Emilgardis:
We also need to prevent

```rust
let _ = if $0let Some(ref mut field) = self.field2 {
    *field = "".to_owned();
    field == other
} else {
    true
}
```

or make it change to

```rust
let _ = self.field2.as_mut().map_or_else(|| {
    true
}, |field| {
    *field = "".to_owned();
    field == other
})
```

however I can't figure out how to get the bindings inside the pattern

crates/ide-assists/src/handlers/replace_if_let_else_option_with_map_or_else.rs
```diff
@@ -0,0 +1,101 @@
...
+            pat = tuple_pat.fields().next()?;
+        }
+        None => return None,
```
Veykril, line 79: Looks like we can just `?` on the result of `single_let`

crates/ide-assists/src/handlers/replace_if_let_else_option_with_map_or_else.rs
Veykril: I don't think this applies here

Veykril:
> this doesn't seem to handle
> 
> ```rust
> if let Some(ref field) = self.field {
>     field.eq_ignore_ascii_case(other)
> } else $0if let Some(field2) = self.field2 {
>     self.something()
> } else {
>     false
> }
> ```
> 
> well, it becomes
> 
> ```rust
> if let Some(ref field) = self.field {
>     field.eq_ignore_ascii_case(other)
> } else self.field2.map_or_else(|| {
>     false
> }, |field2| {
>     self.something()
> })
> ```
> 
> which is obviously wrong.

It's fine if we just don't offer the assist in that case for the time being. better than creating these syntax errors.

Veykril:
> We also need to prevent
> 
> ```rust
> let _ = if $0let Some(ref mut field) = self.field2 {
>     *field = "".to_owned();
>     field == other
> } else {
>     true
> }
> ```
> 
> or make it change to
> 
> ```rust
> let _ = self.field2.as_mut().map_or_else(|| {
>     true
> }, |field| {
>     *field = "".to_owned();
>     field == other
> })
> ```
> 
> however I can't figure out how to get the bindings inside the pattern

You can get the binding for that here https://github.com/rust-lang/rust-analyzer/pull/16031/files#diff-74f54460a4549ebe1f552b2513663b9eac476eef5399254c8895d6177094baa0R70-R72

`tuple_pat` is your pattern in that case, you just gotta inspect it there (`ref mut ident` in that case would be an `IdentPat` with `ref` and `mut` keyword being some)

Veykril:
Closing due to inactivity, feel free to come back to this if you find the time

PR #16035, opened 2023-12-06
fix: Fix diagnostics panicking when resolving to different files due to macros

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/14968


lnicola:
Should we close https://github.com/rust-lang/rust-analyzer/pull/15838?

PR #16041, opened 2023-12-07
Fix WideChar offsets calculation in `line-index`

### LINKED ISSUE #15981: `lib/line-index` 0.1.0 has a bug when handling wide characters
## Description

This is a bug of `lib/line_index` version `0.1.0`. It seems that rust-analyzer is using version 0.1.0-pre.1.

The comment here explains that `start` and `end` are offsets inside a single line:

https://github.com/rust-lang/rust-analyzer/blob/2ee17bc5f256e379c4df954e984b9bd0d79f1cbd/lib/line-index/src/lib.rs#L55-L60

However, during text analysis, the offsets stored in `WideChar` are byte offsets of the entire text.

https://github.com/rust-lang/rust-analyzer/blob/2ee17bc5f256e379c4df954e984b9bd0d79f1cbd/lib/line-index/src/lib.rs#L366-L372

Finally, when calculating `LineCol`, the `start` and `end` in `WideChar` are treated as offsets inside the line again.

https://github.com/rust-lang/rust-analyzer/blob/2ee17bc5f256e379c4df954e984b9bd0d79f1cbd/lib/line-index/src/lib.rs#L118-L132

## Example

The code below will panic.

```rust
fn main() {
    let text = "\n\n\n\n\n宽3456";
    assert_eq!(&text[5..8], "宽");
    assert_eq!(&text[11..12], "6");
    let line_index = LineIndex::new(text);
    let before_6 = TextSize::from(11);
    assert!(line_index.try_line_col(before_6).is_some()); // This will panic!
}
```

The line offset of the character `6` is `6..7`.
The global offset of the character `宽` is `5..8`, and it is mistakenly used as line offset.

roife:
Fix #15981. This PR addresses the issue with the WideChar's offset calculation, ensuring accurate line-specific positions during text analysis in the `lib/line-index` module.

## Changes Made

- Corrected the calculation for `WideChar` offsets, ensuring they reflect positions within respective lines.
- Added tests to verify the accuracy of `WideChar` offset calculations, and correct existing tests.

### DIFF
--- a/lib/line-index/src/tests.rs
+++ b/lib/line-index/src/tests.rs
@@ -1,4 +1,4 @@
-use crate::{LineIndex, TextSize, WideChar};
+use crate::{LineCol, LineIndex, TextSize, WideChar, WideEncoding, WideLineCol};
 
 macro_rules! test {
     (
@@ -102,7 +102,7 @@ test!(
     case: multi_byte_with_new_lines,
     text: "01\t345\n789abcΔf01234567\u{07}9\nbcΔf",
     lines: vec![7, 27],
-    multi_byte_chars: vec![(1, (13, 15)), (2, (29, 31))],
+    multi_byte_chars: vec![(1, (6, 8)), (2, (2, 4))],
 );
 
 test!(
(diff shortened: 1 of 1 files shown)


Veykril:
Thanks!
@bors r+

roife:
> 👀 Test was successful, but fast-forwarding failed: 422 Changes must be made through a pull request.

Oh, what error is this, and is there anything I need to do?

Veykril:
That can be ignored dw

PR #16043, opened 2023-12-07
Bump and unlock some dependencies

Veykril:


Veykril:
Gets rid of all but one duplicate dependency (bitflags)
@bors r+

PR #16061, opened 2023-12-08
fix: Fix fragment parser replacing matches with dummies on incomplete parses

Veykril:
Notably, this caused some completions in the `vec!` macro to no longer work. Fixes https://github.com/rust-lang/rust-analyzer/issues/15016


Veykril:
I'll wait with merging after next weeks release so that any issues that arise wont necessarily be blamed on the spanmap rewrite as this is a merge removal of an if statement that might be more loadbearing that I can tell.

PR #16066, opened 2023-12-09
fix: auto remove unnecessary braces after remove unused imports

Young-Flash:
before
before

after
after


Young-Flash:
could maintainers help review this when have time? 

cc @Veykril @lnicola @HKalbasi

lnicola (approved):
Looks good. I just wasn't sure that `ted` is the best API here, but we can probably merge it like this anyway. I'll just wait until tomorrow for a second review.

Veykril (approved):

PR #16068, opened 2023-12-09
fix: pick up new names when the name exists in 'introduce_named_generic'

roife:
Improve generation of names for generic parameters in `introduce_named_generics`.

### Changes

- Modified `for_generic_parameter` function in `suggest_name.rs` to handle conflicts with existing generic parameters and generate unique names accordingly.
- Update `introduce_named_generic` function and pass existing params to `for_generic_parameter`, enabling the detection and handling of name collisions.


roife:
Here's an example from unit test:

Say we perform `introduce_named_generic` on 

```rust
fn foo<B>(bar: $0impl Bar) {}
```

Before this PR:

```rust
fn foo<B, B: Bar>(bar: B) {} // conflicts!
```

After this PR:

```rust
fn foo<B, B0: Bar>(bar: B0) {}
```

Veykril:
Thanks!
@bors r+

PR #16073, opened 2023-12-09
Replace `doc_comments_and_attrs` with `collect_attrs`

### LINKED ISSUE #16063: Panic on inner attributes in presence of unresolved crate
This code:
```Rust
#[macro_use]
extern crate doesnotexist;

mod _test_inner {
    #![empty_attr]
}
```
Panics:
```
thread 'Worker' panicked at crates/hir/src/lib.rs:1010:21:
cannot find attribute #0
stack backtrace:
   0: rust_begin_unwind
             at /rustc/0f44eb32f1123ac93ab404d74c295263ce468343/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/0f44eb32f1123ac93ab404d74c295263ce468343/library/core/src/panicking.rs:72:14
   2: hir::precise_macro_call_location
   3: hir::emit_def_diagnostic_
   4: hir::Module::diagnostics
   5: ide_diagnostics::diagnostics
   6: std::panicking::try
   7: ide::Analysis::with_db
   8: ide::Analysis::assists_with_fixes
   9: rust_analyzer::handlers::request::handle_code_action
  10: std::panicking::try
  11: core::ops::function::FnOnce::call_once{{vtable.shim}}
```
I investigated a bit, and it seems the problem is that `AttrDocCommentIter` does not iterate over inner attributes, since inner attributes are not direct children of the item.

What is the best way to fix this? Making `AttrDocCommentIter` iterate over descendants instead of children doesn't look great performance-wise, iterating over the node that might contain inner attributes is the solution I think but makes the code a bit ugly. Also, how inner attributes work at all? It seems the responsible code for detecting attributes (both inner and outer) is the `AttrDocCommentIter` which is broken. But inner attributes work, for example `#![doc]` will appear in the hover, so maybe we can drop `AttrDocCommentIter` in favor of reusing that code.

HKalbasi:
fix #16063 

I looked at the other usages of `doc_comments_and_attrs` and it seems all of them are prone to ignoring inner attributes. @Veykril should I replace all of those with `collect_attrs` and remove `doc_comments_and_attrs` (or even `HasDocComments`) entirely?

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -59,7 +59,7 @@ use hir_def::{
     Lookup, MacroExpander, MacroId, ModuleId, StaticId, StructId, TraitAliasId, TraitId,
     TypeAliasId, TypeOrConstParamId, TypeParamId, UnionId,
 };
-use hir_expand::{name::name, MacroCallKind};
+use hir_expand::{attrs::collect_attrs, name::name, MacroCallKind};
 use hir_ty::{
     all_super_traits, autoderef, check_orphan_rules,
     consteval::{try_const_usize, unknown_const_as_generic, ConstEvalError, ConstExt},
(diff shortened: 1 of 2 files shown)


HKalbasi:
I'm going to merge this for now. Will replace remaining usages of `doc_comments_and_attrs` in another PR if desired.

@bors r+

Veykril:
We should replace the other ones as well, though the one in `semantics` needs special casing depending on whether we are working with inner attributes or not.

Also the issue here is actually just 
```rs
   mod _test_inner {
        #![empty_attr]
      //^^^^^^^^^^^^^^ error: unresolved macro `empty_attr`
    }
```
the extern crate is irrelevant

PR #16085, opened 2023-12-11
fix: try obligation of `IndexMut` when infer

### LINKED ISSUE #15842: Type inference unknown when using references
**rust-analyzer version**: c1c9e10f72ffd2e829d20ff1439ff49c2e121731

**rustc version**: rustc 1.75.0-nightly (42b1224e9 2023-10-15)

**relevant settings**: N/A

Sample code:

```rs
use defaultmap::DefaultHashMap;

fn main() {
    let mut map = DefaultHashMap::new(0);
    map["a"] = 1;
}
```

`map` gets the type inference `let mut map: DefaultHashMap<{unknown}, i32>`

Austaras:
Closes #15842.

This issue arises because `K` is ambiguous if only inferred from `Index` trait, but is unique if inferred from `IndexMut`, but r-a doesn't use this info.

### DIFF
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -744,7 +744,7 @@ impl InferenceContext<'_> {
                     (RangeOp::Inclusive, _, None) => self.err_ty(),
                 }
             }
-            Expr::Index { base, index } => {
+            Expr::Index { base, index, is_assignee_expr } => {
                 let base_ty = self.infer_expr_inner(*base, &Expectation::none());
                 let index_ty = self.infer_expr(*index, &Expectation::none());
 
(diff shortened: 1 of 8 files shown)


crates/hir-ty/src/infer/expr.rs
```diff
@@ -772,11 +772,24 @@ impl InferenceContext<'_> {
...
-                        &[index_ty.cast(Interner)],
-                    )
+                    let assoc = self.resolve_ops_index_output();
+                    let res = self.resolve_associated_type_with_params(
+                        self_ty.clone(),
+                        assoc,
+                        &[index_ty.clone().cast(Interner)],
+                    );
+
+                    if *is_assignee_expr {
```
Austaras, line 782: It does no harm to obligate every indexing for `IndexMut` because r-a currently only save obligation that can be solved and it won't need an extra field in HIR, but I believe it's better to be more precise.

Veykril (approved):

PR #16088, opened 2023-12-11
feat: Implement a rust-analyzer span backed proc-macro server mode

Veykril:
This implements the basic span APIs. Basically anything that doesn't require talking back to the client for information access.

This also commits our syntax fixup marker to use an `ErasedAstFileId` of `!0-1` aka `0xffff_fffe`, instead of using a dummy FileId as a marker, as we need that for the `SourceFile` API to be implementable. The reason as to why the server needs to know about this at all is to prevent it from creating invalid fixup spans which could make r-a panic.


crates/hir-expand/src/builtin_fn_macro.rs
```diff
@@ -770,3 +771,15 @@ fn option_env_expand(
 
     ExpandResult::ok(expanded)
 }
+
+fn quote_expand(
```
Veykril, line 775: This is a very interesting macro, it's a builtin macro that is actually backed by a proc-macro that rustc loads from the `proc_macro` library https://doc.rust-lang.org/proc_macro/fn.quote.html! Unsure how we can hook that up for now, but that might require this part to be nightly gated as well

PR #16089, opened 2023-12-11
Query for nearest parent block around the hint to resolve

SomeoneToIgnore:
This way, parameter hints will be found for resolution and https://github.com/rust-lang/rust-analyzer/pull/15522#issuecomment-1835896335 will be fixed

Hopefully that also helps with whatever else (lifetimes', etc.) hints in 
https://github.com/rust-lang/rust-analyzer/issues/13962

> hints are resolved by querying for textDocument/inlayHint with hint's position turned into a [position-1, position+1] range (instead of the original, much wider document range).
>
> This might lead to issues in the future, with e.g. lifetime hints (currently there's nothing to resolve for them and it's fine) that belong to a certain position, but need to have textDocument/inlayHint query for much bigger range than their position+/-1

### DIFF
--- a/crates/ide/src/inlay_hints.rs
+++ b/crates/ide/src/inlay_hints.rs
@@ -424,7 +424,7 @@ fn ty_to_text_edit(
 
 pub enum RangeLimit {
     Fixed(TextRange),
-    NearestParentBlock(TextSize),
+    NearestParent(TextSize),
 }
 
 // Feature: Inlay Hints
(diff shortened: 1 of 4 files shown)


crates/ide/src/inlay_hints.rs
```diff
@@ -458,13 +463,23 @@ pub(crate) fn inlay_hints(
...
                     .for_each(hints),
             },
+            Some(RangeLimit::NearestParentBlock(position)) => {
+                match file
+                    .token_at_offset(position)
+                    .left_biased()
+                    .and_then(|token| token.parent_ancestors().find_map(ast::BlockExpr::cast))
+                {
+                    Some(parent_block) => parent_block.syntax().descendants().for_each(hints),
+                    None => return acc,
```
Veykril, line 480: This won't fix lifetime hints (as they work on `Fn` item nodes which we'll miss here, unless its a local item), maybe a better simple check would be parent block or parent item?

PR #16092, opened 2023-12-11
feat: Introduce term search to rust-analyzer

kilpkonn:
# Introduce term search to `rust-analyzer`
_I've marked this as draft as there might be some shortcomings, please point them out so I can fix them. Otherwise I think it is kind of ready as I think I'll rather introduce extra functionality in follow up PRs._

Term search (or I guess expression search for rust) is a technique to generate code by basically making the types match.
Consider the following program
```rust
fn wrap(arg: i32) -> Option<i32> {
    todo!();
}
```
From the types of values in scope and constructors of `Option`, we can produce the expected result of wrapping the argument in `Option`

Dependently typed languages such as `Idris2` and `Agda` have similar tools to help with proofs, but this can be also used in everyday development as a "auto-complete".

# Demo videos

https://github.com/rust-lang/rust-analyzer/assets/19900308/7b68a1b7-7dba-4e31-9221-6c7485e77d88

https://github.com/rust-lang/rust-analyzer/assets/19900308/0fae530a-aabb-4b28-af71-e19f8d3d64b2

# What does it currently do
- It works well with locals, free functions, type constructors and non-static impl methods that take items by value.
- Works with functions/methods that take shared references, but not with unique references (very conservative).
- Can handle projections to struct fields (eg. `foo.bar.baz`) but this might me more conservative than it has to be to avoid conflicting with borrow checker
- Should create only valid programs (no type / borrow checking errors). Tested with `rust-analyzer analysis-stats /path/to/ripgrep/Cargo.toml --run-term-search --validate-term-search` (basically running `cargo check` on all of the generated programs and only error seems to be due to type inference which is more of issue of testing method.

# Performace / fitness
```txt
ripgrep (latest)
Tail Expr syntactic hits: 130/1692 (7%)
Tail Exprs found: 523/1692 (30%)
Term search avg time: 9ms
Term search:         15.64s, 97ginstr, 8mb

rust-analyzer (on this branch)
Tail Expr syntactic hits: 804/13860 (5%)
Tail Exprs found: 6757/13860 (48%)
Term search avg time: 78ms
Term search:         1088.23s, 6765ginstr, 98mb
```
Highly generic code seems to blow up the search space so currently the amount of generics allowed is functions/methods is limited down to 0 (1 didn't give much improvement and 2 is already like 0.5+s search time)

# Plans for the future (not in this PR)
- ``~~Add impl methods that do not take `self` type (should be quite straight forward)~~ Done
- Be smarter (aka less restrictive) about borrow checking - this seems quite hard but since the current approach is rather naive I think some easy improvement is available.
- ``~~See if it works as a autocomplete while typing~~ Done

_Feel free to ask questions / point of shortcoming either here or on Zulip, I'll be happy to address them. I'm doing this as part of my MSc thesis so I'll be working on it till summer anyway :smile:_


Young-Flash:
interesting, it would be nicer if there give a bit more docs for `crates/hir/src/term_search/*.rs` to help understand the idea and the implementation approach under the hood.

kilpkonn:
> interesting, it would be nicer if there give a bit more docs for `crates/hir/src/term_search/*.rs` to help understand the idea and the implementation approach under the hood.

I've added some more documentation there. If you find some caps / stuff that doesn't make sense etc let me know as it definitely can be improved further :smile:

Veykril:
Just a headsup but it will take some time (I'm potentially speaking months) until this gets reviewed as there is a massive backlog + me having little time atm + holidays coming up.

kilpkonn:
> Just a headsup but it will take some time (I'm potentially speaking months) until this gets reviewed as there is a massive backlog + me having little time atm + holidays coming up.

That's unfortunate :( But thanks for the heads-up, appreciate it!

kilpkonn:
I added also autocomplete part to this PR.
Demo can be seen here:
https://github.com/rust-lang/rust-analyzer/assets/19900308/8d8d3c4d-9f3f-44ed-bc96-3bc3b682bbae

Snippet is a little misleading `auto completion kind for it, but couldn't find any better :p

Veykril:
CI fails because `Generator` stuff was renamed to `Coroutine` (i'll try giving this PR a review this or next week)

Veykril:
Sorry I haven't gotten around to this yet, been somewhat sick last week.

kilpkonn:
> Sorry I haven't gotten around to this yet, been somewhat sick last week.

No worries, hope your well now :smile: 
I've done the rebase so everything should be ready for review again

crates/hir-ty/src/infer/unify.rs
```diff
@@ -82,6 +82,33 @@ pub fn could_unify(
     unify(db, env, tys).is_some()
 }
 
+pub fn could_unify_deeply(
```
Veykril, line 85: Comment for this would be nice (difference between this and `could_unify`)

crates/hir-ty/src/mir/borrowck.rs
```diff
@@ -185,6 +204,149 @@ fn moved_out_of_ref(db: &dyn HirDatabase, body: &MirBody) -> Vec<MovedOutOfRef>
...
+                        let infer = db.infer(def);
+                        let (captures, _) = infer.closure_info(&c);
+                        let parent_subst = ClosureSubst(subst).parent_subst();
+                        captures
+                            .get(f)
+                            .expect("broken closure field")
+                            .ty
+                            .clone()
+                            .substitute(Interner, parent_subst)
+                    },
```
Veykril, line 227: This seems to be re-used for all 4 borrowck "analysis" things, might be nice to extract this into a function if possible: `fn make_fetch_closure_field(db: &dyn HirDatabase) -> impl FnOnce(ClosureId, &Substitution, usize) -> Ty`

crates/hir/src/lib.rs
```diff
@@ -3472,13 +3682,8 @@ impl Impl {
                     .filter(filter),
             )
         });
-        for id in def_crates
-            .iter()
-            .flat_map(|&id| Crate { id }.transitive_reverse_dependencies(db))
-            .map(|Crate { id }| id)
-            .chain(def_crates.iter().copied())
-            .unique()
-        {
```
Veykril, line 3481: This change seems odd. We don't need to traverse all crates in the crate graph, only the ones that can observe the type no?

crates/hir/src/term_search/expr.rs
```diff
@@ -0,0 +1,436 @@
...
+
+/// Helper function to get path to `ModuleDef` as string
+fn mod_item_path_str(
+    sema_scope: &SemanticsScope<'_>,
+    def: &ModuleDef,
+    prefer_no_std: bool,
+    prefer_prelude: bool,
+) -> String {
+    let path = mod_item_path(sema_scope, def, prefer_no_std, prefer_prelude);
+    path.map(|it| it.display(sema_scope.db.upcast()).to_string()).unwrap()
```
Veykril, line 53: bad unwrap

crates/hir/src/term_search/expr.rs
```diff
@@ -0,0 +1,436 @@
...
+                mod_item_path(sema_scope, &ModuleDef::Adt(adt), prefer_no_std, prefer_prelude)
+                    .unwrap();
+            path.pop_segment();
+            let path = path.display(db.upcast()).to_string();
+            match path.is_empty() {
+                true => ty_name,
+                false => format!("{path}::{ty_name}"),
+            }
+        }
+        None => ty.display(db).to_string(),
```
Veykril, line 78: Should probably be `display_source_code`, likewise for most if not all other usages of this here

crates/hir/src/term_search/mod.rs
```diff
@@ -0,0 +1,298 @@
+//! Term search
```
Veykril, line 1: Nit but we don't use `mod.rs` style module hierarchies

crates/hir/src/term_search/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    /// Types queried but not found
+    fn take_types_wishlist(&mut self) -> FxHashSet<Type> {
+        std::mem::take(&mut self.types_wishlist)
+    }
+}
+
+/// Context for the `term_search` function
+#[derive(Debug)]
+pub struct TermSearchCtx<'a, DB: HirDatabase> {
```
Veykril, line 216: This should probably just use `&dyn HirDatabase` everywhere instead of a generic parameter

crates/ide-assists/src/handlers/term_search.rs
```diff
@@ -0,0 +1,230 @@
...
+
+pub(crate) fn term_search(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
+    let unexpanded = ctx.find_node_at_offset::<ast::MacroCall>()?;
+    let syntax = unexpanded.syntax();
+    let goal_range = syntax.text_range();
+
+    let excl = unexpanded.excl_token()?;
+    let macro_name_token = excl.prev_token()?;
+    let name = macro_name_token.text();
+    if name != "todo" {
```
Veykril, line 18: This should fetch the actual todo macro and check whether the two are the same via `FamousDefs` (you can just look at other usages of FamousDefs that should show you what to do (example https://github.com/veykril/rust-analyzer/blob/5860763f1808ef7ba2628bf46c08e77c96804925/crates/ide/src/inlay_hints.rs#L572)

Veykril: Also might be nice to make this work on `todo` and `unimplemented` alike

crates/ide-completion/src/render.rs
```diff
@@ -272,6 +272,79 @@ pub(crate) fn render_resolution_with_import_pat(
...
+
+    let source_range = match ctx.original_token.parent() {
+        Some(node) => match node.ancestors().find_map(|n| ast::Path::cast(n)) {
+            Some(path) => path.syntax().text_range(),
+            None => node.text_range(),
+        },
+        None => ctx.source_range(),
+    };
+
+    let mut item = CompletionItem::new(CompletionItemKind::Snippet, source_range, label.clone());
```
Veykril, line 313: Let's give that a new kind here (that maps to snippet in the LSP layer), that makes it more clear in tests that something is termsearched

crates/ide-diagnostics/src/tests.rs
```diff
@@ -68,6 +68,91 @@ fn check_nth_fix(nth: usize, ra_fixture_before: &str, ra_fixture_after: &str) {
...
 }
 
+pub(crate) fn check_fixes_unordered(ra_fixture_before: &str, ra_fixtures_after: Vec<&str>) {
+    for ra_fixture_after in ra_fixtures_after.iter() {
+        check_has_fix(ra_fixture_before, ra_fixture_after)
+    }
+}
+
+#[track_caller]
+pub(crate) fn check_has_fix(ra_fixture_before: &str, ra_fixture_after: &str) {
```
Veykril, line 78: whats different between this and `check_fix`?

crates/rust-analyzer/src/cli/flags.rs
```diff
@@ -93,6 +93,10 @@ xflags::xflags! {
             /// and annotations. This is useful for benchmarking the memory usage on a project that has
             /// been worked on for a bit in a longer running session.
             optional --run-all-ide-things
+            /// Run term search
```
Veykril, line 96: what does this run it on exactly? Like what expressions etc

crates/rust-analyzer/src/config.rs
```diff
@@ -282,6 +282,8 @@ config_data! {
                 "scope": "expr"
             }
         }"#,
+        /// Whether to enable term search based snippets like `Some(foo.bar().baz())`.
+        completion_term_search_enable: bool = "true",
```
Veykril, line 286: Let's put this to false by default and add a warning that this may slow down completions. Completions are already rather slow atm so I don't feel like it's a good idea to enable this by default right now. It also potentially clutters the completions with things.

Veykril: Also, the key should probably be `completion_termSearch_enable`

Veykril:
@flodiebold could you check out the `hir-ty` changes/additions if you have the time? (if not that's fine)
@HKalbasi could you check out the mir borrowck changes/additions if you have the time? (if not that's fine)

Veykril:
I've skipped over most of the term-search modules for now, might give those a look later (for whether they do something that could be done in a more r-a idiomatic way, not logic since I don't know anything about that here :)

crates/hir-def/src/attr.rs
```diff
@@ -382,27 +382,36 @@ impl AttrsWithOwner {
...
-                        db.span_map(src.file_id).as_ref(),
-                    )
+                    match src.value.get(it.local_id()) {
+                        Some(val) => RawAttrs::from_attrs_owner(
+                            db.upcast(),
+                            src.with_value(val),
+                            db.span_map(src.file_id).as_ref(),
+                        ),
+                        None => RawAttrs::EMPTY,
+                    }
```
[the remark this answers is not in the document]
kilpkonn: This is not a rebasing mistake. I managed to get panics here when running term search on ripgrep project.
The panic came from `src.value[it.local_id()]` as the array access was out of bounds

crates/ide-diagnostics/src/tests.rs
kilpkonn: `check_fix` expects the desired fix to be the first suggested fix (or there to be only 1 fix in most cases).
This doesn't care about the ordering, the fix just has to be there somewhere.

I added this as sepparate method as I wasn't sure if `check_fix` should or shouldn't care where in the list of fixes the desired fix is

crates/rust-analyzer/src/cli/flags.rs
kilpkonn: Currently on all the tail expressions (of functions, block, if statements etc.)

crates/hir-def/src/attr.rs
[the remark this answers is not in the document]
kilpkonn: If I remember correctly then it had something to do with functions such as `fn foo(x: impl Bar)` cuz of the `impl Trait` syntax. I can try to debug that later and create an issue for it etc.

crates/hir-ty/src/infer/unify.rs
```diff
@@ -433,6 +460,18 @@ impl<'a> InferenceTable<'a> {
         true
     }
 
+    /// Unify two relatable values (e.g. `Ty`) and register new trait goals that arise from that.
```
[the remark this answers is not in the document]
flodiebold: Maybe "check whether trait goals which arise from that could be fulfilled".

crates/hir/src/lib.rs
```diff
@@ -1087,6 +1088,27 @@ impl Field {
         Type::new(db, var_id, ty)
     }
 
+    pub fn ty_with_args(&self, db: &dyn HirDatabase, generics: impl Iterator<Item = Type>) -> Type {
```
flodiebold, line 1091: I'm not sure I like this as an API, the way it deals with const args is really weird :sweat_smile: and having it duplicated on every thing that has a type is also not great.

crates/hir/src/lib.rs
```diff
@@ -2951,6 +3123,40 @@ impl GenericDef {
             })
             .collect()
     }
+
+    pub fn type_params(self, db: &dyn HirDatabase) -> Vec<TypeParam> {
```
flodiebold, line 3127: these APIs are also not so great since they lose the order of type/const params. IMO it would be better to have the caller use `type_or_const_params` and filter themselves if they're really only interested in one

kilpkonn: Good point, I'll do that

crates/hir/src/lib.rs
kilpkonn: That is true, however I couldn't really come up with something better right away, especially for stuff that get generics also from impl block.
```rust
impl<T> for Foo<T> {
    fn foo<U>(...) -> Bar<T, U> { ... }
}
```
I guess an option is to turn that impl block into impl block for concrete type, then do the same for function and then get the return value of the function normally. However that kind of duplicates all the definitions themselves so I don't really want to do it

crates/rust-analyzer/src/cli/flags.rs
Veykril: I see, could you add that to the description? It's a bit vague what it tries otherwise

crates/hir-ty/src/infer/unify.rs
```diff
@@ -661,6 +700,38 @@ impl<'a> InferenceTable<'a> {
         }
     }
 
+    fn try_fulfill_obligation(
```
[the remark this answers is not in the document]
kilpkonn: I guess making `try_resolve_obligation` resolve the `solution` instead of bool kind of fixes the duplication as well as confusion. I'm not too sure where the `FIXME` comments of last 2 match arms should go, to the callee or stay here

crates/hir/src/lib.rs
kilpkonn: Hmm, works if I chain the current crate to it. I guess I'll do that then

crates/hir/src/term_search/expr.rs
kilpkonn: Fixed it. Tho most of the `display` calls are to `Name` or `ModPath` that do not implement `HirDisplay`.

crates/hir/src/term_search/mod.rs
kilpkonn: Need it for `Semantics<'a, DB>` or am I missing something?

crates/ide-assists/src/handlers/term_search.rs
kilpkonn: I'm having hard time locating the `todo` macro from standard library.
`core:macros:todo` seems to be the path but in reality it only finds macros that have `macro` keyword which means no `todo` macro.
Something like this is found: https://doc.rust-lang.org/stable/src/core/macros/mod.rs.html#144
But not this: https://doc.rust-lang.org/stable/src/core/macros/mod.rs.html#871-878

HKalbasi (approved):
The MIR changes look good to me. It would be nice to have `PartialMove` as a diagnostic, maybe in a separate PR.

crates/ide-assists/src/handlers/term_search.rs
Veykril: Oh right, those behave different in regards to their path. For the macro rules ones you'll just need the path from the crate root, so `core::todo` should resolve I believe (as the `macro_export` will re-export them from the crate root of core)

crates/hir/src/term_search/mod.rs
Veykril: Oh right, this is due to Semantics being stored here. Forget what I said then.

crates/ide-assists/src/handlers/term_search.rs
kilpkonn: Ty, works like charm!
Any ideas how to use it in tests tho as there is no std lib?

crates/hir-def/src/attr.rs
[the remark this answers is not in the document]
kilpkonn: Created issue here: https://github.com/rust-lang/rust-analyzer/issues/16478

kilpkonn:
I think I've managed to address all issues in one way or another so I guess it's ready for another round of review.

crates/hir/src/lib.rs
```diff
@@ -3258,10 +3454,13 @@ impl Impl {
...
             )
         });
+
         for id in def_crates
             .iter()
             .flat_map(|&id| Crate { id }.transitive_reverse_dependencies(db))
             .map(|Crate { id }| id)
+            .chain(Crate { id: env.krate }.dependencies(db).into_iter().map(|it| it.krate.id))
+            .chain(std::iter::once(env.krate))
             .chain(def_crates.iter().copied())
```
Veykril, line 3464: This line is actually not needed, as its included through transitive_reverse_dependencies
```suggestion
```

crates/hir/src/lib.rs
```diff
@@ -3258,10 +3454,13 @@ impl Impl {
                     .filter(filter),
             )
         });
+
         for id in def_crates
             .iter()
             .flat_map(|&id| Crate { id }.transitive_reverse_dependencies(db))
             .map(|Crate { id }| id)
+            .chain(Crate { id: env.krate }.dependencies(db).into_iter().map(|it| it.krate.id))
+            .chain(std::iter::once(env.krate))
```
Veykril, line 3463: Can you point me to a test that fails without these lines? I'm still a bit confused as to why this is necessary

crates/ide-completion/src/completions/expr.rs
```diff
@@ -328,3 +328,59 @@ pub(crate) fn complete_expr_path(
...
+        return;
+    }
+
+    if !ctx.qualifier_ctx.none() {
+        return;
+    }
+
+    if let Some(ty) = &ctx.expected_type {
+        // Ignore unit types as they are not very interesting
+        if ty.is_unit() {
```
Veykril, line 345: This should probably also ignored error types / unknown types?

crates/ide-completion/src/completions/postfix.rs
```diff
@@ -395,17 +395,21 @@ fn main() {
...
+                sn match    match expr {}
+                sn not      !expr
+                sn ref      &expr
+                sn refm     &mut expr
+                sn unsafe   unsafe {}
+                sn while    while expr {}
+                ex bar
+                ex false
+                ex foo(...)
+                ex true
```
Veykril, line 412: These are not field accesses / method calls so they shouldn't be shown here

crates/ide-completion/src/completions/expr.rs
kilpkonn: Yes, I think functionality vise it already does it but it is faster and more readable to add check so I'll do that!

crates/hir/src/lib.rs
kilpkonn: `handlers::typed_hole::tests::local_item_use_trait` fails without line 3463 as it doesn't see impl right next to function somehow. Starting to wonder if 3462 is necessary as I cant repr issue with types from std

crates/hir/src/lib.rs
Veykril: And I think the `unique` call afterwards won't be necessary either anymore then. I think `transitive_reverse_dependencies` already upholds that property?

crates/ide-completion/src/completions/postfix.rs
kilpkonn: I'll remove completing dot access for now as getting it to consistently complete dot access needs some more work due to universal function call syntax it uses in some cases.

crates/hir/src/lib.rs
kilpkonn: My bad, I'll change the test to not use primitive as it was not the point of the test anyway

crates/hir/src/lib.rs
kilpkonn: According to tests `unique` is not needed so I guess I'll remove that as well

Veykril:
Generally lgtm though the question about the `ty_with_args` API remains. I don't have a nice idea for that either right now :thinking:

kilpkonn:
> Generally lgtm though the question about the `ty_with_args` API remains. I don't have a nice idea for that either right now 🤔

I guess one option would be to limit the visibility to `pub(crate)` for now although that doesn't solve the problem

flodiebold:
I would say it's ok to postpone the question until later, and just add some comments about it.

Veykril:
Then let's merge this I guess :) Excited to try this out!
@bors r+

PR #16094, opened 2023-12-11
Replace `doc_comments_and_attrs` with `collect_attrs`, 2nd round

HKalbasi:
I didn't understand why the one in the semantics is different. Please tell me if it needs special action.

r? @Veykril

### DIFF
--- a/crates/hir-def/src/child_by_source.rs
+++ b/crates/hir-def/src/child_by_source.rs
@@ -5,8 +5,7 @@
 //! node for a *child*, and get its hir.
 
 use either::Either;
-use hir_expand::HirFileId;
-use syntax::ast::HasDocComments;
+use hir_expand::{attrs::collect_attrs, HirFileId};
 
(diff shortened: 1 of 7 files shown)


Veykril:
In semantics in the `descend_into_macros_impl` the attribute's text range is used to strip away all entries from the start of the attribute "list" up the the invoking attribute. But in
```
mod foo {
    #![inner]
}
```
(even though we dont support this yet), we don't wanna strip away stuff in the `mod foo {` range, that is here if the id corresponds to an inner attribute we got strip all text ranges of the outer ones, and then all of the inner ones up to the invoking attribute so that the inbetween is ignored.

Veykril:
But I'm fine with merging this as is, as we do not support user attributes on anything that allows inner attributes currently.

HKalbasi:
I added your comment as a fixme comment for now.

@bors r+

PR #16096, opened 2023-12-12
fix: update VSCode rust-panic problem matcher

LuisFerLCC:
Corrected the `rust-panic` task problem matcher for the VSCode Extension to match the new panic message pattern.

From:
```
thread 'main' panicked at 'PANIC_MESSAGE', src/main.rs:L:C
```
To:
```
thread 'main' panicked at src/main.rs:L:C:
PANIC_MESSAGE
```

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -1689,17 +1689,17 @@
             {
                 "name": "rust-panic",
                 "patterns": [
-					{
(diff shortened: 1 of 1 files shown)


lnicola (approved):
LGTM, but I can't make it work on my system for some reason.

Anyway, a couple of users complained about this problem matcher and we might even drop it at some point.

PR #16099, opened 2023-12-12
internal: Improve import asset perf a bit

Veykril:
And bump the query limit from 40 to 100

### DIFF
--- a/crates/ide-completion/src/tests/flyimport.rs
+++ b/crates/ide-completion/src/tests/flyimport.rs
@@ -597,8 +597,8 @@ fn main() {
 }
 "#,
         expect![[r#"
-            ct SPECIAL_CONST (use dep::test_mod::TestTrait) u8 DEPRECATED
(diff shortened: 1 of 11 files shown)


crates/ide-completion/src/tests/flyimport.rs
```diff
@@ -717,27 +717,27 @@ fn main() {
     check(
         fixture,
         expect![[r#"
-            st Item (use foo::bar::baz::Item) Item
+            st Item (use foo::bar) Item
```
Veykril, line 720: Also changes what we render in the completions, previously we rendered the full path of the item we are going to import, yet that is not always the path we actually import (if we have a path prefix i.e. `foo::Bar$0`). Imo, it makes more sense to render what path we are actually importing, not what the item of interest is that we import.

Veykril:
Also bumps the query limit from 40 to 100, I think this should be fine perf wise now as we do a *lot* less string cloning.
@bors r+

PR #16100, opened 2023-12-12
feat: add assoc func quickfix for `unresolved_method` diagnostic

Young-Flash:
demo

close https://github.com/rust-lang/rust-analyzer/issues/13247

EDIT: I think add a demo gif would be helpful to @lnicola when he make a release change log :)

### DIFF
--- a/crates/ide-diagnostics/src/handlers/unresolved_method.rs
+++ b/crates/ide-diagnostics/src/handlers/unresolved_method.rs
@@ -1,11 +1,14 @@
-use hir::{db::ExpandDatabase, HirDisplay};
+use hir::{db::ExpandDatabase, AssocItem, HirDisplay};
 use ide_db::{
     assists::{Assist, AssistId, AssistKind},
     base_db::FileRange,
     label::Label,
     source_change::SourceChange,
 };
-use syntax::{ast, AstNode, TextRange};
+use syntax::{
+    ast::{self, make, HasArgList},
+    AstNode, SmolStr, TextRange,
+};
 use text_edit::TextEdit;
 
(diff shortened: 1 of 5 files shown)


crates/ide-diagnostics/src/handlers/unresolved_method.rs
```diff
@@ -79,13 +100,178 @@ fn field_fix(
...
+fn assoc_func_fix(ctx: &DiagnosticsContext<'_>, d: &hir::UnresolvedMethodCall) -> Option<Assist> {
+    if let Some(assoc_item_id) = d.assoc_func_with_same_name {
+        let db = ctx.sema.db;
+
+        let expr_ptr = &d.expr;
+        let root = db.parse_or_expand(expr_ptr.file_id);
+        let expr: ast::Expr = expr_ptr.value.to_node(&root);
+
+        let call = ast::MethodCallExpr::cast(expr.syntax().clone())?;
+        let range = call.syntax().text_range();
```
Veykril, line 115:
```suggestion
        let range = InFile::new(expr_ptr.file_id, call.syntax().text_range()).original_node_file_range_rooted(db)?;
```
The range we use for editing needs to accommodate for macros

Veykril: That doesn't matter. In either case, this range is used together with a real file id later (when we construct an assist), so we need to make sure that the text range corresponds to a real file by upmapping it out of any possible macro expansions. A range originating from a macro expansion cannot be used for edits nor diagnostic displaying as it points into a macro expansion, not a file.

PR #16102, opened 2023-12-12
internal: Publish `line-index`

lnicola:


lib/la-arena/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "la-arena"
-version = "0.3.1"
+version = "0.3.2"
```
lnicola, line 3: Not sure if this is worth it, since the only change was removing a `#[warn]`.

PR #16106, opened 2023-12-12
fix: Fix syntax bridge assigning invalid span to lifetime tokens

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/16097
Fixes https://github.com/rust-lang/rust-analyzer/issues/16103


Veykril:
@lnicola I think we might wanna consider doing a point release, this will cause panics for a bunch of people otherwise 😓

honestly speaking I don't know how we do stable releases at this point 😄 but it should be fine to not do an extra changelog (just add whatever to next mondays), don't wanna give you extra work

PR #16114, opened 2023-12-13
fix: self type replacement in inline-function

### LINKED ISSUE #16091: 'inline function' assist does not take `as Self` casts into account
**rust-analyzer version**: 0.4.1765-standalone

**rustc version**: rustc 1.74.0 (79e9716c9 2023-11-13)

**relevant settings**: None

before:
```rust
struct MyStruct {
    number: usize,
}

impl MyStruct {
    fn new(number: usize) -> Self {
        Self { number } as Self
    }
}

// example 1
let s = MyStruct::new(123);

// example 2
let n = u64::from(123u32);
```

after (actual):
```rust
struct MyStruct {
    number: usize,
}

impl MyStruct {
    fn new(number: usize) -> Self {
        Self { number } as Self
    }
}

// example 1
let s = {
    let number = 123;
    MyStruct { number } as Self
};

// example 2
let n = {
    let small = 123u32;
    small as Self
};
```

after (expected):
```rust
struct MyStruct {
    number: usize,
}

impl MyStruct {
    fn new(number: usize) -> Self {
        Self { number } as Self
    }
}

// example 1
let s = {
    let number = 123;
    MyStruct { number } as MyStruct
};

// example 2
let n = {
    let small = 123u32;
    small as u64
};
```

`Self` should be changed to what the `Self` type was in the function (`MyStruct` or `u64` here)

found after reporting #16090, might have the exact same fix but i have little experience here with the codebase

### LINKED ISSUE #16113: Inlining functions with Self
I start with the following code: 
```rust
#[derive(PartialEq, Eq)]
enum Enum {
    A,
    B,
    C,
}

impl Enum {
    fn a_or_b_eq(&self) -> bool {
        self == &Self::A || self == &Self::B
    }

    fn a_or_b_match(&self) -> bool {
        matches!(self, Self::A | Self::B)
    }
}

fn a() -> bool {
    Enum::A.a_or_b_eq() && Enum::A.a_or_b_match()
}
```
Inlining both the functions inside `a()` produces:
```rust
fn a() -> bool {
    ({
        let ref this = Enum::A;
        this || self == &Self::Bthis}) && {
            let ref this = Enum::A;
            matches!this
        }
}
```
which has many issues, including syntax errors. I would expect a result similar to:
```rust
fn a() -> bool {
    ({
        let ref this = Enum::A;
        this == &Enum::A || this == &Enum::B}) && ({
            let ref this = Enum::A;
            matches!(this, Enum::A | Enum::B)
        })
}
```

**rust-analyzer version**: 0.3.1766-standalone

**rustc version**: 1.74.1 (a28077b28 2023-12-04)

**relevant settings**: nothing obviously relevant.

roife:
Fix #16113, fix #16091

The problem described in this issue actually involves three bugs.

Firstly, when using `ted` to modify the syntax tree, the offset of nodes on the tree changes, which causes the syntax range information from `hir` to become invalid. Therefore, we need to edit the AST after the last usage for `usages_for_locals`.

The second issue is that when inserting nodes, it's necessary to use `clone_subtree` for duplication because the `ted::replace` operation essentially moves a node.

The third issue is that we should use `ancestors_with_macros` instead of `ancestors` to handle impl definition in macros.

I have fixed the three bugs mentioned above and added unit tests.

### DIFF
--- a/crates/ide-assists/src/handlers/inline_call.rs
+++ b/crates/ide-assists/src/handlers/inline_call.rs
@@ -373,7 +373,9 @@ fn inline(
     // We should place the following code after last usage of `usages_for_locals`
     // because `ted::replace` will change the offset in syntax tree, which makes
     // `FileReference` incorrect
-    if let Some(imp) = body.syntax().ancestors().find_map(ast::Impl::cast) {
+    if let Some(imp) =
+        sema.ancestors_with_macros(fn_body.syntax().clone()).find_map(ast::Impl::cast)
+    {
         if !node.syntax().ancestors().any(|anc| &anc == imp.syntax()) {
             if let Some(t) = imp.self_ty() {
                 while let Some(self_tok) = body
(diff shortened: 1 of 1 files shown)


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16117, opened 2023-12-13
feat: completion list suggests constructor like & builder methods first

mustakimali:
When typing `MyType::` the completion items' order could be re-ordered based on how likely we want to select those:
* Constructors: `new` like functions to be able to create the type,
* Constructors that take args: Any other function that creates `Self`,
* Builder Methods: any builder methods available,
* Regular methods & associated functions (no change there)

image

In this photo, the order is:
* `new` constructor is first
* `new_builder` second is a builder method
* `aaaanew` is a constructor that takes arguments, is third  and is irrespective of its alphabetical order among names.

---

Another Example using actix `HttpServer` shows preferring constructor without `self` arg first (the `new` method)

image

image

I've dropped my previous idea of highlighting these functions in the rustdoc (https://github.com/rust-lang/rust/pull/107926)

### DIFF
--- a/crates/ide-completion/src/item.rs
+++ b/crates/ide-completion/src/item.rs
@@ -211,8 +211,8 @@ pub enum CompletionRelevancePostfixMatch {
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub struct CompletionRelevanceFn {
-    pub has_args: bool,
-    pub has_self_arg: bool,
+    pub has_params: bool,
+    pub has_self_param: bool,
     pub return_type: CompletionRelevanceReturnType,
 }
 
@@ -310,10 +310,10 @@ impl CompletionRelevance {
                 // Bump Constructor or Builder methods with no arguments,
                 // over them tha with self arguments
                 if fn_score > 0 {
-                    if !asf.has_args {
+                    if !asf.has_params {
                         // bump associated functions
                         fn_score += 1;
-                    } else if asf.has_self_arg {
+                    } else if asf.has_self_param {
                         // downgrade methods (below Constructor)
                         fn_score = 1;
                     }
--- a/crates/ide-completion/src/item.rs
+++ b/crates/ide-completion/src/item.rs
@@ -166,6 +166,8 @@ pub struct CompletionRelevance {
     pub postfix_match: Option<CompletionRelevancePostfixMatch>,
     /// This is set for type inference results
     pub is_definite: bool,
+    /// This is set for items that are function (associated or method)
+    pub function: Option<CompletionRelevanceFn>,
 }
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
@@ -207,6 +209,24 @@ pub enum CompletionRelevancePostfixMatch {
     Exact,
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct CompletionRelevanceFn {
+    pub has_args: bool,
+    pub has_self_arg: bool,
+    pub return_type: CompletionRelevanceReturnType,
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum CompletionRelevanceReturnType {
+    Other,
+    /// Returns the Self type of the impl/trait
+    DirectConstructor,
+    /// Returns something that indirectly constructs the `Self` type of the impl/trait e.g. `Result<Self, ()>`, `Option<Self>`
+    Constructor,
+    /// Returns a possible builder for the type
+    Builder,
+}
+
 impl CompletionRelevance {
     /// Provides a relevance score. Higher values are more relevant.
     ///
@@ -231,6 +251,7 @@ impl CompletionRelevance {
             postfix_match,
             is_definite,
             is_item_from_notable_trait,
+            function,
         } = self;
 
         // lower rank private things
@@ -275,6 +296,33 @@ impl CompletionRelevance {
         if is_definite {
             score += 10;
         }
+
+        score += function
+            .map(|asf| {
+                let mut fn_score = match asf.return_type {
+                    CompletionRelevanceReturnType::DirectConstructor => 15,
+                    CompletionRelevanceReturnType::Builder => 10,
+                    CompletionRelevanceReturnType::Constructor => 5,
+                    CompletionRelevanceReturnType::Other => 0,
+                };
+
+                // When a fn is bumped due to return type:
+                // Bump Constructor or Builder methods with no arguments,
+                // over them tha with self arguments
+                if fn_score > 0 {
+                    if !asf.has_args {
+                        // bump associated functions
+                        fn_score += 1;
+                    } else if asf.has_self_arg {
+                        // downgrade methods (below Constructor)
+                        fn_score = 1;
+                    }
+                }
+
+                fn_score
+            })
+            .unwrap_or_default();
+
         score
     }
 
--- a/crates/ide-completion/src/render/function.rs
+++ b/crates/ide-completion/src/render/function.rs
@@ -8,8 +8,13 @@ use syntax::{format_smolstr, AstNode, SmolStr};
 
 use crate::{
     context::{CompletionContext, DotAccess, DotAccessKind, PathCompletionCtx, PathKind},
-    item::{Builder, CompletionItem, CompletionItemKind, CompletionRelevance},
-    render::{compute_exact_name_match, compute_ref_match, compute_type_match, RenderContext},
+    item::{
+        Builder, CompletionItem, CompletionItemKind, CompletionRelevance, CompletionRelevanceFn,
+        CompletionRelevanceReturnType,
+    },
+    render::{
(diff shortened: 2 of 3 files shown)


crates/ide-completion/src/item.rs
```diff
@@ -164,6 +164,8 @@ pub struct CompletionRelevance {
     pub postfix_match: Option<CompletionRelevancePostfixMatch>,
     /// This is set for type inference results
     pub is_definite: bool,
+    /// Any other bonuses we want to add,
+    pub bonus_score: u32,
```
Veykril, line 168: This defeats the purpose of `CompletionRelevance`, the whole point of it is to have fields representing properties that influence the final score. This field just spreads this final score across multiple locations. This should instead be an enum like

```rs
enum CompletionRelevanceConstructorType {
    /// Not a constructor
    None,
    /// Returns the Self type of the impl/trait
    DirectConstructor,
    /// Returns something that indirectly constructos the `Self` type of the impl/trait e.g. `Result<Self, ()>`, `Option<Self>`
    Constructor,
    /// Returns a possible builder for the type
    Builder,
}
```
and then assign points respectively where we calculate all the other points. Those variants are the interesting cases I can think of at least, `Constructor` might be tricky to implement though, but that can be skipped for now in that case.

crates/ide-completion/src/render/function.rs
```diff
@@ -153,6 +154,34 @@ fn render(
...
 }
 
+/// When typing `::` of a type, the preferred order is:
+/// * Constructors: new like functions to be able to create the type,
+/// * Builder Methods,
+/// * Constructors that take args: Any other function that creates Self
+/// * Regular methods & Associated functions
+///
+fn calculate_bonus(ctx: &RenderContext<'_>, func: hir::Function, db: &dyn HirDatabase) -> u32 {
+    if ctx.token().kind() != syntax::SyntaxKind::COLON2 || func.self_param(db).is_some() {
```
Veykril, line 164: This should check whether we are completing an associated function or not instead of checking the token kind, `func.as_assoc_item(ctx.db()).is_some()` can be used for that

crates/ide-completion/src/render/function.rs
```diff
@@ -153,6 +154,34 @@ fn render(
...
+    let mut bonus = 0;
+
+    let has_args = !func.assoc_fn_params(db).is_empty();
+    let ret_type = func.ret_type(db);
+    if !has_args && !ret_type.is_unit() {
+        // fn() -> A
+        bonus += 30;
+    } else if has_args && !ret_type.is_unit() {
+        // fn(..) -> A
+        bonus += 20;
```
Veykril, line 177: These need to check that the return type is the same as the `Self` type of the trait or impl the function is a part of, otherwise these are generally associated functions being preferred instead of just constructors.

Veykril (changes requested):

crates/ide-completion/src/item.rs
mustakimali: Thanks for some useful pointers, I've pushed a few commits.

> `Constructor` might be tricky to implement though, but that can be skipped for now in that case.

This needed some ugly string comparison - if this is not good enough then I can create a separate PR and attempt to implement it properly later.

crates/ide-completion/src/item.rs
```diff
@@ -269,6 +292,27 @@ impl CompletionRelevance {
...
+                    CompletionRelevanceFnType::DirectConstructor => 15,
+                    CompletionRelevanceFnType::Builder => 10,
+                    CompletionRelevanceFnType::Constructor => 5,
+                    CompletionRelevanceFnType::Other => 0,
+                };
+
+                // Prefer functions with no arguments, then functions with self arguments
+                if score > 0 {
+                    score += if !asf.has_args { 2 } else { 0 };
+                    score -= if asf.has_self_arg { score - 1 } else { 0 };
```
Veykril, line 309:
```suggestion
                    score -= if asf.has_self_arg { 1 } else { 0 };
```

This sets score to `-1`, looks like a mistake. Though subtracting is risky since we start with `score = 0` which could underflow

crates/ide-completion/src/item.rs
```diff
@@ -205,6 +207,26 @@ pub enum CompletionRelevancePostfixMatch {
...
+    pub has_args: bool,
+    pub has_self_arg: bool,
+    pub ty: CompletionRelevanceFnType,
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum CompletionRelevanceFnType {
+    Other,
+    /// Returns a type that is expected in this context
+    ReturnsExpectedType,
```
Veykril, line 221: This variant should not be needed, iirc we set `type_match` according to the return type for functions when needed.

crates/ide-completion/src/render/function.rs
```diff
@@ -153,6 +159,62 @@ fn render(
...
+    ctx: &RenderContext<'_>,
+    func: hir::Function,
+    func_kind: &FuncKind<'_>,
+    db: &dyn HirDatabase,
+) -> Option<CompletionRelevanceFn> {
+    if func.as_assoc_item(ctx.db()).is_none() {
+        return None;
+    }
+
+    let has_args = !func.assoc_fn_params(db).is_empty();
```
Veykril, line 172:
```suggestion
    let has_args = func.num_params(db) > 0;
```

crates/ide-completion/src/render/function.rs
```diff
@@ -153,6 +159,62 @@ fn render(
...
+        .unwrap_or_else(|| (false, false));
+
+    let ty = if !returns_self
+        && !ret_unit_type
+        && ctx.completion.expected_type.as_ref() == Some(&ret_type)
+    {
+        // impl Foo { fn baz(&self) -> u32 { 0 } }
+        // let _: u32 = foo.$0; // baz is preferred as it returns expected u32
+        CompletionRelevanceFnType::ReturnsExpectedType
+    } else if ret_type.display(db).to_string().ends_with("Builder") {
```
Veykril, line 202:
```suggestion
    } else if ret_type.as_adt(db).map_or(false, |adt| adt.name(db).ends_with("Builder")) {
```

crates/ide-completion/src/render/function.rs
```diff
@@ -153,6 +159,62 @@ fn render(
...
+    let has_args = !func.assoc_fn_params(db).is_empty();
+    let ret_type = func.ret_type(db);
+    let has_self_arg = func.self_param(db).is_some();
+    let ret_unit_type = ret_type.is_unit();
+    let self_type = match func_kind {
+        FuncKind::Function(PathCompletionCtx {
+            qualified: Qualified::With { path, .. }, ..
+        }) => path.segment().and_then(|ps| ps.name_ref()).map(|n| n.to_string()),
+        _ => None,
+    }; // self type without any generic args
```
Veykril, line 181: String comparing is certainly the wrong approach here, take a look at how we populate the `type_match` field of the relevance struct for how to possibly do this.

Veykril (changes requested):

crates/ide-completion/src/render/function.rs
mustakimali: I have pushed some commit to do this reliably.

crates/ide-completion/src/item.rs
mustakimali: There was a check ` if score > 0 {` to ensure it never underflows, however I've rewrite this for clarity this time.

crates/ide-completion/src/render/function.rs
```diff
@@ -157,6 +169,65 @@ fn render(
...
+            let Some(hir::PathResolution::Def(hir::ModuleDef::Adt(hir::Adt::Struct(struct_)))) =
+                resolution
+            else {
+                return None;
+            };
+
+            Some(struct_.ty(db))
+        }
+        _ => None,
+    };
```
Veykril, line 200: This is the wrong way to fetch the self type for the impl. For one because it only works for structs, and the other reason being that this discards any generics of the struct. Instead we can make use of the `as_assoc_item` result as that gives us an `AssocItem`,  then you can fetch the self type via `AssocItem::implementing_ty` (its a new function, so you might need to rebase on latest master)

crates/ide-completion/src/render/function.rs
```diff
@@ -157,6 +169,65 @@ fn render(
...
+    let return_type = self_type
+        .map(|self_type| {
+            if match_types(ctx.completion, &self_type, &ret_type).is_some() {
+                // fn([..]) -> Self
+                CompletionRelevanceReturnType::DirectConstructor
+            } else if ret_type
+                .type_arguments()
+                .flat_map(|ret_type_arg| match_types(ctx.completion, &self_type, &ret_type_arg))
+                .next()
+                .is_some()
```
Veykril, line 211:
```suggestion
                .any(|ret_type_arg| match_types(ctx.completion, &self_type, &ret_type_arg).is_some())
```

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16125, opened 2023-12-14
Run rust-analyzer on rustc tests in metrics

### LINKED ISSUE #15947: Run rust-analyzer on the rustc test suit
Rust-analyzer supports several diagnostics, running it against the rustc ui tests and checking the equivalency of them (we don't have equal error messages and spans, but even checking the count of errors will probably work) can help a lot in discovering rare edge case bugs in the analysis and diagnostics logic I think.

We can start having them in metrics (as it will probably have many fails at the start) and later moving them into the CI if there were near zero problems and it was fast to check.

HKalbasi:
fix #15947

### DIFF
--- a/.github/workflows/metrics.yaml
+++ b/.github/workflows/metrics.yaml
@@ -67,7 +67,7 @@ jobs:
   other_metrics:
     strategy:
       matrix:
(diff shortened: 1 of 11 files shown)


Veykril:
Ooh, only 3 minutes for those tests, nice.

PR #16131, opened 2023-12-15
Fix false positive type mismatch in const reference patterns

HKalbasi:


crates/hir-ty/src/infer/pat.rs
```diff
@@ -496,24 +496,28 @@ impl InferenceContext<'_> {
...
+        match &body[pat] {
+            Pat::Tuple { .. }
+            | Pat::TupleStruct { .. }
+            | Pat::Record { .. }
+            | Pat::Range { .. }
+            | Pat::Slice { .. } => true,
+            Pat::Or(pats) => pats.iter().all(|p| self.is_non_ref_pat(body, *p)),
+            Pat::Path(p) => {
+                let v = self.resolve_value_path_inner(p, pat.into());
+                v.is_some_and(|x| !matches!(x.0, hir_def::resolver::ValueNs::ConstId(_)))
```
Veykril, line 510: I think we need to accomodate for enum unit variants and unit structs here as well no? (since we don't inherently treat their resolution as a const)

HKalbasi: This function should return `true` for structs and enum variants, since we need to automatically dereference in those cases. That is, this code is valid:
```Rust
match &&&&&&&e {
    Enum::Variant => ()
}
```
But this is type error:
```Rust
const ENUM_VARIANT_AS_CONST: Enum = Enum::Variant;
match &&&&&&&e {
    ENUM_VARIANT_AS_CONST => ()
}
```

Veykril: Gotcha

PR #16135, opened 2023-12-15
Extend `TargetSpec` functionality to rust-project.json

davidbarsky:
Resolves https://github.com/rust-lang/rust-analyzer/issues/15892.

This PR, as structured, adds runnable support to `rust-project.json`-based workspaces, and that works pretty nicely. It's not as polished as I'd like it to be (namely: interpolation of `test_id` feels a bit janky and undocumented, the), but I'm opening this PR as a draft for general direction/feedback.

I think that this approach can also make https://github.com/rust-lang/rust-analyzer/pull/15476 unnecessary by adding a `flycheck_command` to `rust-project.json` _while also_ extending similar functionality to Cargo only build the impacted crate.

### DIFF
--- a/crates/project-model/src/project_json.rs
+++ b/crates/project-model/src/project_json.rs
@@ -56,7 +56,7 @@ use rustc_hash::FxHashMap;
 use serde::{de, Deserialize};
 use span::Edition;
 
-use crate::cfg_flag::CfgFlag;
+use crate::{cfg_flag::CfgFlag, TargetKind};
 
 /// Roots and crates that compose this Rust project.
 #[derive(Clone, Debug, Eq, PartialEq)]
@@ -73,20 +73,37 @@ pub struct ProjectJson {
 /// useful in creating the crate graph.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Crate {
-    pub(crate) display_name: Option<CrateDisplayName>,
-    pub(crate) root_module: AbsPathBuf,
-    pub(crate) edition: Edition,
-    pub(crate) version: Option<String>,
-    pub(crate) deps: Vec<Dependency>,
-    pub(crate) cfg: Vec<CfgFlag>,
-    pub(crate) target: Option<String>,
-    pub(crate) env: FxHashMap<String, String>,
-    pub(crate) proc_macro_dylib_path: Option<AbsPathBuf>,
-    pub(crate) is_workspace_member: bool,
-    pub(crate) include: Vec<AbsPathBuf>,
-    pub(crate) exclude: Vec<AbsPathBuf>,
-    pub(crate) is_proc_macro: bool,
-    pub(crate) repository: Option<String>,
+    pub display_name: Option<CrateDisplayName>,
+    pub root_module: AbsPathBuf,
+    pub edition: Edition,
+    pub version: Option<String>,
+    pub deps: Vec<Dependency>,
+    pub cfg: Vec<CfgFlag>,
+    pub target: Option<String>,
+    pub env: FxHashMap<String, String>,
+    pub proc_macro_dylib_path: Option<AbsPathBuf>,
+    pub is_workspace_member: bool,
+    pub include: Vec<AbsPathBuf>,
+    pub exclude: Vec<AbsPathBuf>,
+    pub is_proc_macro: bool,
+    pub repository: Option<String>,
+    pub target_spec: Option<TargetSpec>,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct TargetSpec {
+    pub manifest_file: AbsPathBuf,
+    pub target_label: String,
+    pub target_kind: TargetKind,
+    pub runnables: Runnables,
+    pub flycheck_command: Vec<String>,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Runnables {
+    pub check: Vec<String>,
+    pub run: Vec<String>,
+    pub test: Vec<String>,
 }
 
 impl ProjectJson {
@@ -121,6 +138,20 @@ impl ProjectJson {
                         None => (vec![root_module.parent().unwrap().to_path_buf()], Vec::new()),
                     };
 
+                    let target_spec = match crate_data.target_spec {
+                        Some(spec) => {
+                            let spec = TargetSpec {
+                                manifest_file: absolutize_on_base(spec.manifest_file),
+                                target_label: spec.target_label,
+                                target_kind: spec.target_kind.into(),
+                                runnables: spec.runnables.into(),
+                                flycheck_command: spec.flycheck_command,
+                            };
+                            Some(spec)
+                        }
+                        None => None,
+                    };
+
                     Crate {
                         display_name: crate_data
                             .display_name
@@ -149,6 +180,7 @@ impl ProjectJson {
                         exclude,
                         is_proc_macro: crate_data.is_proc_macro,
                         repository: crate_data.repository,
+                        target_spec,
                     }
                 })
                 .collect(),
@@ -171,6 +203,14 @@ impl ProjectJson {
     /// Returns the path to the project's root folder.
     pub fn path(&self) -> &AbsPath {
         &self.project_root
+    }
+
+    pub fn crate_by_root(&self, root: &AbsPath) -> Option<Crate> {
+        self.crates
+            .iter()
+            .filter(|krate| krate.is_workspace_member)
+            .find(|krate| &krate.root_module == root)
+            .cloned()
     }
 }
 
@@ -201,6 +241,8 @@ struct CrateData {
     is_proc_macro: bool,
     #[serde(default)]
     repository: Option<String>,
+    #[serde(default)]
+    target_spec: Option<TargetSpecData>,
 }
 
 #[derive(Deserialize, Debug, Clone)]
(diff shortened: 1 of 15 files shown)


crates/project-model/src/project_json.rs
```diff
@@ -217,6 +260,55 @@ enum EditionData {
...
+pub enum TargetKindData {
+    Bin,
+    /// Any kind of Cargo lib crate-type (dylib, rlib, proc-macro, ...).
+    Lib,
+    Example,
+    Test,
+    Bench,
+    BuildScript,
+    Other,
+}
```
Veykril, line 290: This ties the cargo workspace architecture into project-json, I'm not sure that this was something @matklad had in mind when giving this alternative format for project descriptions.

Is there a need for things like `BuildScript` and `Example`? `Bin`, `Lib` and `Test` (and `Other` ofc) seem fine as they are rather general.

crates/rust-analyzer/src/target_spec.rs
```diff
@@ -1,35 +1,164 @@
...
+}
+
 #[derive(Clone)]
 pub(crate) struct CargoTargetSpec {
     pub(crate) workspace_root: AbsPathBuf,
-    pub(crate) cargo_toml: ManifestPath,
+    /// [`project_manifest`] corresponds to the file defining a crate. With Cargo,
+    /// this will be a Cargo.toml, but with a non-Cargo build system, this might be
+    /// a `TARGETS` or `BUCK` file. Multiple, distinct crates can share a
+    /// single `project_manifest`.
```
Veykril, line 99: This comment confuses me given where its placed, this is the field of a `CargoTargetSpec`, so this should always be a `Cargo.toml`.

docs/dev/lsp-extensions.md
```diff
@@ -372,7 +372,7 @@ interface Runnable {
 }
 ```
 
-rust-analyzer supports only one `kind`, `"cargo"`. The `args` for `"cargo"` look like this:
+rust-analyzer supports two `kind`s of runnables, `"cargo"` and `"rustproject"`. The `args` for `"cargo"` look like this:
```
Veykril, line 375: `projectjson`, not `rustproject` right?

docs/dev/lsp-extensions.md
```diff
@@ -385,6 +385,16 @@ rust-analyzer supports only one `kind`, `"cargo"`. The `args` for `"cargo"` look
 }
 ```
 
+The args for `"jsonproject"` look like:
```
Veykril, line 388: `projectjson`

editors/code/src/run.ts
```diff
@@ -104,24 +104,34 @@ export function prepareEnv(
...
+        const projectJsonRunnableArgs = runnable.args as ra.ProjectJsonRunnable;
+        args = projectJsonRunnableArgs.args;
+
+        definition = {
+            type: tasks.TASK_TYPE,
+            command: args[0],
+            args: args.slice(1),
+            cwd: projectJsonRunnableArgs.workspaceRoot,
+            env: prepareEnv(runnable, config.runnablesExtraEnv),
+            overrideCargo: "buck2",
```
Veykril, line 131: Does this mean only `buck2` is supported for projectjson runnables?

Veykril (commented):
Overall implementation wise it looks good to me, al though I am still a bit on the fence wrt to having flycheck be in here. I am personally still of the opinion that flycheck should not be a direct part of r-a (but a separate extension altogether). Adding it to this file makes that less clear of a separation.

editors/code/src/run.ts
matklad: Yeah, ideally we'd fetch `buck2` string from `projectJsonRunnableArgs`

crates/project-model/src/project_json.rs
```diff
@@ -202,6 +243,8 @@ struct CrateData {
     is_proc_macro: bool,
     #[serde(default)]
     repository: Option<String>,
+    #[serde(default)]
+    target_spec: Option<TargetSpecData>,
```
matklad, line 247: Some notes on JSON side of things here:

I'd maybe call the top-level thing `build_info`, as thre might be multiple targets (in buck2 sense) concerning a single crate. And, well, target is already overloaded, so it's better to avoid pilling more meanings onto that poor word.

I _think_ the core feature here should be an inclusion of opaque blob of data, which is transmitted over LSP so that the custom client may interpret this. So, I'd say the basis should look like

```
build_info: Option<BuildInfo>

struct BuildInfo {
    data: serde_json::Value
}
```

Eg, one can imagine a situation where _all_ builds are remote, and the way you "run" software is by issueing an HTTP request to a remote execution server, and opening a page in the web-browser with results. In this situation, the `data` would be an URL to call, and the client-side extension would take care of HTTPing.

But we also should support "standard" workflows where you just run some command line. 

So, something like (pardon my TypeScript)

```
struct BuildInfo {
    data: serde_json::Value
    runnables: Vec<Runnable>
}

struct Runnable {
    kind: "build" | "check" | "test" | "run" | string,
    label: string,
    argv: string[],
    cwd: string
}
```

This _shouldn't_ use a json map, but rather an array, to be more flexible. Eg, in the future we might want to add `--release` and non-`--release` versions of tests, and with a list it should be simpler. More generally, I've recently read some JSON API guidlines somewhere (don't remember whether though), and one guideline that stuck with me is that you should represent maps in JSON as arrays (assoc lists). I think that's a good guideline and something we should be using in gerneral. 

_Then_, the `kind` should be a semi-open enum. rust-analyzer _should_ treat `build`, `run`, `check` and `test` specially, but it should also be able to show other tasks in GUI. Eg, you can imagine a "flush" task for an embedded build system, or some such. 

We _might_ want to split `check` into two --- one check that is run for the user and displays errors in the terminal, and another check that runs with message-format=json and is used for flycheck. I guess we could call those `check` and `flycheck`?

For _tests_, we need to think how to enable filtering. The core feature is that you can run a single test using `cargo test -- $TEST_NAME`, and all tests using `cargo test`. Note that this might be very different in non-cargo build systems. Eg, it could be `builder test` and `builder test --test-filter=$TEST_NAME`, where `builder test --test-filter=''` would run zero tests, rather than all tests. So I think for test, we should also split it into two: "testall" and "testone". And, for "testone", I _think_ we want to make `argv` a template. That is, argv could look like

```
["builder", "test", "--test-filter=$$TEST_NAME$$", "./buildfile.b"]
```

We shouldn't assume that filter is always last, and that its always a separate argument.

editors/code/src/run.ts
davidbarsky: Oh, yeah, sorry—I meant to pull this string from configuration. Didn’t realize I left it here.

crates/project-model/src/project_json.rs
matklad: I _think_ https://github.com/rust-lang/rust-analyzer/pull/16135#discussion_r1447221391 answers this, but, yeah good catch of the specific smell here!

docs/dev/lsp-extensions.md
davidbarsky: Yes, sorry! I get the two switched up in my head.

crates/rust-analyzer/src/target_spec.rs
davidbarsky: Ah—the intention behind this comment is to communicate that a non-Cargo project can also have a `TargetSpec` (née `CargoTargetSpec`), with the intention of trying to make non-Cargo build systems be more Cargo-like in order to reduce the maintenance burden on `rust-analyzer`, but as mentioned elsewhere, that might be a design bug.

crates/project-model/src/project_json.rs
davidbarsky: They're not really needed, no—they were meant to try and make `project-json` more Cargo-like. I only _really_ need  `Bin`, `Lib` and `Test` if this PR doesn't take the approach that matklad suggested in his comment.

crates/project-model/src/project_json.rs
davidbarsky: Thanks for the feedback here!

> I'd maybe call the top-level thing `build_info`, as thre might be multiple targets (in buck2 sense) concerning a single crate. And, well, target is already overloaded, so it's better to avoid pilling more meanings onto that poor word.

That's a fair point. I didn't intend for the naming to be Buck/Bazel-esque "targets", I mean to invoke the [target](https://doc.rust-lang.org/cargo/reference/cargo-targets.html) concept in _Cargo_, of all things. In any case, `target` is too overloaded and I'll move over to `build_info`.

> I think the core feature here should be an inclusion of opaque blob of data, which is transmitted over LSP so that the custom client may interpret this.

Is the idea that a companion extension to `rust-lang.rust-analyzer` (e.g., something for Buck or Bazel) would be responsible for assembling that opaque blob of data into a some sort of meaningful runnable command or is more that since I'm making this change now, I should make it more futureproof ("[B]ut we also should support "standard" workflows where you just run some command line.")? I ask since I'm working on a variable interpreter/template thingy, which—I think—should make expressing arbitrary build/test commands a lot easier in rust-analyzer itself, which reduces the pressure/need for end-users to install other extensions for runnables.

> This shouldn't use a json map, but rather an array, to be more flexible. Eg, in the future we might want to add `--release` and non-`--release` versions of tests, and with a list it should be simpler. More generally, I've recently read some JSON API guidlines somewhere (don't remember whether though), and one guideline that stuck with me is that you should represent maps in JSON as arrays (assoc lists). I think that's a good guideline and something we should be using in gerneral.

That seems reasonable to me! I was considering how to tackle this for Buck[^1] and this seems like a good approach. 

> Then, the `kind` should be a semi-open enum. rust-analyzer should treat `build`, `run`, `check` and `test` specially, but it should also be able to show other tasks in GUI. Eg, you can imagine a "flush" task for an embedded build system, or some such. 

Yeah, that seems smart.

> We might want to split `check` into two --- one `check` that is run for the user and displays errors in the terminal, and another check that runs with `message-format=json` and is used for flycheck. I guess we could call those `check` and `flycheck`?

I think now's a good time to make this change, especially if the new runnable API is a breaking change.

For tests, we need to think how to enable filtering. The core feature is that you can run a single test using `cargo test -- $TEST_NAME`, and all tests using `cargo test`. Note that this might be very different in non-cargo build systems. Eg, it could be `builder test` and `builder test --test-filter=$TEST_NAME`, where `builder test --test-filter=''` would run zero tests, rather than all tests. So I think for test, we should also split it into two: `"testall"` and `"testone"`. And, for `"testone"`, I _think_ we want to make `argv` a template. That is, argv could look like: [redacted for quoting ease]

I don't think I've fully gotten this into my head yet (sorry, lots of travel!) but I think I see the utility of a change like this.

I think I'd like like a `debug` variant as well, but I haven't really through _this specific aspect_ through.

[^1]: For context, buck2 is moving to a [modifier-style approach](https://github.com/facebook/buck2/blob/main/docs/rfcs/drafts/cfg-modifiers/api.md#cli-modifier). A user can write `buck2 build repo//foo:bar?linux,release`, which translate to "`build repo//foo:bar` for Linux using `release`". there's a more verbose `--modifier`-style API, but I suspect that both options can work when called from rust-analyzer _or_ I suspect the Buck team will be flexible in implementing this.

crates/rust-analyzer/src/target_spec.rs
Veykril: I get that, it's just that the location for this command seems ill placed, though i don't have a better spot for it either aside from somewhere in the module documentation.

Veykril:
> Overall implementation wise it looks good to me, al though I am still a bit on the fence wrt to having flycheck be in here. I am personally still of the opinion that flycheck should not be a direct part of r-a (but a separate extension altogether). Adding it to this file makes that less clear of a separation.

I guess whether the format has that or not makes little difference, the command itself is heavily tied into the rest of the specification here whether it is supported by a separate extension or r-a.

davidbarsky:
> > Overall implementation wise it looks good to me, al though I am still a bit on the fence wrt to having flycheck be in here. I am personally still of the opinion that flycheck should not be a direct part of r-a (but a separate extension altogether). Adding it to this file makes that less clear of a separation.
> 
> 
> 
> I guess whether the format has that or not makes little difference, the command itself is heavily tied into the rest of the specification here whether it is supported by a separate extension or r-a.

While I agree with you that flycheck is a hack, I've found myself of the opinion that `rust-analyzer` (and specifically, its users who are not interested/aware of the need to install a build system-specific extension) can likely get a _ridiculous_ amount of mileage through a templates commands and a promise from build systems to place `rustc` diagnostics in a deterministic location (like, say, stdout).

And yeah, I think the command's specification is largely immaterial to where it runs, or at the very least, it's not a binding commitment in one direction or another.

PR #16136, opened 2023-12-16
fix(completion): make the expected type a tad smarter with `Fn`s

saiintbrisson:
This commit changes how the expected type is calculated when
working with Fn pointers, making the parenthesis stop vanishing
when completing the function name.

I've been bugged by the behavior of parenthesis completion for
a long while now. R-a assumes that the `LetStmt` type is the same
as the function type I've just written. Worse is that all parenthesis
vanish, even from functions that have completely different signatures.
It will now verify if the signature is the same.

While working on this, I noticed that record fields behave the same,
so I also made it prioritize the field type instead of the current
expression when possible, but I'm unsure if this is OK, so input is
appreciated.

ImplTraits as return types will still behave weirdly because lowering
is disallowed at the time it resolves the function types.

image
image
image

### DIFF
--- a/crates/ide-completion/src/context/analysis.rs
+++ b/crates/ide-completion/src/context/analysis.rs
@@ -361,7 +361,12 @@ fn expected_type_and_name(
                     let ty = it.pat()
                         .and_then(|pat| sema.type_of_pat(&pat))
                         .or_else(|| it.initializer().and_then(|it| sema.type_of_expr(&it)))
-                        .map(TypeInfo::original);
+                        .map(TypeInfo::original)
+                        .filter(|ty| {
+                            // don't infer the let type if the expr is a function,
+                            // preventing parenthesis from vanishing
+                            it.ty().is_some() || !ty.is_fn()
+                        });
                     let name = match it.pat() {
                         Some(ast::Pat::IdentPat(ident)) => ident.name().map(NameOrNameRef::Name),
                         Some(_) | None => None,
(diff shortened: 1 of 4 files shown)


crates/ide-completion/src/context/analysis.rs
```diff
@@ -413,20 +414,16 @@ fn expected_type_and_name(
...
+                        (ty, field_name)
                     } else {
                         cov_mark::hit!(expected_type_struct_field_followed_by_comma);
-                        let ty = sema.resolve_record_field(&it)
-                            .map(|(_, _, ty)| ty);
-                        (
-                            ty,
-                            it.field_name().map(NameOrNameRef::NameRef),
-                        )
+                        (field_ty, field_name)
```
saiintbrisson, line 426: I couldn't think of a reason as to why this was done this way, so please enlighten me if there is indeed a reason.

crates/ide-completion/src/context/analysis.rs
```diff
@@ -359,7 +359,8 @@ fn expected_type_and_name(
                     let ty = it.pat()
                         .and_then(|pat| sema.type_of_pat(&pat))
                         .or_else(|| it.initializer().and_then(|it| sema.type_of_expr(&it)))
-                        .map(TypeInfo::original);
+                        .map(TypeInfo::original)
+                        .filter(|ty| it.ty().is_some() || !ty.is_fn());
```
saiintbrisson, line 363: What I'm doing here is only taking `Fn` types into account for the let statement if it is explicitly defined. Is there a case I'm not seeing where this will be problematic?

Veykril: Don't think so but please add a commend for why this is done.

Veykril (approved):

crates/ide-completion/src/context/analysis.rs
Veykril: Unsure why either, your changes look correct here to me.

saiintbrisson:
@Veykril done!

Veykril:
Thanks!
@bors r+

Veykril:
Needs a rebase + reblessing the tests it seems as the output has changed.

PR #16143, opened 2023-12-18
internal: Move proc-macro knowledge out of base-db into hir-expand

Veykril:
It does not make much sense to me to have that live in base-db, additionally, it kind of conflicts with moving span things out into a separate crate


lnicola:
@bors r-

Let's wait a little for the subtree sync if you don't mind.

PR #16153, opened 2023-12-18
fix: Notify user that linkedProjects is set when failing to discover projects

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/15171


crates/rust-analyzer/src/reload.rs
```diff
@@ -174,7 +174,11 @@ impl GlobalState {
 
         if let Err(_) = self.fetch_workspace_error() {
             status.health = lsp_ext::Health::Error;
-            message.push_str("Failed to load workspaces.\n\n");
+            message.push_str("Failed to load workspaces.");
+            if self.config.has_linked_projects() {
+                message.push_str(" Note that linked projects have been specified, it looks like this setting has been misconfigured.");
```
davidbarsky, line 179: if it's alright with you, I've hacked together a change to this that includes the paths of _which_ linked projects have been specified. Would you want to me push to a separate PR with that change?

crates/rust-analyzer/src/reload.rs
```diff
@@ -174,7 +174,11 @@ impl GlobalState {
...
+            }
+            message.push_str("\n\n");
```
davidbarsky, line 181: Sorry, realized that it can be just:
```suggestion
            message.push_str("Failed to load workspaces.");
            if self.config.has_linked_projects() {
                let linked_projects = self.config.linked_projects();
                let (manifest_paths, inline_projects): (Vec<_>, Vec<_>) =
                    linked_projects.iter().partition(|linked_project| {
                        matches!(linked_project, LinkedProject::ProjectManifest(..))
                    });

                message.push_str(
                    "`rust-analyzer.linkedprojects` have been specified, which may cause indexing failures.\n\n ",
                );
                if !manifest_paths.is_empty() {
                    let paths = format!("paths to projects: {:?}", manifest_paths);
                    message.push_str(&paths)
                }

                match inline_projects.len() {
                    0 => (),
                    1 => message.push_str(
                        "The contents of the linked project were redacted due to its length.",
                    ),
                    _ => message.push_str(
                        "The contents of the linked projects were redacted due to its length.",
                    ),
                };
            }
```
feel free to take it or leave it :)

crates/rust-analyzer/src/reload.rs
Veykril: Feel free to PR that instead. I didn't bother putting much effort into this (just noticed I still had myself assigned to the issue :)

PR #16158, opened 2023-12-19
fix(mbe): desugar doc correctly for mbe

### LINKED ISSUE #16110: no code completion nor type information available when using `zerocopy`'s derive macros, e.g. `zerocopy::FromZeroes`
**rust-analyzer version**: 

using vscode ra extension, pre-release channel (also tried release, same result), version:

- Extension version: 0.4.1768
- rust-analyzer 0.3.1766-standalone (457b966b1 2023-12-10)

**rustc version**:

- rustc 1.76.0-nightly (90e321d82 2023-12-02)

**relevant settings**:

nothing special, toolchain install with rustup, extension installed from vscode marketplace, mostly default.

## reproducing

```console
> cargo new demo
> cd demo
> cargo add zerocopy --features derive
```

```rust
use zerocopy::FromZeroes;
#[repr(C)]
#[derive(Debug, Clone, Copy, FromZeroes)]
struct Header {
	magic: u32,
}
```

### if I use the derive macro from `zerocopy`, RA doesn't seem to pickup the impl of the derived trait.

- when I type `let header = Header::`, there's no method of the derived trait in the completion candidate lists, e.g. `FromZeroes::new_zeroed()`;
- if I manually type the method, no inlay type hints shows for the variable;
- when I type `header.`, the candidates list only shows keywords and snippets, no type information available;
- when I hover over the variable, the popup type hint overlay shows `{unknown}`;

screenshot:

image

### only having issue with `zerocopy`

not sure what's special about the `zerocopy` crate, derive macros from other crates (that I have used or tested) all work fine. also, the code compiles and runs without problem.

if I expand the derive macro, it looks like this:

```rust
#[allow(deprecated)]
unsafe impl ::zerocopy::FromZeroes for Header
where
	u32: ::zerocopy::FromZeroes,
{
	fn only_derive_is_allowed_to_implement_this_trait() {
	}
}
```

not knowing the inner working of RA, I would guess maybe it's the `where u32: FromZeroes` clause that confuses RA?

saiintbrisson:
Fixes #16110.

The way rust desugars doc comments when expanding macros is rendering it as raw strings delimited with hashes. Rust-analyzer wasn't aware of this, so the desugared doc comments wouldn't match correctly when on the LHS of macro declarations.

This PR fixes this by porting the code used by rustc:
https://github.com/rust-lang/rust/blob/59096cdad0e527ce4d254b55aca58539f726d3d6/compiler/rustc_ast/src/tokenstream.rs#L662-L671

### DIFF
--- a/crates/hir-def/src/macro_expansion_tests/mbe.rs
+++ b/crates/hir-def/src/macro_expansion_tests/mbe.rs
@@ -1218,8 +1218,10 @@ m! {
 macro_rules! m {
     ($(#[$m:meta])+) => ( $(#[$m])+ fn bar() {} )
 }
-#[doc = " Single Line Doc 1"]
-#[doc = "\n        MultiLines Doc\n    "] fn bar() {}
+#[doc = r" Single Line Doc 1"]
+#[doc = "
+        MultiLines Doc
+    "] fn bar() {}
 "##]],
     );
 }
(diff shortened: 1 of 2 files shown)


Veykril:
Thanks!
@bors r+

Veykril:
This either introduced a bug, or uncovered a bug somewhere else as our xflags invocation errors now 🤔

PR #16160, opened 2023-12-19
minor: Use reserve when removing markdown from text

Waqar144:
After markdown syntax removal the length of the text is roughly the same so we can reserve memory beforehand


Veykril:
Thanks!
@bors r+

PR #16162, opened 2023-12-19
minor: Optimizing Performance with `Promise.all` 🏎

sanjaiyan-dev:
Hello,

This pull request introduces a modest performance improvement by executing asynchronous tasks using `Promise.all`, thereby enhancing the efficient utilization of hardware resources.

Thank you for considering this enhancement.

Ref-: https://github.com/rust-lang/rust/pull/119108

PR #16164, opened 2023-12-19
minor: Bump `@vscode/test-electron`

lnicola:
Fixes this failure in #16162:

image


lnicola:
Looks like it's `code.cmd` instead of `code.exe`. I couldn't figure out which commit fixed it, but whatever.

@bors r+

PR #16167, opened 2023-12-20
fix: Fully remove dummy spans

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/16008

Some of these spans are certainly wrong, but since we discard invisible delimiters currently it doesn't really matter.


crates/hir-expand/src/attrs.rs
```diff
@@ -177,7 +177,7 @@ pub struct Attr {
     pub id: AttrId,
     pub path: Interned<ModPath>,
     pub input: Option<Interned<AttrInput>>,
-    pub ctxt: SyntaxContextId,
+    pub span: Span,
```
Veykril, line 180: This will have an impact on memory usage unfortunately (estimating 10mb on `self`), but we can't really avoid that

PR #16173, opened 2023-12-21
Implement `gen` block and function parsing

mohe2015:
https://github.com/rust-lang/rust-analyzer/issues/16156

Blocked on https://github.com/rust-lang/rust-analyzer/issues/16324

Things I need help with / need some pointers and TODOs:
- [x] ~~Only parse `gen` as keyword in 2024 edition~~
- [x] Hovering function name doesn't show `gen` in hover
- [x] Hovering over `gen` in block doesn't show keyword information. (no doc like https://doc.rust-lang.org/nightly/std/keyword.await.html for `gen` and `yield`)
- [x] When rust-analyzer is loading, vscode doesn't color the keyword correctly (https://github.com/dustypomerleau/rust-syntax/pull/41)
- [x] Yield autocompletion in `gen fn` and `async gen fn`
- [x] inferred type of async gen block is wrong
- [x] https://github.com/compiler-errors/rust/tree/closure-kind desugar to closure, https://github.com/compiler-errors/rust/blob/closure-kind/compiler/rustc_ast_lowering/src/expr.rs, https://github.com/compiler-errors/rust/blob/closure-kind/compiler/rustc_ast/src/ast.rs
- [x] Split off rename generator -> coroutine
- [x] Fix lowering async closures. They may need a nested closure or something
- [x] Fix return type inference of empty `gen fn'
- [x] Duplicate and wrong types in check_infer
- [x] Hover over function shows wrong return type
- [x] Type inference of resume of coroutine broken (seems like already before this PR)
- [x] Lots of tests


Veykril:
> Only parse `gen` as keyword in 2024 edition

Of note here is that our parsing is currently not edition aware at all (we just assume edition 2021 parsing), so we need to fix that first (as a separate PR if any).

Veykril:
Is this ready for review? Note that this won't be merged until we have solved the edition dependent parsing, as `gen` is a heavily used identifier I imagine.

mohe2015:
It's not ready for review (I realize marking as draft is not enough as you also use the S-waiting-on-author label), I mostly opened this PR to avoid duplicate work and to allow others to look at my messy code and use this if they need it. I would definitely need to clean this up before a review and if somebody else wants to take parts of this that's also fine as this is my first PR to rust-analyzer and I only have a very limited understanding of many parts of it that I touched here.

mohe2015:
That's nice! I'm currently writing my thesis so I probably won't have time for this. So anybody feel free to take this over.

Veykril:
Squashed and rebased

snprajwal:
I'm not too sure what needs to be done on top of rebasing this, but I'd be happy to take it up. Does anyone have any pointers?

Veykril:
I'll close this. The PR is quite outdated by now, and a bunch of refactors have landed since (and are still about to land especially the new trait solver). We should attempt this again after those.

PR #16174, opened 2023-12-21
Introduce a new CrateOrigin variant `Member`

alibektas:
CrateGraph has been suffering from ignoring a unified definition of a Workspace : Cargo rightly assumes that each project has its own workspace but in RA this assumption does not hold and this has caused problems for renaming symbols ( #15656 ). What I propose is to add a new variant to `CrateOrigin` called `Member` ( the name is open to discussion :D ). Member will be assigned to workspace members and this is easy to find out as cargo provides this information directly. `CrateOrigin::Local` will now be those crates that workspaces depend on and these crates need to be in the local filesystem and open to editing ( current state of things doesn't really reflect what I describe but anyway... ) and `CrateOrigin::Library` will be the *vendor* crates. By doing this we can now express things like "a symbol is renameable if it originates from a user defined crate ( meaning if `CrateOrigin` is `Local` or `Member`)". This follows in part @ogapo's findings so let's thank them too!

This also mitigates the issue we had [here](https://github.com/rust-lang/rust-analyzer/issues/15656#issuecomment-1857123869) by making all the `proc-macro` enabled crates have `CrateOrigin::Local` thus enabling renaming. So maybe hiding renaming behind a configuration may not be necessary after all


flodiebold:
As mentioned in #15656, this doesn't seem the right approach for determining "can we rename a symbol" to me :thinking:

alibektas:
Ok I am closing this issue as I see now @flodiebold's point after a discussion with @Veykril.

PR #16175, opened 2023-12-21
fix: Correctly set and mark the proc-macro spans

Veykril:
This slows down analysis by 2-3s on self for me unfortunately (~2.5% slowdown)

Noisy diff due to two simple refactoring in the first 2 commits. Relevant changes are [7d762d1](https://github.com/rust-lang/rust-analyzer/pull/16175/commits/7d762d18edb1b05343c7725af54f15ce81cd12de) and [1e1113c](https://github.com/rust-lang/rust-analyzer/pull/16175/commits/1e1113cf5f7318cae702bc31ab5155c24fade6f4) which introduce def site spans and correct marking for proc-macros respectively.

### DIFF
--- a/crates/hir-def/src/db.rs
+++ b/crates/hir-def/src/db.rs
@@ -1,7 +1,7 @@
 //! Defines database & queries for name resolution.
 use base_db::{salsa, CrateId, SourceDatabase, Upcast};
 use either::Either;
-use hir_expand::{db::ExpandDatabase, HirFileId};
+use hir_expand::{db::ExpandDatabase, HirFileId, MacroDefId};
 use intern::Interned;
 use la_arena::ArenaMap;
 use syntax::{ast, AstPtr};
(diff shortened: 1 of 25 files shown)


crates/hir-def/src/db.rs
```diff
@@ -335,10 +335,13 @@ fn macro_def(db: &dyn DefDatabase, id: MacroId) -> MacroDefId {
                 kind: kind(loc.expander, loc.id.file_id(), makro.ast_id.upcast()),
                 local_inner: false,
                 allow_internal_unsafe: loc.allow_internal_unsafe,
+                def_site: db
+                    .span_map(loc.id.file_id())
+                    .span_for_range(db.ast_id_map(loc.id.file_id()).get(makro.ast_id).text_range()),
```
Veykril: slow down comes from this, which I assume misses the cache for the span maps in some occurrences unfortunately

PR #16179, opened 2023-12-21
SCIP: populate new SymbolInformation fields

### LINKED ISSUE #15919: Implement new `SymbolInformation` fields in the SCIP backend
sourcegraph/scip/issues/154 added support for a number of new fields in SCIP's `SymbolInformation`:
 * `kind`
 * `display_name`
 * `signature_documentation`
 * `enclosing_symbol` (what was originally proposed as `owner` or `parent` in the linked issue)

rust-lang/rust-analyzer/pull/15806 recently updated scip to 0.3.1 which makes those new fields available to rust-analyzer. They are currently always defaulted.

@donsbot you said you wanted to add support for those in rust-analyzer, do you still plan on doing that?
If not I'm also interested in working on that, for use in Searchfox.

nicolas-guichard:
Fixes #15919

### DIFF
--- a/crates/ide-db/src/defs.rs
+++ b/crates/ide-db/src/defs.rs
@@ -10,7 +10,7 @@ use hir::{
     Adt, AsAssocItem, AssocItem, AttributeTemplate, BuiltinAttr, BuiltinType, Const, Crate,
     DefWithBody, DeriveHelper, DocLinkDef, ExternCrateDecl, Field, Function, GenericParam,
     HasVisibility, HirDisplay, Impl, Label, Local, Macro, Module, ModuleDef, Name, PathResolution,
-    Semantics, Static, ToolModule, Trait, TraitAlias, TypeAlias, Variant, Visibility,
+    Semantics, Static, ToolModule, Trait, TraitAlias, TypeAlias, Variant, VariantDef, Visibility,
 };
 use stdx::{format_to, impl_from};
 use syntax::{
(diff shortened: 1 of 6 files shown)


crates/ide-db/src/defs.rs
```diff
@@ -134,6 +143,132 @@ impl Definition {
...
+            let name = Some(assoc.name(db)?);
+            let item = trait_.items(db).into_iter().find(|it| it.name(db) == name)?;
+            item.docs(db)
+        })
+    }
+
+    pub fn label(
+        &self,
+        db: &RootDatabase,
+        famous_defs: Option<&FamousDefs<'_, '_>>,
```
Veykril, line 213: This can be removed fully, its only used for its `db` field but the `db` is already being passed here

Veykril:
One small thing, but other lgtm!

crates/ide-db/src/defs.rs
nicolas-guichard: Indeed, I removed it.

nicolas-guichard:
In addition to the change requested above, I've also rebased and added some tests for TraitAlias and TypeAlias which were not previously covered. I think the existing behavior for plain type aliases was unexpected (`type MyTypeAlias = u8;` yielded `[MyTypeAlias]` instead of `MyTypeAlias#`), but do tell me if you would rather revert to that.

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16199, opened 2023-12-26
fix: `extract_struct_from_enum_variant` assist should resolve Self generic arg

Young-Flash:
close https://github.com/rust-lang/rust-analyzer/issues/16197

### DIFF
--- a/crates/ide-assists/src/handlers/extract_struct_from_enum_variant.rs
+++ b/crates/ide-assists/src/handlers/extract_struct_from_enum_variant.rs
@@ -452,6 +452,27 @@ mod tests {
     use super::*;
 
     #[test]
+    fn issue_16197() {
+        check_assist(
+            extract_struct_from_enum_variant,
+            r#"
+enum Foo {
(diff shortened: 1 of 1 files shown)


Young-Flash:
sorry I miss the case that nested generic arg with `Self`, corrected now.

it's ready for review.

crates/ide-assists/src/handlers/extract_struct_from_enum_variant.rs
```diff
@@ -246,6 +246,29 @@ fn create_struct_def(
...
+                                    Some(ted::replace(it, type_arg.syntax()))
+                                } else {
+                                    None
+                                }
+                            })
+                            .count(),
+                    ),
+                    _ => None,
+                })
+                .count();
```
Veykril, line 270: This should use the `PathTransform::transform_path` function to transform the `Self` path

Veykril (changes requested):

Veykril (approved):

Veykril:
Thanks!
@bors r+

Issue #16053, opened 2023-12-08
Completion type match calculations for function items should consider the signature compatability


Veykril:

jmintb:
I am up for working on this :)

Veykril:
Okay, so the bigger picture here would be to have `type_match` consider coercions I think (which should fix this technically). But that seems a lot more tricky to get right, so instead we should just have `compute_type_match` consider `FnDef`s and the like specially which we can do via `Type::as_callable` I think. Roughly, we just need to check whether both types are "callables" and then match them according to the callables instead.

Veykril:
@jmintb I'm sorry, I just realized that was roughly what you were doing in your commit already (how embarassing) 😅 I think I misread something there yesterday when fixing up the PR. Can you just re-implement that part as a new PR (and maybe put the actual equivalence check logic behind a function in `Callable`?

jmintb:
No worries, I will do as you suggest :)

Veykril:
Fixed in https://github.com/rust-lang/rust-analyzer/pull/16136

PR #16205, opened 2023-12-28
Draft Callable type matching

jmintb:
A first attempt at #16053. The type matching is implemented on `Callable` as `Callable::is_unifiable`. We can not import from `ide-completions` and therefore do not have access to `CompletionRelevanceTypeMatch`. `Callable::is_unifiable` therefore returns a `bool` indicating if callables are unifiable or not. The idea is that an exact match is also unifiable. I am not sure what implications this limitation has.

This was what I could get away with, without refactoring too much. It might make sense to go back to the original solution of having a separate function in `render` which differentiates between unifable and exact.

Let me know what you think :)

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -4657,6 +4657,29 @@ impl Callable {
     pub fn return_type(&self) -> Type {
         self.ty.derived(self.sig.ret().clone())
     }
+
+    /// Compute if this `Callable` is unifiable with a `other`. Two callables are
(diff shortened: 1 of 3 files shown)


Veykril:
I think this got fixed with https://github.com/rust-lang/rust-analyzer/pull/16136/files actually (we have a signature check in there now that does the trick already), been going through the PR backlog and that PR made it in first, sorry :( Thanks for looking into this nevertheless!

(An aside, unifying the callable by unifying the parameters and return type would be incorrect, as signatures don't unify in the same way)

PR #16211, opened 2023-12-29
Use Cargo's [workspace.lints.*] to config clippy

tetsuharuohzeki:
This change begin to use [`[workspace.lints.*]`](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-lints-table) that is stabilized since [Rust 1.74](https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html).

By this change, we make the configure more readable and simplify `xargo lint` more.


Veykril:
Thanks!
@bors r+

PR #16213, opened 2023-12-29
internal: update docs/dev/guide.md based on 2024-01-01 release

davidsemakula:
Updates dev guide based on 2024-01-01 release

### DIFF
--- a/docs/dev/guide.md
+++ b/docs/dev/guide.md
@@ -65,11 +65,11 @@ Next, let's talk about what the inputs to the `Analysis` are, precisely.
 
 rust-analyzer never does any I/O itself, all inputs get passed explicitly via
 the `AnalysisHost::apply_change` method, which accepts a single argument, a
-`Change`. [`Change`] is a builder for a single change
+`AnalysisChange`. [`AnalysisChange`] is a builder for a single change
 "transaction", so it suffices to study its methods to understand all of the
 input data.
 
-[`Change`]: https://github.com/rust-lang/rust-analyzer/blob/master/crates/base_db/src/change.rs#L14-L89
+[`AnalysisChange`]: https://github.com/rust-lang/rust-analyzer/blob/guide-2019-01/crates/ra_ide_api/src/lib.rs#L119-L167
 
 The `(add|change|remove)_file` methods control the set of the input files, where
 each file has an integer id (`FileId`, picked by the client), text (`String`)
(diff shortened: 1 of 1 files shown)


lnicola:
So many broken links :confused:. As for `Change`, see https://github.com/rust-lang/rust-analyzer/pull/7806. We should probably use the more recent one, with a SHA in the URL. But I think it's been renamed again in the meanwhile to [`FileChange`](https://github.com/rust-lang/rust-analyzer/blob/35620306a6dc1e108f86e8891c9b3103397ca1e3/crates/base-db/src/change.rs#L36).

davidsemakula:
> We should probably use the more recent one, with a SHA in the URL. But I think it's been renamed again in the meanwhile to [`FileChange`](https://github.com/rust-lang/rust-analyzer/blob/35620306a6dc1e108f86e8891c9b3103397ca1e3/crates/base-db/src/change.rs#L36).

I initially stuck with `AnalysisChange` because the start of the guide says:

> This guide describes the current state of rust-analyzer as of 2019-01-20 (git tag [guide-2019-01](https://github.com/rust-lang/rust-analyzer/tree/guide-2019-01)).

If we change to the more recent `FileChange` type, I think we'll need to make a few more changes to the [Inputs](https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/guide.md#inputs) section (to describe the current API) and [Language Server Protocol](https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/guide.md#language-server-protocol) section (to describe and reference more recent implementations - `GlobalState` instead of `ServerWorldState`, `FileChange` and it's `Change` wrapper from `hir-expand` instead `AnalysisChange` e.t.c).

LMK if that's desirable for this PR🙂

lnicola:
> This guide mostly describes the current state of rust-analyzer as of 2019-01-20 (git tag [guide-2019-01](https://github.com/rust-lang/rust-analyzer/tree/guide-2019-01)).

There, fixed it 😂.

> LMK if that's desirable for this PR🙂

The crate names are all different too. Maybe it would be a good idea to overhaul it a little, but it's fine if you don't feel like doing all that work.

davidsemakula:
> Maybe it would be a good idea to overhaul it a little

I agree, I'll give it a shot 🙂

davidsemakula:
I've made some changes based on the 2024-01-01 release, I think this is the best I can do on my own 🙂

lnicola (approved):

lnicola:
Thanks, this is great!

@bors r+

PR #16223, opened 2024-01-01
feat: add quickfix for redundant_assoc_item diagnostic

Young-Flash:
Happy New Year 😊

follow up https://github.com/rust-lang/rust-analyzer/pull/15990, now it's time to close https://github.com/rust-lang/rust-analyzer/issues/15958, closes https://github.com/rust-lang/rust-analyzer/issues/16269

demo

EDIT: add a demo.git would be more illustrated when making release change log.

### DIFF
--- a/crates/ide-diagnostics/src/handlers/trait_impl_redundant_assoc_item.rs
+++ b/crates/ide-diagnostics/src/handlers/trait_impl_redundant_assoc_item.rs
@@ -1,11 +1,9 @@
 use hir::{db::ExpandDatabase, Const, Function, HasSource, HirDisplay, TypeAlias};
 use ide_db::{
     assists::{Assist, AssistId, AssistKind},
-    base_db::FileRange,
     label::Label,
     source_change::SourceChangeBuilder,
 };
-use syntax::{AstNode, SyntaxKind};
 use text_edit::TextRange;
 
 use crate::{Diagnostic, DiagnosticCode, DiagnosticsContext};
@@ -64,7 +62,7 @@ pub(crate) fn trait_impl_redundant_assoc_item(
     Diagnostic::new(
         DiagnosticCode::RustcHardError("E0407"),
         format!("{redundant_item_name} is not a member of trait `{trait_name}`"),
-        FileRange { file_id: d.file_id.file_id().unwrap(), range: diagnostic_range },
+        hir::InFile::new(d.file_id, diagnostic_range).original_node_file_range_rooted(db),
     )
     .with_fixes(quickfix_for_redundant_assoc_item(
         ctx,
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/trait_impl_redundant_assoc_item.rs
```diff
@@ -10,47 +17,196 @@ pub(crate) fn trait_impl_redundant_assoc_item(
...
+                    .unwrap_or(default_range),
+                format!("\n    type {};", type_alias.name(ctx.sema.db).to_smol_str()),
+            )
+        }
     };
 
     Diagnostic::new(
         DiagnosticCode::RustcHardError("E0407"),
         format!("{redundant_item_name} is not a member of trait `{trait_name}`"),
         FileRange { file_id: d.file_id.file_id().unwrap(), range: diagnostic_range },
```
Veykril, line 67: Oh I just realized I missed this on the review that added this diagnostic, but this unwrap is very bad 😅 We should be using `adjusted_display_range_new` to get the new range for this here.

crates/ide-diagnostics/src/handlers/trait_impl_redundant_assoc_item.rs
```diff
@@ -10,47 +17,196 @@ pub(crate) fn trait_impl_redundant_assoc_item(
...
+        let trait_def_crate = d.trait_.module(db).krate();
+        if trait_def_crate != current_crate {
+            return None;
+        }
+        let trait_def = d.trait_.source(db)?.value;
+        let where_to_insert = trait_def
+            .syntax()
+            .descendants_with_tokens()
+            .find(|it| it.kind() == SyntaxKind::L_CURLY)
+            .map(|it| it.text_range())?;
```
Veykril, line 98:
```suggestion
        let where_to_insert = trait_def
            .assoc_item_list()?
            .l_curly()?
            .text_range();
```

We also need to accomodate for macros here again though, as this range might come from within a macro expansion

crates/ide-diagnostics/src/handlers/trait_impl_redundant_assoc_item.rs
Young-Flash: `adjusted_display_range_new` require a `diag_ptr: InFile<AstPtr<N>>`, regarding get `diag_ptr` from the current `hir::AssocItem`, we could use  `hir::AssocItem.source(db)`, which return an `Option<InFile<Self::Ast>>`, could we `unwarp` or `expect` directly? since `hir::AssocItem.source(db)` actually return `Some(...)`

Veykril: Can also do something similar to `adjusted_display_range_new` since we are skipping the `diag_ptr`. And no unwrapping isn't an option here, the source fetching is technically fully infallible right now, but we might change things at some point where we allow loading libraries with missing sources at which point the unwrap would fail.

saiintbrisson: Wasn't aware this was already discussed here. Opened a issue #16269.

Veykril (approved):

Veykril:
Thanks!
@bors +

saiintbrisson:
@Veykril I guess you forgot the `r` in `r+`

PR #16231, opened 2024-01-02
internal: include backtraces in LSP responses

davidbarsky:
To make integrations with external instrumentation (namely: the lightweight LSP proxy my colleagues and I run at work) easier, I've changed rust-analyzer to return the backtrace, if enabled, in error responses, which saves us the effort of needing to collect logs and correlate editing sessions with individual logs.

### DIFF
--- a/crates/rust-analyzer/src/dispatch.rs
+++ b/crates/rust-analyzer/src/dispatch.rs
@@ -1,10 +1,10 @@
 //! See [RequestDispatcher].
-use std::{fmt, panic, thread};
+use std::{fmt, panic};
 
 use ide::Cancelled;
 use lsp_server::ExtractError;
 use serde::{de::DeserializeOwned, Serialize};
-use stdx::thread::ThreadIntent;
+use stdx::{panic_context, thread::ThreadIntent};
 
 use crate::{
     global_state::{GlobalState, GlobalStateSnapshot},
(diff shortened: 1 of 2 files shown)


Veykril (changes requested):
The implementation here won't quite work if there are nested panic contexts, since the innermost one will clear it before the outermost one (in the handler) is able to read it out. 

So we need to keep track of whether the dropped context is the last one or not when clearing the backtrace. Secondly, we can probably keep track of the backtrace in `with_ctx`, I don't think we need an extra function for that here (but we do still need a second thread local)

davidbarsky:
> The implementation here won't quite work if there are nested panic contexts, since the innermost one will clear it before the outermost one (in the handler) is able to read it out. 

Yeah, that's a good point. Will fix.

> So we need to keep track of whether the dropped context is the last one or not when clearing the backtrace. Secondly, we can probably keep track of the backtrace in `with_ctx`, I don't think we need an extra function for that here (but we do still need a second thread local)

Can make those adjustments, yeah. I think that that the thread local could maybe be a `Vec<(String, Backtrace)>`, where a Backtrace could be disabled. I'll need to double-check.

In any case, I don't feel _too_ strongly about landing this, as I was planning on deploying this in a patch for a few weeks to better attribute the existing panics that we [discussed on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/Common.20Panics.20in.20rust-analyzer). New ones, of course, could crop up, but given that some of the panics I've observed have been around for a few years, I dunno how valuable having this infrastructure would be.

PR #16234, opened 2024-01-03
fix: Fix focus range being discarded in attributes/derives when upmapping

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/16229


Veykril:
@bors r-
there is sync going on right now i guess

PR #16241, opened 2024-01-03
Give a userful error when rustc cannot be found in explicit sysroot

Nilstrieb:
Somehow r-a believed that my sysroot was something weird with no rustc. Probably a me issue, but it was impossible to diagnose since r-a just gave me a plain "No such file or directory". Adding this error makes it clear what happened and allows diagnosing the problem.


Veykril:
Thanks!
@bors r+

PR #16245, opened 2024-01-04
Trait Import completion for Notable Traits

dfireBird:
cc @Veykril  

`ImportMap` and `ImportInfo` are changed to record information related to `notable_traits`. Does the current implementation look okay? If so I can proceed with actual implementation of the completion.

P.S: sorry if I shouldn't ping

### DIFF
--- a/crates/hir-def/src/attr.rs
+++ b/crates/hir-def/src/attr.rs
@@ -207,6 +207,13 @@ impl Attrs {
         })
     }
 
+    pub fn has_doc_notable_trait(&self) -> bool {
+        self.by_key("doc").tt_values().any(|tt| {
+            tt.delimiter.kind == DelimiterKind::Parenthesis &&
+                matches!(&*tt.token_trees, [tt::TokenTree::Leaf(tt::Leaf::Ident(ident))] if ident.text == "notable_trait")
+        })
+    }
+
     pub fn doc_exprs(&self) -> impl Iterator<Item = DocExpr> + '_ {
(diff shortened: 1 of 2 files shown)


Veykril:
Yes that looks like roughly what i had in mind

dfireBird:
Well that was hard to fix the conflicts. Thanks for that now `search_dependencies` is very easily understandable.

Anyways I had some questions, we are currently storing `ItemInNs` of `TraitId` type, but the completion requires us to have `ItemInNs` of `FunctionId` type (and a associated item). I have searched for some function that gives all associated items of trait with `ItemInNs` but I couldn't be able to find any such function. So is that possible to do?

Veykril:
Gonna investigate whether we need to do this actually or not. We might not after all (but we can still make use of the notable trait stuff in another way, for completion priority)

Veykril:
Ye okay after reconsidering we don't need to do this hack, we can just remove the limit in play here. What we can do though is prioritize method ocmpletions that come from notable traits by adding an extra score field to https://github.com/veykril/rust-analyzer/blob/4f75e0f9108192a8d3e115e885cfaf220d1660b8/crates/ide-completion/src/item.rs#L131

Are you interested in doing that instead?

dfireBird:
Oh yes sure I can do that. Just learning more sections in RA :). 

So should I close this PR and recreate a new one?!

Veykril:
Yes lets close this, just ask any questions on zulip, I think you've been there before.

PR #16247, opened 2024-01-04
fix: Fix build scripts not being rebuilt in some occasions

Veykril:
Also makes proc-macro changed flag setting async, we don't wanna block `process_changes` on the database as that is on the main thread!


lnicola:
Needs a `cargo test; git commit -a --amend`.

crates/rust-analyzer/tests/slow-tests/main.rs
```diff
@@ -44,7 +44,7 @@ const PROFILE: &str = "";
 #[test]
 fn completes_items_from_standard_library() {
     if skip_slow_tests() {
-        return;
+        // return;
```
lnicola, line 47:
```suggestion
        return;
```

PR #16248, opened 2024-01-04
Add proc-macro-test crate back to the main workspace

Veykril:
Should fix https://github.com/rust-lang/rust-analyzer/issues/16194

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [workspace]
 members = ["xtask/", "lib/*", "crates/*"]
-exclude = ["crates/proc-macro-srv/proc-macro-test/"]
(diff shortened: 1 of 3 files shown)


lnicola (approved):

pksunkara:
I suspect that this is not the right fix, but I might be wrong. Let's wait and see.

lnicola:
```toml
[dev-dependencies]
proc-macro-test.path = "./proc-macro-test"

[features]
sysroot-abi = ["proc-macro-test/sysroot-abi"]
```

Do we have to remove this dev-dependency somehow?

Veykril:
I think we can drop the feature flag in there, its not like you can test the server without it anyways.

Veykril:
Actually its not that simple since the build script still needs to run on stable ...

PR #16249, opened 2024-01-04
internal: Sync from rust

lnicola:


crates/proc-macro-srv/proc-macro-test/imp/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 //! Exports a few trivial procedural macros for testing.
 
+#![allow(unexpected_cfgs)]
+#![cfg(feature = "sysroot-abi")]
```
lnicola, line 4: Oops, I don't think this does anything.

Issue #14268, opened 2023-03-06
Replace `check_expect` test functions that assert by debug equality


Veykril:
We have a bunch of test fixtures which currently assert outputs via the debug representation of things. These are annoying as the debug representation can change in many ways (even by just upgrading the compiler version which did happen in the past). We should get rid of those and replace them with better expected output.

An example of such a function is the following here https://github.com/rust-lang/rust-analyzer/blob/a360fab9a36a20046fed4c6e9b863c52c7578de0/crates/ide/src/inlay_hints.rs#L541-L546

Urhengulas:
The three PRs should solve this issue

PR #16252, opened 2024-01-04
internal: Switch to `expected.assert_eq` for `ide` tests

Urhengulas:
This PR switches from `assert_debug_eq` to `assert_eq` and only compares parts of the result and not the whole. The aim is to only compare parts which are relevant to the test and also make it more readable.

Part of #14268.

## Questions
- [x] Can I use `Vec`? If not, what is the alternative?
    I assume I cannot because of: https://github.com/rust-lang/rust-analyzer/blob/c3a00b5468576de4e39adc8fa5ceae35a0024e49/docs/dev/architecture.md?plain=1#L413
- [x] Should I group it by file, as proposed by Lukas?
    ```
    file_id 1:
        source_file_edits: 
            - Indel { insert: "foo2", delete: 4..7 }

    file_id 2:
        file_system_edits:
            MoveFile AnchoredPathBuf { anchor: FileId(2), path: "foo2.rs", }
    ```
- [x] Is it okay to ignore `CreateFile` events? They do not have a FileId, which would be problematic, but they do not occur in the existing tests, so I marked them as `unreachable!()` so far.

### DIFF
--- a/crates/ide/src/rename.rs
+++ b/crates/ide/src/rename.rs
@@ -366,7 +366,7 @@ fn text_edit_from_self_param(self_param: &ast::SelfParam, new_name: &str) -> Opt
 #[cfg(test)]
 mod tests {
     use expect_test::{expect, Expect};
-    use ide_db::source_change::{FileSystemEdit, SourceChange};
+    use ide_db::source_change::SourceChange;
     use stdx::trim_indent;
     use test_utils::assert_eq_text;
     use text_edit::TextEdit;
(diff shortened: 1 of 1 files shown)


Veykril:
>  Can I use `Vec`? If not, what is the alternative?

This is meant for test fixtures (the code that is passed to tests as strings), those dont see std library stuff. Testing infra is free to use whatever.

> Should I group it by file, as proposed by Lukas? 

On second though thats probably not worth it given the number of files changed is usually 1 anyways

Veykril (approved):

Issue #15888, opened 2023-11-14
Using a different partitoning in `base_db::input::SourceRoot`


alibektas:
:wave: I have been working on #13529 and (indirectly) on #15741 ( as a matter of fact there should have been a "toml support for rust-analyzer issue". In order for both of these features to be implemented there needs to be a map ( and its inv. )  that maps `SourceFile`s ( so essentially a `FileId` ) to its corresponding (most of the time the closest) `Cargo.toml`/`RA.toml` file. I believe that this should be handled when we get to partition a `Vfs` into `SourceRoot`s. Based on my poor observation we partition with the intention to group as many files as possible into the same `FileSet`. I want to change that in such a way that `Cargo.toml`/`RATOML` files are used as anchors to a `SourceRoot` , meaning that if you encounter , say , a `Cargo.toml` you would know that this is a new `SourceRoot`. This change would enable me to map a `FileId` to the `FileId` of its corresponding `SourceRoot`, as in

```rust
srid = SourceDatabaseExt::file_source_root(db, file_id)
sr = SourceDatabaseExt::source_root(db, srid)
cargo_file_id = sr.cargo_toml_id()  // Option<FileId>
```  

I cannot see any other way of doing this so I wanted to get community's opinion on this.

Veykril:
Can't you just store the `FileId` of the cargo.toml inside the `SourceRoot`? That aside, why do you need such a mapping?

Veykril:
> Can't you just store the `FileId` of the cargo.toml inside the `SourceRoot`?

Answer is yes after some discussion in DMs

PR #16254, opened 2024-01-04
rust-analyzer.toml tracking

alibektas:
Here you can see how far we have gotten with our planned `.rust-analyer.toml` ( aka `RATOML` configuration file ). The related issue is #13529. Our roadmap is roughly : ( this is a modified version of [this](https://github.com/rust-lang/rust-analyzer/issues/13529#issuecomment-1724116909) comment.

# Phase 1 :

- [x] Find out which configs are globally or locally or client-scoped. Create a PR where we possibly publish a document in which we briefly mention why a certain key-value pair is global or 
>  So to make declaring a config as `local`/`global`/`client` simple I separated them into different configs. I am sure that I misplaced one or two configs here and there so we can always amend them. 

- [x] Replace rust_analyzer::config::ConfigData with GlobalConfigData LocalConfigData. As a result of this change rust_analyzer::config::Config's data will be of type GlobalConfigData. 

- [x]  Integrating a deserializer : For this issue, we will use the simplest possible. When #15908 becomes available we can migrate to it and start showing errors and auto-completion to the user. Many thanks to @cormacrelf for taking my ugly work and making wonders out of it!  

- [ ] project-model and all that lsp stuff : How do we get notified when a `toml` file get changed? How do we map a file to its corresponding *RATOML* file?

> Some two months ago : (**REJECTED**) project_model::ProjectManifest is a good place to handle the discovery of the .rust-analyzer.toml file and finding out which configurations apply to a single ProjectManifest. This change will probably entail adding two more variants to ProjectManifest like ProjectManifest::ProjectJsonWithConf and ProjectManifest::CargoTomlWithConf. As I said before names are at this point arbitrary.

> 04.01.24 : This approach would make things more complicated than they already are. We will instead try making some changes to `base-db::input::SourceRoot`.

> 05.01.24 : Making changes to `SourceRoot` ( first discussed under #15888) so that we can express things like *Fetch the related RATOML for this file* has the same design flaw as 16170. So we need to store this information elsewhere. And as a side note, before even considering this option I thought about using a new database where I would map `FileId` ( of a source file ) to a `FileId` ( of a RATOML file ). This obviously would cause much redundancy as there needs to be one entry for each source file. Maybe we could consider using storing the information we want using a mapping of `SourceRootId` -> `FileId` sort.

- [ ] Client settings and RATOML : During first handshakes between the server and the client config data are sent from the client to the server, which is how we have received configuration data so far. I plan to keep this as is and let server do a fresh start if any RATOML files are discovered. But this shouldn't tell you that the configs sent from the client have a higher priority than RATOML. Although I agree with Veykril on the point that our hand is forced to choose either RATOML or client settings over another, the user should still be able to say that they want to use their own configs, which means that we need to add a KV pair that does just that : telling us which source we should prefer.

- [ ] First version of the RATOML : Although I may have sounded like this is something I want to have ready in the first version, I think we should first enable the use of a single RATOML file and gradually introduce nesting. So this version will use two sources to read configs from : (1) lsp_types::InitializeParams as sent by the client (2) RATOML. 

# Phase 2 :
- [ ]  Introduce nesting
- [ ]  Migrate to taplo ( #15908 ). Type annotate each config entry


cormacrelf:
I kept going on config.rs. https://github.com/alibektas/rust-analyzer/pull/1 targets this PR branch.

cormacrelf:
@alibektas hey, what's the plan for config inheritance? If I understand correctly, a `local` config is roughly per-crate, hence things like `imports_prefer_no_std` being local. You would still be able to have a toml at your workspace root that sets these things for the workspace, and then individual crates will override them.

At the moment the deserializer does not track whether a field was set, and collapses straight to the default value. The macro could generate a second struct that has every field `Option<T>` and then a method to perform the cascade. If it did that, then overrides would be at the field level. You would not have e.g. "reset the [imports.granularity] section", only "set imports_granularity_enforce".

Do you want that functionality in the configdata structs, or is it easier to handle this at the level of toml tables / serde_json::Value?

alibektas:
> @alibektas hey, what's the plan for config inheritance? If I understand correctly, a `local` config is roughly per-crate, hence things like `imports_prefer_no_std` being local. You would still be able to have a toml at your workspace root that sets these things for the workspace, and then individual crates will override them.
> 
> At the moment the deserializer does not track whether a field was set, and collapses straight to the default value. The macro could generate a second struct that has every field `Option<T>` and then a method to perform the cascade. If it did that, then overrides would be at the field level. You would not have e.g. "reset the [imports.granularity] section", only "set imports_granularity_enforce".
> 
> Do you want that functionality in the configdata structs, or is it easier to handle this at the level of toml tables / serde_json::Value?

First, to answer your question: Yes I would rather having Option<T> for each config field and not solve this problem during Deserialization. The reason why I have structs like `RootClobalConfigData` etc. was exactly because I had this problem in mind. 

Files are partitioned by the `base-db::input::SourceRoot` into `FileSet`s whose partitioning mechanism is sometimes not ideal for this use case but I was told to stick with it. It more or less partitions files into crates so even though I primarily wanted to let users define a `rust-analyzer.toml` wherever they want within the project however based on how the `SourceRoot` paritioning works I think we will be limited to `crate` level. 

There are other  inheritance issues such as :

1. What if user wants to make use of a `.rust-analyzer.toml` file from `$XDG_CONFIG_HOME/rust-analyzer/.rust-analyzer.toml`?
2. Should `project_root/.rust-analyzer.toml` have precendence over whatever comes as config from the editor? My plan is to add a `client` level config `preferEditorConfig` or sth of that sort and let user decide.

cormacrelf:
Cool. I may as well add that to the macro then.

cormacrelf:
I have thought about the inheritance some more.

If we add a concept of ProjectRoot, ie a FileSet that contains a bunch of SourceRoots, then I think you can actually walk back up the file system from a given crate/SourceRoot to the ProjectRoot and construct an inheritance tree that way. You get your project roots from discovering a workspace + any linkedProjects. You partition the SourceRoots into ProjectRoots much the same way we partition files into SourceRoots. And then for each source root you walk up to its project root and insert under the first rust-analyzer.toml you find along the way. This allows us to find rust-analyzer.tomls in parent directories that aren't SourceRoots themselves.

Any SourceRoots that aren't contained in any project root (eg external crates in your $HOME/.cargo) are just independent roots in the "config forest". Of course everything inherits from RootLocalConfigData in the end, so it's still figuratively a tree.

We can probably allocate FileIds for all the parent directories' TOMLs, including the TOMLs that don't exist, which might make tree construction and modification a bit simpler -- you wouldn't need to insert a new FileId somewhere in a new middle layer of the tree when it is created. And when resolving a config given some subtrees being dirty after a vis change rolls in, the recursive resolver can just clone the Arc of the parent node if there is no actual LocalConfigInput at that level due to the file not actually existing. The resolver could be a salsa query, but I kinda just wrote one with an indextree and an RwLock.

alibektas:
I want to experiment with `SourceRoot` a little more before I give my answer, so I will write back probably in 2 days but let me say this: I want to see how far we can go with what we have at our disposal before considering implementing an alternative representation like `ProjectRoot`,  how preferable it may seem. But I liked the idea of having dummy nodes and having a balanced tree. 

I think there is one important thing to keep in mind if we ever decide to go with something like `ProjectRoot` which is that project discovery is not done recursively but rather for a limited number of steps of walking up and down the tree and I suspect that the maintainers have been explicitly avoiding changing this, which tells me that we wouldn't also be able to do deep project discovery to locate where `ratoml` files are. If that's the case `ProjectRoot` may not be possible.

davidbarsky:
> I think we should first enable the use of a single RATOML file and gradually introduce nesting

@alibektas You're the one working on this, but I'd be happy to send a PR adding support for something like [figment](https://crates.io/crates/figment) or [config](https://crates.io/crates/config).

> 1. What if user wants to make use of a `.rust-analyzer.toml` file from `$XDG_CONFIG_HOME/rust-analyzer/.rust-analyzer.toml`?
> 2. Should `project_root/.rust-analyzer.toml` have precendence over whatever comes as config from the editor? My plan is to add a `client` level config `preferEditorConfig` or sth of that sort and let user decide.

 Speaking as an _extremely_ interested party (I'd like to use this in https://github.com/rust-lang/rust-analyzer/issues/16377, for one!) I think the answer for the first question is to follow what Cargo does and prefer the more local/precise version of the `rust-analyzer.toml` over the `$XDG_CONFIG_HOME`'d file. Similarly, I don't think that `project_root/.rust-analyzer.toml` should have precedence over editor-specified configuration since the more local/precise configuration would be coming from the users's editor, _not_ the cloned repo. This would mean, of course, that editors should avoid sending default configuration values as part of the `InitializeParams` and instead rely on the rust-analyzer binary knowing the default configuration natively.

> I think there is one important thing to keep in mind if we ever decide to go with something like ProjectRoot which is that project discovery is not done recursively but rather for a limited number of steps of walking up and down the tree and I suspect that the maintainers have been explicitly avoiding changing this, which tells me that we wouldn't also be able to do deep project discovery to locate where ratoml files are. If that's the case ProjectRoot may not be possible.

I don't recall the specific behavior in the case of Cargo, but I think that walking up the directory tree is _substantially_ less problematic than searching child directories. The former consists of calling [`std::path::Path::ancestors`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.ancestors) and joining each result with `rust-analyzer.toml`, while the latter might result in an unbounded search of all child directories for the appropriate file(s). Put differently, checking parent directories is `O(h)`, where `h` is the height of the directory tree from the current location, whereas checking children is `O(n)`, where `n` is potentially every file downtree [^1] of the current location.

[^1]: for instance, my employer's monorepo—if fully materialized on disk—is greater than one terrabyte. Searching child directories is (correctly, IMO) nonviable and should remain that way.

cormacrelf:
> @davidbarsky : I'd be happy to send a PR adding support for something like [figment](https://crates.io/crates/figment) or [config](https://crates.io/crates/config).

See https://github.com/alibektas/rust-analyzer/pull/2 for inheritable config, and https://github.com/alibektas/rust-analyzer/pull/3 for a silly implementation of actually inheriting in a tree shape. What libraries like figment/config won't do is update when files change on disk. I also believe the plan is for the configuration to auto-update simply because the files are in the Vfs and get watched, so we do need a way to incrementally update the config. (My silly implementation should have just used salsa for that, but did not.) There is also a lot of custom-ness (aliases, tables to underscores, schema generation, ...) baked onto the json and by extension toml deserializers. 

> Similarly, I don't think that project_root/.rust-analyzer.toml should have precedence over editor-specified configuration

Agree, because the one in XDG_CONFIG_HOME will do the job of setting _defaults_, and does it way better. It will apply to all editors, for example. That leaves a natural job for client configs to override per-editor. So once this stuff lands most people can move what's in their init.lua/vscode config to XDG at their leisure, and eventually only use the client config for client-specific things like "don't insert close parens". No rush, no breaking changes.

> The former consists of calling [std::path::Path::ancestors](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.ancestors) and joining each result with rust-analyzer.toml

That's what I had in mind.

davidbarsky:
> > @davidbarsky : I'd be happy to send a PR adding support for something like [figment](https://crates.io/crates/figment) or [config](https://crates.io/crates/config).
> 
> 
> 
> See https://github.com/alibektas/rust-analyzer/pull/2 for inheritable config, and https://github.com/alibektas/rust-analyzer/pull/3 for a silly implementation of actually inheriting in a tree shape. What libraries like figment/config won't do is update when files change on disk. I also believe the plan is for the configuration to auto-update simply because the files are in the Vfs and get watched, so we do need a way to incrementally update the config. (My silly implementation should have just used salsa for that, but did not.) There is also a lot of custom-ness (aliases, tables to underscores, schema generation, ...) baked onto the json and by extension toml deserializers. 

Oh duh, you're absolutely right. I didn't consider that those files would change over time!

> 
> > Similarly, I don't think that project_root/.rust-analyzer.toml should have precedence over editor-specified configuration
> 
> 
> 
> Agree, because the one in XDG_CONFIG_HOME will do the job of setting _defaults_, and does it way better. It will apply to all editors, for example. That leaves a natural job for client configs to override per-editor. So once this stuff lands most people can move what's in their init.lua/vscode config to XDG at their leisure, and eventually only use the client config for client-specific things like "don't insert close parens". No rush, no breaking changes.
> 
> 
> 
> > The former consists of calling [std::path::Path::ancestors](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.ancestors) and joining each result with rust-analyzer.toml
> 
> 
> 
> That's what I had in mind.

Glad to hear it!

crates/rust-analyzer/src/config.rs
```diff
@@ -2965,7 +3021,7 @@ mod tests {
     #[test]
     fn generate_config_documentation() {
         let docs_path = project_root().join("docs/user/generated_config.adoc");
-        let expected = ConfigData::manual();
+        let expected = ConfigInput::manual();
```
davidbarsky, line 3024: small nit: there's been a pattern of using `Data` inside of rust-analyzer for things that implement `serde::Deserialize`, and while I don't see the larger picture that this branch is painting, I think it might better to keep this suffixed with `Data` instead of `Input`.

cormacrelf: There already is a ConfigData. What should this one be called, given it contains an Option version of every field in ConfigData?

davidbarsky: Ah, I'm sorry. I assumed that `ConfigInput` _was_ `ConfigData`; I didn't realize it was a separate struct with `Options` for fields.

Disregard the nit then!

alibektas:
:wave: Sorry for being inactive for the last couple of days.

There are certain issues that I want to draw your attention to 

- Problem with absolute paths : As the dev docs of ra states that paths are just implementation details. We need to be developing our solution on the interned Ids. So once we have interned a file, we shouldn't really be converting it back to do path manipulation to find the nearest ratoml file. The reason I am mentioning is that if I got the intention right, this should invalidate mapping files to their corresponding `ratoml`s ( given that we don't want to map a source file to `ratoml` by keeping a mapping , once again remember @davidbarsky's 1tb monorepo scenario ).  The alternative is walking up the filesystem, "interning" every folder with a newtype'd id, and storing this in an mapping, just like the `FileId`/`SourceRootId` relation but more fine-grained.  

- Salsa's role : Well, at this point it is premature. On the one hand we know that we will need incremental parsing at some point to enable autocompletion for toml files ( `ratoml`/`Cargo.toml` even regular `toml` files, afaik we can't really exclude non-ratoml/cargo.toml files from not being watched, let's hope that I am wrong ) but it should never be the case that the `config-tree` ,as @cormacrelf calls it, get updated every time salsa db gets and update. So on the materialization level configurations are bounded by the Vfs changes and on the autocompletion level by salsa. 

- Changing some configs means it's time to restart the server but we didn't really do anything to handle this on the config level. In VSCode one would get notified when it's time to restart the server but this message isn't really coming from the `rust-analyzer::config`. We can improve on this.

alibektas:
:wave: Sorry for being inactive for the last couple of days.

There are certain issues that I want to draw your attention to 

- Problem with absolute paths : As the dev docs states paths are just implementation details. We need to be developing our solution on the interned Ids. So once we have interned a file, we shouldn't really be converting it back to do path manipulation to find the nearest ratoml file. The reason I am mentioning is that if I got the intention right, this should invalidate the approach mapping files to their corresponding `ratoml`s ( given that we don't want to map a source file to `ratoml` by keeping a mapping , once again remember @davidbarsky's 1tb monorepo scenario ). 

- Salsa's role : Well, at this point it is premature. On the one hand we know that we will need incremental parsing at some point to enable autocompletion for toml files ( `ratoml`/`Cargo.toml` even regular `toml` files, afaik we can't really exclude non-ratoml/cargo.toml files from not being watched, let's hope that I am wrong ) but it should never be the case that the `config-tree` ,as @cormacrelf calls it, get updated every time salsa db gets and update. So on the materialization level configurations are bounded by the Vfs changes and on the autocompletion level by a SalsaDB that as frequently gets updated as the RootDatabase if that makes sense. 

- Changing some configs means it's time to restart the server but we didn't really do anything to handle this on the config level. In VSCode one would get notified when it's time to restart the server but this message isn't really coming from the `rust-analyzer::config`. We can improve on this.

davidbarsky:
> 👋 Sorry for being inactive for the last couple of days.

No worries, sorry for the delay in repsonding!

> Problem with absolute paths : As the dev docs states paths are just implementation details. We need to be developing our solution on the interned Ids. So once we have interned a file, we shouldn't really be converting it back to do path manipulation to find the nearest ratoml file. The reason I am mentioning is that if I got the intention right, this should invalidate the approach mapping files to their corresponding `ratomls` ( given that we don't want to map a source file to `ratoml` by keeping a mapping , once again remember @davidbarsky's 1tb monorepo scenario).

I'm not sure that _conversions_ after the fact are strictly necessary—I haven't thought about this enough, sorry!—but it seems like `rust-analyzer.toml`'s _will_ need to make use of the paths to determine their respective ordering, correct?

alibektas:
> > 👋 Sorry for being inactive for the last couple of days.
> 
> No worries, sorry for the delay in repsonding!
> 
> > Problem with absolute paths : As the dev docs states paths are just implementation details. We need to be developing our solution on the interned Ids. So once we have interned a file, we shouldn't really be converting it back to do path manipulation to find the nearest ratoml file. The reason I am mentioning is that if I got the intention right, this should invalidate the approach mapping files to their corresponding `ratomls` ( given that we don't want to map a source file to `ratoml` by keeping a mapping , once again remember @davidbarsky's 1tb monorepo scenario).
> 
> I'm not sure that _conversions_ after the fact are strictly necessary—I haven't thought about this enough, sorry!—but it seems like `rust-analyzer.toml`'s _will_ need to make use of the paths to determine their respective ordering, correct?

Oh I forgot to write my follow up comment. Current strategy is the following and this makes my previous comment kinda obsolete.

1. We are notified by Vfs that there is a new ratoml file.
2. We get its ancestors all the way up to  the corresponding workspace root.
3. For all the `ratoml`s that don't exist, the current implementation ask for FileIds as if they had been registered in Vfs. But this I think we will not have in the final version. Instead we will have them stored in a separate db and assign them a `PlaceholderId` until they actually exist.
4. This way the finding out the closest ratoml for a file id is to ask vfs for its path take the path itself and gets its dirname and looking whether it currently exists or not, depending on which we get either a FileId or PlaceholderId and so on.

I think in 2-3 weeks we will have a prototype ready. The most complex example I have been thinking of testing my approach with is a scenario with linked projects and I gotta be honest. I have never dealt with `rust-project.json`. So I would appreciate if you could let me know if there is a certain thing unique to it that you want to see in ratoml, so that I can have it ready for the prototype.

davidbarsky:
> For all the ratomls that don't exist, the current implementation ask for FileIds as if they had been registered in Vfs. But this I think we will not have in the final version. Instead we will have them stored in a separate db and assign them a PlaceholderId until they actually exist.

To confirm my understanding: that separate database will have `PlaceholderId`s corresponding to non-existent files (e.g., `.rust-analyzer.toml`, `/a/foo.rust-analyzer.toml`, `/a/b/foo.rust-analyzer.toml`) until a file is created and they are removed from that database?

> The most complex example I have been thinking of testing my approach with is a scenario with linked projects and I gotta be honest. I have never dealt with rust-project.json. So I would appreciate if you could let me know if there is a certain thing unique to it that you want to see in ratoml, so that I can have it ready for the prototype.

Nothing at this time, but I'll give it a spin when it's ready!

PR #16256, opened 2024-01-04
internal: Only compare relevant parts in  `ide::{runnables,inlay_hints}` tests

Urhengulas:
This PR limits the data being compared. Therefore the tests should be more readable, as well as being more robust to changes to the data structure.

Part of https://github.com/rust-lang/rust-analyzer/issues/14268.

### DIFF
--- a/crates/ide/src/runnables.rs
+++ b/crates/ide/src/runnables.rs
@@ -576,8 +576,14 @@ mod tests {
     fn check_tests(ra_fixture: &str, expect: Expect) {
         let (analysis, position) = fixture::position(ra_fixture);
         let tests = analysis.related_tests(position, None).unwrap();
-        let test_names = tests.into_iter().map(|a| a.nav.name).collect::<Vec<_>>();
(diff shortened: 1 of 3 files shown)


crates/ide/src/runnables.rs
```diff
@@ -1503,23 +1301,7 @@ fn test_foo1() {}
...
                         },
-                        kind: Test {
-                            test_id: Path(
-                                "test_foo1",
-                            ),
-                            attr: TestAttr {
-                                ignore: false,
-                            },
-                        },
-                        cfg: Some(
```
Veykril, line 1514: This is interesting data (as long as its `Some`), same for when `use_name_in_title` is `true`. Would be good to render those conditionally (or otherwise keep them in the output in some way)

Urhengulas: Sure, gonna add that.

lnicola (approved):

Veykril:
Thanks!
@bors r+

PR #16265, opened 2024-01-05
fix: Acknowledge `pub(crate)` imports in import suggestions

Patryk27:
rust-analyzer has logic that discounts suggesting `use`s for private imports, but that logic is unnecessarily strict - for instance given this code:

```rust
mod foo {
    pub struct Foo;
}

pub(crate) use self::foo::*;

mod bar {
    fn main() {
        Foo$0;
    }
}
```

... RA will suggest to add `use crate::foo::Foo;`, which not only makes the code overly verbose (especially in larger code bases), but also is disjoint with what rustc itself suggests.

This commit adjusts the logic, so that `pub(crate)` imports are taken into account when generating the suggestions; considering rustc's behavior, I think this change doesn't warrant any extra configuration flag.

Note that this is my first commit to RA, so I guess the approach taken here might be suboptimal - certainly feels somewhat hacky, maybe there's some better way of finding out the optimal import path 😅

### DIFF
--- a/crates/hir-def/src/find_path.rs
+++ b/crates/hir-def/src/find_path.rs
@@ -551,7 +551,18 @@ fn find_local_import_locations(
         if let Some((name, vis)) = data.scope.name_of(item) {
             if vis.is_visible_from(db, from) {
                 let is_private = match vis {
-                    Visibility::Module(private_to) => private_to.local_id == module.local_id,
+                    Visibility::Module(private_mod, private_vis) => {
+                        if private_mod == def_map.module_id(DefMap::ROOT)
+                            && private_vis.is_explicit()
+                        {
+                            // Treat `pub(crate)` imports as non-private, so
+                            // that we suggest adding `use crate::Foo;` instead
+                            // of `use crate::foo::Foo;` etc.
+                            false
+                        } else {
+                            private_mod.local_id == module.local_id
+                        }
+                    }
                     Visibility::Public => false,
                 };
                 let is_original_def = match item.as_module_def_id() {
@@ -1018,6 +1029,24 @@ $0
             "crate::bar::S",
             "crate::bar::S",
             "crate::bar::S",
+        );
+    }
+
+    #[test]
+    fn promote_pub_crate_imports() {
+        check_found_path(
+            r#"
+//- /main.rs
+mod foo;
+pub mod bar { pub struct S; }
+pub(crate) use bar::S;
+//- /foo.rs
+$0
+        "#,
+            "crate::S",
+            "crate::S",
+            "crate::S",
+            "crate::S",
         );
     }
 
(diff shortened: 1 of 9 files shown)


crates/hir-def/src/find_path.rs
```diff
@@ -551,7 +551,18 @@ fn find_local_import_locations(
         if let Some((name, vis)) = data.scope.name_of(item) {
             if vis.is_visible_from(db, from) {
                 let is_private = match vis {
-                    Visibility::Module(private_to) => private_to.local_id == module.local_id,
+                    Visibility::Module(private_mod, private_meta) => {
+                        if private_mod == def_map.module_id(DefMap::ROOT)
```
Veykril, line 555: Shouldn't it be enough to just check this? I don't see why we need the `is_pub_crate`. If the `private_mod` here is a crate root module then it already has to be `pub(crate)` or something that has t he same semantic meaning.

Veykril: Ah I see, yes that would definitely be unexpected. That deserves an extra test to check that we aren't doing that for that matter! (assuming there is no such test yet).

crates/hir-def/src/visibility.rs
```diff
@@ -94,15 +94,15 @@ impl RawVisibility {
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub enum Visibility {
     /// Visibility is restricted to a certain module.
-    Module(ModuleId),
+    Module(ModuleId, ModuleVisibilityMeta),
```
Veykril, line 97: Let's swap that for
```suggestion
    Module(ModuleId, VisibilityExplicity),
```
and
```rs
enum VisibilityExplicity {
    Explicit,
    Implicit,
}
```
Whether the import is `pub(crate)` or not can be inferred from the flag being `Explicit` and the `ModuleId` being a root module of the same crate. That is we can be more general here.

Veykril: Alternatively
```suggestion
    Module { module: ModuleId, is_explicit: bool },
```
choose whatever you prefer

crates/hir-def/src/find_path.rs
Veykril: Although, what is rustc's behavior in regards to that. Does it also only acknowledge `pub(crate)` re-exports for this?

Patryk27: rustc _does_ suggest `use crate::Foo;`, but I'm not quite sold on it 😅

Veykril: Ye me neither, so let's stick to checking whether the visibility is explicitly provided.

Patryk27:
Ready!

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16271, opened 2024-01-05
Remove sysroot-abi feature flag from proc-macro-test

Veykril:
Closes https://github.com/rust-lang/rust-analyzer/issues/16194

### DIFF
--- a/crates/proc-macro-srv/proc-macro-test/build.rs
+++ b/crates/proc-macro-srv/proc-macro-test/build.rs
@@ -17,10 +17,23 @@ use cargo_metadata::Message;
 
 fn main() {
     println!("cargo:rerun-if-changed=imp");
-    println!("cargo:rerun-if-env-changed=PROC_MACRO_TEST_TOOLCHAIN");
+
+    let has_features = env::var_os("RUSTC_BOOTSTRAP").is_some()
+        || String::from_utf8(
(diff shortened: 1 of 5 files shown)


crates/proc-macro-srv/proc-macro-test/build.rs
```diff
@@ -17,7 +17,6 @@ use cargo_metadata::Message;
 
 fn main() {
     println!("cargo:rerun-if-changed=imp");
-    println!("cargo:rerun-if-env-changed=PROC_MACRO_TEST_TOOLCHAIN");
```
lnicola, line 20: Don't we want this?

Veykril: do we? I checked our CI and bootstrap in rustc, neither of which use this.

Veykril:
Hmm, this doesn't work as easily as I wanted to ...

crates/proc-macro-srv/proc-macro-test/build.rs
lnicola: The build script uses it, so it will catch whoever tries to change it (you?) off-guard.

Veykril: Well no, I removed it from the build script altogether

lnicola: Oh, sorry, I misread the diff.

PR #16274, opened 2024-01-06
Add notable_trait predicate to `CompletionRelevance`

dfireBird:
Given a score of 1 for now, will change as per reviews needed.


crates/ide-completion/src/render/function.rs
```diff
@@ -79,6 +79,11 @@ fn render(
         .and_then(|trait_| trait_.containing_trait_or_trait_impl(ctx.db()))
         .map_or(false, |trait_| completion.is_ops_trait(trait_));
 
+    let is_item_from_notable_trait = func
+        .as_assoc_item(ctx.db())
```
Veykril, line 83: Let's bind the `as_assoc_item` result to var and re-use that as multiple things in this function are interested in it (and the call is not completely free as it accesses the database)

Veykril (commented):
Let's add a test to `crates\ide-completion\src\render.rs` that checks that the relevance fetching works

crates/ide-completion/src/render/function.rs
dfireBird: Is it okay, if bind the result of the `and_then` as well, since seems like that's call to the db as well? I assume both methods (`containing_trait` and `containing_trait_or_trait_impl`) return the needed `Trait`.

Also is `ctx.db()` same as the `db` in the parameter of the render function?

Veykril: The `db` is the same ye. You can (and actually want) to use `containing_trait_or_trait_impl` for both here, I missed that actually (since the attribuite is only on the trait, not trait impls).

dfireBird: I thought the same too but just now saw the implementation and figured to ask the question.

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16275, opened 2024-01-06
fix: Fix `ast::Path::segments` implementation

davidsemakula:
calling `ast::Path::segments` on a qualifier currently returns all the segments of the top path instead of just the segments of the qualifier.

The issue can be summarized by the simple failing test below:
```rust
#[test]
fn path_segments() {
    //use ra_ap_syntax::ast;
    let path: ast::Path = ...; // e.g. `ast::Path` for "foo::bar::item".

    let path_segments: Vec<_> = path.segments().collect();
    let qualifier_segments: Vec<_> = path.qualifier().unwrap().segments().collect();
    assert_eq!(path_segments.len(), qualifier_segments.len() + 1); // Fails because `LHS = RHS`.
}
```

This PR:
- Fixes the implementation of `ast::Path::segments`
- Fixes `ast::Path::segments` callers that either implicitly relied on behavior of previous implementation or exhibited other "wrong" behavior directly related to the result of `ast::Path::segments` (all callers have been reviewed, only one required modification)
- Removes unnecessary (and now unused) `ast::Path::segments` alternatives


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16299, opened 2024-01-07
fix: Fix mir evaluator not respecting block local trait impls

Veykril:


crates/hir-ty/src/mir.rs
```diff
@@ -1016,6 +1016,8 @@ pub enum Rvalue {
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum StatementKind {
+    TraitEnvBlockEnter(hir_def::BlockId),
+    TraitEnvBlockExit,
```
Veykril, line 1020: Not too happy about having to add extra statements for this, but I can't think of a better way to encode this.

flodiebold: Sorry to be making comments from the sidelines since I don't have the whole context, but it's kind of surprising to me that there's any trait solving / method resolution happening during MIR eval at all :thinking:

Veykril: It is surprising, but the reason for that is that we currently try to resolve the concrete method of a trait function "late". That is inference result only resolves the trait function, but MIR later wants the function of the actual trait implementation which happens in mir lowering/eval currrently (and in the IDE layer for goto def etc). I guess we should be able to move that logic into type check (and get rid of the `lookup_impl_method_query`).

flodiebold: Can we do it during MIR lowering?

HKalbasi: I don't think we can get rid of the `lookup_impl_method_query` in MIR eval completely without changing the evaulation of `dyn Trait` which currently only stores the type id in the metadata part, and resolve the impl method on demand. Though I think we can switch to a more direct vtable representation, and then do all the method resolutions in the monomorphization.

Does the current implementation handle block local trait impls for `dyn Trait`? For example:
```Rust
trait Foo {
    fn foo(&self);
}

fn some_dyn_foo() -> Box<dyn Foo> {
    struct St;
    impl Foo for St { fn foo(&self) { println!("foo"); } }
    Box::new(St)
}

fn main() {
    let x = some_dyn_foo();
    x.foo();
}
```
If not, is it possible to handle it in future under current architecture?

flodiebold: > Though I think we can switch to a more direct vtable representation, and then do all the method resolutions in the monomorphization.

Yes, IMO that's what we should do.

PR #16300, opened 2024-01-07
feat: Add debugging command to show all loaded files and their in-memory size

Veykril:


lnicola:
Hmm, any surprises in there?

Veykril:
Yes -> https://github.com/rust-lang/rust-analyzer/issues/16301

Veykril:
Wont be useful anymore once I implemented the VFS changes I have in mind

lnicola:
I still think this could come in handy at times.

Veykril:
Well, not really. My changes are going to get rid of the in memory store fully. Turns out we only need to keep the opened docs in memory, The VFS really only needs to track the file change events, in which case modify and create events can keep the new text around until its uploaded to salsa. So the VFS won't keep any files in memory for longer than a process-change loop. I mean I guess we can keep the command to just dump all the loaded files with file ids, but I'll re-open that as a new one then.

lnicola:
Oh, you actually want to get rid of (most of) it, nice!

PR #16303, opened 2024-01-07
feat: add non-exhaustive-let diagnostic

rosefromthedead:
I want this to have a quickfix to add an else branch but I couldn't figure out how to do that, so here's the diagnostic on its own. It pretends a `let` is a match with one arm, and asks the match checking whether that match would be exhaustive.

Previously the pattern was checked based on its own type, but that was causing a panic in `match_check` (while processing e.g. `crates/hir/src/lib.rs`) so I changed it to use the initialiser's type instead, to align with the checking of actual match expressions. I think the panic can still happen, but I hear that `match_check` is going to be updated to a new version from rustc, so I'm posting this now in the hopes that the panic will magically go away when that happens.


crates/ide-diagnostics/src/handlers/non_exhaustive_let.rs
```diff
@@ -0,0 +1,35 @@
...
+    Diagnostic::new_with_syntax_node_ptr(
+        ctx,
+        DiagnosticCode::RustcHardError("E0005"),
+        format!("non-exhaustive pattern: {}", d.uncovered_patterns),
+        d.pat.clone().map(Into::into),
+    )
+}
+
+#[cfg(test)]
+mod tests {
```
Veykril, line 20: Let's add a test that also checks this does not trigger incorrectly, so test an exhaustive let.

crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -235,6 +243,36 @@ impl ExprValidator {
         }
         pattern
     }
+
+    fn validate_block(&mut self, db: &dyn HirDatabase, expr: &Expr) {
+        let Expr::Block { statements, .. } = expr else { return };
+        let body = db.body(self.owner);
```
Veykril, line 249: We should move this into `ExprValidator` as a field and populate it early once. It's a minor perf nitpick but as this is a query every call of it with trigger an RwLock.

crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -235,6 +243,36 @@ impl ExprValidator {
...
+        let pattern_arena = Arena::new();
```
Veykril, line 250: Likewise, lets make this a field so we can re-use the arena from previous iterations (together with the one for general match checking)

Veykril (commented):
Nice and simple diagnostic, thanks! Just some small things to fix. Regarding quickfix we should just wait until we switch to rustc's exhaustiveness checker which should help in that regard.

crates/hir-ty/src/diagnostics/expr.rs
rosefromthedead: Is it beneficial to reuse an arena when there are no references to the old items? I thought it might be similar logic to reusing a Vec after `.clear()`ing it, but there is no equivalent method on arenas.

Veykril: Clearing isn't too beneficial for an arena, as that will just drop the allocated chunks which includes capacity, unlike a vec where capacity remains. The lifetime issues I can see, given how the arena is used. So without some lifetime transmuting, not much we can do about that.

Nadrieril: I think it drops all but the last chunk, which means there is still some good amortizing of capacity compared to making a new one. Also lifetimes should be cleaner with the new code, it might work now.

Nadrieril: Actually it would make sense to intern the lowered patterns if many parts of r-a start to use them

Nadrieril: I'm actually trying to remove the arena entirely, if this turns out ok perf-wise

rosefromthedead:
Re: the panic I had when I posted this, I never had a solid reproducer, so I can only hope that the pattern analysis update solved that. I'd be happy to mark this diagnostic as experimental if we want to avoid shipping that panic.

crates/hir-ty/src/diagnostics/expr.rs
Nadrieril: Yeah, don't worry about reusing the arena, it's going away soon

Nadrieril:
Yeah don't worry about panics in match analysis, they should all be replaced with an `Err` path soon

Nadrieril:
(and by soon I mean https://github.com/rust-lang/rust-analyzer/pull/16533)

PR #16307, opened 2024-01-08
internal: VFS no longer stores all source files in memory

Veykril:
Turns out there is no need to keep the files around. We either upload them to salsa once processed, or we need to keep them around for the `DidChangeTextDocumentNotification`, but that notification is only valid for opened documents, so instead we can just keep the files around in the `MemDocs`!

Fixes https://github.com/rust-lang/rust-analyzer/issues/16301

### DIFF
--- a/crates/vfs/src/lib.rs
+++ b/crates/vfs/src/lib.rs
@@ -1,8 +1,8 @@
 //! # Virtual File System
 //!
-//! VFS stores all files read by rust-analyzer. Reading file contents from VFS
-//! always returns the same contents, unless VFS was explicitly modified with
-//! [`set_file_contents`]. All changes to VFS are logged, and can be retrieved via
+//! VFS records all file changes pushed to it via [`set_file_contents`].
+//! As such it only ever stores changes, not the actual content of a file at any given moment.
+//! All file changes are logged, and can be retrieved via
 //! [`take_changes`] method. The pack of changes is then pushed to `salsa` and
 //! triggers incremental recomputation.
 //!
@@ -84,14 +84,20 @@ impl FileId {
 /// safe because `FileId` is a newtype of `u32`
 impl nohash_hasher::IsEnabled for FileId {}
 
-/// Storage for all files read by rust-analyzer.
+/// Storage for all file changes and the file id to path mapping.
 ///
 /// For more information see the [crate-level](crate) documentation.
 #[derive(Default)]
 pub struct Vfs {
     interner: PathInterner,
-    data: Vec<Option<Vec<u8>>>,
+    data: Vec<FileState>,
     changes: Vec<ChangedFile>,
+}
+
+#[derive(Copy, PartialEq, PartialOrd, Clone)]
+pub enum FileState {
+    Exists,
+    Deleted,
 }
 
 /// Changed file in the [`Vfs`].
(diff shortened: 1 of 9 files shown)


crates/vfs/src/lib.rs
```diff
@@ -219,26 +222,17 @@ impl Vfs {
         let file_id = self.interner.intern(path);
         let idx = file_id.0 as usize;
         let len = self.data.len().max(idx + 1);
-        self.data.resize_with(len, || None);
+        self.data.resize_with(len, || FileState::Deleted);
```
lnicola, line 225: Nit:

```suggestion
        self.data.resize(len, FileState::Deleted);
```

crates/rust-analyzer/tests/slow-tests/main.rs
```diff
@@ -44,7 +44,7 @@ const PROFILE: &str = "";
 #[test]
 fn completes_items_from_standard_library() {
     if skip_slow_tests() {
-        return;
+        // return;
```
lnicola, line 47:
```suggestion
        return;
```

crates/vfs/src/lib.rs
```diff
@@ -219,26 +222,17 @@ impl Vfs {
...
-    ///
-    /// # Panics
-    ///
-    /// Panics if no file is associated to that id.
-    fn get(&self, file_id: FileId) -> &Option<Vec<u8>> {
-        &self.data[file_id.0 as usize]
-    }
-
-    /// Mutably returns the content associated with the given `file_id`.
+    /// Returns the status of the ifle associated with the given `file_id`.
```
lnicola, line 229:
```suggestion
    /// Returns the status of the file associated with the given `file_id`.
```

lnicola (approved):

lnicola:
@Veykril looks like the sync PR is unlikely to merged today, so feel free to merge anything that doesn't touch the proc macro server parts.

davidbarsky (approved):
having touched this side of things, this is a great improvement :)

lnicola:
Let's get it in the nightly.

@bors r+

roife:
When we read a file (`Vec<u8>`) from `vfs-notify`, we may store the text in two places (`mem_docs` and `salsa`). The text stored in `salsa` is of type `Arc<str>`, while in `mem_docs` it is of type `Vec<u8>`. Every time we operate on the `Vec<u8>` retrieved from `mem_docs`, we need to convert it to `String` or `&str` using `from_utf8`; and after that we need to convert it back to bytes with `Vec<u8>` to store in `mem_docs`. Thus, perhaps we can convert the incoming file to UTF-8 in `vfs-notify` itself, so that we don't have to frequently perform conversions between `into_bytes()` and `from_utf8` in subsequent operations.

Furthermore, I noticed a FIXME in `global_state.rs`: `Consider doing (LineEndings) normalization in the vfs instead?` We could also merge this step into `vfs-notify`, thereby reducing locking.

Veykril:
Hmm, ye we do one round trip unnecessarily there. We can fix that, though that won't really matter too much I believe.

Regarding the `FIXME`, that might work though I'm unsure whether that will get rid of any locking (though it might make the lock holding shorter which would also be nice).

roife:
> Hmm, ye we do one round trip unnecessarily there. We can fix that, though that won't really matter too much I believe.
> 
> Regarding the `FIXME`, that might work though I'm unsure whether that will get rid of any locking (though it might make the lock holding shorter which would also be nice).

If this will not have any negative impact on the project, perhaps I can try to do it; at least it can make the project more concise.

Veykril:
Feel free to take a look!

PR #16312, opened 2024-01-08
fix: Fix rust-analyzer-proc-macro-srv failing to spawn on windows

Veykril:


crates/proc-macro-api/src/process.rs
```diff
@@ -131,12 +131,19 @@ impl Process {
...
         .stdin(Stdio::piped())
         .stdout(Stdio::piped())
-        .stderr(if null_stderr { Stdio::null() } else { Stdio::inherit() })
-        .spawn()
+        .stderr(if null_stderr { Stdio::null() } else { Stdio::inherit() });
+    if cfg!(windows) {
+        let mut path_var = std::env::var_os("PATH").unwrap_or_default();
+        path_var.push(";");
+        path_var.push(path.parent().unwrap().parent().unwrap().as_os_str());
+        path_var.push("\\bin");
```
lnicola, line 143: Maybe we should prepend it to avoid rustup interfering with this, not sure.

PR #16314, opened 2024-01-08
`macro_rules` as macro name

### LINKED ISSUE #15969: should not report error for `macro_rules`
**rust-analyzer version**: 0.3.1740

**rustc version**: rustc 1.76.0-nightly (3a85a5cfe 2023-11-20)

**relevant settings**: nothing

code:

```rs
macro_rules! macro_rules {
    () => {}
}

fn main() {
  macro_rules!{};
}
```

rust-analyzer indicates that there is a syntax error, however, rustc can compile the same code without any issues:

image

image

Urhengulas:
This PR makes RA parse `macro_rules! {}` (note the missing identifier) as a `MACRO_CALL` instead of `MACRO_RULES`.

Fixes #15969.

---

PS. I'd like to have `L_BRACKETS.contains(p.nth(2))` as `p.nth(2).is_l_bracket()`, but I am not sure how I should add this method to `SyntaxKind`, since it is autogenerated.


crates/parser/src/grammar/items.rs
```diff
@@ -59,13 +59,7 @@ pub(super) fn item_or_macro(p: &mut Parser<'_>, stop_on_r_curly: bool) {
     };
 
     if paths::is_use_path_start(p) {
-        match macro_call(p) {
-            BlockLike::Block => (),
-            BlockLike::NotBlock => {
-                p.expect(T![;]);
-            }
-        }
-        m.complete(p, MACRO_CALL);
```
Veykril, line 68: So, everything the parser eats before a `complete` call will be attached to the `completed` node. The reason we eat the semicolon here, instead of constructing the macro call in the `macro_call` function is because that functio nis used for parsing a macro call as an item, or as an expression. The semicolon here is only valid when parsed as an item though, hence why the completion is handled in this weird way. So we want to keep things this way here. This is an important distinction as it changes the parse for stuff like `let foo = macro_rules!();`, the semi here belongs to the let statement, not the macro call.

crates/parser/src/grammar/items.rs
```diff
@@ -414,10 +413,23 @@ fn fn_(p: &mut Parser<'_>, m: Marker) {
...
-fn macro_call(p: &mut Parser<'_>) -> BlockLike {
+// test macro_rules_as_macro_name
+// macro_rules! {}
+// macro_rules! {};
+// macro_rules! ();
+// macro_rules! [];
+
+// test_err macro_rules_as_macro_name
+// macro_rules! ()
+// macro_rules! []
```
Veykril, line 424: These tests only test in item position, but not in expression position. For that we'd need something like
```rs
fn main() {
    let foo = macro_rules!();
}
```

crates/parser/src/grammar/items.rs
```diff
@@ -414,10 +430,14 @@ fn fn_(p: &mut Parser<'_>, m: Marker) {
...
-fn macro_call(p: &mut Parser<'_>) -> BlockLike {
+fn macro_call(p: &mut Parser<'_>, m: Marker) {
     assert!(paths::is_use_path_start(p));
     paths::use_path(p);
-    macro_call_after_excl(p)
+    match macro_call_after_excl(p) {
+        BlockLike::Block => p.eat(SEMICOLON),
+        BlockLike::NotBlock => p.expect(SEMICOLON),
+    };
+    m.complete(p, MACRO_CALL);
```
Veykril, line 440: Let's undo this change here again, this is semantically different from the previous behavior. `foo!{};` is not a valid item parse

Urhengulas: Done

crates/parser/src/grammar/items.rs
```diff
@@ -58,13 +58,25 @@ pub(super) fn item_or_macro(p: &mut Parser<'_>, stop_on_r_curly: bool) {
...
+    if paths::is_use_path_start(p)
+        || (p.at_contextual_kw(T![macro_rules]) && p.nth_at(1, BANG) && !p.nth_at(2, IDENT))
+    {
         match macro_call(p) {
-            BlockLike::Block => (),
-            BlockLike::NotBlock => {
-                p.expect(T![;]);
-            }
-        }
+            BlockLike::Block => p.eat(T![;]),
```
Veykril, line 77:
```suggestion
            BlockLike::Block => (),
```
This is still wrong, `macro_call!{};` is only a valid item without the semicolon

Urhengulas: Ah, then I did change it correctly but need to adapt the test. Because I first had it without the eat, but then the test failed.

crates/parser/src/grammar/items.rs
Urhengulas: But wait. Was your coment about not doing the `eat` or about not having it as `macro_call(p: &mut Parser<'_>, m: Marker)` or about both? Because I very much prefer the to add the marker to the signature to be consistent with the other functions.

PR #16315, opened 2024-01-08
Add possible values in the doc

PitiBouchon:
Add the different possible values the doc for non trivial config


PitiBouchon:
Just adding the possible values so that people can find them on [manual configuation](https://rust-analyzer.github.io/manual.html#configuration) instead of looking in the code

Veykril:
I'll close this. While we definitely want to tackle this, it should be ideally be derived from the types somehow, otherwise this will get out of sync rather easily. https://github.com/rust-lang/rust-analyzer/issues/13178 tracks this and once we are done revamping the configs we can tackle it.

PR #16323, opened 2024-01-09
internal: Use rustc-hash in more places

lnicola:
I think there's still ~~a couple~~ a lot of these left, some in tests and the proc macro server.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -651,6 +651,7 @@ dependencies = [
  "ide-db",
  "itertools",
  "profile",
+ "rustc-hash",
  "smallvec",
  "sourcegen",
  "stdx",
@@ -1329,6 +1330,7 @@ dependencies = [
  "libc",
  "once_cell",
  "perf-event",
+ "rustc-hash",
  "tikv-jemalloc-ctl",
  "winapi",
 ]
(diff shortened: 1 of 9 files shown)


Veykril:
Ideally we should add a clippy.toml and the disallowed-types lint with that set up to forbid using plain `HashMap`, now that lints in Cargo.toml are a thing

lnicola:
Good idea, I forgot that was a thing. It feels like the hardest part is bikeshedding them, e.g. what should we use in `lsp-server`?

Veykril:
Hmm, I guess lsp-server is the exception, it should be fine to use in there?

lnicola:
It also doesn't seem to work very well with derive macros. Anyway, I'll give it a try.

Veykril:
what derive macros we use emit hashmap related things? 🤔  or was that more a general note, not specific to r-a's codebase

lnicola:
We have some hashmaps in `lsp_ext` and i think we have to move them into a submodule to allow the lint there, because they derive Serialize.

Veykril:
Might as well just replace those with `FxHashMap` as well since we have the dependency there.

lnicola:
I don't remember exactly why that didn't work, they might end up in `lsp-types` or something like that.

Veykril:
I see, well excluding that module should be fine anyways.

Veykril:
https://github.com/rust-lang/rust-analyzer/pull/16470 supersedes this

PR #16329, opened 2024-01-09
internal: Some minor perf/memory improvements

Veykril:


Veykril:
image

crates/hir-def/src/db.rs
```diff
@@ -210,13 +210,10 @@ pub trait DefDatabase: InternDatabase + ExpandDatabase + Upcast<dyn ExpandDataba
     #[salsa::invoke(AttrsWithOwner::attrs_query)]
     fn attrs(&self, def: AttrDefId) -> Attrs;
 
+    #[salsa::transparent]
```
Veykril, line 213: This was the majority of the memory improvements fwiw. All this saved was a linear lookup through the attributes which in the majority of cases is not that big anyways (exception being heavily documented items).

PR #16333, opened 2024-01-10
Preserve comments for extracted block expression in 'extract_function'

### LINKED ISSUE #14371: Extract into function remove comments in the first of blocks
```Rust
fn x() {
    let c = /*start*/{
        // the comment
        let a = 2;
        let b = 5;
        a + b
    }/*end*/;
}
```
Selecting the range and using the assist will remove `// the comment` from the result.

roife:
Fix #14371 

Preserve comments for extracted block expression in 'extract_function'.

In the original implementation, `block.statements()` was used to construct a new function, removing the comments within the block. In the updated implementation, we use manual traversal of nodes and `hacky_block_expr` to generate a new block, thereby preserving the comments.

### DIFF
--- a/crates/ide-assists/src/handlers/extract_function.rs
+++ b/crates/ide-assists/src/handlers/extract_function.rs
@@ -25,7 +25,7 @@ use syntax::{
         edit::{AstNodeEdit, IndentLevel},
         AstNode, HasGenericParams,
     },
-    match_ast, ted, SyntaxElement,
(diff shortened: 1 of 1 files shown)


Veykril:
Ugh yea, this is a clear indicator for why we need to switch our AST trivia model https://github.com/rust-lang/rust-analyzer/issues/6584

@bors r+

PR #16339, opened 2024-01-10
Replace SourceRootCrates hashset output with slice for deterministic order

Veykril:
We only iterate over the result, and its pretty small in general so no point for the `HashSet` (additionally this way we get a more defined iteration order).

### DIFF
--- a/crates/base-db/src/lib.rs
+++ b/crates/base-db/src/lib.rs
@@ -50,7 +50,7 @@ pub trait FileLoader {
     /// Text of the file.
     fn file_text(&self, file_id: FileId) -> Arc<str>;
     fn resolve_path(&self, path: AnchoredPath<'_>) -> Option<FileId>;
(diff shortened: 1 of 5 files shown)


crates/base-db/src/lib.rs
```diff
@@ -85,19 +85,20 @@ pub trait SourceDatabaseExt: SourceDatabase {
...
         .iter()
         .filter(|&krate| {
             let root_file = graph[krate].root_file_id;
             db.file_source_root(root_file) == id
         })
-        .collect();
-    Arc::new(res)
+        .collect::<FxHashSet<_>>()
+        .into_iter()
+        .collect()
```
lnicola, line 101: Should we sort here?

Veykril: Yes, I had that before but then rewrote this incorrectly (was contemplating using itertools or not) 😅

PR #16349, opened 2024-01-11
fix: add error recovery for use_tree_list parsing

Young-Flash:
This PR adds error recovery for USE_TREE_LIST parsing, avoid the wrong USE_TREE_LIST making the rest parsing incorrectly.

before

before

after

after

close https://github.com/rust-lang/rust-analyzer/issues/16227

### DIFF
--- a/crates/parser/src/grammar/items/use_item.rs
+++ b/crates/parser/src/grammar/items/use_item.rs
@@ -11,7 +11,7 @@ pub(super) fn use_(p: &mut Parser<'_>, m: Marker) {
 
 // test use_tree
 // use outer::tree::{inner::tree};
-fn use_tree(p: &mut Parser<'_>, top_level: bool) {
+fn use_tree(p: &mut Parser<'_>, top_level: bool) -> bool {
     let m = p.start();
     match p.current() {
         // test use_tree_star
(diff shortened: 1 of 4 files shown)


crates/parser/src/grammar/items/use_item.rs
```diff
@@ -70,24 +70,30 @@ fn use_tree(p: &mut Parser<'_>, top_level: bool) {
...
-            return;
+            return false;
         }
     }
     m.complete(p, USE_TREE);
+    true
 }
 
+pub(super) const USE_TREE_LIST_RECOVERY_SET: TokenSet =
+    TokenSet::new(&[T![;], T![,], T![.], T![ident]]);
```
Veykril, line 81: This should probably contain `pub` and other item keywords like struct etc. Also the `T![.]` makes little sense here I think. `ITEM_RECOVERY_SET)` is probably what we want to include.

crates/parser/src/grammar/items/use_item.rs
```diff
@@ -70,24 +70,30 @@ fn use_tree(p: &mut Parser<'_>, top_level: bool) {
...
-        }
-    }
-    p.expect(T!['}']);
+
+    // test_err use_tree_list_err_recovery
+    // use {a;
+    // use b;
+    // struct T;
+    // fn test() {}
+    delimited(p, T!['{'], T!['}'], T![,], USE_TREE_LIST_RECOVERY_SET, |p: &mut Parser<'_>| {
```
Veykril, line 94: `USE_TREE_LIST_RECOVERY_SET` is wrong here, that param in `delimited` is not for recovery but for the first set. So we want a `USE_TREE_LIST_FIRST_SET` that contains the tokens that use tree may start with.

Veykril (commented):
https://github.com/rust-lang/rust-analyzer/pull/4680#issuecomment-641346748 still applies here fwiw

But I think when parsing a `{}` delimited `UseTree` we can just skip unexpected tokens until we find the closing brace or a new item (ITEM_RECOVERY_SET). That should keep the balancing working.

Veykril (commented):
To re-iterate, I don't think matklad's comment on the previous PR applies here. The invariant that was mentioned is still upheld, either the `}` is eaten as the last child or we recover on an item in which case we stop the parse without consuming any closing one. The reparser checks the invariants for balancedness already so at worst we lose out on the re-parse in this case, but I think that is fine.

BUT, there is one problem I can see here I think. With this setup, the following might change now
```rs

struct implThing;

use crate::{impl$0};
```
This should complete `implThing`, but with the changes here it might no longer do that, can you add a completion test for this? If it no longer works (as I suspect) we need to adjust the recovery strategy by doing an extra lookahead after the keyword token for `,` and `}` and recover accordingly

Young-Flash:
actually the completion still works

image

Veykril:
Nice, thanks!
@bors r+

Issue #11361, opened 2022-01-28
Feat: "One" import granularity


CAD97:
The current options for `rust-analyzer.assist.importGranularity` are "preserve", "crate", "module", and "item". Personally, I group _all_ of my imports into a single `use` block (and don't really understand why nobody else seems to do so); rust-analyzer should support (enforcing) this grouping along the existing `use` groupings.

flodiebold:
This should be called `One` like in [rustfmt](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=import#imports_granularity).

Veykril:
Relevant code is mainly here https://github.com/rust-analyzer/rust-analyzer/blob/93036aa0e201474d43c38be3862f4a899defe267/crates/ide_db/src/helpers/insert_use.rs#L155-L193

We also should adjust our guessing infra to be able to understand that form here https://github.com/rust-analyzer/rust-analyzer/blob/93036aa0e201474d43c38be3862f4a899defe267/crates/ide_db/src/helpers/insert_use.rs#L246-L318

And config changes are needed here https://github.com/rust-analyzer/rust-analyzer/blob/93036aa0e201474d43c38be3862f4a899defe267/crates/rust-analyzer/src/config.rs

petr-tik:
can you please assign this issue to me?

petr-tik:
> We also should adjust our guessing infra to be able to understand that form here 

What's the most resilient way to identify the correct ImportGranularityGuess?

Looking at the rustfmt implementation, it looks like having merged use trees one for each visibility level is obviously One, but what about mixed cases? What should it guess?

https://github.com/rust-lang/rustfmt/pull/4680/files#diff-7bd4f14a75a37432af3ab896bbbe38f88428b490b5d76ed1586f7f473a745e98R1772-R1781

Veykril:
To identify we probably just want to check if there is only one tree that starts as `use { ... };`

If mixed it should fall back to Unknown I'd say.

petr-tik:
i was thinking about writing a test that checks for one tree for each of the possible publicity types 

```rust
use { ... };
pub use { ... };
pub(crate) use { ... };
```

thoughts?

Veykril:
Sounds good

petr-tik:
Some personal stuff made me quite unproductive in February. Looks like @jeremyBanks is actively working on it, so it's only fair I am un-assigned from this issue

PR #16372, opened 2024-01-16
feat: Add "One" import granularity

davidsemakula:
Adds a new import granularity option "One" that merges all imports into a single use statement as long as they have the same visibility and attributes.

This is similar to [rustfmt's `imports_granularity = "One"`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=import#imports_granularity).

Fixes: #11361

### DIFF
--- a/crates/ide-assists/src/handlers/merge_imports.rs
+++ b/crates/ide-assists/src/handlers/merge_imports.rs
@@ -1,5 +1,8 @@
 use either::Either;
-use ide_db::imports::merge_imports::{try_merge_imports, try_merge_trees, MergeBehavior};
+use ide_db::imports::{
+    insert_use::{ImportGranularity, InsertUseConfig},
+    merge_imports::{try_merge_imports, try_merge_trees, MergeBehavior},
+};
 use syntax::{
     algo::neighbor,
     ast::{self, edit_in_place::Removable},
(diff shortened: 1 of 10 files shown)


crates/ide-assists/src/handlers/merge_imports.rs
```diff
@@ -30,14 +33,36 @@ pub(crate) fn merge_imports(acc: &mut Assists, ctx: &AssistContext<'_>) -> Optio
...
         let edits = if let Some(use_item) = tree.syntax().parent().and_then(ast::Use::cast) {
             let mut neighbor = next_prev().find_map(|dir| neighbor(&use_item, dir)).into_iter();
-            use_item.try_merge_from(&mut neighbor)
+            use_item.try_merge_from(&mut neighbor, &ctx.config.insert_use)
         } else {
             let mut neighbor = next_prev().find_map(|dir| neighbor(&tree, dir)).into_iter();
-            tree.try_merge_from(&mut neighbor)
+            let mut edits = tree.clone().try_merge_from(&mut neighbor, &ctx.config.insert_use);
+
+            if edits.is_none() && ctx.config.insert_use.granularity == ImportGranularity::One {
```
Veykril, line 45:
```suggestion
            if ctx.config.insert_use.granularity != ImportGranularity::One {
                tree.clone().try_merge_from(&mut neighbor, &ctx.config.insert_use)
            else {
```
This more reasonable to me, since then we don't do work twice.

Veykril (approved):

crates/ide-assists/src/handlers/merge_imports.rs
davidsemakula: Making this change would break [this test](https://github.com/rust-lang/rust-analyzer/pull/16372/files#diff-697b79caa3c0735c983b89fe0b8e88d6b640e19c929d3bd9019e285ae2774353R290-R298) (and similar ones for `ImportGranularity::One` - the [test case just above it](https://github.com/rust-lang/rust-analyzer/pull/16372/files#diff-697b79caa3c0735c983b89fe0b8e88d6b640e19c929d3bd9019e285ae2774353R281-R289) is the motivation for having this behavior). 

But breaking this behavior (at least for `ImportGranularity::One` only for now) may not be a bad thing because the assist is supposed to "Merges two imports with a common prefix". So may be it's better to provide the functionality equivalent to the test case behavior with a new assist (in a separate PR - something like "normalize import")?

If the above make sense, then I can simplify this implementation even further, while maintaining existing behavior for other import granularities. LMK 🙂

Veykril: I see, then that's probably fine. Could you add a https://docs.rs/cov-mark/latest/cov_mark/ here for one of the corresponding tests? That makes this behavior a bit more  clear

Veykril:
Thanks!
@bors r+

PR #16374, opened 2024-01-16
feat: Show notable trait impls on hover

Veykril:


lnicola:
This seems to work for variables, but looks a little weird for types:

image

PR #16380, opened 2024-01-17
Run metadata on sysroot for CI metrics

Veykril:


Veykril:
Fails as expected https://github.com/rust-lang/rust-analyzer/actions/runs/7553812361/job/20565536169#step:5:48, probably cause the sysroot is immutable

lnicola:
Nah, it just needs an internal feature (#16387).

PR #16385, opened 2024-01-17
Fix intra doc links

### LINKED ISSUE #15848: Incorrect doc links in `src/cmp.rs` lead to invalid hyperlinks in `rust-analyzer`
Original post as [Issue 117666](https://github.com/rust-lang/rust/issues/117666) in `rust-lang/rust/` repo.

## Errors in `rust` source code

- [`PartialEq#derivable`](https://github.com/rust-lang/rust/blob/4e0fb98a5c41be1227845219b81ab147708fd99a/library/core/src/cmp.rs#L245) should be `trait.PartialEq#derivable`
- [`Ord#derivable`](https://github.com/rust-lang/rust/blob/4e0fb98a5c41be1227845219b81ab147708fd99a/library/core/src/cmp.rs#L888) should be `trait.Ord#derivable`
- [`PartialOrd`](https://github.com/rust-lang/rust/blob/4e0fb98a5c41be1227845219b81ab147708fd99a/library/core/src/cmp.rs#L1178) should be `trait.PartialOrd#derivable`

## Links to invalid hyperlinks in `https://doc.rust-lang.org/`

- `PartialEq`: https://doc.rust-lang.org/nightly/src/core/cmp.rs.html#245
- `Ord`: https://doc.rust-lang.org/nightly/src/core/cmp.rs.html#888
- `PartialOrd`: https://doc.rust-lang.org/nightly/src/core/cmp.rs.html#1178

## Incorrect links in `rust-analyzer` provides a link to a nonexistent website

screenshot of `rust-analyzer` providing an invalid link

- https://doc.rust-lang.org/nightly/core/cmp/PartialEq#derivable (incorrect)
- https://doc.rust-lang.org/nightly/core/cmp/trait.PartialEq.html#derivable (correct)
  - has `trait` before `.PartialEq`

## Related Issues in `rust-analyzer`
- https://github.com/rust-lang/rust-analyzer/issues/15835

**IDE**: from `VSCode > Help: About`

```
Version: 1.84.1 (user setup)
Commit: 2b35e1e6d88f1ce073683991d1eff5284a32690f
Date: 2023-11-06T12:35:59.022Z
Electron: 25.9.2
ElectronBuildId: 24603566
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Windows_NT x64 10.0.22621
```

- `rust-analyzer version: 0.3.1722-standalone`

- `rustc 1.75.0-nightly (fee5518cd 2023-11-05)`

- no adjustments made that I am aware of. happy to provide if they are deemed relevant!

Urhengulas:
Fixes #15848

### DIFF
--- a/crates/ide/src/doc_links.rs
+++ b/crates/ide/src/doc_links.rs
@@ -64,13 +64,12 @@ pub(crate) fn rewrite_links(db: &RootDatabase, markdown: &str, definition: Defin
             // * path-based links: `../../module/struct.MyStruct.html`
             // * module-based links (AKA intra-doc links): `super::super::module::MyStruct`
             if let Some((target, title)) = rewrite_intra_doc_link(db, definition, target, title) {
-                return (None, target, title);
+                (None, target, title)
+            } else if let Some(target) = rewrite_url_link(db, definition, target) {
+                (Some(LinkType::Inline), target, title.to_string())
+            } else {
+                (None, target.to_string(), title.to_string())
             }
-            if let Some(target) = rewrite_url_link(db, definition, target) {
-                return (Some(LinkType::Inline), target, title.to_string());
-            }
-
-            (None, target.to_string(), title.to_string())
         }
     });
     let mut out = String::new();
--- a/crates/ide/src/doc_links.rs
+++ b/crates/ide/src/doc_links.rs
@@ -368,16 +368,21 @@ fn rewrite_intra_doc_link(
 ) -> Option<(String, String)> {
     let (link, ns) = parse_intra_doc_link(target);
 
+    let (link, anchor) = match link.split_once('#') {
+        Some((new_link, anchor)) => (new_link, Some(anchor)),
+        None => (link, None),
+    };
+
     let resolved = resolve_doc_path_for_def(db, def, link, ns)?;
     let mut url = get_doc_base_urls(db, resolved, None, None).0?;
 
-    let (_, file, frag) = filename_and_frag_for_def(db, resolved)?;
+    let (_, file, _) = filename_and_frag_for_def(db, resolved)?;
     if let Some(path) = mod_path_of_def(db, resolved) {
         url = url.join(&path).ok()?;
     }
 
     url = url.join(&file).ok()?;
-    url.set_fragment(frag.as_deref());
+    url.set_fragment(anchor);
 
     Some((url.into(), strip_prefixes_suffixes(title).to_string()))
 }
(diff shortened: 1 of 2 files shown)


Urhengulas:
My first step is to add two tests.

First one for intra doc links without an anchor. This already does work in RA, therefore the test should succeed.
Second one for intra doc links with an anchor. This fails atm, therefore the test should fail.

Unfortunately I am not able to get the first test to work. It does not even resolve to a URL at all:

```
---- doc_links::tests::rewrite_intra_doc_link stdout ----

Expect:
----
[PartialEq](https://doc.rust-lang.org/stable/std/cmp/trait.PartialEq.html)
----

Actual:
----
[PartialEq](PartialEq)
----
```

Interestingly the test with anchor does resolve to a URL pointing to the correct domain, but the path is wrong:

```
---- doc_links::tests::rewrite_intra_doc_link_with_anchor stdout ----

Expect:
----
[PartialEq#derivable](https://doc.rust-lang.org/stable/std/cmp/trait.PartialEq.html#derivable)
----

Actual:
----
[PartialEq#derivable](https://doc.rust-lang.org/stable/std/PartialEq#derivable)
----
```

My question: How can I get the first test to work? It has to work because RA already correctly handles this.

alibektas:
Try this `env UPDATE_EXPECT=1 cargo test doc_links::tests` !

Urhengulas:
> Try this `env UPDATE_EXPECT=1 cargo test -p ide doc_links::tests` !

While this would make the tests pass, this does not help me, because the actual output of the tests are wrong. It should be what is in the expect. So my question is how I can make the actual output fit the expected output. For the first test it should work without making changes to the code, because the functionality already works in stable rust-analyzer. To make the second test work I need to change the code (that's what the issue is about), but first I want to get the first test working.

Veykril:
See also https://github.com/rust-lang/rust-analyzer/blob/master/crates/test-utils/src/minicore.rs#L826

Urhengulas:
> //- minicore: eq, derive

That works and makes the first test succeed. Thank youu 🎈

Urhengulas:
I've implemented the functionality with a surprisingly simple method. I try to split the `link` (e.g. `"PartialEq"` or `"PartialEq#derivable"`) at a `#`. If a `#` was present I resolve the url for the part before the `#` (the new `link`) and store the part after the `#` (the `anchor`). After resolving the url I just add the `anchor` to the end of the url.

This makes my test case pass. Are there other cases of intra-doc links with or without anchors where this could fail? None were in the tests so it seems fine.

I want to point out that the PR stopps using the `frag` returned from `filename_and_frag_for_def`, but this does not fail any tests so, again, it seems fine.

Veykril (approved):

PR #16387, opened 2024-01-17
internal: Fix sysroot metadata in metrics

lnicola:


lnicola:
changelog fixup #16380

PR #16391, opened 2024-01-17
Add a new config to allow renaming of non-local defs

alibektas:
With #15656 we started disallowing renaming of non-local items. Although this makes sense there are some false positives that impacted users' workflows. So this config aims to mitigate this by giving users the liberty to disable this feature.

The reason why this is a draft is that I saw one of the tests fail and I am not sure if the "got" result even syntactically makes sense 

Test case is :

```rust
check(
            "Baz",
            r#"
//- /lib.rs crate:lib new_source_root:library
pub struct S;
//- /main.rs crate:main deps:lib new_source_root:local
use lib::S$0;
"#,
            "use lib::Baz;"
);
```

```
Left:
use lib::Baz;

Right:
use lib::Baz;Baz

Diff:
use lib::Baz;Baz
```

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -494,6 +494,9 @@ config_data! {
         /// Exclude imports from find-all-references.
         references_excludeImports: bool = "false",
 
+        /// Allow renaming of items not belonging to any workspace crates.
+        renameExternalItems_enable : bool = "false",
+
         /// Command to be executed instead of 'cargo' for runnables.
         runnables_command: Option<String> = "null",
         /// Additional arguments to be passed to cargo for runnables such as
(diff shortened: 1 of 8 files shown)


Veykril:
The test runner was not aware of renames affecting multiple files so it just concatenated the changes. I fixed that and pushed it to your branch.

crates/rust-analyzer/src/config.rs
```diff
@@ -494,6 +494,9 @@ config_data! {
         /// Exclude imports from find-all-references.
         references_excludeImports: bool = "false",
 
+        /// Allow renaming of items not belonging to any workspace crates.
+        renameExternalItems_enable : bool = "false",
```
Veykril, line 498: Might be better to add the `allow` part to the name, and then to future proof namespace it so `rename_allowExternalItems` or something allong those lines?

crates/rust-analyzer/src/config.rs
```diff
@@ -494,6 +494,9 @@ config_data! {
         /// Exclude imports from find-all-references.
         references_excludeImports: bool = "false",
 
+        /// Allow renaming of items not belonging to any workspace crates.
```
Veykril, line 497:
```suggestion
        /// Allow renaming of items not belonging to the loaded workspaces.
```

PR #16394, opened 2024-01-18
internal: add some `tracing` to {Request, Notification}Dispatch

davidbarsky:
Some of the tracing config would need to be changed in order to benefit more from this (especially `.with_span_events(FmtSpan::CLOSE)`), which provides span events like the following:

```
2024-01-18T02:41:06.675779Z  INFO request{method="textDocument/codeLens" request_id=RequestId(I32(17))}: rust_analyzer::dispatch: close time.busy=61.8µs time.idle=5.29µs
```

I dunno if y'all need `LoggerFormatter` in here, but if you don't, I'd be happy to yeet it out of rust-analyzer. In any case, this provided a pretty decent amount of information in the logs, and I can expand this as needed or we can land this PR and expand later.


Veykril:
`LoggerFormatter` I think we only have because of salsa using log (we can replace this with tracing since we forked it), and https://github.com/matklad/always-assert. I'd love to get rid of the `LoggerFormatter`.

Veykril:
On a related note, would be nice if we could replace our `hprof` thing with tracing / just use tracing for that kind of stuff.

matklad:
Yeah, I'd personally feel pretty strongly about not building two profiling infras. At this point, going all-in on tracing makes sense. 

I used tracing with https://github.com/matklad/tracing-span-tree/ at NEAR, and the end-result was pretty similar to hprof. 

If someone is looking into this seriously, I'd also recommend taking a closer look tracing-tracy:

https://github.com/nagisa/rust_tracy_client

davidbarsky:
I'll swap the hprof infrastructure over to `tracing`. I don't have a strong preference between `tracing-tree` and `tracing-span-tree` since both seem to do the job and they're pretty easy to swap out as needed.

As for the `LoggerFormatter`, I'll send PRs for both those crates to use `tracing`, forking as necessary.

davidbarsky:
Oh, rust-analyzer is already using `tracing-tree`. I'll see if I can make it more hprof-like.

davidbarsky:
Quick update:
- I pushed changes to this branch that switched all callsites of `profile::span` with `let _p = tracing::span!(tracing::Level::INFO, "SPAN_NAME").entered();`. Let me know if this looks reasonable at a glance before I continue on and rewrite the profiling infrastructure in terms of `tracing`. @matklad: for https://github.com/matklad/tracing-span-tree/, did you  have similar filter infrastructure as hprof?
    - Note that the profiling filtering does not yet work. I'll need to look into why it's misbehaving, so this isn't ready to get merged.
- I removed `log` in https://github.com/matklad/always-assert/pull/3 and https://github.com/rust-analyzer/salsa/pull/6. I'll update this PR once those PRs are merged.

Veykril:
>  Let me know if this looks reasonable at a glance before I continue on and rewrite the profiling infrastructure in terms of `tracing`

👍

davidbarsky:
Few updates: I've swapped the hprof infrastructure over to a vendored version of `tracing-span-tree`, which is currently set to be equivalent to `RAPROFILE=*>2`. It works nicely I think, but I'll port the the existing parsing/filtering infrastructure over to tracing in a second.

Cargo.lock
```diff
@@ -1759,7 +1768,7 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 name = "stdx"
 version = "0.0.0"
 dependencies = [
- "always-assert",
```
Veykril, line 1762: Needs a version bump for the dependency

davidbarsky: whoops, ty. fixed.

Veykril:
@bors delegate+
r=me with CI green

davidbarsky:
While CI is passing, I'd like to hold on merging to squash this history and ensure that unreported profiles are, well, actually reported as they were before this change.

davidbarsky:
I've squashed the history, but opted to avoid reporting on unreported spans because my naive implementation of unreported profiles was *really* noisy. My feeling is that I'd rather land this PR and re-implement that later, need be. Anyways, here's a screenshot of the output:

<details>
<summary>Collapsed Screenshot</summary>
A screenshot of VS Code on macOS. It contains a file explorer, a open Rust file (`hprof.rs`), and the `RA_PROFILE` output from `tracing` in the output pane.
</details>

PR #16395, opened 2024-01-18
internal: speedup LineEndings calculation using 'memchr'

roife:
See https://github.com/rust-lang/rust-analyzer/issues/13538


Veykril:
Thanks!
@bors r+

PR #16397, opened 2024-01-18
Refactor `macro_call` to be consistent with other functions

Urhengulas:
https://github.com/rust-lang/rust-analyzer/pull/16314#discussion_r1457324734


Urhengulas:
There is some macos CI failure. Any clue why this is?

Veykril:
That test sometimes fails, we can ignore that
@bors r+

PR #16398, opened 2024-01-18
`cargo clippy --fix`

Urhengulas:
This PR is the result of running `cargo clippy --fix && cargo fmt` in the root of the repository. I did not manually review all the changes, but just skimmed through a few of them. The tests still pass, so it seems fine.

### DIFF
--- a/crates/hir-expand/src/builtin_fn_macro.rs
+++ b/crates/hir-expand/src/builtin_fn_macro.rs
@@ -125,7 +125,7 @@ fn mk_pound(span: Span) -> tt::Subtree {
         vec![crate::tt::Leaf::Punct(crate::tt::Punct {
             char: '#',
             spacing: crate::tt::Spacing::Alone,
-            span: span,
+            span,
         })
         .into()],
         span,
@@ -279,9 +279,9 @@ fn format_args_expand_general(
     let pound = mk_pound(span);
     let mut tt = tt.clone();
     tt.delimiter.kind = tt::DelimiterKind::Parenthesis;
-    return ExpandResult::ok(quote! {span =>
+    ExpandResult::ok(quote! {span =>
         builtin #pound format_args #tt
-    });
+    })
 }
 
 fn asm_expand(
@@ -624,7 +624,7 @@ fn relative_file(
 
 fn parse_string(tt: &tt::Subtree) -> Result<String, ExpandError> {
     tt.token_trees
-        .get(0)
+        .first()
         .and_then(|tt| match tt {
             tt::TokenTree::Leaf(tt::Leaf::Literal(it)) => unquote_str(it),
             _ => None,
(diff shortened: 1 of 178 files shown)


lnicola:
`ASSERT_MAX_FILE_ID_IS_SAME` looks pretty gnarly, I'm not surprised Clippy got confused.

lnicola:
I also see some new warnings:

image

Urhengulas:
> I also see some new warnings:

I am on it

crates/hir-expand/src/builtin_fn_macro.rs
```diff
@@ -396,21 +396,19 @@ fn use_panic_2021(db: &dyn ExpandDatabase, span: Span) -> bool {
...
+    let expn = db.lookup_intern_macro_call(expn);
+    // FIXME: Record allow_internal_unstable in the macro def (not been done yet because it
+    // would consume quite a bit extra memory for all call locs...)
+    // if let Some(features) = expn.def.allow_internal_unstable {
+    //     if features.iter().any(|&f| f == sym::edition_panic) {
+    //         span = expn.call_site;
+    //         continue;
+    //     }
+    // }
+    expn.def.edition >= Edition::Edition2021
```
Veykril, line 411: Please undo this again, this is a loop because of the commented out code and so has meaning once thats fixed again

crates/stdx/src/thread/pool.rs
```diff
@@ -86,6 +86,7 @@ impl Pool {
         self.job_sender.send(job).unwrap();
     }
 
+    #[allow(clippy::len_without_is_empty)]
```
Veykril, line 89: And this is why we don't really run clippy on the code base 😅 the defaults are sometimes questionable

lnicola: It's not Clippy's fault, I'd expect `len` to return either the thread count or the pending queue size, not the number of running jobs.

crates/hir-expand/src/builtin_fn_macro.rs
Urhengulas: Sure

crates/stdx/src/thread/pool.rs
Veykril: Overall yes, I do want us to run clippy. I just don't like some of the default set of lints it has. Related https://github.com/rust-lang/rust-analyzer/issues/15918

crates/hir-expand/src/builtin_fn_macro.rs
Urhengulas: I've reset the branch to not include the manual changes. They will come in a separate PR.

ivan:
https://github.com/rust-lang/rust-analyzer/pull/16398/files#diff-fa039ab572a8591c57be2cea529bb1e287b071b8f30cd5d495da3df98d2a31d1L191-L193 `match` was removed here, but I think this test was specifically testing match patterns.

lnicola:
Looks like some of the other tests in there might be affected too.

Veykril:
Oh, that should definitely be reverted. Forgot we had these weird tests there

Urhengulas:
@ivan thank you for spotting, created an issue for it

PR #16400, opened 2024-01-18
Update bail message referencing config

alibektas:
Even though we changed the name of the config I forgot to update the warning message that referenced it.

### DIFF
--- a/crates/ide-db/src/rename.rs
+++ b/crates/ide-db/src/rename.rs
@@ -81,7 +81,7 @@ impl Definition {
             // Can we not rename non-local items?
             // Then bail if non-local
             if !rename_external && !krate.origin(sema.db).is_local() {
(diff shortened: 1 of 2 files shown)


crates/ide-db/src/rename.rs
```diff
@@ -81,7 +81,7 @@ impl Definition {
             // Can we not rename non-local items?
             // Then bail if non-local
             if !rename_external && !krate.origin(sema.db).is_local() {
-                bail!("Cannot rename a non-local definition. Set `renameExternalItems` to `true` to allow renaming for this item.")
+                bail!("Cannot rename a non-local definition. Set `rename_allowExternalItems` to `true` to allow renaming for this item.")
```
Veykril, line 84:
```suggestion
                bail!("Cannot rename a non-local definition as the config for it is disabled")
```
The config will be somewhat editor dependent so we should be more general

Veykril:
You need to update the tests
@bors delegate+

PR #16401, opened 2024-01-19
Expand lint tables && make clippy happy 🎉 

Urhengulas:
This PR expands the lint tables on `./Cargo.toml` and thereby makes `cargo clippy` exit successfully! 🎉

Fixes #15918 

## How?

In the beginning there are some warnings for rustc.

Next, and most importantly, there is the clippy lint table. There are a few sections in there.

First there are the lint groups.

Second there are all lints which are permanently allowed with the reasoning why they are allowed.

Third there is a huge list of temporarily allowed lints. They should be removed in the mid-term, but incur a substantial amount of work, therefore they are allowed for now and can be worked on bit by bit. 

Fourth there are all lints which should warn.

Additionally there are a few allow statements in the code for lints which should be permanently allowed in this specific place, but not in the whole code base.

## Follow up work
- [ ] Run clippy in CI
- [ ] Remove tidy test (at least @Veykril wrote this in #15017)
- [ ] Work on temporarily allowed lints

PR #16404, opened 2024-01-19
Work through temporarily allowed clippy lints, part 1

Urhengulas:
This is the first batch of not allowing but actually fixing the clippy lints. Each commit removes one lint from the lint table and then fixes the resulting warnings.

Follow-up to #16401


crates/hir-ty/src/mir/eval.rs
```diff
@@ -272,6 +272,7 @@ const STACK_OFFSET: usize = 1 << 30;
 const HEAP_OFFSET: usize = 1 << 29;
 
 impl Address {
+    #[allow(clippy::double_parens)]
     fn from_bytes(it: &[u8]) -> Result<Self> {
         Ok(Address::from_usize(from_bytes!(usize, it)))
     }
```
Veykril, line 278: Feels like this lint should not trigger on macro expansions 🤔

Veykril (approved):

crates/hir-ty/src/mir/eval.rs
Urhengulas: Good point, I will report it and the other bug as well

PR #16412, opened 2024-01-21

lnicola:


lnicola:
@bors r+

I've temporarily reverted 6001c50, will upgrade the auto-published crates and revert the revert next week.

PR #16413, opened 2024-01-21
internal: Add `clippy` to CI

Urhengulas:
Follow-up to #16401

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -105,7 +105,7 @@ jobs:
 
       - name: clippy
         if: matrix.os == 'ubuntu-latest'
-        run: cargo clippy --all-targets -- -D warnings
+        run: cargo clippy --all-targets
 
   # Weird targets to catch non-portable code
   rust-cross:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -90,7 +90,7 @@ jobs:
       - name: Switch to stable toolchain
         run: |
           rustup update --no-self-update stable
-          rustup component add --toolchain stable rust-src
+          rustup component add --toolchain stable rust-src clippy
           rustup default stable
 
       - name: Run analysis-stats on rust-analyzer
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yaml
```diff
@@ -103,6 +103,10 @@ jobs:
           RUSTC_BOOTSTRAP: 1
         run: target/${{ matrix.target }}/debug/rust-analyzer analysis-stats --with-deps $(rustc --print sysroot)/lib/rustlib/src/rust/library/std
 
+      - name: clippy
+        if: matrix.os == 'ubuntu-latest'
+        run: cargo clippy --all-targets -- -D warnings
```
Veykril, line 108: Does this enable *all clippy* warnings or specifically only the ones we list in the `Cargo.toml` files? We only want the latter as we want direct control of the lints.

Urhengulas: It does enable all warnings.

The approach I took is enabling all warnings, and then ignoring (aka. "allow"ing) all warnings we do not care about. Some warnings gonna be ignored forever, but some others are gonna be worked on (as happened in e.g. #16404).

Urhengulas: We could also only enable ("deny" in clippy jargon) the lints we care about, but I prefer the approach of considering all clippy lints, except we find them to be not fitting to our needs.

Veykril: Ah wait now I see. I'm confused as to why we pass `-D warnings` here, thats not a clippy lint group and clippy should check normal rustc warnings by default no? All the other groups that are relevant we have enabled in the lints table already.

Urhengulas: Good point. Gonna remove that.

Urhengulas:
Fixed the new clippy errors which landed on master while this PR is open. Let's try to merge it soon, otherwise we need to fix more clippy errors.

Veykril (approved):

lnicola:
Our builds are a little lopsided, anyone opposed to running Clippy on Windows?

image

Veykril:
👍

flodiebold:
Clippy really belongs on Windows anyway :smile:

PR #16415, opened 2024-01-22
internal: Make TryToNav trait public

Waqar144:
Currently there is no proper way to get a target FileRange for a given Definition.


davidbarsky:
I think this makes sense, but I won't approve this yet. Just to confirm, this for a different binary/library consuming this trait?

Waqar144:
> I think this makes sense, but I won't approve this yet. Just to confirm, this for a different binary/library consuming this trait?

yes, its for a tool.

PR #16416, opened 2024-01-22
internal: VS Code: report conflict with `panicbit.cargo`

Young-Flash:
notification

related to https://github.com/rust-lang/rust-analyzer/issues/16392

### DIFF
--- a/editors/code/src/main.ts
+++ b/editors/code/src/main.ts
@@ -25,7 +25,7 @@ export async function deactivate() {
 export async function activate(
     context: vscode.ExtensionContext,
 ): Promise<RustAnalyzerExtensionApi> {
-    conflictExtDetect();
+    checkConflictingExtensions();
 
     const ctx = new Ctx(context, createCommands(), fetchWorkspace());
     // VS Code doesn't show a notification when an extension fails to activate
@@ -192,7 +192,7 @@ function createCommands(): Record<string, CommandFactory> {
     };
 }
 
-function conflictExtDetect() {
+function checkConflictingExtensions() {
     if (vscode.extensions.getExtension("rust-lang.rust")) {
         vscode.window
             .showWarningMessage(
@@ -205,20 +205,12 @@ function conflictExtDetect() {
     }
 
     if (vscode.extensions.getExtension("panicbit.cargo")) {
-        const isRustAnalyzerCheckOnSave = vscode.workspace
-            .getConfiguration("rust-analyzer")
-            .get("checkOnSave");
-        const isCargoAutomaticCheck = vscode.workspace
-            .getConfiguration("cargo")
-            .get("automaticCheck");
-        if (isRustAnalyzerCheckOnSave && isCargoAutomaticCheck) {
-            vscode.window
-                .showWarningMessage(
-                    `You have Cargo (panicbit.cargo) enabled with 'cargo.automaticCheck' set to true(default), ` +
-                        'you can disable it or set {"cargo.automaticCheck": false} in settings.json to avoid invoke cargo twice',
-                    "Got it",
-                )
-                .then(() => {}, console.error);
-        }
+        vscode.window
+            .showWarningMessage(
+                `You have both the rust-analyzer (rust-lang.rust-analyzer) and Cargo (panicbit.cargo) ` +
+                    'you can disable it or set {"cargo.automaticCheck": false} in settings.json to avoid invoking cargo twice',
+                "Got it",
+            )
+            .then(() => {}, console.error);
     }
 }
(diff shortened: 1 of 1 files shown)


editors/code/src/main.ts
```diff
@@ -200,3 +191,34 @@ function createCommands(): Record<string, CommandFactory> {
...
+    }
+
+    if (vscode.extensions.getExtension("panicbit.cargo")) {
+        const isRustAnalyzerCheckOnSave = vscode.workspace
+            .getConfiguration("rust-analyzer")
+            .get("checkOnSave");
+        const isCargoAutomaticCheck = vscode.workspace
+            .getConfiguration("cargo")
+            .get("automaticCheck");
+        if (isRustAnalyzerCheckOnSave && isCargoAutomaticCheck) {
```
davidbarsky, line 214: maybe a dumb question: is it possible to _just_ check if the Cargo extension is installed? I assume that most of the functionality is now in rust-analyzer, modulo automatically adding dependencies.

editors/code/src/main.ts
```diff
@@ -200,3 +191,34 @@ function createCommands(): Record<string, CommandFactory> {
         revealDependency: { enabled: commands.revealDependency },
     };
 }
+
+function conflictExtDetect() {
```
davidbarsky, line 195: small nit: i would call this function `checkConflictingExtensions()` to match how the other functions are named.

editors/code/src/main.ts
Young-Flash: done

editors/code/src/main.ts
```diff
@@ -200,3 +191,26 @@ function createCommands(): Record<string, CommandFactory> {
...
+                    "both plugins to not work correctly. You should disable one of them.",
+                "Got it",
+            )
+            .then(() => {}, console.error);
+    }
+
+    if (vscode.extensions.getExtension("panicbit.cargo")) {
+        vscode.window
+            .showWarningMessage(
+                `You have both the rust-analyzer (rust-lang.rust-analyzer) and Cargo (panicbit.cargo) ` +
```
Veykril, line 210:
```suggestion
                `You have both the rust-analyzer (rust-lang.rust-analyzer) and Cargo (panicbit.cargo) plugins enabled` +
```

editors/code/src/main.ts
lnicola: Feels like this might be really annoying for someone who wants to use the features of the other extension.

editors/code/src/main.ts
```diff
@@ -200,3 +191,26 @@ function createCommands(): Record<string, CommandFactory> {
         revealDependency: { enabled: commands.revealDependency },
     };
 }
+
+function checkConflictingExtensions() {
+    if (vscode.extensions.getExtension("rust-lang.rust")) {
```
lnicola, line 196: We can probably drop this check in the near future (May?), it's been marked for years as deprecated and it warns on start-up.

Young-Flash: I prefer to keep it in this PR for now. It could be removed in next PR if we do decide to drop it.

lnicola: Yeah, I was proposing end of May or so, to mark two years of depreciation notices.

editors/code/src/main.ts
```diff
@@ -200,3 +191,26 @@ function createCommands(): Record<string, CommandFactory> {
...
+        vscode.window
+            .showWarningMessage(
+                `You have both the rust-analyzer (rust-lang.rust-analyzer) and Rust (rust-lang.rust) ` +
+                    "plugins enabled. These are known to conflict and cause various functions of " +
+                    "both plugins to not work correctly. You should disable one of them.",
+                "Got it",
+            )
+            .then(() => {}, console.error);
```
lnicola, line 204: Nit:

```suggestion
            .catch(console.error);
```

editors/code/src/main.ts
```diff
@@ -200,3 +191,26 @@ function createCommands(): Record<string, CommandFactory> {
...
+    }
+
+    if (vscode.extensions.getExtension("panicbit.cargo")) {
+        vscode.window
+            .showWarningMessage(
+                `You have both the rust-analyzer (rust-lang.rust-analyzer) and Cargo (panicbit.cargo) plugins enabled` +
+                    'you can disable it or set {"cargo.automaticCheck": false} in settings.json to avoid invoking cargo twice',
+                "Got it",
+            )
+            .then(() => {}, console.error);
```
lnicola, line 214:
```suggestion
            .catch(console.error);
```

lnicola (approved):

PR #16420, opened 2024-01-23
Use upstream exhaustiveness checker!

Nadrieril:
Because it has been librarified!

The extra `Apache-2.0 WITH LLVM-exception` license is for `rustc_apfloat`. Also this duplicates `rustc_index` because the other upstream deps are still on an earlier version. They should be bumpable now though. Good thing is that we don't need this new crate to be synchronized with the others, which will make our lives easier.


crates/hir-ty/src/lib.rs
```diff
@@ -15,6 +15,9 @@ extern crate rustc_abi;
 #[cfg(not(feature = "in-rust-tree"))]
 extern crate ra_ap_rustc_abi as rustc_abi;
 
+// Don't use the in-tree one even when we could.
+extern crate ra_ap_rustc_pattern_analysis as rustc_pattern_analysis;
+
```
HKalbasi, line 20: What is the problem with the in tree one?

Nadrieril: I'd like to be able to update them separately. E.g. if I make a breaking change in the in-tree one, I don't need to fix rust-analyzer instantly. It just adds a lot of complexity compared to treating it like a normal crate.

Nadrieril: Said differently, the reason we use the in-tree rustc-lexer etc is because it's important to follow closely the rustc version. There's no such need for exhaustiveness checking

Nadrieril: I'll rephrase the comment actually

HKalbasi: It's fine for now that things are moving fast, but when the dust settles I think it would be better to use the in-tree version here as well. Using the in-tree version ensures that r-a can keep its rustc dependencies update without worrying about api breakages in the rustc side.

Nadrieril: You mean it's a way for the rustc side to be responsible for not breaking r-a? Ok that makes sense. For now I'm taking that responsibility by hand, and yeah when the dust settles we can automate it

HKalbasi: Yes, that's one of the proposes of this in-tree thing.

Using the in-tree version also has the benefit that if the pull request is completely unrelated to the r-a and just fixes the breakages, you can do r-a changes in the same PR in the rust repository, which may make things simpler (1 PR instead of 2 PRs).

Nadrieril: Hm, that's a strong point. Right now I'd have to wait for the rustc one to be merged, wait for republish, and make a PR to r-a.

What happens when a PR that changes r-a is merged into rustc? Does it get merged into r-a master automatically?

lnicola: Subtree syncs happen every week, but they often fail because of bootstrap shenanigans.

Nadrieril: Hm, slow either way. Does rustc CI run rust-analyzer tests?

Nadrieril: Ok, then for now I'd rather not use the in-tree crate, that will make it easier to iterate

Cargo.lock
```diff
@@ -1455,10 +1490,24 @@ version = "0.21.0"
...
  "ra-ap-rustc_lexer",
 ]
 
+[[package]]
+name = "ra-ap-rustc_pattern_analysis"
+version = "0.33.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6c4085e0c771fd4b883930b599ef42966b855762bbe4052c17673b3253421a6d"
+dependencies = [
+ "derivative",
```
lnicola, line 1503: Oof, we went through some pains to get rid of `syn` 1.0. But you do seem to use it in a couple of places, so doing that by hand would probably be too annoying.

Nadrieril: Oh no 🙈 I mean it's not *too* hard to impl by hand

lnicola: It's not, but I don't want to force you to spend time on that right now. It's okay to merge, then clean it up later (ideally).

Veykril: We should just fix up the assist to inline custom proc-macro derive expansions (which currently only supports builtin derives) :)

But to chime in as well, I'd like to see derivative not stick in our dependency tree for long. Iirc it, like syn, has a pretty bad compile time impact. 

> It's okay to merge, then clean it up later (ideally).

☝️

Veykril:
Excited to see what we can do with this!
@bors r+

Cargo.lock
Nadrieril: Maybe https://github.com/mcarton/rust-derivative/pull/118 can save us! If it ever gets merged

Nadrieril: It seems rustc also wants to get rid of syn 1.0 https://github.com/rust-lang/rust/issues/109302

Veykril: Hmm, actually rust-derivative does not seem to be maintained anymore (3 years since the last commit). Anyways, since rustc wants to bump to syn 2 as well T-compiler will probably figure something out there at some point.

Nadrieril: Yeah, that PR I linked was opened recently by the person who's working on moving rustc to syn2, hopefully they'll take some kind of action

lnicola: Fine, I'll do it :joy:.

PR #16426, opened 2024-01-25
minor: Bump rustc crates

lnicola:
This also reverts the earlier revert.


crates/parser/src/lexed_str.rs
```diff
@@ -369,6 +369,7 @@ fn error_to_diagnostic_message(error: EscapeError, mode: Mode) -> &'static str {
             "non-ASCII character in byte string literal"
         }
         EscapeError::NonAsciiCharInByte => "non-ASCII character in raw byte string literal",
+        EscapeError::NulInCStr => "null character in C string literal",
```
lnicola, line 372: Bah, I call these NUL characters, pointers are null, but [maybe](https://en.wikipedia.org/wiki/Null_character) I'm just old and cranky.

PR #16427, opened 2024-01-25
fix: filter out cfg disabled filed when lowering `RecordPat`

Young-Flash:
we should filter out field with disabled cfg when lowering ast `RecordPat` to hir.

close https://github.com/rust-lang/rust-analyzer/issues/16169


Veykril:
Thanks!
@bors r+

PR #16435, opened 2024-01-27
Do not return code lens data after resolving

SomeoneToIgnore:


lnicola (approved):

lnicola:
Looks like we do something similar in `handle_inlay_hints_resolve`, so I guess it's fine. Thanks!

@bors r+

PR #16439, opened 2024-01-27
feat: Support for GOTO def from *inside* files included with include! macro

### LINKED ISSUE #14937: Support for GOTO def from *inside* files included with `include!` macro?
so as of https://github.com/rust-lang/rust-analyzer/pull/14561, goto definition will now correctly locate tokens defined in a file that is included with the `include!` macro. This is wonderful.

Now there is the natural follow-up: would it be possible to have goto work *within* the generated file?

I can see some ways that this is complicated. In particular, there is potentially an `N->1` relationship between `include!` invocations and their targets, which means that there could be different sets (of possibly overlapping) symbols in scope for a given included file. I would imagine that any possible implementation of this would need to be limited to the case where any single file is only included from a single other file.

So... I'm really not sure how feasible this is, but it seems at least like something that would be nice to have eventually.

wasd96040501:
close #14937
Try to implement goto def from *inside* files included with include! macro.
This implementation has two limitations:
1. Only **one** file which calls include! will be tracked. (I think multiple file be included is a rare case and we may let it go)
2. Mapping token from included file to macro call file (semantics.rs:646~658) works fine but I am not sure is this the correct way to implement.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -483,6 +483,7 @@ dependencies = [
  "profile",
  "rustc-hash",
  "smallvec",
+ "span",
  "stdx",
  "syntax",
  "triomphe",
--- a/crates/hir-def/src/db.rs
+++ b/crates/hir-def/src/db.rs
@@ -1,9 +1,10 @@
 //! Defines database & queries for name resolution.
-use base_db::{salsa, CrateId, SourceDatabase, Upcast};
+use base_db::{salsa, CrateId, FileId, SourceDatabase, Upcast};
 use either::Either;
 use hir_expand::{db::ExpandDatabase, HirFileId, MacroDefId};
 use intern::Interned;
 use la_arena::ArenaMap;
+use span::MacroCallId;
 use syntax::{ast, AstPtr};
 use triomphe::Arc;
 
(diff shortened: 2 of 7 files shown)


cmyr:
Thanks for looking into this! I've played around with it a bit in a test project, where it's working fine, and also in my real project where I haven't been able to get it to work.

Can you clarify what you mean by "Only one file which calls include! will be tracked."? Does this mean that we only track only track one `include!` for a given _target_ file, or do you mean something else (like in a given project, only a single file can have `include!` statements that are tracked?)

In my test project it looks like this is working for multiple includes (e.g. `A` includes `A'` and `B` includes `B'` and I can navigate from both the included files.)


If you're curious to see the original project that motivated this issue it's available https://github.com/googlefonts/fontations/; specifically both the `read-fonts` and `write-fonts` crates use a pattern where there is a `generated/` folder containing a bunch of code that is included from corresponding files in the `src/tables` directory.

wasd96040501:
@cmyr Thanks for your positive feedback!
"Only one file which calls include! will be tracked." means, for example, a file `A` is included both by `B` and `C`, the navigation in `A` will respect to only one of `B` and `C`.
Here is a simple code
file `A.rs`
```rust
fn eat() {
    apple();
}
```

file `B.rs`
```rust
fn apple() {
}
include!("A.rs");
```

file `C.rs`
```rust
fn apple() {
}
include!("A.rs");
```
When we goto def in `apple()` of `A`, we will only navigate to one of `B` and `C`, not both.

cmyr:
okay yes I think that is an intuitive and reasonable restriction that A) indicates that the user probably has more serious problems and B) I forget what I was going to say for B).


For what it's worth I would probably phrase this as, "A given file can only be tracked via `include!` to a single location." Or something, I found the original phrasing a bit ambiguous.

Ultimately unimportant nitpicking though, thank you for the patch, this will significantly improve our quality of life. :)

Veykril:
Thanks!
Only supporting a file once per include is fine, we can fix that rather easily but I don't think we gain much from that right now.

The mapping logic looks correct to me.

The main issue here is that non-item level includes won't work with this, but this was known from the start given the lazy nature of rust-analyzer so that will most likely never work.

@bors r+

PR #16450, opened 2024-01-30
internal: Prepare parser interface for editions

Urhengulas:


Veykril:
The one in `project-model` exists for rust-project.json deserialization, `base-db` is the representation for crates and the `parser` one is separate because it does not depend on `base-db` (and shouldn't). On another note, the `base-db` one we might want to move to the `span` crate, I think that makes more sense there. The parser might depend on the span crate in the future in which case we can collapse them, but right now that makes little sense

Veykril:
Rebased and deduplicated the enum

Veykril:
Thanks!
@bors r+

crates/hir-ty/src/method_resolution.rs
```diff
@@ -1157,7 +1157,7 @@ fn iterate_trait_method_candidates(
         {
             // FIXME: this should really be using the edition of the method name's span, in case it
             // comes from a macro
-            if db.crate_graph()[krate].edition < Edition::Edition2021 {
+            if db.crate_graph()[krate].edition < Edition::CURRENT {
```
lnicola, line 1160: Shouldn't this stay as `Edition2021`?

Veykril: Good catch, I misread the comment

Urhengulas:
🙏🏾

PR #16451, opened 2024-01-30
internal: Work through temporarily allowed clippy lints, part 2

Urhengulas:
Another follow-up to https://github.com/rust-lang/rust-analyzer/pull/16401.


Veykril (approved):
Seems like an import is missing or something

Urhengulas:
> Seems like an import is missing or something

Fixed

crates/hir-def/src/hir/format_args.rs
```diff
@@ -177,9 +177,9 @@ pub(crate) fn parse(
     let text = s.text_without_quotes();
     let str_style = match s.quote_offsets() {
         Some(offsets) => {
-            let raw = u32::from(offsets.quotes.0.len()) - 1;
+            let raw = usize::from(offsets.quotes.0.len()) - 1;
             // subtract 1 for the `r` prefix
-            (raw != 0).then(|| raw as usize - 1)
```
Veykril, line 182: This is no longer lazy, so this always underflows when raw = 0

Veykril: Seems weird for clippy to lint this as arithmetic is side effectful in debug builds

crates/ide-completion/src/context/analysis.rs
```diff
@@ -1267,8 +1267,7 @@ fn pattern_context_for(
     pat
         .syntax()
         .ancestors()
-        .skip_while(|it| ast::Pat::can_cast(it.kind()))
-        .next()
```
Veykril, line 1271: This is wrong

crates/hir-def/src/hir/format_args.rs
Urhengulas: Yes, I am also a bit disappointed. But I guess in 99% of the cases doing it not lazy will have the correct behavior.

crates/flycheck/src/lib.rs
```diff
@@ -538,7 +540,7 @@ impl CargoActor {
 }
 
 enum CargoMessage {
-    CompilerArtifact(cargo_metadata::Artifact),
+    CompilerArtifact(Box<cargo_metadata::Artifact>),
```
lnicola, line 543: This came up a couple of times before and I think bjorn3 pointed put that most of the time we're going to have this variant, so making this change will cause a lot of allocations.

I don't know if the cost of moving the large struct around is lower or higher than that of the allocations, though.

crates/ide-completion/src/context/analysis.rs
Veykril: Oh I need to get my eyes checked, I thought the negation was missing, nevermind then!

crates/flycheck/src/lib.rs
Urhengulas: I am fine with any choice. I was just going with the clippy recommendation here.

I'd propose that we go with it and if someone takes the time to compare the two options we potentially revert it and disable the clippy lint for it.

Veykril: https://github.com/rust-lang/rust-analyzer/pull/16470

PR #16452, opened 2024-01-30
add semantic modifier for logical not

x17jiri:
By having a modifier for `!`, I'll be able to highlight it in vscode with more noticable color without affecting the other logical operators.


crates/ide/src/syntax_highlighting/tags.rs
```diff
@@ -77,6 +77,8 @@ pub enum HlMod {
     Macro,
     /// Mutable binding.
     Mutable,
+    /// Operator `!`
+    Not,
```
Veykril, line 81: Using a modifier for this is the wrong way to tackle this. If anything this should be a new `HlTag`, although I'm unsure of the usefulness of doing this for a specific operator, as that might invite the thought of doing this for every operator out there which is a bit overwhelming.

x17jiri: The advantage of using a modifier is that is does not break existing configurations.

For example if someone has this in `.vscode/settings.json` and `!` would become a different HlTag, the setting would need to be updated. If it is just a modifier, the setting will keep working.
```
        "rules": {
            "logical": {
                "foreground": "#ff0000",
```

And I can simply use:
```
        "rules": {
            "logical.not": {
                "foreground": "#ff0000",
```

I agree it might invite the thought of doing this for other operators, but `!` really is special because it completely changes meaning of code and is easy to miss.

Veykril: We only have a limited amount of modifier slots available which is why we aren't using them like this (it's also not really their purpose). VSCode itself has a mechanism for creating hierarchies for semantic tokens which is what we are using instead, though unfortunately that only works for VSCode. The semantic highlighting part of the LSP spec is very inflexible unfortunately.

I don't agree that it is more special than other operators. Wouldn't coloring this operator via the textmate grammar rule work instead? It's something that can always be matched by a textmate grammar as its purely syntactic and does not need to rely on semantics.

x17jiri: I tried TextMate, but the semantic categories take precedence over TextMate, so I would have to disable the semantic syntax highlighting.

I didn't know the number is limited. Do you think there is another way I can implement this without breaking anyone's configuration?

The visibility of `!` in code is of course subjective, but there are more people who think it's easy to overlook :-) https://www.reddit.com/r/rust/comments/8u8o5o/the_not_operator_in_rust/

Veykril: You can disable `rust-analyzer.semanticHighlighting.operator.enable` which should allow text mate to take precedence.

> I didn't know the number is limited. Do you think there is another way I can implement this without breaking anyone's configuration?

No I don't think so. This is also the reason why a lot of the special (non-standard) tokens are configurable. It's a shame the LSP did not flesh out the idea of semantic tokens properly.

Another alternative for you might be to enable `semanticHighlighting.operator.specialization.enable`, then you could color the `logical` and `bitwise` operators (though that will of course color more than just `!`).

x17jiri: > You can disable `rust-analyzer.semanticHighlighting.operator.enable` which should allow text mate to take precedence.

That seems like a good workaround. Thanks

x17jiri: @Veykril Sorry to bother you with this again, but I'm not happy with how the workaround works :-)

What would be your opinion on this:
- add `logicalNot` HlTag
- add `rust-analyzer.semanticHighlighting.operator.specialization.logicalNot.enable` option

The change is a little bigger than this PR, but nothing horrible

Veykril: Not too happy with that to be fair :sweat_smile: Adding this would imply that we'd potentially add this for every single operator out there if someone wants it and that to me seems to be too much specialization for a format that notably doesn't support hierarchies of "classifications". We'd therefore need to add a config for enabling/disabling each of those as well which feels odd given that's the job of the theme really.

Issue #10970, opened 2021-12-09
Feature Request: Diagnostic to replace trailing `return <val>;` with just `<val>`


izik1:
Example:
```rs
#[derive(Copy, Clone, Eq, PartialEq)]
pub struct Id(pub i32);

pub fn path(next: &[Id], width: usize, u: Id, v: Id) -> Vec<Id> {
    let mut u = u;
    let mut path = vec![];
    while u != v {
        u = next[u.0 as usize * width + v.0 as usize];
        path.push(u);
    }
    return<|> path;
}
```
->
```rs
#[derive(Copy, Clone, Eq, PartialEq)]
pub struct Id(pub i32);

pub fn path(next: &[Id], width: usize, u: Id, v: Id) -> Vec<Id> {
    let mut u = u;
    let mut path = vec![];
    while u != v {
        u = next[u.0 as usize * width + v.0 as usize];
        path.push(u);
    }
    <|>path
}
```
(not actually too sure where to put the cursor)

If this _does_ get picked up at some point: edge-case warning: the trailing semi might not exist.

I'd honestly be interested in doing this myself

Veykril:
Instead of an assist this should just be a diagnostic with a fixit.

Relevant code parts to follow are https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/ide_diagnostics/src/lib.rs and https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/hir/src/diagnostics.rs.

Most similar ones to this would probably be the `RemoveThisSemicolon` and/or `ReplaceFilterMapNextWithFindMap` diagnostics, so following how those work should give you insight on how to do this.

gowrizrh:
i'd like to pick this up and work through it

izik1:
So, I have gotten started with this and am still working on it (see: https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/implementing.2010970.20.28diagnostic.20for.20trailing.20.60return.3B.60.29/near/264587951), umm, honestly, I'd like to get this done tonight or tomorrow night, life did what life is oh so good at and distracted me for a while

PR #11020, opened 2021-12-15
feat: Add a diagnostic with a fix to remove `return` at the end of functions

izik1:
#10970, probably not _closes_, because of all the fixmes. But I feel comfortable at least asking for review.
Should I have marked the diagnostic as experimental?

I feel like this should be named along the lines of `unwrap expression` but I can't think of a good name there

PR #16460, opened 2024-01-31
feat: Add diagnostic with fix to replace trailing `return <val>;` with `<val>`

davidsemakula:
Works for functions and closures.
Ignores desugared return expressions (e.g. from desugared try operators).

Fixes: #10970
Completes: #11020


crates/hir/src/diagnostics.rs
```diff
@@ -444,6 +451,20 @@ impl AnyDiagnostic {
                     Err(SyntheticSyntax) => (),
                 }
             }
+            BodyValidationDiagnostic::RemoveTrailingReturn { return_expr } => {
+                if let Ok(source_ptr) = source_map.expr_syntax(return_expr) {
+                    // Filters out desugared return expressions (e.g. desugared try operators).
+                    if ast::ReturnExpr::can_cast(source_ptr.value.kind()) {
```
Veykril, line 457:
```suggestion
                    if let Some(ptr) = source_ptr.value.cast::<ast::ReturnExpr>()) {
```
Then we can put a `Astptr<ast::ReturnExpr>` in the diagnostic and handle that properly

crates/ide-diagnostics/src/handlers/remove_trailing_return.rs
```diff
@@ -0,0 +1,377 @@
...
+        display_range,
+    )
+    .with_fixes(fixes(ctx, d))
+}
+
+fn fixes(ctx: &DiagnosticsContext<'_>, d: &RemoveTrailingReturn) -> Option<Vec<Assist>> {
+    let return_expr = return_expr(ctx, d)?;
+    let stmt = expr_stmt(ctx, d);
+
+    let range = stmt.as_ref().map_or(return_expr.syntax(), AstNode::syntax).text_range();
```
Veykril, line 35: This range might be pointing into a macro expansion, we need to map that up
```diff
    let FileRange { range, file_id } = ctx.sema.original_range_opt(stmt.as_ref().map_or(return_expr.syntax(), AstNode::syntax))?;
```
and then use or check that this file_id is the same as the diagnostic one (only case where this could differ is from includes)

crates/ide-diagnostics/src/handlers/remove_trailing_return.rs
```diff
@@ -0,0 +1,377 @@
...
+    ctx: &DiagnosticsContext<'_>,
+    d: &RemoveTrailingReturn,
+) -> Diagnostic {
+    let display_range = ctx.sema.diagnostics_display_range(InFile {
+        file_id: d.file_id,
+        value: expr_stmt(ctx, d)
+            .as_ref()
+            .map(|stmt| SyntaxNodePtr::new(stmt.syntax()))
+            .unwrap_or_else(|| d.return_expr.into()),
+    });
```
Veykril, line 22: You can use `adjusted_display_range` for this (will need a rebase after https://github.com/rust-lang/rust-analyzer/pull/16459),

Veykril (approved):

Veykril:
Thanks, needs a rebase!
@bors delegate+

PR #16465, opened 2024-01-31
minor: Remove `stdx::eprintln` overwrite

Veykril:


Veykril:
I think we can step this up and use the clippy diagnostic (forgot the name) to disallow usages of the macros and replace the tidy lint fully, now that we run clippy on CI.

Issue #13558, opened 2022-11-06
Extract to function/variable unescapes '\\'


RagnarGrootKoerkamp:
This is similar to #11897, but for extracting to function/variable instead of snippet completion.

**rust-analyzer version**: rust-analyzer 1 (ba28e19b78 2022-10-29)

**rustc version**: rustc 1.62.0-nightly (30f386087 2022-05-05)

**relevant settings**: None, I think

Example:
```
fn main(){
    let a = "\\ $0";
}
```
selecting the enther `let a` line and doing `extract to variable` results in
```
fn main(){
    let var_name = &"\ ";
    let a = var_name;
}
```

I.e., both the `\\` and the `$0` are replaced wrongly.

lnicola:
Fixed in #16467

Issue #11638, opened 2022-03-06
Add structured editing API to insert snipperts


matklad:
Some of our assists don't place cursor correctly:

https://user-images.githubusercontent.com/1711539/156933477-4b7b6e48-4184-473d-8d0a-c0666162e24f.mp4

Here, I'd really want the cursor to end up on A, such that I can rename it. I've looked into quickly fixing this, and it seems we simply don't have an API for that!

What I'd love the API to look like is either this:

```diff
diff --git a/crates/ide_assists/src/handlers/introduce_named_generic.rs b/crates/ide_assists/src/handlers/introduce_named_generic.rs
index 636b05dca..8eb9a462e 100644
--- a/crates/ide_assists/src/handlers/introduce_named_generic.rs
+++ b/crates/ide_assists/src/handlers/introduce_named_generic.rs
@@ -1,5 +1,5 @@
 use syntax::{
-    ast::{self, edit_in_place::GenericParamsOwnerEdit, make, AstNode},
+    ast::{self, edit_in_place::GenericParamsOwnerEdit, make, AstNode, HasGenericParams},
     ted,
 };
 
@@ -39,7 +39,10 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext) ->
             let new_ty = make::ty(&type_param_name).clone_for_update();
 
             ted::replace(impl_trait_type.syntax(), new_ty.syntax());
-            fn_.get_or_create_generic_param_list().add_generic_param(type_param.into())
+            fn_.get_or_create_generic_param_list().add_generic_param(type_param.into());
+            if let Some(generic_param) = fn_.generic_param_list().and_then(|it| it.generic_params().last()) {
+                edit.move_cursor_to(generic_param)
+            }
         },
     )
 }
```

or this

```diff
diff --git a/crates/ide_assists/src/handlers/introduce_named_generic.rs b/crates/ide_assists/src/handlers/introduce_named_generic.rs
index 636b05dca..f361148cb 100644
--- a/crates/ide_assists/src/handlers/introduce_named_generic.rs
+++ b/crates/ide_assists/src/handlers/introduce_named_generic.rs
@@ -34,7 +34,9 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext) ->
 
             let type_param_name = suggest_name::for_generic_parameter(&impl_trait_type);
 
[...]

PR #16467, opened 2024-02-01
internal: Migrate assists to the structured snippet API, part 6/7

DropDemBits:
Continuing from #16082

- `extract_function`
- `generate_getter_or_setter`
- `generate_impl`
- `generate_new`
- `replace_derive_with_manual_impl`

Would've been the final PR in the structured snippet migration series, but I didn't notice that `generate_trait_from_impl` started to use `{insert,replace}_snippet` when I first started the migration 😅. This appears to be a pretty self-contained change, so I'll leave that for a separate future PR.

This also removes the last usages of `render_snippet`, which was a follow up goal of #11638. 🎉

### DIFF
--- a/crates/ide-assists/src/handlers/extract_function.rs
+++ b/crates/ide-assists/src/handlers/extract_function.rs
@@ -1,4 +1,4 @@
-use std::iter;
+use std::{iter, ops::RangeInclusive};
 
 use ast::make;
 use either::Either;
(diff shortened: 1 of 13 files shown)


DropDemBits:
Fixes #13558 too, since the structured snippet API always escapes existing backslashes and dollar signs in code.

Veykril (approved):

Veykril:
Thanks!
@bors r+

Veykril:
Some clippy lints need fixing (we enabled clippy since)
@bors delegate+

PR #16472, opened 2024-02-01
fix: Wrong closure kind inferences for closures with predicates

### LINKED ISSUE #16421: False positive "cannot mutate immutable variable (E0384)" error
**rust-analyzer version**: 0.4.1816-standalone (d410d4a2b 2024-01-21)

**rustc version**: 1.77.0-nightly (d5fd09972 2024-01-22)

Code https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=143f7fbac71c9d440720ae04f8c9018c gives error `cannot mutate immutable variable f` at the line 21, but compiles fine in rustc.

ShoyuVanilla:
Currently, the Rust Analyzer infers the kind(`Fn`, `FnMut` or `FnOnce`) of a closure only with its captures.
But the [Rust compiler utilizes obligations and predicates](https://github.com/rust-lang/rust/blob/11f32b73e0dc9287e305b5b9980d24aecdc8c17f/compiler/rustc_hir_typeck/src/closure.rs#L61)(with [here](https://github.com/rust-lang/rust/blob/11f32b73e0dc9287e305b5b9980d24aecdc8c17f/compiler/rustc_hir_typeck/src/upvar.rs#L192-L268)) to infer its kind whenever available (for example, coercing into a bounded argument), and only uses capture informations when not.
So, this makes false positive errors as we can see in #16421,
or more simpler cases like;
image
which is [compiled and runs](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=00af72541fa3cc1806373c888bb9f885) without any error on Rust compiler.

This PR gathers `FnTrait` predicates from obligations and uses them inside `infer_closures` to infer closure kinds, and fixes some wrong `impl FnX` inferences in preexisting test codes.

Fixes #16421


crates/hir-ty/src/infer/closure.rs
```diff
@@ -957,7 +957,14 @@ impl InferenceContext<'_> {
...
         self.restrict_precision_for_unsafe();
         // closure_kind should be done before adjust_for_move_closure
-        let closure_kind = self.closure_kind();
+        let closure_kind = {
+            let from_capture = self.closure_kind_from_capture();
+            // if predicate.unwrap_or(FnTrait::Fn) < from_capture {
+            //     // Diagnostics here, like compiler does in
+            //     // https://github.com/rust-lang/rust/blob/11f32b73e0dc9287e305b5b9980d24aecdc8c17f/compiler/rustc_hir_typeck/src/upvar.rs#L264
+            // }
+            predicate.unwrap_or(from_capture)
```
Veykril, line 966: Should use `unwrap_or_else(|| self.closure_kind_from_capture())` for lazyiness

crates/hir-ty/src/infer/unify.rs
```diff
@@ -146,6 +147,8 @@ pub(crate) struct InferenceTable<'a> {
     /// Double buffer used in [`Self::resolve_obligations_as_possible`] to cut down on
     /// temporary allocations.
     resolve_obligations_buffer: Vec<Canonicalized<InEnvironment<Goal>>>,
+    fn_trait_predicates: Vec<(Ty, FnTrait)>,
+    cached_fn_trait_ids: Option<CachedFnTraitIds>,
```
Veykril, line 151: Not too happy with this, can't we just iterate through the pending obligations when necessary? (as is what rustc does)

Veykril (commented):
The compiler also inspects the expected type first and foremost, and only then does it deduce_closure_signature_from_predicates if the expected type is a type variable or opaque alias. We probably should do the same (as much as we can, given our infra is not the same), https://github.com/rust-lang/rust/blob/11f32b73e0dc9287e305b5b9980d24aecdc8c17f/compiler/rustc_hir_typeck/src/closure.rs#L214

crates/hir-ty/src/infer/unify.rs
ShoyuVanilla: The chalk resolves obligations like `FnOnce` immediately, so they are not remained in the pending obligations. The most problematic thing is that it infers the closures with `FnOnce` predicate as implementing `FnMut` and `Fn`

ShoyuVanilla:
> The compiler also inspects the expected type first and foremost, and only then does it deduce_closure_signature_from_predicates if the expected type is a type variable or opaque alias. We probably should do the same (as much as we can, given our infra is not the same), https://github.com/rust-lang/rust/blob/11f32b73e0dc9287e305b5b9980d24aecdc8c17f/compiler/rustc_hir_typeck/src/closure.rs#L214

Okay, I'll delve into it more

crates/hir-ty/src/infer/unify.rs
ShoyuVanilla: Maybe I should look into it more, but I'm not sure that it is intended chalk functionality or not

ShoyuVanilla: Anyway, I'm gonna take a deep look into the chalk closure inferences. I might fix this from upstream or maybe RA is asking chalk without some details

ShoyuVanilla: @Veykril I had a closer into RA and chalk, and I think that we couldn't use pending obligations like rustc does.

When the chalk has to solve a goal that whether some closure implements some `FnTrait`;

[1. It asks `RustIrDatabase` for the kind of our closure and checkes if the kind of closure is compatible with the given `FnTrait`](https://github.com/rust-lang/chalk/blob/3b710e1be212912610b0ad4fcefd7f8306f1c33b/chalk-solve/src/clauses/builtin_traits/fn_family.rs#L110-L130)
[2. If so, it pushes the clause for the "FACT" that our closure implements that trait.](https://github.com/rust-lang/chalk/blob/3b710e1be212912610b0ad4fcefd7f8306f1c33b/chalk-solve/src/clauses/builtin_traits/fn_family.rs#L9-L25)

So, solving whether the given closure implements `FnTrait` is heavily depends on `RustIrDatabase` and I think that 's reasonable for it to do so because what chalks wants to be(though it is not going to be actively maintained due to new trait solver in rustc) is [an abstract trait solver](https://rust-lang.github.io/chalk/book/types.html#goal-of-the-chalk-ir-crate), not a very definite rustc inference maker.

But since RA's `ChalkContext`, which is our `RustIrDatabase` implementation, [always returns `Fn` as the kind of a closure](https://github.com/rust-lang/rust-analyzer/blob/a01655552df5a4f7ef2db8296d3ad15746faa22c/crates/hir-ty/src/chalk_db.rs#L359-L366). (I think that this is sort of a heuristic to avoid having other inferences and coercions delayed or unresolved.)
Thus, the obligations like `closure{n}[...] implements FnX` are always solved into `Unique` solutions and not remain as "pending" because they are immediately solved.

Giving the "correct" kind of a closure doesn't work for this because;

1. We need to use "correct" pending obligations for closure kind inferencing
2. But the "correct" pending obligations are what we are going to do with "correct" kind of a closure, a cyclic dependency!

I've tried just giving `FnOnce` as a `closure_kind` but aside from that it makes lots of inferences in our test cases unresolved - if the later were to worked correctly I would have tried fixing those inferences -, it still does not "pend" obligations.
The obligations for closure implementation `FnMut` and `Fn` are solved as `None`, and [they are not pended as well](https://github.com/rust-lang/rust-analyzer/blob/c864953c0f1feeed332c08b8c17192eedbc2535f/crates/hir-ty/src/infer/unify.rs#L675-L707).

So, I have concluded that we would have to modify lots of RA inferencing code to do this as rustc does, and this was why I wrote somewhat sad codes for similar thing 😢  (When I wrote it, I hadn't debugged the chalk much, but tried using pending obligations and it was always empty; so I had similar conclusion without much evidence then 😅 )

crates/hir-ty/src/infer/closure.rs
ShoyuVanilla: I was planning to implement a diagnostic here as the comments above says. I'll try it and will convert into `unwrap_or_else` if it is not possible

ShoyuVanilla:
After some investigations, I found out that I had made some terrible misunderstandings on what rustc does this 😱;

While [this is compiled and runs](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f803b7509b90b237d17bc4c7c2b8a363) as my previous understandings for this issue;
```rust
struct Foo<T: FnOnce()>(T);

fn main() {
    let mut x = 0;
    // let a = || { x = 1; };
    let b = Foo(|| { x = 1 });
    (b.0)();
}
```

[but this is not compiled;](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bdb235ce4fad10f2a8bed90caa760a03)
```rust
struct Foo<T: FnOnce()>(T);

fn main() {
    let mut x = 0;
    let a = || { x = 1; };
    let b = Foo(a);
    (b.0)();
}
```

contrary to my previous misunderstandings, as seen in my wrong test cases.

I'll close this and soon reopen new PR with proper fix, and I think that I could fix the thing in similar way the rustc does, as you said.
Sorry for wasting your time 😢

Veykril:
No problem, and thanks for investigating!

PR #16473, opened 2024-02-01
Stop eagerly resolving inlay hint text edits for VSCode

SomeoneToIgnore:
Send less json over the wire.
After https://github.com/microsoft/vscode/issues/193124 was fixed, this change is not needed anymore.

VSCode 1.86.0 now supports double click for unresolved hint data too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1597,6 +1597,7 @@ dependencies = [
  "rayon",
  "rustc-hash",
  "scip",
+ "semver",
  "serde",
  "serde_json",
  "sourcegen",
--- a/crates/rust-analyzer/Cargo.toml
+++ b/crates/rust-analyzer/Cargo.toml
@@ -42,6 +42,7 @@ triomphe.workspace = true
 nohash-hasher.workspace = true
 always-assert = "0.2.0"
 walkdir = "2.3.2"
+semver.workspace = true
 
 cfg.workspace = true
 flycheck.workspace = true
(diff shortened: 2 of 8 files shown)


lnicola:
You might need to rebase to fix the CI.

@davidbarsky what VS Code are you on?

davidbarsky:
> You might need to rebase to fix the CI.
> 
> 
> 
> @davidbarsky what VS Code are you on?

1.78 right now, but 1.82 in a week or so. We can carry this as a patch, need be.

Veykril:
Fwiw there is version field in the `InitializeParams` that a client can populate. Haven't checked but I'd expect that VSCode populates that. We could read that out and make this conditional on that for a few months. Otherwise if we merge this we should bump the VSCode requirement of the client extension so that we force people to upgrade their VSCode to contain the fix.

SomeoneToIgnore:
I'm not in a rush and can hold this PR, as the whole point of it was to get rid of any code-specific checks.

But if you want to have it nonetheless, let me know and I will push more checks on top.

Veykril:
Rebased (because of clippy issues that were fixed on master) and made the check conditional on the vscode version s owe can just merge it now.

Veykril:
Thanks!
@bors r+

Wilfred:
Looking at the upstream VS Code commit https://github.com/microsoft/vscode/commit/8adf59fc006880fb6a63830e9615f9d7686cabfe, I think the fix shipped in VS Code 1.86 FWIW.

Veykril:
1.86 is what the code in the PR checks for

PR #16474, opened 2024-02-01
internal: even more `tracing`

davidbarsky:
As part of profiling completions, I added some additional spans and moved `TyBuilder::subst_for_def` closer to its usage site (the latter had a small impact on completion performance. Thanks for the tip, Lukas!)

### DIFF
--- a/crates/ide-completion/src/completions/flyimport.rs
+++ b/crates/ide-completion/src/completions/flyimport.rs
@@ -395,6 +395,14 @@ fn import_assets_for_path(
     potential_import_name: &str,
     qualifier: Option<ast::Path>,
 ) -> Option<ImportAssets> {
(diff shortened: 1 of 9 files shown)


Veykril (approved):

crates/ide-completion/src/completions/flyimport.rs
```diff
@@ -395,6 +395,14 @@ fn import_assets_for_path(
     potential_import_name: &str,
     qualifier: Option<ast::Path>,
 ) -> Option<ImportAssets> {
+    let _p = tracing::span!(
+        tracing::Level::INFO,
```
Veykril, line 399: These could just be `tracing::info_span!` instead, though that adds additional nested macro calls so maybe not the best idea

PR #16475, opened 2024-02-02
feat: Support multiple tab stops for completions in VSCode

### LINKED ISSUE #8531: `SnippetTextEdit` interprets `$0` in user code as the cursor position
```rust
    fn prefer_std_paths_over_alloc() {
        let code = r#"
        //- /main.rs crate:main deps:alloc,std
        $0
        "#;
        check_found_path(
            code,
            "std::sync::Arc",
            "std::sync::Arc",
            "std::sync::Arc",
            "std::sync::Arc",
        );
    }
```

Triggering the "inline variable" assist for `code` places the cursor at the `$0` marker and deletes the marker. This shouldn't happen, since the marker is part of user code, not the IDE response.

### LINKED ISSUE #13229: VSCode now supports SnippetTextEdits natively
We should switch to that microsoft/vscode#145374 (comment), https://github.com/rust-lang/rust-analyzer/issues/13213#issuecomment-1245399078

DropDemBits:
Uses the native VSCode support for `SnippetTextEdit`s. Fixes #13229 and fixes #8531.

https://github.com/rust-lang/rust-analyzer/assets/13354275/a2d2c033-bb30-4f34-92ca-bf3f4f744cdc

This is done in a slightly hacky way, as `vscode-languageclient` can't convert RA's `SnippetTextEdit`s into vscode `SnippetTextEdit`s and will appear to use a [different format](https://github.com/microsoft/vscode-languageserver-node/blob/295aaa393fda8ecce110c38880a00466b9320e63/types/src/main.ts#L1501-L1516) in the future.

---

~~Marked as draft since as-is, this will cause completions to double-indent any multi-line code generated.~~
**Update:** This also fixes up edits so that any multi-line code won't be double-indented.

### DIFF
--- a/editors/code/src/commands.ts
+++ b/editors/code/src/commands.ts
@@ -4,7 +4,11 @@ import * as ra from "./lsp_ext";
 import * as path from "path";
 
 import type { Ctx, Cmd, CtxInit } from "./ctx";
-import { applySnippetWorkspaceEdit, applySnippetTextEdits } from "./snippets";
+import {
+    applySnippetWorkspaceEdit,
+    applySnippetTextEdits,
+    type SnippetTextDocumentEdit,
+} from "./snippets";
 import { spawnSync } from "child_process";
 import { type RunnableQuickPick, selectRunnable, createTask, createArgs } from "./run";
 import { AstInspector } from "./ast_inspector";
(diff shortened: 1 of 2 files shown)


Veykril (approved):

DropDemBits:
Added the double indentation workaround.

Initially decided to try and do the workaround in the server, but that turned out to be impossible 😅
Turns out it was easier to do in the client since in there we have access to the indentation that would've been added on on top.

@rustbot ready

Veykril (approved):

Veykril:
Thanks!
@bors r+

Issue #16457, opened 2024-01-31
Upgrade the MacOS runners


lnicola:
The M1 runners are [out of beta now](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/).

We're currently using `macos-11` for releases, which is EOL by Apple and will likely be [dropped](https://github.com/actions/runner-images/issues/9255) in April-June.

I'm not sure what the target version will have to be. We're currently targeting (in theory) Catalina, EOL since 2022.

lnicola:
Done in #18266.

PR #16479, opened 2024-02-03
internal: Use MacOS 12 runners for releases

lnicola:
CC #16457


lnicola:
We could probably bump `MACOSX_DEPLOYMENT_TARGET` too, it's set to 10.15 now, but it's probably 12.7 or something.

lnicola:
Test run at https://github.com/lnicola/rust-analyzer/actions/runs/7768338176.

@bors r+

PR #16482, opened 2024-02-03
Swap Subtree::token_trees from Vec to boxed slice

GnomedDev:
Performs one of the optimizations suggested in #16325, but a little bit more. Boxed slices guarantee `shrink_to_fit` aswell as saving a pointer width as no capacity has to be stored.  

Most of the diff is:
- Changing `vec![]` to `Box::new([])`
- Changing initialize -> fill into fill -> into_boxed_slice
- Working around the lack of an owned iterator or automatic iteration over a `Box<[T]>`

I would like to use my own crate, [small-fixed-array](https://lib.rs/small-fixed-array), although I understand if it isn't mature enough for this. If I'm given the go ahead, I can rework this PR to use it instead.

### DIFF
--- a/crates/hir-expand/src/builtin_fn_macro.rs
+++ b/crates/hir-expand/src/builtin_fn_macro.rs
@@ -361,9 +361,7 @@ fn panic_expand(
     };
 
     // FIXME(slow): quote! have a way to expand to builder to make this a vec!
-    let mut mutable_trees = std::mem::take(&mut call.token_trees).into_vec();
-    mutable_trees.push(tt::TokenTree::Subtree(subtree));
-    call.token_trees = mutable_trees.into_boxed_slice();
+    call.push(tt::TokenTree::Subtree(subtree));
 
     ExpandResult::ok(call)
 }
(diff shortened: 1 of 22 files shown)


crates/hir-expand/src/builtin_fn_macro.rs
```diff
@@ -388,7 +393,12 @@ fn unreachable_expand(
         close: call_site_span,
         kind: tt::DelimiterKind::Parenthesis,
     };
-    call.token_trees.push(tt::TokenTree::Subtree(subtree));
+
+    // FIXME(slow): quote! have a way to expand to builder to make this a vec!
+    let mut mutable_trees = std::mem::take(&mut call.token_trees).into_vec();
+    mutable_trees.push(tt::TokenTree::Subtree(subtree));
```
Veykril, line 399: same here

Veykril (commented):
Overall looks good (except for the `quote` re-allocs, but that isn't too troublesome I reckon). Would be good to do a quick perf measure by running analysis-stats on the repo (I can do that next week otherwise).

GnomedDev:
I've moved the push into a method on Subtree and added the reserve_exact. I'm not sure how to get stable results on `analysis-stats`, so it would probably be better for you to run it.

PR #16484, opened 2024-02-04
fix: preserve where clause when builtin derive

### LINKED ISSUE #16432: There is no suggestion for clone methods and no type inference for the the cloned variables
**rust-analyzer version**: rust-analyzer version: 0.3.1815-standalone (d410d4a2b 2024-01-21)

**relevant settings**:none

while im using burn crate(version = "0.11.1", features = ["wgpu"]),there is something strange

here is the code
```
use burn::{backend::Wgpu, tensor::Tensor};

fn main() {
    let x = Tensor::<Wgpu, 1>::from_data([]);
    let y = x.clone();
}
```

image

there is no suggestion for the clone method

image

and no type inference for y

but the code is ok to build&run, and ra-server gives no log

so... why?

Austaras:
Closes #16432.


crates/hir-ty/src/tests/macros.rs
```diff
@@ -1373,3 +1373,36 @@ pub fn attr_macro() {}
...
+//- minicore: derive, clone
+#[derive(Clone)]
+struct Float;
+
+trait TensorKind: Clone {
+    /// The primitive type of the tensor.
+    type Primitive: Clone;
+}
+
+impl Clone for f64 {}
```
Austaras, line 1390: Don't know why this line is required.

crates/hir-def/src/macro_expansion_tests/builtin_derive_macro.rs
```diff
@@ -157,7 +157,7 @@ where
     generic: Vec<T::InGenericArg>,
 }
 
-impl <T: $crate::clone::Clone, > $crate::clone::Clone for Foo<T, > where T: Trait, T::InFieldShorthand: $crate::clone::Clone, T::InGenericArg: $crate::clone::Clone, {
+impl <T: $crate::clone::Clone, > $crate::clone::Clone for Foo<T, > where <T as Trait>::InWc: Marker, T: Trait, T::InFieldShorthand: $crate::clone::Clone, T::InGenericArg: $crate::clone::Clone, {
```
Austaras, line 160: This aligns with rustc.

crates/hir-ty/src/tests/macros.rs
lnicola: Not sure, we do have `impl const Clone for f64` in `minicore`.

Veykril: You need to enable `builtin_impls`, that is `//- minicore: derive, clone, builtin_impls`

Veykril (approved):
with the test fixed

Veykril:
Thanks!
@bors r+

PR #16487, opened 2024-02-04
Remove unnecessary `.as_ref()` in generate getter assist

### LINKED ISSUE #12389: "Generate getter" assist unnecessarily uses `.as_ref()` when deref coercion would work
```rust
struct S {
    v$0: Vec<u8>,
}
```

Generates:

```rust
impl S {
    fn v(&self) -> &[u8] {
        self.v.as_ref()
    }
}
```

It would probably be better if the body of `v` was just `&self.v` instead of going through `.as_ref()`.

ShoyuVanilla:
Resolves #12389

### DIFF
--- a/crates/ide-assists/src/handlers/generate_getter_or_setter.rs
+++ b/crates/ide-assists/src/handlers/generate_getter_or_setter.rs
@@ -75,10 +75,17 @@ pub(crate) fn generate_setter(acc: &mut Assists, ctx: &AssistContext<'_>) -> Opt
 // Generate a getter method.
 //
 // ```
-// # //- minicore: as_ref
+// # //- minicore: as_ref, deref
 // # pub struct String;
 // # impl AsRef<str> for String {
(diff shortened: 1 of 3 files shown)


lnicola:
Unrelated, but I think we're still not handling `DerefMut`,  right?

ShoyuVanilla:
> Unrelated, but I think we're still not handling `DerefMut`, right?

Yes, [but its return type and return values are merely handled as `&mut T` and `&mut self.value`.](https://github.com/rust-lang/rust-analyzer/blob/7fb639ffc18928cfc2fcdc4a0839b6e05bae4298/crates/ide-assists/src/handlers/generate_getter_or_setter.rs#L207-L211) If similar checks is needed, I think that we could check for `AsMut` and coerce to mutable reference 😄

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16489, opened 2024-02-04
feat!: create alias when renaming an import.

### LINKED ISSUE #15858: Rename type import to alias
If I have `use Type;` imported, I would like to be able to refactor that to `use Type as CustomName;` (in VS Code). Then, I would like all instances that previously used `Type` in that file to now use `CustomName`.
Is this already possible in VS Code somehow?

UserIsntAvailable:
gif

Closes #15858

Implemented:
- [x] - Prevent using `reserved` keywords (e.g self) and `_`.
- [x] - Rename other modules that might be referencing the import.
- [x] - Fix "broken" tests.
- [ ] - Rename **only** "direct" references.
- [ ] - Test more cases.

Future possibilities:
1. Also support `extern crate <name>` syntax.
2. Allow aliasing `self` when it is inside an `UseTreeList`.
~3. If import path already has an alias, "rename" the alias.~
~[4. Create alias even if path is not the last path segment.](https://github.com/rust-lang/rust-analyzer/pull/16489#issuecomment-1930541697)~

### DIFF
--- a/crates/ide/src/rename.rs
+++ b/crates/ide/src/rename.rs
@@ -138,7 +138,7 @@ pub(crate) fn rename(
                         return rename_to_self(&sema, local);
                     }
                 }
-                def.rename(&sema, new_name, rename_external)
+                def.rename(&sema, new_name)
             })
             .collect(),
     };
@@ -2695,7 +2695,8 @@ use qux as frob;
 //- /lib.rs crate:lib new_source_root:library
 pub struct S;
 //- /main.rs crate:main deps:lib new_source_root:local
-use lib::S$0;
+use lib::S;
+fn main() { let _: S$0; }
 "#,
             "error: Cannot rename a non-local definition",
         );
--- a/crates/ide/src/rename.rs
+++ b/crates/ide/src/rename.rs
@@ -9,6 +9,7 @@ use ide_db::{
     base_db::{FileId, FileRange},
     defs::{Definition, NameClass, NameRefClass},
     rename::{bail, format_err, source_edit_from_references, IdentifierKind},
+    source_change::SourceChangeBuilder,
     RootDatabase,
 };
 use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


alibektas:
:wave: Thank you for this PR! I have also seen the your [post](https://github.com/rust-lang/rust-analyzer/issues/15858#issuecomment-1928184388) in the related issue. I am not well acquainted with the inner workings of the renaming infra and honestly I have only taken a short glance at it but maybe you could enlighten me on sth: Why don't you have your aliasing implementation propagate all the way through `ide_db::rename::rename_references` so that,

- You don't have to rewrite some of the stuff that is already there
- If/when a certain thing about renaming changes, so your part of the code adapts to the changes too.
- The actual problem you named would probably be resolved by doing this too, or not? I will have to think about this last point so it may be possible that you see an newer version of this post sometime in the next couple of hours. 

Ofc the editing that an alias would cause is very different to the import statement but on the other hand not so different for the defs affected by it, something we can leverage. 

And one more thing to keep in mind : We don't like `unwrap`s. Whenever possible you should avoid using it and handle it instead.

crates/ide/src/rename.rs
```diff
@@ -131,6 +160,46 @@ pub(crate) fn will_rename_file(
     Some(change)
 }
 
+// FIXME: Allow usage with `extern crate`?
```
Veykril, line 163: Yes, we should support that all the same

crates/ide/src/rename.rs
```diff
@@ -131,6 +160,46 @@ pub(crate) fn will_rename_file(
...
+fn alias_fallback(
+    sema: &Semantics<'_, RootDatabase>,
+    syntax: &SyntaxNode,
+    FilePosition { file_id, offset }: FilePosition,
+    new_name: &str,
+) -> RenameResult<Option<SourceChange>> {
+    match sema.find_node_at_offset_with_descend::<ast::UseTree>(syntax, offset) {
```
Veykril, line 170: Let's not try to handle macros here for now. That will make things more complicated and won't trivially work when the rename happens I reckon (adding a test for that otherwise would be good).

crates/ide/src/rename.rs
```diff
@@ -131,6 +160,46 @@ pub(crate) fn will_rename_file(
...
+        Some(use_tree) => {
+            if let Some(_rename) = use_tree.rename() {
+                return Ok(None); // FIXME: path already has an alias
```
Veykril, line 173: I'd say we should just rename the existing alias in that case. It is a bit weird, but more consistent that way imo

crates/ide/src/rename.rs
```diff
@@ -91,24 +92,52 @@ pub(crate) fn rename(
...
+            .map(|(.., def)| {
+                let mut usages = def.usages(&sema).all();
+
+                // Remove the usage that triggered this rename operation.
+                match usages.references.get_mut(&position.file_id) {
+                    Some(refs) => {
+                        let position = refs
+                            .iter()
+                            .position(|ref_| ref_.range.contains_inclusive(position.offset));
+                        never!(position.is_none());
```
Veykril, line 108: This never doesn't really do anything given we unwrap after anyways

Veykril:
> 4. Create alias even if path is not the last path segment.

This I don't think we should do as that is a more invasive change in the end (splits the imports etc).

Veykril:
The problem of usages relying on the "re-export" of an import is tricky indeed. I am not sure if we already have everything in place to do that properly yet (I don't think we do), as that requires keeping track of what import a usage refers to (this we have), and then resolving the imports step by step (this we are still missing I believe) such that we can check whether the renamed import affects the usage or not.

Though we most likely want a new flag on the usage search for that.

UserIsntAvailable:
> I am not sure if we already have everything in place to do that properly yet (I don't think we do), as that requires keeping track of what import a usage refers to (this we have), and then resolving the imports step by step (this we are still missing I believe) such that we can check whether the renamed import affects the usage or not.

Ok that sounds unfortunate; I'm gonna try implement it on these next days. I gave a look at `FindUsages::search` and there is a lot to take into account :sweat_smile:.

Thanks for the heads up though :+1:.

Veykril:
It's fine if we can't handle that case properly for now though,as long as we open an issue / fixme for that. Most people will use absolute paths for things, especially if they have the default rust-analyzer settings.

alibektas:
@UserIsntAvailable this is mostly my fault, sorry! Whatever you can't/won't do I can do afterwards.

crates/ide/src/rename.rs
UserIsntAvailable: Sure, decided to remove macro support ~I didn't knew `find_node_at_offset_with_descend` was supposed to work also with macros~.

crates/ide/src/rename.rs
UserIsntAvailable: Yeah, seemed weird, but I guess consistency is more important.

crates/ide/src/rename.rs
UserIsntAvailable: It seems that this is also blocked on `usage` not being able to find references to `extern crate` modules.

UserIsntAvailable:
@Veykril

> It's fine if we can't handle that case properly for now though,as long as we open an issue / fixme for that.

I'm fine with that; I just didn't want to leave this on a somewhat broken state.

@alibektas

> this is mostly my fault, sorry! Whatever you can't/won't do I can do afterwards.

Don't worry about it! I wasn't really sure how to if it was fine to merge this as is, then do the other necessary changes on another PR. If you want to do it I don't have any problem with it!

Veykril (commented):
Seems good to me, as noted by the FIXME, this can be done more properly via `ide_db::rename_reference` once we have https://github.com/rust-lang/rust-analyzer/issues/14079 fully done. I believe we can already implement the renaming for `extern crate foo` correctly as we have that in the `Definition`.

Veykril:
Is this considered ready from your PoV? (asking since its still a draft)

UserIsntAvailable:
@Veykril Oops, I thought I marked it as ready. :+1:

Veykril (approved):

Veykril:
Rebased it, thanks!
@bors r+

PR #16493, opened 2024-02-05
fix: emit parser error for missing argument list

Young-Flash:
close https://github.com/rust-lang/rust-analyzer/issues/16362


Veykril:
@bors r+
Thanks!

PR #16497, opened 2024-02-06
Fix incorrect inlining of functions that come from MBE macros

evertedsphere:
Partial fix for https://github.com/rust-lang/rust-analyzer/issues/16471. 

As a reminder, there are two issues there:
1. missing whitespace in parameter types (the first test)
2. the `self` parameter not being replaced by `this` in the function body (the second test) 

The first part is fixed in this PR. See [this comment](https://github.com/rust-lang/rust-analyzer/pull/16497#issuecomment-1934243409) for the second.


lnicola:
Rebasing should fix the CI.

evertedsphere:
It looks like when the function body comes from a macro, `usages_for_locals` can't find the `self` parameter in it and so the renaming fails here:

https://github.com/rust-lang/rust-analyzer/blob/57a454233c57e297450f9f9990b1cad8ad8c845c/crates/ide-assists/src/handlers/inline_call.rs#L361-L370

Specifically, in my second test (at the end of the file), we have
```rs
        self.value.$0hash2(state) // function comes from macro, fails to inline
```
but, as can be seen from other tests, a small change makes it pass:
```rs
        self.value.$0write2_u64(state) // function was written by hand, inlines correctly
```
Perhaps there is some kind of analysis that needs to be run on the function body if it comes from a macro so it won't be "opaque" and we can run usage analysis on it?

Veykril:
The usage only exists in a macro expansion so usage search can't find it. That is currently expected and non-trivial to fix as we can't look through all macro expansions that there are for obvious performance reasons. We could special case this search for locals as the scope of macro expansions is contained to the body the local belongs to, implementing that is already tricky though I believe.

evertedsphere:
@Veykril would you be okay with me removing the second test and merging the other fix for now then? I can try to do the other one separately.

Veykril:
That's fine yes

lnicola:
Can you fix the Clippy warning?

Veykril:
That clippy issue should be fixed on master already, thanks!
@bors r+

PR #16499, opened 2024-02-06
internal: Update rustc_pattern_analysis dependency

Nadrieril:
Just bumping the dependency, as I've been making API changes over on the rustc side. More API changes incoming in the coming weeks.

One benefit of this: we no longer abort in the `DeconstructedPat: Debug` impl, which means we can use `tracing` to investigate issues.

### DIFF
--- a/crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
+++ b/crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
@@ -9,7 +9,7 @@ use rustc_pattern_analysis::{
     index::IdxContainer,
     Captures, TypeCx,
 };
-use smallvec::SmallVec;
+use smallvec::{smallvec, SmallVec};
 use stdx::never;
 use typed_arena::Arena;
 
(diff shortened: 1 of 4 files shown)


crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
```diff
@@ -456,11 +461,27 @@ impl<'p> TypeCx for MatchCheckCtx<'p> {
...
     ) -> fmt::Result {
-        // FIXME: implement this, as using `unimplemented!()` causes panics in `tracing`.
+        let variant =
+            pat.ty().as_adt().and_then(|(adt, _)| Self::variant_id_for_adt(pat.ctor(), adt));
+
+        let db = pat.data().unwrap().db;
+        if let Some(variant) = variant {
+            match variant {
+                VariantId::EnumVariantId(v) => {
+                    write!(f, "{}", db.enum_variant_data(v).name.display(db.upcast()))?;
```
lnicola, line 475: Nit: we should probably clean up these single-argument `write!`s, but we've got a bunch of them, so it doesn't matter right now.

PR #16502, opened 2024-02-06
feat: Add "unnecessary else" diagnostic and fix

### LINKED ISSUE #9457: Add "unnecessary else" diagnostic
```rust
if foo {
    return bar;
} else {
    do_something_else();
}
```
In this snippet the `else {}` block surrounding the `do_something_else()` call is unnecessary and can be omitted. We want to emit a weak diagnostic with a corresponding fixit for this.

Note that the same can apply for `continue` and `break` expressions here and not just `return`.

davidsemakula:
Fixes #9457


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16503, opened 2024-02-06
Move salsa fork in-tree

Veykril:
No one else is supposed to rely on it anyways, this makes it easier to edit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -222,18 +222,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "console"
-version = "0.15.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0e1f83fc076bd6dd27517eacdf25fef6c4dfe5f1d7448bafaaf3a26f13b5e4eb"
-dependencies = [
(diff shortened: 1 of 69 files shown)


Veykril:
TODO: Replace insta with expect-test to cut down on dev-deps

davidbarsky:
oh, thanks for doing this, i was going to send a PR with this, but i'm glad you got there first!

lnicola:
I was worried this would slow down the builds, but it doesn't seem noticeable.

lnicola:
@dbarsky does `tracing` have something like https://docs.rs/test-log/?

image

davidbarsky:
> @dbarsky does `tracing` have something like https://docs.rs/test-log/?
> 
> 
> 
> image
> 
> 
> 
> EDIT: nvm, that's literally `test-log` :grinning:.

slightly incorrect name, but yes, test-log does this.

Veykril:
> I was worried this would slow down the builds, but it doesn't seem noticeable.

Why should it?

lnicola:
We exclude the workspace crates from the GitHub cache.

Issue #12882, opened 2022-07-26
Support flychecking only the current crate


P1n3appl3:
## The issue with checking all crates
Currently when rust-analyzer runs a flycheck it always runs on the entire workspace. In large workspaces (or in my case a large rust-project.json project) running a check build on many crates can significantly increase the time it takes to get results for the current crate. This is especially an issue in build systems that don't support cargo's functionality of streaming diagnostics over stdout as it goes, so if 1 crate in the workspace takes on the order of minutes to run a check build and the rest take milliseconds, you end up having to wait minutes every time a transitive dependency of that long-check-time crate changes to get diagnostics in your editor. Even if you do have streaming diagnostics, not knowing which crate has been modified means the build system can't schedule the check builds to prioritize the one you're actually editing, which causes the same problem of waiting for lots of other crates to check and harming your time-to-first-diagnostic.

## Potential solution
As I understand it rust-analyzer has the required info (the path of the modified file) to only run a check build on the single crate that changed. In the case of a cargo command it could do this automatically by mapping the changed file to the changed crate and passing `-p crate_name` rather than `--workspace` to the flycheck subcommand invocation. For rust-project.json based build systems, if `checkOnSave.overrideCommand` were to expose some sort of variable expansions like `$crate_name`, `$source_root`, and/or `$changed_file` then it would be possible for GN/Bazel/Buck/etc. to use that info to only perform a check build on the modified crate.

## Example: fuchsia's codebase
In fuchsia's codebase (where we use rust-project.json with GN) we [already have support](https://cs.opensource.google/fuchsia/fuchsia/+/main:tools/devshell/contrib/lib/rust/clippy.py;l=130?q=clippy.py) for this "map `$changed_file` to the changed target and only run a check build on that crate" functionality, but it can't be used from the editor because there's no way for rust-analyzer to pass the path of the modified file through to our check build wrapper script. Passing that info along makes the difference between "check on save" being viable or not for our developers due to the latency of checking potentially thousands of crates (including outliers that take minutes to check) vs just one.

## Config/UI concerns
[...]

PR #16510, opened 2024-02-08
feat: Allow cargo check to run on only the current package

Veykril:
cc https://github.com/rust-lang/rust-analyzer/pull/13336, https://github.com/rust-lang/rust-analyzer/issues/12882

### DIFF
--- a/crates/rust-analyzer/src/reload.rs
+++ b/crates/rust-analyzer/src/reload.rs
@@ -503,7 +503,7 @@ impl GlobalState {
             let mut crate_graph_file_dependencies = FxHashSet::default();
 
             let mut load = |path: &AbsPath| {
-                let _p = tracing::span!(tracing::Level::INFO, "switch_workspaces::load").entered();
+                let _p = tracing::span!(tracing::Level::DEBUG, "switch_workspaces::load").entered();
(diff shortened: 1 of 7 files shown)


crates/rust-analyzer/src/reload.rs
```diff
@@ -503,7 +503,7 @@ impl GlobalState {
             let mut crate_graph_file_dependencies = FxHashSet::default();
 
             let mut load = |path: &AbsPath| {
-                let _p = tracing::span!(tracing::Level::INFO, "switch_workspaces::load").entered();
+                let _p = tracing::span!(tracing::Level::DEBUG, "switch_workspaces::load").entered();
```
davidbarsky, line 506: it's up to you, but this change means that `hprof` won't pick up on this span.

Veykril: My base logs became completely useless due to that trace hence the change. Should be fine not to have that included in hprof

lnicola:
@Veykril do you want to keep #12882 open?

Veykril:
Not sure, this only solves the problem for cargo based workspaces. Not for full command overrides.

PR #16519, opened 2024-02-09
internal: Enable some minor lints that we should tackles (take2)

tetsuharuohzeki:
This enables these lints:

- borrowed_box
- derived_hash_with_manual_eq
- forget_non_drop
- needless_doctest_main

### DIFF
--- a/crates/hir-def/src/lib.rs
+++ b/crates/hir-def/src/lib.rs
@@ -721,6 +721,9 @@ impl Clone for Box<dyn OpaqueInternableThing> {
 pub struct InTypeConstId(salsa::InternId);
 impl_intern!(InTypeConstId, InTypeConstLoc, intern_in_type_const, lookup_intern_in_type_const);
 
+// We would like to set `derive(PartialEq)`
+// but the compiler complains about that `.expected_ty` does not implement the `Copy` trait.
+#[allow(clippy::derived_hash_with_manual_eq)]
 #[derive(Debug, Hash, Eq, Clone)]
 pub struct InTypeConstLoc {
     pub id: AstId<ast::ConstArg>,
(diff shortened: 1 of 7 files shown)


crates/hir-def/src/lib.rs
```diff
@@ -721,6 +721,9 @@ impl Clone for Box<dyn OpaqueInternableThing> {
 pub struct InTypeConstId(salsa::InternId);
 impl_intern!(InTypeConstId, InTypeConstLoc, intern_in_type_const, lookup_intern_in_type_const);
 
+// We would like to set `derive(PartialEq)`
+// but the compiler complains about that `.expected_ty` does not implement `Copt` trait.
```
lnicola, line 725:
```suggestion
// but the compiler complains about that `.expected_ty` does not implement the `Copy` trait.
```

crates/ide/src/inlay_hints/implicit_drop.rs
```diff
@@ -1,10 +1,8 @@
 //! Implementation of "implicit drop" inlay hints:
-//! ```no_run
-//! fn main() {
-//!     let x = vec![2];
-//!     if some_condition() {
-//!         /* drop(x) */return;
-//!     }
+//! ```compile_fail
```
lnicola, line 2: We could use `ignore` so we don't try to compile this.

lnicola (approved):

crates/ide/src/inlay_hints/implicit_drop.rs
tetsuharuohzeki: addressed:

- https://github.com/rust-lang/rust-analyzer/pull/16519/commits/a3e60e7f7af383d7b067c12cde78c02d8c15ee26
- https://github.com/rust-lang/rust-analyzer/pull/16519/commits/adddd14afb63b1646b4ac5eb4d33a611c9c864d5

crates/hir-def/src/lib.rs
tetsuharuohzeki: addressed: https://github.com/rust-lang/rust-analyzer/pull/16519/commits/d45cabd02900ed8deed37c9bf338f605aa398dd4

tetsuharuohzeki:
@lnicola Thanks. I updated my changeset.

PR #16533, opened 2024-02-11
Update to latest `rustc_pattern_analysis`

Nadrieril:
Here I go again. Two improvements this time.

1. I've removed the need for an arena for patterns. Turns out this wasn't gaining any performance on the rustc side so may as well allocate normally.
2. I've added a clean error path when types don't match, so rustc_pattern_analysis should never panic except internal logic errors. For now `cx.bug()` calls `never!()` but I'm not sure what `never!()` does. Does it display anything to the user? Otherwise a `debug!()` should be sufficient.

Point 2 should fix https://github.com/rust-lang/rust-analyzer/issues/15883 but I haven't tested it because I'm not sure how to reproduce. Could someone give me pointers as to how to write a test for the pattern code?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1470,12 +1470,12 @@ dependencies = [
 
 [[package]]
 name = "ra-ap-rustc_index"
-version = "0.36.0"
+version = "0.37.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f8a41dee58608b1fc93779ea365edaa70ac9927e3335ae914b675be0fa063cd7"
+checksum = "df5a0ba0d08af366cf235dbe8eb7226cced7a4fe502c98aa434ccf416defd746"
 dependencies = [
  "arrayvec",
- "ra-ap-rustc_index_macros 0.36.0",
+ "ra-ap-rustc_index_macros 0.37.0",
  "smallvec",
 ]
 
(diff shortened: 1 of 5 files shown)


HKalbasi:
`never!` panics with debug assertions enabled and logs otherwise. It's good for the `cx.bug()`.

About the test, I can see some panic with this test:
```Rust
check_diagnostics_no_bails(
            r#"
        //- minicore: option
        fn main() {
            match Some((true, false)) {
                Some(true) | Some(false) => {}
                None => {}
            }
        }
            "#,
        );
```

Nadrieril:
Oh perfect, `check_diagnostics_no_bails` is what I was looking for. This does panic indeed, but only because of the call to `never!()`. If I remove it I get the type error that we should get. Does that mean I shold remove the `never` call?

Nadrieril:
I changed it to `debug!` so that I could add a regression test

HKalbasi:
I think ideally `cx.bug()` should never be called and `Err` should handle these cases and we should use `never` here. Alternatively we can consider calling match check only on bodies with successful type check, similar to rustc, to prevent similar panics and bugs.

But removing the `cx.bug` here requires updating the rustc crate, so let's merge this for now.
@bors r+

Nadrieril:
> Alternatively we can consider calling match check only on bodies with successful type check

That was supposed to be the case, this PR is for the edge cases where we fall to catch the type error

PR #16540, opened 2024-02-12
internal: macro_arg query always returns a TokenTree

Veykril:


crates/hir-def/src/macro_expansion_tests/mbe/meta_syntax.rs
```diff
@@ -68,26 +68,26 @@ m2!();
 "#,
         expect![[r#"
 macro_rules! i1 { invalid }
-/* error: invalid macro definition: expected subtree */
+/* error: macro definition has parse errors */
```
lnicola, line 71: Losing the error info here seems a bit unfortunate?

lnicola: Ah, do we get both errors?

Veykril: I did this intentionally as I found it weird that those are reported at the call site since they are def site errors. `expected subtree` is 1) pointless here since you don't know where the subtree is expected in the definition and 2) could be confusing if someone didnt fully read the message, thinking their invocation is wrong.

Veykril:
my push earlier failed ..
@bors r+

Issue #12500, opened 2022-06-10
Feature: Add `Implement trait` assist for `struct`, `enum` etc.


mylvghb:
This feature already exists in the Intellij Rust plugin, can it be integrated into Rust Analyzer?

![image](https://user-images.githubusercontent.com/35945793/173008601-adc8f38b-cdc0-4a61-b5d7-3d980e7140d4.png)

lnicola:
What does it do?

mylvghb:
@lnicola  You and I don't think alike， here's what I think：

**Steps**

1. The  cursor is positioned on the line of the structure or enum decreation.
   
![image](https://user-images.githubusercontent.com/35945793/173013091-e2e2f397-baa6-4008-8a05-76ce0db18541.png)

2. Click `Implement trait`.

![image](https://user-images.githubusercontent.com/35945793/173013248-43b7f64f-835e-41af-b658-ed8898cab0ef.png)

3. The following code is automatically generated and provides a list of suggested traits

![image](https://user-images.githubusercontent.com/35945793/173015352-42292446-0a9a-4a9c-86e9-466a822003e3.png)

lnicola:
Yeah, that's why I asked what it does. Does it also fill in the required members once you pick a trait from the list?

mylvghb:
> Yeah, that's why I asked what it does. Does it also fill in the required members once you pick a trait from the list?

Yeah, the required members need to be filled.

mylvghb:
**Note**: This feature also applies to generic struct

mylvghb:
**Note:** This feature also applies to generic structs
![image](https://user-images.githubusercontent.com/35945793/173020224-323656c6-60c8-49f4-9550-b903c7c2b379.png)

lnicola:
We can insert the `impl` block, but I don't think we can trigger the completion there.

mylvghb:
> We can insert the `impl` block, but I don't think we can trigger the completion there.

Okay!

jhgg:
> We can insert the `impl` block, but I don't think we can trigger the completion there.

I do wonder though if we can filter completions in this case:
```
impl $0 for T {

}
```

Can we offer completions that are only traits in that position? We don't seem to do that now.

Veykril:
We absolutely can

estnml:
hi ,any progress on this? This would increase DX very much

dfireBird:
[...]

PR #16544, opened 2024-02-12
Add completions to show only traits in trait `impl` statement

dfireBird:
This is prerequisite PR for adding the assist mentioned in #12500 

P.S: If wanted, I will add the implementation of the assist in this PR as well.

### DIFF
--- a/crates/ide-completion/src/completions/flyimport.rs
+++ b/crates/ide-completion/src/completions/flyimport.rs
@@ -239,7 +239,7 @@ fn import_on_the_fly(
                 if matches!(location, TypeLocation::TypeBound) {
                     matches!(ty, ModuleDef::Trait(_))
                 } else if matches!(location, TypeLocation::ImplTrait) {
-                    matches!(ty, ModuleDef::Trait(_)) || matches!(ty, ModuleDef::Module(_))
+                    matches!(ty, ModuleDef::Trait(_) | ModuleDef::Module(_))
                 } else {
                     true
                 }
--- a/crates/ide-completion/src/completions/flyimport.rs
+++ b/crates/ide-completion/src/completions/flyimport.rs
@@ -238,6 +238,8 @@ fn import_on_the_fly(
             (PathKind::Type { location }, ItemInNs::Types(ty)) => {
                 if matches!(location, TypeLocation::TypeBound) {
                     matches!(ty, ModuleDef::Trait(_))
+                } else if matches!(location, TypeLocation::ImplTrait) {
+                    matches!(ty, ModuleDef::Trait(_)) || matches!(ty, ModuleDef::Module(_))
                 } else {
                     true
                 }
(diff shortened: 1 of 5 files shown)


crates/ide-completion/src/completions/type.rs
```diff
@@ -184,6 +184,16 @@ pub(crate) fn complete_type_path(
                         }
                     }
                 }
+                TypeLocation::ImplTrait => {
+                    acc.add_nameref_keywords_with_colon(ctx);
+                    ctx.process_all_names(&mut |name, def, doc_aliases| {
+                        let is_trait = matches!(def, ScopeDef::ModuleDef(hir::ModuleDef::Trait(_)));
```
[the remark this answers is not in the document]
dfireBird: So I added this but while adding I found some other issues. Of which one I fixed in the same commit.

The other issue is caused by the fly_import call in the `ide-completion::completions::complete_name_ref`, for which I have the fix and tested it with debug run, but I'm not sure how to add unit tests for it. Should it be done to `ide-completion::tests::flyimport`?

crates/ide-completion/src/completions/flyimport.rs
```diff
@@ -238,6 +238,8 @@ fn import_on_the_fly(
             (PathKind::Type { location }, ItemInNs::Types(ty)) => {
                 if matches!(location, TypeLocation::TypeBound) {
                     matches!(ty, ModuleDef::Trait(_))
+                } else if matches!(location, TypeLocation::ImplTrait) {
+                    matches!(ty, ModuleDef::Trait(_)) || matches!(ty, ModuleDef::Module(_))
```
Veykril, line 242:
```suggestion
                    matches!(ty, ModuleDef::Trait(_) | ModuleDef::Module(_))
```

crates/ide-completion/src/completions/type.rs
```diff
@@ -31,6 +31,11 @@ pub(crate) fn complete_type_path(
             ScopeDef::ImplSelfType(_) => location.complete_self_type(),
             // Don't suggest attribute macros and derives.
             ScopeDef::ModuleDef(Macro(mac)) => mac.is_fn_like(ctx.db),
+            ScopeDef::ModuleDef(Trait(_) | Module(_))
+                if matches!(location, TypeLocation::ImplTrait) =>
+            {
+                true
+            }
```
Veykril, line 38:
```suggestion
            ScopeDef::ModuleDef(Trait(_) | Module(_)) => matches!(location, TypeLocation::ImplTrait),
```

Veykril (commented):
Two minor simplifications

crates/ide-completion/src/completions/type.rs
dfireBird: I've used match guard because I want it to fall to this match case [here](https://github.com/dfireBird/rust-analyzer/blob/0e47befaf3cdf30fcf851fd34d4d6fdb4d274382/crates/ide-completion/src/completions/type.rs#L40), if it's not `ImplTrait` location.

Veykril: Oh right, don't mind me then.

Veykril:
Thanks!
@bors r+

PR #16555, opened 2024-02-13
performance: Speed up Method Completions By Taking Advantage of Orphan Rules

davidbarsky:
(Continues https://github.com/rust-lang/rust-analyzer/pull/16498)

This PR speeds up method completions by doing two things without regressing `analysis-stats`[^1]:
- Filter candidate traits prior to calling `iterate_path_candidates` by relying on orphan rules (see below for a slightly more in-depth explanation). When generating completions [on `slog::Logger`](https://github.com/oxidecomputer/omicron/blob/5e9e59c312cd787ba50cf6776f220951917dfa14/common/src/ledger.rs#L78) in `oxidecomputer/omicron` as a test, this PR halved my completion times—it's now 454ms cold and 281ms warm. Before this PR, it was 808ms cold and 579ms warm.
- Inline some of the method candidate checks into `is_valid_method_candidate` and remove some unnecessary visibility checks. This was suggested by @Veykril in [this comment](https://github.com/rust-lang/rust-analyzer/pull/16498#issuecomment-1929864427).

We filter candidate traits by taking advantage of orphan rules. For additional details, I'll rely on @WaffleLapkin's explanation  [from Zulip](https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/Trait.20Checking/near/420942417): 

> A type `A` can only implements traits which
> 1. Have a blanket implementation (`impl<T> Trait for T {}`)
> 2. Have implementation for `A` (`impl Trait for A {}`)
> 
> Blanket implementation can only exist in `Trait`'s crate. Implementation for `A` can only exist in `A`'s or `Trait`'s crate.

Big thanks to Waffle for its keen observation!

---

I think some additional improvements are possible:
- `for_trait_and_self_ty` seemingly does not distinguish between `&T`, `&mut T`, or `T`, resulting in seemingly irrelevant traits like `tokio::io::AsyncWrite` being being included for, e.g., `&slog::Logger`. I don't know they're being considered due to the [autoref/autoderef behavior](https://github.com/rust-lang/rust-analyzer/blob/a02a219773629686bd8ff123ca1aa995fa50d976/crates/hir-ty/src/method_resolution.rs#L945-L962), but I wonder if it'd make sense to filter by mutability earlier and not consider trait implementations that require `&mut T` when we only have a `&T`.
- The method completions [spend a _lot_ of time in unification](https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/Trait.20Checking/near/421072356), and while there might be low-hanging fruit there, it might make more sense to wait for the new trait solver in `rustc`. I dunno.

[^1]: The filtering occurs outside of typechecking, after all.

### DIFF
--- a/crates/hir-ty/src/method_resolution.rs
+++ b/crates/hir-ty/src/method_resolution.rs
@@ -254,6 +254,11 @@ impl TraitImpls {
             .flat_map(|v| v.iter().copied())
     }
 
+    /// Queries whether `self_ty` has potentially applicable implementations of `trait_`.
+    pub fn has_impls_for_trait_and_self_ty(&self, trait_: TraitId, self_ty: TyFingerprint) -> bool {
+        self.for_trait_and_self_ty(trait_, self_ty).next().is_some()
+    }
+
     pub fn all_impls(&self) -> impl Iterator<Item = ImplId> + '_ {
         self.map.values().flat_map(|map| map.values().flat_map(|v| v.iter().copied()))
     }
@@ -1170,7 +1175,7 @@ fn iterate_trait_method_candidates(
         for &(_, item) in data.items.iter() {
             // Don't pass a `visible_from_module` down to `is_valid_candidate`,
             // since only inherent methods should be included into visibility checking.
-            let visible = match is_valid_candidate(table, name, receiver_ty, item, self_ty, None) {
+            let visible = match is_valid_method_candidate(table, name, receiver_ty, item, self_ty) {
                 IsValidCandidate::Yes => true,
                 IsValidCandidate::NotVisible => false,
                 IsValidCandidate::No => continue,
@@ -1414,6 +1419,74 @@ fn is_valid_candidate(
     }
 }
 
+/// Checks whether a given `AssocItemId` is applicable for `receiver_ty`.
+///
+/// This method should *only* be called by [`iterate_trait_method_candidates`],
+/// as it is responsible for determining applicability in completions.
+#[tracing::instrument(skip_all, fields(name))]
+fn is_valid_method_candidate(
+    table: &mut InferenceTable<'_>,
+    name: Option<&Name>,
+    receiver_ty: Option<&Ty>,
+    item: AssocItemId,
+    self_ty: &Ty,
+) -> IsValidCandidate {
+    let db = table.db;
+    match item {
+        AssocItemId::FunctionId(fn_id) => {
+            let data = db.function_data(fn_id);
+
+            check_that!(name.map_or(true, |n| n == &data.name));
+
+            table.run_in_snapshot(|table| {
+                let container = fn_id.lookup(db.upcast()).container;
+                let (impl_subst, expect_self_ty) = match container {
+                    ItemContainerId::ImplId(it) => {
+                        let subst = TyBuilder::subst_for_def(db, it, None)
+                            .fill_with_inference_vars(table)
+                            .build();
+                        let self_ty = db.impl_self_ty(it).substitute(Interner, &subst);
+                        (subst, self_ty)
+                    }
+                    ItemContainerId::TraitId(it) => {
+                        let subst = TyBuilder::subst_for_def(db, it, None)
+                            .fill_with_inference_vars(table)
+                            .build();
+                        let self_ty = subst.at(Interner, 0).assert_ty_ref(Interner).clone();
+                        (subst, self_ty)
+                    }
+                    _ => unreachable!(),
+                };
+
+                check_that!(table.unify(&expect_self_ty, self_ty));
+
+                if let Some(receiver_ty) = receiver_ty {
+                    check_that!(data.has_self_param());
+
+                    let fn_subst = TyBuilder::subst_for_def(db, fn_id, Some(impl_subst.clone()))
+                        .fill_with_inference_vars(table)
+                        .build();
+
+                    let sig = db.callable_item_signature(fn_id.into());
+                    let expected_receiver =
+                        sig.map(|s| s.params()[0].clone()).substitute(Interner, &fn_subst);
+
+                    check_that!(table.unify(receiver_ty, &expected_receiver));
+                }
+
+                IsValidCandidate::Yes
+            })
+        }
+        AssocItemId::ConstId(c) => {
+            check_that!(receiver_ty.is_none());
+            check_that!(name.map_or(true, |n| db.const_data(c).name.as_ref() == Some(n)));
+
+            IsValidCandidate::Yes
+        }
+        _ => IsValidCandidate::No,
+    }
+}
+
 enum IsValidCandidate {
     Yes,
     No,
@@ -1441,6 +1514,8 @@ fn is_valid_fn_candidate(
     }
     table.run_in_snapshot(|table| {
         let container = fn_id.lookup(db.upcast()).container;
+
+        let _p = tracing::span!(tracing::Level::INFO, "subst_for_def").entered();
         let (impl_subst, expect_self_ty) = match container {
             ItemContainerId::ImplId(it) => {
                 let subst =
@@ -1460,6 +1535,7 @@ fn is_valid_fn_candidate(
         check_that!(table.unify(&expect_self_ty, self_ty));
 
         if let Some(receiver_ty) = receiver_ty {
+            let _p = tracing::span!(tracing::Level::INFO, "check_receiver_ty").entered();
             check_that!(data.has_self_param());
 
             let fn_subst = TyBuilder::subst_for_def(db, fn_id, Some(impl_subst.clone()))
@@ -1474,6 +1550,7 @@ fn is_valid_fn_candidate(
         }
 
         if let ItemContainerId::ImplId(impl_id) = container {
+            let _p = tracing::span!(tracing::Level::INFO, "check_item_container").entered();
             // We need to consider the bounds on the impl to distinguish functions of the same name
             // for a type.
             let predicates = db.generic_predicates(impl_id.into());
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -4239,6 +4239,10 @@ impl Type {
         }
     }
 
+    pub fn fingerprint_for_trait_impl(&self) -> Option<TyFingerprint> {
+        TyFingerprint::for_trait_impl(&self.ty)
+    }
+
     pub(crate) fn canonical(&self) -> Canonical<Ty> {
         hir_ty::replace_errors_with_variables(&self.ty)
     }
--- a/crates/ide-completion/src/tests/flyimport.rs
+++ b/crates/ide-completion/src/tests/flyimport.rs
@@ -375,6 +375,135 @@ fn main() {
 }
 
 #[test]
+fn trait_method_fuzzy_completion_aware_of_fundamental_boxes() {
+    let fixture = r#"
+//- /fundamental.rs crate:fundamental
+#[lang = "owned_box"]
+#[fundamental]
+pub struct Box<T>(T);
+//- /foo.rs crate:foo
+pub trait TestTrait {
+    fn some_method(&self);
+}
+//- /main.rs crate:main deps:foo,fundamental
+struct TestStruct;
+
+impl foo::TestTrait for fundamental::Box<TestStruct> {
+    fn some_method(&self) {}
+}
+
+fn main() {
+    let t = fundamental::Box(TestStruct);
+    t.$0
+}
+"#;
+
+    check(
+        fixture,
+        expect![[r#"
+        me some_method() (use foo::TestTrait) fn(&self)
+    "#]],
+    );
+
+    check_edit(
+        "some_method",
+        fixture,
+        r#"
+use foo::TestTrait;
+
+struct TestStruct;
+
+impl foo::TestTrait for fundamental::Box<TestStruct> {
+    fn some_method(&self) {}
+}
+
+fn main() {
+    let t = fundamental::Box(TestStruct);
+    t.some_method()$0
+}
+"#,
+    );
+}
+
+#[test]
+fn trait_method_fuzzy_completion_aware_of_fundamental_references() {
+    let fixture = r#"
+//- /foo.rs crate:foo
+pub trait TestTrait {
+    fn some_method(&self);
+}
+//- /main.rs crate:main deps:foo
+struct TestStruct;
+
+impl foo::TestTrait for &TestStruct {
+    fn some_method(&self) {}
+}
+
+fn main() {
+    let t = &TestStruct;
+    t.$0
+}
+"#;
+
+    check(
+        fixture,
+        expect![[r#"
+        me some_method() (use foo::TestTrait) fn(&self)
+    "#]],
+    );
+
+    check_edit(
+        "some_method",
+        fixture,
+        r#"
+use foo::TestTrait;
+
+struct TestStruct;
+
+impl foo::TestTrait for &TestStruct {
+    fn some_method(&self) {}
+}
+
+fn main() {
+    let t = &TestStruct;
+    t.some_method()$0
+}
+"#,
+    );
+}
+
+#[test]
+fn trait_method_fuzzy_completion_aware_of_unit_type() {
+    let fixture = r#"
+//- /test_trait.rs crate:test_trait
+pub trait TestInto<T> {
+    fn into(self) -> T;
+}
+
+//- /main.rs crate:main deps:test_trait
+struct A;
+
+impl test_trait::TestInto<A> for () {
+    fn into(self) -> A {
+        A
+    }
+}
+
+fn main() {
+    let a = ();
+    a.$0
+}
+"#;
+
+    check(
+        fixture,
+        expect![[r#"
+    me into() (use test_trait::TestInto) fn(self) -> T
+    "#]],
+    );
+}
+
+#[test]
 fn trait_method_from_alias() {
     let fixture = r#"
 //- /lib.rs crate:dep
--- a/crates/ide-db/src/imports/import_assets.rs
+++ b/crates/ide-db/src/imports/import_assets.rs
@@ -1,8 +1,9 @@
 //! Look up accessible paths for items.
 
 use hir::{
-    AsAssocItem, AssocItem, AssocItemContainer, Crate, ItemInNs, ModPath, Module, ModuleDef, Name,
-    PathResolution, PrefixKind, ScopeDef, Semantics, SemanticsScope, Type,
+    db::HirDatabase, AsAssocItem, AssocItem, AssocItemContainer, Crate, HasCrate, ItemInNs,
+    ModPath, Module, ModuleDef, Name, PathResolution, PrefixKind, ScopeDef, Semantics,
+    SemanticsScope, Trait, Type,
 };
 use itertools::{EitherOrBoth, Itertools};
 use rustc_hash::{FxHashMap, FxHashSet};
@@ -517,7 +518,7 @@ fn trait_applicable_items(
     let related_traits = inherent_traits.chain(env_traits).collect::<FxHashSet<_>>();
 
     let mut required_assoc_items = FxHashSet::default();
-    let trait_candidates: FxHashSet<_> = items_locator::items_with_name(
+    let mut trait_candidates: FxHashSet<_> = items_locator::items_with_name(
         sema,
         current_crate,
         trait_candidate.assoc_item_name.clone(),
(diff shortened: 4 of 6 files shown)


crates/ide-db/src/imports/import_assets.rs
```diff
@@ -536,6 +537,29 @@ fn trait_applicable_items(
         required_assoc_items.insert(assoc);
         Some(assoc_item_trait.into())
     })
+    .filter(|candidate_trait_id| {
+        // blanket `Trait` implementations for type `A` can only exist in the crate defining
+        // the trait. However, an implementation for `A`` can only exist in `A`'s or `Trait`'s crate,
+        // which allows us to reduce the search space substantially.
```
WaffleLapkin, line 543: As I later pointer out on zulip, this is not correct for fundamental types. Example (please add something similar as a test):

```rust
struct A;

impl fmt::Display for Box<A> { ... }
impl fmt::Display for &A { ... }
```
Both of those impls [compile](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4e9622b393ae5e8aabbb93133e169703) even though the impls are not in `Display` or `Box`/`&`'s definition crate.

Similarly: 
```rust
struct A;

impl Into<A> for () {
    fn into(self) -> A { A }
}
```
Is also valid, even though both `Into` and unit are foreign.

You need to go through all generics.

WaffleLapkin: Note though that `impl fmt::Display for Vec<A> {  }` does **not** compile. So the crates to search for impls are
1. Trait's definition crate
2. Definition crates for all trait's generic arguments
    - This is recursive for _fundamental types_ — `Into<Box<A>> for ()` is OK, but `Into<Vec<A>> for ()` is NOT
3. Self type definition crate
     - This is recursive for _fundamental types_

davidbarsky: i'l add add a test!

crates/hir-ty/src/method_resolution.rs
```diff
@@ -1167,10 +1167,13 @@ fn iterate_trait_method_candidates(
         // trait, but if we find out it doesn't, we'll skip the rest of the
         // iteration
         let mut known_implemented = false;
-        for &(_, item) in data.items.iter() {
+        'items: for &(_, item) in data.items.iter() {
+            if let AssocItemId::TypeAliasId(_) = item {
+                continue 'items;
+            }
```
Veykril, line 1173: This is unnecessary, `is_valid_method_candidate` will immediately bail out already

crates/hir-ty/src/method_resolution.rs
```diff
@@ -1459,6 +1533,7 @@ fn is_valid_fn_candidate(
 
         check_that!(table.unify(&expect_self_ty, self_ty));
 
+        let _p = tracing::span!(tracing::Level::INFO, "check_receiver_ty").entered();
```
Veykril, line 1536: Wouldn't it make more sense to have these two spans inside the respective following if expressions instead of in front?

crates/hir/src/lib.rs
```diff
@@ -266,6 +266,10 @@ impl Crate {
         let data = &db.crate_graph()[self.id];
         data.potential_cfg_options.clone().unwrap_or_else(|| data.cfg_options.clone())
     }
+
+    pub fn id(&self) -> CrateId {
+        self.id
+    }
```
Veykril, line 272: There are `From` impls to go from hir type to id, no need for this

crates/hir/src/lib.rs
```diff
@@ -3691,7 +3695,7 @@ pub enum CaptureKind {
 #[derive(Clone, PartialEq, Eq, Debug, Hash)]
 pub struct Type {
     env: Arc<TraitEnvironment>,
-    ty: Ty,
+    pub ty: Ty,
```
Veykril, line 3698: This should stay private, we don't want to leak hir-ty `Ty`pes into the ide layer

crates/ide-db/src/imports/import_assets.rs
```diff
@@ -536,6 +537,32 @@ fn trait_applicable_items(
...
+        let candidate_trait_id: TraitId = *candidate_trait_id;
+        let definining_crate_for_trait = db.trait_environment(candidate_trait_id.into()).krate;
```
Veykril, line 545:
```suggestion
        let definining_crate_for_trait = Trait::from(candidate_trait_id).krate();
```

crates/ide-db/src/imports/import_assets.rs
```diff
@@ -536,6 +537,32 @@ fn trait_applicable_items(
...
+
+        let Some(receiver_fingerprint) =
+            TyFingerprint::for_trait_impl(&trait_candidate.receiver_ty.ty)
+        else {
+            return false;
+        };
+        let definitions_exist_in_trait_crate = db
+            .trait_impls_in_crate(definining_crate_for_trait)
+            .for_trait_and_self_ty(candidate_trait_id, receiver_fingerprint)
```
Veykril, line 554: Might be nice to add a method `has_impls_for_trait_and_self_ty(..) -> bool`

crates/hir-ty/src/method_resolution.rs
```diff
@@ -1414,6 +1417,75 @@ fn is_valid_candidate(
...
+    table: &mut InferenceTable<'_>,
+    name: Option<&Name>,
+    receiver_ty: Option<&Ty>,
+    item: AssocItemId,
+    self_ty: &Ty,
+) -> IsValidCandidate {
+    let db = table.db;
+    match item {
+        AssocItemId::FunctionId(fn_id) => {
+            let db = table.db;
```
lnicola, line 1435: Nit: this is unused (there's another one above).

crates/hir-ty/src/method_resolution.rs
davidbarsky: done!

crates/hir-ty/src/method_resolution.rs
davidbarsky: yup, it would.

crates/hir/src/lib.rs
davidbarsky: removed!

crates/hir/src/lib.rs
davidbarsky: removed!

crates/hir-ty/src/method_resolution.rs
davidbarsky: whoops, you're right. removed.

davidbarsky:
Right, okay: I addressed the review feedback, except the big one from Waffle. I'm pretty sure I've got a handle on checking fundamental types (I'll pilfer from it from the orphan implementation check), but I'm struggling to get a test with fundamental types working (see the changes in `crates/ide-completion/src/tests/flyimport.rs`). The expect side of things isn't the correct correct, but I can't seem to get the fixtures to be aware of a `#[fundamental]` `Box`—any tips?

Veykril:
Aren't you missing the fundamental check?

davidbarsky:
I am, yeah. I just want to write some tests that invoke fundamental trait impls and then ensure the impl works afterwards (in this PR, to be clear).

Veykril:
Oh wait, your test fixture is not using the fundamental type in any impl, so how should it show anything?

davidbarsky:
Maybe a better of phrasing my question is "do I need to do anything special to introduce a new fundamental type in a fixture, or is the `#[fundamental]` annotation sufficient". Looking at some of the other tests, I'm inclined to say no.

Veykril:
Just using the attribute is enough yes

davidbarsky:
Alright, I updated the PR to include tests for `impl ForeignTrait for Box<LocalStruct>`, `impl ForeignTrait for &LocalStruct`, and `impl Into<A> for ()`. The first two tests pass without issue (in that the filtering already considers fundamental types), but the third test seemingly _does not_ work, despite `.into()` showing up in completions when I build a rust-analyzer binary with these exact changes. 

Notably, I _didn't_ need to make additional changes when filtering to consider fundamental types.

crates/ide-completion/src/tests/flyimport.rs
```diff
@@ -374,6 +374,129 @@ fn main() {
...
+struct A;
+
+impl Into<A> for () {
+    fn into(self) -> A {
+        A
+    }
+}
+
+fn main() {
+    let a = A;
```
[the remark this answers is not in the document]
davidbarsky: I don't believe it is, based on Waffle's comment: https://github.com/rust-lang/rust-analyzer/pull/16555#discussion_r1488545233. 

Note that this test also passes _without_ my filtering changes in `crates/ide-db/src/imports/import_assets.rs`, so I'm inclined to investigate this detail later.

crates/ide-db/src/imports/import_assets.rs
davidbarsky: done!

crates/ide-completion/src/tests/flyimport.rs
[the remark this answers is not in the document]
Veykril: Well the impl is valid sure, but the completion can't possibly trigger since in your snippet `Into` is implemented for `()`, yet you try to call it on `A` which does not implement `Into`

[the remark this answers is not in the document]
Veykril: That is `A.into()` doesn't resolve due to the lack of an impl, `<()>.into()` should resolve though as that impl exists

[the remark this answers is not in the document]
Veykril: Oh, now I see what you mean, right `into` will appears appear given the blanket impls. So any sized type always has at least one applicable `Into` impl

crates/ide-db/src/imports/import_assets.rs
davidbarsky: Wrote my own trait (`pub trait TestInto<T> { fn into(self) -> T; }`, following your suggestion on Zulip) and completions on `()` showed up.

crates/ide-db/src/imports/import_assets.rs
```diff
@@ -536,6 +537,29 @@ fn trait_applicable_items(
...
         Some(assoc_item_trait.into())
     })
+    .filter(|candidate_trait_id| {
+        // we care about the following cases:
+        // 1. Trait's definition crate
+        // 2. Definition crates for all trait's generic arguments
+        //     a. This is recursive for fundamental types: `Into<Box<A>> for ()`` is OK, but `Into<Vec<A>> for ()`` is *not*.
+        // 3. Receiver type definition crate
+        //    a. This is recursive for fundamental types
+        let candidate_trait_id: TraitId = *candidate_trait_id;
```
Veykril, line 547: (style nit) Could be replaced by changing the closure param to `|&candidate_trait_id|`

crates/ide-db/src/imports/import_assets.rs
```diff
@@ -536,6 +537,29 @@ fn trait_applicable_items(
...
+        };
+        let definitions_exist_in_trait_crate = db
+            .trait_impls_in_crate(defining_crate_for_trait.into())
+            .has_impls_for_trait_and_self_ty(candidate_trait_id, receiver_fingerprint);
+
+        let definitions_exist_in_receiver_crate = db
+            .trait_impls_in_crate(trait_candidate.receiver_ty.krate(db).into())
+            .has_impls_for_trait_and_self_ty(candidate_trait_id, receiver_fingerprint);
+
+        definitions_exist_in_trait_crate || definitions_exist_in_receiver_crate
```
Veykril, line 561: I doubt the compiler will optimize this to be lazy due to db accesses, so making the latter a closure to be called in the or expression would a bit better perf wise

crates/hir/src/lib.rs
```diff
@@ -4239,6 +4239,20 @@ impl Type {
...
+                db.struct_data(s).flags.contains(StructFlags::IS_FUNDAMENTAL)
+            }
+            _ => false,
+        }
+    }
+
+    pub fn fingerprint_for_trait_impl(&self) -> Option<TyFingerprint> {
+        TyFingerprint::for_trait_impl(&self.ty)
+    }
+
```
Veykril, line 4255: These aren't needed anymore I think?

crates/ide-db/src/imports/import_assets.rs
davidbarsky: oh, good point. will fix.

crates/hir/src/lib.rs
davidbarsky: `is_fundamental` is not needed, but `fingerprint_for_trait_impl` is. `.has_impls_for_trait_and_self_ty()` is defined in `hir-ty`, but `self` (`hir::Type`) is defined in `hir`, and I think you can imagine why I'm not keen on figuring out how to break that circular dependency 😅

crates/ide-db/src/imports/import_assets.rs
davidbarsky: done!

crates/hir/src/lib.rs
davidbarsky: cool! just removed `is_fundamental` and pushed.

crates/ide-db/src/imports/import_assets.rs
davidbarsky: done!

crates/hir/src/lib.rs
```diff
@@ -40,7 +40,7 @@ use std::{iter, mem::discriminant, ops::ControlFlow};
 use arrayvec::ArrayVec;
 use base_db::{CrateDisplayName, CrateId, CrateOrigin, Edition, FileId};
 use either::Either;
-use hir_def::{
+pub use hir_def::{
```
Veykril, line 43:
```suggestion
use hir_def::{
```

crates/hir/src/lib.rs
```diff
@@ -60,7 +60,7 @@ use hir_def::{
     TypeOrConstParamId, TypeParamId, UnionId,
 };
 use hir_expand::{attrs::collect_attrs, name::name, proc_macro::ProcMacroKind, MacroCallKind};
-use hir_ty::{
+pub use hir_ty::{
```
Veykril, line 63:
```suggestion
use hir_ty::{
```

crates/hir/src/lib.rs
```diff
@@ -110,7 +110,7 @@ pub use {
     cfg::{CfgAtom, CfgExpr, CfgOptions},
     hir_def::{
         attr::{builtin::AttributeTemplate, AttrSourceMap, Attrs, AttrsWithOwner},
-        data::adt::StructKind,
+        data::adt::{StructFlags, StructKind},
```
Veykril, line 113:
```suggestion
        data::adt::StructKind,
```

crates/hir/src/lib.rs
```diff
@@ -4239,6 +4239,10 @@ impl Type {
         }
     }
 
+    pub fn fingerprint_for_trait_impl(&self) -> Option<TyFingerprint> {
```
Veykril, line 4242: I guess for the changes we need to re-export `TyFingerprint` still? If we do that should go in the big re-export use statement

Veykril:
Lgtm with the imports tidied up
@bors delegate+

crates/ide-db/src/imports/import_assets.rs
```diff
@@ -536,6 +537,31 @@ fn trait_applicable_items(
         required_assoc_items.insert(assoc);
         Some(assoc_item_trait.into())
     })
+    .filter(|&candidate_trait_id| {
```
Veykril, line 540: Oh actually, one more thing, I think we can get the same traits multiple times as this point? Especially for an empty fuzzy string search we'll end up with a trait N times (N being its number of assoc items). So maybe we should dedup / collect into a set once first, then filter out by this?

crates/ide-db/src/imports/import_assets.rs
```diff
@@ -536,6 +537,31 @@ fn trait_applicable_items(
...
+        // we care about the following cases:
+        // 1. Trait's definition crate
+        // 2. Definition crates for all trait's generic arguments
+        //     a. This is recursive for fundamental types: `Into<Box<A>> for ()`` is OK, but
+        //        `Into<Vec<A>> for ()`` is *not*.
+        // 3. Receiver type definition crate
+        //    a. This is recursive for fundamental types
```
WaffleLapkin, line 547: How does this implement the recursive parts? It seems from the tests that it works, yet I'm struggling to see where you check the crates. I only see checks for two crates here, not the recursive cases...

crates/hir/src/lib.rs
davidbarsky: weirdly enough, we don't need to reexport `TyFingerprint` for this method. not sure why!

crates/ide-db/src/imports/import_assets.rs
davidbarsky: > How does this implement the recursive parts?

I'm not entirely sure! I found that the tests just worked with the expected behavior? I assume this occurs due to the implementation of `TyFingerprint` inside of `hir_ty`, but to be entirely honest, I don't fully understand that code yet.

Just to confirm, when you say recursive, are you thinking of a case like:

```rust
impl foreign::TestTrait for fundamental::Box<fundamental::Box<local::TestStruct>> {
    fn some_method(&self) {}
}
```

...where the method receiver is a double-Boxed `TestStruct`?

crates/ide-db/src/imports/import_assets.rs
davidbarsky: done!

crates/ide-completion/src/tests/flyimport.rs
```diff
@@ -374,6 +374,135 @@ fn main() {
...
+}
+//- /main.rs crate:main deps:foo
+struct TestStruct;
+
+impl foo::TestTrait for &TestStruct {
+    fn some_method(&self) {}
+}
+
+fn main() {
+    let t = &TestStruct;
```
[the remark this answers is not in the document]
davidbarsky: agreed! weirdly enough, that test fails, despite a rust-analyzer binary working in this case. not sure why, i'm digging, but i don't get an answer shortly, i'm inclined to figure that out later.

crates/ide-db/src/imports/import_assets.rs
WaffleLapkin: Yes, that's what I mean. Although you don't even need double boxes, just a single `Box<TestStruct>` will make you check 3 crates (foreign, fundamental, local)...

crates/ide-completion/src/tests/flyimport.rs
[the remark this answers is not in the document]
WaffleLapkin: Given that the completion currently works, I don't think we should break it in the name of perf.

crates/ide-db/src/imports/import_assets.rs
davidbarsky: I think that's been implemented in the test named `trait_method_fuzzy_completion_aware_of_fundamental_boxes` :)

crates/ide-completion/src/tests/flyimport.rs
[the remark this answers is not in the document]
Veykril: Might be because you need `//- minicore: deref`, not sure but could be that autoderef exits early without the lang item present

crates/ide-db/src/imports/import_assets.rs
WaffleLapkin: I know, thus 
> It seems from the tests that it works

but I can't see why

crates/ide-completion/src/tests/flyimport.rs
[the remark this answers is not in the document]
davidbarsky: `//- minicore: deref` didn't work, unfortunately :/. I'm inclined to treat this variant as *not* a blocker and investigate later. Here's my summary:
- `impl ForeignTrait for LocalStruct` and `impl ForeignTrait for &LocalStruct` provide completions on `let t = &TestStruct` in _both_ tests and a running binary.
- `impl ForeignTrait for &LocalStruct` only provides completions for provide completions on `let t = TestStruct` in a running binary. 
    - This makes me think that there's a bug in `minicore`—or that I might not be understanding some important details—because this behavioral split reproduces with and without my filtering changes.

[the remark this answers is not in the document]
davidbarsky: Made an issue: https://github.com/rust-lang/rust-analyzer/issues/16685

crates/ide-db/src/imports/import_assets.rs
davidbarsky: ah, i misunderstood, sorry! I interpreted " I only see checks for two crates here, not the recursive cases..." as "I don't see tests for the recursive case", not "it's not obvious to me as to why this works".

PR #16559, opened 2024-02-14
Set documentation field in SCIP from doc comment

Wilfred:
Previously, the documentation field was the same as the text shown to users when they hover over that symbol. The documentation should really just be the doc comment, and as of #16179 the signature is already stored in the signatureDocumentation field.

### DIFF
--- a/crates/rust-analyzer/src/cli/scip.rs
+++ b/crates/rust-analyzer/src/cli/scip.rs
@@ -135,12 +135,11 @@ impl flags::Scip {
                     }
 
                     if symbols_emitted.insert(id) {
-                        let documentation = token
-                            .hover
-                            .as_ref()
-                            .map(|hover| hover.markup.as_str())
-                            .filter(|it| !it.is_empty())
-                            .map(|it| vec![it.to_owned()]);
+                        let documentation = match &token.documentation {
+                            Some(doc) => vec![doc.as_str().to_owned()],
+                            None => vec![],
+                        };
+
                         let position_encoding =
                             scip_types::PositionEncoding::UTF8CodeUnitOffsetFromLineStart.into();
                         let signature_documentation =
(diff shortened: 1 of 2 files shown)


crates/ide/src/hover.rs
```diff
@@ -366,6 +367,19 @@ fn hover_ranged(
...
+    scope_node: &SyntaxNode,
+) -> Option<Documentation> {
+    let famous_defs = match &def {
+        Definition::BuiltinType(_) => Some(FamousDefs(sema, sema.scope(scope_node)?.krate())),
+        _ => None,
+    };
+
+    def.docs(sema.db, famous_defs.as_ref())
+}
+
```
Veykril, line 382: This might as well be defined in `static_index`, hover doesn't need this

crates/rust-analyzer/src/cli/scip.rs
```diff
@@ -599,4 +598,22 @@ pub mod example_mod {
...
+        let mut host = AnalysisHost::default();
+        let change_fixture = ChangeFixture::parse(s);
+        host.raw_database_mut().apply_change(change_fixture.change);
+
+        let analysis = host.analysis();
+        let si = StaticIndex::compute(&analysis);
+
+        let file = si.files.first().unwrap();
+        let (_, token_id) = file.tokens.first().unwrap();
+        let token = si.tokens.get(token_id.clone()).unwrap();
```
Veykril, line 615:
```suggestion
        let token = si.tokens.get(token_id.unwrap());
```

Wilfred: That doesn't compile:

```
error[E0599]: no method named `unwrap` found for reference `&ide::TokenId` in the current scope
   --> crates/rust-analyzer/src/cli/scip.rs:615:44
    |
615 |         let token = si.tokens.get(token_id.unwrap());
    |                                            ^^^^^^ method not found in `&TokenId`
```

I've changed the PR to dereference the TokenId instead, which should hopefully also make clippy happy.

lnicola (approved):

PR #16563, opened 2024-02-14
fix: Set `RUSTUP_TOOLCHAIN` and invoke the proxies instead of directly invoking sysroot binaries

Veykril:
cc https://github.com/rust-lang/rust-analyzer/pull/16537 should fix https://github.com/rust-lang/rust-analyzer/issues/16561


darichey:
I think this regresses #15560. If you don't have rustup installed, there are no proxies to invoke (and additionally you don't have `rustc` on your `PATH`). Happy to open a PR if you can advise on the approach.

Veykril:
Right, in that case we don't even need the vars since cargo is no longer involved, https://github.com/rust-lang/rust-analyzer/pull/16693 should fix that I believe.

PR #16567, opened 2024-02-14
fix: False positive unnecessary else block diagnostic

### LINKED ISSUE #16556: `remove-unnecessary-else` bad suggestions and false positives on thread_local macro
**rust-analyzer version**: rust-analyzer 0.3.1839-standalone

**relevant settings**: Unlikely any are relevant

Bad suggestions: Given this code

```rust
fn main() {
    let files = [1];

    let query = if files.is_empty() {
        return;
    } else if files.len() == 1 {
        1
    } else {
        17
    };
    print!("{query}");
}
```
remove-unnecessary-else suggests

```rust
fn main() {
    let files = [1];

    let query = if files.is_empty() {
        return;
    }
    if files.len() == 1 {
        1
    } else {
        17
    };
    print!("{query}");
}
```
which doesn't compile.

I also get false positives on `thread_local!{}` macros, but I don't have a minimal reproduction. In my projects it lints on every `thread_local!{}` call, even those as simple as `thread_local!(static SOMETHING: Cell<u32> = Cell::default());` but I don't get the lints in minimal projects.

ShoyuVanilla:
Fixes #16556

RA detects the else blocks of if-else statements as "unnecessary" if the if-then block's last expression is a never type - e.g. `return` keyword -.
But this shouldn't be done to something like

```rust
let foo = if bar.is_err {
    return;
} else {
    bar.unwrap().0
};
```

### DIFF
--- a/crates/hir-ty/src/diagnostics/expr.rs
+++ b/crates/hir-ty/src/diagnostics/expr.rs
@@ -293,8 +293,23 @@ impl ExprValidator {
             if let Some(else_branch) = else_branch {
                 // If else branch has a tail, it is an "expression" that produces a value,
                 // e.g. `let a = if { ... } else { ... };` and this `else` is not unnecessary
-                if let Expr::Block { tail: Some(_), .. } = body.exprs[*else_branch] {
-                    return;
+                let mut branch = *else_branch;
+                loop {
+                    match body.exprs[branch] {
+                        Expr::Block { tail: Some(_), .. } => return,
+                        Expr::If { then_branch, else_branch, .. } => {
+                            if let Expr::Block { tail: Some(_), .. } = body.exprs[then_branch] {
+                                return;
+                            }
+                            if let Some(else_branch) = else_branch {
+                                // Continue checking for branches like `if { ... } else if { ... } else...`
+                                branch = else_branch;
+                                continue;
+                            }
+                        }
+                        _ => break,
+                    }
+                    break;
                 }
             } else {
                 return;
--- a/crates/hir-ty/src/diagnostics/expr.rs
+++ b/crates/hir-ty/src/diagnostics/expr.rs
@@ -12,6 +12,7 @@ use hir_expand::name;
 use itertools::Itertools;
 use rustc_hash::FxHashSet;
 use rustc_pattern_analysis::usefulness::{compute_match_usefulness, ValidityConstraint};
+use syntax::{ast, AstNode};
 use triomphe::Arc;
 use typed_arena::Arena;
 
@@ -104,7 +105,7 @@ impl ExprValidator {
                     self.check_for_trailing_return(*body_expr, &body);
                 }
                 Expr::If { .. } => {
-                    self.check_for_unnecessary_else(id, expr, &body);
+                    self.check_for_unnecessary_else(id, expr, db);
                 }
                 _ => {}
             }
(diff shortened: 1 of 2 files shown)


ShoyuVanilla:
More details are in my [comment](https://github.com/rust-lang/rust-analyzer/issues/16556#issuecomment-1944271716) of the original issue

davidsemakula:
Thanks for the fix @ShoyuVanilla
I've left suggested changes and a few other considerations as inline comments

ShoyuVanilla:
> Thanks for the fix @ShoyuVanilla I've left suggested changes and a few other considerations as inline comments

Thanks a lot! But where can I see your reviews? BTW I'm in my office now, so it would be few hours later when I reflect them 😄

davidsemakula (changes requested):

davidsemakula:
> > Thanks for the fix @ShoyuVanilla I've left suggested changes and a few other considerations as inline comments
> 
> Thanks a lot! But where can I see your reviews? BTW I'm in my office now, so it would be few hours later when I reflect them 😄

I guess I'd actually forgotten to hit submit 😅 ... you should be able to see them now 🙂

crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -290,7 +290,28 @@ impl ExprValidator {
...
+                                continue;
+                            }
+                        }
+                        _ => break,
+                    }
+                    break;
+                }
+            } else {
                 return;
             }
```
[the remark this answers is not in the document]
ShoyuVanilla: > you should instead be checking whether the if expression has a let statement as a "direct ancestor"

Yes, this would be the right way for handling `let a = if ...` cases and I had considered that method - though quite ugly way iterating exprs once more for let bindings 😅, not as sophisticated as yours(I've learned a lot from your suggestion codes) - but I think that we should also filter out the cases that if-else expression is a tail expression of the other block, like the following example.

```rust
let foo = match bar {
    ...
    _ => {
        if baz {
            return;
        } else {
            a
        }
    }
};
```

I've try to express this in the test code bellow you commented, but yes, as you pointed, that case is not well explaining this and linted as unnecessary.

Thinking about this again, it's tempting to take the method you suggested + extra filtering for tail expression in block. I'll gonna look into this more once I get home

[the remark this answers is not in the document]
davidsemakula: For what it's worth, I still think the example above should still be linted and fixed with below 🙂

```rust
let foo = match bar {
    ...
    _ => {
        if baz {
            return;
        }
        a
    }
};
```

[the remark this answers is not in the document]
ShoyuVanilla: Haha, I was dumb. You're right 👍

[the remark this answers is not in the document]
davidsemakula: Haha, well I'm dumber for writing this diagnostic and not thinking about let statements at all in the first place 😄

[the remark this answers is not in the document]
ShoyuVanilla: Not at all. I think that writing something not exists is the hardest thing 😄 
BTW, after applying your suggestions, I'm gonna think about nicer way to fix the triple if-else case you've mentioned. (I'm still at work 😢 )
Thank you for your kind and smart advices 👍

crates/ide-diagnostics/src/handlers/remove_unnecessary_else.rs
```diff
@@ -41,9 +41,11 @@ fn fixes(ctx: &DiagnosticsContext<'_>, d: &RemoveUnnecessaryElse) -> Option<Vec<
         indent = indent + 1;
     }
     let else_replacement = match if_expr.else_branch()? {
-        ast::ElseBranch::Block(ref block) => {
-            block.statements().map(|stmt| format!("\n{indent}{stmt}")).join("")
-        }
+        ast::ElseBranch::Block(ref block) => block
```
davidsemakula, line 44: nit: this was added by me 😅 ... but you can remove this unnecessary ref 🙂.

crates/ide-diagnostics/src/handlers/remove_unnecessary_else.rs
```diff
@@ -41,9 +41,11 @@ fn fixes(ctx: &DiagnosticsContext<'_>, d: &RemoveUnnecessaryElse) -> Option<Vec<
...
+            .statements()
+            .map(|stmt| format!("\n{indent}{stmt}"))
+            .chain(block.tail_expr().map(|tail| format!("\n{indent}{tail}")))
+            .join(""),
         ast::ElseBranch::IfExpr(ref nested_if_expr) => {
```
davidsemakula, line 49: nit: same here, you can remove this unnecessary ref, again my fault not yours 😅.

davidsemakula (approved):
LGTM! @ShoyuVanilla I left a couple of nits you can fix that were originally added by me 🙂

davidsemakula:
@ShoyuVanilla also would you consider adding the minor changes in https://github.com/rust-lang/rust-analyzer/pull/16575 to this PR? I would then close it in favor of this to reduce the review burden on the maintainers 🙂

ShoyuVanilla:
> @ShoyuVanilla also would you consider adding the minor changes in #16575 to this PR? I would then close it in favor of this to reduce the review burden on the maintainers 🙂

Sure!

crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -288,11 +289,35 @@ impl ExprValidator {
...
+                    // Bail if parent or direct ancestor is a let stmt.
+                    return;
+                }
+                let Some(parent_if_expr) = parent.and_then(ast::IfExpr::cast) else {
+                    // Parent is neither an if expr nor a let stmt.
+                    break;
+                };
+                // Check parent if expr.
+                top_if_expr = parent_if_expr;
+            }
```
Veykril, line 320: This doesn't actually catch all the cases where this is wrong. The transformation is only legal if this if expression is an expression statement.

As `foo(if a { return 1 } else { 0 })` shouldn't trigger this either for example, no `let` statements involved

Veykril: Additionally parsing and such is rather expensive (ideally we could solve this without the AST, but for that we need a tree traversal here) so let's do this check directly before the `self.diagnostics.push(BodyValidationDiagnostic::RemoveUnnecessaryElse { if_expr: id })` line

davidsemakula: @ShoyuVanilla I can also highjack this if that's fine with you ... still relatively early for me 🙂

ShoyuVanilla: Sure you can

davidsemakula:
Changes moved to https://github.com/rust-lang/rust-analyzer/pull/16590

PR #16574, opened 2024-02-15
fix: Fix "needless return" diagnostic for trailing item declarations

### LINKED ISSUE #16566: Incorrect `replace return <expr>; with <expr>` diagnostic
https://github.com/rust-lang/rust-analyzer/blob/1e6cef94dfbc359e7b1e7d392de387c2c539b965/crates/mbe/src/expander/matcher.rs#L184

The transformation here is is not valid so the diagnostic shouldnt' trigger

davidsemakula:
Fixes #16566

### DIFF
--- a/crates/hir-def/src/hir.rs
+++ b/crates/hir-def/src/hir.rs
@@ -352,6 +352,9 @@ pub enum Statement {
         expr: ExprId,
         has_semi: bool,
     },
+    // At the moment, we only use this to figure out if a return expression
+    // is really the last statement of a block. See #16566
+    Item,
 }
 
 impl Expr {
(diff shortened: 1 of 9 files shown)


crates/hir-def/src/hir.rs
```diff
@@ -175,6 +175,7 @@ pub enum Expr {
         statements: Box<[Statement]>,
         tail: Option<ExprId>,
         label: Option<LabelId>,
+        has_trailing_item_decl: bool,
```
Veykril, line 178: Hmm, this feels wrong to me. I think a better approach at fixing this is to not drop item statements on lowering and instead lower them to an empty `Statement::Item` kind for now (we can give it fields once we make actualyl use of that).

davidsemakula: Alright, I'll change to the suggested approach ... I initially went with this because I thought it was the least invasive approach 🙂

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16575, opened 2024-02-15
fix: Fix indenting for child if expr with an else branch for "unnecessary else" diagnostic

davidsemakula:
Indenting for [this if-else](https://github.com/rust-lang/rust-analyzer/pull/16575/files#diff-6a8356f98aa31def8934803aded30183692a3e0d563d72d9395d47aaf83b5a8cR247-R251) expression is one level off without this fix.


davidsemakula:
Closing because changes have been added to https://github.com/rust-lang/rust-analyzer/pull/16567

PR #16577, opened 2024-02-16
internal: Migrate assists to the structured snippet API, part 7/7

DropDemBits:
Continuing from #16467

- `generate_trait_from_impl`

This adds `add_placeholder_snippet_group`, which adds a group of placeholder snippets which are linked together and allows for renaming generated items without going through a separate rename step.

This also removes the last usages of `SourceChangeBuilder::{insert,replace}_snippet`, as all assists have finally been migrated to the structured snippet versions of those methods.

### DIFF
--- a/crates/ide-assists/src/handlers/fix_visibility.rs
+++ b/crates/ide-assists/src/handlers/fix_visibility.rs
@@ -79,7 +79,7 @@ fn add_vis_to_referenced_module_def(acc: &mut Assists, ctx: &AssistContext<'_>) 
         edit.edit_file(target_file);
 
         let vis_owner = edit.make_mut(vis_owner);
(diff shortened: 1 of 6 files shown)


DropDemBits:
@nemethf `generate_trait_from_impl` is the first assist that generates multiple `SnippetTextEdits` 😊

Discovered another snippet rendering issue which causes the placeholders to be placed a little weirdly, but they do still get placed and end up in the final `SnippetTextEdit`s.

DropDemBits:
#16579 fixes the snippet rendering issues I encountered while testing this.

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16588, opened 2024-02-16
internal: Parse (nightly) `const` and `async` trait bounds

compiler-errors:
Both of these bound modifiers were added recently:

* `const` trait bounds: https://github.com/rust-lang/rust/pull/119099
* `async` trait bounds: https://github.com/rust-lang/rust/pull/120392

The latter will certainly will not do the right thing; namely, `async Fn` needs to be mapped to the `AsyncFn` trait. IDK how to do that, so advice would be appreciated, though perhaps we could land this first so the parser isn't complaining about these bounds?

### DIFF
--- a/crates/syntax/rust.ungram
+++ b/crates/syntax/rust.ungram
@@ -614,7 +614,7 @@ TypeBoundList =
 
 TypeBound =
   Lifetime
-| ('?' | '~' 'const')? Type
+| ('~' 'const' | 'const')? 'async'? '?'? Type
 
 //************************//
 //        Patterns        //
--- a/crates/parser/src/grammar/generic_params.rs
+++ b/crates/parser/src/grammar/generic_params.rs
@@ -157,6 +157,16 @@ fn type_bound(p: &mut Parser<'_>) -> bool {
                     p.bump_any();
                     p.expect(T![const]);
                 }
+                // test const_trait_bound
+                // const fn foo(_: impl const Trait) {}
+                T![const] => {
+                    p.bump_any();
+                }
+                // test async_trait_bound
+                // fn async_foo(_: impl async Fn(&i32)) {}
+                T![async] => {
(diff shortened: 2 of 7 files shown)


Veykril (commented):
Can you adjust the grammar file here? https://github.com/rust-lang/rust-analyzer/blob/bb0f93a2608b93da0caab0d51ae71fe6c25f79d0/crates/syntax/rust.ungram#L615-L617
Something like the following should be fine (ungrammar is not an exact representation of the AST)
```text
TypeBound =
  Lifetime
| ('?' | '~')? ('const' | 'async')? Type
```

Then run the `-p syntax` tests to regenerate the AST.

Veykril:
>  IDK how to do that, so advice would be appreciated, though perhaps we could land this first so the parser isn't complaining about these bounds?

Its fine to land the parser changes only, getting rid of the parser errors is the main concern here. The initial lowering of the bounds happen here https://github.com/rust-lang/rust-analyzer/blob/ead369117a061c6a8afd308878cd98d1b8f4b2d3/crates/hir-def/src/hir/type_ref.rs#L345-L380, from there you should be able to follow anything else relevant by usages.

compiler-errors:
👍

crates/syntax/rust.ungram
```diff
@@ -614,7 +614,7 @@ TypeBoundList =
 
 TypeBound =
   Lifetime
-| ('?' | '~' 'const')? Type
+| ('~' 'const' | 'const')? 'async'? '?'? Type
```
compiler-errors, line 617: opted for this because it's a bit more accurate representation of the real grammar rustc parses. I can change it if desired, though.

Veykril: No that's fine as long as the possible tokens are represented.

Veykril:
Thanks!
@bors r+

PR #16590, opened 2024-02-16
fix: Fix false positives for "unnecessary else" diagnostic

### LINKED ISSUE #16556: `remove-unnecessary-else` bad suggestions and false positives on thread_local macro

davidsemakula:
Completes https://github.com/rust-lang/rust-analyzer/pull/16567 by @ShoyuVanilla (see https://github.com/rust-lang/rust-analyzer/pull/16567#discussion_r1492700667)
Fixes #16556

### DIFF
--- a/crates/hir-ty/src/diagnostics/expr.rs
+++ b/crates/hir-ty/src/diagnostics/expr.rs
@@ -109,7 +109,7 @@ impl ExprValidator {
                     self.check_for_trailing_return(*body_expr, &body);
                 }
                 Expr::If { .. } => {
-                    self.check_for_unnecessary_else(id, expr, db);
+                    self.check_for_unnecessary_else(id, expr, &body, db);
                 }
                 Expr::Block { .. } => {
                     self.validate_block(db, expr);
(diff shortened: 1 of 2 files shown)


crates/ide-diagnostics/src/handlers/remove_unnecessary_else.rs
```diff
@@ -387,4 +465,39 @@ fn test() {
...
+"#,
+        );
+        check_diagnostics_with_disabled(
+            r#"
+fn test3() {
+    foo(if a { return 1 } else { 0 })
+}
+"#,
+            std::iter::once("E0308".to_owned()),
+        );
```
davidsemakula, line 501: @Veykril is there a better test for this case?

crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -301,6 +308,36 @@ impl ExprValidator {
...
+                        let ast::Expr::IfExpr(if_expr) = source_ptr.value.to_node(&root) else {
+                            return;
+                        };
+                        let mut top_if_expr = if_expr;
+                        loop {
+                            let parent = top_if_expr.syntax().parent();
+                            let has_parent_expr_stmt_or_stmt_list =
+                                parent.as_ref().map_or(false, |node| {
+                                    ast::ExprStmt::can_cast(node.kind())
+                                        | ast::StmtList::can_cast(node.kind())
```
davidsemakula, line 326: @Veykril it appears if expr that don't end with a semi-colon aren't wrapped in an expr stmt, so I have to check for stmt list parent as well ... is the missing expr stmt a bug?

Veykril: No that is correct, the block tail expression is not considered a statement by our grammar

crates/ide-diagnostics/src/handlers/remove_unnecessary_else.rs
Veykril: That should be fine, but let's add a return type to `test3` so we don't have to supress the type mismatch diagnostic

Veykril:
Thanks!
@bors r+

Veykril:
meh need to rebase again I think
@bors r-

Veykril:
Thanks!
@bors r+

PR #16621, opened 2024-02-21
fix: Recompiles due to RUSTC_BOOTSTRAP

### LINKED ISSUE #15057: rust-analyzer can invoke build scripts with `RUSTC_BOOTSTRAP=1`, wreaking havoc on feature detection
**rust-analyzer version**: 0.3.1541

**rustc version**: rustc 1.70.0 (90c541806 2023-05-31)

**relevant settings**:
- `"rust-analyzer.cargo.buildScripts.useRustcWrapper": false,`

Some crates (for instance anyhow and thiserror) determine whether they are being built by a supported nightly compiler by performing a trial compilation in their build script (think autotools). It seems like rust-analyzer will invoke build scripts while having `RUSTC_BOOTSTRAP` set (perhaps only under some circumstances), which makes those crates think they can use nightly features.

This appears to "work" in rust-analyzer, and "poisons" the build cache with the build script output. A later `cargo run` in the terminal will then reuse the build script output, and build the crate in question *without* having `RUSTC_BOOTSTRAP` set, leading to errors like this:

```
error[E0554]: `#![feature]` may not be used on the stable release channel
   --> /home/sludge/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.71/src/lib.rs:214:32
    |
214 | #![cfg_attr(backtrace, feature(error_generic_member_access, provide_any))]
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0554]: `#![feature]` may not be used on the stable release channel
   --> /home/sludge/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.71/src/lib.rs:214:61
    |
214 | #![cfg_attr(backtrace, feature(error_generic_member_access, provide_any))]
    |                                                             ^^^^^^^^^^^
```

This has previously been reported in https://github.com/dtolnay/thiserror/issues/192 as affecting users that have `rust-analyzer.cargo.buildScripts.useRustcWrapper` *enabled* (the default), while I have it *disabled*, so it seems that the setting may be a red herring and doesn't matter.

txase:
Some packages (e.g. thiserror) force a recompile if the value of the `RUSTC_BOOTSTRAP` env var changes. RA sets the variable to 1 in order to enable rustc / cargo unstable options. This causes flapping recompiles when building outside of RA.

Fixes #15057

### DIFF
--- a/crates/project-model/src/target_data_layout.rs
+++ b/crates/project-model/src/target_data_layout.rs
@@ -32,7 +32,16 @@ pub fn get(
             Sysroot::set_rustup_toolchain_env(&mut cmd, sysroot);
             cmd.envs(extra_env);
             cmd.current_dir(cargo_toml.parent())
-                .args(["rustc", "--", "-Z", "unstable-options", "--print", "target-spec-json"])
+                .args([
+                    "rustc",
+                    "-Z",
+                    "unstable-options",
+                    "--print",
+                    "target-spec-json",
+                    "--",
+                    "-Z",
+                    "unstable-options",
+                ])
                 .env("RUSTC_BOOTSTRAP", "1");
             if let Some(target) = target {
                 cmd.args(["--target", target]);
--- a/crates/project-model/src/build_scripts.rs
+++ b/crates/project-model/src/build_scripts.rs
@@ -138,7 +138,7 @@ impl WorkspaceBuildScripts {
         toolchain: &Option<Version>,
         sysroot: Option<&Sysroot>,
     ) -> io::Result<WorkspaceBuildScripts> {
-        const RUST_1_62: Version = Version::new(1, 62, 0);
+        const RUST_1_75: Version = Version::new(1, 75, 0);
 
         let current_dir = match &config.invocation_location {
             InvocationLocation::Root(root) if config.run_build_script_command.is_some() => {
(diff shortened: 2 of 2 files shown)


crates/project-model/src/target_data_layout.rs
```diff
@@ -32,7 +32,16 @@ pub fn get(
...
+                .args([
+                    "rustc",
+                    "-Z",
+                    "unstable-options",
+                    "--print",
+                    "target-spec-json",
+                    "--",
+                    "-Z",
+                    "unstable-options",
+                ])
```
Veykril, line 44: Is this an accidental change (from debugging perhaps)?

txase: While it looks wonky, it's intended. I explained it in the [commit message](https://github.com/rust-lang/rust-analyzer/pull/16621/commits/2826eb51aa60e2e73d7f35408d2ee2727ec36334) for this change, which I'm pasting here:

> `cargo rustc -- <args>` first builds dependencies then calls `rustc <args>` for the current package. Here, we don't want to build dependencies, we just want to call `rustc --print`. An unstable `cargo rustc` `--print` command bypasses building dependencies first. This speeds up execution of this code path and ensures RA doesn't recompile dependencies with the `RUSTC_BOOTSRAP=1` env var flag set.
> 
> Note that we must pass `-Z unstable-options` twice, first to enable the `cargo` unstable `--print` flag, then later to enable the unstable `rustc` `target-spec-json` print request.

Veykril: Oh I see, I was very confused by the invoked command as it looks somewhat funny :sweat_smile: 

This is a great change thanks, I've been wondering if we could do better as one other downside of the current command is that it doesnt work in virtual workspaces, whereas this does!

Veykril:
Thanks! This is a nice small change! 
@bors r+

PR #16627, opened 2024-02-21
update bug report issue template

Young-Flash:


lnicola:
Maybe we should also add an entry for a repository URL, for larger projects?

lnicola (approved):

Young-Flash:
oops, I don't have enough right. could @lnicola help merge this?

PR #16630, opened 2024-02-21
fix: Wrong closure kind deduction for closures with predicates

### LINKED ISSUE #16421: False positive "cannot mutate immutable variable (E0384)" error

ShoyuVanilla:
Completes #16472, fixes #16421

The changed closure kind deduction is mostly simlar to `rustc_hir_typeck/src/closure.rs`.
Porting closure sig deduction from it seems possible too and I'm considering doing it with another PR


crates/hir-ty/src/infer/unify.rs
```diff
@@ -31,6 +32,72 @@ impl InferenceContext<'_> {
...
+        &mut self,
+        self_ty: InferenceVar,
+    ) -> SmallVec<[WhereClause; 4]> {
+        self.table.resolve_obligations_as_possible();
+
+        let root = self.table.var_unification_table.inference_var_root(self_ty);
+        let pending_obligations = mem::take(&mut self.table.pending_obligations);
+        let obligations = pending_obligations
+            .iter()
+            .filter_map(|obligation| match obligation.value.value.goal.data(Interner) {
```
Veykril, line 46: Could make use of https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain I think?

ShoyuVanilla: Since we are not mutating `obligations` - it should be returned to its original place `pending_obligations` - I think that to use `Vec::retain()`, we should clone the whole `obligations`

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16639, opened 2024-02-22
internal : redesign rust-analyzer::config 

alibektas:
This PR aims to cover the infrastructural requirements for the `rust-analyzer.toml` ( #13529 ) issue. This means, that

1. We no longer have a single config base. The once single `ConfigData` has been divided into 4 : A tree of `.ratoml` files, a set of configs coming from the client ( this is what was called before the `CrateData` except that now values do not default to anything when they are not defined) , a set of configs that will reflect what the contents of a `ratoml` file defined in user's config directory ( e.g `~/.config/rust-analyzer/.rust-analyzer.toml` and finally a tree root that is populated by default values only.
2. Configs have also been divided into 3 different blocks : `global` , `local` , `client`. The current status of a config may change until #13529 got merged.

Once again many thanks to @cormacrelf for doing all the serde work.

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -6,7 +6,6 @@
 //! Of particular interest is the `feature_flags` hash map: while other fields
 //! configure the server itself, feature flags are passed into analysis, and
 //! tweak things like automatic insertion of `()` in completions.
-#![allow(dead_code)]
 use std::{fmt, iter, ops::Not};
 
 use cfg::{CfgAtom, CfgDiff};
@@ -67,7 +66,7 @@ config_data! {
     /// 3. `rust-analyzer.toml` file located at the workspace root
     ///
     /// A config is searched for by traversing a "config tree" in a bottom up fashion. It is chosen by the nearest first principle.
-    global: struct GlobalConfigData <- GlobalConfigInput  -> {
+    global: struct GlobalDefaultConfigData <- GlobalConfigInput -> {
         /// Whether to insert #[must_use] when generating `as_` methods
         /// for enum variants.
         assist_emitMustUse: bool               = false,
@@ -83,7 +82,7 @@ config_data! {
         cargo_allTargets: bool           = true,
         /// Automatically refresh project info via `cargo metadata` on
         /// `Cargo.toml` or `.cargo/config.toml` changes.
-        cargo_autoreload: bool           = true,
+        pub(crate) cargo_autoreload: bool           = true,
         /// Run build scripts (`build.rs`) for more precise code analysis.
         cargo_buildScripts_enable: bool  = true,
         /// Specifies the working directory for running build scripts.
@@ -440,7 +439,7 @@ config_data! {
 config_data! {
     /// Local configurations can be overridden for every crate by placing a `rust-analyzer.toml` on crate root.
     /// A config is searched for by traversing a "config tree" in a bottom up fashion. It is chosen by the nearest first principle.
-    local: struct LocalConfigData <- LocalConfigInput ->  {
+    local: struct LocalDefaultConfigData <- LocalConfigInput ->  {
         /// Toggles the additional completions that automatically add imports when completed.
         /// Note that your client must specify the `additionalTextEdits` LSP client capability to truly have this feature enabled.
         completion_autoimport_enable: bool       = true,
@@ -631,7 +630,7 @@ config_data! {
 config_data! {
     /// Configs that only make sense when they are set by a client. As such they can only be defined
     /// by setting them using client's settings (e.g `settings.json` on VS Code).
-    client: struct ClientConfigData <- ClientConfigInput -> {}
+    client: struct ClientDefaultConfigData <- ClientConfigInput -> {}
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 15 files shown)


alibektas:
I do not think the remaining clippy errors make much sense in this case.

crates/rust-analyzer/src/config.rs
```diff
@@ -618,10 +628,37 @@ pub struct Config {
...
     root_path: AbsPathBuf,
-    data: ConfigData,
     detached_files: Vec<AbsPathBuf>,
     snippets: Vec<Snippet>,
     is_visual_studio_code: bool,
+
+    default_config: ConfigData,
+    client_config: ConfigInput,
+    xdg_config: ConfigInput,
+    ratoml_arena: FxHashMap<SourceRootId, RatomlNode>,
```
Veykril, line 638:
```suggestion
    user_config: ConfigInput,
    ratoml_files: FxHashMap<SourceRootId, RatomlNode>,
```

crates/rust-analyzer/src/config.rs
```diff
@@ -860,15 +900,20 @@ impl Config {
...
+            get_field::<Vec<PathBuf>>(&mut json, &mut errors, "detachedFiles", None)
+                .unwrap_or_default()
                 .into_iter()
                 .map(AbsPathBuf::assert)
                 .collect();
         patch_old_style::patch_json_for_outdated_configs(&mut json);
-        self.data = ConfigData::from_json(json, &mut errors);
-        tracing::debug!("deserialized config data: {:#?}", self.data);
+        let input = ConfigInput::from_json(json, &mut errors);
+        self.client_config = input;
```
Veykril, line 910:
```suggestion
        self.client_config = ConfigInput::from_json(json, &mut errors);
```

crates/rust-analyzer/src/config.rs
```diff
@@ -1894,67 +1956,92 @@ enum SnippetScopeDef {
...
+            slice => slice.serialize(serializer),
+        }
+    }
 }
 
-#[derive(Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(untagged)]
-enum ManifestOrProjectJson {
+pub(crate) enum ManifestOrProjectJson {
```
Veykril, line 2032: Why `pub(crate)`?

crates/rust-analyzer/src/config.rs
```diff
@@ -2088,109 +2175,372 @@ enum ImportPrefixDef {
...
+    global: GlobalConfigData,
+    local: LocalConfigData,
+    client: ClientConfigData,
+}
+
+/// All of the config levels, all fields `Option<T>`, to describe fields that are actually set by
+/// some rust-analyzer.toml file or JSON blob. An empty rust-analyzer.toml corresponds to
+/// all fields being None.
+#[derive(Debug, Clone, Serialize, Default)]
+struct ConfigInput {
```
Veykril, line 2451: It is a bit unfortunate that we store unrelated configs in certain layers (global layer containing client configs for example). Would be nice to split that out somehow

crates/rust-analyzer/src/config.rs
```diff
@@ -2088,109 +2175,372 @@ enum ImportPrefixDef {
...
+            #[allow(unused)]
+            fn clone_with_overrides(&self, input: $input) -> Self {
+                Self {$(
+                    $field: input.$field.unwrap_or_else(|| self.$field.clone()),
+                )*}
+            }
+        }
+
+        impl $input {
+            #[allow(unused)]
```
Veykril, line 2388: >  I do not think the remaining clippy errors make much sense in this case.

That one is due to some invocations leading to an unused (and hence unpushed) vec, just add `#[allow(clippy::ptr_arg)]` to the function

```suggestion
            #[allow(unused, clippy::ptr_arg)]
```

Likewise below

Veykril:
Haven't looked at the split between local global and client yet. Will do that later. Additionally, one thing for a follow up we should look at is whether we can discard the json ptr deserialization approach now that the config structure is fully typed.

crates/rust-analyzer/src/config.rs
alibektas: Two options come to mind:

We can declare non-root nodes as of type `LocalConfigInput`. The problem with this is that once the root node is deleted, one of these non-roots will become a root-node and needs to be upgraded to a `ConfigInput` for which the rest of the configs can either be fetched from vfs but this is not sth we'd want to have as this would mean that `config::Config`needs to have a reference to vfs. The second option is that I somehow compress rest of the data and take the load off of memory and then load it when needed. But I have no idea how this is done.

crates/rust-analyzer/src/config.rs
alibektas: 1. All the fields are now "guarded" behind methods bearing same name as the respective field. 
2. Within `rust-analyzer` crate some field are used as-is, which due to (1) means what was once
looked like `self.config.field_name` now looks like `self.config.field_name()`. There are only a few occurrences of this ( [one example](https://github.com/rust-lang/rust-analyzer/blob/f94ef7378dc6bf415c78eac8b7e8779b165875b4/crates/rust-analyzer/src/reload.rs#L129))
3. `private_interfaces` is on by default. So these methods' return types must be at least `pub(crate)`

alibektas:
I am not satisfied with what I have now, but I also need to move forward with actually implementing ratoml tree to see what I need and what I don't.

crates/rust-analyzer/src/config.rs
alibektas: Or is the question why not `pub`

crates/rust-analyzer/src/config.rs
Veykril: Why would the root node be deletable? There is always one root node (if it doesn't exist as a file it should just be set to the default)

crates/rust-analyzer/src/config.rs
```diff
@@ -6,7 +6,7 @@
 //! Of particular interest is the `feature_flags` hash map: while other fields
 //! configure the server itself, feature flags are passed into analysis, and
 //! tweak things like automatic insertion of `()` in completions.
-
+#![allow(dead_code)]
```
Veykril, line 9: This should be removed ideally

alibektas: there are parts that actually belong to what is to be implemented, so I can't remove this just yet.

alibektas:
For integration tests it is necessary to discuss what will belong to global/local/client. After that, this part can be actually be merged.

PR #16647, opened 2024-02-23
fix: replace_filter_map_next_with_find_map shouldn't work for dyn trait

Young-Flash:
close https://github.com/rust-lang/rust-analyzer/issues/16596

### DIFF
--- a/crates/hir-ty/src/diagnostics/expr.rs
+++ b/crates/hir-ty/src/diagnostics/expr.rs
@@ -160,7 +160,7 @@ impl ExprValidator {
                 });
             }
 
-            let receiver_ty = self.infer[*receiver].strip_references().clone();
+            let receiver_ty = self.infer[*receiver].clone();
             checker.prev_receiver_ty = Some(receiver_ty);
         }
     }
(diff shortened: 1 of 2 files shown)


crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -149,17 +150,18 @@ impl ExprValidator {
...
+                FilterMapNextChecker::new(&self.owner.resolver(db.upcast()), db)
+            });
+
+            if checker.check(call_id, receiver, &callee).is_some() {
                 self.diagnostics.push(BodyValidationDiagnostic::ReplaceFilterMapNextWithFindMap {
                     method_call_expr: call_id,
                 });
             }
+
+            let receiver_ty = self.infer[*receiver].strip_references().clone();
```
Veykril, line 163: Why are we stripping it here? If this is due to coercions and autoref then we should check the adjustments:
```rs
    let coerced = self.infer
            .expr_adjustments
            .get(receiver)
            .and_then(|adjusts| adjusts.last().map(|adjust| adjust.target.clone()));
```

Veykril: Ah nvm you want to get the dyn trait, then we should do the stripping where we check for the dyn trait, that seems more proper.

PR #16651, opened 2024-02-23
Add assist for filling fields by replacing ellipsis in record syntax

dfireBird:
I'm not sure if the tests cover the most cases, I'll add more if suggested.

### DIFF
--- a/crates/ide-assists/src/handlers/fill_record_pattern_fields.rs
+++ b/crates/ide-assists/src/handlers/fill_record_pattern_fields.rs
@@ -188,6 +188,84 @@ fn foo(bar: Bar) {
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


Veykril (commented):
A test case for a struct or enum variant that is generated from a macro would be good (to check whether it works there, or whether it is correctly inapplicable)

dfireBird:
I have never worked on macros (i.e the one writing macros not using them), so if you don't mind, can you point out a snippet somewhere and see how tests can be done?

Veykril:
Yes, that is exactly a test case I had in mind

dfireBird:
I added those tests along with a new check and corresponding test.

Veykril:
Thanks!
@bors r+

PR #14589, opened 2023-04-17
Basic test explorer for VSCode based on testing API

ShuiRuTian:
Another implementation for #3601 

The client side controls life-cycles of tests in this impl.

Another basic idea is to based on `runnables` to reuse existing logic.

> NOTE: the names of many variables and API might be reconsidered.

You might want to read the "Known issue" section in the new added editors/code/src/test_explorer/README.md.
These issues might need to be fixed before the first release, waiting for opinions :)

Although we still lack some setting to control the behavior, the basic logic is done.

Behavior:
Bench, example and build script is not listed in the test explorer.

Features:

Basic VSCode test explorer feature
1. Test item tree and status update:

Note that, run will update the test item on the fly(as long as the output is logged), but debug will update the status when the debug session is over.
image

2. Test run history and error binding

Note that the error is bound to the item rahter than the line which occurs error.
image

Some special feature
3. test item flatten
This is the "normal" test item tree
image
Workspace - package - target - test modules - test
However, this might be kind of verbose, if you have
- only one workspace, then maybe you do not want to click the only workspace to expand. 
- or only one package in the workspace,
- or only one target in package

So we choose to flatten the test items, if there is only one workspace/package/target, we will ignore to build its corresponding test item.
image
In this picture, the target level is flattened.

And, if there is multiple workspaces/packages, but there is only one which has tests, we will not flatten it. but we will not show the ones which does not have tests. So, you might still see the result as if it's not flattened.
image
In this picture, there are two workspaces in fact, but the other does not have any tests.
So as if the workspace is not flattened.

However, this might be kind of confused, because it introduces inconsistent, and VSCode would not remember the last run result if the test item tree is changed.

Please leave your opinions whether the behavior is reasonable.

Issues:
1. Not support select multi items, like:
image
In this screenshot, user chooses two items, however, click run/debug button will get a notification like "Please only pick one to run/debug"

2. Because we load test items lazily now, when running a test suite, whose test cases are not loaded, their status is not updated. And if we load them later, it will be shown as if they are not executed:
image

PR #16662, opened 2024-02-24
Add test explorer

### LINKED ISSUE #3601: Support VS Code testing API
The Java plugin for VS Code comes with a built-in test explorer panel. This panel shows all tests in the project, and allows running the full suite ([link](https://code.visualstudio.com/docs/java/java-testing#_test-explorer)).

Rust Analyzer currently only provides a way to run a single unit test. This opens the console, and runs it there. There is no way to run multiple tests, or browse the tests. Making progress towards this would probably help carry the IDE experience forward.

## Example

report
 
_Java test explorer workflow example ([source](https://code.visualstudio.com/docs/java/java-testing#_test-explorer))_

HKalbasi:
This PR implements the vscode testing api similar to #14589, this time using a set of lsp extensions in order to make it useful for clients other than vscode, and make the vscode client side logic simpler (its now around ~100 line of TS code)

Fix #3601

### DIFF
--- a/crates/rust-analyzer/src/hack_recover_crate_name.rs
+++ b/crates/rust-analyzer/src/hack_recover_crate_name.rs
@@ -1,0 +1,25 @@
+//! Currently cargo does not emit crate name in the `cargo test --format=json`, which needs to be changed. This
+//! module contains a way to recover crate names in a very hacky and wrong way.
+
+// FIXME(hack_recover_crate_name): Remove this module.
+
+use std::sync::{Mutex, MutexGuard, OnceLock};
+
+use ide_db::FxHashMap;
+
+static STORAGE: OnceLock<Mutex<FxHashMap<String, String>>> = OnceLock::new();
+
+fn get_storage() -> MutexGuard<'static, FxHashMap<String, String>> {
+    STORAGE.get_or_init(|| Mutex::new(FxHashMap::default())).lock().unwrap()
+}
+
+pub(crate) fn insert_name(name_with_crate: String) {
+    let Some((_, name_without_crate)) = name_with_crate.split_once("::") else {
+        return;
+    };
+    get_storage().insert(name_without_crate.to_owned(), name_with_crate);
+}
+
+pub(crate) fn lookup_name(name_without_crate: String) -> Option<String> {
+    get_storage().get(&name_without_crate).cloned()
+}
--- a/crates/rust-analyzer/src/main_loop.rs
+++ b/crates/rust-analyzer/src/main_loop.rs
@@ -1,14 +1,15 @@
 //! The main loop of `rust-analyzer` responsible for dispatching LSP
 //! requests/replies and notifications back to the client.
-use crate::lsp::ext;
+
 use std::{
     fmt,
     time::{Duration, Instant},
 };
 
 use always_assert::always;
-use crossbeam_channel::{select, Receiver};
+use crossbeam_channel::{never, select, Receiver};
 use ide_db::base_db::{SourceDatabase, SourceDatabaseExt, VfsPath};
+use itertools::Itertools;
 use lsp_server::{Connection, Notification, Request};
 use lsp_types::notification::Notification as _;
 use stdx::thread::ThreadIntent;
@@ -19,8 +20,9 @@ use crate::{
     diagnostics::fetch_native_diagnostics,
     dispatch::{NotificationDispatcher, RequestDispatcher},
     global_state::{file_id_to_url, url_to_file_id, GlobalState},
+    hack_recover_crate_name,
     lsp::{
-        from_proto,
+        from_proto, to_proto,
         utils::{notification_is, Progress},
     },
     lsp_ext,
@@ -58,6 +60,7 @@ enum Event {
     QueuedTask(QueuedTask),
     Vfs(vfs::loader::Message),
     Flycheck(flycheck::Message),
+    TestResult(flycheck::CargoTestMessage),
 }
 
 impl fmt::Display for Event {
@@ -68,6 +71,7 @@ impl fmt::Display for Event {
             Event::Vfs(_) => write!(f, "Event::Vfs"),
             Event::Flycheck(_) => write!(f, "Event::Flycheck"),
             Event::QueuedTask(_) => write!(f, "Event::QueuedTask"),
+            Event::TestResult(_) => write!(f, "Event::TestResult"),
         }
     }
 }
@@ -81,9 +85,10 @@ pub(crate) enum QueuedTask {
 #[derive(Debug)]
 pub(crate) enum Task {
     Response(lsp_server::Response),
-    ClientNotification(ext::UnindexedProjectParams),
+    ClientNotification(lsp_ext::UnindexedProjectParams),
     Retry(lsp_server::Request),
     Diagnostics(Vec<(FileId, Vec<lsp_types::Diagnostic>)>),
+    DiscoverTest(lsp_ext::DiscoverTestResults),
     PrimeCaches(PrimeCachesProgress),
     FetchWorkspace(ProjectWorkspaceProgress),
     FetchBuildData(BuildDataProgress),
@@ -127,6 +132,7 @@ impl fmt::Debug for Event {
             Event::QueuedTask(it) => fmt::Debug::fmt(it, f),
             Event::Vfs(it) => fmt::Debug::fmt(it, f),
             Event::Flycheck(it) => fmt::Debug::fmt(it, f),
+            Event::TestResult(it) => fmt::Debug::fmt(it, f),
         }
     }
 }
@@ -214,6 +220,10 @@ impl GlobalState {
 
             recv(self.flycheck_receiver) -> task =>
                 Some(Event::Flycheck(task.unwrap())),
+
+            recv(self.test_run_session.as_ref().map(|s| s.receiver()).unwrap_or(&never())) -> task =>
+                Some(Event::TestResult(task.unwrap())),
+
         }
     }
 
@@ -322,6 +332,18 @@ impl GlobalState {
                     self.handle_flycheck_msg(message);
                 }
             }
+            Event::TestResult(message) => {
+                let _p =
+                    tracing::span!(tracing::Level::INFO, "GlobalState::handle_event/test_result")
+                        .entered();
+                self.handle_cargo_test_msg(message);
+                // Coalesce many test result event into a single loop turn
+                while let Some(message) =
+                    self.test_run_session.as_ref().and_then(|r| r.receiver().try_recv().ok())
+                {
+                    self.handle_cargo_test_msg(message);
+                }
+            }
         }
         let event_handling_duration = loop_start.elapsed();
 
@@ -367,7 +389,8 @@ impl GlobalState {
             let update_diagnostics = (!was_quiescent || state_changed || memdocs_added_or_removed)
                 && self.config.publish_diagnostics();
             if update_diagnostics {
-                self.update_diagnostics()
+                self.update_diagnostics();
+                self.update_tests();
             }
         }
 
@@ -488,6 +511,55 @@ impl GlobalState {
         });
     }
 
+    fn update_tests(&mut self) {
+        let db = self.analysis_host.raw_database();
+        let subscriptions = self
+            .mem_docs
+            .iter()
+            .map(|path| self.vfs.read().0.file_id(path).unwrap())
+            .filter(|&file_id| {
+                let source_root = db.file_source_root(file_id);
+                !db.source_root(source_root).is_library
+            })
+            .collect::<Vec<_>>();
+        tracing::trace!("updating tests for {:?}", subscriptions);
+
+        // Updating tests are triggered by the user typing
+        // so we run them on a latency sensitive thread.
+        self.task_pool.handle.spawn(ThreadIntent::LatencySensitive, {
+            let snapshot = self.snapshot();
+            move || {
+                let tests = subscriptions
+                    .into_iter()
+                    .filter_map(|f| snapshot.analysis.crates_for(f).ok())
+                    .flatten()
+                    .unique()
+                    .filter_map(|c| snapshot.analysis.discover_tests_in_crate(c).ok())
+                    .flatten()
+                    .collect::<Vec<_>>();
+                for t in &tests {
+                    hack_recover_crate_name::insert_name(t.id.clone());
+                }
+                let scope = tests
+                    .iter()
+                    .filter_map(|t| Some(t.id.split_once("::")?.0))
+                    .unique()
+                    .map(|it| it.to_owned())
+                    .collect();
+                Task::DiscoverTest(lsp_ext::DiscoverTestResults {
+                    tests: tests
+                        .into_iter()
+                        .map(|t| {
+                            let line_index = t.file.and_then(|f| snapshot.file_line_index(f).ok());
+                            to_proto::test_item(&snapshot, t, line_index.as_ref())
+                        })
+                        .collect(),
+                    scope,
+                })
+            }
+        });
+    }
+
     fn update_status_or_notify(&mut self) {
         let status = self.current_status();
         if self.last_reported_status.as_ref() != Some(&status) {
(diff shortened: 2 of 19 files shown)


HKalbasi:
@ShuiRuTian Can you please review this? You can probably detect its problems since you have implemented this functionality one time :)

ShuiRuTian:
Hi @HKalbasi Thanks for the ping.

I have not reviewed the details of this PR, and I even forgot some details about my PR, so the questions might be stupid.

1. Client side or server side, it's a problem. It's indeed arguable.
    - I did the heavy job on client side based on this comment https://github.com/rust-lang/rust-analyzer/issues/3601#issuecomment-945778285. It's also easy to use the rich feature of the client(VSCode).
    - It's also fine to do the work on server side, [BSP](https://build-server-protocol.github.io/docs/specification#testreport) is a good example.

2. Some flaws: 
    - No debug support. 
    - No cancel support.
    - No refresh support.
    - No enqueued status in front end

4. Is it totally different with Runnables?

5. is test_id a crate? Does this means it always get all tests in a crate and send them to update?

I would try to pick some time to review this PR in detail.

HKalbasi:
@ShuiRuTian thanks for your comment.

> Client side or server side, it's a problem. It's indeed arguable.

My understanding from #16515 was that your PR was closed since it was heavy on the client side, which is not desired. Specially about the https://github.com/rust-lang/rust-analyzer/issues/3601#issuecomment-945778285 the part that this PR differs from it is that it runs the `cargo test` on the server side. The benefit of doing this is that otherwise n clients should each implement a cargo specific logic, but now they can use a regular lsp interface.

> No debug support.

I intentionally deferred it to the future. Running the debug on the server is challenging (but is possible) and running it on the client is against the point of this PR, and test explorer is very useful even without debugging (one can double click on the test in the test explorer, and then use the old debug lens) so I ignored it for now.

> No cancel support.

I added it in my last commit.

> No refresh support.

I don't know exactly what is this. If you hint me about it, I can try to implement it.

> No enqueued status in front end

I don't know much about this one either. When should it be called, and what is its effect in the UI?

> Is it totally different with Runnables?

It doesn't reuse the runnables lsp extension, to keep the client side logic simpler. The runnables lsp extension is not a holy thing to keep it as is, and I think a polished version of this new extension set can replace (most of) the runnables extension in future when it gains support from more clients.

Tangentially, I think in some point we can retire the code lenses for run and debugging tests in favor of the client native UI.

> is test_id a crate? Does this means it always get all tests in a crate and send them to update?

At the protocol level, it can be any module, and client side implementation can handle that, but the server side implementation currently work at the crate level, and return all tests in the crates of the open files and for resolve request. It is not hard to fix it, but it works well enough on the r-a repo so I don't think it is a problem and we can always fix it if it becomes one.

> I would try to pick some time to review this PR in detail.

Thank you!

ShuiRuTian:
Oops, my bad, I forgot to explain some terms.

> No refresh support

This is a "refresh" icon. TestController has a method called "refreshHandler", which would be called. This feature is useful when something goes wrong(according to my experience, mainly because sync issues). We could rebuild everything.

> No enqueued status in front end

This means the tests user chooses to run, but they might not be started yet. Just call them immediately when user click "run" button. TestRun has enqueued method. To know the status of tests, you could run command with `cargo test -Z unstable-options --format=json` (you might also find that --show-output and --report-time are helpful) and analytics the output events.

> Tangentially, I think in some point we can retire the code lenses for run and debugging tests in favor of the client native UI.

Agree, but as you said, it might be not easy to add Debug support. And, according to my experience, again, it might need totally refactor when you need some new features. I reimplemented the whole world once.

---- 

Also, I am kind of curious what it looks like now.

I implemented some feature to try to make it look more naturally:

- some pretty icons:

image

- And if it's not a workspace, we only display unit tests and integration tests, the crate level are totally ignored.

HKalbasi:
I added some icons, basic debug support, and refresh handler.

About enqueued status, my cargo only emits the started event. Do you mean to emit enqueued in that situation? Or there is some other events gated behind a flag (I tried `--show-output` and it didn't add enqueued events).

> And if it's not a workspace, we only display unit tests and integration tests, the crate level are totally ignored.

I omitted this one for the simplicity of the implementation, but it can be easily added in a separate PR.

HKalbasi:
I'm going to merge this, I think this is in the scope of the project, since any other extensions implementing this functionality needs to either ask cargo or r-a to provide it, both are not ideal, and this PR is small enough, specially client side, to be included in the main extension. And this PR is certainly not perfect, but we can incrementally improve it and we shouldn't let perfect becomes the enemy of good.

@bors r+

docs/dev/lsp-extensions.md
```diff
@@ -385,6 +385,106 @@ rust-analyzer supports only one `kind`, `"cargo"`. The `args` for `"cargo"` look
...
+    // True if this test may have children not available eagerly
+    canResolveChildren: boolean;
+    // The id of the parent test in the test tree. If not present, this test
+    // is a top level test.
+    parent?: string | undefined;
+    // The information useful for running the test. The client can use `runTest`
+    // request for simple execution, but for more complex execution forms
+    // like debugging, this field is useful.
+    runnable?: Runnable | undefined;
+};
```
nemethf, line 424: for completeness (and when the client desynchronizes) would it make sense  to include a `state: TestState` field as well?  With possible additional states like: `scheduled` (scheduled to run) and `initial`.

nemethf (commented):
I've only read the documentation, so my "review" is not well informed. Sorry.

ShuiRuTian:
I did not image this PR would be merged so quickly...

I am still curious what does it looks like for now, anyway, I could try next publishment.

Veykril:
> I did not image this PR would be merged so quickly...

Me neither. I am not too happy with all those hacks in this honestly, r-a already has way too much tech debt as is and this is just adding on to it.

crates/rust-analyzer/src/hack_recover_crate_name.rs
```diff
@@ -0,0 +1,25 @@
+//! Currently cargo does not emit crate name in the `cargo test --format=json`, which needs to be changed. This
```
Veykril, line 1: Then we should get that fixed in cargo instead of creating this mess here

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -367,7 +389,8 @@ impl GlobalState {
             let update_diagnostics = (!was_quiescent || state_changed || memdocs_added_or_removed)
                 && self.config.publish_diagnostics();
             if update_diagnostics {
-                self.update_diagnostics()
+                self.update_diagnostics();
+                self.update_tests();
```
Veykril, line 393: This shouldn't run unless the client actually makes use of it.

Veykril (commented):
I haven't looked at the actual logic here yet, unsure when I'll get to that since I'll need to brush up on what was said in https://github.com/rust-lang/rust-analyzer/issues/3601 designwise to compare.

editors/code/src/lsp_ext.ts
```diff
@@ -68,6 +68,37 @@ export const viewItemTree = new lc.RequestType<ViewItemTreeParams, string, void>
...
+};
+export type TestItem = {
+    id: string;
+    label: string;
+    icon: "package" | "module" | "test";
+    canResolveChildren: boolean;
+    parent?: string | undefined;
+    textDocument?: lc.TextDocumentIdentifier | undefined;
+    range?: lc.Range | undefined;
+    runnable?: Runnable | undefined;
```
ShuiRuTian, line 84: Ah, do we have `runnable` at the first commit?

The info in Runnable is rich, I believe it already tells `range`, `textDocument`, and `parent`. It would be kind of confused if there is no single truth of source.

editors/code/src/lsp_ext.ts
```diff
@@ -68,6 +68,37 @@ export const viewItemTree = new lc.RequestType<ViewItemTreeParams, string, void>
...
+};
+export type DiscoverTestResults = { tests: TestItem[]; scope: string[] };
+export type TestState = { tag: "failed"; message: string } | { tag: "passed" } | { tag: "started" };
```
ShuiRuTian, line 87: At lease there is also `cancel` and `ignore` state. [BSP](https://build-server-protocol.github.io/docs/specification/#teststatus) is a good exmple.

editors/code/src/ctx.ts
```diff
@@ -103,6 +105,10 @@ export class Ctx implements RustAnalyzerExtensionApi {
     ) {
         extCtx.subscriptions.push(this);
         this.statusBar = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left);
+        this.testController = vscode.tests.createTestController(
```
ShuiRuTian, line 108: Maybe add an option to control whether the feature is enabled? The default might be `true`(although it's kind of aggressive for me, because the API are all experimental), but an option is a good escape hatch if there is something wrong.

Veykril: Agreed, config (defaulted to false) and a capability is the least we should do here. https://github.com/rust-lang/rust-analyzer/pull/16773

docs/dev/lsp-extensions.md
```diff
@@ -385,6 +385,106 @@ rust-analyzer supports only one `kind`, `"cargo"`. The `args` for `"cargo"` look
...
+    // client should remove old tests not included in the response.
+    scope: string[];
+}
+```
+
+**Method:** `experimental/discoveredTests`
+
+**Notification:** `DiscoverTestResults`
+
+This notification is sent from the server to the client when the
```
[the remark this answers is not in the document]
Veykril: This needs a capability to opt into imo, https://github.com/rust-lang/rust-analyzer/pull/16773

crates/rust-analyzer/src/hack_recover_crate_name.rs
HKalbasi: I will try to fix it in cargo, By the way @ShuiRuTian did you had better solution to this in your PR?

editors/code/src/lsp_ext.ts
HKalbasi: It doesn't tell `parent`, but it tells `range` and `textDocument`. I will remove the duplicate fields of runnable here.

editors/code/src/ctx.ts
HKalbasi: Why should the default be false? Does the presence of this feature interfere with normal operation?

crates/rust-analyzer/src/main_loop.rs
Veykril: Did this in https://github.com/rust-lang/rust-analyzer/pull/16773 fwiw

editors/code/src/ctx.ts
Veykril: This is hot of the press, it might have issues we don't know about and we don't know the perf impact yet. Turning it off by default (and then stating that in the changelog how to enable) seems like the better starting experience, I reckon (at least judging from the issue) that enough people are interested in this and enable to it to give us relevant testing feedback.

Veykril:
@lnicola when you do the changelog for this, would be nice if you could mention that the config introduced by https://github.com/rust-lang/rust-analyzer/pull/16773 has this feature disabled by default, that is it is opt-in

docs/dev/lsp-extensions.md
HKalbasi: Does the client desynchronization happens frequently? The server does not store the test state currently, and it only report it to the client. In the vscode I think it stores the previous results independent of the server, does it solve this problem?

docs/dev/lsp-extensions.md
```diff
@@ -385,6 +385,106 @@ rust-analyzer supports only one `kind`, `"cargo"`. The `args` for `"cargo"` look
...
+This notification is sent from the client to the server when the user is no longer
+interested in the test results. The server should clean up its resources and send
+a `experimental/endRunTest` when is done.
+
+**Method:** `experimental/changeTestState`
+
+**Notification:** `ChangeTestStateParams`
+
+```typescript
+type TestState = { tag: "failed"; message: string }
```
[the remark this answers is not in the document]
HKalbasi: It is the stderr of the test, and it usually has no color unless the program explicitly prints some ansi colored text, which in this case I think it is reported as is.

crates/rust-analyzer/src/hack_recover_crate_name.rs
ShuiRuTian: Not sure what here want to do.
If it's trying to find the crate of running tests, it's not a big problem in my PR. Because we do have the entire logic test tree(creates-testkinds-modules-tests) and we only allows one test to be run meanwhile, so it's easy to get the crate of the test that user clicked by calling get parent.

HKalbasi: Aha. This PR handles running multiple tests so it hits this problem.

It seems there is an effort in upstream (see https://github.com/rust-lang/testing-devex-team/issues/1 and https://github.com/rust-lang/rfcs/pull/3558) about reworking the test json output that will hopefully fix this problem.

docs/dev/lsp-extensions.md
nemethf: Desynchronization is rare, and it is usually a result of a bug.  But these bugs are hard to track down, so any additional info helps to find them.  At any rate, the server does seem to store something about the tests otherwise how it can detect changes in existing tests and send the experimental/discoveredTests notification.  

Maybe TestItem should not contain its state if that's to hard to do, but a `queue-position` would be definitely useful showing how many tests are before it if rust-analyzer scheduled the test to run.

docs/dev/lsp-extensions.md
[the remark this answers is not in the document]
nemethf: I think it would be helpful to extend the documentation explaining that the message is the standard error of the test.

docs/dev/lsp-extensions.md
HKalbasi: I think there is value in keeping the discovery phase and running phase separate. They are separate in the vscode api, and in the r-a case, in the discovery phase the server uses its knowledge but in the running phase it just act as a proxy between cargo and the client. So I think not adding test state and queue position in the discovery related requests is good. But maybe maybe we can add more information to the run related requests for the desynchronization problem. But maybe if it is a rare and bug related event, then the user can just tolerate it rerun the tests?

editors/code/src/lsp_ext.ts
HKalbasi: I ended up keeping those fields, since some of them should be kept in sync with the old runnables in order to function correctly without breaking the old runnables, but are not ideal for the test explorer usage. And it makes the implementation for clients that already implement the runnable extension simpler.

docs/dev/lsp-extensions.md
```diff
@@ -385,6 +385,106 @@ rust-analyzer supports only one `kind`, `"cargo"`. The `args` for `"cargo"` look
...
+interface TestItem {
+    // A unique identifier for the test
+    id: string;
+    // The file containing this test
+    textDocument?: lc.TextDocumentIdentifier | undefined;
+    // The range in the file containing this test
+    range?: lc.Range | undefined;
+    // A human readable name for this test
+    label: string;
+    icon: "package" | "module" | "test";
```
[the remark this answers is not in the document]
HKalbasi: Fixed in #16794

docs/dev/lsp-extensions.md
```diff
@@ -385,6 +385,106 @@ rust-analyzer supports only one `kind`, `"cargo"`. The `args` for `"cargo"` look
...
+This notification is sent from the server to the client when the
+server detect changes in the existing tests. The `DiscoverTestResults` is
+the same as the one in `experimental/discoverTest` response.
+
+**Method:** `experimental/runTest`
+
+**Request:** `RunTestParams`
+
+```typescript
+interface DiscoverTestParams {
```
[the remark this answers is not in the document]
HKalbasi: Fixed in #16794

docs/dev/lsp-extensions.md
nemethf: Probably.  Thanks for this and other clarifications and updates of the documentation.

PR #16670, opened 2024-02-25
Document nvim 0.10 `inlay_hint`

rikhuijzer:
It took me a few hours to figure out how to enable inlay type hints on Nvim 0.10 with `lspconfig`. `rustaceanvim` has already dropped support for them (https://github.com/mrcjkb/rustaceanvim/discussions/46#discussioncomment-7518669) and most other plugins, like `rust-tools`, are depreciated in favor of `rustaceanvim`.

This PR documents how to enable the inlay hints on Neovim 0.10 and later. I've tested this and changing the `on_attach` function is the only step that is required.

### DIFF
--- a/docs/user/manual.adoc
+++ b/docs/user/manual.adoc
@@ -378,6 +378,8 @@ lspconfig.rust_analyzer.setup({
 })
 ----
 
+Note that the hints are only visible after `rust-analyzer` has finished loading **and** you have to edit the file to trigger a re-render.
+
 See https://sharksforarms.dev/posts/neovim-rust/ for more tips on getting started.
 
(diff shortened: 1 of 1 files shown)


Veykril:
Shouldn't such documentation belong to the rustaceanvim docs?

rikhuijzer:
> Shouldn't such documentation belong to the rustaceanvim docs?

Good question. I don't think so. I myself run without any `rust-tools` or `rustaceanvim` and when I compare the GitHub stars of `rustaceanvim` and `lspconfig`, it looks like many people do so too.

This documentation change is in the `lspconfig` section of the docs. Inlay hints require only Neovim, `coc-rust-analyzer`, and `lspconfig`.

rikhuijzer:
Thanks, @Veykril

PR #16677, opened 2024-02-26
fix: filter the DOS device path prefix

### LINKED ISSUE #16666: modding with absolute path disables autocomplete
**rust-analyzer version**: 0.3.1850-standalone (68c506fd6 2024-02-19)
**rustc version**: rustc 1.77.0-nightly (30dfb9e04 2024-01-14)

I'm writing a build script that automatically *mod*s some other files. The build script generates it's output to some directory which then is included with the include! macro. The main file behaves correcty, the autocomplete treats the injected code as if the include macro didn't exist (which is expected). In the included files compile-errors are displayed correctly as well. Just autocomplete is not working. When modding the file manually in main.rs autocomplete starts working. Note that the modded file is in a different directory than src (still within the same vscode session tho)

Hanamiyuzu:
fix #16666

### DIFF
--- a/crates/hir-def/src/nameres/mod_resolution.rs
+++ b/crates/hir-def/src/nameres/mod_resolution.rs
@@ -154,6 +154,14 @@ impl DirPath {
     fn join_attr(&self, mut attr: &str, relative_to_parent: bool) -> String {
         let base = if relative_to_parent { self.parent().unwrap() } else { &self.0 };
 
+        if cfg!(windows) {
+            for dos_prefix in ["\\\\?\\", "\\\\.\\"] {
+                if attr.starts_with(dos_prefix) {
+                    attr = &attr[dos_prefix.len()..];
+                    break;
+                }
+            }
+        }
         if attr.starts_with("./") {
             attr = &attr["./".len()..];
         }


Hanamiyuzu:
gotcha!

lnicola:
> Is there a reason why we replace "\" with "/"?

I do wonder myself, CC @matklad.

crates/hir-def/src/nameres/mod_resolution.rs
```diff
@@ -154,6 +154,14 @@ impl DirPath {
     fn join_attr(&self, mut attr: &str, relative_to_parent: bool) -> String {
         let base = if relative_to_parent { self.parent().unwrap() } else { &self.0 };
 
+        if cfg!(windows) {
```
matklad, line 157: There shouldn’t be any cfgs in the analysis parts of rust-analyzer.

A rust-analyzer server running on a Unix host should successfully analyzer code of a windows-only project. 

Host operating system shouldn’t be an input to analysis.

If there’s some need to vary semantics of paths, it should be triggered by an explicit property of a Crate data structure stored in salsa database.

matklad (commented):
> Is there a reason why we replace "\" with "/"? From my experience on windows a mix of "/" and "\" works fine.
Alternatively canonicalize could be used as a fallback.

Perhaps a better question to ask is why this code doesn’t just use Path? The reason for that is that Path is platform specific: std::os::Path behaves differently on windows and Linux.

But we explicitly don’t want that in rust-analyzer: analysis is a pure function of the safe of the code, running rust-analyzer on a codebase should give *exactly* the same, determinist result irrespective of the host operating system.

So, we need to model our own `Path` in cross-platform way, and we do that by using String and normalizing path delimiters.


I _think_ the code here predates introduction of

https://github.com/rust-lang/rust-analyzer/blob/master/crates/vfs/src/anchored_path.rs

anchored paths is probably the right abstraction we should be using here.

matklad:
Note also the docs at the top of 

https://github.com/rust-lang/rust-analyzer/blob/master/crates/vfs/src/anchored_path.rs

Hanamiyuzu:
I will review the relevant code again, or maybe we should do nothing at all...

PR #16679, opened 2024-02-26
fix: Fix modules in blocks not resolving in ide layer

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/16511


crates/ide/src/syntax_highlighting/test_data/highlight_block_mod_items.html
```diff
@@ -0,0 +1,64 @@
...
+    <span class="parenthesis">(</span><span class="punctuation">$</span>foo<span class="colon">:</span>ident<span class="parenthesis">)</span> <span class="operator">=</span><span class="angle">&gt;</span> <span class="brace">{</span>
+        <span class="keyword">mod</span> y <span class="brace">{</span>
+            <span class="keyword">struct</span> <span class="punctuation">$</span>foo<span class="semicolon">;</span>
+        <span class="brace">}</span>
+    <span class="brace">}</span><span class="semicolon">;</span>
+<span class="brace">}</span>
+<span class="keyword">fn</span> <span class="function declaration">main</span><span class="parenthesis">(</span><span class="parenthesis">)</span> <span class="brace">{</span>
+    <span class="comment">// foo!(Foo);</span>
+    <span class="keyword">mod</span> <span class="module declaration">module</span> <span class="brace">{</span>
+        <span class="unresolved_reference">foo</span><span class="macro_bang">!</span><span class="parenthesis macro">(</span><span class="none macro">Bar</span><span class="parenthesis macro">)</span><span class="semicolon">;</span>
```
Veykril, line 57: There is still a bug here, but I am unsure what the cause is as our early name resolution is quite complicated (we should really clean that up)

PR #16681, opened 2024-02-26
fix: ProjectWorkspace::DetachedFiles does not contain toolchain and target_layout

oooutlk:
Avoid using these fields in `load_cargo::load_crate_graph()`.

### DIFF
--- a/crates/load-cargo/src/lib.rs
+++ b/crates/load-cargo/src/lib.rs
@@ -309,10 +309,6 @@ fn load_crate_graph(
     vfs: &mut vfs::Vfs,
     receiver: &Receiver<vfs::loader::Message>,
 ) -> AnalysisHost {
-    let (ProjectWorkspace::Cargo { toolchain, target_layout, .. }
-    | ProjectWorkspace::Json { toolchain, target_layout, .. }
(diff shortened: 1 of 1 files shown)


oooutlk:
I noticed that using these two fields are not mistakes but an incomplete effort to add them for DetachedFiles. Unfortunately I failed in compiling these code while my project is depending on 0.0.202, that’s why I thought it was a bug. Isn’t it? However I’ll close this pr.

lnicola:
Note that auto-publishing is currently broken, probably because of some breaking change in a dependency of `cargo-workspaces`: https://github.com/rust-lang/rust-analyzer/actions/runs/8044749724/job/21968837388.

CC @pksunkara

pksunkara:
Looks like gix-actor 0.30.1 patch release is broken. @byron Are you willing to yank it and/or release a fix?

Byron:
The version of `gix-actor` you are referring to [is already yanked](https://crates.io/crates/gix-actor/0.30.1). It can't be fixed differently, unfortunately.

lnicola:
Thanks, publishing succeeded in https://github.com/rust-lang/rust-analyzer/actions/runs/8044749724/job/22001956273.

PR #16690, opened 2024-02-27
fix: use 4 spaces for indentation in macro expansion

### LINKED ISSUE #16471: inline assist not working properly on `Hash::hash`
**rust-analyzer version**: 0.4.1810-standalone

**relevant settings**: None

with this code,
```rust
pub struct MyStruct {
    value: u64,
}

impl std::hash::Hash for MyStruct {
    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
        self.value.hash(state);
    }
}

impl MyStruct {
    pub fn as_u64(&self) -> u64 {
        self.value
    }
}
```

place the cursor on `hash` in `self.value.hash(state);` and apply the inline function assist

this is the result:
```rust
...
        {
            let this = &self.value;
            let state: &mutH = state;
          state.write_u64(*self)
        };
...
```

- `&mutH` should be `&mut H`
- `*self` should be `*this`
- the first two lines are indented with 4 spaces (normal), but `state.write_u64(*self)` only with 2, this is automatically fixed with the formatter but still doesnt look nice

the cause could be that the `Hash` impl for `u64` is macro generated, but i dont know exactly

roife:
Partial fix for #16471.
 
In the previous code, the indentation produced by macro expansion was set to 2 spaces. This PR modifies it to 4 spaces for the sake of consistency.

### DIFF
--- a/crates/ide-assists/src/handlers/inline_call.rs
+++ b/crates/ide-assists/src/handlers/inline_call.rs
@@ -1348,8 +1348,8 @@ macro_rules! define_foo {
 define_foo!();
 fn bar() -> u32 {
     {
-      let x = 0;
-      x
+        let x = 0;
+        x
     }
 }
 "#,
@@ -1662,7 +1662,7 @@ fn main() {
     let a: A = A{};
     let b = {
         let a = a;
-      a as A
+        a as A
     };
 }
 "#,
@@ -1781,7 +1781,7 @@ fn _hash2(self_: &u64, state: &mut u64) {
     {
         let inner_self_: &u64 = &self_;
         let state: &mut u64 = state;
-      _write_u64(state, *inner_self_)
+        _write_u64(state, *inner_self_)
     };
 }
 "#,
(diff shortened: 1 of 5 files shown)


crates/ide-assists/src/handlers/inline_call.rs
```diff
@@ -358,16 +358,15 @@ fn inline(
...
-                })
-                .for_each(|usage| {
-                    ted::replace(usage, &this());
-                });
-        }
+
+        body.syntax()
+            .descendants_with_tokens()
+            .filter_map(NodeOrToken::into_token)
+            .filter(|tok| tok.kind() == SyntaxKind::SELF_KW)
```
Veykril, line 365: That approach won't work, as `self` can also refer to the current module

Veykril:
I don't think we can do much about `1.` right now as our search infra is just not built in a way that allows to search like that

crates/ide-assists/src/handlers/inline_call.rs
roife: You are right, I forgot the other semantics of `self` 😱

roife:
I agree that replacement of `self` keyword is non-trivial since it may refer the 'current module'. I have changed the title and the description of this PR.

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16693, opened 2024-02-27
fix: rust-project.json projects not preferring sysroot rustc

Veykril:


Wilfred:
@Veykril thank you for the quick fix! :)

PR #16694, opened 2024-02-27
feat: add onBackspace listener

Young-Flash:
# Explain

This is one of my favorite IntelliJ detailed feature. When the user is on a line with only characters for indentation (spaces or tabs) , pressing the Backspace key deletes the entire line(and if user select a range and press Backspace, it just delete the selected range) and comes to the end of the previous line, user don't need to press Backspace many times to delete all the indentation, which improve development experience a lot.

# Implementation

It's relatively simple, implemented only in client's TS code, I also expose a config("rust-analyzer.enableBackspaceListener") to turn off it in case someone don't like it.

# Demo

onBackspace-20240226105423-tm39yns

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -518,7 +518,7 @@
                 "rust-analyzer.enableBackspaceListener": {
                     "markdownDescription": "Enable the automatically remove whilespace functionality when press backspace key",
                     "type": "boolean",
-                    "default": true
+                    "default": false
                 },
                 "$generated-start": {},
                 "rust-analyzer.assist.emitMustUse": {
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -316,7 +316,7 @@
             {
                 "command": "rust-analyzer.onBackspace",
                 "key": "backspace",
-                "when": "editorTextFocus && editorLangId == rust"
+                "when": "config.rust-analyzer.enableBackspaceListener && editorTextFocus && editorLangId == rust"
             }
         ],
         "configuration": {
@@ -953,6 +953,11 @@
                     "items": {
                         "type": "string"
                     }
+                },
+                "rust-analyzer.enableBackspaceListener": {
+                    "markdownDescription": "Enable the automatically remove whilespace functionality when press backspace key",
+                    "type": "boolean",
+                    "default": true
                 },
                 "rust-analyzer.files.excludeDirs": {
                     "markdownDescription": "These directories will be ignored by rust-analyzer. They are\nrelative to the workspace root, and globs are not supported. You may\nalso need to add the folders to Code's `files.watcherExclude`.",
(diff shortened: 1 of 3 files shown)


editors/code/package.json
```diff
@@ -949,6 +954,11 @@
                         "type": "string"
                     }
                 },
+                "rust-analyzer.enableBackspaceListener": {
```
Veykril, line 957: You'll have to add this before the `                "$generated-start": {},` marker as everything after is auto generated

editors/code/package.json
```diff
@@ -312,6 +312,11 @@
                 "command": "rust-analyzer.joinLines",
                 "key": "ctrl+shift+j",
                 "when": "editorTextFocus && editorLangId == rust"
+            },
+            {
+                "command": "rust-analyzer.onBackspace",
+                "key": "backspace",
+                "when": "config.rust-analyzer.enableBackspaceListener && editorTextFocus && editorLangId == rust"
```
lnicola, line 319: Wow, so we could have made `onEnter` configurable? :exploding_head:.

editors/code/package.json
```diff
@@ -949,6 +954,11 @@
                         "type": "string"
                     }
                 },
+                "rust-analyzer.enableBackspaceListener": {
+                    "markdownDescription": "Enable the automatically remove whilespace functionality when press backspace key",
+                    "type": "boolean",
+                    "default": true
```
Veykril, line 960: We should probably disable this by default. It will most likely be confusing to people to have this behavior in their opened documents in one language but not another.

editors/code/package.json
Young-Flash: maybe 😄

Young-Flash:
CI / TypeScript failed at [Run npm run format:check](https://github.com/rust-lang/rust-analyzer/actions/runs/8063451044/job/22025286789?pr=16694#step:8:19), but it passed in my local, am I missing something?

image

lnicola:
Dunno, works for me:

image

There's also this:

> NOTE: run `cargo test` locally and commit the updated files

editors/code/package.json
Young-Flash: This was supposed to be a feature implemented by the editor, unfortunately not supported by vscode. I think it should be enable by default (I can turn it off if you insist)

flodiebold:
This seems in no way Rust-specific? :confused:

Young-Flash:
> This seems in no way Rust-specific?

yeah it's not rust-specific, much more a editor level feature.

flodiebold:
There's also an open issue in VSCode for this: https://github.com/microsoft/vscode/issues/42785 which links an extension that seems to do what you want: https://marketplace.visualstudio.com/items?itemName=jasonlhy.hungry-delete

Veykril:
I agree with flodiebold here, we shouldn't add non specific rust things to r-a. Our extension isn't meant to be a general purpose extension.

PR #16696, opened 2024-02-27
fix: Goto definition for `deref_mut`

### LINKED ISSUE #16520: Misdirect the definition of `deref_mut` to `deref`
# Problem:
As shown in the [reference of std::ops::DerefMut](https://doc.rust-lang.org/std/ops/trait.DerefMut.html#examples), the definition of the `*` operator in `*x = 'b';` should be `deref_mut`. However, it shows `deref` instead.
image
(This is the screenshot of neovim)

# Versions

**rust-analyzer version**: rust-analyzer 0.3.1831-standalone

**relevant settings**: All of the settings of rust are default. My `coc-rust-analyzer` version is 0.75.1.

ShoyuVanilla:
Fixes #16520

https://github.com/rust-lang/rust-analyzer/blob/a3236be9d7a8179ac4a997858138a4d6c260a451/crates/hir/src/source_analyzer.rs#L375-L393

As we can see from the above, current implementation routes all dereferencing prefix operations to `Deref::deref` implementation, not regarding mutabilities.

https://github.com/rust-lang/rust-analyzer/blob/a3236be9d7a8179ac4a997858138a4d6c260a451/crates/hir-ty/src/infer/mutability.rs#L134-L151

Since we are resolving them already in mutability inferences, we can use those results for proper `deref` / `deref_mut` routing.

### DIFF
--- a/crates/hir/src/source_analyzer.rs
+++ b/crates/hir/src/source_analyzer.rs
@@ -42,7 +42,7 @@ use hir_ty::{
 use itertools::Itertools;
 use smallvec::SmallVec;
 use syntax::{
-    ast::{self, AstNode},
+    ast::{self, AstNode, BinExpr, Expr, IdentPat, PathExpr},
     SyntaxKind, SyntaxNode, TextRange, TextSize,
 };
 use triomphe::Arc;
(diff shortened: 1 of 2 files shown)


Veykril (approved):

Veykril:
Thanks! I wonder if we record that for indexing expressions as well so that we can resolve `IndexMut` in a similar fashion.
@bors r+

ShoyuVanilla:
> I wonder if we record that for indexing expressions as well so that we can resolve IndexMut in a similar fashion.
Good point. I'll check for that case tomorrow

PR #16700, opened 2024-02-27
config: Add Configuration for Progress Reporting Verbosity

davidbarsky:
At work, some of the indexing would take up the entire status bar because of `buck-out` has some *pretty* long paths. This is an option to make it less chatty/loud.

### DIFF
--- a/docs/user/generated_config.adoc
+++ b/docs/user/generated_config.adoc
@@ -793,6 +793,13 @@ This config takes a map of crate names with the exported proc-macro names to ign
 --
 Internal config, path to proc-macro server executable.
 --
+[[rust-analyzer.progressReporting.verbosity]]rust-analyzer.progressReporting.verbosity (default: `[]`)::
++
+--
+Configures the level of detail rust-analyzer will report while scanning files.
+
+If not set, this will default to including parent directories while scanning.
+--
 [[rust-analyzer.references.excludeImports]]rust-analyzer.references.excludeImports (default: `false`)::
 +
 --
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -505,6 +505,11 @@ config_data! {
         /// Internal config, path to proc-macro server executable.
         procMacro_server: Option<PathBuf>          = "null",
 
+        /// Configures the level of detail rust-analyzer will report while scanning files.
+        ///
+        /// If not set, this will default to including parent directories while scanning.
+        progressReporting_verbosity: Vec<ProgressReportingConfig> = "[]",
+
         /// Exclude imports from find-all-references.
         references_excludeImports: bool = "false",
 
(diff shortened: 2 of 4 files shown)


docs/user/generated_config.adoc
```diff
@@ -793,6 +793,13 @@ This config takes a map of crate names with the exported proc-macro names to ign
 --
 Internal config, path to proc-macro server executable.
 --
+[[rust-analyzer.progressReporting.verbosity]]rust-analyzer.progressReporting.verbosity (default: `[]`)::
++
+--
+Configures the level of detail rust-analyzer will report while scanning files.
+
+If not set, this will default to including parent directories while scanning.
```
lnicola, line 801: This feels a bit confusing (I'd expect `[]` to mean that nothing is reported, with `null` being the default).

But I also don't see what makes it behave like the docs say.

Veykril:
This feels like a very odd config to me I have to say 😅 Does VSCode hide elements in the status bar if the messages are too long? (As in does it overflow into other elements?)

davidbarsky:
> This feels like a very odd config to me I have to say 😅 Does VSCode hide elements in the status bar if the messages are too long?

It _is_ a weird config, I agree.

> As in does it overflow into other elements?

Yeah, it overflows and flashes in a weird, somewhat unpleasant way. This config working around that.

docs/user/generated_config.adoc
davidbarsky: Ah crud, I think I got the options here flipped. Lemme fix that.

Veykril:
Ugh, I wonder if we can override the handling in the client code and just filter too long ones / replace the middle or something with `...` to get below a character limit.

Veykril:
So I am still not on board with the idea of a config for this honestly speaking, not sure what to do here.

Though now I'm wondering, are you saying absolute paths in there? Or are your projects just that deep that the paths relative to the project root are very long? If they are absolute than that sounds like we have a mismatch between paths and the project root path for you here.

davidbarsky:
> So I am still not on board with the idea of a config for this honestly speaking, not sure what to do here.
> 
> 
> 
> Though now I'm wondering, are you saying absolute paths in there? Or are your projects just that deep that the paths relative to the project root are very long? If they are absolute than that sounds like we have a mismatch between paths and the project root path for you here.

I completely forgot about this PR!

> Or are your projects just that deep that the paths relative to the project root are very long?

It's this one. The paths easily extend beyond 100 characters and it leads to a lot of flashing in the status bar. It's generally not an issue with something like `~/.cargo/bin/registry`, but proc macros under Buck have some _pretty deeply_ nested paths.

Veykril:
I wonder if we should just do the easy thing and truncate the paths to some arbitrary length (counting from the end of the path)?

Veykril:
I'll close this for now, this does not feel like something we'd want to have configuration for

PR #16709, opened 2024-02-28
fix: Goto definition for `index_mut`

ShoyuVanilla:
Mostly same with #16696.
https://github.com/rust-lang/rust-analyzer/blob/0ac05c05271f31c43d31017cbd288e8737a0edb0/crates/hir-ty/src/infer/mutability.rs#L103-L133
Thankfully, we are doing similar method resolutions so we can use them like the mentioned PR.
As there are only three `LangItem`s having `Mut` in there names; `FnMut`, `DerefMut` and `IndexMut`, I think that this is the last one 😄


Veykril (approved):

Veykril:
Let's see if bors works again
@bors r+

PR #16710, opened 2024-02-28
docs: document a few node types

dev-ardi:


lnicola:
Not to spoil your fun, but that file is generated automatically and shouldn't be edited.

dev-ardi:
Uh... Where do the actual definitions come from?

lnicola:
https://github.com/rust-lang/rust-analyzer/blob/master/crates/syntax/rust.ungram

dev-ardi:
Uh so it's impossible to document them?

lnicola:
`ungram` could parse [the comments](https://github.com/rust-lang/rust-analyzer/blob/master/crates/syntax/rust.ungram#L229-L231) and emit them as rustdoc.

dev-ardi:
How hard would that be to achieve?

PR #16723, opened 2024-02-29
fix: autocomplete constants inside format strings

### LINKED ISSUE #16608: No autocompletion for consts inside format strings.
Code sample:
```rs
fn main() {
    let variable1 = 1;
    let variable2 = 2;
    
    const CONSTANT = 3;
    println!("Some format string {}");
}
```

Current behavior:

https://github.com/rust-lang/rust-analyzer/assets/69449791/5b2944d2-6308-403e-b11c-3f6f2520988e

When applied to format strings, only variables, not constant, will be included in the autocompletion list. Is this by design or is it a bug?

norskeld:
Hi! This PR adds autocompletion for constants (including statics) inside format strings and closes #16608.

I'm not sure about adding the `constants` field to the `CompletionContext`. It kinda makes sense, since it's in line with the `locals` field, and this way everything looks a bit cleaner, but at the same time does it really need to be there?

Anyway, let me know if anything should/can be changed. :)

### DIFF
--- a/crates/ide-completion/src/completions/format_string.rs
+++ b/crates/ide-completion/src/completions/format_string.rs
@@ -1,6 +1,7 @@
 //! Completes identifiers in format string literals.
 
-use ide_db::syntax_helpers::format_string::is_format_string;
+use hir::{ModuleDef, ScopeDef};
+use ide_db::{syntax_helpers::format_string::is_format_string, SymbolKind};
 use itertools::Itertools;
 use syntax::{ast, AstToken, TextRange, TextSize};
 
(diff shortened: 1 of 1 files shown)


lnicola:
This is arguably cosmetic, but I don't think shadowing is handled correctly:

```rust
const a: i32 = 10;

fn main() {
    let a = 10;
    format_args!("{a");
}
```

Here we complete the constant `a`, not the local variable.

Veykril:
I think in that case we'd show both even?

Veykril:
> I'm not sure about adding the `constants` field to the `CompletionContext`. It kinda makes sense, since it's in line with the locals field, and this way everything looks a bit cleaner, but at the same time does it really need to be there?

I think instead of adding that we should switch that completion part over to use `process_all_names` directly. We only collect locals eagerly in the context because of param completions which need to do some special stuff I think?

lnicola:
No, just one, maybe they're de-duped at one point:

image

norskeld:
> Here we complete the constant `a`, not the local variable.

I'm so used to naming constants using uppercase that didn't even think about a case like that! It'd nice to handle this correctly for sure.

I did some digging and I think the culprit here is `hir_def::resolver::Resolver::names_in_scope` which is used in `process_all_names`. It handles shadowing so that if there's a binding and a constant with the same `Name` (like in your example) in the resulting index map there'll be only the constant.

@Veykril @lnicola What should I do here? 

> I think instead of adding that we should switch that completion part over to use `process_all_names` directly. We only collect locals eagerly in the context because of param completions which need to do some special stuff I think?

Yeah, makes sense, I'll change that then.

> I don't even know this icon

And statics use a different one for some reason... 😁

image

lnicola:
> What should I do here?

Ignore it?

> I never know any of those VSCode icons as they are completely unreadable :D

Ah, I take it you're not a Visual C++ user. 😀

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16738, opened 2024-03-02
Update rustc_pattern_analysis

Nadrieril:
This should fix https://github.com/rust-lang/rust-analyzer/issues/16656 but I can't check because we don't have a reproducer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,7 +88,7 @@ ra-ap-rustc_lexer = { version = "0.35.0", default-features = false }
 ra-ap-rustc_parse_format = { version = "0.35.0", default-features = false }
 ra-ap-rustc_index = { version = "0.35.0", default-features = false }
 ra-ap-rustc_abi = { version = "0.35.0", default-features = false }
-ra-ap-rustc_pattern_analysis = { version = "0.37.0", default-features = false }
+ra-ap-rustc_pattern_analysis = { version = "0.40.0", default-features = false }
 
 # local crates that aren't published to crates.io. These should not have versions.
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -88,7 +88,7 @@ ra-ap-rustc_lexer = { version = "0.35.0", default-features = false }
 ra-ap-rustc_parse_format = { version = "0.35.0", default-features = false }
 ra-ap-rustc_index = { version = "0.35.0", default-features = false }
 ra-ap-rustc_abi = { version = "0.35.0", default-features = false }
-ra-ap-rustc_pattern_analysis = { version = "0.37.0", default-features = false }
+ra-ap-rustc_pattern_analysis = { version = "0.40.0", default-features = false }
```
lnicola, line 91:
```suggestion
ra-ap-rustc_pattern_analysis = { version = "0.41.0", default-features = false }
```

I don't think 0.40 actually includes your fix, and 0.41 will hopefully come out in a couple of hours.

Nadrieril: I checked, it does include the fix :) It was published shortly after the fix landed, to my own surprise

lnicola: Ah, sorry then. Still, can you bump it to the new version? It's out now.

lnicola:
Eh, whatever. We can upgrade next time.

@bors r+

Cargo.toml
Nadrieril: Just saw that ^^ Yeah I'll bump next time it's useful

PR #16742, opened 2024-03-03
internal: internal: Implement parent-child relation for `SourceRoot`s

alibektas:
This commit adds the said relation by keeping a map of type `FxHashMap<SourceRootId,Option<SourceRootId>>` inside the `GlobalState`. Its primary use case is reading `rust-analyzer.toml`(#13529)  files that can be placed in every local source root. As a config will be found by traversing this "tree" we need the parent information for every local source root. This commit omits defining this relation for library source roots entirely.


crates/load-cargo/src/lib.rs
```diff
@@ -246,7 +246,7 @@ impl ProjectFolders {
 #[derive(Default, Debug)]
 pub struct SourceRootConfig {
     pub fsc: FileSetConfig,
-    pub local_filesets: Vec<usize>,
+    pub local_filesets: Vec<u64>,
```
alibektas, line 249: This may look ugly but fst uses u64 as its index type so using usize was, I believe, not the right call in the first place. I am aware that such changes are frowned upon but it is either this or I am going to use some usize::try_from().unwrap() in some places :smile:

crates/load-cargo/src/lib.rs
```diff
@@ -265,6 +265,36 @@ impl SourceRootConfig {
...
             .collect()
     }
+
+    /// Maps local source roots to their parent source roots by bytewise comparing of root paths .
+    /// If a source root doesn't have a parent then its parent is declared as None.
+    pub fn source_root_parent_map(&self) -> FxHashMap<SourceRootId, Option<SourceRootId>> {
+        let roots = self.fsc.roots();
+        let mut map = FxHashMap::<SourceRootId, Option<SourceRootId>>::default();
+
+        'outer: for (idx, (root, root_id)) in roots.iter().enumerate() {
```
alibektas, line 275: Looks very ugly to me. I would appreciate suggestions for something more idiomatic.

crates/load-cargo/src/lib.rs
```diff
@@ -413,4 +443,203 @@ mod tests {
...
+        use crate::SourceRootConfig;
+
+        macro_rules! virp {
+            ($s : literal) => {
+                VfsPath::new_virtual_path(format!($s))
+            };
+        }
+
+        #[test]
+        fn test1() {
```
alibektas, line 460: What do we actually think about enumerating tests under a module that groups all tests with similar purpose?

Veykril: Makes it tough to reason about what each test is testing

crates/load-cargo/src/lib.rs
```diff
@@ -265,6 +265,36 @@ impl SourceRootConfig {
             })
             .collect()
     }
+
+    /// Maps local source roots to their parent source roots by bytewise comparing of root paths .
+    /// If a source root doesn't have a parent then its parent is declared as None.
+    pub fn source_root_parent_map(&self) -> FxHashMap<SourceRootId, Option<SourceRootId>> {
```
Veykril, line 271: Is the `None` case useful at all? We can just not insert an entry for the `None` case otherwise

crates/vfs/src/file_set.rs
```diff
@@ -123,6 +123,17 @@ impl FileSetConfig {
...
+    /// Get the lexicographically ordered vector of the underlying map.
+    pub fn roots(&self) -> Vec<(Vec<u8>, u64)> {
+        let mut stream = self.map.stream();
+        let mut vc = vec![];
+        while let Some((pth, idx)) = stream.next() {
+            vc.push((pth.to_vec(), idx));
+        }
+
+        vc
+    }
```
Veykril, line 135: https://docs.rs/fst/latest/fst/map/struct.Stream.html#method.into_byte_vec

crates/load-cargo/src/lib.rs
```diff
@@ -265,6 +265,36 @@ impl SourceRootConfig {
...
+        let roots = self.fsc.roots();
+        let mut map = FxHashMap::<SourceRootId, Option<SourceRootId>>::default();
+
+        'outer: for (idx, (root, root_id)) in roots.iter().enumerate() {
+            if !self.local_filesets.contains(root_id) {
+                continue;
+            }
+
+            for (_, (root2, root2_id)) in roots.iter().enumerate().take(idx).rev() {
+                if root2.iter().enumerate().all(|(i, c)| &root[i] == c) {
```
Veykril, line 281:
```suggestion
                if root.starts_with(&**root2) {
```

crates/load-cargo/src/lib.rs
```diff
@@ -265,6 +265,36 @@ impl SourceRootConfig {
...
+                    // We are interested in parents if they are also local source roots.
+                    // So instead of a non-local parent we may take a local ancestor as a parent to a node.
+                    if self.local_filesets.contains(root2_id) {
```
Veykril, line 284: This check should be done before the `starts_with` check probably

crates/load-cargo/src/lib.rs
davidbarsky: not really, i'd just name each test after what it does and move it under the `test` module above—it's not as if this is gated behind a `cfg`.

crates/load-cargo/src/lib.rs
```diff
@@ -413,4 +443,203 @@ mod tests {
...
+        use ide_db::base_db::SourceRootId;
+        use vfs::{file_set::FileSetConfigBuilder, VfsPath};
+
+        use crate::SourceRootConfig;
+
+        macro_rules! virp {
+            ($s : literal) => {
+                VfsPath::new_virtual_path(format!($s))
+            };
+        }
```
davidbarsky, line 457: i'm not sure that this macro is carrying its weight, IMO. i'd remove it?

crates/load-cargo/src/lib.rs
```diff
@@ -413,4 +443,203 @@ mod tests {
...
+
+        #[test]
+        fn test6() {
+            let mut builder = FileSetConfigBuilder::default();
+            let root = vec![virp!("/ROOT/abc")];
+            let root2 = vec![virp!("/ROOT/def")];
+            let root3 = vec![virp!("/ROOT/def/ghi/jkl")];
+            builder.add_file_set(root);
+            builder.add_file_set(root2);
+            builder.add_file_set(root3);
```
davidbarsky, line 551: i would inline each root into the builder?

crates/load-cargo/src/lib.rs
alibektas: No it is not useful. Sorry!

Veykril (commented):
A bit annoying that `source_root_parent_map` is quadratic over the roots, but I think that shouldn't really be a problem in practice

PR #16748, opened 2024-03-04
internal: Compute syntax validation errors on demand

Veykril:
The LRU cache causes us to re-parse trees quite often, yet we don't use the validation errors at all. With this we push calculating them off to the caller who is interested in them.


lnicola:
image

PR #16749, opened 2024-03-04
internal: Some method resolution cleanups

Veykril:


crates/hir-ty/src/method_resolution.rs
```diff
@@ -1028,80 +1025,69 @@ fn iterate_method_candidates_with_autoref(
...
-            Some(&receiver_ty),
-            Some(receiver_adjustments.clone()),
-            &mut callback,
-        )?
-    }
+    table.run_in_snapshot(|table| {
+        // We're looking for methods with *receiver* type receiver_ty. These could
+        // be found in any of the derefs of receiver_ty, so we have to go through
+        // that, including raw derefs.
+        table.run_in_snapshot(|table| {
```
flodiebold, line 1061: why is this inside the other `run_in_snapshot`? couldn't it just come before that? :thinking:

Veykril: Oh, right this isn't necessary anymore. It was required before due to the `let receiver_ty = table.instantiate_canonical(receiver_ty.clone());` being used in both but that's no longer the case

crates/hir-ty/src/infer/unify.rs
```diff
@@ -438,6 +438,7 @@ impl<'a> InferenceTable<'a> {
     where
         T: HasInterner<Interner = Interner> + TypeFoldable<Interner>,
     {
+        // TODO check this vec here
```
davidbarsky, line 441: what sort of check would happen here?

davidbarsky:
When I tried this out yesterday, this yield a ~10ms improvement on completions!

crates/hir-ty/src/infer/unify.rs
Veykril: Oh that was a note to myself to check if we can cache the allocation for that vec here

Veykril:
Hopefully the saves don't come from https://github.com/rust-lang/rust-analyzer/pull/16749/commits/fcc246e8faadb64f30187ea5191a84531d82c2f0 as that is likely to be invalid

Veykril:
Dropped fcc246e8faadb64f30187ea5191a84531d82c2f0 for now as it fails the test I've added after.

PR #16752, opened 2024-03-04
fix: Destructure struct with no public fields

Lindronics:
Unfortunately I missed this case in #16638.

If a struct only has private members, the assist should not be applicable. Though valid syntax exists (`Foo { .. }`), it isn't particularly useful. Since this case applies to a lot of common types (`Arc`, `Vec`, ...), it probably makes the most sense to hide the action.

This also fixes the `make::record_pat_field_list` function to produce valid syntax if the field list is empty, as it is used in other places too.

## Current behaviour
```rust
// In crate `other_crate`
pub struct Foo { bar: i32 }

// In current crate
fn do_something(foo: other_crate::Foo) {}

// Becomes
fn do_something(other_crate::Foo { , .. }: other_crate::Foo) {}
```

## Fixed behaviour
Assist should not be applicable in this case anymore.


Veykril (approved):
Thanks!

PR #16755, opened 2024-03-05
For toolchain binaries use the full path found in $PATH

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/16754

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1336,7 +1336,6 @@ name = "proc-macro-test"
 version = "0.0.0"
 dependencies = [
  "cargo_metadata",
- "toolchain",
 ]
 
 [[package]]
--- a/crates/ide/src/expand_macro.rs
+++ b/crates/ide/src/expand_macro.rs
@@ -189,7 +189,7 @@ fn _format(
     let &crate_id = db.relevant_crates(file_id).iter().next()?;
     let edition = db.crate_graph()[crate_id].edition;
 
-    let mut cmd = std::process::Command::new(toolchain::rustfmt());
+    let mut cmd = std::process::Command::new(toolchain::Tool::Rustfmt.path());
     cmd.arg("--edition");
     cmd.arg(edition.to_string());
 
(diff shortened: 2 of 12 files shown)


Veykril:
@MabezDev this reverts your recent PR, but if I understood the issue you had correctly, my changes in here should not break your workflow again (we now invoke the full path found in $PATH instead of just the binary name if we find a working path in $PATH).

MabezDev:
Unfortunately, this does break my workflow. I checked out this PR and installed locally, replacing my binary which contained my original PR.

```
2024-03-05T13:29:59.280330Z ERROR rust_analyzer::main_loop: FetchBuildDataError:
error: package `block-device-adapters v0.1.0 (/home/mabez/development/rust/embedded/util/embedded-fatfs/block-device-adapters)` cannot be built because it requires rustc 1.75 or newer, while the currently active rustc version is 1.74.1
```

It's still trying to use my system Rust, and not the rustup installation.

Veykril:
That's odd, you said " My `$PATH` has `$HOME/.cargo/bin` before `/usr/bin`." in your reported issue before, so this should pick that up. Do you have the `RUSTC` env var set for some reason?

Veykril:
Oh wait, you literally have `$HOME/.cargo/bin` in your path I assume? Is that allowed ? If yes then the logic here won't resolve env vars in there so that would explain the issue

MabezDev:
So I added the source to `.zprofile` and it started working :shrug:. This is odd because it's already in my `.zshenv` which I thought was _always_ sourced, but I guess maybe not? Anyways, this change LGTM now.

davidbarsky:
i'll try this with a buck2-built sysroot once [some issues at work](https://www.theverge.com/2024/3/5/24091207/meta-outage-facebook-instagram-threads-down-signed-out) are resolved.

Veykril:
Does this work?
@bors delegate=davidbarsky

Veykril:
it does!

davidbarsky:
Sorry, am a little sick, so I was out of commission yesterday! A buck2-build sysroot is still successfully indexed with these changes. Thanks for waiting to let me check that.

@bors r=@Veykril

PR #16756, opened 2024-03-05
internal: Adjust a few things for trait assoc item hovers

Veykril:
https://github.com/rust-lang/rust-analyzer/pull/15938 (minor turned major wrt diff because of test changes)

### DIFF
--- a/crates/hir/src/display.rs
+++ b/crates/hir/src/display.rs
@@ -596,33 +596,32 @@ impl HirDisplay for Trait {
         write_generic_params(def_id, f)?;
         write_where_clause(def_id, f)?;
 
-        let assoc_items = self.items(f.db);
-        let assoc_items_size = assoc_items.len();
-        let limited_size = f.limited_size.unwrap_or(assoc_items_size);
-        if assoc_items.is_empty() {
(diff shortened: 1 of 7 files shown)


crates/hir/src/display.rs
```diff
@@ -596,33 +596,32 @@ impl HirDisplay for Trait {
...
+                        AssocItem::Const(cst) => cst.hir_fmt(f),
+                        AssocItem::TypeAlias(type_alias) => type_alias.hir_fmt(f),
+                    }?;
+                    f.write_str(";\n")?;
+                }
+
+                if assoc_items.len() > count {
+                    f.write_str("    /* … */\n")?;
+                }
+                f.write_str("}")?;
```
Veykril, line 623: @Young-Flash sorry for changing stuff myself here (but I figured this wasn't worth going through additional review rounds given how long the PR has been up).

Made some small stylistic changes (`;` instead of `,` for seperators, given items), and more importantly if no limit is set we don't show any items or braces at all. As before this we'd still see `{\n    ...\n}` for every trait hover which is rather noisy. If someone wants that they can set the limit to `0` instead of `null`.

PR #16759, opened 2024-03-05
fix: goto-definition for constants inside range pattern

### LINKED ISSUE #15653: No definition found for `const` when referenced from match pattern by range
After `Go to Definition`

изображение

**rust-analyzer version**: 0.3.1665-standalone (05666441b 2023-09-17)
**rustc version**: 1.72.0 (5680fa18f 2023-08-23)

roife:
Fix #15653.

This PR addresses the issue where it was not possible to navigate to constants in range patterns, specifically including two major changes:

1. Previously, both the `start` and `end` fields in `Pat::Range` were of type `LiteralOrConst`. When performing `goto-definition` on constants inside range patterns, r-a would use `resolve_bind_pat_to_const` to find their definitions. However, because the content of a `Const` is not `Pat` but `Path`, it was not stored in the `source_map`, so `resolve_bind_pat_to_const` would returns `None`. This PR changes them to `Const(PatId)`, so that during the lowering process, they are considered as a `pat`, allowing their definitions to be found later through `resolve_bind_pat_to_const`.

2. The process related to range patterns in MIR-lowering has been modified to correctly handle the above changes.

### DIFF
--- a/crates/hir-def/src/body/lower.rs
+++ b/crates/hir-def/src/body/lower.rs
@@ -1413,16 +1413,10 @@ impl ExprCollector<'_> {
                         ast::Pat::LiteralPat(it) => {
                             Some(Box::new(LiteralOrConst::Literal(pat_literal_to_hir(it)?.0)))
                         }
-                        ast::Pat::IdentPat(p) => {
-                            let name =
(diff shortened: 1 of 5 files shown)


Veykril (approved):

Veykril:
Thanks
@bors r+

PR #16762, opened 2024-03-05
Added QuickFix for unresolved field.

wyatt-herkamp:
Adds an AssistKind::QuickFix for unresolved field. 

Currently, Snippets are not available for Diagnostics. So the type defaults to the Unit type if the type is not able to be inferred. 

https://github.com/rust-lang/rust-analyzer/assets/11785959/bd4c67bd-f8cb-4254-9a87-641ae80a7a28

Closes https://github.com/rust-lang/rust-analyzer/issues/4563

### DIFF
--- a/crates/ide-diagnostics/src/handlers/unresolved_field.rs
+++ b/crates/ide-diagnostics/src/handlers/unresolved_field.rs
@@ -1,4 +1,6 @@
-use hir::{db::ExpandDatabase, Adt, HasSource, HirDisplay, InFile};
+use std::iter;
+
+use hir::{db::ExpandDatabase, Adt, HasSource, HirDisplay, InFile, Struct, Union};
 use ide_db::{
     assists::{Assist, AssistId, AssistKind},
     base_db::FileRange,
@@ -7,8 +9,13 @@ use ide_db::{
     source_change::{SourceChange, SourceChangeBuilder},
 };
 use syntax::{
-    ast::{self, edit::IndentLevel, make},
-    AstNode, AstPtr, SyntaxKind,
+    algo,
+    ast::{self, edit::IndentLevel, make, FieldList, Name, Visibility},
+    AstNode, AstPtr, Direction, SyntaxKind, TextSize,
+};
+use syntax::{
+    ast::{edit::AstNodeEdit, Type},
+    SyntaxNode,
 };
 use text_edit::TextEdit;
 
@@ -52,12 +59,12 @@ pub(crate) fn unresolved_field(
 fn fixes(ctx: &DiagnosticsContext<'_>, d: &hir::UnresolvedField) -> Option<Vec<Assist>> {
     let mut fixes = if d.method_with_same_name_exists { method_fix(ctx, &d.expr) } else { None };
     if let Some(fix) = add_field_fix(ctx, d) {
-        fixes.get_or_insert_with(Vec::new).push(fix);
+        fixes.get_or_insert_with(Vec::new).extend(fix);
     }
     fixes
 }
 
-fn add_field_fix(ctx: &DiagnosticsContext<'_>, d: &hir::UnresolvedField) -> Option<Assist> {
+fn add_field_fix(ctx: &DiagnosticsContext<'_>, d: &hir::UnresolvedField) -> Option<Vec<Assist>> {
     // Get the FileRange of the invalid field access
     let root = ctx.sema.db.parse_or_expand(d.expr.file_id);
     let expr = d.expr.value.to_node(&root);
@@ -65,10 +72,6 @@ fn add_field_fix(ctx: &DiagnosticsContext<'_>, d: &hir::UnresolvedField) -> Opti
     let error_range = ctx.sema.original_range_opt(expr.syntax())?;
     // Convert the receiver to an ADT
     let adt = d.receiver.as_adt()?;
-    let Adt::Struct(adt) = adt else {
-        return None;
-    };
-
     let target_module = adt.module(ctx.sema.db);
 
     let suggested_type =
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/unresolved_field.rs
```diff
@@ -46,15 +50,88 @@ pub(crate) fn unresolved_field(
...
 
+fn add_field_fix(ctx: &DiagnosticsContext<'_>, d: &hir::UnresolvedField) -> Option<Assist> {
+    // Get the FileRange of the invalid field access
+    let root = ctx.sema.db.parse_or_expand(d.expr.file_id);
+    let expr = d.expr.value.to_node(&root);
+
+    let error_range = ctx.sema.original_range_opt(expr.syntax())?;
+    // Convert the receiver to an ADT
+    let adt = d.receiver.as_adt()?;
+    let Adt::Struct(adt) = adt else {
```
Veykril, line 68: We should be able to support `union`s as well here

crates/ide-diagnostics/src/handlers/unresolved_field.rs
```diff
@@ -46,15 +50,88 @@ pub(crate) fn unresolved_field(
...
+    }
+    let adt_source = adt.source(ctx.sema.db)?;
+    let adt_syntax = adt_source.syntax();
+    let range = adt_syntax.original_file_range(ctx.sema.db);
+
+    // Get range of final field in the struct
+    let (offset, needs_comma, indent) = match adt.fields(ctx.sema.db).last() {
+        Some(field) => {
+            let last_field = field.source(ctx.sema.db)?.value;
+            let hir::FieldSource::Named(record_field) = last_field else {
```
Veykril, line 94: Maybe leave a FIXME for tuple struct support, no need to implement that now as its less useful (and probably more complicated to handle given its index based, not name based), but would be nice to have that tracked in some way.

crates/ide-diagnostics/src/handlers/unresolved_field.rs
```diff
@@ -46,15 +50,88 @@ pub(crate) fn unresolved_field(
...
+        return None;
+    }
+    let adt_source = adt.source(ctx.sema.db)?;
+    let adt_syntax = adt_source.syntax();
+    let range = adt_syntax.original_file_range(ctx.sema.db);
+
+    // Get range of final field in the struct
+    let (offset, needs_comma, indent) = match adt.fields(ctx.sema.db).last() {
+        Some(field) => {
+            let last_field = field.source(ctx.sema.db)?.value;
```
Veykril, line 93:
```suggestion
            let last_field = adt_syntax.field_list()?.fields().last();
```
somethign like that, no need to go via `source` since we have the main source already

crates/ide-diagnostics/src/handlers/unresolved_field.rs
```diff
@@ -46,15 +50,88 @@ pub(crate) fn unresolved_field(
...
+
+    if !is_editable_crate(target_module.krate(), ctx.sema.db) {
+        return None;
+    }
+    let adt_source = adt.source(ctx.sema.db)?;
+    let adt_syntax = adt_source.syntax();
+    let range = adt_syntax.original_file_range(ctx.sema.db);
+
+    // Get range of final field in the struct
+    let (offset, needs_comma, indent) = match adt.fields(ctx.sema.db).last() {
```
Veykril, line 91: As noted, we can probably just check the `adt_syntax.field_list()` here

crates/ide-diagnostics/src/handlers/unresolved_field.rs
```diff
@@ -46,15 +50,88 @@ pub(crate) fn unresolved_field(
...
+        Some(field) => {
+            let last_field = field.source(ctx.sema.db)?.value;
+            let hir::FieldSource::Named(record_field) = last_field else {
+                return None;
+            };
+            let last_field_syntax = record_field.syntax();
+            let last_field_imdent = IndentLevel::from_node(last_field_syntax);
+            (
+                last_field_syntax.text_range().end(),
+                !last_field_syntax.to_string().ends_with(','),
```
Veykril, line 101: This should probably be something like `algo::skip_trivia_token(field_list.r_curly_token()?, Direction::Prev)` and then check if that token is a comma token. (algo is a module from the syntax crate)

crates/ide-diagnostics/src/handlers/unresolved_field.rs
```diff
@@ -46,15 +50,88 @@ pub(crate) fn unresolved_field(
...
+                last_field_imdent,
+            )
+        }
+        None => {
+            // Empty Struct. Add a field right before the closing brace
+            let indent = IndentLevel::from_node(adt_syntax.value) + 1;
+            let record_field_list =
+                adt_syntax.value.children().find(|v| v.kind() == SyntaxKind::RECORD_FIELD_LIST)?;
```
Veykril, line 109:
```suggestion
                matches!(adt_syntax.value.field_list()?, ast::FieldList::RecordFieldList(_));
```

Veykril (commented):
Some tests would be nice to add (to check if the quickfix is available, the lines with 💡 mark those that have fixes available)

crates/ide-diagnostics/src/handlers/unresolved_field.rs
Veykril: What type is `adt_syntax` here? I assumed it is `ast::Struct`

wyatt-herkamp: So 
```rust
adt_syntax.value // Returns a SyntaxNode<RustLanguage>
adt_syntax.value.field_list() // Is not implemented. U=
```

Now you do have a field_list under adt_source.value
As `adt_source.value` returns a Struct

Veykril:
Thanks!
@bors r+

PR #16768, opened 2024-03-06
mark tokio::test as test

PSeitz:
this is to support the `excludeTests` option to also handle `#[tokio::test]`

### DIFF
--- a/crates/hir-def/src/attr.rs
+++ b/crates/hir-def/src/attr.rs
@@ -202,6 +202,12 @@ impl Attrs {
                 .rev()
                 .zip(["core", "prelude", "v1", "test"].iter().rev())
                 .all(|it| it.0.as_str() == Some(it.1))
+        }) || self.iter().any(|it| {
+            it.path()
+                .segments()
+                .iter()
+                .zip(["tokio", "test"].iter())
+                .all(|it| it.0.as_str() == Some(it.1))
         })
     }
 
--- a/crates/ide/src/references.rs
+++ b/crates/ide/src/references.rs
@@ -319,12 +319,19 @@ fn func() {
 fn test() {
     test_func();
 }
+
+#[tokio::test]
+fn async_test() {
+    test_func();
+}
+
 "#,
             expect![[r#"
                 test_func Function FileId(0) 0..17 3..12
 
                 FileId(0) 35..44
                 FileId(0) 75..84 Test
+                FileId(0) 128..137 Test
             "#]],
         );
 


crates/hir-def/src/attr.rs
```diff
@@ -202,6 +202,12 @@ impl Attrs {
                 .rev()
                 .zip(["core", "prelude", "v1", "test"].iter().rev())
                 .all(|it| it.0.as_str() == Some(it.1))
+        }) || self.iter().any(|it| {
+            it.path()
+                .segments()
+                .iter()
+                .zip(["tokio", "test"].iter())
+                .all(|it| it.0.as_str() == Some(it.1))
```
Veykril, line 210: That's the wrong way to fix this, we need to go up the expansion hierarchy to find the necessary test attribute when categorizing the references

PSeitz: I thought that's too slow as mentioned here: https://github.com/rust-lang/rust-analyzer/pull/16441#discussion_r1471087307

Although I would prefer to have a full filtered result which takes a little bit longer

crates/ide/src/references.rs
```diff
@@ -319,12 +319,19 @@ fn func() {
 fn test() {
     test_func();
 }
+
+#[tokio::test]
```
Young-Flash, line 323: `#[tokio::test]` is already mark as test, cause it will be expand to `#[::core::prelude::v1::test]`

crates/hir-def/src/attr.rs
Veykril: Right, that was about general traversal which we do, we are missing traversing up the macro expansions still though. Seems that was overlooked in the PR. That won't make much of a difference I guess given most finds shouldn't be in macro calls anyways.

crates/ide/src/references.rs
PSeitz: I thought it's not working because I saw a lot of references in `tokio::test`, but the problem with detection seems more general. 

e.g. a `vec![]` is enough to break the current filter

```rust
struct Point {
    x: f64,
    y: f64,
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn it_works() {
        assert_eq!(2 + 2, 4);
        let vec = vec![Point { x: 1.0, y: 2.0 }];
        assert_eq!(vec.len(), 1);
    }
}

```

crates/hir-def/src/attr.rs
PSeitz: I thought it's related to `tokio::test`, but it's not:
https://github.com/rust-lang/rust-analyzer/pull/16768#discussion_r1514463398

PSeitz:
@Veykril @Young-Flash 
I'm not familiar enough with rust-analyzer.

`sema.to_def(&it)` what does this do? The `it_works` function should appear with the `func.is_test(sema.db)` on the example below, so my theory would be that this look up fails, or the ancestors traversal does not work like expected.


```rust
fn is_name_ref_in_test(sema: &Semantics<'_, RootDatabase>, name_ref: &ast::NameRef) -> bool {
    name_ref.syntax().ancestors().any(|node| match ast::Fn::cast(node) {
        Some(it) => sema.to_def(&it).map_or(false, |func| func.is_test(sema.db)),
        None => false,
    })
}
```


```rust
struct Point {
    x: f64,
    y: f64,
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn it_works() {
        assert_eq!(2 + 2, 4);
        let vec = vec![Point { x: 1.0, y: 2.0 }];
        assert_eq!(vec.len(), 1);
    }
}
```

PR #16771, opened 2024-03-06
internal: Don't desugar self param into a pattern

Veykril:
Small experiment to see if this simplifies things


Veykril:
This does cause some extra code duplication but it reflects better when it comes to source mapping back, as now strictly pattern syntax is tied to lowered patterns, instead of a nasty either. Imo it feels nicer this way even if it costs us a bit of extra code.

PR #16772, opened 2024-03-06
internal: Add tracing spans to macro generated database

Veykril:


crates/salsa/salsa-macros/src/query_group.rs
```diff
@@ -235,13 +235,24 @@ pub(crate) fn query_group(args: TokenStream, input: TokenStream) -> TokenStream
 
         queries_with_storage.push(fn_name);
 
+        let tracing = if let QueryStorage::Memoized = query.storage {
+            let s = format!("{trait_name}::{fn_name}");
```
davidbarsky, line 239: it's been a second since I last looked at writing proc macros, but I think that [`quote::format_ident`](https://docs.rs/quote/latest/quote/macro.format_ident.html) is preferable to allocating here. if that doesn't work, i'd consider making the trait the target here.

Veykril: This intentional is a string, because we want to expand it to a string literal, not an ident, Also this allocation happens at expansion time so it doesn't really matter as it doesn't happen when running the query

PR #16776, opened 2024-03-07
fix: Don't force draw a dependency edge to the real_span_map query

Veykril:
This can cause extra invalidations as like direct `AstIdMap` dependencies


lnicola:
This had quite an impact on the metrics:

image

PR #16778, opened 2024-03-07
Update `rustc_pattern_analysis` to 0.42.0

Nadrieril:
There was an important API change in 0.41.0, and (hopefully) a fix for https://github.com/rust-lang/rust-analyzer/issues/16774 in 0.42.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,10 +84,10 @@ tt = { path = "./crates/tt", version = "0.0.0" }
 vfs-notify = { path = "./crates/vfs-notify", version = "0.0.0" }
 vfs = { path = "./crates/vfs", version = "0.0.0" }
 
-ra-ap-rustc_lexer = { version = "0.40.0", default-features = false }
-ra-ap-rustc_parse_format = { version = "0.40.0", default-features = false }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -88,7 +88,7 @@ ra-ap-rustc_lexer = { version = "0.40.0", default-features = false }
 ra-ap-rustc_parse_format = { version = "0.40.0", default-features = false }
 ra-ap-rustc_index = { version = "0.40.0", default-features = false }
 ra-ap-rustc_abi = { version = "0.40.0", default-features = false }
-ra-ap-rustc_pattern_analysis = { version = "0.40.0", default-features = false }
+ra-ap-rustc_pattern_analysis = { version = "0.42.0", default-features = false }
```
Veykril, line 91: Can we bump the other ra-ap-rustc crates here as well? Otherwise we get duped dependencies of those and usually they don't contain any abi breakage

Nadrieril: Done, there wa sonly a tiny API change

lnicola (approved):

PR #16787, opened 2024-03-08
Remove unncessary check for macro call

pksunkara:
Since `macro_rules` is a contextual keyword, it is an `IDENT` token and thus `is_path_start` already identifies it correctly. You can tell the previous check is unnecessary because the relevant tests still pass.


lnicola:
Yeah, I'm not sure what this is trying to do, but it looks like a no-op today.

@bors r+

Issue #13904, opened 2023-01-07
Structs that implement Parser from clap::Parser throw an error no-such-field


lever1209:
**rust-analyzer version**: 0.4.1352-standalone

**rustc version**: rustc 1.66.0 (69f9c33d7 2022-12-12)

**relevant settings**:
```json
{
	"rust-analyzer.files.excludeDirs": [
		"target"
	],
	"rust-analyzer.cargo.features": ["interactive-cli", "debug"]
}
```
```toml
[dependencies]
clap = { version = "4.0.32", features = ["derive"] }
```

```rs
#[derive(Parser, Debug)]
#[command(author, version, about, long_about = None)]
struct Args {
	#[arg(action = ArgAction::Set, id = "config_path", help = "path to an alternate config file.", short, long)]
	config: String,

	#[cfg(feature = "linux-qt")]
	#[arg(action = ArgAction::SetTrue, long = "qt")]
	use_qt: bool,

	#[cfg(feature = "linux-gtk")]
	#[arg(action = ArgAction::SetTrue, long = "gtk")]
	use_gtk: bool,

	#[cfg(any(
		feature = "linux-gtk",
		feature = "linux-qt",
		feature = "windows-uwp",
		feature = "windows-native"
	))]
	#[arg(action = ArgAction::SetTrue, long = "nogui")]
	nogui: bool,
}
```

lnicola:
You haven't said which field is missing. Does `config` work?

lever1209:
use_qt, use_gtk, and nogui, anything that is currently disabled because of features

lnicola:
You'll have to enable those features for the code to be visible.

lever1209:
im fully aware, just replicate it and see what i mean

lowr:
Given

```rust
#[derive(clap::Parser)]
struct Args {
    #[cfg(feature = "linux-qt")]
	use_qt: bool,
}
```

we expand the derive macro to

```rust
#[deny(clippy::correctness)]
impl clap::FromArgMatches for Args {
    // ...
    fn from_arg_matches_mut(
        __clap_arg_matches: &mut clap::ArgMatches,
    ) -> ::std::result::Result<Self, clap::Error> {
        #![allow(deprecated)]
        let v = Args {
            use_qt: /* ... */
        };
        ::std::result::Result::Ok(v)
    }
    // ...
}
```

[...]

Issue #11657, opened 2022-03-08
no-such-field for `#[cfg(not(test))]` with `StructOpt`


kolloch:
```rust
use structopt::StructOpt;

#[derive(StructOpt)]
struct Opt {
    #[cfg(not(test))]
    arg: String, // <- error for the whole line
}

fn main() {
    #[cfg(not(test))]
    let opt = Opt::from_args();
    #[cfg(not(test))]
    println!("arg {}", opt.arg);
}
```
=>
```
no such fieldrust-analyzer[no-such-field](https://rust-analyzer.github.io/manual.html#no-such-field)
```

`cargo test` `cargo build` `cargo build --release` don't show any errors.

The error occurs only in combination with `#[derive(StructOpt)]`.

Full source:
https://github.com/kolloch/no-such-field

## itemtree.rs

Strips `#[cfg(not(test))]` items from main but not from struct:

```rust
pub(self) use structopt::StructOpt;

#[derive(StructOpt)]  // AttrId { ast_index: 0 }
pub(self) struct Opt {
    #[cfg(not (test))]  // AttrId { ast_index: 0 }
    pub(self) arg: String,
}

// flags = 0x2
pub(self) fn main() -> ();
```

## cargo expand

```
❯ cargo expand --tests
   Compiling no-such-field v0.1.0 (/home/peter/nexxiot/no-such-field)
    Finished test [unoptimized + debuginfo] target(s) in 0.09s
```
```rust
#![feature(prelude_import)]
#[prelude_import]
use std::prelude::rust_2021::*;
#[macro_use]
extern crate std;
use structopt::StructOpt;
struct Opt {}
#[allow(unused_variables)]
#[allow(unknown_lints)]
#[allow(
    clippy::style,
    clippy::complexity,
    clippy::pedantic,
    clippy::restriction,
    clippy::perf,
    clippy::deprecated,
    clippy::nursery,
    clippy::cargo
)]
#[deny(clippy::correctness)]
#[allow(dead_code, unreachable_code)]
impl ::structopt::StructOpt for Opt {
    fn clap<'a, 'b>() -> ::structopt::clap::App<'a, 'b> {
        let app = ::structopt::clap::App::new("no-such-field");
        <Self as ::structopt::StructOptInternal>::augment_clap(app)
    }
    fn from_clap(matches: &::structopt::clap::ArgMatches) -> Self {
        Opt {}
    }
}
#[allow(unused_variables)]
#[allow(unknown_lints)]
[...]

PR #16789, opened 2024-03-08
feat: Configure derive macro inputs

### LINKED ISSUE #8434: Custom derive attribute stripping is insufficient
In

```rust
#[derive(Derive)]
#[cfg_attr(never, derive(Clone))]
struct S {
    #[cfg(never)]
    invisible_field: u8,
    #[cfg(not(never))]
    visible_field: u8,
}
```

The `Derive` macro should be passed the token stream corresponding to

```rust
struct S {
    #[cfg(not(never))]
    visible_field: u8,
}
```

Currently, we pass

```rust
#[cfg_attr(never, derive(Clone))]
struct S {
    #[cfg(never)]
    invisible_field: u8,
    #[cfg(not(never))]
    visible_field: u8,
}
```

Since we have to evalutate `cfg`s to do this correctly, I propose moving the `remove_derive_attrs` code to `hir_expand`.

wyatt-herkamp:
This will attempt to process cfg attributes and cfg_attr attributes for proc macro expansion. 
image

Closes #8434 , #11657, and #13904


wyatt-herkamp:
I do have a few questions for the team that works on Rust Analyzer. 
1. Do you want me to throw this behind a feature flag as an unstable feature while it is tested?
2. The changes to [fixup.rs](https://github.com/wyatt-herkamp/rust-analyzer/blob/cfg_attr/crates/hir-expand/src/fixup.rs) I changed the type for remove from SyntaxNode to SyntaxElement because I need to be able to remove the comma as well. Is this the best choice. Or can I remove the comma another way?
3. Are there edge cases when I need to pass the `cfg` and `cfg_attr` attribute to the actual proc macro?
4. Can [this match](https://github.com/wyatt-herkamp/rust-analyzer/blob/f45b080965036d5386087c61b66fd93dff406cdc/crates/hir-expand/src/cfg_process.rs#L149-L175) statement be simplified down?

wyatt-herkamp:
image
`#[cfg_attr(..,..)]` Is parsing

crates/cfg/src/cfg_expr.rs
```diff
@@ -62,6 +72,69 @@ impl CfgExpr {
         }
     }
 }
+fn next_cfg_expr_from_syntax<I>(iter: &mut Peekable<I>) -> Option<CfgExpr>
```
Veykril, line 75: Let's move this stuff into `cfg_process`, it doesn't sit right with me that the `cfg` crate depends on `syntax`.

crates/hir-expand/src/cfg_process.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+    if !found_comma {
+        warn!("No comma found in {}", meta_tt);
+        return None;
+    }
+    // Remove the right paren
+    remove.insert(meta_tt.r_paren_token()?.into());
+    Some(remove)
+}
+fn add_comma(item: &impl AstNode, res: &mut FxHashSet<SyntaxElement>) {
```
Veykril, line 105: The name is a bit confusing, since technically this causes us to remove the comma from the input :)

crates/hir-expand/src/cfg_process.rs
```diff
@@ -0,0 +1,204 @@
...
+                    remove.insert(attr.syntax().clone().into());
+                    continue;
+                }
+            }
+        }
+    }
+    Some(())
+}
+
+fn remove_tokens_within_cfg_attr(meta: Meta) -> Option<FxHashSet<SyntaxElement>> {
```
Veykril, line 67: We should probably remove all tokens after the initial subtree that we wanna keep post expansion (that is `#[cfg_attr(foo, bar, baz)]` should be censored into `#[bar]`, instead of `#[bar, baz]` while still incorrect, that probably prevents more issues from invalid attributes being passed to proc-macros.)

Veykril (commented):
Annoying we have to do this on the syntax tree but there is no way around that. note that
```rs
struct S {
    field: fn(#[cfg(FALSE)] ()) -> ()
}
```
here we also need to configure the param away. It's fine to not support this right now though, I think we should probably have this infra here traverse the token tree in an untyped manner and just look for attributes and then handle things accordingly? Unsure.

Anyways with the few things I've raised I'm fine with merging this for now, since you said you were planning on tackling full `cfg_attr` supports, cleaning up and what not can be done afterwards (since the changes are all tied closely to a single module).

crates/hir-expand/src/cfg_process.rs
wyatt-herkamp: So I was going to handle that case next. https://github.com/rust-lang/rust-analyzer/issues/10110

The reason I did not was I am needing to append tokens not just remove them. Because to do that properly I would need to create a second seperate attribute. I know this is a bug. However, currently `syntax_node_to_token_tree_modified`
The `append` input is a `FxHashMap<SyntaxElement, Vec<tt::Leaf<SpanData<Ctx>>>>` and I am not sure however to convert a full attribute into that. So I was going to let that be a bug on its own. When I tested this case. 

What we see is a unexpected comma. I would be happy to fix this in this PR. However, I would need some guidance on appending a new attribute onto the code.

Veykril: I was mainly saying to do damage mitigation until we fully implement this, as syn proc-macros will just error out if we give them a malformed attribute  over one that is just missing information. We need to adjust `append`'s type here, since attributes consist of token trees and we are unable to model those with just leaves. We used to have `Subtree`s in there before but I removed that due to the unnecessary delimiters created. I think we want to change that type into `FxHashMap<SyntaxElement, Vec<Either<tt::Leaf<SpanData<Ctx>>, tt::Subtree<SpanData<Ctx>>>>>` or similar. A bit ugly but that is the most permiissible that should support he use case needed here (can re-evaluate after we see whats needed for the implementation). Then we can represent an attribute with a `#` leaf punc node and a `[...]` subtree

Veykril:
Forgot to answer your questions:

> Do you want me to throw this behind a feature flag as an unstable feature while it is tested?

If this is easily doable (which it looks like it should be), then yes let's do that for now so we can test it a bit.

> The changes to [fixup.rs](https://github.com/wyatt-herkamp/rust-analyzer/blob/cfg_attr/crates/hir-expand/src/fixup.rs) I changed the type for remove from SyntaxNode to SyntaxElement because I need to be able to remove the comma as well. Is this the best choice. Or can I remove the comma another way?

Seems right to do this

> Are there edge cases when I need to pass the cfg and cfg_attr attribute to the actual proc macro?

I don't think so

> Can [this match](https://github.com/wyatt-herkamp/rust-analyzer/blob/f45b080965036d5386087c61b66fd93dff406cdc/crates/hir-expand/src/cfg_process.rs#L149-L175) statement be simplified down?

Not really

Veykril:
> How do I get the settings from the expand db?

We'd have to thread that through the database which is kind of annoying actually. Should be fine to not put this behind a config then. It shouldn't really break anything anyways (or rather make it worse than the status quo).

Also would be nice to add a test that this works as expected for our builtin derives in `crates\hir-def\src\macro_expansion_tests\builtin_derive_macro.rs`

Veykril:
> Are there edge cases when I need to pass the cfg and cfg_attr attribute to the actual proc macro?

Actually there is another one, `#[cfg_attr(foo, cfg_attr(bar, baz))]`, that is we need to recurse on this

wyatt-herkamp:
> > Are there edge cases when I need to pass the cfg and cfg_attr attribute to the actual proc macro?
> 
> Actually there is another one, `#[cfg_attr(foo, cfg_attr(bar, baz))]`, that is we need to recurse on this

Do you want that here? Or should that be done when we handle multiple attributes?

Veykril:
that is fine to skip for now, its a very rare occurence

wyatt-herkamp:
> that is fine to skip for now, its a very rare occurence
Awesome. I think it should be good now. 

Comparing against output from cargo expand is good. 
Code_TtasZpkNeL

Veykril (approved):
Alright, lgtm 👍 (a few things I'd clean up slightly, but I think that can wait until your follow up and I might just do that myself, don't wanna nitpick and block a PR that way)

Veykril:
Very nice to see this, this has been a pain point for a lot of folks, especially diesel I think cc @weiznich :)
Thanks!
@bors r+

PR #16795, opened 2024-03-09
remove unused struct Snap in lsif

mu001999:
Detected by #121752, see https://github.com/rust-lang/rust/pull/121752#issuecomment-1986378328


Veykril:
Thanks!
@bors r+

PR #16805, opened 2024-03-11
feat: Implement resolving and lowering of Lifetimes (no inference yet)

dfireBird:


crates/hir-def/src/resolver.rs
```diff
@@ -418,6 +425,24 @@ impl Resolver {
...
+            match scope {
+                Scope::GenericParams { def, params } => {
+                    if let Some(id) = params.find_lifetime_by_name(&lifetime.name, *def) {
+                        return Some(LifetimeNs::LifetimeParam(id));
+                    }
+                }
+                _ => continue,
+            }
+        }
+        None
```
Veykril, line 443:
```suggestion
        self.scopes().find_map(|scope| {
            match scope {
                Scope::GenericParams { def, params } => {
                    params.find_lifetime_by_name(&lifetime.name, *def).map(|id| LifetimeNs::LifetimeParam(id))
                }
                _ => None,
            }
        })
```

crates/hir-ty/src/display.rs
```diff
@@ -938,18 +939,32 @@ impl HirDisplay for Ty {
...
+                        const_params,
+                        _impl_trait_params,
+                        lifetime_params,
+                    ) = generics.provenance_split();
+                    let total_len =
+                        parent_params + self_param + type_params + const_params + lifetime_params;
                     // We print all params except implicit impl Trait params. Still a bit weird; should we leave out parent and self?
                     if total_len > 0 {
                         // `parameters` are in the order of fn's params (including impl traits),
                         // parent's params (those from enclosing impl or trait, if any).
```
Veykril, line 955: Would be good to add where the lifetime params here are

crates/hir-ty/src/display.rs
```diff
@@ -1327,16 +1355,12 @@ fn hir_fmt_generics(
         } else {
             parameters.as_slice(Interner)
         };
-        if !parameters_to_write.is_empty() || lifetime_args_count != 0 {
+        //FIXME: Should handle when creating substitutions
```
Veykril, line 1358: Should handle what? The ordering I presume, would be good to make that clear

crates/hir-ty/src/lower.rs
```diff
@@ -275,9 +277,10 @@ impl<'a> TyLoweringContext<'a> {
                 let inner_ty = self.lower_ty(inner);
                 TyKind::Slice(inner_ty).intern(Interner)
             }
-            TypeRef::Reference(inner, _, mutability) => {
+            TypeRef::Reference(inner, lifetime, mutability) => {
                 let inner_ty = self.lower_ty(inner);
-                let lifetime = static_lifetime();
+                let lifetime =
+                    lifetime.as_ref().map_or_else(static_lifetime, |lr| self.lower_lifetime(lr));
```
Veykril, line 283: Might be good to add a FIXME that this should infer the elided lifetime instead of setting it static

crates/hir-ty/src/lower.rs
```diff
@@ -867,15 +887,46 @@ impl<'a> TyLoweringContext<'a> {
                     }
                 }
             }
+
+            for arg in generic_args
+                .args
+                .iter()
+                .filter(|arg| matches!(arg, GenericArg::Lifetime(_)))
+                .skip(skip)
```
Veykril, line 895: I think this skip is wrong? It accounts for the self type which is a `GenericArg::TypeParam`, those are already filtered out by the filter at this point though, so we might skip a lifetime here. I don't think the skip is needed at all in this case.

crates/hir-ty/src/lower.rs
```diff
@@ -867,15 +887,46 @@ impl<'a> TyLoweringContext<'a> {
...
+                .take(expected_num)
```
Veykril, line 896:
```suggestion
                .take(lifetime_params)
```

crates/hir-ty/src/lower.rs
```diff
@@ -837,9 +856,9 @@ impl<'a> TyLoweringContext<'a> {
                 fill_self_params();
             }
             let expected_num = if generic_args.has_self_type {
-                self_params + type_params + const_params
+                self_params + type_params + const_params + lifetime_params
             } else {
-                type_params + const_params
+                type_params + const_params + lifetime_params
             };
```
Veykril, line 862: This is used after the `.filter(|arg| !matches!(arg, GenericArg::Lifetime(_)))` on line 868, so I don't think it should account for lifetimes?

crates/hir-ty/src/utils.rs
```diff
@@ -270,64 +270,129 @@ pub(crate) struct Generics {
 }
 
 impl Generics {
```
Veykril, line 272: The changes in this module are not too great perf wise, instead of borrowing we do a lot of cloning now I wonder if we can get around this somehow.

crates/hir-ty/src/lower.rs
dfireBird: It is wrong, I thought to remove it but then I forgot about it.

crates/hir-ty/src/lower.rs
dfireBird: I thought about it too, but I didn't change since it's used in take and take only returns what it has available even in the case of bigger number, that's why I didn't change it. Also enforces that `expected_num` includes lifetimes too. This is just my point of view. I could remove it if you want.

crates/hir-ty/src/display.rs
Veykril: No worries, I sometimes make those kinds of vague FIXMEs myself still 😅

crates/hir-ty/src/utils.rs
dfireBird: Is there any specific reason why there is separate arena for lifetimes in `GenericParams`, if we could combine both. It could potential solve this as well having ordering of lifetimes issue.

flodiebold: I don't think it would solve the ordering issues. I think you could just add something like this:
```rust
enum GenericParamDataRef<'a> {
    TypeParamData(&'a TypeParamData),
    ConstParamData(&'a ConstParamData),
    LifetimeParamData(&'a LifetimeParamData),
}
```

dfireBird: I wasn't sure whether I could a add new enum since there is type with exact same data.

I was thinking that could solve the issue but now that I think about it, It's not going to solve that ordering issues and will also have other problematic things.
But what do you think about adding a flag for the presence of Self type and generating the iterator with Self at the start and then other lifetimes and types if the flag is set?

dfireBird:
Well this is fun, clippy shows some suggestions when the suggested changes are made, it's not compiling :sweat_smile:

Veykril:
You should be able to replace both `'a` in the return type with `'_` and drop the remaining `'a`'s in the signature.

dfireBird:
I have tried that, but the closures inside `iter_parent` requires the lifetime to be specified, so I had to create it here as well.
I feel like something is wrong there but I just don't know what is.

Veykril:
Right, I recall that lint having a few false positives. It's fine to annotate it iwth an `#[allow(...)]` then

Veykril:
Looks like the frequent cloning that was replaced with `GenericParamDataRef` brought down the perf decrese considerably. Type inference went from ~68 to ~70 seconds on my machine which is within possible noise.

Veykril:
I'd say lets test it since its still the start of the week :) Thanks!
@bors r+

dfireBird:
> I'd say lets test it since its still the start of the week :) 
Let's hope everything goes good.

PR #16808, opened 2024-03-11
feat: Add proc macro semantic token type

### LINKED ISSUE #11529: Add `proc-macro` semantic token type
We currently have `attribute`, `macro` and `derive` types for highlighting as well as the `attribute` modifier for attribute invocation tokens.

being able to show if a macro call is a proc-macro seems like a good addition, so I believe adding a `procMacro` token type for specifically function like proc-macros seems like a good addition.

The other option would be to add a `procMacro` modifier and tag attributes, derives and function-like proc-macros with it. I feel like this offers less value though, given attributes and derives are always proc-macros anyways.

On another note, given the attribute modifier we may as well want a `macro` modifier that we tag all tokens inside a macro! call with.

ShoyuVanilla:
Closes  #11529


crates/hir/src/semantics.rs
```diff
@@ -1236,6 +1236,15 @@ impl<'db> SemanticsImpl<'db> {
         sa.resolve_macro_call(self.db, macro_call)
     }
 
+    pub fn is_proc_macro_call(&self, macro_call: &ast::MacroCall) -> bool {
+        let sa = match self.analyze(macro_call.syntax()) {
+            Some(it) => it,
+            None => return false,
+        };
+        let macro_call = self.find_file(macro_call.syntax()).with_value(macro_call);
+        sa.is_proc_macro_call(self.db, macro_call)
```
Veykril, line 1245:
```suggestion
        self.resolve_macro_call(macro_call).map_or(false, |m| matches!(m.id, MacroId::ProcMacroId(..))
```

No need for `SourceAnalyzer::is_proc_macro_call` then

crates/ide/src/syntax_highlighting.rs
```diff
@@ -298,8 +299,9 @@ fn traverse(
                         ast::Item::Fn(_) | ast::Item::Const(_) | ast::Item::Static(_) => {
                             bindings_shadow_count.clear()
                         }
-                        ast::Item::MacroCall(_) => {
+                        ast::Item::MacroCall(ref macro_call) => {
                             inside_macro_call = true;
+                            inside_proc_macro_call = sema.is_proc_macro_call(macro_call);
```
Veykril, line 304:
```suggestion
                        ast::Item::MacroCall(macro_call) => {
                            inside_macro_call = true;
                            inside_proc_macro_call = sema.is_proc_macro_call(&macro_call);
```

crates/ide/src/syntax_highlighting.rs
```diff
@@ -519,7 +522,11 @@ fn traverse(
                 highlight |= HlMod::Attribute
             }
             if inside_macro_call && tt_level > 0 {
-                highlight |= HlMod::Macro
+                if inside_proc_macro_call {
+                    highlight |= HlMod::ProcMacro
+                } else {
+                    highlight |= HlMod::Macro
+                }
```
Veykril, line 529:
```suggestion
                if inside_proc_macro_call {
                    highlight |= HlMod::ProcMacro
                }
                highlight |= HlMod::Macro
```
I think this makes more sense, that way one doesn't have to handle the proc macro modifier specifically if they are only interested in macros in general

crates/ide/src/syntax_highlighting.rs
ShoyuVanilla: The borrow checker grumbles here 😅 

```
error[E0382]: borrow of partially moved value: `item`
   --> crates/ide/src/syntax_highlighting.rs:310:52
    |
287 |                 if let Some(item) = ast::Item::cast(node.clone()) {
    |                             ---- this reinitialization might get skipped
...
302 |                         ast::Item::MacroCall(macro_call) => {
    |                                              ---------- value partially moved here
...
310 |                         if sema.is_attr_macro_call(&item) {
    |                                                    ^^^^^ value borrowed here after partial move
    |
    = note: partial move occurs because value has type `syntax::ast::MacroCall`, which does not implement the `Copy` trait
help: borrow this binding in the pattern to avoid moving the value
    |
302 |                         ast::Item::MacroCall(ref macro_call) => {
    |                                              +++

```

Veykril: Oh, I didn't notice the `continue`s, nevermind then.

Veykril:
Thanks!
@bors r+

PR #16812, opened 2024-03-11
fix: Goto implementation to impls inside blocks

### LINKED ISSUE #3739: '0 implementations' on type nested in fn
I have the following:

```rust
#[derive(Clone)]
struct Foo;
fn bar() {
    #[derive(Clone)]
    struct Bar;
}
````

which shows '1 implementation' on `Foo` but '0 implementations' on `Bar`. I'm using version 0.2.100 as listed in VSCode.

### PR DESCRIPTION
**Pull request by ShoyuVanilla**
Fixes #3739

### DIFF
--- a/crates/ide/src/goto_implementation.rs
+++ b/crates/ide/src/goto_implementation.rs
@@ -340,4 +340,75 @@ impl Tr for S {
 "#,
         );
     }
+
+    #[test]
+    fn goto_adt_implementation_inside_block() {
+        check(
+            r#"
+//- minicore: copy, derive
+trait Bar {}
+
+fn test() {
+    #[derive(Copy)]
+  //^^^^^^^^^^^^^^^
+    struct Foo$0;
+
+    impl Foo {}
+       //^^^
+
+    trait Baz {}
+
+    impl Bar for Foo {}
+               //^^^
+
+    impl Baz for Foo {}
+               //^^^
+}
+"#,
+        );
+    }
+
+    #[test]
+    fn goto_trait_implementation_inside_block() {
+        check(
+            r#"
+struct Bar;
+
+fn test() {
+    trait Foo$0 {}
+
+    struct Baz;
+
+    impl Foo for Bar {}
+               //^^^
+
+    impl Foo for Baz {}
+               //^^^
+}
+"#,
+        );
+        check(
+            r#"
+struct Bar;
+
+fn test() {
+    trait Foo {
+        fn foo$0() {}
+    }
+
+    struct Baz;
+
+    impl Foo for Bar {
+        fn foo() {}
+         //^^^
+    }
+
+    impl Foo for Baz {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/hir-ty/src/method_resolution.rs [hunk 1]
```diff
@@ -216,7 +259,30 @@ impl TraitImpls {
...
+                        _ => None,
+                    };
+                    if let Some(body) = body {
+                        let body = db.body(body);
+                        for (_, block_def_map) in body.blocks(db.upcast()) {
+                            Self::collect_def_map(db, map, &block_def_map, query_depth);
+                        }
+                    }
                 }
             }
```
ShoyuVanilla, line 287: The logic is quite verbose, but I couldn't find any good ways that can probe into those blocks 🤔

crates/ide/src/goto_implementation.rs
```diff
@@ -337,6 +337,42 @@ impl Tr for S {
...
+    #[derive(Copy)]
+  //^^^^^^^^^^^^^^^
+    struct Foo$0;
+}
+"#,
+        );
+    }
+
+    #[test]
+    fn goto_implementation_in_nested_blocks_2() {
```
Veykril, line 359: This test we cannot support due to performance reasons, see the main review message

crates/ide/src/goto_implementation.rs
```diff
@@ -337,6 +337,42 @@ impl Tr for S {
     const C: usize = 4;
         //^
 }
+"#,
+        );
+    }
+
+    #[test]
+    fn goto_implementation_in_nested_blocks_1() {
```
Veykril, line 345: This we can support

Veykril (changes requested):
We intentionally do not look into all blocks there are, as an IDE we try to be lazy and only analyze the crate structure on a global scale, which allows us to skip out on having to look into every body which would slow us down immensely. So supporting local impls like this isn't really possible. 

So what we should do instead to make the first case work is have `goto_implementation` look into the block defmaps of the ADT/trait definition and iterate those. We can query the relevant impls via the `*_impls_in_block` queries (check for uses of those queries, that should show you how to use thisI think, if not just ask)

ShoyuVanilla:
I'm seeking some performant implementation; is there any way to find `BlockId` with current cursor location? 🤔

Veykril:
You can turn block expression ast nodes into blocks via https://github.com/rust-lang/rust-analyzer/blob/0dd89d7ee7800c53dc10cec941c4e740562d54df/crates/hir/src/semantics/source_to_def.rs#L205 (Note that will only yield `Some` for blocks that actually have local items or macro calls in them due to perf optimizations)

ShoyuVanilla:
Thank you! I'll look into that point 😄

Veykril:
Thanks!
@bors r+

PR #16824, opened 2024-03-12
minor: Fix unwanted leading whitespace in hover text

Wilfred:
PR #16366 moved layout information to a separate line, so the leading whitespace is no longer necessary.


lnicola (approved):

lnicola:
Thanks!

@bors r+

PR #16825, opened 2024-03-12
RFC: Hide size/layout information by default

Wilfred:
Most of the time, I'm not interested in the exact layout of types. However, rust-analyzer shows the information very prominently, even before the doc comment.

Screenshot 2024-03-12 at 12 38 03

I've fixed the whitespace in #16824, but I feel like opt-in is a better default for layout information. I know that this has been the behaviour for a while, but the layout information is much more visible after cf905cff7613fc2425244facb4849b7f23208ef5.

What do you think?

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -1089,7 +1089,7 @@
                 },
                 "rust-analyzer.hover.memoryLayout.enable": {
                     "markdownDescription": "Whether to show memory layout data on hover.",
-                    "default": true,
+                    "default": false,
                     "type": "boolean"
                 },
                 "rust-analyzer.hover.memoryLayout.niches": {
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -370,7 +370,7 @@ config_data! {
         /// How to render the align information in a memory layout hover.
         hover_memoryLayout_alignment: Option<MemoryLayoutHoverRenderKindDef> = "\"hexadecimal\"",
         /// Whether to show memory layout data on hover.
-        hover_memoryLayout_enable: bool = "true",
+        hover_memoryLayout_enable: bool = "false",
         /// How to render the niche information in a memory layout hover.
         hover_memoryLayout_niches: Option<bool> = "false",
         /// How to render the offset information in a memory layout hover.
--- a/docs/user/generated_config.adoc
+++ b/docs/user/generated_config.adoc
@@ -500,7 +500,7 @@ Use markdown syntax for links on hover.
 --
 How to render the align information in a memory layout hover.
 --
-[[rust-analyzer.hover.memoryLayout.enable]]rust-analyzer.hover.memoryLayout.enable (default: `true`)::
+[[rust-analyzer.hover.memoryLayout.enable]]rust-analyzer.hover.memoryLayout.enable (default: `false`)::
 +
 --
 Whether to show memory layout data on hover.


editors/code/package.json
```diff
@@ -1089,7 +1089,7 @@
                 },
                 "rust-analyzer.hover.memoryLayout.enable": {
                     "markdownDescription": "Whether to show memory layout data on hover.",
-                    "default": true,
+                    "default": false,
```
davidbarsky, line 1092: instead of changing this here, you're going to `crates/rust-analyzer/src/config.rs` and change this line: https://github.com/rust-lang/rust-analyzer/blob/a2e274142f35d21fd28d28655f4af8e8531ab649/crates/rust-analyzer/src/config.rs#L373. You'll then want to run `cargo test`, as that will run a test that updates a `package.json` (and a few other files, iirc) to the appropriate default.

Put differently, this isn't the source of truth: `config.rs` is.

davidbarsky (changes requested):

editors/code/package.json
Wilfred: Thanks, fixed.

HKalbasi:
I think it should be enabled by default due discoverability reasons. People can disable features they found annoying and noisy, but they can't enable features that they don't know exist.

What about placing it in its old place? Or some other place, like at the end of doc comments, or even making its place configurable, with its default being something less invasive.

Wilfred:
FWIW there is some precedent for off-by-default features: some of the inlay hint settings and the niches option are opt-in. I suppose it depends what proportion of users want to see size and alignment info at all times.

That said, I didn't find this information obtrusive previously, so putting it back on its own line would also suit me :)

Veykril:
The placement I personally think is good where it is now (as I made that change :^), putting after docs is a plain bad choice because docs can be long and I don't want to scroll down all the way when I actually care about the info. Putting it where it was before (behind the closing brace) is imo also no ideal, we render more info than we used to for on hovers (like where clauses and fields now) which makes it less visually appealing if it comes after, and usually you put comments above the thing they describe, not below.

Enabling by default due to discoverability is always a point to have, and its pretty subjective by the team whether we deem that sufficient or not (as was raised, some inlay hints are off by default due to them being rather noisy in general even though that means people won't necessarily discover them). This problem would be solved entirely if we had a landing page on installation in VSCode https://github.com/rust-lang/rust-analyzer/issues/13351 which could introduce all the features. I think it would be fine for us to change the default here, in most cases this info is not relevant to people and we've had this on for a while that the ones that do make use of it will wonder where the info went and presumably look into the config to check if they can re-enable it. 

Another question this raises to me is if we want to maybe change it so they are enabled by default for types with actual stable layouts (aka `repr(C)` etc), but disabled otherwise?

lnicola:
This doesn't seem too noisy to me, and it's probably easy enough to disable. I think we'd get more angry users (like with `needless_return`) if it was a problem.

> This problem would be solved entirely if we had a landing page on installation in VSCode

We already have a lot of settings, and a lot of users would probably space out after going through 20 or so and just accept the defaults.

davidbarsky:
I think my biggest complaint with the size/layout information is that they look like comments, which throws me for a loop. If they rendered with the system UI font—and I'm not sure how feasible that is—then I think I'd be perfectly fine with them as the default.

> Another question this raises to me is if we want to maybe change it so they are enabled by default for types with actual stable layouts (aka `repr(C)` etc), but disabled otherwise?

I *really* like that approach and would strongly advocate for it. If people are using non-`#[repr(Rust)]` for their structs/enums, then they probably _do_ care about size/layout information, but people who are using `Arc<Mutex<SomeStruct>>` probably don't—it's just noise.

(If people care about size/layout information and they're *not* using annotating things with-`#[repr(Rust)]`, then I think they're the sort of power user who would be comfortable changing a setting.)

HKalbasi:
> but people who are using Arc<Mutex<SomeStruct>> probably don't—it's just noise

One usage of those annotations is to optimize the memory usage, so the size of `SomeStruct` might be important even when it is not `#[repr(C)]` and is used in a `Arc<Mutex>`.

lnicola:
Right. If a struct is `#[repr(C)]` because you're calling into C, you don't care about its layout because you can't do anything about it.

Veykril:
> At least, when formatted as is.

Happy to take alternative formatting ideas if you have any fwiw

HKalbasi:
For the record, clangd render it in this way:
image

It is enabled by default, but only appears in definition hover, not in usage hover.

Wilfred:
@HKalbasi ooh, that seems like a really nice heuristic. Showing the layout information on definition sites means the information is available, but not too prominent.

It also looks like they're taking advantage of the fact that `textDocument/hover` can return multiple items, so the size can be rendered as markdown rather than source code.

Veykril:
> It also looks like they're taking advantage of the fact that textDocument/hover can return multiple items, so the size can be rendered as markdown rather than source code.

We are already returning "multiple results" as well (its still just one hover result, but concatenated markdown delimited via `---`), the main reason for us using comment syntax is that it was simple to come up with.

Wilfred:
Apparently this is a UI quirk of Emacs that only shows the first line, I've also created a discussion here: https://github.com/joaotavora/eglot/discussions/1422

PR #16835, opened 2024-03-14
Have Derive Attribute share a token tree with it's proc macros.

wyatt-herkamp:
The goal of this PR is to stop creating a token tree for each derive proc macro. 

This is done by giving the derive proc macros an id to its parent derive element. 

From running the analysis stat on the rust analyzer project I did see a small memory decrease.

```
Inference:           42.80s, 362ginstr, 591mb
MIR lowering:        8.67s, 67ginstr, 291mb
Mir failed bodies: 18 (0%)
Data layouts:        85.81ms, 609minstr, 8mb
Failed data layouts: 135 (6%)
Const evaluation:    440.57ms, 5235minstr, 13mb
Failed const evals: 1 (0%)
Total:               64.16s, 552ginstr, 1731mb
```
After Change
```
Inference:           40.32s, 340ginstr, 593mb
MIR lowering:        7.95s, 62ginstr, 292mb
Mir failed bodies: 18 (0%)
Data layouts:        87.97ms, 591minstr, 8mb
Failed data layouts: 135 (6%)
Const evaluation:    433.38ms, 5226minstr, 14mb
Failed const evals: 1 (0%)
Total:               60.49s, 523ginstr, 1680mb
```

Currently this breaks the expansion for the actual derive attribute. 

## TODO
- [x] Pick a better name for the function `smart_macro_arg`

### DIFF
--- a/crates/hir-def/src/nameres/collector.rs
+++ b/crates/hir-def/src/nameres/collector.rs
@@ -14,6 +14,7 @@ use crate::item_tree::Mod;
 use crate::item_tree::ModKind;
 use crate::macro_call_as_call_id_with_eager;
 
+use crate::nameres::attr_resolution::derive_attr_macro_as_call_id;
 use crate::nameres::mod_resolution::ModDir;
 
 use crate::item_tree::ItemTree;
@@ -1258,9 +1259,7 @@ impl DefCollector<'_> {
                         Some(def) if def.is_attribute() => def,
                         _ => return Resolved::No,
                     };
-                    // We will treat derive macros as an attribute as a reference for the input to derives
-                    let call_id =
-                        attr_macro_as_call_id(self.db, file_ast_id, attr, self.def_map.krate, def);
+
                     if let MacroDefId {
                         kind:
                             MacroDefKind::BuiltInAttr(
@@ -1289,8 +1288,16 @@ impl DefCollector<'_> {
                                 return recollect_without(self);
                             }
                         };
-                        let ast_id = ast_id.with_value(ast_adt_id);
 
+                        let ast_id = ast_id.with_value(ast_adt_id);
+                        // the call_id for the actual derive macro. This is used so all the derive proc macros can share a token tree
+                        let call_id = derive_attr_macro_as_call_id(
+                            self.db,
+                            &ast_id,
+                            attr,
+                            self.def_map.krate,
+                            def,
+                        );
                         match attr.parse_path_comma_token_tree(self.db.upcast()) {
                             Some(derive_macros) => {
                                 let mut len = 0;
@@ -1338,7 +1345,8 @@ impl DefCollector<'_> {
 
                         return recollect_without(self);
                     }
-
+                    let call_id =
+                        attr_macro_as_call_id(self.db, file_ast_id, attr, self.def_map.krate, def);
                     // Skip #[test]/#[bench] expansion, which would merely result in more memory usage
                     // due to duplicating functions into macro expansions
                     if matches!(
--- a/crates/hir-def/src/nameres/collector.rs
+++ b/crates/hir-def/src/nameres/collector.rs
@@ -17,8 +17,6 @@ use crate::macro_call_as_call_id_with_eager;
 use crate::nameres::mod_resolution::ModDir;
 
 use crate::item_tree::ItemTree;
-use base_db::{CrateId, Dependency, FileId};
-use cfg::{CfgExpr, CfgOptions};
 
 use crate::item_tree::TreeId;
 
--- a/crates/hir-def/src/nameres/collector.rs
+++ b/crates/hir-def/src/nameres/collector.rs
@@ -1232,6 +1232,7 @@ impl DefCollector<'_> {
                         Some(def) if def.is_attribute() => def,
                         _ => return Resolved::No,
                     };
+
                     let call_id =
                         attr_macro_as_call_id(self.db, file_ast_id, attr, self.def_map.krate, def);
                     if let MacroDefId {
@@ -1264,7 +1265,6 @@ impl DefCollector<'_> {
                         };
 
                         let ast_id = ast_id.with_value(ast_adt_id);
-                        // the call_id for the actual derive macro. This is used so all the derive proc macros can share a token tree
 
                         match attr.parse_path_comma_token_tree(self.db.upcast()) {
                             Some(derive_macros) => {
@@ -2305,7 +2305,7 @@ impl ModCollector<'_, '_> {
 
     fn collect_macro_call(
         &mut self,
-        &MacroCall { ref path, ast_id, expand_to, ctxt, .. }: &MacroCall,
+        &MacroCall { ref path, ast_id, expand_to, ctxt }: &MacroCall,
         container: ItemContainerId,
     ) {
         let ast_id = AstIdWithPath::new(self.file_id(), ast_id, ModPath::clone(path));
@@ -2444,10 +2444,7 @@ mod tests {
     use base_db::SourceDatabase;
     use test_fixture::WithFixture;
 
-    use crate::{
-        nameres::{DefMapCrateData, ModuleData, ModuleOrigin},
-        test_db::TestDB,
-    };
+    use crate::{nameres::DefMapCrateData, test_db::TestDB};
 
     use super::*;
 
--- a/crates/hir-def/src/nameres/collector.rs
+++ b/crates/hir-def/src/nameres/collector.rs
@@ -3,85 +3,60 @@
 //! `DefCollector::collect` contains the fixed-point iteration loop which
 //! resolves imports and expands macros.
 
-use std::{iter, mem};
+use std::{cmp::Ordering, iter, mem, ops::Not};
 
-use crate::attr::Attrs;
-use crate::item_tree::Fields;
-use crate::item_tree::FileItemTreeId;
-use crate::item_tree::MacroCall;
-use crate::item_tree::MacroRules;
-use crate::item_tree::Mod;
-use crate::item_tree::ModKind;
-use crate::macro_call_as_call_id_with_eager;
-
-use crate::nameres::attr_resolution::derive_attr_macro_as_call_id;
-use crate::nameres::mod_resolution::ModDir;
-
-use crate::item_tree::ItemTree;
-
-use crate::item_tree::TreeId;
-
-use crate::LocalModuleId;
-use crate::{
-    item_tree::{ExternCrate, ItemTreeId, Macro2, ModItem},
-    nameres::{
-        diagnostics::DefDiagnostic, proc_macro::parse_macro_name_and_helper_attrs,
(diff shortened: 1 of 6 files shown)


Veykril (commented):
That's a smart idea, re-using the `#[derive]` attribute for the query, since all the inner derives should share the arg! Though this raises the question what to do about the "pseudo" expansion that we use this for right now... Guess it's time to move that hack into a different layer, the question is how

wyatt-herkamp:
> That's a smart idea, re-using the `#[derive]` attribute for the query, since all the inner derives should share the arg! 

Thank you! I have changed how I did this though I forgot to commit the code. I will have that up soon.

> Though this raises the question what to do about the "pseudo" expansion that we use this for right now... Guess it's time to move that hack into a different layer, the question is how

So I was wanting to add a new setting and stuff.

The default would be 
If you request the expansion on the `#[derive]` it would just return the struct but with all the "fixups" applied. Including see what `cfg_attr` go are removed and what is applied

The second would be collect all the derives and return them as one expansion. 

This would only really matter for the `rust-analyzer/expandMacro` command or request. We would need to add seperate handling for this under [expand_macro](https://github.com/wyatt-herkamp/rust-analyzer/blob/948280391f8764b5b4946d5036a1b3162b8bd825/crates/ide/src/expand_macro.rs#L26)

wyatt-herkamp:
https://github.com/wyatt-herkamp/rust-analyzer/commit/8b78a18e5401acdc1db432a85a436acf4917a885 adds a new MacroCallKind [DeriveAttr](https://github.com/wyatt-herkamp/rust-analyzer/blob/8b78a18e5401acdc1db432a85a436acf4917a885/crates/hir-expand/src/lib.rs#L232) 

The reason I did this is because handling `cfg`/`cfg_attr` and other derive related censors. This also allows us to go about handling expansion requests on the derive attribute a bit easier. We also can only apply it on an `Adt`(Struct, Enum, Union).

crates/hir-expand/src/db.rs
```diff
@@ -338,13 +339,39 @@ pub(crate) fn parse_with_map(
...
+/// Imagine the word smart in quotes.
+///
+/// This resolves the [MacroCallId] to check if it is a derive macro if so get the [macro_arg] for the derive.
+/// Other wise return the [macro_arg] for the macro_call_id.
+///
+/// This is not connected to the database so it does not cached the result. However, the inner [macro_arg] query is
+///
+/// FIXME: Pick a better name
+fn smart_macro_arg(db: &dyn ExpandDatabase, id: MacroCallId) -> MacroArgResult {
+    let loc = db.lookup_intern_macro_call(id);
```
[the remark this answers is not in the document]
wyatt-herkamp: I think I did this correct?

wyatt-herkamp:
@Veykril should any tests be added for this PR?

crates/hir-expand/src/db.rs
```diff
@@ -338,13 +339,26 @@ pub(crate) fn parse_with_map(
...
 }
+/// This is just to ensure the types of smart_macro_arg and macro_arg are the same
+type MacroArgResult = (Arc<tt::Subtree>, SyntaxFixupUndoInfo, Span);
+/// Imagine the word smart in quotes.
+///
+/// This resolves the [MacroCallId] to check if it is a derive macro if so get the [macro_arg] for the derive.
+/// Other wise return the [macro_arg] for the macro_call_id.
+///
+/// This is not connected to the database so it does not cached the result. However, the inner [macro_arg] query is
+fn macro_arg_considering_derives(db: &dyn ExpandDatabase, id: MacroCallId) -> MacroArgResult {
```
Veykril, line 350: We should add this as a `#[salsa::transparent]` query to the `ExpandDatabase`, as `ExpansionInfo::new` should use this instead of the `macro_arg` query as well.

crates/hir-expand/src/db.rs
```diff
@@ -338,13 +339,26 @@ pub(crate) fn parse_with_map(
...
+    let macro_call_kind = db.lookup_intern_macro_call(id).kind;
+    // FIXME: We called lookup_intern_macro_call twice.
```
Veykril, line 352: Just add the `MacroCallKind` as an additional param here, a bit ugly but that's not really too problematic in comparison to the extra lookup

crates/hir-expand/src/lib.rs
```diff
@@ -225,6 +225,13 @@ pub enum MacroCallKind {
         derive_attr_index: AttrId,
         /// Index of the derive macro in the derive attribute
         derive_index: u32,
+        /// The "parent" macro call.
+        /// We will resolve the same token tree for all derive macros in the same derive attribute.
+        derive_macro_id: MacroCallId,
+    },
+    DeriveAttr {
+        ast_id: AstId<ast::Adt>,
+        invoc_attr_index: AttrId,
```
Veykril, line 234: We actually used to have this (as a field for `Attr`), but I got rid of that because it kind of infected the rest of the codebase with various checks which wasn't that great. Generally speaking there isn't really a reason for us to differentiate those two attributes. The only reason we'd need to here is to make our life easier wrt the IDE layer and derive attribute inputs. So I'd rather not we added this distinction back and instead figure out if we can explicitly encode this hack as a separate thing somehow (especially because there are other attributes, specifically builtin ones) where I was planning on employing this trick). I'll think about that a bit.

crates/hir-expand/src/db.rs
```diff
@@ -338,13 +339,39 @@ pub(crate) fn parse_with_map(
...
+type MacroArgResult = (Arc<tt::Subtree>, SyntaxFixupUndoInfo, Span);
+/// Imagine the word smart in quotes.
+///
+/// This resolves the [MacroCallId] to check if it is a derive macro if so get the [macro_arg] for the derive.
+/// Other wise return the [macro_arg] for the macro_call_id.
+///
+/// This is not connected to the database so it does not cached the result. However, the inner [macro_arg] query is
+///
+/// FIXME: Pick a better name
+fn smart_macro_arg(db: &dyn ExpandDatabase, id: MacroCallId) -> MacroArgResult {
```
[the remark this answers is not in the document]
wyatt-herkamp: > not really great of a name either but better than a non descriptive word like smart I think?

A reason the comment `/// Imagine the word smart in quotes.` 😂

crates/hir-expand/src/lib.rs
wyatt-herkamp: I think separating also provides a higher understanding that these are different. Like we can see the expansion of the individual derive elements.

wyatt-herkamp: I agree we should treat derive as a normal everyday attribute macro as much as possible. But, treating it specially  does two things.

We ensure the type of the syntax tree connected is an Adt. We also can do any other processing that could be required. Such as cfg stripping and stripping the `#[derive]`

I could undo this and at [MacroCallKind::Attr { ast_id, invoc_attr_index, .. }](https://github.com/wyatt-herkamp/rust-analyzer/blob/c376addfcc46f7eeec02d765027c8febc17a5825/crates/hir-expand/src/db.rs#L454-L455)
We would need to grab the attribute. Get it's name. Check if it is derive. If derive we would also have to pull the syntax tree and try to convert it to a an Adt. 

If that succeeds. We then can run the derive stripping and cfg stripping.

I personally thinks this adds complexity here. I think explicitly treating derive specially allows for us to have an easier experience when expanding them.

wyatt-herkamp: That makes a lot of sense. I will get on it!

crates/hir-def/src/nameres/collector.rs
```diff
@@ -1252,7 +1262,9 @@ impl DefCollector<'_> {
                                 return recollect_without(self);
                             }
                         };
+
                         let ast_id = ast_id.with_value(ast_adt_id);
+                        // the call_id for the actual derive macro. This is used so all the derive proc macros can share a token tree
```
Veykril, line 1267: This comment seems misplaced

crates/hir-def/src/nameres/collector.rs
```diff
@@ -2289,7 +2305,7 @@ impl ModCollector<'_, '_> {
 
     fn collect_macro_call(
         &mut self,
-        &MacroCall { ref path, ast_id, expand_to, ctxt }: &MacroCall,
+        &MacroCall { ref path, ast_id, expand_to, ctxt, .. }: &MacroCall,
```
Veykril, line 2308: This change should be unnecessary

crates/hir-def/src/nameres/collector.rs
```diff
@@ -2428,7 +2444,10 @@ mod tests {
     use base_db::SourceDatabase;
     use test_fixture::WithFixture;
 
-    use crate::{nameres::DefMapCrateData, test_db::TestDB};
+    use crate::{
+        nameres::{DefMapCrateData, ModuleData, ModuleOrigin},
+        test_db::TestDB,
+    };
```
Veykril, line 2450: likewise

crates/hir-expand/src/db.rs
```diff
@@ -339,12 +346,26 @@ pub(crate) fn parse_with_map(
     }
 }
 
-// FIXME: for derive attributes, this will return separate copies of the same structures! Though
-// they may differ in spans due to differing call sites...
-fn macro_arg(
+/// Imagine the word smart in quotes.
```
Veykril, line 349: outdated comment

crates/hir-expand/src/db.rs
```diff
@@ -414,6 +435,8 @@ fn macro_arg(
             }
             return (Arc::new(tt), SyntaxFixupUndoInfo::NONE, span);
         }
+
+        // MacroCallKind::Derive should not be here. As we are getting the argument for the derive macro
```
Veykril, line 439: Let's replace this match arm with just an 
```unreachable!("`ExpandDatabase::macro_arg` called with `MacroCallKind::Derive`")```

File: crates/hir-expand/src/db.rs [hunk 11]
```diff
@@ -603,7 +637,7 @@ fn proc_macro_span(db: &dyn ExpandDatabase, ast: AstId<ast::Fn>) -> Span {
 
 fn expand_proc_macro(db: &dyn ExpandDatabase, id: MacroCallId) -> ExpandResult<Arc<tt::Subtree>> {
     let loc = db.lookup_intern_macro_call(id);
-    let (macro_arg, undo_info, span) = db.macro_arg(id);
+    let (macro_arg, undo_info, span) = db.macro_arg_considering_derives(id, &loc.kind.clone());
```
Veykril, line 640:
```suggestion
    let (macro_arg, undo_info, span) = db.macro_arg_considering_derives(id, &loc.kind);
```

Veykril (commented):
`speculative_expand` also needs to call the new macro arg stuff

Veykril:
Thanks!
@bors r+

PR #16839, opened 2024-03-14
Refactor extension to support arbitrary shell command runnables

Wilfred:
Currently, the extension assumes that all runnables invoke cargo. Arguments are sometimes full CLI arguments, and sometimes arguments passed to a cargo subcommand.

Refactor the extension so that tasks are just a `program` and a list of strings `args`, and rename `CargoTask` to `RustTask` to make it generic.

(This was factored out of #16135 and tidied.)

### DIFF
--- a/editors/code/src/tasks.ts
+++ b/editors/code/src/tasks.ts
@@ -2,7 +2,6 @@ import * as vscode from "vscode";
 import * as toolchain from "./toolchain";
 import type { Config } from "./config";
 import { log } from "./util";
-import { unwrapUndefinable } from "./undefinable";
 
 // This ends up as the `type` key in tasks.json. RLS also uses `cargo` and
 // our configuration should be compatible with it so use the same key.
@@ -10,10 +9,10 @@ export const TASK_TYPE = "cargo";
 export const TASK_SOURCE = "rust";
 
 export interface RustTargetDefinition extends vscode.TaskDefinition {
+    program: string;
     args: string[];
     cwd?: string;
     env?: { [key: string]: string };
-    overrideCargo?: string;
 }
 
 class RustTaskProvider implements vscode.TaskProvider {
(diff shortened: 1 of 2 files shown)


editors/code/src/tasks.ts
```diff
@@ -2,22 +2,20 @@ import * as vscode from "vscode";
...
+export interface RustTargetDefinition extends vscode.TaskDefinition {
+    program: string;
+    args: string[];
     cwd?: string;
     env?: { [key: string]: string };
-    overrideCargo?: string;
 }
 
-class CargoTaskProvider implements vscode.TaskProvider {
+class RustTaskProvider implements vscode.TaskProvider {
```
davidbarsky, line 18: nit: I'm not sure there will be any *other* kind of Task run by rust-analyzer, so I wouldn't qualify with `Rust`, but I also don't feel too strongly.

editors/code/src/tasks.ts
```diff
@@ -39,14 +37,15 @@ class CargoTaskProvider implements vscode.TaskProvider {
             { command: "run", group: undefined },
         ];
 
+        const cargoPath = await toolchain.cargoPath();
```
davidbarsky, line 40: I don't know if this should happen in *this* PR, but I wonder whether `toolchain.ts` should _even exist_ or if the toolchain contents shouldn't be synced over from rust-analyzer itself over a request, given the work that Lukas did to make toolchain discovery more reliable. there's two separate implementations of this discovery mechanism, and the one in the extension is substantially less robust.

if we're to switch over, i think it's worth considering keeping the old and new versions of runnables around before removing the older version.

editors/code/src/tasks.ts
Wilfred: I think it needs *some* name, `class TaskProvider implements vscode.TaskProvider` would be confusing. I don't feel strongly about the name though. Do you have any suggestions? `RustAnalyzerTaskProvider` is maybe a little long.

editors/code/src/tasks.ts
Wilfred: This PR (especially with #16840 being a larger change) is probably better as a small focused PR, but I agree it would be nice to follow up on how toolchains are handled :)

Veykril: Yes, we should ideally replicate the toolchain discovery here / allow clients to query the server for the correct toolchain stuff. Not doing that can cause recompilations and random failures. (Agree that that can go into a separate PR though)

Issue #12105, opened 2022-04-27
Feature request: Add a bazel Runnable kind


petr-tik:
As a developer using rust-analyzer with bazel (using rules_rust and rust-project.json generated by @rules_rust//tools/rust_analyzer:gen_rust_project)
I would like rust-analyzer to use `bazel test` for test runnables 
so that I can enjoy the same UX regardless of the build system. 

The Runnables extension is a great productivity boost and it's very easy to run a test
from a code lens. According to my experiments and the lsp-extensions.md documentation, it 
currently only supports one Runnable.kind = "cargo". 

Given the consistency and stability of the bazel test intereface, we should be able to use bazel test and pass additional command line arguments 

```bash
bazel test //path_to_rust_package:rust_test_name --test_arg $THE_TEST_SELECTED_BY_USER
```

While the --test_arg is definitely something that rust-analyzer can already reliable populate, working out the `//path_to_rust_package:rust_test_name` might require an expansion to rules_rust to add test targets to rust-project.json file. 

Once this is implemented, the same interface could be used for the Related Tests extension. 

For rust-analyzer to automatically work out the runnable kind is "bazel", rules_rust can add a field to rust-project.json that declares runnable kind to be bazel (or we can heuristically guess?). 

For the rest, rust-analyzer works like a charm with bazel, so thanks very much.

googleson78:
Another idea is to have `overrideCommand` for the tests, like what is done for `checkOnSave`

matklad:
Yup, we should have something like this. The runnables infra was originally designed with an eye towards something extensible like this (hence runnable kind existance). 

One designed constraint here is that we should glue bazel and rust-analyzer here in such a way so as not to hard-code knowledge about bazel. 

Ideally, the runnables LSP request exposes only semantic knowledge (eg, "there's a test in target `spam` with name `foo::bar::baz`"), and then there's some other extension (`bazel-rust-analyzer`) which cooks up the specific command line string to launch.

The reason for this enterprisey setup is that there are many ways to launch things. Ie, we can imagine running a test, in debugger, on embedded device and such. So we should provide customization points outside of rust-analyzer process proper to do these fancy things.

PR #16840, opened 2024-03-15
Allow rust-project.json to include arbitrary shell commands for runnables

Wilfred:
This is a follow-up on #16135, resolving the feedback raised :)

Allow rust-project.json to include shell runnables, of the form:

```
{
  "build_info": {
    "label": "//project/foo:my-crate",
    "target_kind": "bin",
    "shell_runnables": [
      {
        "kind": "run",
        "program": "buck2",
        "args": ["run", "//project/foo:my-crate"]
      },
      {
        "kind": "test_one",
        "program": "test_runner",
        "args": ["--name=$$TEST_NAME$$"]
      }
    ]
  }
}

```

If these runnable configs are present for the current crate in rust-project.json, offer them as runnables in VS Code.

This PR required some boring changes to APIs that previously only handled cargo situations. I've split out these changes as commits labelled 'refactor', so it's easy to see the interesting changes.

### DIFF
--- a/crates/project-model/src/project_json.rs
+++ b/crates/project-model/src/project_json.rs
@@ -33,7 +33,7 @@
 //!
 //! * file on disk
 //! * a field in the config (ie, you can send a JSON request with the contents
-//!   of rust-project.json to rust-analyzer, no need to write anything to disk)
+//!   of `rust-project.json` to rust-analyzer, no need to write anything to disk)
 //!
 //! Another possible thing we don't do today, but which would be totally valid,
 //! is to add an extension point to VS Code extension to register custom
@@ -55,8 +55,7 @@ use rustc_hash::FxHashMap;
 use serde::{de, Deserialize, Serialize};
 use span::Edition;
 
-use crate::cfg::CfgFlag;
-use crate::ManifestPath;
+use crate::{cfg::CfgFlag, ManifestPath, TargetKind};
 
 /// Roots and crates that compose this Rust project.
 #[derive(Clone, Debug, Eq, PartialEq)]
@@ -68,6 +67,10 @@ pub struct ProjectJson {
     project_root: AbsPathBuf,
     manifest: Option<ManifestPath>,
     crates: Vec<Crate>,
+    /// Configuration for CLI commands.
+    ///
+    /// Examples include a check build or a test run.
+    runnables: Vec<Runnable>,
 }
 
 /// A crate points to the root module of a crate and lists the dependencies of the crate. This is
@@ -88,6 +91,86 @@ pub struct Crate {
     pub(crate) exclude: Vec<AbsPathBuf>,
     pub(crate) is_proc_macro: bool,
     pub(crate) repository: Option<String>,
+    pub build: Option<Build>,
+}
+
+/// Additional, build-specific data about a crate.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Build {
+    /// The name associated with this crate.
+    ///
+    /// This is determined by the build system that produced
+    /// the `rust-project.json` in question. For instance, if buck were used,
+    /// the label might be something like `//ide/rust/rust-analyzer:rust-analyzer`.
+    ///
+    /// Do not attempt to parse the contents of this string; it is a build system-specific
+    /// identifier similar to [`Crate::display_name`].
+    pub label: String,
+    /// Path corresponding to the build system-specific file defining the crate.
+    ///
+    /// It is roughly analogous to [`ManifestPath`], but it should *not* be used with
+    /// [`crate::ProjectManifest::from_manifest_file`], as the build file may not be
+    /// be in the `rust-project.json`.
+    pub build_file: Utf8PathBuf,
+    /// The kind of target.
+    ///
+    /// Examples (non-exhaustively) include [`TargetKind::Bin`], [`TargetKind::Lib`],
+    /// and [`TargetKind::Test`]. This information is used to determine what sort
+    /// of runnable codelens to provide, if any.
+    pub target_kind: TargetKind,
+}
+
+/// A template-like structure for describing runnables.
+///
+/// These are used for running and debugging binaries and tests without encoding
+/// build system-specific knowledge into rust-analyzer.
+///
+/// # Example
+///
+/// Below is an example of a test runnable. `{label}` and `{test_id}`
+/// are explained in [`Runnable::args`]'s documentation.
+///
+/// ```json
+/// {
+///     "program": "buck",
+///     "args": [
+///         "test",
+///          "{label}",
+///          "--",
+///          "{test_id}",
+///          "--print-passing-details"
+///     ],
+///     "cwd": "/home/user/repo-root/",
+///     "kind": "testOne"
+/// }
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Runnable {
+    /// The program invoked by the runnable.
+    ///
+    /// For example, this might be `cargo`, `buck`, or `bazel`.
+    pub program: String,
+    /// The arguments passed to [`Runnable::program`].
+    ///
+    /// The args can contain two template strings: `{label}` and `{test_id}`.
+    /// rust-analyzer will find and replace `{label}` with [`Build::label`] and
+    /// `{test_id}` with the test name.
+    pub args: Vec<String>,
+    /// The current working directory of the runnable.
+    pub cwd: Utf8PathBuf,
+    pub kind: RunnableKind,
+}
+
+/// The kind of runnable.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RunnableKind {
+    Check,
+
+    /// Can run a binary.
+    Run,
+
+    /// Run a single test.
+    TestOne,
 }
 
 impl ProjectJson {
@@ -95,6 +178,7 @@ impl ProjectJson {
     ///
     /// # Arguments
     ///
+    /// * `manifest` - The path to the `rust-project.json`.
     /// * `base` - The path to the workspace root (i.e. the folder containing `rust-project.json`)
     /// * `data` - The parsed contents of `rust-project.json`, or project json that's passed via
     ///            configuration.
@@ -109,6 +193,7 @@ impl ProjectJson {
             sysroot_src: data.sysroot_src.map(absolutize_on_base),
             project_root: base.to_path_buf(),
             manifest,
+            runnables: data.runnables.into_iter().map(Runnable::from).collect(),
             crates: data
                 .crates
                 .into_iter()
@@ -125,6 +210,15 @@ impl ProjectJson {
                             (absolutize(src.include_dirs), absolutize(src.exclude_dirs))
                         }
                         None => (vec![root_module.parent().unwrap().to_path_buf()], Vec::new()),
+                    };
+
+                    let build = match crate_data.build {
+                        Some(build) => Some(Build {
+                            label: build.label,
+                            build_file: build.build_file,
+                            target_kind: build.target_kind.into(),
+                        }),
+                        None => None,
                     };
 
                     Crate {
@@ -146,6 +240,7 @@ impl ProjectJson {
                         exclude,
                         is_proc_macro: crate_data.is_proc_macro,
                         repository: crate_data.repository,
+                        build,
                     }
                 })
                 .collect(),
@@ -167,7 +262,15 @@ impl ProjectJson {
         &self.project_root
     }
 
-    /// Returns the path to the project's manifest file, if it exists.
+    pub fn crate_by_root(&self, root: &AbsPath) -> Option<Crate> {
+        self.crates
+            .iter()
+            .filter(|krate| krate.is_workspace_member)
+            .find(|krate| krate.root_module == root)
+            .cloned()
+    }
+
+    /// Returns the path to the project's manifest, if it exists.
     pub fn manifest(&self) -> Option<&ManifestPath> {
         self.manifest.as_ref()
     }
@@ -176,6 +279,10 @@ impl ProjectJson {
     pub fn manifest_or_root(&self) -> &AbsPath {
         self.manifest.as_ref().map_or(&self.project_root, |manifest| manifest.as_ref())
     }
+
+    pub fn runnables(&self) -> &[Runnable] {
+        &self.runnables
+    }
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
@@ -183,6 +290,8 @@ pub struct ProjectJsonData {
     sysroot: Option<Utf8PathBuf>,
     sysroot_src: Option<Utf8PathBuf>,
     crates: Vec<CrateData>,
+    #[serde(default)]
+    runnables: Vec<RunnableData>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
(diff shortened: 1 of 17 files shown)


davidbarsky (changes requested):

docs/dev/lsp-extensions.md
```diff
@@ -385,6 +385,17 @@ rust-analyzer supports only one `kind`, `"cargo"`. The `args` for `"cargo"` look
 }
 ```
 
+The args for `"shell"` look like this:
+
+```typescript
+{
+    kind: string;
```
nemethf, line 392: How should the LSP client use the `kind`  field?  Does it describe what the runnable does? Then how does it differ then from `label` field of the parent `Runnable`?

Also should the "shell" kind have a `expectTest` field just like the "cargo" kind?

BTW, how should the LSP client interpret the `workspaceRoot`, `expectTest`, and `overrideCargo` fields?

Additionally, I think it would make sense to copy the example from the Pull Request into the lsp-extension.md file, or just reference a user-level documentation describing this feature.

Thanks

nemethf: Also, shouldn't this PR explain the new build_info field of rust-project.json in manual.adoc?

davidbarsky:
> > This is a follow-up on https://github.com/rust-lang/rust-analyzer/pull/16135, resolving the feedback raised :)
> 
> 
> 
> So this means #16135 can be closed in favor of this one?

Yup! Just closed. Wilfred or I will push to this branch as needed.

docs/dev/lsp-extensions.md
Veykril: When `expectTest` is set the client should set the `UPDATE_EXPECT` env var to `1` when invoking the runnable. ideally there should be a field for arbitrary env var overrides in the payload, then this field could be replaced by the server instructing the client to just set the env var with that general mechanism as we'll need this for other things as well.

`overrideCargo` I think should just replace the cargo command that is being used by the client.

Veykril (approved):
Generally looks good to me (aside from what has been raised already by others)

crates/project-model/src/project_json.rs
```diff
@@ -87,6 +87,21 @@ pub struct Crate {
     pub(crate) exclude: Vec<AbsPathBuf>,
     pub(crate) is_proc_macro: bool,
     pub(crate) repository: Option<String>,
+    pub build_info: Option<BuildInfo>,
```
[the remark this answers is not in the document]
cormacrelf: @davidbarsky If I'm reading this right, you want the runnables to be stored at the root of the project.json. That will require:

- a substitution mechanism like `$label` so that the "runnable templates" can be turned into real commands for a specific crate
- probably RA using the `build_info.kind` field to figure out which runnables to offer for a given target

I also think we also want custom runnables in `build_info` for specific crates anyway. Right? Because the build system might simply know better. Substitution is nice because it saves a few MB of JSON. But ultimately you want the ability to add more to specific crates.

Because I believe you're going to want runnables per-crate anyway, and they are completely capable of handling anything substitution can, I think we can merge this PR pretty much as-is. I think the substitution mechanism is more of an optimisation than anything else.

I am happy to put together a PR that allows flycheck to use these.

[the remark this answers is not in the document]
davidbarsky: @cormacrelf 

> a substitution mechanism like `$label` so that the "runnable templates" can be turned into real commands for a specific crate

Yup: the substitution mechanism is thankfully not _too_ complex: I think it boils down to *just* a [`str::replace`](https://doc.rust-lang.org/stable/std/string/struct.String.html#method.replace).

(If you squint, I'd argue that the substitution already exists in rust-analyzer today via Cargo, but the substitution knowledge is pretty implicit...)

> probably RA using the `build_info.kind` field to figure out which runnables to offer for a given target

Will need to repage some context in, but I'm pretty sure the answer is also "yup".

> I also think we also want custom runnables in `build_info` for specific crates anyway. Right? Because the build system might simply know better. Substitution is nice because it saves a few MB of JSON. But ultimately you want the ability to add more to specific crates.

I don't know if I agree: while I'm assuming that while there *is* a smart build system, I don't think there isn't much surface area for more specialized commands for individual crates. There's also the fact that rust-analyzer is pretty darn fast even compared to buck2, which is the fastest build system I'm aware of—single digit milliseconds versus ~300 milliseconds. There's also the matter of the duplication being somewhat concerning for the [Fuschia folks](https://github.com/rust-lang/rust-analyzer/issues/15892#issuecomment-1919859373), as their `rust-project.json` is already pretty large and I'm not really inlined to make things more complicated for them, however slightly.

> Because I believe you're going to want runnables per-crate anyway, and they are completely capable of handling anything substitution can, I think we can merge this PR pretty much as-is. I think the substitution mechanism is more of an optimisation than anything else.

You are right it's an optimization, but it's a pretty cheap one. Wilfred and I need to get this over the finish line, but it's genuinely not too much work to implement it.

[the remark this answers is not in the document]
cormacrelf: OK. A good metric for the design is whether `rust-project.json` can encode all the cargo-related logic. If you set `build_info.label` to a cargo crate name, and a root level fly check template of `cargo check -p $label --message-format=json`, then we almost have that mission in the bag. The main missing piece is flycheck for a crate vs for the whole workspace, which is configurable through `check.workspace` in RA. Having two more runnable kinds, `flycheck` and `flycheckWorkspace`, should do it.

cormacrelf:
If it helps, I have been using a fork of RA along the lines that @davidbarsky was talking about. It includes almost exactly those project_json.rs changes you just pushed; the main addition is the flycheck command generation and label substitution. Basically `["rust-project", "check", "$$LABEL$$"]` as one of the root runnables with type "flycheck" works. It also keeps compatibility with `"$saved_file"`. If you want to rip anything out of my branch, go for it. The useful stuff is in the flycheck crate, you'll find it by walking back through the commits.

https://github.com/cormacrelf/rust-analyzer/commit/flycheck_label

And the rust-project changes that mirror it (this is working pretty well for us):

https://github.com/cormacrelf/buck2/tree/rust-project-check-label (IIRC)

davidbarsky:
> If it helps, I have been using a fork of RA along the lines that @davidbarsky was talking about. It includes almost exactly those project_json.rs changes you just pushed; the main addition is the flycheck command generation and label substitution. Basically `["rust-project", "check", "$$LABEL$$"]` as one of the root runnables with type "flycheck" works. It also keeps compatibility with `"$saved_file"`. If you want to rip anything out of my branch, go for it. The useful stuff is in the flycheck crate, you'll find it by walking back through the commits.
> 
> https://github.com/cormacrelf/rust-analyzer/tree/flycheck_label
> 
> And the rust-project changes that mirror it:
> 
> https://github.com/cormacrelf/buck2/tree/rust-project-check-target
> 
> Overall this has been working really well.

@cormacrelf Thanks! I've just been updating this PR to address all the things I've wanted to (runnable command interpolation; etc). I've also renamed some of the fields in `rust-project.json` made use of `{test_id}` and `${label}` for interpolation. I'll make another PR to rust-analyzer *after* this change lands in order to have `flycheck` accept a target label.

I also have a corresponding change to rust-project in [D58150427](https://www.internalfb.com/diff/D58150427); I'm landing that diff today/tomorrow and update this PR with the commit on GitHub. That might make it possible for you to reduce how many commits you/patches you need to carry in buck and `rust-analyzer`.

davidbarsky:
@Veykril can you give this a once over/delegate `bors`? i'll add the docs for `rust-project.json` here as well.

petr-tik:
I think this might address
#12105

crates/project-model/src/project_json.rs
```diff
@@ -54,9 +54,9 @@ use paths::{AbsPath, AbsPathBuf, Utf8PathBuf};
 use rustc_hash::FxHashMap;
 use serde::{de, Deserialize, Serialize};
 use span::Edition;
+use std::path::PathBuf;
```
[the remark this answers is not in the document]
davidbarsky: > The usages of this should be (now that we depend on it) Utf8Pathbuf.

Whoops, I'll switch it over.

> Though now I also wonder, shouldn't all usages of paths in the file here ideally be `AbsPathBuf`?

I think that's a larger change, but one I'd be happy to make later. I believe that relative paths work today (I did write [this code](https://github.com/facebook/buck2/blob/04e1b643492318c4362f249e07c713d088ba092a/integrations/rust-project/src/main.rs#L106-L108), after all...), so it might be a bit surprising to make this change *now*.

davidbarsky:
@cormacrelf here's the commit in rust-project: https://github.com/facebook/buck2/commit/71697233345d46ae43c55871a6f23ff8583c6749.

@petr-tik yes, I think that `bazelbuild/rules_rust` would be able to take advantage of this feature once this lands.

crates/project-model/src/project_json.rs
```diff
@@ -93,31 +94,83 @@ pub struct Crate {
     pub build: Option<Build>,
 }
 
-/// Additional metadata about a crate, used to configure runnables.
+/// Additional, build-specific data about a crate.
+///
+/// This is primarily used to
```
Wilfred, line 99: Unfinished.

crates/project-model/src/project_json.rs
```diff
@@ -67,8 +67,9 @@ pub struct ProjectJson {
     project_root: AbsPathBuf,
     manifest: Option<ManifestPath>,
     crates: Vec<Crate>,
-    /// Configuration for commands, such as CLI invocations for
-    /// a check build or a test run.
+    /// Configuration for commands.
+    ///
+    /// Examples include a check build or a test run.
```
Wilfred, line 72: I'd probably explicitly say this is CLI invocations.

crates/project-model/src/project_json.rs
```diff
@@ -93,31 +94,83 @@ pub struct Crate {
...
     /// Path corresponding to the build system-specific file defining the crate.
+    ///
+    /// It is roughly analagous to [ManifestPath], but none of the indexing/loading
+    /// operations can be directly applied/used with this file.
     pub build_file: Utf8PathBuf,
-    /// What kind of target is this crate? For example, we don't want
-    /// to offer a 'run' button for library crates.
+    /// The kind of target.
+    ///
+    /// Examples (non-exhaustively) include [TargetKind::Lib], [TargetKind::Lib],
```
Wilfred, line 118: Lib twice here.

crates/project-model/src/project_json.rs
```diff
@@ -93,31 +94,83 @@ pub struct Crate {
...
+    /// and [TargetKind::Test]. This information is used to determine what sort
+    /// runnable codelens to provide, if any.
     pub target_kind: TargetKind,
 }
 
+/// A template for runnables.
+///
+/// This to create runnables for running/debugging binaries and tests.
+/// In the future, it can be used to setup Flycheck.
```
Wilfred, line 127: I'd focus on what it can do today, rather than future work.

crates/project-model/src/project_json.rs
```diff
@@ -93,31 +94,83 @@ pub struct Crate {
...
+/// {
+///     "program": "buck",
+///     "args": [
+///         "test",
+///          "{label}",
+///          "--",
+///          "{test_id}",
+///          "--print-passing-details"
+///     ],
+///     "cwd": "/Users/dbarsky/fbsource",
```
Wilfred, line 144: How about something more generically named here? Buck conventions aren't super useful.

davidbarsky: whoops, i thought i removed this! I'll make this something like `/home/user/repo-root/`.

davidbarsky (approved):

crates/project-model/src/project_json.rs
```diff
@@ -93,31 +94,81 @@ pub struct Crate {
...
+    ///
+    /// It is roughly analogous to [ManifestPath], but none of the indexing/loading
+    /// operations can be directly applied/used with this file.
     pub build_file: Utf8PathBuf,
-    /// What kind of target is this crate? For example, we don't want
-    /// to offer a 'run' button for library crates.
+    /// The kind of target.
+    ///
+    /// Examples (non-exhaustively) include [TargetKind::Bin], [TargetKind::Lib],
+    /// and [TargetKind::Test]. This information is used to determine what sort
```
lnicola, line 117: > what sort _of_

crates/project-model/src/project_json.rs
```diff
@@ -93,31 +94,81 @@ pub struct Crate {
...
-    /// The name associated with this crate, according to the custom
-    /// build system being used.
+    /// The name associated with this crate.
+    ///
+    /// This is determined by the build system that produced
+    /// the `rust-project.json` in question. For instance, if buck were used,
+    /// the label might be something like `//ide/rust/rust-analyzer:rust-analyzer`.
+    ///
+    /// Do not attempt to parse the contents of this string; it is a build system-specific
+    /// identifier similar to [Crate::display_name].
```
lnicola, line 107: Nit: Crate::display_name will only show up as monospace in the docs if you wrap it in backquotes.

davidbarsky: thanks! fixed.

crates/project-model/src/project_json.rs
```diff
@@ -88,13 +91,93 @@ pub struct Crate {
...
+
+/// A template-like structure for describing runnables.
+///
+/// These are used for running and debugging binaries and tests without encoding
+/// build system-specific knowledge into rust-analyzer.
+///
+/// # Example
+///
+/// Below is an example of a test runnable. `{label}` and `{test_id}`
+/// are explained in [Runnable::args]'s documentation.
```
lnicola, line 130:
```suggestion
/// are explained in the [`Runnable::args`] documentation.
```

crates/project-model/src/project_json.rs
```diff
@@ -88,13 +91,93 @@ pub struct Crate {
...
+///     "kind": "testOne"
+/// }
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Runnable {
+    /// The program invoked by the runnable.
+    ///
+    /// For example, this might be `cargo`, `buck`, or `bazel`.
+    pub program: String,
+    /// The arguments passed to [Runnable::program].
```
lnicola, line 152: Since you've changed one..

```suggestion
    /// The arguments passed to [`Runnable::program`].
```

crates/project-model/src/project_json.rs
```diff
@@ -88,13 +91,93 @@ pub struct Crate {
...
+    /// the `rust-project.json` in question. For instance, if buck were used,
+    /// the label might be something like `//ide/rust/rust-analyzer:rust-analyzer`.
+    ///
+    /// Do not attempt to parse the contents of this string; it is a build system-specific
+    /// identifier similar to [`Crate::display_name`].
+    pub label: String,
+    /// Path corresponding to the build system-specific file defining the crate.
+    ///
+    /// It is roughly analogous to [`ManifestPath`], but none of the indexing/loading
+    /// operations can be directly applied/used with this file.
```
Veykril, line 112: I am not sure I can tell what indexing/loading operations are in this context

davidbarsky: ah, sorry: it's in reference to `ProjectManifest::from_manifest_file`. you can't run `cargo metadata` on a `rust-project.json` or Starlark-ified `TARGETS` file.

(Arguably, it should, but that's more of a discussion for https://github.com/rust-lang/rust-analyzer/pull/17246. I'll clarify this comment, sorry.)

gvozdvmozgu:
It seems that after this PR, the tests have stopped running.

```
 *  Executing task: cargo cargo test --package sqruff-lib --lib -- rules::layout::LT13::tests::test_fail_leading_whitespace_comment --exact --show-output 

error: no such command: `cargo`

        View all installed commands with `cargo --list`
        Find a package to install `cargo` with `cargo search cargo-cargo`
```

PR #16860, opened 2024-03-17
feat: Syntax highlighting improvements

Veykril:
Specifically
- Adds a new `constant` modifier, attached to keyword `const` (except for `*const ()` and `&raw const ()`), `const` items and `const` functions
- Adds (or rather reveals) `associated` modifier for associated items
- Fixes usage of the standard `static` modifier, now it acts like `associated` except being omitted for methods.
- Splits `SymbolKind::Function` into `Function` and `Method`. We already split other things like that (notable self param from params), so the split makes sense in general as a lot special cases around it anyways.


Veykril:
Last commit changes the ordering of the syntax tree to be no longer grammar order reliant (instead everything is lexicographically ordered now).

PR #16863, opened 2024-03-17
Bump dependencies and use in-tree `rustc_pattern_analysis`

Nadrieril:
One last `pattern_analysis` API change. I don't have any more planned! So we can now use the in-tree version when available.


Nadrieril:
I say that and I just had a new idea that requires API change x) Anyway, no _major_ change in sight

PR #16871, opened 2024-03-18
fix: Skip problematic cyclic dev-dependencies

Veykril:
Implements a workaround for https://github.com/rust-lang/rust-analyzer/issues/14167, notably it does not implement the ideas surfaced in the issue, but takes a simpler to implement approach (and one that is more consistent).

Effectively, all this does is discard dev-dependency edges that go from a workspace library target to another workspace library target. This means, using a dev-dependency to another workspace member inside unit tests will always fail to resolve for r-a now, (instead of being order dependent and causing problems elsewhere) while things will work out fine in integration tests, benches, examples etc. This effectively acknowledges package cycles to be okay, but crate graph cycles to be invalid:

Quoting https://github.com/rust-lang/rust-analyzer/issues/14167#issuecomment-1864145772
> Though, if you have “package cycle” in integration tests, you’d have “crate cycle” in unit test.

We disallow the latter here, while continuing to support the former

(What's missing is to supress diagnostics for such unit tests, though not doing so might be a good deterrent, making devs avoid the pattern altogether)

### DIFF
--- a/crates/project-model/src/workspace.rs
+++ b/crates/project-model/src/workspace.rs
@@ -1091,6 +1091,24 @@ fn cargo_to_crate_graph(
                     continue;
                 }
 
+                // If the dependency is a dev-dependency with both crates being member libraries of
+                // the workspace we discard the edge. The reason can be read up on in
+                // https://github.com/rust-lang/rust-analyzer/issues/14167
+                // but in short, such an edge usually causes some form of cycle in the crate graph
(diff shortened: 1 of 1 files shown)


Veykril:
Well, this obviously didn't work out as I expected. Not sure why I thought this would be a good solution, one does quite often crate dev-only dependencies in their own workspace (r-a is an example of that). Guess we'll need to be a bit smarter about this after all.

PR #16877, opened 2024-03-18
fix: Fix `impl Trait<Self>` causing stackoverflows

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/15646


Veykril:
Going to wait with merging this until https://github.com/rust-lang/rust-analyzer/pull/16805 lands since its easier to resolve conflicts in this PR then vice versa

PR #16879, opened 2024-03-18
Add fuel to match checking 

Nadrieril:
Exhaustiveness checking is NP-hard hence can take extremely long to check some specific matches. This PR makes ehxaustiveness bail after a set number of steps. I chose a bound that takes ~100ms on my machine, which should be more than enough for normal matches.

I'd like someone with less recent hardware to run the test to see if that limit is low enough for them. Also curious if the r-a team thinks this is a good ballpark or if we should go lower/higher. I don't have much data on how complex real-life matches get, but we can definitely go lower than `500 000` steps.

The second commit is a drive-by soundness fix which doesn't matter much today but will matter once `min_exhaustive_patterns` is stabilized.

Fixes https://github.com/rust-lang/rust-analyzer/issues/9528 cc @matklad

### DIFF
--- a/crates/hir-ty/src/diagnostics/expr.rs
+++ b/crates/hir-ty/src/diagnostics/expr.rs
@@ -11,7 +11,6 @@ use hir_def::{ItemContainerId, Lookup};
 use hir_expand::name;
 use itertools::Itertools;
 use rustc_hash::FxHashSet;
-use rustc_pattern_analysis::usefulness::{compute_match_usefulness, PlaceValidity};
 use syntax::{ast, AstNode};
 use tracing::debug;
 use triomphe::Arc;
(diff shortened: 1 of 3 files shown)


Veykril:
Running `time cargo test -p ide-diagnostics exponential_match --release` gives me ~344ms (debug is more like 2 seconds). I think that should be fine. Might be good to add a 

```rs
    if skip_slow_tests() {
        return;
    }
```

in front given usually the test runs in debug. With that it gets skipped by default when people run the test suite (while CI will still always run it).

Nadrieril:
I'll make the match even bigger actually, to be sure it would hang CI ^^

Veykril:
Thanks 👍 
@bors r+

PR #16889, opened 2024-03-19
internal: Enforce utf8 paths

Veykril:
Cargo already requires this, and I highly doubt r-a works with non-utf8 paths generally either. This just makes dealing with paths a lot easier.

### DIFF
--- a/crates/paths/src/lib.rs
+++ b/crates/paths/src/lib.rs
@@ -1,4 +1,4 @@
-//! Thin wrappers around `std::path`, distinguishing between absolute and
+//! Thin wrappers around `std::path`/`camino::path`, distinguishing between absolute and
 //! relative paths.
 
 #![warn(rust_2018_idioms, unused_lifetimes)]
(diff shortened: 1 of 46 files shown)


crates/paths/src/lib.rs
```diff
@@ -45,20 +65,30 @@ impl Borrow<AbsPath> for AbsPathBuf {
...
+    type Error = Utf8PathBuf;
+    fn try_from(path_buf: Utf8PathBuf) -> Result<AbsPathBuf, Utf8PathBuf> {
+        if !path_buf.is_absolute() {
+            return Err(path_buf);
+        }
+        Ok(AbsPathBuf(path_buf))
+    }
+}
+
 impl TryFrom<PathBuf> for AbsPathBuf {
```
matklad, line 78: It feels like this impl should now be deleted? Not it is `try` for two orthogonal reasons:

* either the path is not valid utf-8
* or the path is not absolute

Better let the caller to explictily handle both of the possiblieties and do something like

```
AbsPathBuf::try_from(Utf8PathBuf::try_from(PathBuf)?)
```

PR #16896, opened 2024-03-19
Improve text description of check command

Wilfred:
Previously, rust-analyzer assumed that the command was always `clippy check` when multiple commands were running. This isn't true when users have overridden the check command.


Veykril:
I already fixed this a few days ago in 😄  https://github.com/rust-lang/rust-analyzer/pull/16883

PR #16913, opened 2024-03-21
fix: Some file watching related vfs fixes

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/15554, additionally it seems that client side file watching was broken on windows this entire time, this PR switches `DidChangeWatchedFilesRegistrationOptions` to use relative glob patterns which do work on windows in VSCode.

### DIFF
--- a/crates/rust-analyzer/src/handlers/notification.rs
+++ b/crates/rust-analyzer/src/handlers/notification.rs
@@ -241,7 +241,7 @@ pub(crate) fn handle_did_change_watched_files(
     state: &mut GlobalState,
     params: DidChangeWatchedFilesParams,
 ) -> anyhow::Result<()> {
-    for change in params.changes {
+    for change in params.changes.iter().unique_by(|&it| &it.uri) {
         if let Ok(path) = from_proto::abs_path(&change.uri) {
             state.loader.handle.invalidate(path);
         }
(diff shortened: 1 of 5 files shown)


lnicola:
Huh, it works for me :stuck_out_tongue:.

Veykril:
Inbefore their file watcher works differently well between linux and windows

Veykril:
Okay so VSCode just does not send a `workspace/didChangeWatchedFiles` notification when a file gets created even if it matches the glob pattern ... which I guess makes sense if you snapshot all existing files when we tell it what to watch? But at the same time, when will it ever send a created notification then ...?

Veykril:
And no, this does not work for me whatsoever. Running `cargo add` yields no events from VSCode even 😕

lnicola:
WDYM?

```
$ touch src/foo.rs # =>

[Trace - 9:57:23 PM] Sending notification 'workspace/didChangeWatchedFiles'.
Params: {
    "changes": [
        {
            "uri": "file:///Projects/hello/src/foo.rs",
            "type": 1
        },
        {
            "uri": "file:///Projects/hello/src/foo.rs",
            "type": 1
        }
    ]
}
```

Veykril:
I don't get any such events from vscode :)

lnicola:
Sounds like file watching is completely broken on Windows?

Veykril:
Looks like it, tried it on my linux machine quickly and there touching does sent an event

Veykril:
Huh, switching over to using relative patterns works! That's weird

PR #16915, opened 2024-03-21
fix: Prevent stack overflow in recursive const types

6d7a:
In the evaluation of const values of recursive types certain declarations could cause an endless call-loop within the interpreter (hir-ty’s create_memory_map), which would lead to a stack overflow.
This commit adds a check that prevents values that contain an address in their value (such as TyKind::Ref) from being allocated at the address they contain.
The commit also adds a test for this edge case.

### DIFF
--- a/crates/hir-ty/src/consteval/tests.rs
+++ b/crates/hir-ty/src/consteval/tests.rs
@@ -2828,7 +2828,7 @@ fn unsized_local() {
 
 #[test]
 fn recursive_adt() {
-    check_answer(
+    check_fail(
         r#"
         //- minicore: coerce_unsized, index, slice
         pub enum TagTree {
@@ -2849,6 +2849,6 @@ fn recursive_adt() {
             TAG_TREE
         };
     "#,
-        |b, _| assert_eq!(b[0] % 8, 0),
+        |e| matches!(e, ConstEvalError::MirEvalError(MirEvalError::StackOverflow)),
     );
 }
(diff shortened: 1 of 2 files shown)


HKalbasi:
@6d7a Thanks for the PR! I didn't understand how your patch addresses the problem. If I understand correctly, the problem is that `TAG_TREE` and `VARIANT_TAG_TREE` will get same address which will break things (here, and probably in other places too) but I'm not sure how this PR prevents that and how well it generalize to other cases.

In addition to that bug which we need to fix, infinite constants might exist, so we might need to add a depth limit or something like that to the `create_memory_map` function.

6d7a:
@HKalbasi thank you for your review. My reasoning in case of the fix was to prevent that reference values point to the location that they themselves reside in. However I can see that this is more of a hotfix for my particular case instead of a sustainable solution. Too much wishful thinking on my part.
I have removed the fix and added a stack depth check by propagating the `Evaluator`'s stack depth limit along the call chain of `create_memory_map`. That prevents a terminal stack overflow, which makes rust-analyzer impossible to use in certain projects, and instead returns a soft internal error. I adapted the test accordingly.
I'd be interested in finding a way to better handle recursive const cases, though, but I guess that's for another PR.

HKalbasi:
Thanks!
@bors r+

> I'd be interested in finding a way to better handle recursive const cases, though, but I guess that's for another PR.

AFAIK there is no way to create infinite constants in stable Rust, so this is good for now.

Though we should find out why it is allocating the same address for different constants, as it can cause problems even without recursive types.

PR #16919, opened 2024-03-22
fix: handle self::super when lowering UseTree

### LINKED ISSUE #16800: Importing From `self::super` Leads To Unresolved Identifiers
This bug causes strange effects within Rust Analyzer. The basic gist is importing anything from `self::super` doesn't "resolve", and attempting to do anything semantic with any identifiers imported from there doesn't work. For example, hovering and control clicking these identifiers does nothing, and hovering over most variables created from let bindings made with expressions using these identifiers show a type of `{unknown}`. Type checking still appears to work, however I find this bug degrades the developer performance, and sometimes even [causes strange bogus errors](16801).

I am looking to implement a fix for this issue. Let me know if this is something the team wants to be fixed.

Minimum Reproducible Code
-------------------------------------------
Here is the minimum reproducible code along with some comments.
```rs
#![allow(dead_code)]

struct Thing;

impl Thing {
	fn do_it(self) -> usize {
		69420
	}
}

mod a {
	mod b {
		/// Self in Front of Supers
		/// -----------------------
		/// Works in rustc, causes this bug in rust-analyzer. You can tell this is
		/// causes a bug by triggering a hover hint over the `thing` variable; it
		/// reads `{unknown}`, same with the `result` variable.
		use self::super::super::Thing;

		/// Just Supers
		/// -----------
		/// Works in rust-analyzer and rustc, variables are resolved correctly.
		//use super::super::Thing;

		/// Self in Middle of Supers
		/// ------------------------
		/// Hard error in rust-analyzer and rustc.
		//use super::self::super::Thing;

		fn use_it() {
			let thing = Thing;
			let result = thing.do_it();
			println!("{}", result)
		}
	}
}

fn main() {}
```

Here's the output from `cargo check` on the code as is.
image

Here's what happens when you hover over `let thing`.
image

Version Information
-----------------------------
**rust-analyzer version**: 1.76.0 (07dca48 2024-02-04)

Related Issues
---------------------
#10770

roife:
fix #16800.

### DIFF
--- a/crates/hir-expand/src/mod_path.rs
+++ b/crates/hir-expand/src/mod_path.rs
@@ -225,6 +225,26 @@ fn convert_path(
     let mut segments = path.segments();
 
     let segment = &segments.next()?;
+    let handle_super_kw = &mut |init_deg| {
+        let mut deg = init_deg;
+        let mut next_segment = None;
+        for segment in segments.by_ref() {
+            match segment.kind()? {
+                ast::PathSegmentKind::SuperKw => deg += 1,
+                ast::PathSegmentKind::Name(name) => {
+                    next_segment = Some(name.as_name());
+                    break;
+                }
+                ast::PathSegmentKind::Type { .. }
+                | ast::PathSegmentKind::SelfTypeKw
+                | ast::PathSegmentKind::SelfKw
+                | ast::PathSegmentKind::CrateKw => return None,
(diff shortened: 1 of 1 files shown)


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16924, opened 2024-03-22
feat: Add `rust-analyzer.cargo.allTargets` to configure passing `--all-targets` to cargo invocations

### LINKED ISSUE #16859: Build script build is run with `--all-targets` even if `rust-analyzer.check.allTargets` is false
This is easily confirmed by checking out [hashbrown](https://github.com/rust-lang/hashbrown/) and trying to work on it with stable while also using
```
{
    "rust-analyzer.cargo.buildScripts.useRustcWrapper": false,
    "rust-analyzer.check.allTargets": false
}
```
It is understandable that `allTargets: false` is needed, since `cargo check --all-targets` fails on stable on hashbrown. (Though this default can be quite confusing when one doesn't know about this toggle.)

But with the configuration above I would then expect RA to work fine on hashbrown, and yet it still tries to build the benchmarks:

```
2024-03-17T10:08:52.660154Z ERROR rust_analyzer::main_loop: FetchBuildDataError:
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> benches/insert_unique_unchecked.rs:3:12
  |
3 | #![feature(test)]
  |            ^^^^


error: aborting due to 1 previous error


For more information about this error, try `rustc --explain E0554`.

error: could not compile `hashbrown` (bench "insert_unique_unchecked") due to 2 previous errors
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> benches/bench.rs:5:12
  |
5 | #![feature(test)]
  |            ^^^^


error: aborting due to 1 previous error


For more information about this error, try `rustc --explain E0554`.

error: could not compile `hashbrown` (bench "bench") due to 2 previous errors


```

**rust-analyzer version**: `rust-analyzer version: 0.3.1877-standalone`

poliorcetics:
Closes #16859

## Unresolved question:

Should this be a setting for build scripts only ? All the other `--all-targets` I found where already covered by `checkOnSave.allTargets`

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -71,6 +71,9 @@ config_data! {
         /// How many worker threads to handle priming caches. The default `0` means to pick automatically.
         cachePriming_numThreads: ParallelCachePrimingNumThreads = "0",
 
+        /// Pass `--all-targets` to cargo invocation. Overridden by `#rust-analyzer.check.allTargets#`
+        /// when the latter is set.
+        cargo_allTargets: bool           = "true",
         /// Automatically refresh project info via `cargo metadata` on
         /// `Cargo.toml` or `.cargo/config.toml` changes.
         cargo_autoreload: bool           = "true",
@@ -163,8 +166,8 @@ config_data! {
         /// Run the check command for diagnostics on save.
         checkOnSave | checkOnSave_enable: bool                         = "true",
 
-        /// Check all targets and tests (`--all-targets`).
-        check_allTargets | checkOnSave_allTargets: bool                  = "true",
+        /// Check all targets and tests (`--all-targets`). Overrides `#rust-analyzer.cargo.allTargets#`.
+        check_allTargets | checkOnSave_allTargets: Option<bool>          = "null",
         /// Cargo command to use for `cargo check`.
         check_command | checkOnSave_command: String                      = "\"check\"",
         /// Extra arguments for `cargo check`.
@@ -1273,6 +1276,7 @@ impl Config {
         let sysroot_query_metadata = self.data.cargo_sysrootQueryMetadata;
 
         CargoConfig {
+            all_targets: self.data.cargo_allTargets,
             features: match &self.data.cargo_features {
                 CargoFeaturesDef::All => CargoFeatures::All,
                 CargoFeaturesDef::Selected(features) => CargoFeatures::Selected {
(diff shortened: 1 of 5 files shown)


jakzale:
> Closes #16859
> 
> ## Unresolved question:
> Should this be a setting for build scripts only ? All the other `--all-targets` I found where already covered by `checkOnSave.allTargets`

When working with targets that
- are building their own `std`, and
- are being built with `restricted_std`,

passing `--all-targets` will break both `cargo build` and `cargo check`, as `cargo` will try to build the `test` crate, which is not marked with `restricted_std`.

See the PR referenced above.

Veykril:
> Should this be a setting for build scripts only ? All the other `--all-targets` I found where already covered by `checkOnSave.allTargets`

Yes, all other invocations should be check related so thats fine. What we should do is have the check setting inherit the cargo setting here if its unset. Thats how all the other ones work for this pairing

poliorcetics:
> Yes, all other invocations should be check related so thats fine. What we should do is have the check setting inherit the cargo setting here if its unset. Thats how all the other ones work for this pairing

So the `checkOnSave.allTargets` setting would change from `bool` to `Option<bool>` and become `None` by default ?

Veykril (approved):

Veykril:
Thanks!
@bors r+

editors/code/package.json
```diff
@@ -546,6 +546,11 @@
                     "minimum": 0,
                     "maximum": 255
                 },
+                "rust-analyzer.cargo.allTargets": {
+                    "markdownDescription": "Pass `--all-targets` to cargo invocation. Overridden by `#rust-analyzer.check.allTargets#`\nwhen the latter is set.",
```
RalfJung, line 550: Is it truly overridden? It seems to me like it is only partially overridden. Specifically, the build script / proc macro build will always use the value of `rust-analyzer.cargo.allTargets` even if `rust-analyzer.check.allTargets` is set.

If that's correct then the docs are wrong or at least misleading.

crates/rust-analyzer/src/config.rs
```diff
@@ -1273,6 +1276,7 @@ impl Config {
         let sysroot_query_metadata = self.data.cargo_sysrootQueryMetadata;
 
         CargoConfig {
+            all_targets: self.data.cargo_allTargets,
```
RalfJung, line 1279: If `check_allTargets` is intended to overwrite `cargo_allTargets`, then that would have to be done here as well.

editors/code/package.json
Veykril: Indeed, I misread this.

poliorcetics: I should have precised "for check calls" maybe

Veykril: I adjusted it in https://github.com/rust-lang/rust-analyzer/pull/16988 making it the same as the docs for `allFeatures`

PR #16925, opened 2024-03-22
Allow post-fix match to parse

RossSmyth:
I am not very confident in how I handled the float recovery logic, but the RAST looks like how I believe it should.

re: rust-lang/rust#121619 rust-lang/rust#121618


RossSmyth:
Well there doesn't seem to be too much excitement for this PR.

PR #16927, opened 2024-03-22
fix: Rename `func_like` to `FuncLike`

### LINKED ISSUE #16926: `FuncLike` proc-macro expansion no longer works due to improper rename
Commit https://github.com/rust-lang/rust-analyzer/commit/8e324e98a1db06b3be105c8e3e905665653233a1 renamed the `FuncLike` kind to `Bang` and added an alias for backward-compatibility called `func_like`, while it should be `FuncLike`. This currently causes the following error in my project when using `#[tokio::main]`:
```console
proc macro `main` not expanded: failed to read response: unknown variant `FuncLike`, expected one of `CustomDerive`, `Attr`, `func_like` at line 1 column 178
```
Minimal reproducible example (*really*):
```rs
// cargo add tokio -F full
#[tokio::main]
async fn main() {}
```
The following lines cause this:
https://github.com/rust-lang/rust-analyzer/blob/8e324e98a1db06b3be105c8e3e905665653233a1/crates/proc-macro-api/src/lib.rs#L39-L42
A simple fix should just be to rename it to `FuncLike`:
```diff
-    #[serde(rename(serialize = "func_like", deserialize = "func_like"))]
+    #[serde(rename(serialize = "FuncLike", deserialize = "FuncLike"))]
```
I'll try to PR a fix. **edit:** #16927

Sculas:
Should fix #16926. Please check the issue for more information.

### DIFF
--- a/crates/proc-macro-api/src/lib.rs
+++ b/crates/proc-macro-api/src/lib.rs
@@ -38,7 +38,7 @@ pub enum ProcMacroKind {
     Attr,
     // This used to be called FuncLike, so that's what the server expects currently.
     #[serde(alias = "bang")]
-    #[serde(rename(serialize = "func_like", deserialize = "func_like"))]
+    #[serde(rename(serialize = "FuncLike", deserialize = "FuncLike"))]
     Bang,
 }
 


Veykril:
Thanks!
@bors r+

PR #16933, opened 2024-03-24
Add static modifier to static variables

### LINKED ISSUE #16894: `static` variables do not have `static` modifier
It seems that after #16860 PR local variables declarations became indistinguishable from static variables declarations.

**rust-analyzer version**:  rust-analyzer version: 0.4.1887-standalone (b91697de8 2024-03-19)

**relevant settings**: 
Previously I used the following config, this allowed me to diversify between `static`, `static mut`, `const`, `let` and `let mut`
```json
        "[ThemeName]": {
         "enabled": true,
         "rules": {
             "variable": "#FFFFFF",
              "variable.mutable": {
                  "italic": true,
                  "foreground": "#FFFFFF"
              },
             "*.static": {
                 "bold": true,
                 "foreground": "#cc711d",
             },
             "*.static.mutable": {
                 "bold": true,
                 "foreground": "#cc711d",
             },
             "*.static.constant": {
                 "bold": true,
                 "foreground": "#F0F080",
             },
            }
        }
    }
````

(Can be triggered via `Developer : Inspect Editor Tokens and Scopes` with the cursor on a symbol name)
```rust
// semantic token type: variable
// modifiers: declaration
static FOO: AtomicUsize = AtomicUsize::new(0);

// semantic token type: variable
// modifiers: declaration constant
const BAR: usize = 0;

fn test() {
    // semantic token type: variable
    // modifiers: declaration constant
    let foo: usize = 0;
}
```

maxwase:
Fixes #16894

It seems that there is a name collision between `SymbolKind::Static` and `HlMod::Static` resulting in a wired looking test diff (`static mutable static unsafe`), that's why I decided to rename `SymbolKind::Static`. I don't think that any of these are exposed to a user, however, not fully sure, I was not able to break my configuration with either of these name changes.

Tested with the following config static colors configuration:

```json
        "enabled": true,
        "rules": {
            "namespace": "#80E0E0",
            "function": {
                "bold": true,
                "foreground": "#80F080",
            },
            "variable": "#FFFFFF",
            "variable.static": {
                "bold": true,
                "foreground": "#cc711d",
            },
            "variable.constant": {
                "bold": true,
                "foreground": "#F0F080",
            },
        }
    }
```

image

```rs
// semantic token type : modifiers

// variable : declaration constant
const A: u32 = 0;
// variable : declaration static
static B: u32 = 0;

fn foo() {
    // variable : declaration
    let _c = 0;
}
```

### DIFF
--- a/crates/ide/src/syntax_highlighting/highlight.rs
+++ b/crates/ide/src/syntax_highlighting/highlight.rs
@@ -493,7 +493,7 @@ pub(super) fn highlight_def(
                 h |= HlMod::Unsafe;
             }
 
-            h
+            h | HlMod::Static
         }
         Definition::SelfType(_) => Highlight::new(HlTag::Symbol(SymbolKind::Impl)),
         Definition::GenericParam(it) => match it {
--- a/crates/ide/src/syntax_highlighting/tags.rs
+++ b/crates/ide/src/syntax_highlighting/tags.rs
@@ -154,7 +154,7 @@ impl HlTag {
                 SymbolKind::Module => "module",
                 SymbolKind::SelfParam => "self_keyword",
                 SymbolKind::SelfType => "self_type_keyword",
-                SymbolKind::Static => "static",
+                SymbolKind::Static => "static_kind",
                 SymbolKind::Struct => "struct",
                 SymbolKind::ToolModule => "tool_module",
                 SymbolKind::Trait => "trait",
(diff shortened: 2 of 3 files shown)


Veykril:
The name collision is fine. I remove the use of the static modifier for static items because I assumed they are used for static members usually (think of static methods in languages like java). It felt somewhat wrong to me to use it for `static` (as in the item) but arguably its not clear what it should be used for as the LSP spec (given its a standard modifiers) doesn't specify it.

maxwase:
@Veykril I see your point, however I think it is probably ok to have it here, at least for the sake of diversification, without it there is no way to semantically distinguish those

Veykril:
I'm still not too convinced by this being used for static items 😅 I opened https://github.com/rust-lang/rust-analyzer/pull/17074 which will add it as an item type instead (and `const` as well) as that seems more correct to me

PR #16935, opened 2024-03-24
Handle panicking like rustc CTFE does

### LINKED ISSUE #16907: Hang when hovering over constant
**rust-analyzer version**:
- 0.3.1885-standalone (b6d1887bc 2024-03-17)
- 0.4.1889-standalone (5e276ae51 2024-03-20)
- 0.0.0 (8d74705b43 2024-03-21)

**rustc version**: rustc 1.79.0-nightly (a7e4de13c 2024-03-19)

**relevant settings**: N/A

```rust
pub enum AA {
    BB,
}

pub fn hang() -> AA {
    const CONST: AA = AA::BB;
    CONST
}
```

Hovering over either `CONST` starts the hang and rust-analyzer becomes unresponsive, consuming about ~200MiB RAM per second in this case, though in other cases it can be much faster.

The server cannot be killed through hovering on the `rust-analyzer` button on the lower bar or with `rust-analyzer: Stop server`, and must be killed through the command line. It is either non-responsive or doesn't kill the process.

I can reproduce this consistently with recent rust-analyzer client and server, but only after updating `rustc` from `nightly-2024-03-19` (`rustc 1.79.0-nightly (3c85e5624 2024-03-18)`) to `nightly-2024-03-20` (`rustc 1.79.0-nightly (a7e4de13c 2024-03-19)`) or later with rustup (override works too).

Nilstrieb:
Instead of using `core::fmt::format` to format panic messages, which may in turn panic too and cause recursive panics and other messy things, redirect `panic_fmt` to `const_panic_fmt` like CTFE, which in turn goes to `panic_display` and does the things normally. See the tests for the full call stack.

The tests don't work yet, I probably missed something in minicore.

fixes #16907 in my local testing, I also need to add a test for it

### DIFF
--- a/crates/hir-ty/src/mir/eval/shim.rs
+++ b/crates/hir-ty/src/mir/eval/shim.rs
@@ -13,7 +13,7 @@ use crate::mir::eval::{
     name, pad16, static_lifetime, Address, AdtId, Arc, BuiltinType, Evaluator, FunctionId,
     HasModule, HirDisplay, Interned, InternedClosure, Interner, Interval, IntervalAndTy,
     IntervalOrOwned, ItemContainerId, LangItem, Layout, Locals, Lookup, MirEvalError, MirSpan,
-    ModPath, Mutability, Result, Substitution, Ty, TyBuilder, TyExt,
+    Mutability, Result, Substitution, Ty, TyBuilder, TyExt,
 };
 
 mod simd;
@@ -158,6 +158,25 @@ impl Evaluator<'_> {
         Ok(false)
     }
 
+    pub(super) fn detect_and_redirect_special_function(
+        &mut self,
+        def: FunctionId,
+    ) -> Result<Option<FunctionId>> {
+        // `PanicFmt` is redirected to `ConstPanicFmt`
+        if let Some(LangItem::PanicFmt) = self.db.lang_attr(def.into()) {
+            let resolver =
+                self.db.crate_def_map(self.crate_id).crate_root().resolver(self.db.upcast());
+
+            let Some(hir_def::lang_item::LangItemTarget::Function(const_panic_fmt)) =
+                self.db.lang_item(resolver.krate(), LangItem::ConstPanicFmt)
+            else {
+                not_supported!("const_panic_fmt lang item not found or not a function");
+            };
+            return Ok(Some(const_panic_fmt));
+        }
+        Ok(None)
+    }
+
     /// Clone has special impls for tuples and function pointers
     fn exec_clone(
         &mut self,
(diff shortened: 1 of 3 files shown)


crates/hir-ty/src/mir/eval/shim.rs
```diff
@@ -309,43 +334,6 @@ impl Evaluator<'_> {
         let mut args = args.iter();
         match it {
             BeginPanic => Err(MirEvalError::Panic("<unknown-panic-payload>".to_owned())),
-            PanicFmt => {
```
HKalbasi, line 312: Don't we need to have something like this? How `const_panic_fmt` can return `MirEvalError::Panic` on its own?

Nilstrieb: https://github.com/rust-lang/rust/blob/6a92312a1eaf5d9cffe3bdf9aaecbfdf8e2e35d9/library/core/src/panicking.rs#L259
it calls `panic_display` which is in turn hooked again https://github.com/rust-lang/rust/blob/6a92312a1eaf5d9cffe3bdf9aaecbfdf8e2e35d9/library/core/src/panicking.rs#L190

crates/hir-ty/src/mir/eval/tests.rs
```diff
@@ -87,6 +88,42 @@ fn main() {
...
+fn panic_fmt() {
+    // panic!
+    // -> panic_2021 (builtin macro redirection)
+    //   -> #[lang = "panic_fmt"] core::panicking::panic_fmt (hooked by CTFE for redirection)
+    //   -> core::panicking::const_panic_fmt
+    //     -> #[rustc_const_panic_str] core::panicking::panic_display (hooked by CTFE for builtin panic)
+    //       -> Err(ConstEvalError::Panic)
+    check_pass(
+        r#"
+//- minicore: fmt, panic
```
Nilstrieb, line 101: somehow these tests don't work correctly yet, they currently fail MIR building (also, they shouldn't be `check_pass` but `check_panic` or whatever). I am currently [unable to debug the build failure](https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/print.20path.20of.20FunctionId), so help welcome

Nilstrieb:
Someone else who had this problem in their project (making it unusable) confirmed that this also fixed their problem. I can drop the tests from this PR and do them in a follow up PR (I promise I won't forget about it :3) to merge this ASAP.

crates/hir-ty/src/mir/eval/tests.rs
Nilstrieb: I dropped this test, will do it in a follow-up PR. I added a simpler test for the hover behavior, which should be enough end-to-end.

Nilstrieb:
I created #16938 to figure out the tests, so this PR should now work (and have a small hover test to ensure that end-to-end behavior doesn't regress). So this should be ready for merging now.

crates/hir-ty/src/mir/eval/shim.rs
HKalbasi: So we need to hook `panic_display` in r-a as well? What is the output of `rust-analyzer(debug-command) interpret function` on a function that panics? It should be `MirEvalError::Panic("panic message")` but I guess it is now `execution limit exceeded` or something like that as it will try to execute `panic_fmt` and `panic_display` over and over.

Nilstrieb: Hooking that function is what my `rustc_const_panic_str` change does.

HKalbasi: Ah I see. So we now lose the panic message? but let's track it in #16938 and merge this one.

PR #16938, opened 2024-03-24
Implement `BeginPanic` handling in const eval

Nilstrieb:
for #16935, needs some figuring out of how to write these tests correctly


Veykril:
I pushed a fix to the minicore fixture which changes the failure of one of the tests

Veykril:
There were some more issues in the minicore definition (it was written with signatures in mind only so some functions were just lacking bodies etc), now both tests are running into execution limits (probably executing `loop {}`s but I am not sure why, I don't know hwo the current formatting infra works wrt to const eval)

Veykril:
Found another issue, now we actually run into trying to interpret `BeginPanic` which I don't think is implemented? I don't know

Veykril:
> Found another issue, now we actually run into trying to interpret `BeginPanic` which I don't think is implemented? I don't know
> Panic with message:
> "<unknown-panic-payload>"
@HKalbasi is that the case? Do we need to implement the `BeginPanic` interpretation now or is  that error deliberate as to that we should never hit it? I am not quite sure on this...

HKalbasi:
Previously, `BeginPanic` hit only in panics with custom payload. Now that we hook one layer below `PanicFmt`, I think we should try to extract the panic message if the underlying panic payload type is `&str` or `String`.

Nilstrieb:
thanks for continuing this, i completely forgot about it :skull:

PR #16944, opened 2024-03-25
Revert to the crates.io version of rustc_pattern_analysis

Nadrieril:
The API hasn't fully settled yet, and there's an extra wrinkle with `IdxContainer` which blocked the [subtree update](https://github.com/rust-lang/rust/pull/122981). Let's just keep using the crates.io version for a bit longer.

r? @lnicola


lnicola (approved):

lnicola:
That was easy :joy:.

@bors r+

PR #16957, opened 2024-03-27
fix: check for client support of relative glob patterns before using them

### LINKED ISSUE #16955: RA sends relative glob patterns even when the LSP client does not allow it
Commit fe28e470cdf7d864fb69411e780b1e9791d6742a introduces relative glob patterns but it does not check whether the LSP client supports it.   Rust-analyzer should only use relative patterns if the client has a relativePatternSupport in its [DidChangeWatchedFilesClientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#didChangeWatchedFilesClientCapabilities).

It is debatable whether this is a serious regression in nightly and it's important to fix it before the next release.  For me, it is.

CC: @Veykril

poliorcetics:
Fixes #16955

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -1013,6 +1013,17 @@ impl Config {
         )
     }
 
+    pub fn did_change_watched_files_relative_pattern_support(&self) -> bool {
+        try_or_def!(
+            self.caps
+                .workspace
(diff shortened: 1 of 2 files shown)


lnicola (approved):

lnicola:
Thanks!

@bors r+

PR #16965, opened 2024-03-28
fix: use lldb when debugging with C++ extension on MacOS

roife:
See https://github.com/rust-lang/rust-analyzer/issues/16901#issuecomment-2024486941

This PR resolves the issue of being unable to debug using the C++ extension on macOS. By using special configurations for the `MIMode` on macOS, it enables the C++ extension to connect to lldb when debugging (without affecting other platforms).

### DIFF
--- a/editors/code/src/debug.ts
+++ b/editors/code/src/debug.ts
@@ -203,6 +203,10 @@ function getCCppDebugConfig(
         cwd: cargoWorkspace || runnable.args.workspaceRoot,
         sourceFileMap,
         env,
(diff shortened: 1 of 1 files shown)


roife:
cc @lnicola. I believe this PR resolves the issue of debugging with the C++ extension.

lnicola:
I take it this works without also setting `miDebuggerPath`?

roife:
By setting `MIMode`, VScode will automatically use `lldb` in the `PATH`.

PR #16967, opened 2024-03-28
fix: ADT hover considering only type or const len not lifetimes

### LINKED ISSUE #16963: Panic on hovering over type name with lifetime argument
When hovering over the name of a type with a lifetime parameter, a panic occurs and [this error](https://github.com/rust-lang/rust-analyzer/files/14781790/stderr.txt) is printed to stderr. VS Code shows a popup with a "Request textDocument/hover failed." message.

**rust-analyzer version**: 0.4.1899-standalone (4b33850c3 2024-03-26)

**rustc version**: rustc 1.79.0-nightly (47ecded35 2024-03-26)

Create a new empty project, then paste the following snippet:
```rust
fn main() {
    println!("Hello, world!");
}

struct Pedro<'a> {
    hola: &'a str
}
```
Now hover your mouse over the word `Pedro` and the panic will occur.

dfireBird:
I feel like test doesn't do much. Please suggest if I should improve it?

Fixes #16963

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -1418,7 +1418,8 @@ impl Adt {
     }
 
     pub fn layout(self, db: &dyn HirDatabase) -> Result<Layout, LayoutError> {
-        if db.generic_params(self.into()).iter().count() != 0 {
+        let generic_params = &db.generic_params(self.into());
+        if generic_params.iter().next().is_some() || generic_params.iter_lt().next().is_some() {
             return Err(LayoutError::HasPlaceholder);
         }
         let krate = self.krate(db).id;
--- a/crates/ide/src/hover/tests.rs
+++ b/crates/ide/src/hover/tests.rs
@@ -7856,3 +7856,25 @@ impl Iterator for S {
         "#]],
     );
 }
+
+#[test]
+fn hover_lifetime_regression_16963() {
+    check(
+        r#"
+struct Pedro$0<'a> {
+    hola: &'a str
+}
+"#,
+        expect![[r#"
+            *Pedro*
+
+            ```rust
+            test
+            ```
+
+            ```rust
+            struct Pedro<'a>
+            ```
+        "#]],
+    )
+}


crates/hir/src/lib.rs
```diff
@@ -1418,7 +1418,8 @@ impl Adt {
     }
 
     pub fn layout(self, db: &dyn HirDatabase) -> Result<Layout, LayoutError> {
-        if db.generic_params(self.into()).iter().count() != 0 {
+        let generic_params = &db.generic_params(self.into());
+        if generic_params.iter().count() != 0 || generic_params.iter_lt().count() != 0 {
```
lnicola, line 1422: Nit: `.iter().next().is_some()`

dfireBird: Yeah, that could be better, since count consumes the entire iterator.

lnicola (approved):

lnicola:
Thanks!

@bors r+

lnicola:
The test does its job (crash without the fix), it's fine.

crates/hir/src/lib.rs
```diff
@@ -1418,7 +1418,8 @@ impl Adt {
     }
 
     pub fn layout(self, db: &dyn HirDatabase) -> Result<Layout, LayoutError> {
-        if db.generic_params(self.into()).iter().count() != 0 {
+        let generic_params = &db.generic_params(self.into());
+        if generic_params.iter().next().is_some() || generic_params.iter_lt().next().is_some() {
```
Veykril, line 1422: Having a `GenericParam::is_empty` method would probably be useful here (or given there is more than just params in it, a `GenericParam::has_params`)

dfireBird: Should I open a PR with the proposed change of this and the proposed change in another PR?

Feel like a opening a lot of small PRs 😅.

Veykril: Putting both into one is fine

PR #16968, opened 2024-03-28
fix: silence mismatches involving unresolved projections

### LINKED ISSUE #16801: Functions Returning Unresolved Types Wrapped In A Tuple Wrapped In `impl Try` Causes Strange Errors When `?`ing & Destructuring
This bug requires a very specific condition to occur. First, you must have a function that returns an `Option` that contains a tuple that contains a [type that is not resolved](16800). Once you have that, you must use the function, apply the `?` operator on to it's result, and destructure the tuple. After doing that, Rust Analyzer will throw a bogus error on the let statement stating that "non-exhaustive pattern: `_` not covered".

This title is quite the mouth full, but this is a rather obscure bug. I'm looking to implement a fix for this issue, so let me know if the team wants this issue fixed; there's a chance this doesn't need to be fixed as this bug inherently requires another bug to trigger.

Minimum Reproducible Code
-------------------------------------------
Here is the minimum reproducible code.
```rs
#![allow(dead_code, unused_variables)]

struct Thing;

mod module {
	// This uses another bug to cause rust-analyzer to not resolve `Thing`.
	// More information at issue #16800.
	use self::super::Thing;

	fn create() -> Option<(i32, Thing)> {
		Some((69420, Thing))
	}

	fn consume() -> Option<()> {
		// Error triggered on next line.
		let (number, thing) = create()?;
		Some(())
	}
}

fn main() {}
```

Here's the output from `cargo check` on the code.
image

Here's what happens when you hover over the `let` statement.
image

Version Information
-----------------------------
**rust-analyzer version**: 1.76.0 (07dca48 2024-02-04)

roife:
fix #16801


crates/hir-ty/src/chalk_ext.rs
```diff
@@ -93,6 +94,11 @@ impl TyExt for Ty {
         self.data(Interner).flags.contains(TypeFlags::HAS_ERROR)
     }
 
+    fn is_projection(&self) -> bool {
+        matches!(self.kind(Interner), TyKind::Alias(AliasTy::Projection(_))) ||
```
lnicola, line 98: Nit: these can be merged.

crates/hir/src/lib.rs
```diff
@@ -1679,6 +1679,12 @@ impl DefWithBody {
             acc.extend(AnyDiagnostic::inference_diagnostic(db, self.into(), d, &source_map));
         }
         for (pat_or_expr, mismatch) in infer.type_mismatches() {
+            // Silence mismatches involving unresolved projections
+            let is_unknown_proj = |ty: &Ty| ty.contains_unknown() && ty.is_projection();
+            if is_unknown_proj(&mismatch.expected) || is_unknown_proj(&mismatch.actual) {
```
lnicola, line 1684: What about types like `S<<T as Tr>::U>`?

roife: Hmm... It's a little complicated. Perhaps we need to recursively check whether it satisfies `has_unknown && is_projection`?

lnicola: I think so, see https://github.com/rust-lang/rust-analyzer/blob/fc0d51ae655dd17ffe73fdf0b9a4819d262a8e1b/crates/hir-ty/src/infer.rs#L698-L708 and https://github.com/rust-lang/rust-analyzer/blob/fc0d51ae655dd17ffe73fdf0b9a4819d262a8e1b/crates/hir-ty/src/infer.rs#L1647 :grimacing:.

roife: I've read the code of `chalk`, and the logic for recursive checking seems somewhat similar to the part where flags are computed here: [https://github.com/rust-lang/chalk/blob/12eda8042b764134509ec2c12883786b9674cccd/chalk-ir/src/lib.rs#L688](https://github.com/rust-lang/chalk/blob/12eda8042b764134509ec2c12883786b9674cccd/chalk-ir/src/lib.rs#L688).

roife: Thanks!

crates/hir-ty/src/infer.rs
```diff
@@ -699,7 +699,7 @@ impl<'a> InferenceContext<'a> {
             mismatch.expected = table.resolve_completely(mismatch.expected.clone());
             mismatch.actual = table.resolve_completely(mismatch.actual.clone());
             chalk_ir::zip::Zip::zip_with(
-                &mut UnknownMismatch(self.db),
+                &mut UnknownMismatch(self.db, |ty| matches!(ty.kind(Interner), TyKind::Error)),
```
lnicola, line 702: Not sure, but it feels like we should be checking here for projections?

roife: I have tried doing this at first time, but it seems to cause some errors in MIR (as shown below). I suspect that MIR uses `type_mismatch` in `borrowck` to avoid certain paths, so I only removed the corresponding error in the outer layer. However, I haven't studied this part carefully yet 🙊. If necessary, I can further investigate this error in MIR to ensure it's not another potential bug.

```
thread 'Worker' panicked at crates/hir-ty/src/mir.rs:185:21:
internal error: entered unreachable code: Only tuple has tuple field
```

lnicola: No, you're probably right, MIR is not resilient to code that doesn't typecheck.

Veykril: We should probably add a flag the type check results whether there were errors or other problems and have mir rely on that for exiting out early. Since us stripping diagnostics could cause MIR lowering to try lowering an non-successfully typed body.

roife: Agree. I'll try to implement it.

roife: I added an `ignore_in_diagnostics` field to the `TypeMismatch` struct, enabling us to retain type-mismatches for MIR lowering while removing them from diagnostics. It's more elegant than the previous solution.

roife: Hmmm, I found that after adding the field, we need to modify some code about tests and statistics to filter out ignored diagnostics. (done)

crates/hir-ty/src/infer.rs
```diff
@@ -258,6 +258,7 @@ pub enum InferenceDiagnostic {
 pub struct TypeMismatch {
     pub expected: Ty,
     pub actual: Ty,
+    pub ignored_in_diagnostics: bool,
```
Veykril, line 261: Instead of this, let's just add a `has_errors` field to `InferenceResult` and set that to true if there are any type mismatches pre filtering, then change mir lowering to error out if that flag is true

crates/hir-ty/src/infer.rs
```diff
@@ -1721,7 +1746,7 @@ impl chalk_ir::zip::Zipper<Interner> for UnknownMismatch<'_> {
                 zip_substs(self, None, &fn_ptr_a.substitution.0, &fn_ptr_b.substitution.0)?
             }
             (TyKind::Error, TyKind::Error) => (),
-            (TyKind::Error, _) | (_, TyKind::Error) => return Err(chalk_ir::NoSolution),
+            _ if (self.1)(a) || (self.1)(b) => return Err(chalk_ir::NoSolution),
```
Veykril, line 1749:
```suggestion
            (TyKind::Error, _) | (_, TyKind::Error) | (TyKind::Alias(AliasTy::Projection(_)) | TyKind::AssociatedType(_, _), _) | (_, TyKind::Alias(AliasTy::Projection(_)) | TyKind::AssociatedType(_, _)) => return Err(chalk_ir::NoSolution),
```
No need for the closures and zipping twice, we can just inline things here. It's very unlikely that we want to reuse this zipper for other things, so keeping it simple is better here. (Also adjust the comment again)

roife: Oh, sorry, previously I chose to skip mismatches in diagnostics, so I reused `UnknownMismatch`, and I didn't change it after using `has_error` field.

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #16972, opened 2024-03-29
Make `cargo run` always available for binaries

joshka:
Previously, items for `cargo test` and `cargo check` would appear as in
the `Select Runnable` quick pick that appears when running
`rust-analyzer: Run`, but `run` would only appear as a runnable if a
`main`` function was selected in the editor. This change adds `cargo
run` as an always available runnable command for binary packages.

This makes it easier to develop cli / tui applications, as now users can
run application from anywhere in their codebase.

### DIFF
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -836,10 +836,13 @@ pub(crate) fn handle_runnables(
     let config = snap.config.runnables();
     match cargo_spec {
         Some(spec) => {
-            let all_targets = !snap.analysis.is_crate_no_std(spec.crate_id)?;
-            for cmd in ["check", "test"] {
+            for cmd in ["check", "run", "test"] {
+                if cmd == "run" && spec.target_kind != TargetKind::Bin {
+                    continue;
+                }
                 let mut cargo_args =
                     vec![cmd.to_owned(), "--package".to_owned(), spec.package.clone()];
+                let all_targets = cmd != "run" && !snap.analysis.is_crate_no_std(spec.crate_id)?;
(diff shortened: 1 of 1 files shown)


joshka:
I'm unsure where exactly to write tests for this functionality.

I manually tested this using `cargo xtask install --server` against both a binary crate and a library crate.

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -836,10 +836,13 @@ pub(crate) fn handle_runnables(
     let config = snap.config.runnables();
     match cargo_spec {
         Some(spec) => {
-            let all_targets = !snap.analysis.is_crate_no_std(spec.crate_id)?;
```
Veykril, line 839: Let's keep this out of the loop, as this work only needs to be done once

Veykril:
Sounds reasonable I think, though there might be one problem. If inside a main function, with this change we'll show the run action twice if I am not mistaken.

joshka:
> If inside a main function, with this change we'll show the run action twice if I am not mistaken

Yep - although only for a main function inside the main src folder, not in a main function in the examples folder.

I'm not sure this is particularly a problem though as there's a rationale reason that both should be supported. When my cursor is located in a main function I'd expect to see the current behavior which is the same for tests / examples - I want to run the current contextual item. I'd also want to generally have a muscle memory type ahead just work (e.g. `Cmd+R run ENTER` selects the run cargo command and runs it. Also, I think this will be set as the most recent selection automatically so that becomes `Cmd+R ENTER` most of the time. Removing it contextually breaks the muscle memory part of this.

image

PR #16979, opened 2024-03-30
pattern analysis: Use contiguous indices for enum variants

Nadrieril:
The main blocker to using the in-tree version of the `pattern_analysis` crate is that rustc requires enum indices to be contiguous because it uses `IndexVec`/`BitSet` for performance. Currently we swap these out for `FxHashMap`/`FxHashSet` when the `rustc` feature is off, but we can't do that if we use the in-tree crate.

This PR solves the problem by using contiguous indices on the r-a side too.

### DIFF
--- a/crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
+++ b/crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
@@ -3,7 +3,7 @@
 use std::fmt;
 use tracing::debug;
 
-use hir_def::{DefWithBodyId, EnumVariantId, HasModule, LocalFieldId, ModuleId, VariantId};
+use hir_def::{DefWithBodyId, EnumId, EnumVariantId, HasModule, LocalFieldId, ModuleId, VariantId};
 use rustc_hash::FxHashMap;
 use rustc_pattern_analysis::{
     constructor::{Constructor, ConstructorSet, VariantVisibility},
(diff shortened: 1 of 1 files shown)


crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
```diff
@@ -36,6 +36,32 @@ pub(crate) type WitnessPat<'p> = rustc_pattern_analysis::pat::WitnessPat<MatchCh
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub(crate) enum Void {}
 
+/// An index type for enum variants. This ranges from 0 to `variants.len()`, whereas `EnumVariantId`
+/// can take arbitrary large values (and hence mustn't be used with `IndexVec`/`BitSet`).
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub(crate) struct EnumVariantContiguousIndex(usize);
+
+impl EnumVariantContiguousIndex {
+    fn from_enumvariantid(
```
lnicola, line 45:
```suggestion
    fn from_enum_variant_id
```

?

crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
```diff
@@ -89,9 +115,18 @@ impl<'p> MatchCheckCtx<'p> {
...
+    fn variant_id_for_adt(
+        db: &'p dyn HirDatabase,
+        ctor: &Constructor<Self>,
+        adt: hir_def::AdtId,
+    ) -> Option<VariantId> {
         match ctor {
-            &Variant(id) => Some(id.into()),
+            Variant(id) => {
+                let hir_def::AdtId::EnumId(eid) = adt else {
+                    panic!("bad constructor {ctor:?} for adt {adt:?}")
```
lnicola, line 126: Maybe `stdx::never` and `return None` here so we don't panic?

Nadrieril: If we return `None` we panic only a tiny bit later because most of the calls of this function `unwrap()` the output

Nadrieril: Want me to change that anyway?

crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
```diff
@@ -36,6 +36,32 @@ pub(crate) type WitnessPat<'p> = rustc_pattern_analysis::pat::WitnessPat<MatchCh
...
+        target_evid: EnumVariantId,
+    ) -> Option<Self> {
+        // Find the index of this variant in the list of variants.
+        db.enum_data(eid)
+            .variants
+            .iter()
+            .map(|(evid, _name)| *evid)
+            .enumerate()
+            .find(|(_, evid)| *evid == target_evid)
+            .map(|(i, _)| EnumVariantContiguousIndex(i))
```
Veykril, line 57: This can be simplified to
```suggestion
target_evid.lookup(self.db.upcast()).index as usize
```

(we store the index in the item tree to prevent a linear lookup here)

crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
Veykril: panicking early seems better then

Veykril:
Thanks!
@bors delegate+

crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
Nadrieril: Oh nicee

PR #16981, opened 2024-03-31
Limit chalk by time instead of iteration count

stepantubanov:
Makes these cases terminate quickly instead of hanging:
- https://github.com/rust-lang/rust-analyzer/issues/7680
- https://github.com/rust-lang/rust-analyzer/issues/9813

### DIFF
--- a/crates/hir-ty/src/traits.rs
+++ b/crates/hir-ty/src/traits.rs
@@ -1,6 +1,9 @@
 //! Trait solving using Chalk.
 
-use std::env::var;
+use std::{
+    env::var,
+    time::{Duration, Instant},
+};
 
 use chalk_ir::{fold::TypeFoldable, DebruijnIndex, GoalData};
 use chalk_recursive::Cache;
@@ -22,7 +25,7 @@ use crate::{
 };
 
 /// This controls how much 'time' we give the Chalk solver before giving up.
-const CHALK_SOLVER_FUEL: i32 = 1000;
+const CHALK_TIMEOUT: Duration = Duration::from_secs(1);
 
 #[derive(Debug, Copy, Clone)]
 pub(crate) struct ChalkContext<'a> {
@@ -144,16 +147,16 @@ fn solve(
     tracing::debug!("solve goal: {:?}", goal);
     let mut solver = create_chalk_solver();
 
-    let fuel = std::cell::Cell::new(CHALK_SOLVER_FUEL);
-
+    let start = Instant::now();
     let should_continue = || {
         db.unwind_if_cancelled();
-        let remaining = fuel.get();
-        fuel.set(remaining - 1);
-        if remaining == 0 {
-            tracing::debug!("fuel exhausted");
+
+        let elapsed = start.elapsed();
+        if elapsed > CHALK_TIMEOUT {
+            tracing::warn!("chalk solver timed out ({elapsed:.1?})");
+            return false;
         }
-        remaining > 0
+        true
     };
 
     let mut solve = || {


lnicola:
I don't understand how this works. Do 1000 steps take longer than one second?

stepantubanov:
> I don't understand how this works. Do 1000 steps take longer than one second?

In these cases yes, it gets exponentially slower with each step. Does one second sound like a reasonable limit?

flodiebold:
This would make trait solving non-deterministic though. I'm not sure that's a good idea.

Veykril:
> This would make trait solving non-deterministic though. I'm not sure that's a good idea.

I am actually curious whether this would cause problems? I guess what could happen is that we a goal stays unsolved due a timeout even if it could get calculated after a change? Though in that case it feels like the query should've been invalidated anyways meaning that shouldn't happen.

Also note that the new trait solver is likely to have timed fuel as well instead of counter as we have it right now (learned that this week from a discussion with lcnr)

Veykril:
Reading up on https://github.com/rust-lang/rust-analyzer/issues/4315 (an issue we had due to nondeterminism), I don't think this would be a problem here, as the trait solve query is not LRU cached (which I believe is the main issue wrt to nondeterminism).

matklad:
> Also note that the new trait solver is likely to have timed fuel as well instead of counter as we have it right now (learned that this week from a discussion with lcnr)

This is probably the most consequential thing in this issue. Is there any write-up for why it is the case? It seems that deterministic fuel should be fairly easy to add. In the limit, one can measure CPU instructions or some such, but there's bound to be some sort of primitive "step" which can be counted.

And I'd say that, over the long run, deterministic fuel is much more attractive. While it is true that with the current strategy of never forgetting the queries time based limiting should work, it closes over the possibility to adjust that strategy later. 

Salsa as an abstract API does not specify when or why it recomputes the queries. And we can imagine wanting to experiment with this down the line. What if we do LRU auto-magically for everything? When we implement on disk persistence, what if some queries turn out to be faster to recompute than to deserialise from disk? What if, when running concurrently, salsa would _race_ two identical queries rather than block?

So, irrespective of the current implications of the specific change, I'd be very much in favor of preserving valuable deterministic properties. 

----

There's a separate issue that non-determinism is pain in the back in day-to-day development. If the user submits an issue where rust-analyzer _sometimes_ colors the code red, is it due to inherent timing non-determinism, or do we have some missing vfs update somewhere? If you have built-in non-determinism, then any such issue becomes doubly mysterious.

Veykril:
>  Is there any write-up for why it is the case?

I don't remember the why, @lcnr can you briefly re-iterate why you'd prefer a time based fuel for the new trait solver opposed to a step / unit of work count?

lcnr:
> In these cases yes, it gets exponentially slower with each step. Does one second sound like a reasonable limit?

This is the reason why I'd expect a time based fuel to work better. The number of steps taken by the trait solver is a bad estimate for its runtime. While we could spend some effort getting a better estimate (considering the *size* of goals/returned constraints as well), I expect that having a fuel which is both prevents hangs while correctly checking all desired goals is pretty much impossible. We have the same issue when trying to prevent hangs in rustc itself, where I've given up on that goal for now.

My perspective from RustNation - and my experience using r-a by default while writing tests for the solver -  was that hangs in r-a are quite a lot worse than hangs in rustc as r-a runs by default and in the background.

nikomatsakis:
> 🤔 I wonder if fuel should be a generalized salsa feature (cc @nikomatsakis)? Kill a query if it is too deep in the query graph or if its dependency set is too wide? Sort of conservative approximation for “cycle detection”.

Interesting thought. Wouldn't help here because chalk isn't salsa-integrated, although that's long been an ambition of my mine (though the current new trait solver design didn't go that route, but rather continued to have special case dependency tracking).

Veykril:
Closing this as it seems undesirable to have. Would be good to check if we could add another dimension of a limit here (like goal size or similar?), though that will require changes to chalk.

PR #16996, opened 2024-04-02
internal: Lower outlive goals, respect them in display impls

Veykril:


crates/hir-ty/src/lower.rs
```diff
@@ -1325,7 +1349,21 @@ impl<'a> TyLoweringContext<'a> {
...
+                            DebruijnIndex::INNERMOST,
+                            bound_var.index,
+                        ))
+                        .intern(Interner),
+                        None => it,
+                    },
+                    None => static_lifetime(),
+                },
+            })
+            .intern(Interner)
```
Veykril, line 1366: This is kind of ugly, not sure if there is a better way (while also avoiding allocations, or iterating twice). We lower all bounds in a `with_shifted_in(DebruijnIndex::ONE` which isn't correct for the lifetime of the trait object,

PR #17000, opened 2024-04-02
fix: set the right postfix snippets competion source range

mrnossiom:
Hi :wave:,

Changed the completion item `source_range` to match the replaced text. Though in VS Code it may not be disturbing because the snippet is previewed in a box, but in Helix editor, it's previewed by applying the main text edit.

Before :
image

After :
image

Thanks


Veykril:
Thanks!
@bors r+

pascalkuthe:
thanks for working on this! I just wanted to say that this doesn't just fix the preview (which is just a visual thing) but also multicursor completions. M

Additional text edits can't be mapped to multiple cursor (since they aren't associated with the cursor as per spec and you con't want to repeat imports for each cursor for example).

I see that RA does this as a workaround for LSP only allowing a single edit to reach completion item. I think that generally makes sense for associated edits like imports but I can't think of a practical case where multiple edits would be associated with the cursor. I suspect the RA completion edit could be brought closer to the lsp completion edit and avoid impedance mismatches like this bug

PR #17003, opened 2024-04-03
Fix off-by-one error converting to LSP UTF8 offsets with multi-byte char

krobelus:
On this file,

```rust
fn main() {
    let 된장 = 1;
}
```

when using `"positionEncodings":["utf-16"]` I get an "unused variable" diagnostic on the variable
name (codepoint offset range `8..10`). So far so good.

When using `positionEncodings":["utf-8"]`, I expect to get the equivalent range in bytes (LSP:
"Character offsets count UTF-8 code units (e.g bytes)."), which is `8..14`, because both
characters are 3 bytes in UTF-8.  However I actually get `10..14`.

Looks like this is because we accidentally treat a 1-based index as an offset value: when
converting from our internal char-indices to LSP byte offsets, we look at one character to many.
This causes wrong results if the extra character is a multi-byte one, such as when computing
the start coordinate of 된장.

Fix that by actually passing an offset. While at it, fix the variable name of the line number,
which is not an offset (yet).

Originally reported at https://github.com/kakoune-lsp/kakoune-lsp/issues/740

### DIFF
--- a/crates/rust-analyzer/src/diagnostics/to_proto.rs
+++ b/crates/rust-analyzer/src/diagnostics/to_proto.rs
@@ -68,8 +68,13 @@ fn location(
     let range = {
         let position_encoding = snap.config.position_encoding();
         lsp_types::Range::new(
-            position(&position_encoding, span, span.line_start, span.column_start),
-            position(&position_encoding, span, span.line_end, span.column_end),
+            position(
+                &position_encoding,
+                span,
+                span.line_start,
+                span.column_start.saturating_sub(1),
+            ),
+            position(&position_encoding, span, span.line_end, span.column_end.saturating_sub(1)),
         )
     };
     lsp_types::Location::new(uri, range)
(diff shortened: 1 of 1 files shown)


Veykril:
I believe the reasoning is that we need to do less work for utf8, given rust natively handles utf8 strings.

krobelus:
Right, LSP doesn't specify (what a surprise) whether the client preference or server preference should win.
So the best workaround is for the client to send only `["utf16"]` in that case.
Fortunately this is irrelevant usually.

Veykril:
Thanks!
@bors r+

BenjaminBrienen:
> Right, LSP doesn't specify (what a surprise) whether the client preference or server preference should win.
> So the best workaround is for the client to send only `["utf16"]` in that case.
> Fortunately this is irrelevant usually.

Isn't this the exact kind of "why" that would be useful to have as a code comment? Also, that unit of behavior for selecting an encoding could definitely be extracted into its own method. It's very self contained and solves a single problem.

PR #17004, opened 2024-04-03
Try caching macro calls more aggressively in Semantics

Veykril:


Wilfred:
@Veykril dumb question: why is this cached outside of salsa? Do you have a good rule of thumb for deciding which state should be in salsa versus in the struct directly like this example?

Veykril:
Most of the state here depends on syntax trees, those have no business being inside salsa because they are heavy memory wise and very unstable.

PR #17019, opened 2024-04-05
fix: VFS should not confuse paths with source roots that have the same prefix

Wilfred:
Previously, the VFS would assign paths to the source root that had the longest string prefix match. This would break when we had source roots in subdirectories:

```
/foo
/foo/bar
```

Given a file `/foo/bar_baz.rs`, we would attribute it to the `/foo/bar` source root, which is wrong.

As a result, we would attribute paths to the wrong crate when a crate was in a subdirectory of another one. This is more common in larger monorepos, but could occur in any Rust project.

Fix this in the VFS, and add a test.

### DIFF
--- a/crates/vfs/src/file_set.rs
+++ b/crates/vfs/src/file_set.rs
@@ -132,6 +132,10 @@ impl FileSetConfig {
     ///
     /// `scratch_space` is used as a buffer and will be entirely replaced.
     fn classify(&self, path: &VfsPath, scratch_space: &mut Vec<u8>) -> usize {
+        // `path` is a file, but r-a only cares about the containing directory. We don't
+        // want `/foo/bar_baz.rs` to be attributed to source root directory `/foo/bar`.
+        let path = path.parent().unwrap_or_else(|| path.clone());
+
         scratch_space.clear();
         path.encode(scratch_space);
         let automaton = PrefixOf::new(scratch_space.as_slice());
--- a/crates/vfs/src/file_set/tests.rs
+++ b/crates/vfs/src/file_set/tests.rs
@@ -40,3 +40,26 @@ fn name_prefix() {
     let partition = file_set.partition(&vfs).into_iter().map(|it| it.len()).collect::<Vec<_>>();
     assert_eq!(partition, vec![1, 1, 0]);
 }
+
+/// Ensure that we don't consider `/foo/bar_baz.rs` to be in the
+/// `/foo/bar/` root.
+#[test]
+fn name_prefix_partially_matches() {
+    let mut file_set = FileSetConfig::builder();
+    file_set.add_file_set(vec![VfsPath::new_virtual_path("/foo".into())]);
+    file_set.add_file_set(vec![VfsPath::new_virtual_path("/foo/bar".into())]);
+    let file_set = file_set.build();
+
+    let mut vfs = Vfs::default();
+
+    // These two are both in /foo.
+    vfs.set_file_contents(VfsPath::new_virtual_path("/foo/lib.rs".into()), Some(Vec::new()));
+    vfs.set_file_contents(VfsPath::new_virtual_path("/foo/bar_baz.rs".into()), Some(Vec::new()));
+
+    // Only this file is in /foo/bar.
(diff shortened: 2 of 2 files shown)


crates/vfs/src/file_set.rs
```diff
@@ -170,6 +170,13 @@ impl FileSetConfigBuilder {
                 for p in paths {
                     let mut buf = Vec::new();
                     p.encode(&mut buf);
+
+                    // A root is a directory, so append a trailing
+                    // separateor to the string used in prefix
```
davidbarsky, line 175:
```suggestion
                    // separator to the string used in prefix
```

crates/vfs/src/file_set.rs
```diff
@@ -170,6 +170,13 @@ impl FileSetConfigBuilder {
...
+                    // checks. This ensures that we don't think that
```
davidbarsky, line 176:
```suggestion
                    // checks. This ensures that r-a doesn't think that
```

crates/vfs/src/file_set.rs
```diff
@@ -170,6 +170,12 @@ impl FileSetConfigBuilder {
                 for p in paths {
                     let mut buf = Vec::new();
                     p.encode(&mut buf);
+
+                    // A root is a directory, so append a trailing separator to the string
+                    // used in prefix checks. This ensures that r-a doesn't think that
+                    // `/foo/bar.rs` lives in the `/foo/b/` root.
+                    buf.extend(std::path::MAIN_SEPARATOR.to_string().as_bytes());
```
Veykril, line 177:
```suggestion
                    buf.extend(std::path::MAIN_SEPARATOR_STR.as_bytes());
```

Veykril:
Thanks!
@bors r+

Veykril:
🤔 why is this failing on merge only, judging from the commit parents this is a recent branch so there shouldn't be any semantic conflicts I think?

ShoyuVanilla:
> 🤔 why is this failing on merge only, judging from the commit parents this is a recent branch so there shouldn't be any semantic conflicts I think?

According to CI log, the following tests fail on Windows

```
 failures:
    tests::ancestor_can_be_parent
    tests::ancestor_can_be_parent_2
    tests::basic_child_parent
    tests::basic_child_parent_with_unrelated_parents_sib
    tests::child_binds_ancestor_if_parent_nonlocal
    tests::non_locals_are_skipped
```
and it seems that CI runs Windows tests on merges, but not on PRs, so it fails on merge only.

https://github.com/rust-lang/rust-analyzer/blob/7a8374c162c64c17e865b98aad282d16b16e96d6/.github/workflows/ci.yaml#L93-L95

Wilfred:
Aha, there is a `.parent()` method, so I've updated the PR to use that instead. Let me know what you think.

crates/vfs/src/vfs_path.rs
```diff
@@ -76,7 +76,7 @@ impl VfsPath {
     }
 
     /// Returns `true` if `other` is a prefix of `self`.
-    pub fn starts_with(&self, other: &VfsPath) -> bool {
+    pub fn starts_with(&self, other: &VfsPath) -> bool { // ?
```
davidbarsky, line 79: is this question mark intentional?

ShoyuVanilla:
> huh, that seems unfortunate. Maybe PRs should also run Windows tests?

Though I don't know about the r-a CI policy but in my naive opinion, running tests for each major platforms would be good for cases like filesystem or environment variables level changes like in this issue and running it wouldn't slow down CI much because each platform CI is done in different instances.

Wilfred:
@Veykril I think this PR is good to merge now, it's using `.parent()` which should do the right thing on Windows too.

Veykril:
Thanks!
@bors r+

bill-myers:
Is this enough? Seems to me this will still result in /foo/bar_baz/quux.rs being attributed to /foo/bar rather than /foo.

The correct solution is of course to add a final slash to prefixes, and add a final slash to the string to search if matching the prefix itself is needed.

PR #17021, opened 2024-04-06
Support hovering limits for adts

### LINKED ISSUE #17009: hover a struct can't show its whole definition,but hover a enum can
截图 2024-04-04 11-17-11
截图 2024-04-04 11-16-31

roife:
Fix #17009

1. Currently, r-a supports limiting the number of struct fields displayed when hovering. This PR extends it to support enum variants and union fields. Since the display of these three (ADTs) is similar, this PR extends 'hover_show_structFields' to 'hover_show_adtFieldsOrVariants'.
2. This PR also resolved the problem that the layout of ADT was not restricted by display limitations when hovering on the Self type.
3. Additionally, this PR changes the default value of display limitations to `10` (instead of the original `null`), which helps users discover this feature.

### DIFF
--- a/crates/hir/src/display.rs
+++ b/crates/hir/src/display.rs
@@ -188,7 +188,7 @@ impl HirDisplay for Struct {
             StructKind::Record => {
                 let has_where_clause = write_where_clause(def_id, f)?;
                 if let Some(limit) = f.entity_limit {
-                    display_fields(&self.fields(f.db), has_where_clause, limit, f)?;
+                    display_fields(&self.fields(f.db), has_where_clause, limit, false, f)?;
                 }
             }
             StructKind::Unit => _ = write_where_clause(def_id, f)?,
@@ -225,7 +225,7 @@ impl HirDisplay for Union {
 
         let has_where_clause = write_where_clause(def_id, f)?;
         if let Some(limit) = f.entity_limit {
-            display_fields(&self.fields(f.db), has_where_clause, limit, f)?;
+            display_fields(&self.fields(f.db), has_where_clause, limit, false, f)?;
         }
         Ok(())
     }
@@ -235,10 +235,12 @@ fn display_fields(
     fields: &[Field],
     has_where_clause: bool,
     limit: usize,
+    in_line: bool,
     f: &mut HirFormatter<'_>,
 ) -> Result<(), HirDisplayError> {
     let count = fields.len().min(limit);
-    f.write_char(if !has_where_clause { ' ' } else { '\n' })?;
+    let (indent, separator) = if in_line { ("", ' ') } else { ("    ", '\n') };
+    f.write_char(if !has_where_clause { ' ' } else { separator })?;
     if count == 0 {
         if fields.is_empty() {
             f.write_str("{}")?;
@@ -246,15 +248,19 @@ fn display_fields(
             f.write_str("{ /* … */ }")?;
         }
     } else {
-        f.write_str("{\n")?;
+        f.write_char('{')?;
+        f.write_char(separator)?;
         for field in &fields[..count] {
-            f.write_str("    ")?;
+            f.write_str(indent)?;
             field.hir_fmt(f)?;
-            f.write_str(",\n")?;
+            f.write_char(',')?;
+            f.write_char(separator)?;
         }
 
         if fields.len() > count {
-            f.write_str("    /* … */\n")?;
+            f.write_str(indent)?;
+            f.write_str("/* … */")?;
+            f.write_char(separator)?;
         }
         f.write_str("}")?;
     }
@@ -345,21 +351,10 @@ impl HirDisplay for Variant {
                 }
                 f.write_char(')')?;
             }
-            VariantData::Record(fields) => {
-                f.write_str(" {")?;
-                let mut first = true;
-                for (_, field) in fields.iter() {
-                    if first {
-                        first = false;
-                        f.write_char(' ')?;
-                    } else {
-                        f.write_str(", ")?;
-                    }
-                    // Enum variant fields must be pub.
-                    write!(f, "{}: ", field.name.display(f.db.upcast()))?;
-                    field.type_ref.hir_fmt(f)?;
+            VariantData::Record(_) => {
+                if let Some(limit) = f.entity_limit {
+                    display_fields(&self.fields(f.db), false, limit, true, f)?;
                 }
-                f.write_str(" }")?;
             }
         }
         Ok(())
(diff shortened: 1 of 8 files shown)


Young-Flash (approved):
Thanks!

small nit: IMO default "10" is a little big, 10 fields exceeding the hover windows

image

roife:
How about changing the default limitation it to 5?

l1nxy:
I'm wondering if the change in configuration should be documented on https://github.com/rust-lang/rust-analyzer/issues/4604.

l1nxy:
I'm wondering if the change in configuration should be documented on https://github.com/rust-lang/rust-analyzer/issues/4604.

lnicola:
> I'm wondering if the change in configuration should be documented on https://github.com/rust-lang/rust-analyzer/issues/4604.

That's mainly for protocol changes (extensions), but we do have a compatibility mechanism for renamed settings.

crates/hir/src/display.rs
```diff
@@ -251,22 +222,44 @@ impl HirDisplay for Union {
...
-        if !fields.is_empty() {
-            f.write_char(if !has_where_clause { ' ' } else { '\n' })?;
-            f.write_str("{\n")?;
-            for field in self.fields(f.db) {
-                f.write_str("    ")?;
-                field.hir_fmt(f)?;
-                f.write_str(",\n")?;
-            }
-            f.write_str("}")?;
+fn display_fields_or_variants<T: HirDisplay>(
```
Veykril, line 234: I think we should split this into two, one for fields one  for variants. Reason being for an enum like
```rs
enum Enum {
    Variant {},
    Variant2 { field: Field },
    Variant3(),
    Variant4(Field),
    Variant5,
}
```
hovering the enum should yield
```rs
enum Enum {
    Variant {},
    Variant2 { .. }, // or the field if the limit is enabled and non-zero
    Variant3(),
    Variant4(,.), , // or the field if the limit is enabled and non-zero
    Variant5,
}
```
That is hovering an enum should only show its shape, not its vairants fields. To see those the person can hover the variant instead.

crates/ide/src/hover/tests.rs
```diff
@@ -51,13 +51,17 @@ fn check(ra_fixture: &str, expect: Expect) {
 }
 
 #[track_caller]
-fn check_hover_struct_limit(count: usize, ra_fixture: &str, expect: Expect) {
+fn check_hover_adt_fields_or_variants_limit(
+    count: Option<usize>,
```
Veykril, line 55: A neat trick for tests is to take a `impl Into<Option<usize>>` (might need to be `impl Into<Option<u32>>` for fallback reasons). That way you can just use `None` or `5` at call sites instead of `Some(5)`.

crates/ide/src/hover/tests.rs
```diff
@@ -1431,15 +1624,18 @@ impl Thing {
...
                 ```
             "#]],
     );
-    check(
+    check_hover_adt_fields_or_variants_limit(
+        None,
         r#"
 struct Thing { x: u32 }
 impl Thing {
-    fn new() -> Self$0 { Self { x: 0 } }
```
Veykril, line 1442: why was this touched?

crates/ide/src/hover/tests.rs
```diff
@@ -1456,9 +1652,9 @@ impl Thing {
     );
     check(
         r#"
-enum Thing { A }
+struct Thing { x: u32 }
 impl Thing {
-    pub fn new() -> Self$0 { Thing::A }
```
Veykril, line 1461: why was this touched?

crates/ide/src/hover/tests.rs
```diff
@@ -968,6 +977,190 @@ fn hover_record_struct_limit() {
...
+            ```rust
+            // size = 12 (0xC), align = 4
+            struct Foo { /* … */ }
+            ```
+        "#]],
+    )
+}
+
+#[test]
+fn hover_enum_limit() {
```
Veykril, line 1019: Would be good to add the snippet I posted earlier as a test case as well then

crates/ide/src/hover/tests.rs
roife: I didn't change it, just added a check before it for coverage (so the original `check` was moved below, see the check after it). 
But probably the diff algorithm of GitHub doesn't think so, so it incorrectly identifies this line as a modification rather than an insertion.. 🫤

crates/hir/src/display.rs
roife: > Variant2 { .. }, // or the field if the limit is enabled and non-zero

Here I use a single config to limit the num of fields or variants of ADTs, should we separate it into two different ones?

Veykril: AH wait I made a mistake, hovering enums should ***not*** render the variants fields.

Splitting the two into two configs sounds reasonable nevertheless.

roife: So, for hovering on enums, we can display it as 🤔

```rust
enum Enum {
    Variant {},
    Variant2 { /* .. */ },
    Variant3(),
    Variant4( /* .. */ ),
    Variant5,
}
```

crates/ide/src/hover/render.rs
```diff
@@ -410,8 +410,20 @@ pub(super) fn definition(
         Definition::Trait(trait_) => {
             trait_.display_limited(db, config.max_trait_assoc_items_count).to_string()
         }
-        Definition::Adt(Adt::Struct(struct_)) => {
-            struct_.display_limited(db, config.max_struct_field_count).to_string()
+        Definition::Adt(adt @ (Adt::Struct(_) | Adt::Union(_))) => {
```
Veykril, line 413: We should do this branch for `Definition::Variant` as well

crates/rust-analyzer/src/config.rs
```diff
@@ -312,8 +312,10 @@ config_data! {
         /// How to render the size information in a memory layout hover.
         hover_memoryLayout_size: Option<MemoryLayoutHoverRenderKindDef> = Some(MemoryLayoutHoverRenderKindDef::Both),
 
-        /// How many fields of a struct to display when hovering a struct.
-        hover_show_structFields: Option<usize> = None,
+        /// How many variants of an enum to display when hovering on. Show none if empty.
+        hover_show_enumVariants: Option<usize> = Some(5),
+        /// How many fields of a struct or union to display when hovering on. Show none if empty.
+        hover_show_structOrUnionFields: Option<usize> = Some(5),
```
Veykril, line 318:
```suggestion
        hover_show_fields: Option<usize> = Some(5),
```
since this should also work for enum variants, so just fields should be good enough

Veykril (commented):
Last review round

Veykril (approved):

Veykril (approved):

PR #17024, opened 2024-04-06
fix: handle escaped chars in doc comments

### LINKED ISSUE #16980: `#[doc ="..."]` comments no longer allow escaped newlines
In rust-analyzer `rust-analyzer version: 0.3.1885-standalone (b6d1887bc 2024-03-17)` in vscode, when writing a doc comment

```
#[doc = "\n\nThis is a \"quote\".\n\nFoo"]
fn main() {}
```
hovering over `main()` shows the doc comment as expected, with newlines. However, in rust-analyzer `rust-analyzer version: 0.3.1896-standalone`, while the quote characters display correctly, literal `\n` (backslash n) tokens appear in the tooltip.

roife:
fix #16980.

For `ast::LiteralKind::String`, store the original string value.

### DIFF
--- a/crates/hir-expand/src/attrs.rs
+++ b/crates/hir-expand/src/attrs.rs
@@ -1,5 +1,5 @@
 //! A higher level attributes based on TokenTree, with also some shortcuts.
-use std::{fmt, ops};
+use std::{borrow::Cow, fmt, ops};
 
 use base_db::CrateId;
 use cfg::CfgExpr;
@@ -297,6 +297,20 @@ impl Attr {
         }
     }
 
+    pub fn string_value_unescape(&self) -> Option<Cow<'_, str>> {
+        match self.input.as_deref()? {
+            AttrInput::Literal(it) => match it.text.strip_prefix('r') {
+                Some(it) => {
+                    it.trim_matches('#').strip_prefix('"')?.strip_suffix('"').map(Cow::Borrowed)
+                }
+                None => {
+                    it.text.strip_prefix('"')?.strip_suffix('"').and_then(unescape).map(Cow::Owned)
+                }
+            },
+            _ => None,
+        }
+    }
+
     /// #[path(ident)]
     pub fn single_ident_value(&self) -> Option<&tt::Ident> {
         match self.input.as_deref()? {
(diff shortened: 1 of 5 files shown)


roife:
cc @Veykril See https://github.com/rust-lang/rust-analyzer/issues/16980#issuecomment-2028614924

I'm not sure if this is the best approach to handle escaped chars, but if we store text instead of not storing the string value, then we'll need to unescape it later on, which makes it complex. Hope you can provide some suggestions.

Veykril:
This won't work, both our mbe parsing and the proc-macro logic expect `tt::Literal`s to actually contain literal values, which in the case of strings would be a proper string literal value. I'd prefer if we could store the text verbatim so the span fits (ideally this will make syntax injection work properly for doc attributes, I haven't gotten to that yet though).

I don't think having to unescape at call sites should be too bad though, we can have the `string_value` function do just that (or have a separate one if they are callees that don't care about it).

roife:
If we unescape characters in `string_value`, then we have to return an `Option<String>` (instead of the original `Option<&str>`), which will affect many functions in the calling chain.

Veykril:
That is, in https://github.com/rust-lang/rust-analyzer/blob/4135696ea7d5b09f30f79676a3795f91610a025d/crates/ide-db/src/documentation.rs there are some calls that care about unescaping and I presume (haven't checked), https://github.com/rust-lang/rust-analyzer/blob/4135696ea7d5b09f30f79676a3795f91610a025d/crates/hir-def/src/nameres/collector.rs#L1920 this one also cares about unescaping. The rest does not seem to care so we can keep the escaped versions there.

crates/hir-expand/src/attrs.rs
```diff
@@ -346,6 +360,39 @@ impl Attr {
...
+                }
+                _ => return None,
+            }
+        } else {
+            res.push(c);
+        }
+    }
+    Some(res)
+}
+
```
Veykril, line 395: shouldve said that before but we have an unescape function we can use from the rustc_lexer, see what we do here https://github.com/rust-lang/rust-analyzer/blob/062e1b9b81542680afde1aa47fa5316ac5133461/crates/syntax/src/ast/token_ext.rs#L215-L250

crates/hir-expand/Cargo.toml
```diff
@@ -33,6 +33,7 @@ mbe.workspace = true
 limit.workspace = true
 span.workspace = true
 parser.workspace = true
+ra-ap-rustc_lexer.workspace = true
```
Veykril, line 36: Let's re-export the unescape stuff from the parser or syntax crate (reason being that the `ra-ap-rustc` crates are special and require extra `extern crate` prelude annotations, a re-export saves us from that here

Veykril:
One more annoyance, but otherwise lgtm! Thanks!
@bors delegate+

PR #17037, opened 2024-04-09
internal: improve `TokenSet` implementation and add reserved keywords

### LINKED ISSUE #16858: Reserved words in paths are not prepended with 'r#'
**rust-analyzer version**: rust-analyzer version: 0.3.1877-standalone (574e23ec5 2024-03-09)

When sources are auto-generated, e.g. from protobuf specification, it may contain modules/types/… named with reserved words, like `priv`, `mod`. It is possible to use them from Rust code using `r#priv`, `r#mod`. The problem is that when you use auto-import, imported path don't augment such path components with `r#`.

davidsemakula:
The current `TokenSet` type represents "A bit-set of `SyntaxKind`s" as a newtype `u128`.
Internally, the flag for each `SyntaxKind` variant in the bit-set is set as the n-th LSB (least significant bit) via a bit-wise left shift operation, where n is the discriminant. 

Edit: This is problematic because there's currently ~121 token `SyntaxKind`s, so adding new token kinds for missing reserved keywords increases the number of token `SyntaxKind`s above 128, thus making this ["mask"](https://github.com/rust-lang/rust-analyzer/blob/7a8374c162c64c17e865b98aad282d16b16e96d6/crates/parser/src/token_set.rs#L31-L33) operation overflow.
~~This is problematic because there's currently 266 SyntaxKinds, so this ["mask"](https://github.com/rust-lang/rust-analyzer/blob/7a8374c162c64c17e865b98aad282d16b16e96d6/crates/parser/src/token_set.rs#L31-L33) operation silently overflows in release mode.~~
~~This leads to a single flag/bit in the bit-set being shared by multiple `SyntaxKind`s~~.

This PR:
- Changes the wrapped type for `TokenSet` from `u128` to `[u64; 3]` ~~`[u*; N]` (currently `[u16; 17]`) where `u*` can be any desirable unsigned integer type and `N` is the minimum array length needed to represent all token `SyntaxKind`s without any collisions~~.
- Edit: Add assertion that `TokenSet`s only include token `SyntaxKind`s
- Edit: Add ~7 missing [reserved keywords](https://doc.rust-lang.org/stable/reference/keywords.html#reserved-keywords)
- ~~Moves the definition of the `TokenSet` type to grammar codegen in xtask, so that `N` is adjusted automatically (depending on the chosen `u*` "base" type) when new `SyntaxKind`s are added~~.
- ~~Updates the `token_set_works_for_tokens` unit test to include the `__LAST` `SyntaxKind` as a way of catching overflows in tests.~~

~~Currently `u16` is arbitrarily chosen as the `u*` "base" type mostly because it strikes a good balance (IMO) between unused bits and readability of the generated `TokenSet` code (especially the [`union` method](https://github.com/rust-lang/rust-analyzer/blob/7a8374c162c64c17e865b98aad282d16b16e96d6/crates/parser/src/token_set.rs#L26-L28)), but I'm open to other suggestions or a better methodology for choosing `u*` type.~~

~~I considered using a third-party crate for the bit-set, but a direct implementation seems simple enough without adding any new dependencies. I'm not strongly opposed to using a third-party crate though, if that's preferred.~~

~~Finally, I haven't had the chance to review issues, to figure out if there are any parser issues caused by collisions due the current implementation that may be fixed by this PR - I just stumbled upon the issue while adding "new" keywords to solve #16858~~

Edit: fixes #16858

### DIFF
--- a/crates/parser/src/token_set.rs
+++ b/crates/parser/src/token_set.rs
@@ -4,32 +4,46 @@ use crate::SyntaxKind;
 
 /// A bit-set of `SyntaxKind`s
 #[derive(Clone, Copy)]
-pub(crate) struct TokenSet(u128);
+pub(crate) struct TokenSet([u64; 3]);
+
+const LAST_TOKEN_KIND_DISCRIMINANT: usize = SyntaxKind::SHEBANG as usize;
 
 impl TokenSet {
-    pub(crate) const EMPTY: TokenSet = TokenSet(0);
+    pub(crate) const EMPTY: TokenSet = TokenSet([0; 3]);
 
     pub(crate) const fn new(kinds: &[SyntaxKind]) -> TokenSet {
-        let mut res = 0u128;
+        let mut res = [0; 3];
         let mut i = 0;
         while i < kinds.len() {
-            res |= mask(kinds[i]);
+            let kind = kinds[i];
+            debug_assert!(
+                kind as usize <= LAST_TOKEN_KIND_DISCRIMINANT,
+                "Expected a token `SyntaxKind`"
+            );
+            let idx = kind as usize / 64;
+            res[idx] |= mask(kind);
             i += 1;
         }
         TokenSet(res)
     }
 
     pub(crate) const fn union(self, other: TokenSet) -> TokenSet {
-        TokenSet(self.0 | other.0)
+        TokenSet([self.0[0] | other.0[0], self.0[1] | other.0[1], self.0[2] | other.0[2]])
     }
 
     pub(crate) const fn contains(&self, kind: SyntaxKind) -> bool {
-        self.0 & mask(kind) != 0
+        debug_assert!(
+            kind as usize <= LAST_TOKEN_KIND_DISCRIMINANT,
+            "Expected a token `SyntaxKind`"
+        );
+        let idx = kind as usize / 64;
+        self.0[idx] & mask(kind) != 0
     }
 }
 
-const fn mask(kind: SyntaxKind) -> u128 {
-    1u128 << (kind as usize)
+const fn mask(kind: SyntaxKind) -> u64 {
+    debug_assert!(kind as usize <= LAST_TOKEN_KIND_DISCRIMINANT, "Expected a token `SyntaxKind`");
+    1 << (kind as usize % 64)
 }
 
 #[test]
--- a/crates/parser/src/token_set.rs
+++ b/crates/parser/src/token_set.rs
@@ -6,6 +6,9 @@ use crate::SyntaxKind;
 #[derive(Clone, Copy)]
 pub(crate) struct TokenSet([u64; 3]);
 
+/// `TokenSet`s should only include token `SyntaxKind`s, so the discriminant of any passed/included
+/// `SyntaxKind` must *not* be greater than that of the last token `SyntaxKind`.
+/// See #17037.
 const LAST_TOKEN_KIND_DISCRIMINANT: usize = SyntaxKind::SHEBANG as usize;
 
 impl TokenSet {
@@ -15,13 +18,13 @@ impl TokenSet {
         let mut res = [0; 3];
         let mut i = 0;
         while i < kinds.len() {
-            let kind = kinds[i];
+            let discriminant = kinds[i] as usize;
             debug_assert!(
-                kind as usize <= LAST_TOKEN_KIND_DISCRIMINANT,
+                discriminant <= LAST_TOKEN_KIND_DISCRIMINANT,
                 "Expected a token `SyntaxKind`"
             );
-            let idx = kind as usize / 64;
-            res[idx] |= mask(kind);
+            let idx = discriminant / 64;
+            res[idx] |= 1 << (discriminant % 64);
             i += 1;
         }
         TokenSet(res)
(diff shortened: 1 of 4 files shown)


lnicola:
How come this never fails in practice? Do we never construct a `TokenSet` with something higher than `RET_TYPE`?

Veykril:
`TokenSet`s only ever contain token syntax kinds which populate the low integer variants only

lnicola:
I mean, we're cutting it pretty close with `LIFETIME_IDENT` (119), but I can't find any higher than that. Of course, `T![]` doesn't make it any easier to check. But we could at least add an assertion for it.

Veykril:
We should certainly have a `(debug_)assert` here

lnicola:
But can we fit them all in 0 to 64 though? :sweat_smile:.

Veykril:
I mean we are using a u128 rigt now, that suffices no?

lnicola:
Yeah, but if we could fit them in 64 bits, we'd save 8 full bytes of memory for a token set!

davidsemakula:
> `TokenSet`s only ever contain token syntax kinds which populate the low integer variants only, so in practice things never collide here.

Gotcha, that explains why it hasn't been problematic before 🙂


> So while this is a nice idea here, we don't really need to do this (should be fine to just debug assert that we only ever pass a token syntax kind to these)

As mentioned earlier, I ran into the issue while solving #16858

I attempted it to add about ~7 [reserved keywords](https://doc.rust-lang.org/stable/reference/keywords.html#reserved-keywords) (e.g. `abstract`, `final` e.t.c) which triggered the overflow(s).

I could solve the issue another way, but I thought adding the reserved keywords [here](https://github.com/rust-lang/rust-analyzer/blob/a5feb4f05f09adca661c869b1bf2324898cbaa43/xtask/src/codegen/grammar/ast_src.rs#L67) was the best solution.

Veykril:
I see, adding the keywords as kinds seems like the proper approach here. Can we change the token set to just contain a `[u64; 3]` then? I don't think splitting it into smaller primitives makes much sense

Veykril:
Codegenerating this seems a bit overkill, let's just keep the manual impl, I expect us to only ever need to bump the backing limit once (when you add the new contextual keywords).

davidsemakula:
Done, also added fix for #16858

crates/parser/src/token_set.rs
```diff
@@ -4,32 +4,46 @@ use crate::SyntaxKind;
 
 /// A bit-set of `SyntaxKind`s
 #[derive(Clone, Copy)]
-pub(crate) struct TokenSet(u128);
+pub(crate) struct TokenSet([u64; 3]);
+
+const LAST_TOKEN_KIND_DISCRIMINANT: usize = SyntaxKind::SHEBANG as usize;
```
lnicola, line 9: This might be worth a comment.

crates/parser/src/token_set.rs
```diff
@@ -4,32 +4,46 @@ use crate::SyntaxKind;
...
+    pub(crate) const EMPTY: TokenSet = TokenSet([0; 3]);
 
     pub(crate) const fn new(kinds: &[SyntaxKind]) -> TokenSet {
-        let mut res = 0u128;
+        let mut res = [0; 3];
         let mut i = 0;
         while i < kinds.len() {
-            res |= mask(kinds[i]);
+            let kind = kinds[i];
+            debug_assert!(
```
lnicola, line 19: This is also checked in `mask`.

crates/parser/src/token_set.rs
```diff
@@ -4,32 +4,46 @@ use crate::SyntaxKind;
...
     }
 
     pub(crate) const fn union(self, other: TokenSet) -> TokenSet {
-        TokenSet(self.0 | other.0)
+        TokenSet([self.0[0] | other.0[0], self.0[1] | other.0[1], self.0[2] | other.0[2]])
     }
 
     pub(crate) const fn contains(&self, kind: SyntaxKind) -> bool {
-        self.0 & mask(kind) != 0
+        debug_assert!(
```
lnicola, line 35: Ditto.

crates/parser/src/token_set.rs
```diff
@@ -4,32 +4,46 @@ use crate::SyntaxKind;
...
+            "Expected a token `SyntaxKind`"
+        );
+        let idx = kind as usize / 64;
+        self.0[idx] & mask(kind) != 0
     }
 }
 
-const fn mask(kind: SyntaxKind) -> u128 {
-    1u128 << (kind as usize)
+const fn mask(kind: SyntaxKind) -> u64 {
```
lnicola, line 44: I don't feel too strongly, but we might as well inline this, since half of the logic happens in the caller.

crates/parser/src/token_set.rs
```diff
@@ -4,32 +4,46 @@ use crate::SyntaxKind;
...
-    pub(crate) const EMPTY: TokenSet = TokenSet(0);
+    pub(crate) const EMPTY: TokenSet = TokenSet([0; 3]);
 
     pub(crate) const fn new(kinds: &[SyntaxKind]) -> TokenSet {
-        let mut res = 0u128;
+        let mut res = [0; 3];
         let mut i = 0;
         while i < kinds.len() {
-            res |= mask(kinds[i]);
+            let kind = kinds[i];
```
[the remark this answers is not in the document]
davidsemakula: Nope, no `for` loops in `const fn`

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17051, opened 2024-04-11
fix: support auto-closing for triple backticks

### LINKED ISSUE #16051: Auto-closing backticks does not work well combined with triple backtick code blocks.
**rust-analyzer version**: `0.3.1756-standalone`

**rustc version**: `rustc 1.75.0-beta.5`

When writing doc comments with examples, attempting to create a code block by pressing backtick three times ends up creating 4 backticks, which can get quite annoying when writing a lot of them.

Furthermore, unlike normal string literals, extra backticks are not error highlighted of course, which means documentation
can end up with random double backticks scattered throughout the documentation.

Ideally I would line an option to at least disable the autoclosing backticks without having to disable all autoclosing pairs, because any more sophisticated solution will still run into the second issue.

roife:
It might fix #16051, see https://github.com/rust-lang/rust-analyzer/issues/16051#issuecomment-2042606030

### DIFF
--- a/editors/code/language-configuration.json
+++ b/editors/code/language-configuration.json
@@ -19,7 +19,8 @@
         { "open": "(", "close": ")" },
         { "open": "\"", "close": "\"", "notIn": ["string"] },
         { "open": "/*", "close": " */", "notIn": ["string"] },
-        { "open": "`", "close": "`", "notIn": ["string"] }
+        { "open": "`", "close": "`", "notIn": ["string"] },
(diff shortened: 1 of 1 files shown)


roife:
If auto-closing for backticks only works in comments, that would be even better.

Perhaps we can submit a related API (something like `onlyIn`) to VS Code. 🤔

Veykril:
Sounds reasonable! Thanks!
@bors r+

PR #17054, opened 2024-04-12
Revert "fix: set the right postfix snippets competion source range"

### LINKED ISSUE #17036: Postfix snippets no longer work in vscode
**rust-analyzer version**: 0.4.1909-standalone

**editor or extension**: VSCode with extension v0.4.1909 (pre-release)

Postfix snippets stopped working after I've updated vscode extension. 

After downgrading a few times I found that the last version that has postfix snippets working is v0.4.1908. Upgrading it to v0.4.1909 breaks them.

v0.4.1908
image

v0.4.1909
image

roife:
This reverts commit c5686c89413b3c6eead7b39637381ee353f8c05c.

Fix #17036. See https://github.com/rust-lang/rust-analyzer/issues/17036#issuecomment-2050185641


roife:
Perhaps we should consider adding an `assert` statement for it. 👀

PR #17055, opened 2024-04-12
fix: Replace Just the variable name in Unused Variable Diagnostic Fix

### LINKED ISSUE #17053: Bug Report: Rust Analyzer Panic when destructuring Array with Tail Binding
When attempting to destructure an array and bind the tail using the pattern `[x, y @ ..] = arr`, Rust Analyzer encounters a panic and crashes with an assertion failure message. The panic specifically occurs with the message `thread 'Worker' panicked at crates\ide-db\src\label.rs:33:9: assertion failed: label.starts_with(char::is_uppercase) && !label.ends_with('.')`.

**rust-analyzer version**: 0.3.1916-standalone

**rustc version**: 1.77.2

**editor or extension**: VSCode: rust-analyzer v0.3.1916 on Windows 10

**code snippet to reproduce (literally just this)**: 
```rust
fn main() {
    let arr = [1, 2, 3, 4, 5];
    let [x, y @ ..] = arr;
}

```

Presumable, the problem here is that the tail binding ends in a `.` ? Is this a known issue?

wyatt-herkamp:
Changes Unused Variable diagnostic to just look at the variable name, not the entire syntax range. 

Also added a test for an unused variable in an array destructure. 

Closes #17053

### DIFF
--- a/crates/ide-diagnostics/src/handlers/unused_variables.rs
+++ b/crates/ide-diagnostics/src/handlers/unused_variables.rs
@@ -30,10 +30,10 @@ pub(crate) fn unused_variables(
         .primary_source(ctx.sema.db)
         .name()
         .map(|v| v.syntax().original_file_range_rooted(ctx.sema.db))
-        .filter(|it| Some(it.file_id) == ast.file_id.file_id())?;
-    if !diagnostic_range.range.contains_range(name_range.range) {
-        return None;
-    }
+        .filter(|it| {
+            Some(it.file_id) == ast.file_id.file_id()
+                && diagnostic_range.range.contains_range(it.range)
+        });
     let var_name = d.local.name(ctx.sema.db);
     Some(
         Diagnostic::new_with_syntax_node_ptr(
@@ -42,13 +42,9 @@ pub(crate) fn unused_variables(
             "unused variable",
             ast,
         )
-        .with_fixes(fixes(
-            ctx.sema.db,
-            var_name,
-            name_range.range,
-            diagnostic_range,
-            ast.file_id.is_macro(),
-        ))
+        .with_fixes(name_range.and_then(|it| {
+            fixes(ctx.sema.db, var_name, it.range, diagnostic_range, ast.file_id.is_macro())
+        }))
         .experimental(),
     )
 }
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/unused_variables.rs
```diff
@@ -20,32 +21,41 @@ pub(crate) fn unused_variables(
...
         return None;
     }
-    let diagnostic_range = ctx.sema.diagnostics_display_range(ast);
-    let var_name = d.local.primary_source(ctx.sema.db).syntax().to_string();
+    let file_id = ctx.sema.diagnostics_display_range(ast).file_id;
+    let (var_name, text_range) = d
+        .local
+        .primary_source(ctx.sema.db)
+        .name()
+        .map(|v| (v.value.text().to_owned(), v.value.syntax().text_range()))?;
```
Veykril, line 29: The text range here might be in a macro expansion instead of a real file. Should be fine to keep the `diagnostic_range` as before

wyatt-herkamp:
The changes in #16877 fixes the panic. However, it would replace the entire syntax for the test. Not just the variable name.

Veykril:
Oh, so that's where I put my version of that fix (I completely forgot I had that in that PR).

crates/ide-diagnostics/src/handlers/unused_variables.rs
```diff
@@ -23,6 +24,13 @@ pub(crate) fn unused_variables(
         return None;
     }
     let diagnostic_range = ctx.sema.diagnostics_display_range(ast);
+    // The range for the Actual Name. We don't want to replace the entire declarition. Using the diagnostic range causes issues within in Array Destructuring.
+    let name_range =
+        d.local.primary_source(ctx.sema.db).name().map(|v| v.syntax().value.text_range())?;
+    // Make sure we are within the diagnostic range for the variable
+    if !diagnostic_range.range.contains_range(name_range) {
```
Veykril, line 31: This check is still wrong, `diagnostic_range` is from a real file where as `name_range` could still originate from a macro expansion.

```rust
let name_range = d.local.primary_source(ctx.sema.db).name().map(|v| v.syntax().clone()).as_ref().original_file_range_rooted(self.db.upcast()).filter(|it| it.file_id == ast.file_id);
```
should give the right thing, and if it's `None` (can happen in theory via proc-macros/includes I believe), we should just not offer the quickfix

Veykril: That is, ideally this function shouldnt ever return `None`, its a temporary fix for another issue. As returning `None` means the diagnostic disappears.

Veykril:
Thanks!
@bors r+

PR #17058, opened 2024-04-13
feat: TOML based config for rust-analyzer

alibektas:
> Important
> 
> We don't promise _**any**_ stability with this feature yet, any configs exposed may be removed again, the grouping may change etc.

# TOML Based Config for RA 

This PR ( addresses #13529 and this is a follow-up PR on #16639 ) makes rust-analyzer configurable by configuration files called `rust-analyzer.toml`. Files **must** be named `rust-analyzer.toml`. There is not a strict rule regarding where the files should be placed, but it is recommended to put them near a file that triggers server to start (i.e., `Cargo.{toml,lock}`, `rust-project.json`).

## Configuration Types

Previous configuration keys are now split into three different classes.

1. Client keys: These keys only make sense when set by the client (e.g., by setting them in `settings.json` in VSCode). They are but a small portion of this list. One such example is `rust_analyzer.files_watcher`, based on which either the client or the server will be responsible for watching for changes made to project files.
2. Global keys: These keys apply to the entire workspace and can only be set on the very top layers of the hierarchy. The next section gives instructions on which layers these are.
3. Local keys: Keys that can be changed for each crate if desired.

### How Am I Supposed To Know If A Config Is Gl/Loc/Cl ?

#17101

## Configuration Hierarchy

There are 5 levels in the configuration hierarchy. When a key is searched for, it is searched in a bottom-up depth-first fashion.

### Default Configuration

**Scope**: Global, Local, and Client

This is a hard-coded set of configurations. When a configuration key could not be found, then its default value applies.

### User configuration

**Scope**: Global, Local

If you want your configurations to apply to **every** project you have, you can do so by setting them in your `$CONFIG_DIR/rust-analyzer/rust-analyzer.toml` file, where `$CONFIG_DIR` is :

| Platform | Value                                 | Example                                  |
| ------- | ------------------------------------- | ---------------------------------------- |
| Linux   | `$XDG_CONFIG_HOME` or `$HOME`/.config | /home/alice/.config                      |
| macOS   | `$HOME`/Library/Application Support   | /Users/Alice/Library/Application Support |
| Windows | `{FOLDERID_RoamingAppData}`           | C:\Users\Alice\AppData\Roaming           |

### Client configuration

**Scope**: Global, Local, and Client

Previously, the only way to configure rust-analyzer was to configure it from the settings of the Client you are using. This level corresponds to that. 

> With this PR, you don't need to port anything to benefit from new features. You can continue to use your old settings as they are.

### Workspace Root Configuration

**Scope**: Global, Local

Rust-analyzer already used the path of the workspace you opened in your Client. We used this information to create a configuration file that won't affect your other projects and define global level configurations at the same time. 

### Local Configuration

**Scope**: Local

You can also configure rust-analyzer on a crate level. Although it is not an error to define global ( or client ) level keys in such files, they won't be taken into consideration by the server. Defined local keys will affect the crate in which they are defined and crate's descendants. Internally, a Rust project is split into what we call `SourceRoot`s. This, although with exceptions, is equal to splitting a project into crates.

> You may choose to have more than one `rust-analyzer.toml` files within a `SourceRoot`, but among them, the one closer to the project root will be


crates/rust-analyzer/src/config.rs
```diff
@@ -540,81 +466,181 @@ config_data! {
...
+        /// Some clients like vs-code issue new searches on result filtering and don't require all results to be returned in the initial search.
+        /// Other clients requires all results upfront and might require a higher limit.
+        workspace_symbol_search_limit: usize = 128,
+        /// Workspace symbol search scope.
+        workspace_symbol_search_scope: WorkspaceSymbolSearchScopeDef = WorkspaceSymbolSearchScopeDef::Workspace,
+    }
+}
+
+config_data! {
+    /// Local configurations can be overridden for every crate by placing a `rust-analyzer.toml` on crate root.
```
davidbarsky, line 505: Does a configuration need to be _right next _ a `Cargo.toml`, or...?

```suggestion
    /// Configuration can be overridden for every crate by placing a `rust-analyzer.toml` at thecrate root.
```

crates/rust-analyzer/src/config.rs
```diff
@@ -540,81 +466,181 @@ config_data! {
...
+    /// A config is searched for by traversing a "config tree" in a bottom up fashion. It is chosen by the nearest first principle.
```
davidbarsky, line 506:
```suggestion
    /// A configuration is found using a depth-first search. The `rust-analyzer.toml` closest to a given file is selected.
```

crates/rust-analyzer/src/config.rs
```diff
@@ -827,16 +1177,45 @@ impl Config {
         caps: ClientCapabilities,
         workspace_roots: Vec<AbsPathBuf>,
         visual_studio_code_version: Option<Version>,
+        user_config_path: Option<Utf8PathBuf>,
```
davidbarsky, line 1180: question: are there non-`user` config paths? i would just rename all usage here to `config_path`, IMO.

(i see a "root config path", but I'd remove the `user` qualifier entirely.)

crates/rust-analyzer/src/config.rs
```diff
@@ -827,16 +1177,45 @@ impl Config {
...
             visual_studio_code_version,
+            client_config: ClientConfig::default(),
+            user_config: None,
+            ratoml_files: FxHashMap::default(),
+            default_config: ConfigData::default(),
+            source_root_parent_map: Arc::new(FxHashMap::default()),
+            user_config_path,
+            root_ratoml: None,
+            root_ratoml_path,
+            should_update: false,
```
davidbarsky, line 1218: I'd do two things:
1. keep `ConfigData` named `data` instead of `default_config`.
2. I'd group all the configuration file-related values into their own wrapper struct.

crates/rust-analyzer/src/config.rs
```diff
@@ -846,135 +1225,336 @@ impl Config {
...
+            snippets: self.snippets.clone().to_vec(),
+            limit: self.completion_limit().to_owned(),
+            enable_term_search: self.completion_termSearch_enable().to_owned(),
+            prefer_prelude: self.imports_preferPrelude(source_root).to_owned(),
+        }
+    }
+
+    pub fn detached_files(&self) -> &Vec<AbsPathBuf> {
+        // FIXME @alibektas : This is the only config that is confusing. If it's a proper configuration
+        // why is it not among the others? If it's client only which I doubt it is current state should be alright
```
davidbarsky, line 1304: it's weird because the VS Code extension, by default, cares about what folder rust-analyzer is running in. it's debatable whether it makes sense in this context, but I'm inclined to think it does.

davidbarsky (commented):
some nits; i'll try to review more later.

crates/rust-analyzer/src/global_state.rs
```diff
@@ -298,14 +330,26 @@ impl GlobalState {
...
+                if let Some(("rust-analyzer", Some("toml"))) = vfs_path.name_and_extension() {
+                    // Remember ids to use them after `apply_changes`
+                    modified_ratoml_files.insert(file.file_id, vfs_path.clone());
+                    if vfs_path == self.config.user_config_path() {
+                        user_config_file_id = Some(file.file_id);
+                    }
+
+                    if vfs_path == &root_vfs_path {
+                        root_path_ratoml = Some(file.file_id);
+                    }
```
Veykril, line 350: We can do these checks below when going through the `modified_ratoml_files`, gets rid of these two extra variables here. Can also put the id into the `_text` vars instead then, allowing us to omit the unwraps below

crates/rust-analyzer/src/global_state.rs
```diff
@@ -358,6 +422,56 @@ impl GlobalState {
 
         self.analysis_host.apply_change(change);
 
+        let config_change = {
+            let mut change = ConfigChange::default();
+            let inst = self.analysis_host.analysis();
```
Veykril, line 427: that var name bothers me :^)

crates/rust-analyzer/src/global_state.rs
```diff
@@ -358,6 +422,56 @@ impl GlobalState {
...
+                // belongs to a non-local source root, we will ignore it.
+                // As it doesn't make sense a users to use external config files.
+                if let Ok(source_root) = inst.source_root_id(file_id) {
+                    if let Ok(true) = inst.is_local_source_root(source_root) {
+                        if let Some((old_file, old_path, old_text)) =
+                            change.change_ratoml(source_root, file_id, vfs_path.clone(), text)
+                        {
+                            // SourceRoot has more than 1 RATOML files. In this case lexicographically
+                            // smaller wins.
+                            // FIXME @alibektas : What about symlinks?
```
Veykril, line 440: can be ignored I'd say. They don't really matter for this

crates/rust-analyzer/src/config.rs
Veykril: Ideally it should be (its tied to a source root right now which just coincidentally aligns with crate roots usually)

crates/rust-analyzer/src/config.rs
```diff
@@ -641,20 +650,303 @@ pub struct Config {
...
     root_path: AbsPathBuf,
-    detached_files: Vec<AbsPathBuf>,
     snippets: Vec<Snippet>,
     visual_studio_code_version: Option<Version>,
 
     default_config: ConfigData,
-    client_config: ConfigInput,
-    user_config: ConfigInput,
+    client_config: ClientConfig,
+    user_config: Option<RatomlNode>,
```
Veykril, line 658: doc comment would be good as to what the user config is

Veykril (commented):
Also not a complete review yet, will give the previous PR another pass and merge that first (then adjust things as I see to speed up things, instead of doing more review rounds on that)

crates/rust-analyzer/src/config.rs
alibektas: Ideally yes but this PR tries to abstract away VfsPaths as much as possible, so I just the current criterion is that a there is mapping from `SourceRootId`s to `FileId`s ( of RATOML files ). So you are right @_davidbarsky pointing this out. I should change this comment.

crates/rust-analyzer/src/config.rs
Veykril: This is indeed a very odd config and only necessary (I hope) because we are lacking proper support for handling detached files currently.

crates/rust-analyzer/src/config.rs
alibektas: Yeah naming things is not my strength. I though of XDG_CONFIG_.. but that is also not very accurate either. Matter of fact it was Veykril's suggestion to change it to user_config. In any case the name is still the same but I added comment.

crates/rust-analyzer/src/config.rs
```diff
@@ -651,23 +654,304 @@ pub struct Config {
...
+    file_id: FileId,
+}
+
+#[derive(Debug, Clone, Default)]
+struct ClientConfig {
+    node: FullConfigInput,
+    detached_files: Vec<AbsPathBuf>,
+}
+
+impl Serialize for RatomlNode {
```
Veykril, line 708: Is this really needed? When would we ever serialize this (there are more serialize additions that I am unsure why they are there, also noticed we derive it for things in `_config_data!` which also doesn't seem necessary?)

crates/rust-analyzer/src/config.rs
```diff
@@ -858,23 +1142,37 @@ pub struct ClientCommandsConfig {
...
+#[derive(Debug, Default)]
+pub struct ConfigError(Vec<ConfigErrorInner>);
+
+impl ConfigError {
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
 }
 
+impl ConfigError {}
```
Veykril, line 1159:
```suggestion
```

crates/rust-analyzer/src/config.rs
```diff
@@ -651,23 +654,304 @@ pub struct Config {
     workspace_roots: Vec<AbsPathBuf>,
     caps: lsp_types::ClientCapabilities,
     root_path: AbsPathBuf,
-    detached_files: Vec<AbsPathBuf>,
```
Veykril, line 654: Let's keep that here, it's a special thing so we should really handle it specially. (it being encoded in the config is technically wrong)

crates/rust-analyzer/src/config.rs
```diff
@@ -651,23 +654,304 @@ pub struct Config {
...
+            ConfigNodeKey::User => serializer.serialize_str("user"),
+        }
+    }
+}
+
+#[derive(Debug, Serialize)]
+enum ConfigNodeValue<'a> {
+    /// `rust-analyzer::config` module works by setting
+    /// a mapping between `SourceRootId` and `ConfigInput`.
+    /// Storing a `FileId` is mostly for debugging purposes.
```
Veykril, line 744: This comment feels out of place, as there is no file id to be seen here

crates/rust-analyzer/src/config.rs
```diff
@@ -651,23 +654,304 @@ pub struct Config {
...
+    pub fn user_config_path(&self) -> &VfsPath {
+        &self.user_config_path
+    }
+
+    pub fn should_update(&self) -> bool {
+        self.should_update
+    }
+
+    // FIXME @alibektas : Server's health uses error sink but in other places it is not used atm.
+    pub fn apply_change(&self, change: ConfigChange, error_sink: &mut ConfigError) -> Config {
```
Veykril, line 794: Let's make a convenience method for this as well `pub fn apply_change_whatever(&self, change: ConfigChange) -> (Config, ConfigError) {`

crates/rust-analyzer/src/config.rs
```diff
@@ -651,23 +654,304 @@ pub struct Config {
...
+        let mut toml_errors = vec![];
+        let mut json_errors = vec![];
+
+        config.should_update = false;
+
+        if let Some((file_id, change)) = change.user_config_change {
+            config.user_config = Some(RatomlNode {
+                file_id,
+                node: GlobalLocalConfigInput::from_toml(
+                    toml::from_str(change.to_string().as_str()).unwrap(),
```
Veykril, line 805: Nit

```suggestion
                    toml::from_str(&change.to_string()).unwrap(),
```

more importantly though, that unwrap is a bad idea, I am fairly sure this code path can be reached with an invalid toml file

crates/rust-analyzer/src/config.rs
```diff
@@ -651,23 +654,304 @@ pub struct Config {
...
+                }
+            }
+            config.should_update = true;
+        }
+
+        if let Some((file_id, change)) = change.root_ratoml_change {
+            config.root_ratoml = Some(RatomlNode {
+                file_id,
+                node: GlobalLocalConfigInput::from_toml(
+                    toml::from_str(change.to_string().as_str()).unwrap(),
```
Veykril, line 840: likewise

crates/rust-analyzer/src/config.rs
```diff
@@ -651,23 +654,304 @@ pub struct Config {
...
+
+        if let Some(change) = change.ratoml_file_change {
+            for (source_root_id, (file_id, _, text)) in change {
+                if let Some(text) = text {
+                    config.ratoml_files.insert(
+                        source_root_id,
+                        RatomlNode {
+                            file_id,
+                            node: GlobalLocalConfigInput::from_toml(
+                                toml::from_str(&text).unwrap(),
```
Veykril, line 855: likewise

crates/rust-analyzer/src/global_state.rs
```diff
@@ -350,10 +377,30 @@ impl GlobalState {
...
+                    change.change_file(file_id, Some(text.clone()));
+                    if let Some(vfs_path) = modified_ratoml_files.get(&file_id) {
+                        if vfs_path == self.config.user_config_path() {
+                            user_config_file = Some((file_id, Some(text.clone())));
+                        } else if vfs_path == &root_vfs_path {
+                            root_path_ratoml = Some((file_id, Some(text.clone())));
+                        } else {
+                            ratoml_text_map.insert(file_id, (vfs_path.clone(), Some(text.clone())));
+                        }
+                    }
```
Veykril, line 403: Can be simplified by having the lbocks return `None` or `Some(text)`, assign that to a  var and then branching afterwards using the var output. The only difference in the two arms is the insertion to `line_endings_map` after all

crates/rust-analyzer/src/global_state.rs
```diff
@@ -365,6 +412,54 @@ impl GlobalState {
 
         self.analysis_host.apply_change(change);
 
+        let config_change = {
+            let mut change = ConfigChange::default();
+            let snap = self.analysis_host.analysis();
```
Veykril, line 417:
```suggestion
            let snap = self.analysis_host.raw_database();
```
no need to create a snapshot here

crates/rust-analyzer/src/global_state.rs
```diff
@@ -365,6 +412,54 @@ impl GlobalState {
...
+                change.change_root_ratoml(Some((file_id, txt)));
+            }
+
+            change
+        };
+
+        let mut error_sink = ConfigError::default();
+        let config = self.config.apply_change(config_change, &mut error_sink);
+
+        if config.should_update() {
```
Veykril, line 456: We should return the `should_update` bool when calling `apply_change` instead of encoding it in the config as a field

crates/rust-analyzer/src/global_state.rs
```diff
@@ -365,6 +412,54 @@ impl GlobalState {
 
         self.analysis_host.apply_change(change);
 
+        let config_change = {
```
Veykril, line 415: We should skip all of the green diff here if `ratoml_text_map.is_empty() && user_config_file.is_none() && root_path_ra_tom.is_none()` (otherwise we unnecessarily replace the config with an equal copy/do unnecessary work). This function we are in runs on the main thread and also cancels all outstanding tasks, so as long as this function runs it will block the entire main loop from progressing. So we need ot be careful with the amount of work we do here.

crates/rust-analyzer/tests/slow-tests/main.rs
```diff
@@ -1373,3 +1377,1027 @@ version = "0.0.0"
 
     server.request::<WorkspaceSymbolRequest>(Default::default(), json!([]));
 }
+
```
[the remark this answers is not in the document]
alibektas: Following Barsky's suggestion I had removed the in-file module `ratoml` where these tests were , so I am reverting things back to its former state, assuming that you didn't mean that I migrate these to a separate file.

crates/rust-analyzer/src/config.rs
alibektas: Before I made this PR I thought I would implement a lsp_ext for users to use when they think that config infra isn't doing what it is supposed to be doing. Serializing was meant for this but now I removed everything related.

crates/rust-analyzer/src/config.rs
alibektas: I added sth but I am sure I misunderstood either what you mean or the purpose of this method :smile: .

crates/rust-analyzer/src/cli/scip.rs
```diff
@@ -35,12 +37,18 @@ impl flags::Scip {
...
 
         if let Some(p) = self.config_path {
             let mut file = std::io::BufReader::new(std::fs::File::open(p)?);
             let json = serde_json::from_reader(&mut file)?;
-            config.update(json)?;
+            let mut change = ConfigChange::default();
+            change.change_client_config(json);
+            let mut error_sink = ConfigError::default();
+            config = config.apply_change(change, &mut error_sink);
+            // FIXME @alibektas : What happens to errors without logging?
```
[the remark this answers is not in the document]
alibektas: I am holding onto this one as I couldn't quite understand what I need to do.

crates/rust-analyzer/tests/slow-tests/support.rs
```diff
@@ -207,8 +207,8 @@ impl Project<'_> {
 
         change.change_client_config(self.config);
         let mut error_sink = ConfigError::default();
-        assert!(error_sink.is_empty());
-        config = config.apply_change(change, &mut error_sink);
+        assert!(error_sink.is_empty(), "{error_sink:?}");
```
Veykril, line 210: Heh, that assert should be after the apply call actually

crates/rust-analyzer/src/config.rs
```diff
@@ -903,43 +802,48 @@ impl Config {
                 serde_json::Error::custom("expected a non-empty string"),
             ));
         }
-        config
+        (config, should_update)
+    }
+
+    pub fn apply_change_whatever(&self, change: ConfigChange) -> (Config, ConfigError) {
```
Veykril, line 808: The whatever was me being too lazy to come up with a proper name here 😅 Let's call this `apply_change` and rename the other to `apply_change_with_sink` and migrate the callers of that function to this one (probably need to change the return type to `(Config, ConfigError, bool)` then. I don' think any caller cares about setting up the error sink themselves (so we could probably get rid of the public API that takes the error sink as well)

Veykril (commented):
2 more things otherwise I think this is good to go

alibektas:
My splitting the config keys into global/local/client wasn't really meticulous, because I was thinking that you would intervene at some point :smile:. Are we planning to this in the near future or do you want to take a look before I actually merge this PR?

Veykril:
I haven't given the split a look yet tbf, I can do that afterwards / once you've fixed up the rest here

cbiffle:
Caveat: I am not familiar with this codebase and have no idea how big of a request I am making. I'm also not a VScode user, and some of the comments in the code seem VScode-specific, so it's possible I'm not understanding things correctly.

It would be nice for rust-embedded use cases if the following things were available as crate-local rather than workspace-level settings; from my read of the current code they are global:

- Control of features, in particular whether or not to pass `--all-features` -- for Cargo reasons, a lot of projects in the rust-embedded space use features to control processor variants, and cannot build with `--all-features` set. (This is not Hubris-specific.) While I assume your workspace-level feature specification allows for the `cratename/featurename` style of specification (since workspace-level features aren't a thing), it would be nice to specify these in crates themselves to avoid maintaining a big list at the root.
- Target triple. It's possible the target triple can be set here already -- rust-analyzer using the word "target" to mean both "a thing that can be built" and "the target triple" means it's difficult for me, not being super familiar with the codebase, to tell which case is referring to which. But since Cargo still has no stable way of indicating a preference to build a crate for a specific triple, generated rust-analyzer client configurations tend to fill in for this rn, and this PR could eliminate that if it were configurable at a crate level.

I assume that sufficiently flexible/configurable clients are able to override all of the settings? (This seems important if config files are checked in but you want to do something different in a specific editor.)

Veykril:
> It would be really nice for this to be at least a warning. Otherwise I foresee frustration where we set a key and it silently doesn't take effect.

We ought to have that indeed thats a good point to raise.

Regarding the preference for crate-local configs, the ones you raised are certainly targets for those but they will require some more follow up work to this. This PR is mainly bringing the infra up that allows us to have these config files for now.

otiv-emiel-vanseveren:
Hi, I'm really looking forward to this feature, what's stopping it from landing?

alibektas:
Sorry for the delay, I have been busy with other stuff. It will be merged in no time ( 3-4 days at most )

otiv-emiel-vanseveren:
@alibektas, please don't forget about us!

Veykril:
Ali is on vacation now, they unfortunately didn't manage to fix this up in time, I'll see if I can get to it instead.

otiv-emiel-vanseveren:
@Veykril, If possible, I would like to take: https://github.com/rust-lang/rust-analyzer/issues/17101, when merged?

Veykril:
Note to myself for next time, just merge these big PRs and disable their functionality if they are blocked on a minor change. Rebasing this ain't no fun

Veykril:
Okay diagnostics work now, though the UX is horrible for now and this generally needs some cleanup that I'll task @alibektas with as a follow up. It also seems like the server hangs up when you create a fresh toml file so I'll investigate that later

Veykril:
Quoting my comment on https://github.com/rust-lang/rust-analyzer/issues/13529 for visibility:

> #17058 will implement the core features here, there are still some issues with it (the user config is not working correctly yet), as well some restrictions with crate specific configs due to architecture reasons (they shouldn't really be noticeably in most cases). A lot of configs are also still missing from it and the diagnostics UX (when an invalid config is supplied) is really bad. All of that will be resolved in the near or far future (time will tell).
> 
> Important
> 
> We don't promise _**any**_ stability with this feature yet, any configs exposed may be removed again, the ordering may change etc.
> 
> Once the PR lands (should today), this issue will be used to track any updates on the feature.

PR #17073, opened 2024-04-15
Better inline preview for postfix completion

roife:
Better inline preview for postfix completion, a proper implementation of c5686c89413b3c6eead7b39637381ee353f8c05c.

Here editors may filter completion item with the text within `delete_range`, so we need to include the `receiver text` in the `lookup` (aka `FilterText` in LSP spec) for editors to find the completion item. (See https://github.com/rust-lang/rust-analyzer/issues/17036#issuecomment-2056614180, Thanks to [pascalkuthe](https://github.com/pascalkuthe))

### DIFF
--- a/crates/ide-completion/src/completions/dot.rs
+++ b/crates/ide-completion/src/completions/dot.rs
@@ -1035,7 +1035,7 @@ fn main() {
     fn postfix_drop_completion() {
         cov_mark::check!(postfix_drop_completion);
         check_edit(
-            "drop",
+            "x.drop",
             r#"
 //- minicore: drop
 struct Vec<T>(T);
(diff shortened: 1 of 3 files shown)


Veykril (approved):

davidbarsky:
Just as a heads up, this PR doesn't work too well with VS Code. On c0371309b40181baadda50f5fe6649f219be15d0, these are the completions I'm getting:

Screenshot 2024-04-15 at 11 23 34 AM

With af728741de6ad74517f34053efae864d7d75ac5d, I'm getting the postfix completions first:

Screenshot 2024-04-15 at 11 25 57 AM

I'm not sure how intentional that is, but it certainly feels a bit like a bug to me.

PR #17078, opened 2024-04-15
internal: Improve diagnostics performance

Veykril:


crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -531,8 +534,16 @@ fn types_of_subpatterns_do_match(pat: PatId, body: &Body, infer: &InferenceResul
...
+            None if *has_type_mismatches => (),
             None => {
-                body[pat].walk_child_pats(|subpat| walk(subpat, body, infer, has_type_mismatches))
+                let pat = &body[pat];
+                if let Pat::ConstBlock(expr) | Pat::Lit(expr) = *pat {
+                    *has_type_mismatches |= infer.type_mismatch_for_expr(expr).is_some();
+                    if *has_type_mismatches {
+                        return;
+                    }
+                }
```
Veykril, line 545: The literal patterns don't have the mismatch associated with them, but contained expression within has. This fixes the `bug` call we had in one of the tests, so I replaced that with `never!` again.

PR #17088, opened 2024-04-17
internal: Implement a linear storage type for salsa queries that use single (newtyped) integers as keys

Veykril:
My hope was to be able to remove the outer `RwLock` around the hashmap, since the only write is to insert into it, though unfortunately I haven't found a concurrent append-only vec / slab that supports what we'd need here.


Veykril:
Managed to get rid of the table lock but there is no visible perf improvement from this so..

PR #6246, opened 2020-10-15
Follow symlinks when walking project trees

dfoxfranke:
Fixes #3691.

~~WIP pending further testing~~:

- [X] Verify that symlinked files get indexed.
- [x] Verify that files in symlinked directories get indexed.
- [x] Verify that inotify events are properly received and handled when the target of a symlink resides outside the project tree.

PR #17093, opened 2024-04-17
fix: VFS should not walk circular symlinks

Wilfred:
As of #6246, rust-analyzer follows symlinks. This can introduce an infinite loop if symlinks point to parent directories.

Considering that #6246 was added in 2020 without many bug reports, this is clearly a rare occurrence. However, I am observing rust-analyzer hang on projects that have symlinks of the form:

```
test/a_symlink -> ../../
```

Ignore symlinks that only point to the parent directories, as this is more robust but still allows typical symlink usage patterns.

### DIFF
--- a/crates/vfs-notify/src/lib.rs
+++ b/crates/vfs-notify/src/lib.rs
@@ -9,7 +9,10 @@
 
 #![warn(rust_2018_idioms, unused_lifetimes)]
 
-use std::fs;
+use std::{
+    fs,
+    path::{Component, Path},
+};
 
 use crossbeam_channel::{never, select, unbounded, Receiver, Sender};
 use notify::{Config, RecommendedWatcher, RecursiveMode, Watcher};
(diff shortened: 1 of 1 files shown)


crates/vfs-notify/src/lib.rs
```diff
@@ -258,3 +263,23 @@ fn read(path: &AbsPath) -> Option<Vec<u8>> {
...
+/// possible to construct cycle using two or more symlinks), but it
+/// catches common cases.
+fn path_is_parent_symlink(path: &Path) -> bool {
+    // We can't canonicalize this path, presume it's not a symlink.
+    let Ok(canonical) = path.canonicalize() else {
+        return false;
+    };
+
+    // The canonicalized path is the same, it's not a symlink.
+    if path == canonical {
```
Veykril, line 280: This will always fail on windows, as canonicalizing paths on windows will give back an [extended length path](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file)

Wilfred: Updated to use `read_link()` instead :)

Veykril (approved):

PR #17094, opened 2024-04-17
feat: Add convert From to TryFrom assist

Lindronics:
Adds a new code assist to convert a `From` impl into a `TryFrom` impl.

This is useful in situations where it turns out after or halfway through writing a `From` implementation that the conversion is actually fallible.

## Example

https://github.com/rust-lang/rust-analyzer/assets/26360861/872ec7c4-c9ff-451c-9453-4baaaad47326


Veykril:
One minor thing, thanks!
@bors delegate+

Veykril:
Thanks!
@bors r+

PR #17105, opened 2024-04-19
Make test harness arguments configurable and not `--nocapture`.

### LINKED ISSUE #12737: Make `--nocapture` optional
When running tests, rust-analyzer [unconditionally](https://github.com/rust-lang/rust-analyzer/blob/361cddc16ca85b9981d3d892eda474aa57d92bcc/crates/rust-analyzer/src/cargo_target_spec.rs) adds `--nocapture` to the test arguments. I would like this to be optional, because using `--nocapture` leads to sometimes-unreadable interleaved output, especially when two tests fail simultaneously, or when successful tests are producing verbose output intended to be read only if the test fails.

kpreid:
* Added config `runnables.extraTestBinaryArgs` to control the args.
* The default is `--show-output` rather than `--nocapture` to prevent unreadable output when 2 or more tests fail or print output at once.
* Renamed variables in `CargoTargetSpec::runnable_args()` for clarity.

Fixes #12737.

Suggested changelog info:

> New Features
>
> * add `rust-analyzer.runnables.extraTestBinaryArgs` to configure test harness options when running tests; replaces a previously hard-coded `--nocapture` option.

I'm not sure I made the right choices in vocabulary, between “binary”, “executable”, and “runnable”, and “launch” vs. “execute”, so let me know if I missed something to be consistent with in the naming and documentation.

### DIFF
--- a/crates/rust-analyzer/src/cargo_target_spec.rs
+++ b/crates/rust-analyzer/src/cargo_target_spec.rs
@@ -33,53 +33,55 @@ impl CargoTargetSpec {
         kind: &RunnableKind,
         cfg: &Option<CfgExpr>,
     ) -> (Vec<String>, Vec<String>) {
-        let mut args = Vec::new();
-        let mut extra_args = Vec::new();
+        let extra_test_binary_args = snap.config.runnables().extra_test_binary_args;
+
+        let mut cargo_args = Vec::new();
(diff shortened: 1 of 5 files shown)


Veykril (approved):

Veykril:
TIL about `--show-output`, thanks!
@bors r+

PR #17114, opened 2024-04-20
minor: Add in-rust-tree feature to hir-expand

lnicola:
CC https://github.com/rust-lang/rust/pull/124199#issuecomment-2067701168


Veykril:
Right that is still needed then ugh

@bors r+

PR #17115, opened 2024-04-20
Try to generate more meaningful names in json converter

leviska:
I just found out about rust-analyzer json converter, but I think it would be more convenient, if names were more useful, like using the names of the keys.

Let's look at some realistic arbitrary json:

```json
{
    "user": {
        "address": {
            "street": "Main St",
            "house": 3
        },
        "email": "example@example.com"
    }
}
```
I think, new generated code is much easier to read and to edit, than the old:
```rust
// Old
struct Struct1{ house: i64, street: String }
struct Struct2{ address: Struct1, email: String }
struct Struct3{ user: Struct2 }

// New
struct Address1{ house: i64, street: String }
struct User1{ address: Address1, email: String }
struct Root1{ user: User1 }
```

Ideally, if we drop the numbers, I can see it being usable just as is (may be rename root)
```rust
struct Address{ house: i64, street: String }
struct User{ address: Address, email: String }
struct Root{ user: User }
```

Sadly, we can't just drop them, because there can be multiple fields (recursive) with the same name, and we can't just easily retroactively add numbers if the name has 2 instances due to parsing being single pass.
We could ignore the `1` and add number only if it's > 1, but I will leave this open to discussion and right now made it the simpler way

In sum, even with numbers, I think this PR still helps in readability

### DIFF
--- a/crates/ide-diagnostics/src/handlers/json_is_not_rust.rs
+++ b/crates/ide-diagnostics/src/handlers/json_is_not_rust.rs
@@ -30,7 +30,8 @@ struct State {
 impl State {
     fn generate_new_name(&mut self, name: &str) -> ast::Name {
         let name = stdx::to_camel_case(name);
-        let count = if let Some(count) = self.names.get(&name) {
+        let count = if let Some(count) = self.names.get_mut(&name) {
+            *count += 1;
             *count
         } else {
             self.names.insert(name.clone(), 1);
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/json_is_not_rust.rs
```diff
@@ -22,15 +22,22 @@ use crate::{fix, Diagnostic, DiagnosticCode, DiagnosticsConfig, Severity};
...
-    fn generate_new_name(&mut self) -> ast::Name {
-        self.struct_counts += 1;
-        make::name(&format!("Struct{}", self.struct_counts))
+    fn generate_new_name(&mut self, name: &str) -> ast::Name {
+        let name = stdx::to_camel_case(name);
+        let count = if let Some(count) = self.names.get_mut(&name) {
+            *count += 1;
+            *count
+        } else {
+            self.names.insert(name.clone(), 1);
```
leviska, line 37: I don't want to copy if name exists, so this long if

leviska:
I just remembered, that the name in json can be invalid struct name... I'll try to find a way to check that I guess

leviska:
Actually, the old code fails too, because it doesn't check the names for struct variables...

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17138, opened 2024-04-25
feature: Make generate function assist generate a function as a constructor if the generated function has the name "new" and is an asscociated function.

### LINKED ISSUE #17050: Generate function for `Foo::new` should assume a `Self` return type
For

```rust
struct Foo {
    field1: String,
    field2: u32,
}

let foo = Foo::new();
```

using the generate function assist on `new` could generate

```rust
impl Foo {
    fn new() -> Self {
        Self {
            field1: todo!(),
            field2: todo!(),
        }
    }
}
```

instead of

```rust
impl Foo {
    fn new() -> _ {
        todo!()
    }
}
```

which I think would be a much more helpful in many cases.

Kohei316:
close #17050
This PR makes `generate function assist` generate a function as a constructor if the generated function has the name "new" and is an asscociated function.
If the asscociate type is a record struct, it generates the constructor like this.
```rust
impl Foo {
    fn new() -> Self {
        Self { field_1: todo!(), field_2: todo!() }
    } 
}
``` 
If the asscociate type is a tuple struct, it generates the constructor like this. 
```rust
impl Foo {
    fn new() -> Self {
        Self(todo!(), todo!())
    } 
}
``` 
If the asscociate type is a unit struct, it generates the constructor like this. 
```rust
impl Foo {
    fn new() -> Self {
        Self
    } 
}
``` 
If the asscociate type is another adt, it generates the constructor like this. 
```rust
impl Foo {
    fn new() -> Self {
        todo!()
    } 
}
```

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -1489,6 +1489,14 @@ impl Adt {
             .map(|arena| arena.1.clone())
     }
 
+    pub fn as_struct(&self) -> Option<Struct> {
+        if let Self::Struct(v) = self {
+            Some(*v)
+        } else {
+            None
+        }
+    }
+
     pub fn as_enum(&self) -> Option<Enum> {
         if let Self::Enum(v) = self {
             Some(*v)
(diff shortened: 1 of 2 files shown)


Veykril (commented):
A test case with a `new` function that takes arguments would be nice as well (which I don't think this change here handles)

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17140, opened 2024-04-25
handle {self} when removing unused imports

### LINKED ISSUE #17139: removing unused imports generates invalid `use module_name::self`
**rust-analyzer version**: rust-analyzer version: 0.3.1932-standalone (47a901b9b 2024-04-21)

**rustc version**: rustc 1.79.0-nightly (ef8b9dcf2 2024-04-24)

**editor or extension**: VScode v0.3.1932

```rust

mod inner {
    pub struct X();
    pub struct Y();
}

mod z {
    use super::inner::{self, X}$0;

    fn f() {
        let y = inner::Y();
    }
}

```

becomes

```
mod inner {
    pub struct X();
    pub struct Y();
}

mod z {
    use super::inner:self;

    fn f() {
        let y = inner::Y();
    }
}
```
which is invalid. It should be `use super::inner` instead.

harrysarson:
Fixes #17139 

On master

```rs
mod inner {
    pub struct X();
    pub struct Y();
}

mod z {
    use super::inner::{self, X}$0;

    fn f() {
        let y = inner::Y();
    }
}
```

becomes

```rs
mod inner {
    pub struct X();
    pub struct Y();
}

mod z {
    use super::inner:self;

    fn f() {
        let y = inner::Y();
    }
}
```

with this fix it instead becomes

```

```rs
mod inner {
    pub struct X();
    pub struct Y();
}

mod z {
    use super::inner;

    fn f() {
        let y = inner::Y();
    }
}
```

### DIFF
--- a/crates/syntax/src/ast/node_ext.rs
+++ b/crates/syntax/src/ast/node_ext.rs
@@ -378,9 +378,26 @@ impl ast::UseTreeList {
 
     /// Remove the unnecessary braces in current `UseTreeList`
     pub fn remove_unnecessary_braces(mut self) {
+        // Returns true iff there is a single subtree and it is not the self keyword. The braces in
+        // `use x::{self};` are necessary and so we should not remove them.
+        let has_single_subtree_that_is_not_self = |u: &ast::UseTreeList| {
+            if let Some((single_subtree,)) = u.use_trees().collect_tuple() {
+                // We have a single subtree, check whether it is self.
+
+                let is_self = single_subtree.path().as_ref().map_or(false, |path| {
+                    path.segment().and_then(|seg| seg.self_token()).is_some()
+                        && path.qualifier().is_none()
+                });
+
+                !is_self
+            } else {
+                // Not a single subtree
+                false
+            }
+        };
+
         let remove_brace_in_use_tree_list = |u: &ast::UseTreeList| {
-            let use_tree_count = u.use_trees().count();
-            if use_tree_count == 1 {
+            if has_single_subtree_that_is_not_self(u) {
                 if let Some(a) = u.l_curly_token() {
                     ted::remove(a)
                 }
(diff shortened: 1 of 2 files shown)


Veykril:
That transform is generally not valid, `use foo::bar::{self}` imports only the type namespace, where as `use foo::bar;` imports value, type and macro namespace. So transforming like that can actually cause name collisions.

Now there is still a bug here in that we produce malformed syntax but we need to keep the braces.

crates/syntax/src/ast/node_ext.rs
```diff
@@ -376,11 +376,24 @@ impl ast::UseTreeList {
             .filter_map(|it| it.into_token().filter(|it| it.kind() == T![,]))
     }
 
+    pub fn is_just_self(&self) -> bool {
+        if let Some((single_subtree,)) = self.use_trees().collect_tuple() {
+            fn path_is_self(path: &ast::Path) -> bool {
```
Veykril, line 381: This can be a closure

harrysarson: have made this into a closure

harrysarson:
> That transform is generally not valid, 

The normalize import IDE assist (https://github.com/rust-lang/rust-analyzer/blob/master/crates/ide-assists/src/handlers/normalize_import.rs) does transform `{self}`. I.e. the following unit test would parse if added to normalize_import.rs

```rs

    #[test]
    fn test_collapses_self() {
        check_assist_variations!("std::fmt::{self}", "std::fmt");
    }
```

Is that a bug?

Veykril:
Unless the assist there checks that `fmt` only refers to a type namespace there I'd consider that a bug

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17144, opened 2024-04-26
manual: remove suggestion of rust-project.json example

cbiffle:
The manual has been linking to the repo

https://github.com/rust-analyzer/rust-project.json-example/tree/master

This repo does not contain a rust-project.json, does not appear to have _ever_ contained a rust-project.json, and my bug report about this has gone untouched: https://github.com/rust-analyzer/rust-project.json-example/issues/4

Since I can't figure out an example, this commit removes the link pending better documentation.

### DIFF
--- a/docs/user/manual.adoc
+++ b/docs/user/manual.adoc
@@ -787,8 +787,6 @@ There are three ways to feed `rust-project.json` to rust-analyzer:
 
 Relative paths are interpreted relative to `rust-project.json` file location or (for inline JSON) relative to `rootUri`.
 
-See https://github.com/rust-analyzer/rust-project.json-example for a small example.
-
 You can set the `RA_LOG` environment variable to `rust_analyzer=info` to inspect how rust-analyzer handles config and project loading.
 
 Note that calls to `cargo check` are disabled when using `rust-project.json` by default, so compilation errors and warnings will no longer be sent to your LSP client.


lnicola:
image

That repo has an example of the third option, just like it was pointed out in https://github.com/rust-analyzer/rust-project.json-example/issues/4#issuecomment-2028007200.

cbiffle:
That seems like an example of how to write a vscode configuration file. Is this section intended to be vscode-specific? Perhaps it would be worth telling readers to look in the `.vscode` directory (which I skipped, not being a vscode user, looking for a `rust-project.json` file that doesn't exist)?

lnicola:
> Specify `"rust-analyzer.linkedProjects": [ { "roots": […​], "crates": […​] }]` inline.

It's literally that. It's not Code-specific, every LSP client will support changing the server settings. But the syntax depends on the client, unfortunately.

Veykril:
I wonder if we can make a simple CLI thing for r-a that turns a cargo project into a `rust-project.json` file. That feels like a nice way to help people figure out the format 🤔 Also let's archive that repo, it being vscode specific isn't really helpful.

Veykril:
Also there is https://json.schemastore.org/rust-project.json which might help (I dont remember if we have docs for rust-project.json)

lnicola:
We have https://rust-analyzer.github.io/manual.html#non-cargo-based-projects, which may or may not be up-to-date.

Veykril:
I'll merge this, I am not too fond of that example, especially because that third option isn't meant to be written manually anyways.
@bors r+

PR #17146, opened 2024-04-26
minor: let diagnostics.enable's scope be limited to DiagnosticsCode::Ra only

### LINKED ISSUE #17048: "Fill struct fields" assist has stopped showing up in many cases
**rust-analyzer version**: 0.3.1916-standalone

**rustc version**: rustc 1.77.1 (7cf61ebde 2024-03-27)

**relevant settings**:

```json
"rust-analyzer.hover.documentation.keywords.enable": false,
"rust-analyzer.cargo.buildScripts.useRustcWrapper": false,
"rust-analyzer.workspace.symbol.search.kind": "all_symbols",
"rust-analyzer.diagnostics.enable": false,
"rust-analyzer.inlayHints.closureStyle": "rust_analyzer",
"rust-analyzer.cargo.sysrootQueryMetadata": true,
```

screenshot_24-04-10_22:17:50

This is not the first time I've seen this happen, but this time it isn't happening in a macro (which I understand are difficult to handle in IDEs), and I see no other reason why "fill struct fields" shouldn't show up here. I've written `wgpu::BindGroupLayoutEntry { }` and while field completions work, there is no option to fill *all* missing fields like there used to.

alibektas:
The documentation for the config `rust-analyzer.diagnostics.enable` states :  *"Whether to show native rust-analyzer diagnostics"*. However as I further investigated #17048 I came to the conclusion that the thing it currently does is to disable/enable diagnostics completely. I am not sure if that's the desired behavior. If so, this PR should better be closed. 
If not this PR changes the said config to regulate whether we should show diagnostics with diagnostics code `DiagnosticCode::Ra` or not and thereby fixes #17048. 

As to the new test : It only serves the purpose of demonstrating what the change means. As soon as it becomes clear that this PR has any sense I will write a test under ide_diagnostics

### DIFF
--- a/crates/ide-diagnostics/src/lib.rs
+++ b/crates/ide-diagnostics/src/lib.rs
@@ -408,7 +408,8 @@ pub fn diagnostics(
     }
 
     res.retain(|d| {
-        !(ctx.config.disabled.contains(d.code.as_str())
+        !(matches!(d.code, DiagnosticCode::Ra(..)) && !ctx.config.enabled
+            || ctx.config.disabled.contains(d.code.as_str())
             || ctx.config.disable_experimental && d.experimental)
     });
 
--- a/crates/ide/src/lib.rs
+++ b/crates/ide/src/lib.rs
@@ -686,7 +686,7 @@ impl Analysis {
         };
 
         self.with_db(|db| {
-            let diagnostic_assists = if diagnostics_config.enabled && include_fixes {
+            let diagnostic_assists = if include_fixes {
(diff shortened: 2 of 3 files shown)


Veykril (commented):
The code in question here seems correct to me, in that these parts configure the native diagnostics only (opposed to flycheck ones). Us mapping some of r-a's diagnostics to rustc error codes is irrelevant to the setting in my eyes. I'd say the point of the setting is to just not get real time diagnostics which is what it currently configures.

Veykril:
I'll close this, as the behavior looks correct to me

PR #17147, opened 2024-04-26
docs: Fix typo in VS Code setting description

Wilfred:
Replace 'futureg' with 'future'.


Veykril:
Needs a `cargo test -p rust-analyzer` to update the generated config stuff (I feel like we should move that into a prepare step for publishing instead little reason in this being a test honestly)

PR #17148, opened 2024-04-27
fix: Tracing span names should match function names

Wilfred:
When viewing traces, it's slightly confusing when the span name doesn't match the function name. Ensure the names are consistent.

(It might be worth moving most of these to use `#[tracing::instrument]` so the name can never go stale. @davidbarsky suggested that is marginally slower, so I've just done the simple change here.)

### DIFF
--- a/crates/base-db/src/change.rs
+++ b/crates/base-db/src/change.rs
@@ -51,7 +51,7 @@ impl FileChange {
     }
 
     pub fn apply(self, db: &mut dyn SourceDatabaseExt) {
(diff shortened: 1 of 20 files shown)


davidbarsky:
> (It might be worth moving most of these to use `#[tracing::instrument]` so the name can never go stale. @davidbarsky suggested that is marginally slower, so I've just done the simple change here.)

it's also kinda more annoying if you only want to record _some_ parameters. The runtime speed should be the same, it's just a papercut in compile times.

Veykril:
Fwiw, using the attribute is also worse for rust-analyzer analysis performance

PR #17150, opened 2024-04-27
internal: add no-new-root check to josh pull

RalfJung:
We probably don't want a second root to show up...

Also clear the rust-version file while we are at it.


lnicola (approved):

lnicola:
Thanks!

@bors r+

PR #17165, opened 2024-04-30
Omit changelog header from GitHub release notes

### LINKED ISSUE #16455: GitHub release changelog number misinterpreted as issue reference
The description of the GitHub releases under https://github.com/rust-lang/rust-analyzer/releases uses the format
```
= Changelog #{changelog_n}

...
```

The problem is that GitHub misinterprets this `#<number>` as issue / PR reference, which is quite confusing:
Changelog screenshot

It seems escaping the `#` with a `\` has no effect (at least not in the preview), so maybe it would be best to simply omit the `#` (or add a zero-width space or similar between)?

Marcono1234:
Resolves #16455

I wasn't sure whether one of you maintainers wanted to implement this, so I gave it a try. Feedback is appreciated! No worries if you reject this pull request (and implement the changes yourself) in case these changes here are not clean enough.

I have tested this using `test_data/input.adoc` (renamed to `2024-01-01.adoc`) with the following command:
```
cargo run --package xtask --bin xtask publish-release-notes --dry-run xtask/test_data/2024-01-01.adoc
```
It seems to work correctly, but would be good if you could double check.

### DIFF
--- a/xtask/src/publish.rs
+++ b/xtask/src/publish.rs
@@ -9,6 +9,22 @@ impl flags::PublishReleaseNotes {
     pub(crate) fn run(self, sh: &Shell) -> anyhow::Result<()> {
         let asciidoc = sh.read_file(&self.changelog)?;
         let mut markdown = notes::convert_asciidoc_to_markdown(std::io::Cursor::new(&asciidoc))?;
+
+        // Remove changelog header because GitHub misinterprets the changelog number as issue reference
+        let changelog_header_prefix = "# Changelog #";
+        let mut changelog_header_start =
+            markdown.find(changelog_header_prefix).expect("should contain changelog header");
+        changelog_header_start =
+            markdown[..changelog_header_start].rfind('\n').map_or(0, |i| i + 1);
+        let mut changelog_header_end = changelog_header_start + changelog_header_prefix.len();
+        changelog_header_end += markdown[changelog_header_end..].find('\n').unwrap() + 1;
+        while changelog_header_end < markdown.len()
+            && markdown.as_bytes()[changelog_header_end] == b'\n'
+        {
+            changelog_header_end += 1;
+        }
+        markdown.replace_range(changelog_header_start..changelog_header_end, "");
+
         let file_name = check_file_name(self.changelog)?;
         let tag_name = &file_name[0..10];
         let original_changelog_url = create_original_changelog_url(&file_name);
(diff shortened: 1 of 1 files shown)


xtask/src/publish.rs
```diff
@@ -9,6 +9,22 @@ impl flags::PublishReleaseNotes {
     pub(crate) fn run(self, sh: &Shell) -> anyhow::Result<()> {
         let asciidoc = sh.read_file(&self.changelog)?;
         let mut markdown = notes::convert_asciidoc_to_markdown(std::io::Cursor::new(&asciidoc))?;
+
+        // Remove changelog header because GitHub misinterprets the changelog number as issue reference
```
lnicola, line 13:
```suggestion
        // Remove changelog header because it's not going to be relevant for the GitHub release notes
```

xtask/src/publish.rs
```diff
@@ -9,6 +9,22 @@ impl flags::PublishReleaseNotes {
...
+        let changelog_header_prefix = "# Changelog #";
+        let mut changelog_header_start =
+            markdown.find(changelog_header_prefix).expect("should contain changelog header");
+        changelog_header_start =
```
lnicola, line 17: This seems a little too complicated, and I'm not sure why it's needed.

My suggestion is to edit the AsciiDoc source, not the derived Markdown, since the latter is less predictable. That way, we only need to trim off the first line (until the first `'\n'`).

We could also strip off the `Commit: ...` line because it's redundant, but that seems less important.

Marcono1234: Good point; but it seems the AsciiDoc to Markdown conversion currently expects a title, and otherwise fails with "Error: document title not found". And I did not want to remove that code.

So I have changed it now to only remove the `#` from the changelog number. But since this deviates a bit from what this pull request was originally about, so I have marked it as Draft for now again. What do you think?

lnicola: Sorry, I didn't realize the title was needed for AsciiDoc.

I'd still prefer to strip off the line, since it's useless there.

Marcono1234: Maybe the changelog number is a bit useful though? For example when someone refers to it because they read it on the project website. I don't think GitHub itself numbers the releases.

In case you still think the complete title should be removed, should I then also remove the code which tries to convert this AsciiDoc title to Markdown? Or how should I handle the title being required currently?

(Also, in case it is easier for you to directly take this over, just let me know; this PR has "Allow edits by maintainers" enabled in case that is helpful.)

PR #17174, opened 2024-05-02
Fix: infer type of async block with tail return expr

### LINKED ISSUE #17106: return inside async blocks hints `impl Future<Output = !>`
VSC: rust-analyzer version: 0.3.1924-standalone
`rust-analyzer --version`: rust-analyzer 1.77.0 (aedd173 2024-03-17)

rustc 1.77.0 (aedd173a2 2024-03-17)

both VSCodium
```
Version: 1.87.2
Release: 24072
Commit: f7ed6417a7cfae92f98129384c13b505bb9cfc53
Date: 2024-03-12T18:30:28.994Z
Electron: 27.3.2
ElectronBuildId: undefined
Chromium: 118.0.5993.159
Node.js: 18.17.1
V8: 11.8.172.18-electron.0
OS: Linux x64 5.15.0-102-generic
```

and helix 24.3 (2cadec0b)

```rust
#[allow(unused_variables)]
#[tokio::main]
async fn main() {
    // indicates impl Future<Output = i32>
    let a1 = async { 8 };
    let a2 = async {
        return 8; // indicates a impl Future<Output = !>
    };
    let a3 = async {
        loop {
            return 8;
        } // suddenly fine impl Future<Output = i32>
    };

    let t1 = tokio::spawn(async {
        return 8; // indicates a JoinHandle<!>
    });
    let t2 = tokio::task::spawn_blocking(|| {
        return 8; // this is as expected JoinHandle<i32>
    });
}
```

Despite hintng Future<Output = !> and JoinHandle<!> the futures actually return correctly the value. Just the hint is incorrect.
Btw I am not a fan of writing explicitly `return` within async blocks, but if the compiler supports it, the hints should be correct as well.

scrnshot with hints:
image

Kohei316:
Fixes #17106
The `infer_async_block` method calls the `infer_block` method internally, which returns the never type without coercion when `tail_expr` is `None` and `ctx.diverges` is `Diverges::Always`.This is the reason for the bug in this issue.
https://github.com/rust-lang/rust-analyzer/blob/cfce2bb46da62950a8b70ddb0b2a12332da1b1e1/crates/hir-ty/src/infer/expr.rs#L1411-L1413

This PR solves the bug by adding a process to coerce after calling `infer_block` method.

This code passes all the tests, including tests I added for this isuue, however, I am not sure if this solution is right. I think that this solution is an ad hoc solution. So, I would appreciate to have your review.
I apologize if I'm off the mark, but `infer_async_block` method should be rewritten to share code with the process of infering type of `expr::Closure` instead of the `infer_block` method. That way it will be closer to the infer process of rustc.

### DIFF
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -933,6 +933,7 @@ impl InferenceContext<'_> {
         let prev_ret_coercion =
             mem::replace(&mut self.return_coercion, Some(CoerceMany::new(ret_ty.clone())));
 
+        // FIXME: We should handle async blocks like we handle closures
         let expected = &Expectation::has_type(ret_ty);
         let (_, inner_ty) = self.with_breakable_ctx(BreakableKind::Border, None, None, |this| {
             let ty = this.infer_block(tgt_expr, *id, statements, *tail, None, expected);
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -933,8 +933,23 @@ impl InferenceContext<'_> {
         let prev_ret_coercion =
             mem::replace(&mut self.return_coercion, Some(CoerceMany::new(ret_ty.clone())));
 
+        let expected = &Expectation::has_type(ret_ty);
         let (_, inner_ty) = self.with_breakable_ctx(BreakableKind::Border, None, None, |this| {
-            this.infer_block(tgt_expr, *id, statements, *tail, None, &Expectation::has_type(ret_ty))
+            let ty = this.infer_block(tgt_expr, *id, statements, *tail, None, expected);
+            if let Some(target) = expected.only_has_type(&mut this.table) {
+                match this.coerce(Some(tgt_expr), &ty, &target) {
+                    Ok(res) => res,
+                    Err(_) => {
+                        this.result.type_mismatches.insert(
+                            tgt_expr.into(),
+                            TypeMismatch { expected: target.clone(), actual: ty.clone() },
+                        );
+                        target
+                    }
+                }
+            } else {
+                ty
+            }
         });
 
         self.diverges = prev_diverges;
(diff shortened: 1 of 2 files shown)


crates/hir-ty/src/infer/expr.rs
```diff
@@ -933,8 +933,23 @@ impl InferenceContext<'_> {
         let prev_ret_coercion =
             mem::replace(&mut self.return_coercion, Some(CoerceMany::new(ret_ty.clone())));
 
+        let expected = &Expectation::has_type(ret_ty);
         let (_, inner_ty) = self.with_breakable_ctx(BreakableKind::Border, None, None, |this| {
-            this.infer_block(tgt_expr, *id, statements, *tail, None, &Expectation::has_type(ret_ty))
+            let ty = this.infer_block(tgt_expr, *id, statements, *tail, None, expected);
```
Veykril, line 938: I think it suffices to merely replace this by `self.infer_return(tgt_expr)`?

Kohei316: I do not fully understand these functions and may be misunderstanding, but I think that we cannot use this way. First, `infer_return` method only coerces and does not return a value. Next, the method internally calls `infer_expr_inner`, which internally calls `infer_async_block`. This results in an infinite loop.

Veykril: Right

Veykril:
> I apologize if I'm off the mark, but infer_async_block method should be rewritten to share code with the process of infering type of expr::Closure instead of the infer_block method. That way it will be closer to the infer process of rustc.

No I think that is correct, we should in fact handle it more like closures, as really that's what it kind of is.

crates/hir-ty/src/infer/expr.rs
```diff
@@ -933,8 +933,23 @@ impl InferenceContext<'_> {
         let prev_ret_coercion =
             mem::replace(&mut self.return_coercion, Some(CoerceMany::new(ret_ty.clone())));
 
+        let expected = &Expectation::has_type(ret_ty);
```
Veykril, line 936:
```suggestion
        // FIXME: We should handle async blocks like we handle closures
        let expected = &Expectation::has_type(ret_ty);
```

Veykril:
Thanks for looking into this!
@bors r+

PR #17175, opened 2024-05-02
fix: Fix implicit ty args being lowered where they shouldn't

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/17173


crates/hir-def/src/item_tree/lower.rs
```diff
@@ -662,7 +662,7 @@ impl<'a> Ctx<'a> {
             let attrs = RawAttrs::new(self.db.upcast(), &param, self.body_ctx.span_map());
             debug_assert!(self.generic_param_attr_buffer.insert(item, attrs).is_none());
         };
-
+        self.body_ctx.take_impl_traits_bounds();
```
Veykril, line 665: Issue was basically this line missing. Without it, we start allocating impl trait args for things that are not related to the generic params / where clauses

Veykril:
cc @dfireBird this fixes an issue that your recent PR introduced just so you are aware
@bors r+

PR #17203, opened 2024-05-08
Fix OOM caused by term search

kilpkonn:
The issue came from multi Cartesian product for exprs with many (25+) arguments, each having multiple options.
The solution is two fold:
### Avoid blowing up in Cartesian product
**Before the logic was:**
    1. Find expressions for each argument/param - there may be many
    2. Take the Cartesian product (which blows up in some cases)
    4. If there are more than 2 options throw them away by squashing them to `Many` 
**Now the logic is:**
    1. Find expressions for each argument/param and squash them to `Many` if there are more than 2 as otherwise we are guaranteed to also have more than 2 after taking the product which means squashing them anyway.
    2. Take the Cartesian product on iterator
    3. Start consuming it one by one
    4. If there are more than 2 options throw them away by squashing them to `Many`  (same as before)

This is also why I had to update some tests as the expressions get squashed to many more eagerly.

### Use fuel to avoid long search times and high memory usage
Now all the tactics use `should_continue:  Fn() -> bool` to chech if they should keep iterating _(Similarly to chalk)_.
This reduces the search times by a magnitude, for example from ~139ms/hole to ~14ms/hole for `ripgrep` crate.
There are slightly less expressions found, but I think speed gain worth it for usability.
Also note that syntactic hits decreases more because of squashing so you simple need to run search multiple times to get full terms.
Also the worst case time (For example `nalgebra` crate cus it has tons of generics) has search times mostly under 200ms. 

Benchmarks on `ripgrep` crate
Before:
```
Tail Expr syntactic hits: 291/1692 (17%)
Tail Exprs found: 1253/1692 (74%)
Term search avg time: 139ms
````
After:
```
Tail Expr syntactic hits: 239/1692 (14%)
Tail Exprs found: 1226/1692 (72%)
Term search avg time: 14ms
```

### DIFF
--- a/crates/ide-completion/src/render.rs
+++ b/crates/ide-completion/src/render.rs
@@ -1808,8 +1808,7 @@ fn f() { A { bar: b$0 }; }
                 fn baz() [type]
                 ex baz() [type]
                 ex bar() [type]
-                ex A { bar: baz() }.bar [type]
-                ex A { bar: bar() }.bar [type]
+                ex A { bar: ... }.bar [type]
                 st A []
                 fn f() []
             "#]],
@@ -1947,8 +1946,8 @@ fn main() {
 }
             "#,
             expect![[r#"
-                ex core::ops::Deref::deref(&T(S)) (use core::ops::Deref) [type_could_unify]
                 ex core::ops::Deref::deref(&t) (use core::ops::Deref) [type_could_unify]
+                ex core::ops::Deref::deref(&T(S)) (use core::ops::Deref) [type_could_unify]
                 lc m [local]
                 lc t [local]
                 lc &t [type+local]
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: crates/ide-completion/src/render.rs [hunk 1]
```diff
@@ -1947,8 +1946,8 @@ fn main() {
 }
             "#,
             expect![[r#"
-                ex core::ops::Deref::deref(&T(S)) (use core::ops::Deref) [type_could_unify]
                 ex core::ops::Deref::deref(&t) (use core::ops::Deref) [type_could_unify]
+                ex core::ops::Deref::deref(&T(S)) (use core::ops::Deref) [type_could_unify]
                 lc m [local]
```
Veykril, line 1951: Are these results non-deterministic-ish?

kilpkonn: Not really, will debug later why exactly the ordering changed.

Veykril:
Thanks!
@bors r+

crates/ide-completion/src/render.rs
Veykril: Would be great if you could take a quick look at that, will merge this given the ordering change being possibly non-deterministic shouldn't be related to this PR.

kilpkonn: Checked that collapsing terms before Cartesian product  https://github.com/rust-lang/rust-analyzer/pull/17203/files#diff-5797341fbf7c8fcf87bd0bfd50c533bff963b96d1185fadbda31cfebc114fdbaR130-R135 causes the order to change. Since I think there is no preferred order I think the current order is find.
If you got any good suggestions for preferred order then I'm happy to change it - would like to switch from `Vec` to `FxHashSet` anyway for its uniqueness guarantees, but couldn't really do it as it would cause "random" order

PR #17207, opened 2024-05-08
fix: Report all LSP protocol errors with invalid_data

Wilfred:
Previously we did not use invalid_data for serde errors, making it harder to understand errors when the client sends malformed data to the server.


davidbarsky (approved):

lnicola (approved):

lnicola:
Thanks!

@bors r+

PR #17208, opened 2024-05-08
fix: Report both IO errors and main_loop errors

Wilfred:
If rust-analyzer receives a malformed LSP request, the IO thread terminates with a meaningful error, but then closes the channel.

Once the channel has closed, the main_loop also terminates, but it only has RecvError and can't show a meaningful error. As a result, rust-analyzer would incorrectly claim that the client forgot to shutdown.

```
$ buggy_lsp_client | rust-analyzer
Error: client exited without proper shutdown sequence
```

Instead, include both error messages when the server shuts down.

### DIFF
--- a/crates/rust-analyzer/src/bin/main.rs
+++ b/crates/rust-analyzer/src/bin/main.rs
@@ -259,9 +259,15 @@ fn run_server() -> anyhow::Result<()> {
         config.rediscover_workspaces();
     }
 
-    rust_analyzer::main_loop(config, connection)?;
+    // If the io_threads have an error, there's usually an error on the main
+    // loop too because the channels are closed. Ensure we report both errors.
+    match (rust_analyzer::main_loop(config, connection), io_threads.join()) {
+        (Err(loop_e), Err(join_e)) => anyhow::bail!("{loop_e}\n{join_e}"),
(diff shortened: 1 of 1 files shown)


crates/rust-analyzer/src/bin/main.rs
```diff
@@ -259,9 +259,15 @@ fn run_server() -> anyhow::Result<()> {
...
     }
 
-    rust_analyzer::main_loop(config, connection)?;
+    // If the io_threads have an error, there's usually an error on the main
+    // loop too because the channels are closed. Ensure we report both errors.
+    match (rust_analyzer::main_loop(config, connection), io_threads.join()) {
+        (Err(loop_e), Err(join_e)) => anyhow::bail!("{loop_e}\n{join_e}"),
+        (Ok(_), Err(join_e)) => anyhow::bail!("{join_e}"),
+        (Err(loop_e), Ok(_)) => anyhow::bail!("{loop_e}"),
+        _ => {}
```
davidbarsky, line 268: this variant is just `Ok(_), Ok(_)`, right? do you mind handling that explicitly here?

davidbarsky (approved):

crates/rust-analyzer/src/bin/main.rs
Wilfred: Done :)

PR #17216, opened 2024-05-12
fix: extract mod to file should respect path attribute

Young-Flash:
close https://github.com/rust-lang/rust-analyzer/issues/17181

### DIFF
--- a/crates/ide-assists/src/handlers/move_module_to_file.rs
+++ b/crates/ide-assists/src/handlers/move_module_to_file.rs
@@ -1,6 +1,7 @@
 use std::iter;
 
 use ast::edit::IndentLevel;
+use hir::HasAttrs;
 use ide_db::base_db::AnchoredPathBuf;
 use itertools::Itertools;
 use stdx::format_to;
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/move_module_to_file.rs
```diff
@@ -107,6 +116,72 @@ mod tests {
...
+fn test() {}
+mod $0expr {
+    struct A {}
+}
+"#,
+            r#"
+//- /parser/__mod.rs
+fn test() {}
+mod expr;
+//- /parser/expr.rs
```
Veykril, line 137: This won't be correct, the module would need to be at `/parser/__mod/expr.rs` for this to be loaded no?

Young-Flash: `parser/expr.rs` can be loaded.

mod_with_path

IMO, with #path specified, `parser/__mod.rs` for mod `parser`(in the following code), can be seen as `parser/mod.rs`
```rust
#[path="parser/__mod.rs"]
mod parser;
```

Veykril: Right, thanks for double checking.

PR #17217, opened 2024-05-12
docs: Prioritize rustup over manual installation

senekor:
Unless I'm missing something, it seems weird to have the manual installation instructions above the `rustup` one. It's more complicated, platform-dependent and doesn't update automatically.

Many of the individual editor instructions link to this section (Helix, Neovim, Emacs) so it would be great if this is simple and straight-forward.

### DIFF
--- a/docs/user/manual.adoc
+++ b/docs/user/manual.adoc
@@ -133,6 +133,18 @@ Specifically, `rustup` includes a proxy called `rust-analyzer`, which can cause 
 === rust-analyzer Language Server Binary
 
 Other editors generally require the `rust-analyzer` binary to be in `$PATH`.
+
+==== rustup
+
+`rust-analyzer` is available in `rustup`:
+
+[source,bash]
+----
+$ rustup component add rust-analyzer
+----
+
+==== Manual installation
+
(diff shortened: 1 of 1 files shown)


lnicola:
Sorry, but we don't want this. That version doesn't get any backports after beta branches and can lag for other reasons too. It will be months old by the time it reaches stable.

On GitHub we have weekly releases, so we recommend using those instead.

senekor:
That's ok, thanks for the quick response. I have two follow-up questions:

- Would it be good to add a note about that to the rustup installation section? I've been running the rustup installation for a long time, not knowing it is not the best experience. I can open another PR.
- How are users supposed to keep their installation up to date? Run this manual setup once per week? Write their own scripts / cron jobs? Or is there an easier way?

lnicola:
I thought there was a note about this in the manual, but I can't really check until much later today. We should definitely add it, or put it in a better place if possible.

There's no recommended way to update, but replacing the binary is enough. Arch Linux has a rust-analyzer package which gets very fast updates (couple of hours, usually). Not sure about brew or anything else.

We don't plan to add automatic updates, for various reasons.

PR #17227, opened 2024-05-14
fix: Hash file contents to verify whether file actually changed

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/16580

### DIFF
--- a/crates/hir-def/src/item_tree.rs
+++ b/crates/hir-def/src/item_tree.rs
@@ -29,6 +29,7 @@
 //!
 //! In general, any item in the `ItemTree` stores its `AstId`, which allows mapping it back to its
 //! surface syntax.
+#![allow(unexpected_cfgs)]
 
 mod lower;
 mod pretty;
@@ -467,13 +468,12 @@ macro_rules! mod_items {
         pub enum GenericModItem {
             $(
                 $(
-                    #[cfg_attr(FALSE, $generic_params)]
+                    #[cfg_attr(ignore_fragment, $generic_params)]
                     $typ(FileItemTreeId<$typ>),
                 )?
             )+
         }
 
-        #[allow(unexpected_cfgs)]
         impl From<GenericModItem> for ModItem {
             fn from(id: GenericModItem) -> ModItem {
                 match id {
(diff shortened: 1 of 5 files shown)


crates/vfs/src/lib.rs
```diff
@@ -93,20 +96,13 @@ impl nohash_hasher::IsEnabled for FileId {}
 pub struct Vfs {
     interner: PathInterner,
     data: Vec<FileState>,
-    // FIXME: This should be a HashMap<FileId, ChangeFile>
-    // right now we do a nasty deduplication in GlobalState::process_changes that would be a lot
-    // easier to handle here on insertion.
-    changes: Vec<ChangedFile>,
-    // The above FIXME would then also get rid of this probably
-    created_this_cycle: Vec<FileId>,
+    changes: IndexMap<FileId, ChangedFile, BuildHasherDefault<FxHasher>>,
```
lnicola, line 99: I think we have a `FxIndexMap` somewhere around.

crates/vfs/src/lib.rs
```diff
@@ -217,41 +209,74 @@ impl Vfs {
...
+        let mut set_data = |change_kind| {
+            self.data[file_id.0 as usize] = match change_kind {
+                &Change::Create(_, hash) | &Change::Modify(_, hash) => FileState::Exists(hash),
+                Change::Delete => FileState::Deleted,
+            };
+        };
+
         let changed_file = ChangedFile { file_id, change: change_kind };
-        self.changes.push(changed_file);
+        match self.changes.entry(file_id) {
```
lnicola, line 234: I'm missing the overall picture, but can we merge `Change::Create` and `Change::Modify` into a single variant, similar to `FileState::Exists`? That would simplify the aggregation below.

crates/hir-def/src/item_tree.rs
```diff
@@ -467,13 +468,12 @@ macro_rules! mod_items {
         pub enum GenericModItem {
             $(
                 $(
-                    #[cfg_attr(FALSE, $generic_params)]
+                    #[cfg_attr(ignore_fragment, $generic_params)]
```
lnicola, line 471: Can we use `#[cfg(any())]` instead?

Veykril: I like using the condition as the description of what this is even doing tbf 😅 (obviously we'd not need this if `$ignore{...}` was stable

crates/vfs/src/lib.rs
Veykril: We do in `ide-db` I think which isn't a dependency here

lnicola (approved):

PR #17246, opened 2024-05-16
feature: teach rust-analyzer to discover `linked_projects`

davidbarsky:
This PR's been a long-time coming, but like the title says, it introduces server-side project discovery and removes the extension hooks I previously introduced. I don't think this PR is ready to land, but here are the things I'm feeling squishy about:
- I don't think I like the idea of introducing the `cargo-metadata` command-but-for-everything-else in the `flycheck` module, but the progress reporting infrastructure was too convenient to pass up. Happy to move it elsewhere.

Here are the things I _know_ I need to change:
- For progress reporting, I'm extracting from a `serde_json::Value` that corresponds to `tracing_subsciber::fmt::Layer`'s JSON output. I'd like to make this a bit more structured/documented than the current nonsense I wrote.
- The progress reporting currently hardcodes "Buck"; it should be deriving that from the previously mentioned more-structured-output.
- This doesn't handle *reloading* when a corresponding buildfile is changed. It should be doing that.

<details>
<summary>Anyway, here's a video of rust-analyzer discovering a Buck target.</summary>

https://github.com/rust-lang/rust-analyzer/assets/2067774/be6cd9b9-2c9a-402d-847f-05f860a91df1
</details>

### DIFF
--- a/crates/flycheck/src/project_json.rs
+++ b/crates/flycheck/src/project_json.rs
@@ -1,0 +1,152 @@
+//! A `cargo-metadata`-equivalent for non-Cargo build systems.
+use std::{io, process::Command};
+
+use crossbeam_channel::Sender;
+use paths::{AbsPathBuf, Utf8Path, Utf8PathBuf};
+use project_model::ProjectJsonData;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+use crate::command::{CommandHandle, ParseFromLine};
+
+pub const ARG_PLACEHOLDER: &str = "{arg}";
+
+/// A command wrapper for getting a `rust-project.json`.
+///
+/// This is analogous to `cargo-metadata`, but for non-Cargo build systems.
+pub struct Discover {
+    command: Vec<String>,
+    sender: Sender<DiscoverProjectMessage>,
+}
+
+#[derive(PartialEq, Clone, Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub enum DiscoverArgument {
+    Path(#[serde(serialize_with = "serialize_abs_pathbuf")] AbsPathBuf),
+    Buildfile(#[serde(serialize_with = "serialize_abs_pathbuf")] AbsPathBuf),
+}
+
+fn serialize_abs_pathbuf<S>(path: &AbsPathBuf, se: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    let path: &Utf8Path = path.as_ref();
+    se.serialize_str(path.as_str())
+}
+
+impl Discover {
+    /// Create a new [Discover].
+    pub fn new(sender: Sender<DiscoverProjectMessage>, command: Vec<String>) -> Self {
+        Self { sender, command }
+    }
+
+    /// Spawn the command inside [Discover] and report progress, if any.
+    pub fn spawn(&self, discover_arg: DiscoverArgument) -> io::Result<DiscoverHandle> {
+        let command = &self.command[0];
+        let args = &self.command[1..];
+
+        let args: Vec<String> = args
+            .iter()
+            .map(|arg| {
+                if arg == ARG_PLACEHOLDER {
+                    serde_json::to_string(&discover_arg).expect("Unable to serialize args")
+                } else {
+                    arg.to_owned()
+                }
+            })
+            .collect();
+
+        let mut cmd = Command::new(command);
+        cmd.args(args);
+
+        Ok(DiscoverHandle { _handle: CommandHandle::spawn(cmd, self.sender.clone())? })
+    }
+}
+
+/// A handle to a spawned [Discover].
+#[derive(Debug)]
+pub struct DiscoverHandle {
+    _handle: CommandHandle<DiscoverProjectMessage>,
+}
+
+/// An enum containing either progress messages, an error,
+/// or the materialized `rust-project`.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(tag = "kind")]
+#[serde(rename_all = "snake_case")]
+enum DiscoverProjectData {
+    Finished { buildfile: Utf8PathBuf, project: ProjectJsonData },
+    Error { error: String, source: Option<String> },
+    Progress { message: String },
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum DiscoverProjectMessage {
+    Finished { project: ProjectJsonData, buildfile: AbsPathBuf },
+    Error { error: String, source: Option<String> },
+    Progress { message: String },
+}
+
+impl DiscoverProjectMessage {
+    fn new(data: DiscoverProjectData) -> Self {
+        match data {
+            DiscoverProjectData::Finished { project, buildfile, .. } => {
+                let buildfile = buildfile.try_into().expect("Unable to make path absolute");
+                DiscoverProjectMessage::Finished { project, buildfile }
+            }
+            DiscoverProjectData::Error { error, source } => {
+                DiscoverProjectMessage::Error { error, source }
+            }
+            DiscoverProjectData::Progress { message } => {
+                DiscoverProjectMessage::Progress { message }
+            }
+        }
+    }
+}
+
+impl ParseFromLine for DiscoverProjectMessage {
+    fn from_line(line: &str, _error: &mut String) -> Option<Self> {
+        // can the line even be deserialized as JSON?
+        let Ok(data) = serde_json::from_str::<Value>(line) else {
+            let err = DiscoverProjectData::Error { error: line.to_owned(), source: None };
+            return Some(DiscoverProjectMessage::new(err));
+        };
+
+        let Ok(data) = serde_json::from_value::<DiscoverProjectData>(data) else {
+            return None;
+        };
+
+        let msg = DiscoverProjectMessage::new(data);
+        Some(msg)
+    }
+
+    fn from_eof() -> Option<Self> {
+        None
+    }
+}
+
+#[test]
+fn test_deserialization() {
+    let message = r#"
+    {"kind": "progress", "message":"querying build system","input":{"files":["src/main.rs"]}}
+    "#;
+    let message: DiscoverProjectData =
+        serde_json::from_str(message).expect("Unable to deserialize message");
+    assert!(matches!(message, DiscoverProjectData::Progress { .. }));
+
+    let message = r#"
+    {"kind": "error", "error":"failed to deserialize command output","source":"command"}
+    "#;
+
+    let message: DiscoverProjectData =
+        serde_json::from_str(message).expect("Unable to deserialize message");
+    assert!(matches!(message, DiscoverProjectData::Error { .. }));
+
+    let message = r#"
+    {"kind": "finished", "project": {"sysroot": "foo", "crates": [], "runnables": []}, "buildfile":"rust-analyzer/BUILD"}
+    "#;
+
+    let message: DiscoverProjectData =
+        serde_json::from_str(message).expect("Unable to deserialize message");
+    assert!(matches!(message, DiscoverProjectData::Finished { .. }));
+}
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -328,6 +328,102 @@ config_data! {
         /// `textDocument/rangeFormatting` request. The rustfmt option is unstable and only
         /// available on a nightly build.
         rustfmt_rangeFormatting_enable: bool = false,
+
+        /// Enables automatic discovery of projects using [`DiscoverWorkspaceConfig::command`].
+        ///
+        /// [`DiscoverWorkspaceConfig`] also requires setting `progress_label` and `files_to_watch`.
+        /// `progress_label` is used for the title in progress indicators, whereas `files_to_watch`
+        /// is used to determine which build system-specific files should be watched in order to
+        /// reload rust-analyzer.
+        ///
+        /// Below is an example of a valid configuration:
+        /// ```json
+        /// "rust-analyzer.workspace.discoverConfig": {
+        ///     "command": [
+        ///         "rust-project",
+        ///         "develop-json",
+        ///         {arg}
+        ///     ],
+        ///     "progressLabel": "rust-analyzer",
+        ///     "filesToWatch": [
+        ///         "BUCK",
+        ///     ],
+        /// }
+        /// ```
+        ///
+        /// ## On `DiscoverWorkspaceConfig::command`
+        ///
+        /// **Warning**: This format is provisional and subject to change.
+        ///
+        /// [`DiscoverWorkspaceConfig::command`] *must* return a JSON object
+        /// corresponding to `DiscoverProjectData::Finished`:
+        ///
+        /// ```norun
+        /// #[derive(Debug, Clone, Deserialize, Serialize)]
+        /// #[serde(tag = "kind")]
+        /// #[serde(rename_all = "snake_case")]
+        /// enum DiscoverProjectData {
+        ///     Finished { buildfile: Utf8PathBuf, project: ProjectJsonData },
+        ///     Error { error: String, source: Option<String> },
+        ///     Progress { message: String },
+        /// }
+        /// ```
+        ///
+        /// As JSON, `DiscoverProjectData::Finished` is:
+        ///
+        /// ```json
+        /// {
+        ///     // the internally-tagged representation of the enum.
+        ///     "kind": "finished",
+        ///     // the file used by a non-Cargo build system to define
+        ///     // a package or target.
+        ///     "buildfile": "rust-analyzer/BUILD",
+        ///     // the contents of a rust-project.json, elided for brevity
+        ///     "project": {
+        ///         "sysroot": "foo",
+        ///         "crates": []
+        ///     }
+        /// }
+        /// ```
+        ///
+        /// It is encouraged, but not required, to use the other variants on
+        /// `DiscoverProjectData` to provide a more polished end-user experience.
+        ///
+        /// `DiscoverWorkspaceConfig::command` may *optionally* include an `{arg}`,
+        /// which will be substituted with the JSON-serialized form of the following
+        /// enum:
+        ///
+        /// ```norun
+        /// #[derive(PartialEq, Clone, Debug, Serialize)]
+        /// #[serde(rename_all = "camelCase")]
+        /// pub enum DiscoverArgument {
+        ///    Path(AbsPathBuf),
+        ///    Buildfile(AbsPathBuf),
+        /// }
+        /// ```
+        ///
+        /// The JSON representation of `DiscoverArgument::Path` is:
+        ///
+        /// ```json
+        /// {
+        ///     "path": "src/main.rs"
+        /// }
+        /// ```
+        ///
+        /// Similarly, the JSON representation of `DiscoverArgument::Buildfile` is:
+        ///
+        /// ```
+        /// {
+        ///     "buildfile": "BUILD"
+        /// }
+        /// ```
+        ///
+        /// `DiscoverArgument::Path` is used to find and generate a `rust-project.json`,
+        /// and therefore, a workspace, whereas `DiscoverArgument::buildfile` is used to
+        /// to update an existing workspace. As a reference for implementors,
+        /// buck2's `rust-project` will likely be useful:
+        /// https://github.com/facebook/buck2/tree/main/integrations/rust-project.
+        workspace_discoverConfig: Option<DiscoverWorkspaceConfig> = None,
     }
 }
 
@@ -924,6 +1020,21 @@ impl Config {
         );
         (config, e, should_update)
     }
+
+    pub fn add_linked_projects(&mut self, data: ProjectJsonData, buildfile: AbsPathBuf) {
+        let linked_projects = &mut self.client_config.0.global.linkedProjects;
+
+        let new_project = ManifestOrProjectJson::DiscoveredProjectJson { data, buildfile };
+        match linked_projects {
+            Some(projects) => {
+                match projects.iter_mut().find(|p| p.manifest() == new_project.manifest()) {
+                    Some(p) => *p = new_project,
+                    None => projects.push(new_project),
+                }
+            }
+            None => *linked_projects = Some(vec![new_project]),
+        }
+    }
 }
 
 #[derive(Default, Debug)]
@@ -986,6 +1097,14 @@ impl From<ProjectJson> for LinkedProject {
     fn from(v: ProjectJson) -> Self {
         LinkedProject::InlineJsonProject(v)
     }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DiscoverWorkspaceConfig {
+    pub command: Vec<String>,
+    pub progress_label: String,
+    pub files_to_watch: Vec<String>,
 }
 
 pub struct CallInfoConfig {
@@ -1528,15 +1647,27 @@ impl Config {
     pub fn has_linked_projects(&self) -> bool {
         !self.linkedProjects().is_empty()
     }
-    pub fn linked_manifests(&self) -> impl Iterator<Item = &Utf8Path> + '_ {
+
+    pub fn linked_manifests(&self) -> impl Iterator<Item = &AbsPath> + '_ {
         self.linkedProjects().iter().filter_map(|it| match it {
             ManifestOrProjectJson::Manifest(p) => Some(&**p),
-            ManifestOrProjectJson::ProjectJson(_) => None,
+            // despite having a buildfile, using this variant as a manifest
+            // will fail.
+            ManifestOrProjectJson::DiscoveredProjectJson { .. } => None,
+            ManifestOrProjectJson::ProjectJson { .. } => None,
         })
     }
+
     pub fn has_linked_project_jsons(&self) -> bool {
-        self.linkedProjects().iter().any(|it| matches!(it, ManifestOrProjectJson::ProjectJson(_)))
+        self.linkedProjects()
+            .iter()
+            .any(|it| matches!(it, ManifestOrProjectJson::ProjectJson { .. }))
     }
+
+    pub fn discover_workspace_config(&self) -> Option<&DiscoverWorkspaceConfig> {
+        self.workspace_discoverConfig().as_ref()
+    }
+
     pub fn linked_or_discovered_projects(&self) -> Vec<LinkedProject> {
         match self.linkedProjects().as_slice() {
             [] => {
@@ -1560,6 +1691,12 @@ impl Config {
                             .map_err(|e| tracing::error!("failed to load linked project: {}", e))
                             .ok()
                             .map(Into::into)
+                    }
+                    ManifestOrProjectJson::DiscoveredProjectJson { data, buildfile } => {
+                        let root_path =
+                            buildfile.parent().expect("Unable to get parent of buildfile");
+
+                        Some(ProjectJson::new(None, root_path, data.clone()).into())
                     }
                     ManifestOrProjectJson::ProjectJson(it) => {
                         Some(ProjectJson::new(None, &self.root_path, it.clone()).into())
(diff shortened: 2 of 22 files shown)


crates/rust-analyzer/src/lsp/utils.rs
```diff
@@ -75,25 +75,21 @@ impl GlobalState {
...
-                self.send_notification::<lsp_types::notification::ShowMessage>(
-                    lsp_types::ShowMessageParams { typ: lsp_types::MessageType::ERROR, message },
-                );
             }
         }
+        self.show_message(
+            lsp_types::MessageType::ERROR,
+            message,
+            tracing::enabled!(tracing::Level::ERROR),
+        );
```
Veykril, line 92: This unnecessarily opens the log button notification now when there is no additional info in the logs

Veykril (commented):
Haven't looked at the main_loop.rs and reload.rs changes yet, will do that tomorrow

Veykril:
> I don't think I like the idea of introducing the cargo-metadata command-but-for-everything-else in the flycheck module, but the progress reporting infrastructure was too convenient to pass up. Happy to move it elsewhere.

The flycheck crate is a mess now (at least name wise) anyways as it also does stuff for the test explorer, so that doesn't really make things a lot worse now...

crates/rust-analyzer/src/config.rs
```diff
@@ -929,6 +930,18 @@ impl Config {
         self.workspace_roots.extend(paths);
     }
 
+    pub fn add_linked_projects(&mut self, projects: Vec<ProjectJsonData>) {
+        let linked_projects = &mut self.client_config.global.linkedProjects;
+
+        let mut new_projects: Vec<_> =
+            projects.into_iter().map(ManifestOrProjectJson::ProjectJson).collect();
```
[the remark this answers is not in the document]
davidbarsky: i don't think that a `mut` can be dropped in the first borrow, but i've deferred collecting until the match arms.

crates/rust-analyzer/src/config.rs
```diff
@@ -272,6 +272,11 @@ config_data! {
         /// The warnings will be indicated by a blue squiggly underline in code
         /// and a blue icon in the `Problems Panel`.
         diagnostics_warningsAsInfo: Vec<String> = vec![],
+        /// Enables automatic discovery of projects using the discoverCommand.
+        ///
+        /// Setting this command will result in rust-analyzer starting indexing
+        /// only once a Rust file has been opened.
+        discoverCommand: Option<Vec<String>> = None,
```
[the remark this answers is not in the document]
davidbarsky: fixed!

crates/project-model/Cargo.toml
```diff
@@ -14,6 +14,7 @@ doctest = false
 [dependencies]
 anyhow.workspace = true
 cargo_metadata.workspace = true
+crossbeam-channel.workspace = true
```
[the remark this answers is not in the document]
davidbarsky: removed, good point

crates/rust-analyzer/src/lsp/utils.rs
davidbarsky: where should error messages from flycheck be printed? flycheck eagerly gets that, but printing with additional information interrupts the error message.

that's probably a sign that the the stuff in flycheck isn't sufficient here, huh

crates/rust-analyzer/src/config.rs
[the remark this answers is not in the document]
Veykril: Ah sorry, with the borrow I meant instead of
```diff
-Some(projects) => projects.append(&mut new_projects.collect::<Vec<_>>()),
+Some(projects) => projects.extend(new_projects),
```
didn't realize you were using append, was thinking of extend :)

crates/rust-analyzer/src/lsp/utils.rs
Veykril: I don't think I understand, why is flycheck relevant here? This is just for logging errors, if the error has no extra information we show all the info in the popup, if it has additional info we log that and tell the user to check the logs by giving the pop up a button. I'm a bit confused as to why you need to make changes here.

davidbarsky: right, sorry: it's the interaction of several different features. 
-  `tracing::enabled!(tracing::Level::ERROR)` is almost always true (since it's the default level for `tracing-subscriber`)
- `CargoActor` in `flycheck` will be printing out the lines it couldn't parse as JSON (which will often contain the error itself...), but that's interleaved with the `additional_info`.

If I'm *not* make this change, the end result is the following (where the `additional_info` is "unable to do the things"): 

<details>
<summary>Screenshot of mangled logs</summary>
Screenshot 2024-05-23 at 11 18 14 AM
</details>

> that's probably a sign that the the stuff in flycheck isn't sufficient here, huh

This is what I meant when I said "that's probably a sign that the the stuff in flycheck isn't sufficient here, huh". However, having thought about it, it's not that flycheck is incorrect! It's that the CLI that i'm calling here, rust-project](https://github.com/facebook/buck2/tree/main/integrations/rust-project), that's wrong. It should communicate the underlying error in a more structured matter.

I'll hack on something today and propose something in this PR, which would let me undo the change here.

crates/rust-analyzer/src/reload.rs
```diff
@@ -81,9 +82,10 @@ impl GlobalState {
                 &self.config.lru_query_capacities_config().cloned().unwrap_or_default(),
             );
         }
+
         if self.config.linked_or_discovered_projects() != old_config.linked_or_discovered_projects()
         {
-            self.fetch_workspaces_queue.request_op("discovered projects changed".to_owned(), false)
+            self.fetch_workspaces_queue.request_op("discovered projects changed".to_owned(), true)
```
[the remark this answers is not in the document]
davidbarsky: I can change this to an enum, since I'm doing a bunch of surgery here anyways: would something like (bad namesincoming!) `ReloadBehavior { Force, Done }` be a decent starting point?

crates/rust-analyzer/src/config.rs
```diff
@@ -929,6 +931,16 @@ impl Config {
...
     }
 
+    pub fn add_linked_projects(&mut self, projects: Vec<ProjectJsonData>) {
+        let linked_projects = &mut self.client_config.global.linkedProjects;
+
+        let new_projects = projects.into_iter().map(ManifestOrProjectJson::ProjectJson);
+        match linked_projects {
+            Some(projects) => projects.append(&mut new_projects.collect::<Vec<_>>()),
+            None => *linked_projects = Some(new_projects.collect::<Vec<_>>()),
+        }
```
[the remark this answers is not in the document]
davidbarsky: i'm not the biggest fan of combinators, but i also don't feel particularly strongly about this.

davidbarsky:
For "reload rust-analyzer reasons when a `TARGETS`/`BUCK` file changes"; this PR is based atop of https://github.com/rust-lang/rust-analyzer/pull/16840.

Veykril:
Is this now ready for re-review?

davidbarsky:
> Is this now ready for re-review?

Almost. I'm got a few things I want to get fixed first, but it should be ready by tomorrow. I'll comment here when it is.

crates/flycheck/src/json_workspace.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn from_line(line: &str, _error: &mut String) -> Option<Self> {
+        let Ok(value) = serde_json::from_str::<serde_json::Value>(line) else {
+            return Some(DiscoverProjectMessage::Error { message: line.to_owned(), context: None });
+        };
+
+        if let Ok(project) = serde_json::from_value::<FinishedOutput>(value.clone()) {
+            return Some(DiscoverProjectMessage::Finished(project));
+        }
+
+        if let Some(message) = value.pointer("/fields/message") {
```
davidbarsky, line 78: I'm not a fan of this approach—it currently does some shotgun parsing `tracing-subscriber`'s default JSON output, but this should be more of a defined JSON schema (comprised of serde-annotated structs, not [_the_ JSON Schema](https://json-schema.org/)) that rustlings, rustc, or Bazel's `rules_rust` can make use of. I'm thinking of making it a reusable `tracing` layer.

crates/rust-analyzer/Cargo.toml
```diff
@@ -39,6 +39,7 @@ mimalloc = { version = "0.1.30", default-features = false, optional = true }
 lsp-server.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
+pretty_assertions = "1.4"
```
davidbarsky: whoops, removing this.

crates/rust-analyzer/src/config.rs
```diff
@@ -1565,8 +1594,13 @@ impl Config {
...
                     }
-                    ManifestOrProjectJson::ProjectJson(it) => {
-                        Some(ProjectJson::new(None, &self.root_path, it.clone()).into())
+                    ManifestOrProjectJson::ProjectJson { data, buildfile } => {
+                        let root_path = match buildfile {
+                            Some(buildfile) => AbsPathBuf::try_from(buildfile.clone())
+                                .expect("unable to get AbsPathBuf"),
+                            None => self.root_path.clone(),
+                        };
+                        Some(ProjectJson::new(None, &root_path, data.clone()).into())
```
davidbarsky, line 1605: I wrote this in a haze, but I don't think it should be necessary to clone here.

crates/rust-analyzer/src/handlers/notification.rs
```diff
@@ -152,13 +154,15 @@ pub(crate) fn handle_did_save_text_document(
         // Re-fetch workspaces if a workspace related file has changed
         if let Some(abs_path) = vfs_path.as_path() {
             if reload::should_refresh_for_change(abs_path, ChangeKind::Modify) {
-                state
-                    .fetch_workspaces_queue
-                    .request_op(format!("workspace vfs file change saved {abs_path}"), false);
+                state.fetch_workspaces_queue.request_op(
+                    format!("workspace vfs file change saved {abs_path}"),
+                    (Some(abs_path.to_owned()), false),
```
davidbarsky, line 159: I was lazy here; I should make this a proper struct instead of a tuple. For context, I'm using this in order to pass the saved file over to `GlobalState::fetch_workspaces` so that I can kick off the `discover_handle` to do its thing.

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -146,11 +158,13 @@ impl GlobalState {
             self.register_did_save_capability();
         }
 
-        self.fetch_workspaces_queue.request_op("startup".to_owned(), false);
-        if let Some((cause, force_crate_graph_reload)) =
-            self.fetch_workspaces_queue.should_start_op()
-        {
-            self.fetch_workspaces(cause, force_crate_graph_reload);
+        if self.config.discover_command().is_none() {
```
davidbarsky, line 161: this feels a little gross, but I can't think of a better way to do this

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -190,7 +204,12 @@ impl GlobalState {
                     lsp_types::DocumentFilter {
                         language: None,
                         scheme: None,
-                        pattern: Some("**/rust-analyzer.toml".into()),
+                        pattern: Some("**/BUCK".into()),
+                    },
+                    lsp_types::DocumentFilter {
+                        language: None,
+                        scheme: None,
+                        pattern: Some("**/TARGETS".into()),
```
davidbarsky, line 212: I shouldn't be hard-coding this here; this should be coming from configuration.

(i think it's a hard blocker until this is a config, to be clear.)

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -787,6 +847,28 @@ impl GlobalState {
         }
     }
 
+    fn handle_discover_msg(&mut self, message: flycheck::DiscoverProjectMessage) {
+        match message {
+            flycheck::DiscoverProjectMessage::Finished(output) => {
+                self.report_progress("Buck", Progress::End, None, None, None);
```
davidbarsky, line 853: `"Buck"` shouldn't be hard-coded here; it should either come from configuration or the crate's label.

(hard blocker again, IMO.)

crates/rust-analyzer/src/reload.rs
```diff
@@ -81,9 +83,11 @@ impl GlobalState {
                 &self.config.lru_query_capacities_config().cloned().unwrap_or_default(),
             );
         }
+
         if self.config.linked_or_discovered_projects() != old_config.linked_or_discovered_projects()
         {
-            self.fetch_workspaces_queue.request_op("discovered projects changed".to_owned(), false)
+            self.fetch_workspaces_queue
+                .request_op("discovered projects changed".to_owned(), (None, true))
```
davidbarsky, line 90: this worked, but it's not entirely clear to me as to why...

crates/rust-analyzer/src/reload.rs
```diff
@@ -217,6 +226,10 @@ impl GlobalState {
                 .filter_map(Result::ok)
                 .collect();
             let cargo_config = self.config.cargo();
+            let discover_command = self.config.discover_command();
+            let is_quiescent = !(self.discover_workspace_queue.op_in_progress()
+                || self.vfs_progress_config_version < self.vfs_config_version
+                || self.vfs_progress_n_done < self.vfs_progress_n_total);
```
davidbarsky, line 232: this part feels extremely hacky and I don't like it, but its seemingly necessary to avoid spawning another discover command when `fetch_workspaces` is repeatedly called.

crates/rust-analyzer/src/reload.rs
```diff
@@ -234,19 +247,40 @@ impl GlobalState {
...
+                                #[allow(clippy::needless_return)]
+                                return Ok(workspace);
+                            };
+
+                            if let Some(build) = it.crate_by_buildfile(path) {
+                                if is_quiescent {
+                                    let arg = DiscoverProjectParam::Label(build.label);
+                                    sender.send(Task::DiscoverLinkedProjects(arg)).unwrap();
+                                }
+                            }
```
davidbarsky, line 275: clippy is unhappy about this (as am I) but I _think_ fixing this requires moving the `discover_command` functionality to live in `ProjectWorkspace`, alongside some additional changes in `ProjectWorkspace` to avoid repeatedly calling `discover_command` and rely on some internally-cached `rust-project.json` when `GlobalState::fetch_workspaces` is called.

davidbarsky (commented):
@Veykril I commented on things I know are subpar/things I'd like to still change, but feel free to review now :)

crates/rust-analyzer/src/config.rs
```diff
@@ -2284,11 +2320,20 @@ mod single_or_array {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 #[serde(untagged)]
 enum ManifestOrProjectJson {
     Manifest(Utf8PathBuf),
-    ProjectJson(ProjectJsonData),
+    ProjectJson { data: ProjectJsonData, buildfile: Option<Utf8PathBuf> },
```
davidbarsky, line 2327: oh, i just realized that this change alone would break (de)serialization for callers to want to pass the `rust-project.json` inline.

crates/rust-analyzer/src/config.rs
davidbarsky: this is less gross now, but a conversion from an `Utf8PathBuf` to an `AbsPathBuf is still unfortunately necessary. Fixing that would be a larger refactor, but it's not one I'm opposed to making.

crates/rust-analyzer/src/main_loop.rs
davidbarsky: this is still gross, and i'm struggling to make this better with the current architecture with OpQueues—the state machine is pretty defuse across all of the rust-analyzer crate. I'll explain the pitfalls of the approach I took in a larger comment.

crates/rust-analyzer/src/main_loop.rs
davidbarsky: this is now driven by user-specified config.

crates/rust-analyzer/src/reload.rs
davidbarsky: this is gross, but it's a bit less gross now.

davidbarsky:
Alright, this is in a mostly fine state. Here's the situation:
- Any build system-specific integrations strings (Buck, Bazel) have been removed and are instead driven through configuration.
- The workspace loading/discovery mechanism was already pretty complex and introducing laziness to a system that _desperately_ wants to be eager only makes it more complex.

The latter bit is causing some subtle bugs that I'm struggling to fix. Here's the sequence of events that causes the main one that I'm struggling to fix with small checks:
1. The first event arrives: the user opens a Rust file. `GlobalState::handle_event` is called; the `textDocument/didOpen` notification is dispatched to the handler.
    1. The handler, seeing that lazy project discovery is enabled creates a deferred task to check whether this new file has been indexed by rust-analyzer (why deferred? to avoid panics; see https://github.com/rust-lang/rust-analyzer/issues/15837).
2. The remainder of `GlobalState::handle_event` runs. `self.fetch_workspaces_queue.should_start_op` is called, and if there's a value—which there is—it calls `self.fetch_workspaces`. 
  a. The first call to `self.fetch_workspaces` fails almost immediately because no workspace has been discovered. This results in a stray error being shown to the user/the status bar might briefly flash red in VS Code.
3. The next turn of `GlobalState::handle_event` and checks whether the opened text document is indexed due to step 1. rust-analyzer learns that the file isn't, and spawns the discover command.
4. the discover workspace command completes successfully and precedes with the the remainder of workspace crate loading.
5. Some time later, the user decides to change something in their settings. The setting change takes effect, but `GlobalState::fetch_workspace_error` reports an error, presumably due to failed the failed workspace fetch in step 2.
    1. I don’t know how to get out of this state, but rust-analyzer also works just fine: the only thing wrong is it reporting its internal state incorrectly?
    
Stepping back, `OpQueue` doesn't feel particularly great to work with—particularly when interlocking states are concerned—so I'm not sure if it makes sense to unify `fetch_workspaces_queue` with `discover_workspaces_queue` or replace them wholesale with more explicit state machine with transitions. I briefly tried unifying them, but the issue is that a subsequent request to the `OpQueue` can override a pending-but-not-yet-started request and the caller is required to maintain invariants.

Anyway, this isn't particularly organized, but this is a dump of my thinking/debugging.

davidbarsky:
Noticed a smaller error in that rust-analyzer assumes an error `source` when, strictly speaking, it doesn't need one.

davidbarsky:
After some discussion [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/Thinking.20about.20Lazy.20Project.20Discovery.20and.20OpQueues/near/447224346) and noodling around with different approaches today, I think the right approach to [resolve this issue in the short term](https://github.com/rust-lang/rust-analyzer/pull/17246#issuecomment-2185259122) would be to make the currently eager (that is, they occur on startup/workspace folder change) `ProjectManifest::discover_all` + `cargo metadata`-style operations a lazy, `rust-analyzer.workspace.DiscoverCommand`-style operation that only happens _after_ startup. This would mean two things:
- Project discovery/indexing wouldn't start until a user opens a Rust file.
- `cargo_metadata::MetadataCommand::new()` would become the default mode for `flycheck/src/json_workspace.rs`. That file would no longer be JSON workspace-specific, but in doing so, it would make "project discovery" a first-class concept in rust-analyzer.

The thing I'm currently thinking about is whether `GlobalState::workspaces` or `Config::discovered_projects` should be the source of truth, and I'm somewhat inclined to make `GlobalState::workspaces` the source of truth. I think to properly handle linkedProjects, discovery would be either be the default project discovery command or a closure that returns an inline ProjectJson (basically, a `() -> ProjectJson`).

crates/flycheck/src/json_workspace.rs
```diff
@@ -0,0 +1,102 @@
+//! A `cargo-metadata`-equivalent for non-Cargo build systems.
```
Veykril, line 1: Is there any reason for this being here? This might as well live in the main rust-analyzer crate

Veykril: Ah I see this deopends on the command handle stuff ... we should really rename this crate then

crates/flycheck/src/json_workspace.rs
Veykril: Why does tracing play into this at all? Tracing sounds very wrong here

crates/rust-analyzer/src/config.rs
```diff
@@ -1806,7 +1862,7 @@ impl Config {
     }
 
     pub fn cargo_autoreload_config(&self) -> bool {
```
Veykril, line 1864: An aside, this config confuses me a bit tbf, I don't see purpose of it

crates/rust-analyzer/src/config.rs
```diff
@@ -1565,6 +1613,14 @@ impl Config {
                             .ok()
                             .map(Into::into)
                     }
+                    ManifestOrProjectJson::DiscoveredProjectJson { data, buildfile } => {
+                        let root: &AbsPath =
+                            buildfile.as_path().try_into().expect("unable to get AbsPathBuf");
```
Veykril, line 1618: Feels like `DiscoveredProjectJson` should contain an `AbsPathBuf` instead, having the validation done somewhere up the call chain

davidbarsky: yes, but unfortunately, the `DiscoveredProjectJson` contains a `project_model::ProjectJsonData`, which currently needs (I think?) to have non-absolute paths so that we can emit SCIP Protobufs to [Glean](https://glean.software/docs/indexer/lsif-rust/) for code navigation in web browsers.

davidbarsky:
I couldn't answer https://github.com/rust-lang/rust-analyzer/pull/17246#discussion_r1660880157 inline, so:

> Why does tracing play into this at all? Tracing sounds very wrong here.

it's a _bit_ wrong, but I'm using `tracing` as a building block for a `cargo_metadata::Message` equivalent (see the third example on [it's documentation](https://docs.rs/cargo_metadata/latest/cargo_metadata/)). I think `tracing` should be an implementation detail here, with a more defined format that allows for communicating errors, progress, and successful `rust-project.json` generation. At the moment, it's, uh, a bit ad-hoc, but it shouldn't be.

crates/rust-analyzer/src/config.rs
davidbarsky: oh, i misunderstood the question. yeah, this can be an `AbsPathBuf`.

davidbarsky: just pushed a change that moved over to `AbsPathBuf` for the buildfile, but this required adding `#[serde(serialize_with = "..."]`/`#[serde(deserialize_with = "..."]` annotations to avoid making `AbsPath`/`AbsPathBuf` implement `Serialize`/`Deserialize`. I'd like to avoid implementing `Serialize`/`Deserialize` on `AbsPathBuf` because of this comment: https://github.com/rust-lang/rust-analyzer/blob/1b283db47f8de1412c851c92bb4ce4ef039ff8ff/crates/paths/Cargo.toml#L16-L18

I think hack could go away if/when `Config::linked_projects` isn't used as a source of truth anymore.

Veykril:
Ugh I forgot about this sorry
@bors delegate+

Merge after you rebase it once more (though do wait until monday's release just to be safe should this break anything accidentally)

davidbarsky:
> Ugh I forgot about this sorry
> 
> @bors delegate+
> 
> 
> 
> Merge after you rebase it once more (though do wait until monday's release just to be safe should this break anything accidentally)

No worries! I've got a four-day weekend, so I don't expect to open GitHub on my computer :). I'll add some proper serde deserialization that is more independent of `tracing`'s output and then merge this on Monday or Tuesday.

crates/rust-analyzer/src/reload.rs
```diff
@@ -244,23 +257,40 @@ impl GlobalState {
 
                 sender.send(Task::FetchWorkspace(ProjectWorkspaceProgress::Begin)).unwrap();
 
+                if let (Some(_command), Some(path)) = (&discover_command, &path) {
+                    let build = linked_projects.iter().find_map(|project| match project {
+                        LinkedProject::InlineJsonProject(it) => it.crate_by_buildfile(path),
+                        _ => None,
+                    });
```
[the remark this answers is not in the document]
davidbarsky: hmm, good points. For context, I'm trying to draw an analogue here between a Cargo.toml and a `BUCK`/`BUILD` file, with the latter being superset of a Cargo.toml.

> A BUCK file has two targets, and one of them is a leaf crate (//a:helpers) but will nevertheless claim to own a/BUCK and then load up zero useful crates. Should we load a Vec of labels instead?

Maybe. I designed this around rust-project's current behavior, which would find the buildfile corresponding to a given target and load _all_ Rust targets defined in said buildfile, so the hope is that we'd reload _all_ relevant targets in said buildfile (rust-analyzer can't be more granular than that, unfortunately.)

> Imagine A has a big dependency graph; we just asked RA to reindex it all twice, which can take maybe a minute on a medium sized codebase.

That does currently suck, but that'll be the thing I'll be working on fixing next: https://github.com/rust-lang/rust-analyzer/issues/17537

> Should reloading the workspace maybe load whatever was loaded last time by storing the argument passed to the discover command? 

Maybe, but that target might've gone away in the interim. I guess we still have this issue with the PR as-is, though.

crates/flycheck/src/project_json.rs
```diff
@@ -0,0 +1,155 @@
...
+    let message = r#"
+    {"kind": "error", "error":"failed to deserialize command output","source":"command"}
+    "#;
+
+    let message: DiscoverProjectData =
+        serde_json::from_str(message).expect("Unable to deserialize message");
+    assert!(matches!(message, DiscoverProjectData::Error { .. }));
+
+    let message = r#"
+    {"kind": "finished", "project": {"sysroot": "foo", "crates": [], "runnables": []}, "buildfile":"/Users/dbarsky/Developer/rust-analyzer"}
```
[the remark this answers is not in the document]
davidbarsky: >  Should buildfile be called root_dir instead?

No, I botched the copy-and-paste job for this test. It should be something like `Users/dbarsky/Developer/rust-analyzer/BUCK`.

> For reviewing, it flows through: [redacted steps]

This is correct, yes.

crates/flycheck/src/project_json.rs
```diff
@@ -0,0 +1,155 @@
...
+    command: Vec<String>,
+    sender: Sender<DiscoverProjectMessage>,
+}
+
+#[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum Arguments {
+    Path(
+        #[serde(serialize_with = "serialize_abs_pathbuf")]
+        #[serde(deserialize_with = "deserialize_abs_pathbuf")]
```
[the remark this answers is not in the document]
davidbarsky: > Don't actually need a Deserialize implementation for this type that only exists to be serialized.

Whoops, yup. Will remove.

> Also, since this it copy pasted below, maybe you can just implement the serde traits on AbsPathBuf itself?

I don't want to that because taking a dependency on serde in the crate where `AbsPathBuf` is defined (`paths`) slows down compilation because `serde_derive` now needs to be compiled *way* sooner.

crates/flycheck/src/project_json.rs
```diff
@@ -0,0 +1,155 @@
...
+///
+/// This is analogous to `cargo-metadata`, but for non-Cargo build systems.
+pub struct Discover {
+    command: Vec<String>,
+    sender: Sender<DiscoverProjectMessage>,
+}
+
+#[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum Arguments {
```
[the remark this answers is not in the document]
davidbarsky: yeah, i'll add those docs today. thanks!

(the tl;dr is that it assumes a somewhat expansive client like Buck's `rust-project` because the smarts can't live in rust-analyzer.)

crates/flycheck/src/project_json.rs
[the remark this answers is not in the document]
davidbarsky: @cormacrelf Sorry, I forgot to respond to this:

> Forcing RP to emit a buildfile seems a bit unnecessarily limiting, assumes you want one BUCK file at a time as your RA workspace and can somewhat haphazardly/heuristically add extra targets. What if I want to make RP support rust-project develop-json -t //... $JSON to make sure I don't end up with an important crate stuck as is_in_workspace: false? What should be the buildfile then?

I'm risking going off-topic about rust-project/Buck-specific concepts, but the change to rust-project would be to return a Vec of `DiscoverProjectData::Finished` instead of the current output. We'd know the set of related crates through by setting `with_rust_workspace` in a PACKAGE (not build!) file, which would behave like a Cargo workspace equivalent (note that we recently broke `with_rust_workspace`, so don't try to use it yet!).

Alternatively, you could have all your crates in a single `BUCK` file, but that loses the nice 1:1 correspondence between a Cargo.toml file and a `BUCK` file.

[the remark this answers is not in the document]
cormacrelf: What would RA even do with two separate json projects with about 600 dependencies in common?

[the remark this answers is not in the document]
davidbarsky: > What would RA even do with two separate json projects with about 600 dependencies in common?

It'll index them both and not perform and deduplication in terms of work or memory. This isn't great, but I think the correct fix is to make the crate graph an interned Salsa DB, which I'll be doing next. The state you're describing will likely be possible for a little while.

crates/rust-analyzer/src/reload.rs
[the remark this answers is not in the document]
davidbarsky: I switched it over. It is now:

```
#[derive(PartialEq, Clone, Debug, Serialize)]
#[serde(rename_all = "camelCase")]
pub enum DiscoverArgument {
    Path(#[serde(serialize_with = "serialize_abs_pathbuf")] AbsPathBuf),
    Buildfile(#[serde(serialize_with = "serialize_abs_pathbuf")] AbsPathBuf),
}
```

Not the biggest fan of the name as-is, but if I can't think of anything better, I'll land it today and make changes in subsequent PRs.

crates/flycheck/src/project_json.rs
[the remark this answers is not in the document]
cormacrelf: My instinct here is to leverage Buck a bit more, and let it do the deduplication. It's really good at that. rust-project already supports passing in multiple target expressions and that's the main way I use it. So I would want RA to be able to drive it that way, and if it can't, I'll probably end up submitting/carrying a patch that allows overriding the targets for the develop-json subcommand, to ignore what RA says it's interested in, but still use the automatic discovery feature to reload.

A good interface for users to describe by their behaviour which targets to root the graph on is hard to pin down, and I can see why it'll take more than this PR to figure it out. But it doesn't matter too much to me because I have the luxury of a small enough repo where //... is almost usable.

crates/rust-analyzer/src/reload.rs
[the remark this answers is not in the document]
cormacrelf: Lgtm

crates/flycheck/src/project_json.rs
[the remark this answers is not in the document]
davidbarsky: > So I would want RA to be able to drive it that way, and if it can't, I'll probably end up submitting/carrying a patch that allows overriding the targets for the develop-json subcommand, to ignore what RA says it's interested in, but still use the automatic discovery feature to reload.

At least on the rust-analyzer side of things, I _think_ that it should be pretty easy to support `//...` if `{arg}` is no longer required—rust-analyzer will still pass around the path to buildfile or Rust source file, but it'll never actually be _used_ if `{arg}` isn't, well, used. The tricky implementation bit would be on the `rust-project` side of things, but if I'm not mistaken, a regular `rust-project develop --stdout //...` might also _just work_ (but without any of the progress reporting).

I can imagine an alternative design where `{arg}` is still required, but it allows specifying literals like `{arg: '//...'}`, which would be used unconditionally, which would the behavior a bit more specified.

You know, I'll think I'll remove the hard requirement of specifying `{arg}` for now.

PR #17250, opened 2024-05-17
internal: Correct Neovim 0.10 inlay hints config example

adaszko:
This change is what I had to do to make inlay hints work on Neovim 0.10.  The current example produces errors about wrong argument type to `.enable()`.


davidbarsky (approved):
Thanks!

lnicola (approved):

PR #17251, opened 2024-05-17
fix: resolve extern prelude for local mods in block modules

roife:
fix https://github.com/rust-lang/rust-analyzer/issues/17057, https://github.com/rust-lang/rust-analyzer/issues/17032.

We should use `ModuleOrigin` to check if the current module is a pseudo-module introduced by blocks (where names might be shadowed), rather than checking `block_def_map`.

### DIFF
--- a/crates/hir-def/src/nameres/path_resolution.rs
+++ b/crates/hir-def/src/nameres/path_resolution.rs
@@ -18,9 +18,7 @@ use crate::{
     db::DefDatabase,
     item_scope::{ImportOrExternCrate, BUILTIN_SCOPE},
     item_tree::Fields,
-    nameres::{
-        sub_namespace_match, BlockInfo, BuiltinShadowMode, DefMap, MacroSubNs, ModuleOrigin,
-    },
+    nameres::{sub_namespace_match, BlockInfo, BuiltinShadowMode, DefMap, MacroSubNs},
     path::{ModPath, PathKind},
     per_ns::PerNs,
     visibility::{RawVisibility, Visibility},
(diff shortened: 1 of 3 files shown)


roife:
I'm not sure how to add unittests for it. 🤔

crates/hir-def/src/nameres/path_resolution.rs
```diff
@@ -470,9 +472,9 @@ impl DefMap {
         };
 
         let extern_prelude = || {
-            if self.block.is_some() {
-                // Don't resolve extern prelude in block `DefMap`s, defer it to the crate def map so
-                // that blocks can properly shadow them
+            if matches!(self[module].origin, ModuleOrigin::BlockExpr { .. }) {
```
Veykril, line 475:
```suggestion
            if self.block.is_some() && module == DefMap::ROOT {
```

If I understand the code here correctly, likjewise below. Though I am unsure whether this is the correct fix here

Veykril: Though it should be more correct than the status quo

roife: They are equivalent. If a module is a `BlockExpr`, its `local_id` should be equivalent to `ROOT`, and it should be within a block; conversely, this is also true. Should I change it to the latter?

Veykril: Yes, sorry I wasn't clear. With unsure whether this is the correct fix I meant the general approach here. My code suggestion is the more correct way to do your check here (they are equivalent, it's just that the origin shouldn't be used to check this property).

crates/hir-def/src/nameres/path_resolution.rs
```diff
@@ -515,8 +518,8 @@ impl DefMap {
             None => self[Self::ROOT].scope.get(name),
         };
         let from_extern_prelude = || {
-            if self.block.is_some() {
-                // Don't resolve extern prelude in block `DefMap`s.
+            if matches!(self[module].origin, ModuleOrigin::BlockExpr { .. }) {
+                // Don't resolve extern prelude in pseudo-module of a block.
```
Veykril, line 522: Same thing here

Veykril:
@bors delegate+
r=me with the review addressed

PR #17252, opened 2024-05-17
internal: refactor `prefer_no_std`/`prefer_prelude` bools into a struct

davidbarsky:
I noticed that there's a large number of functions/arguments during an unrelated change that take two booleans and realized they're _probably_ better off being in a single struct—less error-prone, etc.

Feel free to suggest a better name than `ImportPathConfig`/close this entirely! I can also make these args enums; just hopefully making this a little more misuse-resistant.


Veykril:
Rebased it given my PR had a bunch of conflicts touching the same parts :) The new `ignore_local_imports: bool,` remains now though, as its private API and shouldn't belong to that new struct (though that means its also not exposed like the other two were so that should be fine)
@bors r+

Veykril:
small rebase mishap
@bors r+

PR #17254, opened 2024-05-18
Sync from rust

lnicola:
Using `josh`, this time.

Fails to build right now because we need to upgrade to a newer version of `rustc_abi`. It should have come out last Sunday, but the workflow didn't run.


lnicola:
https://github.com/rust-lang/rust-analyzer/pull/17152 only shows 250 commits (as opposed to 1742 here), but that's just a display issue.

Veykril:
> It should have come out last Sunday, but the workflow didn't run.

The workflow got disabled due to inactivity again, I re-enabled it now though feel free to trigger a manual publish there if you don't want to wait. I'm not planning on merging anything this week anymore. (though looking at it we don't have manual ones setup in the CI)

lnicola:
This needs to wait for 1.79 because `rustc-abi` is now using the new `NonZero`.

Veykril:
Meh, that's annoying, guess we should've realized that and called it out since we did get pinged on that PR due to something else.

Will you need to make a new sync PR if we merge anything before merging this? Assuming no conflicts happen or is that gonna cause issues?

Veykril:
Otherwise I'm sure upstream would be fine to undo that change for that crate until 1.79 releases

lnicola:
> Will you need to make a new sync PR if we merge anything before merging this?

It was easy to make, I can just make a new one next month. The only issue is that we can't update the rustup version.

> Otherwise I'm sure upstream would be fine to undo that change for that crate until 1.79 releases

Do you think so? I'm not sure.

Veykril:
> It was easy to make, I can just make a new one next month. 

I take that as the josh debugging has finally worked out then? 🎉 

> The only issue is that we can't update the rustup version.

Shouldn't be too problematic, I don't think any critical bug is in the current rustup version that is fixed right now, nor are there any interesting proc-macro server things outstanding iirc.

> Do you think so? I'm not sure.

It's very minor, I'd be surprised if they said no.

lnicola:
> I take that as the josh debugging has finally worked out then? 🎉

Hopefully? But only thanks to RalfJung :-).

> It's very minor, I'd be surprised if they said no.

I filed https://github.com/rust-lang/rust/pull/125240, but I suspect it's going to be deemed as too much churn.

lnicola:
To be fair, `rustc_pattern_analysis` will also need to be bumped. There are some breaking changes there, CC @Nadrieril :grimacing:.

Issue #16195, opened 2023-12-25
When generating implementation for a trait, emit `async fn` for functions that return `impl Future (+ Send)`


GoldsteinE:
With AFITs + RPITITs being stabilized in 1.75 without RTN, the following pattern will probably become common:

```rust
trait AlwaysSend {
    fn foo(&self) -> impl Future<Output = ()> + Send;
}

impl AlwaysSend for SomeType {
    // We can use `async fn` syntax here, yay!
    async fn foo(&self) { /* ... */ }
}
```

Unfortunately, the “fill missing methods” assist will not generate a nice clean `async fn` in this case, instead simply copying `fn() -> impl Future<Output = ()> + Send`.

Note that it doesn’t seem to apply to traits generated with `trait-variant`, but it’s still unfortunate to have to use procmacros to get a nice IDE experience.

Possibly related to #11315

GoldsteinE:
Glancing at the relevant code, this actually seems doable by applying one more transform to the methods:

https://github.com/rust-lang/rust-analyzer/blob/85fb463fc586594925f05fc8e285b1568f98f41a/crates/ide-assists/src/utils.rs#L182-L188

I’ll try to look into it later this week.

ryo33:
As it might help someone who is interested in this feature as like me, I'd like to leave a quick fix to workaround the functionality.
In my Neovim this works: `%s/fn \([^(]*\)(\([a-zA-Z& {}:\n,<>']*\)) -> impl [a-z:]*Future<Output = \(.*\)>[+ A-Za-z<>:,'\n]* {/async fn \1(\2) -> \3 {/`
It can be done in any editor that supports regex search and replacement, including VSCode, or external replacement tools like sed etc.

PR #17258, opened 2024-05-18
Add `toggle_async_sugar` assist code action

### LINKED ISSUE #17010: Code action to toggle `async fn` and `impl Future`
It's better to have a code action that toggles a function or trait method between `async fn method() -> Ret` and `fn method() -> impl Future<Output = Ret>`. Conversion from `async fn` to `impl Future` should be lossless, but, of course, the opposite is not.

It's convenient in cases like:
- like to convert an existing `-> impl Future` function to `async fn` especially it was generated (e.g. https://github.com/rust-lang/rust-analyzer/issues/16195)
- like to add `Send` bound to an existing `async fn` to assert the implementation is Send.

maxwase:
Implement code action for sugaring and de-sugaring asynchronous functions.

This code action does not import `Future` trait when de-sugaring and does not touch function boby, I guess this can be implemented later if needed. This action also does not take into consideration other bounds because IMO it's usually "let me try to use sugared version here".

Feel free to request changes, that's my first code action implementation 😄 

Closes #17010 
Relates to #16195


crates/ide-assists/src/handlers/toggle_async_sugar.rs
```diff
@@ -0,0 +1,460 @@
...
+// }
+// ```
+// ->
+// ```
+// pub fn foo() -> impl Future<Output = usize> {
+//     0
+// }
+// ```
+pub(crate) fn toggle_async_sugar(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
+    let function: ast::Fn = ctx.find_node_at_offset()?;
```
Veykril, line 30: Let's instead look for an `async` token or if that fails for a return type node, and then check if the parent of either is a function and continue as written. Reason being we don't want to show this assist whenever the cursor is in any part of the function as that clutters things.

crates/ide-assists/src/handlers/toggle_async_sugar.rs
```diff
@@ -0,0 +1,460 @@
...
+    }
+    let future_output = unwrap_future_output(main_trait_path)?;
+
+    acc.add(
+        AssistId("toggle_async_sugar", AssistKind::RefactorRewrite),
+        "Convert `impl Future` into async",
+        function.syntax().text_range(),
+        |builder| {
+            match future_output {
+                ast::Type::TupleType(_) => {
```
Veykril, line 74: This ought to check that the tuple is empty, otherwise we delete other tuples than just the unit type

crates/ide-assists/src/handlers/toggle_async_sugar.rs
```diff
@@ -0,0 +1,460 @@
...
+                async_range = TextRange::new(async_range.start(), whitespace_range.end());
+            }
+            builder.delete(async_range);
+
+            match return_type {
+                Some(ret_type) => builder.replace(
+                    ret_type.syntax().text_range(),
+                    format!("impl Future<Output = {ret_type}>"),
+                ),
+                None => builder.insert(rparen.text_range().end(), " -> impl Future<Output = ()>"),
```
Veykril, line 139: Instead of insertin a bare `Future` or importing the `Future` trait I think it'd be best to qualify the `Future` path. Then the user can invoke an assist to turn that into an import if they'd like. You can do that via ```rs

let scope = ctx.sema.scope(return_type.syntax())?;
let module = scope.module();
let future_trait = FamousDefs(&ctx.sema, scope.krate()).core_future_Future()?;
let trait_path = module.find_path(
    ctx.db(),
    ModuleDef::Trait(future_trait),
    ImportPathConfig {
        prefer_no_std: ctx.config.prefer_no_std,
        prefer_prelude: ctx.config.prefer_prelude,
    },
)?;
trait_path.display(db) // use this in the format! invocation
```

File: crates/ide-assists/src/handlers/toggle_async_sugar.rs [hunk 4]
```diff
@@ -0,0 +1,460 @@
...
+    famous_defs::FamousDefs,
+};
+use syntax::{
+    ast::{self, HasVisibility},
+    AstNode, NodeOrToken, SyntaxKind, SyntaxNode, SyntaxToken, TextRange,
+};
+
+use crate::{AssistContext, Assists};
+
+// Assist: toggle_async_sugar
```
Veykril, line 12: I'd prefer if we split the assist into two distinct labeled ones (still keep them in the same module here as they are short enough). `desugar_async_function` and `resugar_async_function` or similar, and then break apart the `toggle_async_sugar` function into two entry points that are called in the `lib.rs` file

crates/ide-assists/src/handlers/toggle_async_sugar.rs
maxwase: Changed, perhaps `unnecessary_async` assist should also be restricted

crates/ide-assists/src/handlers/toggle_async_sugar.rs
maxwase: Thanks for great suggestion, I was looking for something like this! I've changed the `scope` from a `return_type` to a `function` because `return_type` is an `Option`

crates/ide-assists/src/handlers/toggle_async_sugar.rs
maxwase: Changed, now it's `desugar_async_into_impl_future` and `sugar_impl_future_into_async`, other names were a bit ambiguous to me

Veykril (approved):

Veykril:
Thanks!
@bors r+

Issue #11039, opened 2021-12-17
Quick fix to declare an unresolved lifetime


matklad:
With inband lifetimes going away (https://github.com/rust-lang/rust/issues/44524), I believe we should just fix the ergonomics at the ide layer.

Typing `fn foo(&'a str)` should emit an error for unresolved `'a`, and a quick fix to add a lifetime.

See how `NoSuchField` diagnostics is treated, which has a similar flow with a fix: https://github.com/rust-analyzer/rust-analyzer/search?q=NoSuchField.

@Veykril I believe you did name resolution for lifetimes, could you add a code link for where it is happening?

Veykril:
Since we don't use lifetimes in type inference and such yet we only actually resolve them in the ide layer(through Semantics).

So I guess we should first actually lower/resolve them in type inference(without using them afterwards, we can just keep the static lifetime hack) and then emit diagnostics from there which I think should happen somewhere around here? https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/hir_ty/src/lower.rs#L162-L351

PR #17265, opened 2024-05-20
Implement assist to add missing lifetime to function

ThouCheese:
To address #11039

### DIFF
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -252,6 +252,10 @@ pub enum InferenceDiagnostic {
         expr: ExprId,
         expected: Ty,
     },
+    FunctionMissingLifetime {
+        function: ExprId,
+        lifetime: LifetimeRef,
+    },
 }
 
 /// A mismatch between an expected and an inferred type.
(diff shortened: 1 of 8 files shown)


Veykril:
This should ideally be a quickfix for an unresolved lifetime diagnostic. The origin for the diagnostic to emit would be here https://github.com/rust-lang/rust-analyzer/blob/16fd9aa373eeb36b1203bab1fde9cf3b9320a424/crates/hir-ty/src/lower.rs#L1380, `TyLoweringContext` has no infrastructure for reporting diagnostics yet though so we'd need to add that as well, best would probably be via callback (`Option<&'lt mut dyn FnMut(DiagnosticType)>`) that is stored in it and that can be set by the caller when creating a `TyLoweringContext`. 



Ultimately it should make its way into the main diagnostics here https://github.com/rust-lang/rust-analyzer/blob/16fd9aa373eeb36b1203bab1fde9cf3b9320a424/crates/hir-ty/src/infer.rs#L421 and then be processed like the rest in there.

Veykril:
I believe https://github.com/rust-lang/rust-analyzer/pull/18074 will make this easier and less messy (no callbacks), so I'll put this into draft until then

PR #17268, opened 2024-05-21
feat: More callable info

Veykril:
With this PR we retain more info about callables other than functions, allowing for closure parameter type inlay hints to be linkable as well as better signature help around closures and `Fn*` implementors.


crates/hir/src/lib.rs
```diff
@@ -2167,17 +2166,24 @@ impl From<hir_ty::Mutability> for Access {
...
-    pub fn parent_fn(&self) -> Function {
-        self.func
+    pub fn parent_fn(&self) -> Option<Function> {
+        match self.func {
+            Callee::Def(CallableDefId::FunctionId(f)) => Some(f.into()),
+            _ => None,
+        }
     }
 
+    // pub fn parent_closure(&self) -> Option<Closure> {
```
lnicola, line 2183: Leftover code?

Veykril: Ah, yes

PR #15993, opened 2023-11-30
Debug use cargo workspace root as `cwd`

meowtec:
fixes #13022

Issue #13022, opened 2022-08-15
VSCode "Debug" test cwd is workspace root, not package root for tests


Crzyrndm:
Setup: https://github.com/Crzyrndm/ra-executing-directory

A package nested in a workspace, the tests (unit + integration) just print `std::env::current_dir()` and then panic (so that the print is always visible)

* `cargo test --no-fail-fast`: prints the package root directory for both unit and integration no matter what directory it is run in
* VSCode "Run test" button: prints the package root directory
* VSCode "Debug" button: prints the _workspace_ root directory

This makes debugging tests that need to load a data file quite awkward

![image](https://user-images.githubusercontent.com/8765278/184589044-c2e34101-4312-4f33-9ab5-3d36c2e0df29.png)

**rust-analyzer version**: rust-lang.rust-analyzer 0.3.1170 (current stable)

**rustc version**: rustc 1.63.0 (4b91a6ea7 2022-08-08)

Tested OS is Windows 10

**relevant settings**:

Crzyrndm:
Disclaimer: I have never used JS/TS so take the following as you will

From what I can tell, the Debug executing directory is set [here](https://github.com/rust-lang/rust-analyzer/blob/c567771605631c786af8b3d21204abad857e04ac/editors/code/src/run.ts#L142) and is unconditionally the workspace/open folder root directory.

This is correct for the basic case (`cargo run -p <package> ...`) but it looks like cargo is localising test execution to the package (which makes sense. Don't won't test pass/fail depending on whether it's in a workspace or not). I haven't tested examples / benchmarks to see if similar package localisation is happening but I can see that being a likely default as well

The [target](https://github.com/rust-lang/rust-analyzer/blob/c567771605631c786af8b3d21204abad857e04ac/editors/code/src/run.ts#L148) appears to be known in the same scope although I don't know if that's package/folder/something else at this point. Presumably whether it's a run/test(/example/benchmark) or not can be obtained from [this](https://github.com/rust-lang/rust-analyzer/blob/c567771605631c786af8b3d21204abad857e04ac/editors/code/src/run.ts#L140)

esiegel:
In addition it doesn't include any of the cargo environment variables that could be used to help normalize a path.  For example, `CARGO_MANIFEST_DIR` is missing

PR #17275, opened 2024-05-22
Fix inconsistent cwd of `run` and `debug` command in client

### LINKED ISSUE #17012: Run and Debug codelens link buttons for main (non test) entry point launch with differing cwd paths
Observed behaviour: In the scenario of an bin package inside a virtual workspace, such as in this minimal example (https://github.com/Keshi/minimal-virtual-workspace), choosing to Run or Debug the main function using the codelens buttons results in different working directories.
If the application is one that expects to dynamically load a settings or asset file relative to the cwd, then the application will always fail to launch through one of either Run or Debug.

Expected/Desired behaviour: Run and Debug codelens buttons both launch the application using the same cwd behaviour (as observed by the application) as Run currently does.

**rust-analyzer version**: rust-analyzer version: 0.3.1906-standalone (e4a405f87 2024-03-31)
**rustc version**: rustc 1.77.1 (7cf61ebde 2024-03-27)
**editor or extension**: VSCode, rust-analyzer 0.3.1906

**repository link (if public, optional)**: 
https://github.com/Keshi/minimal-virtual-workspace

roife:
Fix #17012. Also related to #13022 and #15993.

When the `kind` of runnable is `bin`, Cargo would use the workspace root as the cwd for the `run` command; otherwise, Cargo defaults to the package root as the cwd for `run`.

Initially, r-a assumed the workspace root as the cwd for all runnables in `debug` command, which led to issue `#13022`. In this case, during unit testing, the `run` command would use the package root while `debug` would use the workspace root, causing inconsistency.

PR `#15993` addressed this problem by using the package root as the cwd for `debug` command. However, it also resulted in an inconsistency: when executing the `run` command within the main fn of a package (whose target is `bin`), Cargo would use the workspace root, whereas `debug` would use the package root, leading to issue `#17012`.

The preferable approach is to determine the cwd based on the runnable's type. To resolve this, this PR introduces a new `cwd` field within `CargoRunnable`, allowing r-a to decide the appropriate cwd depending on the specific kind of the runnable.


Veykril (approved):
Sounds reasonable to me, cc @davidbarsky since you've been dabbling around this area wrt to rust-project.json, probably not too relevant but given https://github.com/rust-lang/rust-analyzer/pull/16840 is in a similar-ish vein

davidbarsky:
yeah, this seems reasonable to me, I think.

davidbarsky (approved):

PR #17277, opened 2024-05-22
fix: Various find path fixes

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/17271

### DIFF
--- a/crates/hir-def/src/find_path.rs
+++ b/crates/hir-def/src/find_path.rs
@@ -596,7 +596,10 @@ fn find_local_import_locations(
 
 #[cfg(test)]
 mod tests {
+    use expect_test::{expect, Expect};
     use hir_expand::db::ExpandDatabase;
+    use itertools::Itertools;
+    use stdx::format_to;
     use syntax::ast::AstNode;
     use test_fixture::WithFixture;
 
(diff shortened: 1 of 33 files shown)


crates/hir-def/src/find_path.rs
```diff
@@ -70,45 +67,41 @@ pub enum PrefixKind {
...
     prefer_prelude: bool,
+    ignore_local_imports: bool,
 }
 
 /// Attempts to find a path to refer to the given `item` visible from the `from` ModuleId
 fn find_path_inner(ctx: FindPathCtx<'_>, item: ItemInNs, from: ModuleId) -> Option<ModPath> {
     // - if the item is a builtin, it's in scope
     if let ItemInNs::Types(ModuleDefId::BuiltinType(builtin)) = item {
-        return Some(ModPath::from_segments(PathKind::Plain, Some(builtin.as_name())));
+        return Some(ModPath::from_segments(PathKind::Plain, once(builtin.as_name())));
```
davidbarsky, line 100: the tiniest of nits: do an absolute `std::iter::once` instead of importing it?

crates/hir-def/src/find_path.rs
```diff
@@ -70,45 +67,41 @@ pub enum PrefixKind {
...
 }
 
 #[derive(Copy, Clone)]
 struct FindPathCtx<'db> {
     db: &'db dyn DefDatabase,
-    prefixed: Option<PrefixKind>,
+    prefix: PrefixKind,
     prefer_no_std: bool,
     prefer_prelude: bool,
+    ignore_local_imports: bool,
```
davidbarsky, line 93: i suspect that https://github.com/rust-lang/rust-analyzer/pull/17252 might help here: i found myself getting confused by the booleans in other contexts 😅

davidbarsky (approved):

crates/hir-def/src/find_path.rs
Veykril: Right, guess I'll fix up the conflicts after merging this then 😬

PR #17282, opened 2024-05-22
Feat: hide double underscored symbols from symbol search

### LINKED ISSUE #17272: Hide unhygenic tokens generated by macros from symbol search
Due to lifetime issues, our macros need to generate a variety of intermediary structs/statics. These show up in symbol search, which clutters up the search significantly. These are technically unhygenic.

I would be happy to fix this with a little bit of guidance.

Screenshot 2024-05-21 at 5 20 43 PM

jkelleyrtp:
Fixes #17272 by changing the default behavior of query to skip results that start with `__` (two underscores).

Not sure if this has any far reaching implications - a review would help to understand if this is the right place to do the filtering, and if it's fine to do it by default on the query.

If you type `__` as your search, then we'll show the matching double unders, just in case you actually need the symbol.

### DIFF
--- a/crates/ide-db/src/symbol_index.rs
+++ b/crates/ide-db/src/symbol_index.rs
@@ -53,7 +53,6 @@ pub struct Query {
     case_sensitive: bool,
     only_types: bool,
     libs: bool,
-    include_hidden: bool,
 }
 
 impl Query {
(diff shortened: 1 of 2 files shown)


crates/ide-db/src/symbol_index.rs
```diff
@@ -66,9 +67,14 @@ impl Query {
             mode: SearchMode::Fuzzy,
             assoc_mode: AssocSearchMode::Include,
             case_sensitive: false,
+            include_hidden: false,
```
Veykril, line 70: I don't think we have a need for this config here, it is implicitly configured by prefixing `__` to your query

Veykril:
Fixed the review myself (going through a PR cleanup), hope you don't mind :) Thanks!
@bors r+

Veykril:
Wrong button
@bors r+

PR #17295, opened 2024-05-25
Use correct format for setting environment variables when debugging with cpptools

0xJonas:
The RA VSCode extension uses an incorrect format for the environment variables in the `launch.json` when debugging with the C/C++ Extension. This extension uses a different format than CodeLLDB or NativeDebug, which means that the environment variables are not actually set for the debuggee.

What it currently looks like:
```json
"env": {
  "NAME": "VALUE"
}
```

What the C/C++ extension expects:
```json
"environment": [
  { "name": "NAME", "value": "VALUE" }
]
```

For reference: https://code.visualstudio.com/docs/cpp/launch-json-reference#_environment

### DIFF
--- a/editors/code/src/debug.ts
+++ b/editors/code/src/debug.ts
@@ -195,7 +195,7 @@ function getCCppDebugConfig(
         cwd: runnable.args.cwd || runnable.args.workspaceRoot || ".",
         sourceFileMap,
         environment: Object.entries(env).map(entry => {
-            return { "name": entry[0], "value": entry[1] }
+            return { "name": entry[0], "value": entry[1] };
         }),
         // See https://github.com/rust-lang/rust-analyzer/issues/16901#issuecomment-2024486941
         osx: {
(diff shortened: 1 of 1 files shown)


editors/code/src/debug.ts
```diff
@@ -194,7 +194,9 @@ function getCCppDebugConfig(
         args: runnable.args.executableArgs,
         cwd: runnable.args.cwd || runnable.args.workspaceRoot || ".",
         sourceFileMap,
-        env,
+        environment: Object.entries(env).map((entry) => {
+            return { name: entry[0], value: entry[1] };
+        }),
```
mathew-horner, line 199: nit: `return` shouldn't be needed here.
```suggestion
        environment: Object.entries(env).map((entry) => ({
            name: entry[0], value: entry[1]
        })),
```
Not sure what the proper formatting is.

0xJonas: You're right, I forgot you could just put parens around the object. Thanks!

Veykril:
Thanks!
@bors r+

PR #17296, opened 2024-05-25
Avoid clone when constructing runnable label.

mathew-horner:
I stumbled across this when reading this code. This seems like an unnecessary allocation (though likely small?)


davidbarsky (approved):
still a good idea to fix :)

PR #17297, opened 2024-05-26
Feature: Add IDE assist for to move ADT definitions + impls to separate files

kitlangton:
I've added a new IDE assist for moving the selected struct or enum, along with all of its impls, to a separate file as a new module. 

I've relied on similar refactorings when using IDEs for Scala. It's quite nice to be able to define new structs in the current file, allowing one to experiment more rapidly without getting bogged down by module organization.

I've added a good number of tests here, ensuring that it handles various module hierarchies correctly. It also attempts to preserve the visibility of the initial ADT definition, so that this operation doesn't break compilation.

### Example

**Before**
```rust
// main.rs
struct Foo {
    x: i32,
}

impl Foo {
    fn new(x: i32) -> Self {
        Self { x }
    }
}
```

**After**
```rust
// main.rs
mod foo;
use foo::*;

// foo.rs
use super::*;
struct Foo {
    x: i32,
}

impl Foo {
    fn new(x: i32) -> Self {
        Self { x }
    }
}
```

### TODOs

- [x] Copy all necessary `use` statements into the new module, as to not break compilation.
  - I've simply added `use super::*`. Which works, but might be too heavy handed.
- [x] Fix windows/linux tests

#### Note

In my excitement, I hadn't read `CONTRIBUTING.md` until just now, which recommends opening an issue before a PR 😅. Alas, as I've already written this code, I might as well share it. However, I will remain perfectly happy if this is rejected outright. It was a lovely learning experience. Thank you for your consideration and for building such a wonderful tool! 🥰

### DIFF
--- a/crates/ide-assists/src/handlers/move_definition_to_file.rs
+++ b/crates/ide-assists/src/handlers/move_definition_to_file.rs
@@ -1,44 +1,14 @@
 use std::iter;
 
+use crate::{AssistContext, AssistId, AssistKind, Assists};
 use hir::HasAttrs;
 use ide_db::base_db::AnchoredPathBuf;
-use itertools::Itertools;
-use stdx::{format_to, to_lower_snake_case};
+use std::path::PathBuf;
+use stdx::to_lower_snake_case;
 use syntax::{
     ast::{self, HasName, HasVisibility},
     AstNode, SmolStr,
 };
-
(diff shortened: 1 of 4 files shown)


Veykril:
We already have an assist (combo) for this 😅 You can select stuff in a file, invoke `Extract Module` and then invoke `Extract Module to file` on the new module, so I'll close this, sorry, but thanks for trying to improve things anyways!

PR #17308, opened 2024-05-29
Add preference modifier for workspace-local crates when using auto import.

mathew-horner:
@joshka pointed out some odd behavior of auto import ordering. It doesn't seem that the current heuristics were applying any sort of precedence to imports from the workspace. I've went ahead and added that.

I hope to get some feedback on the modifier numbers here. I just went with something that felt like it balanced giving more power to workspace crates without completely ignoring relative path distance.

closes https://github.com/rust-lang/rust-analyzer/issues/17303

### DIFF
--- a/crates/ide-assists/src/handlers/auto_import.rs
+++ b/crates/ide-assists/src/handlers/auto_import.rs
@@ -272,8 +272,10 @@ fn module_distance_heuristic(db: &dyn HirDatabase, current: &Module, item: &Modu
     // cost of importing from another crate
     let crate_boundary_cost = if current.krate() == item.krate() {
         0
-    } else if item.krate().is_builtin(db) {
+    } else if item.krate().origin(db).is_local() {
         2
+    } else if item.krate().is_builtin(db) {
+        3
     } else {
         4
     };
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/auto_import.rs
```diff
@@ -272,8 +272,10 @@ fn module_distance_heuristic(db: &dyn HirDatabase, current: &Module, item: &Modu
     // cost of importing from another crate
     let crate_boundary_cost = if current.krate() == item.krate() {
         0
-    } else if item.krate().is_builtin(db) {
+    } else if item.krate().origin(db).is_local() {
```
joshka, line 275: Does this pick up examples / binaries defined in the crate?

davidbarsky: it should. `is_local` generally means “everything that can reasonably be edited by the user/isn’t a crate downloaded from crates.io”.

mathew-horner: > Does this pick up examples / binaries defined in the crate?

Are you asking if it picks up _items that are defined in_ examples / binaries from library code? If so, correct me if I'm wrong, but I don't think you can do that in Rust?

If you mean the opposite, yes this does that. `is_local` specifically checks if the crate origin is local to the workspace.

joshka: I meant the opposite (picking up library defined types from the example code)

Veykril: examples, binaries, library and integration tests of a package are all different crates, so the `==` condition above will fail for those. `is_local` is about where the package is coming from, local being its a workspace member (opposed to coming from crates-io)

Veykril: Checking whether things are from the same package can't really be done (we discard that informaton). We could specifically try to rank better by checking if the item crate is in the direct dependencies of the current crate though, that will make things work fine for examples and the like

Veykril:
I'll merge this then, my review could be addressed as a follow up improvement if anyone wants to, thanks!
@bors r+

PR #17312, opened 2024-05-29
Add `Function::fn_ptr_type(…)` for obtaining name-erased function type

regexident:
cc @flodiebold

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -1884,6 +1884,14 @@ impl Function {
         Type::from_value_def(db, self.id)
     }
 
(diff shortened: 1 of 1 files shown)


hamirmahal (commented):
Is it feasible to add unit tests and documentation for `fn_ptr_type`?

They would give people new to the codebase more context on how to use it.

regexident:
Definitely! Could you point me towards the appropriate place to add a test for this?

Veykril:
The HIR api is generally not directly tested and nothing in our code makes use of this so I don't see a place to add a test here.

PR #17325, opened 2024-06-01
Allow expanding inert attributes to aid IDE features

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/11929

We can't unfortunately turn all inert attribute calls into actual macro calls, as that results in 100mb+ additional memory usage 😓 but only a select few are worth it anyways


Veykril:
Closing this for now, as this requires more invasive changes than I'd like right now

PR #17329, opened 2024-06-02
Don't mark `#[rustc_deprecated_safe_2024]` functions as unsafe

Nilstrieb:
`std::env::set_var` will be unsafe in edition 2024, but not before it. I couldn't quite figure out how to check for the span properly, so for now we just turn the false positives into false negatives, which are less bad.


Veykril:
> I couldn't quite figure out how to check for the span properly

We don't carry hygiene around right now (aside from `$crate` resolution)

@bors r+

PR #17346, opened 2024-06-05
Changed package.json so vscode extension settings have submenus

ChosenName:
There are a lot of options that are a part of rust-analyzer, sometimes it can be hard to find an option that you are looking for. To fix this I have put all configurations into categories based on their names. I have also changed the schema in `crates/rust-analyzer/src/config.rs` to reflect this.

Currently for each generated entry the title is redeclared, this does function but I am prepared to change this if it is a problem.

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -3080,8 +3080,10 @@ mod tests {
         let package_json_path = project_root().join("editors/code/package.json");
         let mut package_json = fs::read_to_string(&package_json_path).unwrap();
 
-        let start_marker = "            { \"title\": \"$generated-start\" },\n";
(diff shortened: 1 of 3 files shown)


Veykril:
Sounds reasonable, we can tweak this afterwards if need be, thanks!
@bors r+

mibes404:
@ChosenName this is really good. 

Can these categories be shown and hidden dynamically? 

Right now, user-oriented high-level settings like inlayHints are mixed in with highly technical low-level settings like lru capacity. It would be nice to have a drop-down menu to select a typical user profile, which would then show settings depending on the user profile: basic, power, or advanced. That way, the amount of settings can be limited to what a typical user in that category needs.

PR #17352, opened 2024-06-06
fix: do not resolve prelude within block modules

### LINKED ISSUE #17338: Type inference for `Result` wrong for inner functions.
**rust-analyzer version**: v0.3.1975

**editor or extension**: VSCode 1.89.1 (system setup)

```rust
// src/error.rs

pub enum Error {
...
}

pub type Result<T> = std::result::Result<T, Error>;
```

```rust
// src/.../tests.rs

use crate::error::Result;

#[tokio::test]
async fn test_database_get_secret_key() {
    async fn inner(database: &Database) -> Result<String> {
        database.create_table_auth_if_not_exists().await?;
        database.create_secret_key_if_not_exists().await?;

        let secret_key = database.get_secret_key().await?;

        Ok(secret_key)
    }

    let database = get_database().await;

    let result = inner(&database).await;
    reset(database).await;

    assert_eq!(result.unwrap().len(), 44);

    sleep_async(1).await;
}
```

I've used `use crate::error::Result;`, but analyzer cannot infer it correctly.

image

It infered to `Result<String, {unknown}>`.

This doesn't happen to outer functions.

The tests can be run successfully.

The only thing annoying me is that it doesn't offer code completion.

roife:
fix #17338 (continuing from #17251).

In #17251, we injected preludes into non-top-level modules, which leading to r-a to directly resolve names in preludes in block modules. This PR fix it by checking whether the module is a pseudo-module introduced by blocks. (similar to what we do for extern preludes)


Veykril:
Thanks!
@bors r+

PR #17353, opened 2024-06-06
fix: incorrect formatting of hover actions

### LINKED ISSUE #12728: Incorrect formatting of hover action
**rust-analyzer version**: rust-analyzer version: 0.0.0 (75b22326d 2022-07-03)
**rustc version**: rustc 1.64.0-nightly (1517f5de0 2022-07-07)

**relevant settings**: rust-analyzer.hover.actions.references.enable

image

You can see there is a triple underscore `___` between references and `▶︎`. This is caused by the .join() here:
https://github.com/rust-lang/rust-analyzer/blame/995a17fbd9f7cdedfdafec946b1d9e44eff448b4/editors/code/src/client.ts#L61
Is the `___` there for a particular reason, @vsrs ?

roife:
fix #12728.

### Changes

- Use ` | ` as the separator for actions. (I'm not sure why we use `___` previously)
- Reorder actions to match codelens

### DIFF
--- a/crates/ide/src/hover.rs
+++ b/crates/ide/src/hover.rs
@@ -375,8 +375,8 @@ pub(crate) fn hover_for_definition(
     HoverResult {
         markup: render::process_markup(sema.db, def, &markup, config),
         actions: [
-            show_implementations_action(sema.db, def),
             show_fn_references_action(sema.db, def),
+            show_implementations_action(sema.db, def),
             runnable_action(sema, def, file_id),
             goto_type_action_for_def(sema.db, def, &notable_traits),
         ]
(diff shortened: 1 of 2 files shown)


roife:
# Before

image


# After

image

roife:
It's weird that we haven't fixed it🫨. I used to think it was a feature. 🙈

Veykril (approved):
I think the idea was to have that as `\n___\n` which will render a line in markdown, though this works as well  I think

roife:
> I think the idea was to have that as `\n___\n` which will render a line in markdown, though this works as well I think

That makes sense.

PR #17356, opened 2024-06-06
chore: Prefer tracing span shorthand macros

Wilfred:
Use `info_span!()` and `debug_span!()` rather than the more verbose `tracing::span!(tracing::Level::INFO)`.

### DIFF
--- a/crates/base-db/src/change.rs
+++ b/crates/base-db/src/change.rs
@@ -51,7 +51,7 @@ impl FileChange {
     }
 
     pub fn apply(self, db: &mut dyn SourceDatabaseExt) {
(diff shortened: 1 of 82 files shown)


lnicola (approved):
Thanks, I wonder why we didn't use these right from the start :grinning:.

davidbarsky:
> Thanks, I wonder why we didn't use these right from the start :grinning:.

I picked the other option because I'd like to make `span!` default to `Level::INFO` (no need to inundate users with those details early on...) in the future and the codemod to _that_ state is pretty easy. I can't make that change  in tracing 0.1 due to backwards compatibility concerns, however.

PR #17364, opened 2024-06-07
feat: show type bounds from containers when hovering on functions

### LINKED ISSUE #12917: hover over associated functions should show the impl block they are defined in
consider this code:
```rust
struct Foo<T>(T);

impl<T: Eq> Foo<T> {
    fn new(t: T) -> Self {
        Self(t)
    }
}

fn main() {
    let foo = Foo::new(1.2);
}
```

This doesn't compile because `f64` doesn't implement `Eq`, however the hover and popup that shows for new has no mention of the bound `T: Eq` from the impl block.
It just shows:
```rust
playground::Foo
fn new(t: T) -> Self
```

I think it should also show the impl block definition, as that is an integral part of the functions signature:
```rust
playground::Foo
impl<T: Eq> Foo<T>
fn new(t: T) -> Self
```

roife:
fix #12917.

### Changes

1. Added Support for displaying the container and type bounds from it when hovering on functions with generic types.
2. Added a user config to determine whether to display container bounds (enabled by default).
3. Added regression tests.
4. Simplified and refactored `hir/display.rs` to improve readability.

### DIFF
--- a/crates/hir/src/display.rs
+++ b/crates/hir/src/display.rs
@@ -8,7 +8,7 @@ use hir_def::{
     },
     lang_item::LangItem,
     type_ref::{TypeBound, TypeRef},
-    AdtId, GenericDefId,
+    AdtId, GenericDefId, ItemContainerId, ItemTreeLoc, Lookup,
 };
 use hir_ty::{
     display::{
@@ -22,7 +22,7 @@ use itertools::Itertools;
 
 use crate::{
     Adt, AsAssocItem, AssocItem, AssocItemContainer, Const, ConstParam, Enum, ExternCrateDecl,
-    Field, Function, GenericParam, HasCrate, HasVisibility, LifetimeParam, Macro, Module,
+    Field, Function, GenericParam, HasCrate, HasVisibility, Impl, LifetimeParam, Macro, Module,
     SelfParam, Static, Struct, Trait, TraitAlias, TupleField, TyBuilder, Type, TypeAlias,
     TypeOrConstParam, TypeParam, Union, Variant,
 };
@@ -35,11 +35,18 @@ impl HirDisplay for Function {
         let mut module = self.module(db);
 
         match container {
-            Some(AssocItemContainer::Impl(_)) => {
+            Some(AssocItemContainer::Trait(trait_)) => {
+                write_trait_header(&trait_, f)?;
+                f.write_str("\n")?;
+            }
+            Some(AssocItemContainer::Impl(impl_)) => {
+                write_impl_header(&impl_, f)?;
+                f.write_str("\n")?;
+
                 // Block-local impls are "hoisted" to the nearest (non-block) module.
                 module = module.nearest_non_block_module(db);
             }
-            _ => {}
+            None => {}
         }
         let module_id = module.id;
         write_visibility(module_id, self.visibility(db), f)?;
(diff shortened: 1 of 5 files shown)


roife:
Some examples:

image

image

image

crates/hir/src/display.rs
```diff
@@ -122,6 +140,28 @@ impl HirDisplay for Function {
...
+
+    f.write_str("impl")?;
+    let def_id = GenericDefId::ImplId(impl_.id);
+    write_generic_params(def_id, f)?;
+
+    if let Some(trait_) = impl_.trait_(db) {
+        f.write_str(" ")?;
+        let trait_data = db.trait_data(trait_.id);
+        let name = trait_data.name.display(db.upcast()).to_string();
+        f.write_str(&name)?;
```
Veykril, line 154:
```suggestion
        write!(f, "{}", trait_data.name.display(db.upcast()))?;
```

crates/hir/src/display.rs
```diff
@@ -122,6 +140,28 @@ impl HirDisplay for Function {
...
 
+fn write_impl_header(impl_: &Impl, f: &mut HirFormatter<'_>) -> Result<(), HirDisplayError> {
+    let db = f.db;
+
+    f.write_str("impl")?;
+    let def_id = GenericDefId::ImplId(impl_.id);
+    write_generic_params(def_id, f)?;
+
+    if let Some(trait_) = impl_.trait_(db) {
+        f.write_str(" ")?;
```
Veykril, line 151:
```suggestion
        f.write_char(' ')?;
```

crates/hir/src/display.rs
```diff
@@ -122,6 +140,28 @@ impl HirDisplay for Function {
...
+        let trait_data = db.trait_data(trait_.id);
+        let name = trait_data.name.display(db.upcast()).to_string();
+        f.write_str(&name)?;
+        f.write_str(" for")?;
+    }
+
+    f.write_str(" ")?;
```
Veykril, line 158:
```suggestion
    f.write_str(' ')?;
```

crates/rust-analyzer/src/config.rs
```diff
@@ -479,6 +479,8 @@ config_data! {
         /// How to render the size information in a memory layout hover.
         hover_memoryLayout_size: Option<MemoryLayoutHoverRenderKindDef> = Some(MemoryLayoutHoverRenderKindDef::Both),
 
+        /// Whether to show type bounds from containers for functions.
+        hover_show_containerBounds: bool = true,
```
Veykril, line 483: I don't think we want a config for this to be honest

crates/hir/src/display.rs
```diff
@@ -562,6 +591,16 @@ fn write_where_clause(
 ) -> Result<bool, HirDisplayError> {
     let params = f.db.generic_params(def);
 
+    let container = match def {
```
Veykril, line 583: Not a fan of having this here as it clutters the function a bit more. Since we already return whether we wrote any predicates or not via the bool return, can we instead have the extra logic in the caller of this? I'd rather have the `where` part emission duplicated to the caller if nothing got printed than making this function here more complicated I think

Veykril (changes requested):

crates/rust-analyzer/src/config.rs
roife: So should we display type bounds all the time? I added this config because I’m concerned that some users might find them too noisy in more complex type signature.

Veykril: Yes we are already doing the same for the function bounds themselves so I think this is fine for now, we can add it later if need be

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17372, opened 2024-06-09
feat: Compute native diagnostics in parallel

Veykril:


lnicola (approved):
LGTM. Ideally we'd have a way to prioritize tasks (so we could use all threads if available, not just a quarter), but that ends only leads to more questions about fairness etc.

PR #17375, opened 2024-06-09
Don't intern attribute inputs as their spans make them unique

Veykril:


Veykril:
Hmm, this regressed memory usage actually. I guess we do have these duplicates between item tree and attr query, in which case an arc would probably be the better choice

PR #17377, opened 2024-06-10
internal: better print style for hir

Young-Flash:
before:
before

after:
after


crates/hir-def/src/body/pretty.rs
```diff
@@ -48,22 +48,38 @@ pub(super) fn print_body_hir(db: &dyn DefDatabase, body: &Body, owner: DefWithBo
...
+            p.buf.push(',');
+            p.buf.push(' ');
         });
+        // remove the last ", " in param list
+        if body.params.len() > 0 {
+            p.buf.truncate(p.buf.len() - 2);
+        }
         p.buf.push(')');
         p.buf.push(' ');
+        // return type
```
lnicola, line 77: Nit: maybe use `push_str`?

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17381, opened 2024-06-10
fix: ensure that the parent of a SourceRoot cannot be itself

### LINKED ISSUE #17378: Version `0.3.1992` rust-analyzer process spins out of control in macOS VSCode
**rust-analyzer version**: `0.3.1992`

**rustc version**: rustc 1.78.0 (9b00956e5 2024-04-29)

**relevant settings**: Standard Rustup installation

**repository link (if public, optional)**: n/a

n/a

VSCode updated my extensions today and on the latest rust-analyzer version things are spinning out of control:

CleanShot 2024-06-10 at 10 08 43@2x

I've worked around it by downgrading the version on VSCode itself:

CleanShot 2024-06-10 at 10 12 14@2x

roife:
fix #17378.

In `FileSetConfig.map`, different roots might be mapped to the same `root_id` due to deduplication in `ProjectFolders::new`:

```rust
// Example from rustup
/Users/roife/code/rustup/target/debug/build/rustup-863a063426b56c51/out
/Users/roife/code/rustup
```

In `source_root_parent_map`, r-a might encounter paths where their SourceRootId (i.e. `root_id`) is identical, yet one the them is the parent of the another. This situation can cause the `root_id` to be its own parent, potentially leading to an infinite loop.

This PR resolves such cases by adding a check.

### DIFF
--- a/crates/load-cargo/src/lib.rs
+++ b/crates/load-cargo/src/lib.rs
@@ -285,7 +285,6 @@ impl SourceRootConfig {
     /// If a `SourceRoot` doesn't have a parent and is local then it is not contained in this mapping but it can be asserted that it is a root `SourceRoot`.
     pub fn source_root_parent_map(&self) -> FxHashMap<SourceRootId, SourceRootId> {
         let roots = self.fsc.roots();
-        let mut i = 0;
         roots
             .iter()
             .enumerate()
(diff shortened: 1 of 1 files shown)


roife:
could a circular reference occur here?

For example:

```
rootA: ["/ROOT/a", "/ROOT/a/b/c/d"]
rootB: ["/ROOT/a/b"]
```

This might cause `rootA` to be the parent of `rootB` while `rootB` is the parent of `rootA`.🤯

rami3l:
@roife I can confirm that your test build on [73399](https://github.com/rust-lang/rust-analyzer/pull/17381/commits/733995c0d7088f0a193aa77204d68f3fecdf561a) works for me. Nice work!

Veykril:
We already depend on petgraph so we could just use a proper graph impl for this instead that then deals with accidental cycle handling if they can actually occur. Though this entire thing is very hacky in the first place...

Thanks a lot for investigating!
@bors r+

PR #17395, opened 2024-06-11
chore: remove `UnindexinedProject` notification 

davidbarsky:
This PR is split out from https://github.com/rust-lang/rust-analyzer/pull/17246/ (and contains its changes, which is a little annoying from a review perspective...). I'd like to land this change a week or so after #17246 lands in order to give any users of the unindexed project notification time to adopt migrate.


Veykril:
Just to check, is this resolved within#17246, or does this contain other changes (and just needs a rebase)?

davidbarsky:
This one contains a few other changes and needs a rebase. I'll land it after monday's release in cas there are any users of this.

editors/code/src/main.ts
```diff
@@ -6,25 +6,14 @@ import { type CommandFactory, Ctx, fetchWorkspace } from "./ctx";
 import * as diagnostics from "./diagnostics";
 import { activateTaskProvider } from "./tasks";
 import { setContextValue } from "./util";
-import type { JsonProject } from "./rust_project";
 
 const RUST_PROJECT_CONTEXT_NAME = "inRustProject";
 
-// This API is not stable and may break in between minor releases.
-export interface RustAnalyzerExtensionApi {
```
Veykril, line 14: Let's keep the type itself just with the `client` property, that's what we used to have iirc

davidbarsky: done; kept it.

PR #17407, opened 2024-06-12
fix: avoid doubling cargo args in runnables

### LINKED ISSUE #17402: cargo runnables no longer work
It seems that after this PR, the tests have stopped running.

```
 *  Executing task: cargo cargo test --package sqruff-lib --lib -- rules::layout::LT13::tests::test_fail_leading_whitespace_comment --exact --show-output 

error: no such command: `cargo`

        View all installed commands with `cargo --list`
        Find a package to install `cargo` with `cargo search cargo-cargo`
```

_Originally posted by @gvozdvmozgu in https://github.com/rust-lang/rust-analyzer/issues/16840#issuecomment-2162227412_

davidbarsky:
Fixes #17402.

Sorry about this—I think I missed up the rebase!

### DIFF
--- a/editors/code/src/tasks.ts
+++ b/editors/code/src/tasks.ts
@@ -1,8 +1,7 @@
 import * as vscode from "vscode";
-import * as toolchain from "./toolchain";
 import type { Config } from "./config";
 import { log } from "./util";
-import { unwrapUndefinable } from "./undefinable";
+import { expectNotUndefined, unwrapUndefinable } from "./undefinable";
 
 // This ends up as the `type` key in tasks.json. RLS also uses `cargo` and
 // our configuration should be compatible with it so use the same key.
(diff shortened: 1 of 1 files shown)


editors/code/src/tasks.ts
```diff
@@ -145,11 +145,13 @@ async function cargoToExecution(
         const cargoPath = await toolchain.cargoPath();
         const cargoCommand = definition.overrideCargo?.split(" ") ?? [cargoPath];
 
-        const args = [definition.command].concat(definition.args ?? []);
-        const fullCommand = [...cargoCommand, ...args];
-        const processName = unwrapUndefinable(fullCommand[0]);
+        const args = expectNotUndefined(
+            definition.args,
```
lnicola, line 149: I'm not really familiar with the code here, but shouldn't we use `definition.command` just like before?

davidbarsky: i'm not sure what you mean—can you expand?

lnicola: We're not using `definition.command` in the `cargo` case. I guess it's also present in `definition.args`, then?

Veykril: Shouldn't we drop the entire `cargoCommand` querying here? If I look at the previous PR, `createTask` already handles this now no?

davidbarsky: uh... yes. i think we can drop it.

PR #17410, opened 2024-06-12
internal: Prefer .expect() over .unwrap() in flycheck

Wilfred:
I managed to hit these `.unwrap()` calls a few times when debugging, so replace them with `.expect()` calls with unique text descriptions.

### DIFF
--- a/crates/flycheck/src/lib.rs
+++ b/crates/flycheck/src/lib.rs
@@ -328,7 +328,8 @@ impl FlycheckActor {
                     tracing::debug!(flycheck_id = self.id, "flycheck finished");
 
                     // Watcher finished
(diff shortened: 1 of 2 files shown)


Wilfred:
@Veykril I'm not entirely sure, I was playing with some changes to `rust-project.json` generation tools.

Feel free to close the PR if you don't think it's worthwhile, I just figured it might make future debugging easier.

Veykril:
Generally we do prefer expects but for these they don't really add any value imo, when these fail the other half of the channels dropped which only happens if the main thread dies, likewise the message doesn't really add any more information imo so I'll close this

lnicola:
Not exactly this, but see https://github.com/rust-lang/rust-analyzer/issues/17445.

PR #17411, opened 2024-06-13
fix: Improve hover text in unlinked file diagnostics

Wilfred:
Use full sentences, and mention how to disable the diagnostic if users are intentionally working on unowned files.

Screenshot 2024-06-12 at 5 55 48 PM

(Full disclosure: I've tested a rust-analyzer build in VS Code, but the pop-up logic is currently disabled due to #17062, so I haven't tested that.)


Veykril:
We used to pop a notification for unlinked files (the code is still there even I think), its just disabled because it was buggy and annoying.

Wilfred:
@Veykril Yep, the code is still there, it was just disabled in #17062. Would you like me to put up a PR to remove the dead code?

Veykril (approved):

Veykril:
I'm not yet decided on what to do with the notification pop up, I think we want to keep it and re-purpose it with the landing page stuff maybe. 
@bors r+

PR #17412, opened 2024-06-13
fix: add a breaker to avoid infinite loops from source root cycles

### LINKED ISSUE #17409: Version `0.3.1995` rust-analyzer process spins out of control in macOS VSCode
Followup of #17378 - as similar to https://github.com/rust-lang/rust-analyzer/issues/17378#issuecomment-2162333916 I am also still experiencing the same issue of high CPU on version `0.3.1995` (and also on `main` at fa486e6).

Doing some debugging locally, I find that the infinite loop mentioned in https://github.com/rust-lang/rust-analyzer/issues/17378#issuecomment-2158379807 seems to still the culprit.

I inserted a debug print:

```diff
$ git diff 
diff --git a/crates/rust-analyzer/src/config.rs b/crates/rust-analyzer/src/config.rs
index e8504979b..546f1e49a 100644
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -2532,6 +2532,7 @@ macro_rules! _impl_for_config_data {
                 $vis fn $field(&self, source_root: Option<SourceRootId>) -> &$ty {
                     let mut par: Option<SourceRootId> = source_root;
                     while let Some(source_root_id) = par {
+                        dbg!(source_root_id, &self.source_root_parent_map);
                         par = self.source_root_parent_map.get(&source_root_id).copied();
                         if let Some((config, _)) = self.ratoml_files.get(&source_root_id) {
                             if let Some(value) = config.$field.as_ref() {
```

... and this is my output:

```
[crates/rust-analyzer/src/config.rs:336:1] source_root_id = SourceRootId(
    200,
)
[crates/rust-analyzer/src/config.rs:336:1] &self.source_root_parent_map = {
    SourceRootId(
        202,
    ): SourceRootId(
        200,
    ),
    SourceRootId(
        200,
    ): SourceRootId(
        202,
    ),
    // elided for brevity
}
[crates/rust-analyzer/src/config.rs:336:1] source_root_id = SourceRootId(
    202,
)
[crates/rust-analyzer/src/config.rs:336:1] &self.source_root_parent_map = {
    SourceRootId(
        202,
    ): SourceRootId(
        200,
    ),
    SourceRootId(
        200,
    ): SourceRootId(
        202,
    ),
    // elided for brevity
}
// and repeat ad infinitum between source root 202 and 200
```

So I think the patch in #17381 is probably insufficient because here I don't have a source root with a parent of itself; instead somehow I have a cycle between two source roots 202 and 200.

I guess there's brute force solutions to prevent this infinite loop like bailing out if the number of iterations ever exceeds the total size of `source_root_parent_map`. Or maybe there's similar ideas to #17381 that can prune this graph.

... I would go further and submit a patch but I really need to go to sleep, sorry 🙈

cc @Veykril @roife

davidhewitt:
See #17409 

This patch prevents infinite looping from cycles by giving up if the number of source roots checked for a config value reaches the total number of source roots.

Alternative more precise options include creating a set of all source roots visited and giving up as soon as a cycle is encountered, but I wasn't sure how costly an allocation would be here for performance.

Can confirm that locally this fixes the problem for me.

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -2531,12 +2531,21 @@ macro_rules! _impl_for_config_data {
                 #[allow(non_snake_case)]
                 $vis fn $field(&self, source_root: Option<SourceRootId>) -> &$ty {
                     let mut par: Option<SourceRootId> = source_root;
+                    let mut traversals = 0;
                     while let Some(source_root_id) = par {
                         par = self.source_root_parent_map.get(&source_root_id).copied();
                         if let Some((config, _)) = self.ratoml_files.get(&source_root_id) {
                             if let Some(value) = config.$field.as_ref() {
                                 return value;
                             }
+                        }
+                        // Prevent infinite loops caused by cycles by giving up when it's
+                        // clear that we must have either visited all source roots or
+                        // encountered a cycle.
+                        traversals += 1;
+                        if traversals >= self.source_root_parent_map.len() {
+                            // i.e. no source root contains the config we're looking for
+                            break;
                         }
                     }
 


Veykril:
Certainly not the way to fix this but it should do for now
@bors r+

davidhewitt:
Thanks; completely agree that this just hides the problem at the datastructure level but at least it'll save a few watts of energy on users' machines 😂

It also doesn't seem the _worst_ thing to have a way to break out of the infinite loop as an emergency valve. 👀

Issue #17397, opened 2024-06-12
Unused/inactive highlighting when file is not linked


cynecx:
**rust-analyzer version**: 4af21ffb026c7ec3a97a484ca27b36f703eb5fb1

Before an unlinked file (or a `#[cfg(FALSE)]` module) just had a small diagnostic for the first line. Now the whole file gets the visually marked as "unused", which was implemented here: https://github.com/rust-lang/rust-analyzer/pull/17350.

The PR mentions:

> Whilst this is marginally more prominent, it's less invasive than a pop-up, and users do want to know why they're getting no rust-analyzer support in certain files.

I agree it is less "invasive" compared to pop-up but now it's also more visually distracting (or for me straight up annoying). My workflow quite often consists of workflows where I've got unlinked files in the work-tree and having the whole file "grayed out" it not great visually speaking.

There must be a better to do this. It might be interesting to see what RustRover is doing here. But for now I've reverted this change locally, so it's not bugging me anymore.

Veykril:
RustRover shows a banner-like thing  at the top of the file I believe. VSCode doesn't really let us to do that unfortunately.

@davidbarsky @Wilfred I'll put this on your plate to figure out :) I figured people would find this annoying as that was the exact reason why the diagnostic was limited to the first few characters of the file

RalfJung:
Yeah this definitely sounds quite annoying, e.g. when working on ui test files in rustc and Miri which are expected to be "unlinked".

lnicola:
How about a code lens at the top of the file? It might still be too easy to miss though.

RalfJung:
I think a big part of the problem here is dealing with intentionally unlinked files. If there's a better way to avoid false warnings then it is less of a problem to make the warning more visible.

cynecx:
I agree with Ralf here. But what can we do in the meantime until this gets figured out? I'd be in favor of reverting that PR or at the very least put it behind a ff/config.

lnicola:
Disabling the diagnostic (`rust-analyzer.diagnostics.disabled`) probably works.

cynecx:
Oh… I didn't know that setting existed 😅 Thanks!

davidbarsky:
Yeah, I think it's reasonable that some people are annoyed by this, which is why I think should be controlled at the language/build system level, not the IDE level[^1]. However, _that_ discussion isn't immediately relevant to this issue.

[...]

PR #17415, opened 2024-06-13
fix: Only show unlinked-file diagnostic on first line during startup

Wilfred:
This partially reverts #17350, based on the feedback in #17397.

If we don't have an autofix, it's more annoying to highlight the whole file. This autofix heuristic fixes the diagnostic being overwhelming during startup.


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17417, opened 2024-06-13
refactor: Prefer plain trait definitions over macros for impl_intern_value_trivial

Wilfred:
`impl_intern_value_trivial` can be defined with a trait directly, so prefer that over a macro definition.


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17419, opened 2024-06-14
Filter builtin macro expansion

### LINKED ISSUE #14177: confusing macro expansion for asm!
The ability to recursively expand macros is wonderful, and I find myself using it quite often to "see" what a particular construct is doing. So far, it's been great (it's so useful I often find myself looking for it under the inline context menu and have to remind myself it's in the command palette), but today I ran into this expansion of [a macro from the riscv crate](https://github.com/rust-embedded/riscv/blob/620c5bf225557dce465785571195a5c76a7bfa5a/src/register/mstatus.rs#L212):

screenshot of an expansion for the read_csr_as! macro with a `loop` block highlighted

The highlighted loop block in the middle looks like it's the result of expanding [this part of the `read_csr!` macro](https://github.com/rust-embedded/riscv/blob/620c5bf225557dce465785571195a5c76a7bfa5a/src/register/macros.rs#L10
):

```rust
#[cfg(riscv)]
() => {
    let r: usize;
    core::arch::asm!(concat!("csrrs {0}, ", stringify!($csr_number), ", x0"), out(reg) r);
    r
}
```
via this rust-analyzer notion of what `asm!` represents: https://github.com/rust-lang/rust-analyzer/blob/a6603fc21d50b3386a488c96225b2d1fd492e533/crates/hir-expand/src/builtin_fn_macro.rs#L299-L302

The question I was hoping to answer by expanding the macro was approximately "what are the semantics of reading the mstatus register here?", and while `{loop{}};` is a syntactically valid way to spell `!` it is, in this case, shadowing the actual `csrrs` instruction that's populating the `r` local.

From the git history, I infer that this is largely because macro expansion is doing at least triple duty here: in addition to providing me with visibility into the system, it's being used internally both for syntax highlighting and diagnostics (and probably also completion?). I can't speak to the other builtin macros, but for `asm!` (and `global_asm!`) in particular, that would seem to set an extremely high bar for rust-analyzer: in addition to being a full Rust lsp implementation, fully implementing `asm!` would require embedding an architecture-dependent language server implementation for each possible rustc target.

Instead, I think I would have preferred that the expansion handle the `concat!` and `stringify!` but leave the `asm!` and so end up looking something like this:

```rust
let r: usize;
core::arch::asm!("csrrs {0}, 0x300, x0", out(reg) r);
r
```

but, if leaving the macro un-expanded is impossible for Reasons (as I suspect it may be), then I would have still found it less confusing to find something like:

```rust
let r: usize;
{
    ///! expanded from
    ///!     core::arch::asm!(concat!("csrrs {0}, ", stringify!($csr_number), ", x0"), out(reg) r);
    /// internal to rust-analyzer only, this represents `!` (divergence) as a substitute for the above assembly semantics
    loop {}
};
r
```

What do you think?

**rust-analyzer version**: `rust-analyzer version: 0.3.1394-standalone`

**rustc version**: `rustc 1.69.0-nightly (2d14db321 2023-02-15)`

**relevant settings**: N/A

ishanjain28:
This PR adds a filter on the types of built in macros that are allowed to be expanded. 

Currently, This list of allowed macros contains, `stringify, cfg, core_panic, std_panic, concat, concat_bytes, include, include_str, include_bytes, env` and `option_env`.

Fixes #14177

### DIFF
--- a/crates/hir/src/semantics.rs
+++ b/crates/hir/src/semantics.rs
@@ -19,8 +19,12 @@ use hir_def::{
     AsMacroCall, DefWithBodyId, FunctionId, MacroId, TraitId, VariantId,
 };
 use hir_expand::{
-    attrs::collect_attrs, db::ExpandDatabase, files::InRealFile, name::AsName, InMacroFile,
-    MacroCallId, MacroFileId, MacroFileIdExt,
+    attrs::collect_attrs,
+    builtin_fn_macro::{BuiltinFnLikeExpander, EagerExpander},
+    db::ExpandDatabase,
+    files::InRealFile,
+    name::AsName,
+    InMacroFile, MacroCallId, MacroFileId, MacroFileIdExt,
 };
 use itertools::Itertools;
 use rustc_hash::{FxHashMap, FxHashSet};
@@ -319,6 +323,30 @@ impl<'db> SemanticsImpl<'db> {
         } else {
             sa.expand(self.db, macro_call)?
         };
+        let macro_call = self.db.lookup_intern_macro_call(file_id.macro_call_id);
+
+        match macro_call.def.kind {
+            hir_expand::MacroDefKind::BuiltIn(
+                _,
+                BuiltinFnLikeExpander::Cfg
+                | BuiltinFnLikeExpander::StdPanic
+                | BuiltinFnLikeExpander::Stringify
+                | BuiltinFnLikeExpander::CorePanic,
+            )
+            | hir_expand::MacroDefKind::BuiltInEager(
+                _,
+                EagerExpander::Env
+                | EagerExpander::Concat
+                | EagerExpander::Include
+                | EagerExpander::OptionEnv
+                | EagerExpander::IncludeStr
+                | EagerExpander::ConcatBytes
+                | EagerExpander::IncludeBytes,
+            ) => {
+                // Do nothing and allow matching macros to be expanded
+            }
+            _ => return None,
+        }
 
         let node = self.parse_or_expand(file_id.into());
         Some(node)
--- a/crates/hir/src/semantics.rs
+++ b/crates/hir/src/semantics.rs
@@ -345,7 +345,13 @@ impl<'db> SemanticsImpl<'db> {
             ) => {
                 // Do nothing and allow matching macros to be expanded
             }
-            _ => return None,
+
+            hir_expand::MacroDefKind::BuiltIn(_, _)
+            | hir_expand::MacroDefKind::BuiltInAttr(_, _)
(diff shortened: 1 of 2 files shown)


ishanjain28:
Added `format_args` to the list of expandable macros

Veykril:
> Added `format_args` to the list of expandable macros

Ah right we can't filter that here as it would break other features that require to look into the macro call. Change of plans, let's keep `Semantics::expand` as is and duplicate it to a new function `Semantics::expand_no_builtins` or similar that has the filtering for the given builtins we listed. Then use that new function for just this feature.

ishanjain28:
@veykril Does it look right now ? I created a `expand_allowed_builtins` function and update the call in `expand_macro.rs` to call this function instead of the old `expand` function.

crates/hir/src/semantics.rs
```diff
@@ -324,6 +328,53 @@ impl<'db> SemanticsImpl<'db> {
...
+        let macro_call = self.db.lookup_intern_macro_call(file_id.macro_call_id);
+
+        match macro_call.def.kind {
+            hir_expand::MacroDefKind::BuiltIn(
+                _,
+                BuiltinFnLikeExpander::Cfg
+                | BuiltinFnLikeExpander::StdPanic
+                | BuiltinFnLikeExpander::Stringify
+                | BuiltinFnLikeExpander::CorePanic
+                | BuiltinFnLikeExpander::FormatArgs,
```
Veykril, line 351: We can skip expanding this one as well now, it expands to unstable syntax so the user shouldn't need to see it

crates/ide/src/expand_macro.rs
```diff
@@ -228,6 +229,32 @@ mod tests {
...
+            $0concat!("test", 10, 'b', true);"#,
+            expect![[r#"
+                concat!
+                "test10btrue""#]],
+        )];
+        for (test, expect) in tests {
+            check(test, expect);
+        }
+    }
+
```
Veykril, line 257: Let's just write these out as tests like the rest, that is one test for the failure case, one for success. This here makes it harder to read what is being tested

ishanjain28: Noted, I'll update

crates/hir/src/semantics.rs
ishanjain28: hey @Veykril Are you referring to all 4 or some subset specifically ?

ishanjain28: I assume you meant `format_args`. I have removed it.

Veykril:
Thanks!
Pushed a commit to this myself cleaning up the match a bit subjectively, didn't wanna bother you there with another review :)
@bors r+

Issue #17433, opened 2024-06-16
Heuristics for omitting parameter inlay hints are not documented nor are they configurable


Lunaphied:
Right now, a parameter inlay hint is automatically omitted using a series of heuristics defined [here](https://github.com/rust-lang/rust-analyzer/blob/6b8b8ff4c56118ddee6c531cde06add1aad4a6af/crates/ide/src/inlay_hints/param_name.rs#L92-L99). This isn't referenced in the documentation at all, and while useful, also cannot be tweaked or disabled.

As a result it can appear that inlay hints for parameters are broken, since quite frequently these heuristics will match and they will be omitted.

I think an appropriate resolution would be to both document the heuristics being used (something I'm planning to PR in a moment), and to allow disabling these heuristics. Ideally each heuristic could be individually disabled and/or configured but a global disable for the heuristic omission would also be ideal.

PR #17434, opened 2024-06-16
docs: document omission heuristics for parameter inlay hints

Lunaphied:
These are not currently documented and could cause users to think that their rust-analyzer configuration is broken.

Partially addresses #17433.


Veykril:
Thanks!
@bors r+

PR #17438, opened 2024-06-17
feat: add `toggleLSPLogs` command

### LINKED ISSUE #8233: Add: "show LSP logs" command
To get LSP logs in vscode, one needs to set a rather obscure setting: `"rust-analyzer.trace.server": "verbose"`. This should be a command which is discoverable via command palette. 

It should be implemented by the client, and in language agnostic way. The best way to arrive at that state though is to implement it ourselves and then poke editor's maintainers with "see what we have here! You should have this too". 

This is mostly client-side work:

* add command to package.json
* implement command in commands.ts by writing go the settings (not sure what's the best scope would be -- probably project-wide) and maybe opening the corresponding vscode view
* documenting it in the manual.adoc (preferably with equivalents for other editors)
* adjusting docs in `dev` folder if they mention logs.

jjoeldaniel:
Implement client-side command to toggle LSP logs in VSCode.

The command replaces the need to add/remove the `"rust-analyzer.trace.server": "verbose"` setting each time one wants to display logs. I've also updated the docs/ instances that reference the now outdated manual method.

The command labeled `rust-analyzer: Toggle LSP Logs` enables the setting project-wide and opens the relevant trace output channel.

Closes #8233

### DIFF
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -145,7 +145,7 @@ To log all communication between the server and the client, there are two choice
   ```
   env RA_LOG=lsp_server=debug code .
   ```
-* You can log on the client side, by enabling `"rust-analyzer.trace.server": "verbose"` workspace setting.
+* You can log on the client side, by the `rust-analyzer: Toggle LSP Logs` command or enabling `"rust-analyzer.trace.server": "verbose"` workspace setting.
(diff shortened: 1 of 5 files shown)


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17439, opened 2024-06-17
Properly prime all crate def maps in parallel_prime_caches

Veykril:


davidbarsky:
I tried this against [`app/buck2/bin/buck2.rs`](https://github.com/facebook/buck2/blob/9528c3dedfc255fafa9c14e1456a810358698baf/app/buck2/bin/buck2.rs). With these changes, the load time for the first hover after indexing was about a second. Without these changes, that same action took about 10 seconds, so this is a great improvement. Even though more crates are indexed, the more informative visual progress indicator is _wildly_ helpful. Thanks for this change!

---

Stepping back a bit, it'd be nice if it were possible to create a work-done progress indicator if a go-to-def or whatever takes more than a second, but I don't think that's really possible/feasible until if/when rust-analyzer uses async internally. I imagine this would look like a [`Timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html), but instead of canceling the other future, it'd create a new future.

crates/hir-expand/src/fixup.rs
```diff
@@ -276,6 +275,62 @@ pub(crate) fn fixup_syntax(
                         ]);
                     }
                 },
+                ast::RecordExprField(it) => {
```
davidbarsky, line 278: sorry, dumb question: why are the fixup changes included in this PR?

Veykril: Because I was too lazy to put them into a separate PR earlier and only later realized I won't self-merge this PR immediately as I usually do

PR #17460, opened 2024-06-20
minor: Sync from rust

lnicola:


lnicola:
@RalfJung a lot of empty commits here, but it seems to match https://github.com/rust-lang/rust-analyzer/pull/17152, so I suppose it's fine?

RalfJung:
I guess?

I don't know why it included `tests/run-make: update to use fs_wrapper `, but that is a strange commit already in rustc -- it has an empty diff there. Usually does not mirror a commit when there's no diff after applying the filters, but maybe there is an exception when the commit was already empty before applying the filter? Cc @christian-schilling

RalfJung:
... is that normal?^^

lnicola:
Yeah, it happens all the time :grinning:.

christian-schilling:
> I guess?
> 
> 
> 
> I don't know why it included `tests/run-make: update to use fs_wrapper`, but that is a strange commit already in rustc -- it has an empty diff. Usually josh does not mirror a commit when there's no diff after applying the filters, but maybe there is an exception when the commit was already empty before applying the filter? Cc @christian-schilling 
> 
> 
> 
> EDIT: We had the same in Miri, https://github.com/rust-lang/miri/pull/3681 also mirrored that empty commit.

Yes. You guessed right. Empty diffs are skipped unless the original was also empty.
This special rule is necessary because otherwise an empty diff created in the partial history would immediately break the round trip.
On an empty diff there is no way to tell where it belongs to, so we have to include it everywhere.

RalfJung:
Ah, that makes sense - thanks!

PR #17466, opened 2024-06-20
chore: Fix diagnostic name in macro_error.rs

chenyukang:


Veykril:
We should probably derive these more structurally from the function or module name maybe so they don't go out of sync
Thanks!
@bors r+

PR #17467, opened 2024-06-20
feat: add bool_to_enum assist for parameters 

### LINKED ISSUE #17400: `bool_to_enum` assist should work on parameters
https://github.com/rust-lang/rust-analyzer/blob/4af21ffb026c7ec3a97a484ca27b36f703eb5fb1/crates/ide-assists/src/handlers/bool_to_enum.rs currently only works on "local variables, fields, constants, and statics", parameters should be included in the local variables part but that is not the case as we look for `LetStmt` here https://github.com/rust-lang/rust-analyzer/blob/4af21ffb026c7ec3a97a484ca27b36f703eb5fb1/crates/ide-assists/src/handlers/bool_to_enum.rs#L96-L117 opposed to looking for patterns.

So instead of looking for the statement this should look for the `IdentPat` directly, then check if that is within a let statement or parameter and act accordingly.

winstxnhdw:
## Summary

This PR adds parameter support for `bool_to_enum` assists. Essentially, the assist can now transform this:

```rs
fn function($0foo: bool) {
    if foo {
        println!("foo");
    }
}
```

To this,

```rs
#[derive(PartialEq, Eq)]
enum Bool { True, False }

fn function(foo: Bool) {
    if foo == Bool::True {
        println!("foo");
    }
}
```

Thanks to @/davidbarsky  for the test skeleton (:

Closes #17400

### DIFF
--- a/crates/ide-assists/src/handlers/bool_to_enum.rs
+++ b/crates/ide-assists/src/handlers/bool_to_enum.rs
@@ -1,3 +1,4 @@
+use either::Either;
 use hir::{ImportPathConfig, ModuleDef};
 use ide_db::{
     assists::{AssistId, AssistKind},
@@ -93,27 +94,30 @@ struct BoolNodeData {
 fn find_bool_node(ctx: &AssistContext<'_>) -> Option<BoolNodeData> {
     let name: ast::Name = ctx.find_node_at_offset()?;
 
-    if let Some(let_stmt) = name.syntax().ancestors().find_map(ast::LetStmt::cast) {
-        let bind_pat = match let_stmt.pat()? {
-            ast::Pat::IdentPat(pat) => pat,
-            _ => {
-                cov_mark::hit!(not_applicable_in_non_ident_pat);
-                return None;
-            }
-        };
-        let def = ctx.sema.to_def(&bind_pat)?;
+    if let Some(ident_pat) = name.syntax().parent().and_then(ast::IdentPat::cast) {
+        let def = ctx.sema.to_def(&ident_pat)?;
         if !def.ty(ctx.db()).is_bool() {
             cov_mark::hit!(not_applicable_non_bool_local);
             return None;
         }
 
-        Some(BoolNodeData {
-            target_node: let_stmt.syntax().clone(),
-            name,
-            ty_annotation: let_stmt.ty(),
-            initializer: let_stmt.initializer(),
-            definition: Definition::Local(def),
-        })
+        let local_definition = Definition::Local(def);
+        match ident_pat.syntax().parent().and_then(Either::<ast::Param, ast::LetStmt>::cast)? {
+            Either::Left(param) => Some(BoolNodeData {
+                target_node: param.syntax().clone(),
+                name,
+                ty_annotation: param.ty(),
+                initializer: None,
+                definition: local_definition,
+            }),
+            Either::Right(let_stmt) => Some(BoolNodeData {
+                target_node: let_stmt.syntax().clone(),
+                name,
+                ty_annotation: let_stmt.ty(),
+                initializer: let_stmt.initializer(),
+                definition: local_definition,
+            }),
+        }
     } else if let Some(const_) = name.syntax().parent().and_then(ast::Const::cast) {
         let def = ctx.sema.to_def(&const_)?;
         if !def.ty(ctx.db()).is_bool() {
--- a/crates/ide-assists/src/handlers/bool_to_enum.rs
+++ b/crates/ide-assists/src/handlers/bool_to_enum.rs
@@ -518,6 +518,98 @@ mod tests {
     use crate::tests::{check_assist, check_assist_not_applicable};
 
     #[test]
+    fn parameter_with_first_param_usage() {
+        check_assist(
+            bool_to_enum,
+            r#"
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -517,6 +539,30 @@ mod tests {
 
     use crate::tests::{check_assist, check_assist_not_applicable};
 
+    #[test]
+    fn parameter_with_usage() {
+        check_assist(
+            bool_to_enum,
+            r#"
+fn function($0foo: bool) {
```
Young-Flash, line 547: add tests for different param location (the second param, the last param,....) would be nice

crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -93,7 +93,29 @@ struct BoolNodeData {
 fn find_bool_node(ctx: &AssistContext<'_>) -> Option<BoolNodeData> {
     let name: ast::Name = ctx.find_node_at_offset()?;
 
-    if let Some(let_stmt) = name.syntax().ancestors().find_map(ast::LetStmt::cast) {
+    if let Some(param) = name.syntax().ancestors().find_map(ast::Param::cast) {
```
Veykril, line 96: This can be simplified a bit (that is we can mostly collapse this and the let statement branch) by first looking for the ident pat, then branching on the parents of that instead (since both let statements and param positions take patterns)
```rs
if let Some(param) = name.syntax().ancestors().find_map(ast::IdentPat::cast) {
    .. do type checks and stuff as before
    match param.syntax().ancestors().find_map(Either::<ast::Param, ast::LetStmt>::cast)? {
        Either::Left(param) => populate BoolNodeData accordingly
        Either::Right(let_stmt=> populate BoolNodeData accordingly
    }
```

crates/ide-assists/src/handlers/bool_to_enum.rs
Veykril: especially one test for closure params would be good

crates/ide-assists/src/handlers/bool_to_enum.rs
winstxnhdw: I didn't know about `Either`, thank you!

crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -93,27 +94,36 @@ struct BoolNodeData {
...
-    if let Some(let_stmt) = name.syntax().ancestors().find_map(ast::LetStmt::cast) {
-        let bind_pat = match let_stmt.pat()? {
-            ast::Pat::IdentPat(pat) => pat,
-            _ => {
-                cov_mark::hit!(not_applicable_in_non_ident_pat);
-                return None;
-            }
-        };
-        let def = ctx.sema.to_def(&bind_pat)?;
+    if let Some(ident_pat) = name.syntax().ancestors().find_map(ast::IdentPat::cast) {
```
Veykril, line 97:
```suggestion
    if let Some(ident_pat) = name.syntax().parent().and_then(ast::IdentPat::cast) {
```

crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -93,27 +94,36 @@ struct BoolNodeData {
...
 
-        Some(BoolNodeData {
-            target_node: let_stmt.syntax().clone(),
-            name,
-            ty_annotation: let_stmt.ty(),
-            initializer: let_stmt.initializer(),
-            definition: Definition::Local(def),
-        })
+        let local_definition = Definition::Local(def);
+        match ident_pat.syntax().ancestors().find_map(Either::<ast::Param, ast::LetStmt>::cast)? {
```
Veykril, line 105:
```suggestion
        match ident_pat.syntax().parent().and_then(Either::<ast::Param, ast::LetStmt>::cast)? {
```

crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -93,27 +94,36 @@ struct BoolNodeData {
...
+            Either::Right(let_stmt) => match let_stmt.pat()? {
+                ast::Pat::IdentPat(_) => Some(BoolNodeData {
+                    target_node: let_stmt.syntax().clone(),
+                    name,
+                    ty_annotation: let_stmt.ty(),
+                    initializer: let_stmt.initializer(),
+                    definition: local_definition,
+                }),
+                _ => {
+                    cov_mark::hit!(not_applicable_in_non_ident_pat);
```
Veykril, line 122: Let's remove the `cov_mark` for this, this is covered by the proposed changes above (via the `?` bubbling), otherwise this just adds noise to the code imo.

Veykril (commented):
Okay one more simplification, I didn't realize the code generally assumes a top level ident pattern only, so using ancestor is wrong. We want to check if the ident pattern is the top most patterns in the statement/param.

Also could you squash the commits? There is a lot of very small commits that I don't think add any value here

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17469, opened 2024-06-21
fix: use ItemInNs::Macros to convert ModuleItem to ItemInNs

### LINKED ISSUE #17425: Some custom snippet definitions do not work
- `rustc` version: `rustc 1.79.0 (129f3b996 2024-06-10)`
- `rust-analyzer` version: `rust-analyzer 0.3.1995-standalone (c86d6230f 2024-06-10)`
- `vscode` version:
```
Version: 1.90.0 (Universal)
Commit: 89de5a8d4d6205e5b11647eb6a74844ca23d2573
Date: 2024-06-04T19:34:44.157Z
Electron: 29.4.0
ElectronBuildId: 9593362
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Darwin x64 23.5.0
```
- Zed version:
```
Zed: v0.140.4 (Zed Preview)
OS: macOS 14.5.0
Memory: 32 GiB
Architecture: x86_64
```

### Discussed in https://github.com/rust-lang/rust-analyzer/discussions/17354

<div type='discussions-op-text'>

<sup>Originally posted by **failable** June  6, 2024</sup>
This snippet work in VSCode and Zed.

```
              "derive (serde)": {
                "prefix": ["ds"],
                "body": ["#[derive(Serialize, Deserialize)]$0"],
                "requires": ["serde::Serialize", "serde::Deserialize"],
                "scope": "item"
              },
```

image

--------------------------

This does not work in VSCode and Zed
```
              "thiserror": {
                "prefix": ["te"],
                "body": [
                  "#[derive(Error, Debug)]",
                  "pub enum ${1:Error} {",
                  "\t$0",
                  "}"
                ],
                "requires": ["thiserror::Error"],
                "scope": "item"
              }
```
image

-------------------------
However, this one works

```
          "thiserror": {
            "prefix": ["te"],
            "body": [
              "#[derive(Error, Debug)]",
              "pub enum ${1:Error} {",
              "\t$0",
              "}"
            ],
            "requires": ["serde::Serialize", "serde::Deserialize"],// For testing
            "scope": "item"
          }
```

image

Why this `thiserror` dependency does not work? Thanks.</div>

roife:
fix #17425.

When converting `PathResolution` to `ItemInNs`, we should convert `ModuleDef::Macro` to `ItemInNs::Macros` to ensure that it can be found in `DefMap`.

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -2784,6 +2784,7 @@ impl From<ModuleDef> for ItemInNs {
             ModuleDef::Static(_) | ModuleDef::Const(_) | ModuleDef::Function(_) => {
                 ItemInNs::Values(module_def)
             }
+            ModuleDef::Macro(it) => ItemInNs::Macros(it),
             _ => ItemInNs::Types(module_def),
         }
     }


roife:
I'm not sure whether it's correct since I am not very familiar with Ns. What's the difference between `ItemInNs::Types(ModuleDef::Macro)` and `ItemInNs::Macros` 👀

lnicola:
> What's the difference between ItemInNs::Types(ModuleDef::Macro) and ItemInNs::Macros?👀

Not completely sure, but I think hte first one is a type defined in a macro expansion, while the second is a macro?

Veykril:
`ItemInNs` carries the namespace this resolution is coming from/supposed to be treated as being in. So `ItemInNs::Types(ModuleDef::Macro)` means a `Macro` in a type namespace context which is technically non-sensical, as macros can only exist in the macro namespace. The reason for this `ItemInNs` system is that some things can be in different namespaces, a unit struct for example can be in the type namespace given its a type but also in the value namespace as it can refer to its unit constructor.

roife:
Thanks, I get it. Then this PR should be OK.

PR #17483, opened 2024-06-23
minor : fixes for ratoml module

alibektas:
This is a follow-up PR to #17058.

- Parse errors are reflected as such by defining a new variant called `ConfigError::ParseError`
- New error collection has been added to store config level agnostic errors.

EDIT : Some things that this PR promised to solve are removed and will be addressed by other PRs

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -892,7 +892,6 @@ impl Config {
                         SnippetScopeDef::Type => SnippetScope::Type,
                         SnippetScopeDef::Item => SnippetScope::Item,
                     };
-                    #[allow(clippy::single_match)]
                     match Snippet::new(
                         &def.prefix,
                         &def.postfix,
@@ -2833,7 +2832,7 @@ fn get_field<T: DeserializeOwned>(
         })
 }
 
-fn get_field_toml<T: DeserializeOwned>(
+fn get_field_toml<T: DeserializeOwned + fmt::Debug>(
     toml: &toml::Table,
     error_sink: &mut Vec<(String, toml::de::Error)>,
     field: &'static str,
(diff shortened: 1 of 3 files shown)


alibektas:
https://github.com/rust-lang/rust-analyzer/blob/bd9758bd5d780ab41068aa6e9c3ec0b3d3d593db/crates/rust-analyzer/src/config.rs#L3610-L3612

I would also like to ( since it is relevant to this issue ) ask why this is supposed to be an error. If my assumption is correct it is related to the hyphenated key. If that's the case toml actually allows such [keys](https://toml.io/en/v1.0.0#keys).

crates/rust-analyzer/src/config.rs
```diff
@@ -758,6 +766,35 @@ impl Config {
...
+                for (name, def) in snips.iter() {
+                    if def.prefix.is_empty() && def.postfix.is_empty() {
+                        continue;
+                    }
+                    let scope = match def.scope {
+                        SnippetScopeDef::Expr => SnippetScope::Expr,
+                        SnippetScopeDef::Type => SnippetScope::Type,
+                        SnippetScopeDef::Item => SnippetScope::Item,
+                    };
+                    #[allow(clippy::single_match)]
```
Veykril, line 780: This shouldn't be triggering here?

crates/rust-analyzer/src/config.rs
```diff
@@ -758,6 +766,35 @@ impl Config {
...
+                    match Snippet::new(
+                        &def.prefix,
+                        &def.postfix,
+                        &def.body,
+                        def.description.as_ref().unwrap_or(name),
+                        &def.requires,
+                        scope,
+                    ) {
+                        Some(snippet) => config.snippets.push(snippet),
```
Veykril, line 789: we need to clear the snippets in here once, see https://github.com/rust-lang/rust-analyzer/issues/17485

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -1038,4 +1041,43 @@ impl GlobalState {
             .finish();
         Ok(())
     }
+
+    fn fetch_aux_files(&mut self) {
```
Veykril, line 1045: This is still not the right way to handle this. We need to make sure these files are registered to the file watcher, which will then handle the rest for us instead of us loading the files once at startup.

Veykril:
> https://github.com/rust-lang/rust-analyzer/blob/bd9758bd5d780ab41068aa6e9c3ec0b3d3d593db/crates/rust-analyzer/src/config.rs#L3610-L3612
> 
> I would also like to ( since it is relevant to this issue ) ask why this is supposed to be an error. If my assumption is correct it is related to the hyphenated key. If that's the case toml actually allows such [keys](https://toml.io/en/v1.0.0#keys).


Sprry forgot to reply to this. the error here is that `sysroot` isn't a table config, so the  the `non-table` key being used here should error

crates/rust-analyzer/src/config.rs
```diff
@@ -798,6 +797,13 @@ pub struct Config {
     /// Clone of the value that is stored inside a `GlobalState`.
     source_root_parent_map: Arc<FxHashMap<SourceRootId, SourceRootId>>,
 
+    /// Use case : It is an error to have an empty value for `check_command`.
+    /// Since it is a `global` command at the moment, its final value can only be determined by
+    /// traversing through `global` configs and the `client` config. However the non-null value constraint
+    /// is config level agnostic, so this requires an independent error storage
+    /// FIXME : bad name I know...
+    other_errors: ConfigErrors,
```
Veykril, line 805: `validation_errors`?

PR #17487, opened 2024-06-24
internal: Some more small memory optimizations

Veykril:
Not a big impact on metrics, though there are some more savings in queries mainly used by the IDE layer from this

### DIFF
--- a/crates/hir/src/term_search/expr.rs
+++ b/crates/hir/src/term_search/expr.rs
@@ -209,7 +209,7 @@ impl Expr {
                 }
             }
             Expr::Variant { variant, generics, params } => {
-                let generics = non_default_generics(db, (*variant).into(), generics);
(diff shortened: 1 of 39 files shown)


crates/hir/src/term_search/expr.rs
```diff
@@ -209,7 +209,7 @@ impl Expr {
                 }
             }
             Expr::Variant { variant, generics, params } => {
-                let generics = non_default_generics(db, (*variant).into(), generics);
+                let generics = non_default_generics(db, variant.parent_enum(db).into(), generics);
```
Veykril, line 212: @kilpkonn heads up but the logic here is/was wrong. `non_default_generics` only queries the generics of the definition, not the parent so querying it for an enum variant always resulted in an empty set of generics. Hence why this now changes the test outputs for generics in the assist below. (technically this is not more correct)

PR #17488, opened 2024-06-24
minor : Fix duplicate snippets showing up on hover.

### LINKED ISSUE #17485: Duplicate method completion from custom snippets
image

I noticed there is an option `rust-analyzer.completion.snippets.custom` for this:

https://github.com/rust-lang/rust-analyzer/blob/2fd803cc13dc11aeacaa6474e3f803988a7bfe1a/docs/user/generated_config.adoc#L356

But I manually set `rust-analyzer.completion.snippets.custom` an empty `{}`, these duplicated completions still come out.

alibektas:
With each `config::apply_change` duplicate configs were being added.
Now we first drain the vec that holds these and then start adding. This fixes #17485

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -844,6 +844,9 @@ impl Config {
             config.source_root_parent_map = source_root_map;
         }
 
+        // IMPORTANT : This holds as long as ` completion_snippets_custom` is declared `client`.
+        config.snippets = vec![];
+
         let snips = self.completion_snippets_custom().to_owned();
 
         for (name, def) in snips.iter() {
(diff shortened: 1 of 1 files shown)


crates/rust-analyzer/src/config.rs
```diff
@@ -844,6 +844,9 @@ impl Config {
             config.source_root_parent_map = source_root_map;
         }
 
+        // IMPORTANT : This holds as long as ` completion_snippets_custom` is declared `client`.
+        config.snippets = vec![];
```
Veykril, line 848:
```suggestion
        config.snippets.clear();
```
No need to allocate a fresh vec

PR #17492, opened 2024-06-25
Inspect named constants for `DefMap` collecting during method resolution

### LINKED ISSUE #17031: non-exhaustive pattern false positive
**rust-analyzer version**: v0.3.1916 (taken from the VSCode plugin page, the given command does nothing on my machine)

**editor or extension**:  VSCode with rust-analyzer v0.3.1916

**relevant settings**: NA

**repository link (if public, optional)**: NA

```rust
pub trait Foo {
    type Bar;
    fn foo(bar: Self::Bar);
}

pub struct FooImpl;

const foo_impl: () = {
    impl Foo for FooImpl {
        type Bar = ();

        fn foo(_bar: Self::Bar) {
            let () = _bar; // non-exhaustive pattern: `_` not covered reported here
        }
    }
};
```
This code compiles fine with rustc.

Any of the following changes make the error go away:
1. Renaming `_bar` by `bar`
2. Replacing `fn foo(_bar: Self::Bar)` by `fn foo(_bar: ())`
3. Removing the `const foo_impl: () = {...}` wrapper

The original issue for me is when using proptest_derive, I get the error with this code:

```rust
#[derive(Arbitrary)]
pub enum Foo {
    Bar,
    Baz
}
```
The generated code looks similar to the repro above.

ShoyuVanilla:
Fixes #17031 

Besides that issue, this won't fix the cases like;

```rust
pub trait Foo {
    type Bar;
    fn foo(bar: Self::Bar);
}

pub struct FooImpl;

fn foo_impl() {
    impl Foo for FooImpl {
        type Bar = ();

        fn foo(_bar: Self::Bar) {
            let () = _bar; // non-exhaustive pattern: `_` not covered reported here
        }
    }
}
```

which simply replaces the const block in the example code of the original issue with a function 😢 

Rust Analyzer doesn't inspect all the blocks and bodies but only some "closer" ones for implementation searching and I think that makes sense since for IDE, we cannot wait very long time for such full-search.
But sometimes this makes some false-positive diagnostics like in the above code or some failures in associated type inferences.

I wonder what is the proper way to handle such cases.
Maybe we can just do the same as now for the "first pass", and if there are some inference failures or diagnostics, (rollback to a snapshot if needed and) search more blocks and bodies 🤔

### DIFF
--- a/crates/hir-ty/src/method_resolution.rs
+++ b/crates/hir-ty/src/method_resolution.rs
@@ -210,9 +210,21 @@ impl TraitImpls {
                 map.entry(target_trait).or_default().entry(self_ty_fp).or_default().push(impl_id);
             }
 
-            // To better support custom derives, collect impls in all unnamed const items.
+            // To better support custom derives, collect impls in all named and unnamed const items.
+            // const FOO: () = { ... };
             // const _: () = { ... };
-            for konst in module_data.scope.unnamed_consts() {
+            for konst in module_data
+                .scope
+                .declarations()
+                .filter_map(|def| {
+                    if let hir_def::ModuleDefId::ConstId(konst) = def {
+                        Some(konst)
+                    } else {
+                        None
+                    }
+                })
+                .chain(module_data.scope.unnamed_consts())
+            {
                 let body = db.body(konst.into());
                 for (_, block_def_map) in body.blocks(db.upcast()) {
                     Self::collect_def_map(db, map, &block_def_map);
--- a/crates/ide-diagnostics/src/handlers/non_exhaustive_let.rs
+++ b/crates/ide-diagnostics/src/handlers/non_exhaustive_let.rs
@@ -44,4 +44,28 @@ fn main() {
 "#,
         );
     }
+
+    #[test]
+    fn regression_issue_17031() {
+        check_diagnostics(
+            r#"
+pub trait Foo {
+    type Bar;
+    fn foo(bar: Self::Bar);
+}
+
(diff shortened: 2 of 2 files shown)


Veykril:
Hm, that case you mentioned here we should be able to support though, generally we consider all top level impls + any impls from the current block and parent blocks (searching upwards is fine as that is still cheap and not global), which means we should be able to resolve this example. So we aren't considering relevant block impls here somewhere.

crates/hir-ty/src/method_resolution.rs
```diff
@@ -210,9 +210,21 @@ impl TraitImpls {
...
+                .declarations()
+                .filter_map(|def| {
+                    if let hir_def::ModuleDefId::ConstId(konst) = def {
+                        Some(konst)
+                    } else {
+                        None
+                    }
+                })
+                .chain(module_data.scope.unnamed_consts())
+            {
```
Veykril, line 227: This is intentional, `_` is special cased as a macro pattern to support here. Considering all consts for this is too expensive hence why this is done. Future rust will also lint against this "impl in non-`_` const" pattern

ShoyuVanilla: Oh, thanks. Then I'll try in another way

Veykril:
The `TraitEnv` when we solve for `_bar`'s type here should definitely contain the `foo_impl` block, which ought to respect its containing impls then. Anything else here is a bug.

PR #17494, opened 2024-06-25
do not normalize `use foo::{self}` to `use foo`

harrysarson:
It changes behaviour and can cause collisions. E.g. for the following snippet

```rs
mod foo {

    pub mod bar {}

    pub const bar: i32 = 8;
}

// transforming the below to `use foo::bar;` causes the error:
//
//   the name `bar` is defined multiple times
use foo::bar::{self};

const bar: u32 = 99;

fn main() {
    let local_bar = bar;
}
```

we still normalize

```rs
use foo::bar;
use foo::bar::{self};
```

to `use foo::bar;` because this cannot cause collisions.

See: https://github.com/rust-lang/rust-analyzer/pull/17140#issuecomment-2079189725

### DIFF
--- a/crates/ide-assists/src/handlers/merge_imports.rs
+++ b/crates/ide-assists/src/handlers/merge_imports.rs
@@ -490,6 +490,25 @@ use foo::bar;
     }
 
     #[test]
+    fn test_merge_nested_empty_and_self_with_other() {
+        check_assist(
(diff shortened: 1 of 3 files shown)


harrysarson:
cc: @Veykril

Veykril (approved):

PR #17495, opened 2024-06-25
pass cargo extra args when debugging

### LINKED ISSUE #17128: VSCode: Debug run does not respect `rust-analyzer.runnables.extraArgs`
**rust-analyzer version**: 0.3.1932-standalone
**rustc version**: rustc 1.79.0-nightly (7f2fc33da 2024-04-22)

**editor or extension**: VSCode - rust-analyzer v0.3.1932

**relevant settings**: `rust-analyzer.cargo.target`

**repository link (if public, optional)**: [Minimal reproducible example](https://github.com/schurwanzn-stabl/ra-issue1-mre) (needs nightly toolchain, but that's not relevant to the issue itself)

**code snippet to reproduce**: Code snippet does not suffice to reproduce, see linked repository

The MRE (minimal reproducible example) enables `-Z build-std` via it's `.cargo/cargo.toml`. The build-std feature requires that any cargo build/test/run/... command be called with the `--target=...` option  - compilation fails without it. This makes it ideal to showcase the issue.
The `--target x86_64-unknown-linux-gnu` option is set by `rust-analyzer.runnables.extraArgs`. Using, for example, code lens to *Run* the test or the main function of the MRE will work as expected (try it without the `rust-analyzer.runnables.extraArgs` setting - it will fail).
However, running the *Debug* code lens will fail with `error: -Zbuild-std requires --target`.

---

Comparing the relevant code for debug running
https://github.com/rust-lang/rust-analyzer/blob/47a901b9bf1f99b1ec5222d478684fc412d526a5/editors/code/src/debug.ts#L179-L188
with the equivalent code for non-debug running
https://github.com/rust-lang/rust-analyzer/blob/47a901b9bf1f99b1ec5222d478684fc412d526a5/editors/code/src/run.ts#L144-L153
seems to confirm that `rust-analyzer.runnables.extraArgs` is ignored on debug runs.

listochkin:
fixes #17128

### DIFF
--- a/editors/code/src/debug.ts
+++ b/editors/code/src/debug.ts
@@ -5,7 +5,7 @@ import type * as ra from "./lsp_ext";
 
 import { Cargo, getRustcId, getSysroot } from "./toolchain";
 import type { Ctx } from "./ctx";
-import { prepareEnv } from "./run";
+import { createCargoArgs, prepareEnv } from "./run";
 import { isCargoRunnableArgs, unwrapUndefinable } from "./util";
 
 const debugOutput = vscode.window.createOutputChannel("Debug");
@@ -180,7 +180,8 @@ async function getDebugExecutable(
     env: Record<string, string>,
 ): Promise<string> {
     const cargo = new Cargo(runnableArgs.workspaceRoot || ".", debugOutput, env);
-    const executable = await cargo.executableFromArgs(runnableArgs.cargoArgs);
+    const args = createCargoArgs(runnableArgs);
+    const executable = await cargo.executableFromArgs(args);
 
     // if we are here, there were no compilation errors.
     return executable;


bittailor:
Hi

When I currently use the prerelease rust-analyzer VsCode extension (v0.4.2025) I'm not able anymore to debug indivitual tests via the 'debug link' above the test.

image

it fails with 
```
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.00s
     Running unittests src/lib.rs (target/debug/deps/vs_debug_issue-c35e030d2028369f)
error: Unrecognized option: 'message-format'
error: test failed, to rerun pass `-p vs-debug-issue --lib`
```

With the release version (v0.3.2020) it still works.

I think I tracked it down to the change in this PR as it adds `--` into the arguments list that is passed to the find the test artifact while later (in [toolchian.ts](https://github.com/rust-lang/rust-analyzer/blob/f2afcb874e7410121c366ae601660abe327e320b/editors/code/src/toolchain.ts#L29) `--message-format=json` is added to the end of the arguments list.

So currently on master the argument list looks like the following which has the `--message-format=json` after the `--` and in my view causes the issue:
image
 
 while running the release tag v0.3.2020 the argument list looks like 
 
 
image

Shelim:
Confirming, the issue is happening in the wild and there are very few hits on Google yet.

To clarify (TL;DR): the current workaround is to downgrade `rust-analyser` in VsCode to `Release version` via extension menu

PR #17496, opened 2024-06-26
minor: Bump `actions/download-artifact` and `upload-artifact`

lnicola:
These will stop working in a couple of days. I think we could still use them in a more efficient way, but more tweaking is needed for that.


lnicola:
Test release at https://github.com/rust-lang/rust-analyzer/releases/tag/nightly

Issue #17497, opened 2024-06-26
Invalid RA diagnostic error: expected 2 arguments, found 1


mckenfra:
`rust-analyzer version: 0.3.2011-standalone (2fd803cc1 2024-06-22)`
`rustc 1.78.0 (9b00956e5 2024-04-29)`
`VSCode, Version 0.3.2011, Server Version 0.3.2011-standalone (2fd803cc1 2024-06-22)`

While working on a rust project in VSCode, I noticed that RA was reporting errors with perfectly valid code. After stripping back all surrounding code, I have come up with the below minimal example that reproduces the error.

The error message I see in VSCode is: `expected 2 arguments, found 1` rust-analyzer [E0107](https://doc.rust-lang.org/stable/error_codes/E0107.html)

This occurs on the line `vec.push("hello");`

```rust
#![allow(dead_code)]
use core::ops::Deref;

trait Trait1 {
    type Assoc: Deref<Target = String>;
}
trait Trait2: Trait1 {
}
trait Trait3 {
    type T1: Trait1;
    type T2: Trait2;
    fn push(&self, x: bool, y: bool);
}

fn problem_method<T: Trait3>() {
    let mut vec = Vec::new();
    vec.push("hello"); // Rustc ok, RA errors (mismatched args)
}
```

lnicola:
Fixed in #17511.

Issue #17233, opened 2024-05-14
rust-analyzer seems to be getting confused between trait methods and inherent methods


TDplai:
**rust-analyzer version**: rust-analyzer 1.78.0 (9b00956 2024-04-29)

Can also reproduce with rust-analyzer 1.80.0-nightly (ab14f94 2024-05-13)

**rustc version**: rustc 1.78.0 (9b00956e5 2024-04-29)

**editor or extension**: Neovim 0.9.5

**relevant settings**:

- `CARGO_HOME` is set to `$HOME/.local/share/cargo`
- `RUSTUP_HOME` is set to `$HOME/.local/share/rustup`

**code snippet to reproduce**: 
```rust
pub trait A {
    // Removing this associated type makes everything work
    type X: B;
}
// Making `B` not a subtrait of `A` makes everything work
pub trait B: A {
    // Changing this to take `&self` makes everything work
    fn confused_name(self, _: i32);
}

pub struct Foo;
impl Foo {
    // Changing this to take `self` by value makes everything work
    pub fn confused_name(&self) {}
}

pub fn repro<T: A>() {
    // rust-analyzer reports "expected 1 argument, found 0"
    Foo.confused_name();
}
```

Note that the above code does compile.

dfireBird:
This seems like a problem with chalk, I traced it to this `trait_solve_query` call: https://github.com/rust-lang/rust-analyzer/blob/6524922b1b1cfdf16d1e08785dec03f352f693a2/crates/hir-ty/src/method_resolution.rs#L1180

Which here should return `None` but returns a `Some` and it correctly returns `None` if any of the changes mentioned in the example code. I'm not sure what more info should I add here, if there is anything needed please ask.

flodiebold:
Probably https://github.com/rust-lang/chalk/issues/750.

PR #17506, opened 2024-06-28
#17497 - Invalid RA diagnostic error: expected 2 arguments, found 1

mckenfra:
Fix for #17497 and  #17233

The issue occurs because in some configurations of traits where one of them has `Deref` as a supertrait, RA's type inference algorithm fails to fully resolve a trait's self type, and instead uses a `TyKind::BoundVar`. This then erroneously matches with types that do not implement the trait. In other words, unconnected types appear to inherit the trait's methods.

The fix is to insert a heuristic when invoking `lookup_method()` that allows the caller to determine if the returned method is a good match. If the returned method would result in diagnostic errors, the caller instructs the algorithm to continue iterating the possible methods to see if a better one exists (i.e. one that won't result in diagnostic errors).

Includes a unit test that only passes after applying the fixes in this commit.

### DIFF
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -378,6 +378,7 @@ impl InferenceContext<'_> {
                     &param_tys,
                     &indices_to_skip,
                     is_varargs,
+                    false,
                 );
                 self.normalize_associated_types_in(ret_ty)
             }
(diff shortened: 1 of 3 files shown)


flodiebold:
I don't think this is a good approach. We want to match rustc's method resolution, if we did a heuristic like this we would miss cases where rustc will report an error. It's unfortunate that there's a bug in trait resolution, but that doesn't mean that we should work around it by making method resolution inaccurate.

mckenfra:
@flodiebold Thank you, your comment prompted me to read Rust's method resolution algorithm described in [Method-call expressions](https://doc.rust-lang.org/reference/expressions/method-call-expr.html#method-call-expressions), which I was not aware of before. The RA code does appear to follow the Rust guidelines with respect to the order in which the method candidates are considered. I agree with you that my solution is not a good approach, so I will close this pull request.

FYI - the reason I worked on this pull request was because I was frustrated that my Rust project was showing errors in VSCode, and yet it was compiling perfectly fine using rustc!

PR #17511, opened 2024-06-29
#17497 - Invalid RA diagnostic error: expected 2 arguments, found 1

mckenfra:
Fix for #17497

The issue occurs because in some configurations of traits where one of them has `Deref` as a supertrait, RA's type inference algorithm fails to resolve the `Deref::Target` type, and instead uses a `TyKind::BoundVar` (i.e. an unknown type). This "autoderefed" type then incorrectly acts as if it implements all traits in scope.

The fix is to re-apply the same sanity-check that is done in [`iterate_method_candidates_with_autoref()`](https://github.com/rust-lang/rust-analyzer/blob/9463d9eea4b87e651e7d8ed8425a9c92f23b1cdf/crates/hir-ty/src/method_resolution.rs#L1008), that is: don't try to resolve methods on unknown types. This same sanity-check is now done on each autoderefed type for which trait methods are about to be checked. If the autoderefed type is unknown, then the iterating of the trait methods for that type is skipped.

Includes a unit test that only passes after applying the fixes in this commit.

Includes a change to the assertion count in test `syntax_highlighting::tests::benchmark_syntax_highlighting_parser` as suggested by Lukas Wirth during review.

Includes a change to the sanity-check code as suggested by Florian Diebold during review.

### DIFF
--- a/crates/hir-ty/src/method_resolution.rs
+++ b/crates/hir-ty/src/method_resolution.rs
@@ -30,8 +30,8 @@ use crate::{
     utils::all_super_traits,
     AdtId, Canonical, CanonicalVarKinds, DebruijnIndex, DynTyExt, ForeignDefId, GenericArgData,
     Goal, Guidance, InEnvironment, Interner, Mutability, Scalar, Solution, Substitution,
-    TraitEnvironment, TraitRef, TraitRefExt, Ty, TyBuilder, TyExt, TyKind, VariableKind,
-    WhereClause,
+    TraitEnvironment, TraitRef, TraitRefExt, Ty, TyBuilder, TyExt, TyKind, TyVariableKind,
+    VariableKind, WhereClause,
 };
 
 /// This is used as a key for indexing impls.
@@ -1083,6 +1083,11 @@ fn iterate_method_candidates_by_receiver(
     table.run_in_snapshot(|table| {
         let mut autoderef = autoderef::Autoderef::new(table, receiver_ty.clone(), true);
         while let Some((self_ty, _)) = autoderef.next() {
+            if matches!(self_ty.kind(Interner), TyKind::InferenceVar(_, TyVariableKind::General)) {
+                // don't try to resolve methods on unknown types
+                return ControlFlow::Continue(());
+            }
+
             iterate_trait_method_candidates(
                 &self_ty,
                 autoderef.table,
--- a/crates/hir-ty/src/tests/method_resolution.rs
+++ b/crates/hir-ty/src/tests/method_resolution.rs
@@ -2099,3 +2099,42 @@ fn test() {
 "#,
     );
 }
+
+#[test]
+fn mismatched_args_due_to_supertraits_with_deref() {
+    check_no_mismatches(
+        r#"
+//- minicore: deref
+use core::ops::Deref;
+
+trait Trait1 {
+    type Assoc: Deref<Target = String>;
+}
+
+trait Trait2: Trait1 {
+}
+
+trait Trait3 {
+    type T1: Trait1;
+    type T2: Trait2;
+    fn bar(&self, x: bool, y: bool);
+}
+
+struct Foo;
+
+impl Foo {
+    fn bar(&mut self, _: &'static str) {}
+}
+
+impl Deref for Foo {
+    type Target = u32;
+    fn deref(&self) -> &Self::Target { &0 }
+}
+
+fn problem_method<T: Trait3>() {
+    let mut foo = Foo;
+    foo.bar("hello"); // Rustc ok, RA errors (mismatched args)
+}
+"#,
+    );
+}
--- a/crates/ide/src/syntax_highlighting/tests.rs
+++ b/crates/ide/src/syntax_highlighting/tests.rs
@@ -1238,7 +1238,7 @@ fn benchmark_syntax_highlighting_parser() {
             })
             .count()
     };
-    assert_eq!(hash, 1169);
+    assert_eq!(hash, 1167);
 }
 
 #[test]


crates/hir-ty/src/method_resolution.rs
```diff
@@ -1058,6 +1058,8 @@ fn iterate_method_candidates_by_receiver(
     mut callback: &mut dyn FnMut(ReceiverAdjustments, AssocItemId, bool) -> ControlFlow<()>,
 ) -> ControlFlow<()> {
     let receiver_ty = table.instantiate_canonical(receiver_ty);
+    let is_receiver_ty_unknown = TyFingerprint::for_trait_impl(&receiver_ty).is_none();
```
flodiebold, line 1061: this seems a really roundabout way to check for unknown? if we mean unbound type variables, just `receiver_ty.inference_var(Interner).is_some()` should do it after the instantiate_canonical, or `receiver_ty.is_general_var(Interner, &receiver_ty.binders)` before it. ... but actually, if the receiver_ty was unknown, we wouldn't reach this point in the first place because of the early return in `iterate_method_candidates_with_autoref`, would we? I'm not sure I understand the logic here. (`TyFingerprint::for_trait_impl(&receiver_ty).is_none()` just includes a bunch of other type kinds that probably shouldn't be?)

crates/hir-ty/src/method_resolution.rs
```diff
@@ -1079,6 +1081,14 @@ fn iterate_method_candidates_by_receiver(
     table.run_in_snapshot(|table| {
         let mut autoderef = autoderef::Autoderef::new(table, receiver_ty.clone(), true);
         while let Some((self_ty, _)) = autoderef.next() {
+            if !is_receiver_ty_unknown {
+                let canonical_self_ty = autoderef.table.canonicalize(self_ty.clone());
+                if canonical_self_ty.value.is_general_var(Interner, &canonical_self_ty.binders) {
```
flodiebold, line 1086: I think here that can just be `self_ty.inference_var(Interner).is_some()` then?

crates/hir-ty/src/method_resolution.rs
mckenfra: Unfortunately, using either of your suggestions here causes the following test to fail:

```
---- syntax_highlighting::tests::benchmark_syntax_highlighting_parser stdout ----
syntax highlighting parser: 4.64s, 0b
thread 'syntax_highlighting::tests::benchmark_syntax_highlighting_parser' panicked at crates/ide/src/syntax_highlighting/tests.rs:1231:5:
assertion `left == right` failed
  left: 1167
 right: 1169


failures:
    syntax_highlighting::tests::benchmark_syntax_highlighting_parser
```

In fact, I only introduced this code involving `is_receiver_ty_unknown` because, having made my other changes, I found that this exact test was failing. I did some investigation involving comparing the test data generated by this test before and after applying my other changes. I found that `TyFingerprint::for_trait_impl(&receiver_ty).is_none()` was sufficient to ensure this test passed.

crates/hir-ty/src/method_resolution.rs
mckenfra: Unfortunately, when I replace my `canonical_self_ty.value.is_general_var(Interner, &canonical_self_ty.binders)` with your suggested `self_ty.inference_var(Interner).is_some()` here, the following tests (which were passing before) now fail:

```
test consteval::tests::trait_method ... FAILED
test tests::traits::trait_method_with_scalar_fallback ... FAILED
test tests::traits::variable_kinds_2 ... FAILED
```

crates/hir-ty/src/method_resolution.rs
Veykril: That test failing is generally fine I believe. The assertion there is not really meaningful for the most part in terms of correctness as there are a bunch of unresolved types in the fixture of it anyways. I am actually not even sure why we have the assert in there in the first place, as its meant as a benchmark.

mckenfra: Ok sure, I will remove this code involving `is_receiver_ty_unknown` entirely. I will instead update the assertion count in the test as you suggest.

crates/hir-ty/src/method_resolution.rs
flodiebold: Maybe try `autoderef.table.resolve_ty_shallow(&self_ty).inference_var(Interner).is_some()` then. I don't think canonicalizing makes sense here.

mckenfra: Using `autoderef.table.resolve_ty_shallow(&self_ty).inference_var(Interner).is_some()` the same 3 tests still fail:

```
failures:

---- consteval::tests::trait_method stdout ----
thread 'consteval::tests::trait_method' panicked at crates/hir-ty/src/consteval/tests.rs:583:5:
Error in evaluating goal: UnresolvedMethod("f")

---- tests::traits::trait_method_with_scalar_fallback stdout ----
thread 'tests::traits::trait_method_with_scalar_fallback' panicked at crates/hir-ty/src/tests/traits.rs:4328:5:
assertion `left == right` failed: type annotation differs at 195..196
  left: "{unknown}"
 right: "i32"

---- tests::traits::variable_kinds_2 stdout ----
thread 'tests::traits::variable_kinds_2' panicked at crates/hir-ty/src/tests/traits.rs:3244:5:
assertion `left == right` failed: type annotation differs at 99..106
  left: "{unknown}"
 right: "u128"


failures:
    consteval::tests::trait_method
    tests::traits::trait_method_with_scalar_fallback
    tests::traits::variable_kinds_2
```

flodiebold: Ok, what you actually need is `matches!(self_ty.kind(Interner), TyKind::InferenceVar(_, TyVariableKind::General))`, because integer or float variables can actually have methods.

mckenfra: Yes that works. I will make the change as you suggest and update the pull request.

lnicola:
Can you rebase or something? This looks like it includes some newer changes.

lnicola:
Sorry, it doesn't, I was looking at the wrong diff.

lnicola:
This does seem to introduce a couple of extra unknown types and one type mismatch with `borrow`:

> crates/hir-expand/src/files.rs 270:57-270:84: Expected SyntaxNode<RustLanguage>, got &{unknown}

PR #17513, opened 2024-06-29
fix: completions after async kw

### LINKED ISSUE #17500: No completions offered after async keyword
_This issue is a follow-up of #17452_

**rust-analyzer version**: Using Neovim, so the version released in 2024-06-24.

**rustc version**: (eg. output of `rustc -V`) 1.79.0 (129f3b996 2024-06-10)

**editor or extension**: (eg. VSCode, Vim, Emacs, etc. For VSCode users, specify your extension version; for users of other editors, provide the distribution if applicable) Neovim, but the issue also repros in VS Code.

**Reproduction steps**:
1. In a Rust file type `async f`.
2. Request completions.
3. Notice how there are no completions from rust-analyzer, while the expectation is for items like `fn` to be provided.

roife:
fix #17500 

### Changes

1. fix completions after async kw
2. fix completions for `async` kw in trait

### DIFF
--- a/crates/ide-completion/src/completions/item_list.rs
+++ b/crates/ide-completion/src/completions/item_list.rs
@@ -82,17 +82,30 @@ fn add_keywords(acc: &mut Completions, ctx: &CompletionContext<'_>, kind: Option
     let no_vis_qualifiers = ctx.qualifier_ctx.vis_node.is_none();
     let in_block = kind.is_none();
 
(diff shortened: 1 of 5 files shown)


Veykril (approved):
Ideally we'd also honor the required order of keywords, `async unsafe` is valid whereaas `unsafe async` is not

PR #17518, opened 2024-06-30
fix: Fix expression scope calculation when within macro expansions

Veykril:


Cargo.toml
```diff
@@ -10,9 +10,7 @@ license = "MIT OR Apache-2.0"
 authors = ["rust-analyzer team"]
 
 [profile.dev]
-# Disabling debug info speeds up builds a bunch,
-# and we don't rely on it for debugging that much.
-debug = 0
```
Veykril, line 15: Imo the comment here is just wrong, I frequently need to enable this when debugging panics within unit tests :)

This ultimately requires to recompile everything all the time which means it doesn't speed up builds at all.

PR #17523, opened 2024-07-01
Add an option to use "::" for the external crate prefix.

### LINKED ISSUE #11823: Option to have auto-imports of items from external crates use global `::`-prefixed paths
I would find it very useful to have an option to enable auto-import on an ident like `File` to always generate the following `use` statement, for consistency and explicitness that we're pulling from the extern prelude:

```rs
use ::std::fs::File;
```

wada314:
Fixes #11823 .
Hi I'm very new to rust-analyzer and not sure how the review process are. Can somebody take a look at this PR? thanks!


crates/rust-analyzer/src/config.rs
```diff
@@ -359,6 +359,8 @@ config_data! {
         imports_preferPrelude: bool                       = false,
         /// The path structure for newly inserted paths to use.
         imports_prefix: ImportPrefixDef               = ImportPrefixDef::Plain,
+        /// Whether to prefix external (including std, core) crate imports with `::`. e.g. "use ::std::io::Read;".
+        imports_prefix_external_crate: bool = false,
```
Veykril, line 363: This name collides with the above (one being the `prefix` object, the other being a field in said object, we should probably sanity check that those configs don't exist in a test) which will cause issues, `imports_prefixExternalPrelude` would be better

wada314: Thanks for taking a look at my PR! Renamed it.

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17533, opened 2024-07-02
minor: Don't show layout info when hovering on locals

### LINKED ISSUE #17096: Hover should only show type layout when hovering the definition, not a reference
See https://github.com/rust-lang/rust-analyzer/pull/16825#issuecomment-2058225309

Wilfred:
This is useful when the cursor is on the type name, but layout information tends to distract when the cursor is just on a local variable of that type.

Fixes #17096

### DIFF
--- a/crates/ide/src/hover/render.rs
+++ b/crates/ide/src/hover/render.rs
@@ -499,9 +499,6 @@ pub(super) fn definition(
         Definition::TypeAlias(it) => {
             render_memory_layout(config.memory_layout, || it.ty(db).layout(db), |_| None, |_| None)
         }
-        Definition::Local(it) => {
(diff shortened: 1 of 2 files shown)


Veykril:
That doesn't address the issue. I like the layout info on locals a lot in fact as it is the easiest way to check the size of some arbitrary instantiated type.

As that linked issue states, `Hover should only show type layout when hovering the definition, not a reference`. This just removes hovers for locals though which is something completely different.

Veykril:
To add to this, we should differentiate here https://github.com/rust-lang/rust-analyzer/blob/cbcb9779f52a5d7f2de6272432077dc21a3524f1/crates/ide/src/hover.rs#L224-L238
If we deal with a `IdentClass::NameClass` we should render the layout info, otherwise we should omit it

Veykril:
I'll close this for now as I outlined what we should be doing instead

PR #17542, opened 2024-07-04
feat: go-to-def and find-references on control-flow keywords

### LINKED ISSUE #17517: More IDE features for control flow keywords
We currently highlight relevant control flow structures when the cursor is on a relevant keyword, an example being highlighting all `await`s when on an async fn's `async` keyword, or all loop control flow keywords when it is on the loops keyword. We can do more things here I'd say. `Go to def/decl` on `return` could go the relevant closure/fn, `continue` and `break` to the corresponding loop (or block for labeled blocks), go to references could do the same as the highlighting

roife:
fix #17517.

This PR implements **go-to-definition** and **find-references** functionalities for control flow keywords, which is similar to the behaviors in the `highlight-related` module. Besides, this PR also fixes some incorrect behaviors in `highlight-related`.

## Changes

1. **Support for go-to-definition on control flow keywords**:  
   This PR introduces functionality allowing users to navigate on the definition of control flow keywords (`return`, `break`, `continue`).
   Commit: 2a3244ee147f898dd828c06352645ae1713c260f..7391e7a608634709db002a4cb09229de4d12c056.

2. **Bug fixes and refactoring in highlight-related**:
   - **Handling return/break/continue within try_blocks**:  
     This PR adjusted the behavior of these keywords when they occur within `try_blocks`. When encounter these keywords, the program should exit the outer function or loop which containing the `try_blocks`, rather than the `try_blocks` itself; while the `?` will cause the program to exit `try_blocks`.  
     Commit: 59d697e807f0197f59814b37dca1563959da4aa1.
   - **Support highlighting keywords in macro expansion for highlight-related**:  
     Commit: 88df24f01727c23a667a763ee3ee0cec22d5ad52.
   - Detailed description for the bug fixes
     + The previous implementation of `preorder_expr` incorrectly treated `try_blocks` as new contexts, thereby r-a will not continue to traverse inner `return` and `break/continue` statements. To resolve this, a new function `preorder_expr_with_ctx_checker` has been added, allowing users to specify which expressions to skip.
       * For example, when searching for the `?` in the context, r-a should skip `try_blocks` where the `?` insides just works for `try_blocks`. But when search for the `return` keyword, r-a should collect both the `return` keywords inside and outside the `try_blocks`
     + Thus, this PR added `WalkExpandedExprCtx` (builder pattern). It offers the following improvements: customizable context skipping, maintenance of loop depth (for `break`/`continue`), and handling macro expansion during traversal.

3. **Support for find-references on control flow keywords**:  
   This PR enables users to find all references to control flow keywords.  
   Commit: 9202a33f81218fb9c2edb5d42e6b4de85b0323a8.

### DIFF
--- a/crates/ide/src/goto_definition.rs
+++ b/crates/ide/src/goto_definition.rs
@@ -2,10 +2,10 @@ use std::{iter, mem::discriminant};
 
 use crate::{
     doc_links::token_as_doc_comment, navigation_target::ToNav, FilePosition, NavigationTarget,
-    RangeInfo, TryToNav, UpmappingResult,
+    RangeInfo, TryToNav,
 };
 use hir::{
-    AsAssocItem, AssocItem, DescendPreference, InFile, MacroFileIdExt, ModuleDef, Semantics,
+    AsAssocItem, AssocItem, DescendPreference, FileRange, InFile, MacroFileIdExt, ModuleDef, Semantics
 };
 use ide_db::{
     base_db::{AnchoredPath, FileLoader},
@@ -14,11 +14,12 @@ use ide_db::{
     FileId, RootDatabase,
 };
 use itertools::Itertools;
+
 use syntax::{
-    ast::{self, HasLoopBody, Label},
+    ast::{self, HasLoopBody},
     match_ast, AstNode, AstToken,
     SyntaxKind::*,
-    SyntaxToken, TextRange, T,
+    SyntaxNode, SyntaxToken, TextRange, T,
 };
 
 // Feature: Go to Definition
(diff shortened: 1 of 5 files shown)


crates/ide/src/goto_definition.rs
```diff
@@ -190,6 +202,200 @@ fn try_filter_trait_item_definition(
...
+        None
+    }
+
+    let token_kind = token.kind();
+    sema.descend_into_macros(DescendPreference::None, token.clone())
+        .into_iter()
+        .filter_map(|descended| {
+            let file_id = sema.hir_file_for(&descended.parent()?);
+
+            // Try to find the function in the macro file
```
Veykril, line 290: I think you want to go via `Semantics::token_ancestors_with_macros` instead,

crates/ide/src/goto_definition.rs
```diff
@@ -190,6 +202,200 @@ fn try_filter_trait_item_definition(
...
+            _ => None,
+        }
+    };
+
+    sema.descend_into_macros(DescendPreference::None, token.clone())
+        .into_iter()
+        .filter_map(|descended| {
+            let file_id = sema.hir_file_for(&descended.parent()?);
+
+            // Try to find the function in the macro file
```
Veykril, line 378: same here

crates/ide/src/goto_definition.rs
```diff
@@ -190,6 +202,200 @@ fn try_filter_trait_item_definition(
...
+                    let for_tok = for_.for_token()?.into();
+                    let nav = NavigationTarget::from_expr(db, InFile::new(file_id, expr), for_tok);
+                    return Some(nav);
+                }
+                ast::Expr::BlockExpr(blk)
+                    if blk.label().is_some() && label_matches(blk.label()) =>
+                {
+                    let expr = ast::Expr::BlockExpr(blk.clone());
+                    let lbl_tok = blk.label().unwrap().lifetime()?.lifetime_ident_token()?.into();
+                    let nav = NavigationTarget::from_expr(db, InFile::new(file_id, expr), lbl_tok);
```
Veykril, line 355: here we can navigate to the label directly

crates/ide/src/navigation_target.rs
```diff
@@ -153,6 +153,22 @@ impl NavigationTarget {
         )
     }
 
+    pub(crate) fn from_expr(
```
Veykril, line 156: Hm actually I think this is the wrong way to go about this, let's construct these manually in the relevant modules and instead expose the `orig_range_with_focus` functions to the rest of the crate

crates/ide/src/highlight_related.rs
```diff
@@ -517,6 +531,20 @@ fn find_defs(sema: &Semantics<'_, RootDatabase>, token: SyntaxToken) -> FxHashSe
...
+fn original_range(
+    db: &dyn db::ExpandDatabase,
+    file_id: HirFileId,
+    text_range: Option<TextRange>,
+) -> Option<TextRange> {
+    if text_range.is_none() || !file_id.is_macro() {
+        return text_range;
+    }
+
+    InFile::new(file_id, text_range.unwrap())
```
Veykril, line 543:
```rs
    InFile::new(file_id, text_range?)
```
the file check shouldn't be necessary, the method call after already does that

crates/ide/src/goto_definition.rs
```diff
@@ -190,6 +202,200 @@ fn try_filter_trait_item_definition(
...
+                        // For async token, we navigate to itself, which triggers
+                        // VSCode to find the references
+                        let focus_token = if matches!(cursor_token_kind, T![async]) {
+                            fn_.async_token()?
+                        } else {
+                            fn_.fn_token()?
+                        };
+                        let focus_range = InFile::new(file_id, focus_token.text_range())
+                            .original_node_file_range_opt(db)
+                            .map(|(frange, _)| frange.range);
```
Veykril, line 249: Needs to filter by `frange.file_id == nav.file_id`

crates/ide/src/highlight_related.rs
```diff
@@ -506,6 +531,101 @@ fn find_defs(sema: &Semantics<'_, RootDatabase>, token: SyntaxToken) -> FxHashSe
...
+        .original_node_file_range_opt(db)
+        .map(|(frange, _)| frange.range)
```
Veykril, line 545: Needs to filer by `frange.file_id == highlight_references::file_id` (that is the file_id of the file the feature is running on)

Veykril (commented):
One thing to look out for is the turning `FileRange` into `TextRange` after upmapping. All of those need to verify that the file id of the `FileRange` is in fact the same as the file we are doing the feature stuff in, as `include!` has its input in a different file than where the call is.

crates/ide/src/goto_definition.rs
roife: Actually they are different. Here's an example:

```rust
macro_rules! M {
    ($blk:expr) => {
        fn f1() {
            $blk
        }

        $blk
    };
}

fn main() {
    M!({ return; /* goto-def on this return*/ });
}
```

When we perform on the return in the macro call-site, it should navigate to both `f1` and `main`. So we should take all the occurrence of `return` (both real files and macro files) into consideration.

Veykril: Yes but that is covered by iterating through all the `descend_into_macros` results. The `return` token should yield two tokens in `descend_into_macros`, and on each we'd call `token_ancestors_with_macros` which ought to yield us the 2 functions

roife: Oh, I get it. I misunderstood what you meant earlier.

crates/ide/src/navigation_target.rs
roife: So, should I implement `keywords_to_nav` in the relevant module, rather than directly calling `orig_range_with_focus_r` in `from_expr`?

Veykril: No let's put that into the calling module as well, not here I'd say

roife: If we move `from_expr` to the mod `goto_definition`, we will have to make `from_syntax` and `orig_range_with_focus_r` be `pub(crate)`, which seems a bit strange. 👀

roife:
I squashed some commits related to `goto-def` and updated the PR description for simplicity.

Veykril (approved):

PR #17544, opened 2024-07-05
feat: add inlay hints for generic parameters

### LINKED ISSUE #11091: inlay-hints for generic arguments
We should offer the ability to show inlay hints for generic arguments, just like we do for function arguments already.
```rs
fn foo<'a, T, const C: usize>() {}
foo::<'b, u32, 0>()
// should render as something like
foo::<'a: 'b, T: u32, C: 0>()
```
Each of these 3 should have a config flag, with only the const generics one being enabled by default I'd say.

MikeWalrus:
fixes #11091

By default, only hints for const generic parameters are shown, and this can be configured through `rust-analyzer.inlayHints.genericParameterHints.enable`.

Probably needs more testing.

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -2885,19 +2885,6 @@ fn field_props(field: &str, ty: &str, doc: &[&str], default: &str) -> serde_json
                 "Only show discriminant hints on fieldless enum variants."
             ]
         },
-        "GenericParameterHintsDef" => set! {
-            "type": "string",
-            "enum": [
-                "always",
-                "never",
-                "const_only"
-            ],
-            "enumDescriptions": [
-                "Always show generic parameter hints.",
-                "Never show generic parameter hints.",
-                "Only show const generic parameter hints."
-            ]
-        },
         "AdjustmentHintsModeDef" => set! {
             "type": "string",
             "enum": [
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -10,9 +10,9 @@ use dirs::config_dir;
 use flycheck::{CargoOptions, FlycheckConfig};
 use ide::{
     AssistConfig, CallableSnippets, CompletionConfig, DiagnosticsConfig, ExprFillDefaultMode,
-    HighlightConfig, HighlightRelatedConfig, HoverConfig, HoverDocFormat, InlayFieldsToResolve,
-    InlayHintsConfig, JoinLinesConfig, MemoryLayoutHoverConfig, MemoryLayoutHoverRenderKind,
-    Snippet, SnippetScope, SourceRootId,
+    GenericParameterHints, HighlightConfig, HighlightRelatedConfig, HoverConfig, HoverDocFormat,
+    InlayFieldsToResolve, InlayHintsConfig, JoinLinesConfig, MemoryLayoutHoverConfig,
+    MemoryLayoutHoverRenderKind, Snippet, SnippetScope, SourceRootId,
 };
 use ide_db::{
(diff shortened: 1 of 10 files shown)


crates/ide/src/inlay_hints/param_name/generic.rs
```diff
@@ -0,0 +1,288 @@
...
+
+use crate::{inlay_hints::GenericParameterHints, InlayHint, InlayHintsConfig, InlayKind};
+
+use super::{is_argument_similar_to_param_name, render_label};
+
+pub(crate) fn hints(
+    acc: &mut Vec<InlayHint>,
+    sema: &hir::Semantics<'_, RootDatabase>,
+    config: &InlayHintsConfig,
+    node: SyntaxNode,
```
Veykril, line 16:
```suggestion
    node: AnyHasGenericArgs,
```
[`c08d419` (#17555)](https://github.com/rust-lang/rust-analyzer/pull/17555/commits/c08d419fba1c7ad3c6e80f38d9412942c8516b9a) adds that, we should use it here (will need a rebase)

crates/ide/src/inlay_hints/param_name/generic.rs
```diff
@@ -0,0 +1,288 @@
...
+                    let const_param = ast::ConstParam::cast(syntax)?;
+                    const_param.name()?.syntax().clone()
+                }
+                hir::GenericParam::LifetimeParam(it) => {
+                    if const_generics_only {
+                        return None;
+                    }
+                    sema.source(it)?.value.syntax().clone()
+                }
+            };
```
Veykril, line 65: Querying the source can be expensive, so lets move this down to where we actually need it so we give the other filtering things a run first

crates/ide/src/inlay_hints/param_name/generic.rs
```diff
@@ -0,0 +1,288 @@
...
+
+            if should_hide {
+                return None;
+            }
+
+            let linked_location = sema.original_range_opt(&source_syntax);
+            let label = render_label(param_name, config, linked_location);
+            Some(InlayHint {
+                range: sema.original_range_opt(arg.syntax())?.range,
```
Veykril, line 82: Likewise lets move this in front of the `let source_syntax` (after that was moved down) as its a cheaper filter

crates/ide/src/inlay_hints/param_name/generic.rs
```diff
@@ -0,0 +1,288 @@
...
+                position: crate::InlayHintPosition::Before,
+                pad_left: false,
+                pad_right: true,
+                kind: InlayKind::Parameter,
```
Veykril, line 86: Let's add a new one for this, GenericParameter` and map that to `PARAMETER`

Veykril (commented):
Generally looks pretty good!

Veykril (commented):
A test for when the user supplied mismatching args wrt to the params might be nice as well

MikeWalrus:
@Veykril I've pushed again. Please let me know if I miss anything. :)

crates/rust-analyzer/src/config.rs
```diff
@@ -2874,6 +2885,19 @@ fn field_props(field: &str, ty: &str, doc: &[&str], default: &str) -> serde_json
...
+                "always",
+                "never",
+                "const_only"
+            ],
+            "enumDescriptions": [
+                "Always show generic parameter hints.",
+                "Never show generic parameter hints.",
+                "Only show const generic parameter hints."
+            ]
+        },
```
Veykril, line 2900:
```suggestion
```

Veykril:
Looks good, thanks!
@bors r+

Issue #4604, opened 2020-05-25
Client Changes


matklad:
This issue exists to announce changes which might affect rust-analyzer plugins for different editors. 
If you maintain such a plugin, consider subscribing to this issue. 

All our protocol extensions ~~are~~ will be documented at [lsp-extensions.md](https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/lsp-extensions.md). I'll try to remember to leave a comment on this issue if I modify that file.

matklad:
After https://github.com/rust-analyzer/rust-analyzer/pull/4632, all our current extensions are documented. There were a bunch of minor changes here and there. Most notably some requests were moved from `rust-analyzer/foo` to `experimental/foo`, in hope that this might be useful for other servers, and to mark the extensions we intend to upstream to the protocol itself.

Notably, inlay hints and runnables are are still unchanged: i want to upstream these two as well, but I am not too happy with their current state, and I don't have time to polish them right now. So i opted for documenting them as is, and cleaning up later.

matklad:
https://github.com/rust-analyzer/rust-analyzer/pull/4688 documents our `initilizationOptions`.

matklad:
#4710 changes API for runnables. The idea is that it's the client who is reponsible for running stuff, to make it possible to, eg, inject debugger support or other custom tools. So, rust-analzyer provides a more abstract representation of a cargo command, to make it easier for the client to make sense of it.

matklad:
#4717 adds API for lazily computing code action edits.

matklad:
#5188 adds status notification.

matklad:
adds "memory usage" command: https://github.com/rust-analyzer/rust-analyzer/pull/5244

matklad:
https://github.com/rust-analyzer/rust-analyzer/pull/4972 will switch to gzipping release artifacts.

matklad:
rust-analyzer is available via rustup

```
15:04:42|~/projects/rust-analyzer/editors/code|master✓
λ rustup component add --toolchain nightly rust-analyzer-preview
info: component 'rust-analyzer-preview' for target 'x86_64-unknown-linux-gnu' is up to date

15:04:46|~/projects/rust-analyzer/editors/code|master✓
λ rustup run nightly rust-analyzer --version
rust-analyzer f5a4a4b
```

matklad:
https://github.com/rust-analyzer/rust-analyzer/pull/5782

[...]

PR #17547, opened 2024-07-06
internal: Clean up runnable lsp extension

Veykril:
This feels like a natural addition to me, and also allows us to drop the expect-test hardcoding from the extension. Additionally, `cargoExtraArgs` is pointless, all the client will do is merge it with `cargoArgs` so the server can do that instead of delegating that to the client.


docs/dev/lsp-extensions.md
```diff
@@ -376,12 +376,29 @@ rust-analyzer supports two `kind`s of runnables, `"cargo"` and `"shell"`. The `a
 
 ```typescript
 {
+    /**
+     * Environment variables to set before running the command.
+     */
+    environment?: Record<string, string>;
```
Veykril, line 382: Open to alternative names, should we be explicit with `environmentVariables`?

docs/dev/lsp-extensions.md
```diff
@@ -376,12 +376,29 @@ rust-analyzer supports two `kind`s of runnables, `"cargo"` and `"shell"`. The `a
...
+    /**
+     * The working directory to run the command in.
+     */
+    cwd: string;
```
Veykril, line 386: I feel like this being mandatory gives a more unique experience across clients. We should not leave this decision to clients as that could result in runnables sometimes working sometimes not. (notable it is required by the shell version already)

davidbarsky (approved):
this looks good and reasonable to me.

docs/dev/lsp-extensions.md
davidbarsky: could also call it `envVariables`?

docs/dev/lsp-extensions.md
davidbarsky: yeah, reducing the number of possible options is great. the combinatorial explosion of options made this really difficult to reason about.

ia0:
I'm a bit confused by https://github.com/rust-lang/rust-analyzer/issues/4604#issuecomment-2211788144, in particular:

> gets rid of `cargoExtraArgs` and bakes that into `cargoArgs`

Does this refer to `rust-analyzer.cargo.extraArgs`? I don't see any `rust-analyzer.cargo.args` in the [manual](https://rust-analyzer.github.io/manual.html). Or does this refer to something else?

davidbarsky:
> I'm a bit confused by [#4604 (comment)](https://github.com/rust-lang/rust-analyzer/issues/4604#issuecomment-2211788144), in particular:
> 
> > gets rid of `cargoExtraArgs` and bakes that into `cargoArgs`
> 
> Does this refer to `rust-analyzer.cargo.extraArgs`? 

It does not. It refers to the field in the runnable sent from the server to the client.

Veykril:
It refers to this https://github.com/rust-lang/rust-analyzer/blob/78d5f05e704dfe54ac70806973876608af3e2ba3/docs/dev/lsp-extensions.md#L382 which is the `rust-analyzer.cargo.extraArgs` config yes. `rust-analyzer.cargo.args` isn't a thing, as the `cargoArgs` in the runnable is populated by the server depending on the task, its not something the user configures.

docs/dev/lsp-extensions.md
Veykril: Eh, I'll stick with `environment` I think.

davidbarsky: fair enough! not terribly attached.

ia0:
Thanks a lot for the clarification!

PR #17549, opened 2024-07-06
fix: Fix runnables being incorrectly constructed

Veykril:
I've misunderstood parts of the code here which caused runnables to arbitrarily break :) (I have yet to understand the conditions that made them break though, there is some odd caching involved I feel like ...)
Fixes https://github.com/rust-lang/rust-analyzer/issues/17402

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -323,14 +323,6 @@
             {
                 "title": "general",
                 "properties": {
-                    "rust-analyzer.cargoRunner": {
-                        "type": [
-                            "null",
-                            "string"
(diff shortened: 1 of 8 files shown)


editors/code/package.json
```diff
@@ -323,14 +323,6 @@
             {
                 "title": "general",
                 "properties": {
-                    "rust-analyzer.cargoRunner": {
```
Veykril, line 326: Also removes this, the custom runner was broken for ages already so clearly no one is depending on it

PR #17558, opened 2024-07-07
fix: Fix double rounding of `f32` literals

### LINKED ISSUE #17556: `f32` literals are sometimes double rounded
**rust-analyzer version**: 0.3.2020-standalone

**rustc version**: rustc 1.79.0 (129f3b996 2024-06-10)

**editor or extension**: VSCode, extension version v0.3.2020

**relevant settings**: N/A

```rust
const FOO: f32 = 1.9999999403953552;
```

The correctly rounded `f32` here is `1.9999999` (as can be seen when hovering over the literal itself), but when hovering over `FOO`, `const FOO: f32 = 2.0` is displayed instead, which is the result of `1.9999999403953552_f64 as f32` instead of `1.9999999403953552_f32`.

beetrees:
Fixes #17556 by delaying parsing until the type is known. Also adds a test to check the issue is fixed.

### DIFF
--- a/crates/hir-def/src/hir.rs
+++ b/crates/hir-def/src/hir.rs
@@ -56,29 +56,29 @@ pub struct Label {
 }
 pub type LabelId = Idx<Label>;
 
-// We convert float values into bits and that's how we don't need to deal with f32 and f64.
-// For PartialEq, bits comparison should work, as ordering is not important
+// We leave float values as a string to avoid double rounding.
+// For PartialEq, string comparison should work, as ordering is not important
 // https://github.com/rust-lang/rust-analyzer/issues/12380#issuecomment-1137284360
-#[derive(Default, Debug, Clone, Copy, Eq, PartialEq)]
-pub struct FloatTypeWrapper(u64);
+#[derive(Default, Debug, Clone, Eq, PartialEq)]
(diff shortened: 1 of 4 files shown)


crates/hir-def/src/hir.rs
```diff
@@ -56,29 +56,29 @@ pub struct Label {
...
 
-// We convert float values into bits and that's how we don't need to deal with f32 and f64.
-// For PartialEq, bits comparison should work, as ordering is not important
+// We leave float values as a string to avoid double rounding.
+// For PartialEq, string comparison should work, as ordering is not important
 // https://github.com/rust-lang/rust-analyzer/issues/12380#issuecomment-1137284360
-#[derive(Default, Debug, Clone, Copy, Eq, PartialEq)]
-pub struct FloatTypeWrapper(u64);
+#[derive(Default, Debug, Clone, Eq, PartialEq)]
+pub struct FloatTypeWrapper(String);
```
Veykril, line 63:
```suggestion
pub struct FloatTypeWrapper(Box<str>);
```
let's use a boxed string to shrink the type a bit (and ensure no left over allocation)

Veykril:
Thanks!
@bors r+

PR #17559, opened 2024-07-07
Encode ident rawness and literal kind separately in tt::Leaf

Veykril:


Veykril:
@lnicola this PR is now blocked on us syncing with rust-lang/rust as this touches the proc-macro stuff causing CI to run on nightly 😬 so would be nice if we could do a sync by next week, the PR isn't finished yet and once it is I will only merge it at the start of the week. Just letting you know :)

lnicola:
Yup, will do one soon!

crates/proc-macro-api/src/msg.rs
```diff
@@ -19,8 +19,10 @@ pub const VERSION_CHECK_VERSION: u32 = 1;
 pub const ENCODE_CLOSE_SPAN_VERSION: u32 = 2;
 pub const HAS_GLOBAL_SPANS: u32 = 3;
 pub const RUST_ANALYZER_SPAN_SUPPORT: u32 = 4;
+/// Whether literals encode their kind as an additional u32 field and idents their rawness as a u32 field
+pub const EXTENDED_LEAF_DATA: u32 = 5;
 
-pub const CURRENT_API_VERSION: u32 = RUST_ANALYZER_SPAN_SUPPORT;
+pub const CURRENT_API_VERSION: u32 = EXTENDED_LEAF_DATA;
```
Veykril, line 25: We should be able to start dropping support for some proc-macro server API versions soon should the code here become too complicated.

Veykril:
I reckon this PR will break stuff, will install the modified proc-macro server and use that for the week to see if something is breaking there as well

crates/proc-macro-api/src/msg.rs
```diff
@@ -19,8 +19,10 @@ pub const VERSION_CHECK_VERSION: u32 = 1;
 pub const ENCODE_CLOSE_SPAN_VERSION: u32 = 2;
 pub const HAS_GLOBAL_SPANS: u32 = 3;
 pub const RUST_ANALYZER_SPAN_SUPPORT: u32 = 4;
+/// Whether literals encode their kind as an additional u32 field and idents their rawness as a u32 field
+pub const EXTENDED_LEAF_DATA: u32 = 5;
```
Veykril, line 23: Headsup @vlad20012, this PR and https://github.com/rust-lang/rust-analyzer/pull/17601 make changes to the proc-macro server API adding a new version, adding two more u32s to literals on the wire, and one more u32 to idents. I have not really given this much thought tbf so please comment if you think we can do better than that. Its not set in stone until the next sync to rustc which will update the proc-macro server then.

lnicola: Note that syncs are currently on hold until we figure out the new `rustc_pattern_analysis` story.

Veykril: that's fine, probably better even as it gives me more time to test the proc-macro server :D

crates/tt/src/lib.rs
```diff
@@ -145,23 +203,85 @@ pub struct Punct<S> {
...
+    /// by a non-punctuation token.
+    ///
+    /// Converts to `proc_macro::Spacing::Alone`, but
+    /// `proc_macro::Spacing::Alone` converts back to `token::Spacing::Alone`.
+    /// Because of that, pretty-printing of `TokenStream`s produced by proc
+    /// macros is unavoidably uglier (with more whitespace between tokens) than
+    /// pretty-printing of `TokenStream`'s produced by other means (i.e. parsed
+    /// source code, internally constructed token streams, and token streams
+    /// produced by declarative macros).
+    JointHidden,
```
vlad20012, line 263: The changes in Spacing seems unrelated to the wire changes :thinking:

crates/proc-macro-api/src/msg.rs
vlad20012: @Veykril Thanks for mentioning! THe changes looks good to me

crates/tt/src/lib.rs
Veykril: Yes that is not yet used, and iirc it won't ever make it to the proc-macro server either so you should be safe to ignore that. For you only the changes in proc-macro-api ought to be really relevant.

lnicola:
@Veykril did this break `cargo xtask install`?

Veykril:
Ah no I didn't

PR #17561, opened 2024-07-07
Add --keep-going to the check command

mo8it:
Fixes https://github.com/rust-lang/rustlings/issues/1628

@Veykril I am not sure about what you meant with "unconditionally" in https://github.com/rust-lang/rustlings/issues/1628#issuecomment-2212492230, but I didn't find out how to get the version of the toolchain anyway to do a check like in [this snippet](https://github.com/rust-lang/rust-analyzer/blob/a5b21ea0aa644dffd7cf958b43f11f221d53404e/crates/project-model/src/build_scripts.rs#L125-L127). Is this check even required if rust-analyzer was installed with the toolchain?

`--keep-going` was [stabilized in 1.74](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-174-2023-11-16)

### DIFF
--- a/crates/flycheck/src/lib.rs
+++ b/crates/flycheck/src/lib.rs
@@ -426,6 +426,8 @@ impl FlycheckActor {
                     }
                 }
 
+                cmd.arg("--keep-going");
+
                 options.apply_on_command(&mut cmd);
                 (cmd, options.extra_args.clone())
             }


Veykril:
The check isn't require as we don't support toolchains that old anyways (it is impossible to reliably support a big range of rust versions)
Thanks!
@bors r+

lnicola:
Yeah, we can probably drop it, it's been a while.

huntc:
> The check isn't require as we don't support toolchains that old anyways (it is impossible to reliably support a big range of rust versions) Thanks! @bors r+

This has caused a problem with a couple of my projects that are pinned to 1.72 of Rust. Is there a way we can make this conditional e.g. don't include the `check` if the compiler is pre-1.74?

Please see https://github.com/rust-lang/rust-analyzer/issues/17662.

devanlai:
> > The check isn't require as we don't support toolchains that old anyways (it is impossible to reliably support a big range of rust versions) Thanks! @bors r+
> 
> This has caused a problem with a couple of my projects that are pinned to 1.72 of Rust. Is there a way we can make this conditional e.g. don't include the `check` if the compiler is pre-1.74?
> 
> Please see #17662.
> 
> I also understand that there's no strict backward-compability statement for RA, but Zed and VSC don't appear to honour rust-toolchain rust-analyzer components, so we're a bit stuck. Also see [zed-industries/zed#4883 (comment)](https://github.com/zed-industries/zed/issues/4883#issuecomment-2241984180)

I ran into the same issue with being pinned to an older rust version. For VSCode at least, I've found that using [rust-analyzer's config](https://rust-analyzer.github.io/manual.html#toolchain) to use the stable toolchain instead seems to work fine, presuming that for analysis purposes, the stable toolchain is fine.

Veykril:
> but Zed and VSC don't appear to honour rust-toolchain rust-analyzer components, so we're a bit stuck. 

That is a good point to raise, we should be able to support the toolchain component in an overwrite if its installed imo. Though that is the job of the client as that is the entity responsible for spawning the process. We can do this in VSCode but Zed would need to do that themselves.

Veykril:
I'll try to look into the VSCode thing here https://github.com/rust-lang/rust-analyzer/issues/17663, zed can then probably try to do something similar

PR #17562, opened 2024-07-07
Delete `CloneAny` from `rust-analyzer`'s fork of `AnyMap`

WaffleLapkin:
...because it's very sketchy and ~~causes~~ will cause FCWs (this is a blocker for https://github.com/rust-lang/rust/pull/120248).
In this case it *is* actually sound, but still.

I should write a better fork of anymap...

### DIFF
--- a/crates/stdx/src/anymap.rs
+++ b/crates/stdx/src/anymap.rs
@@ -68,8 +68,6 @@ pub type RawMap<A> = hash_map::HashMap<TypeId, Box<A>, BuildHasherDefault<TypeId
 /// The type parameter `A` allows you to use a different value type; normally you will want
 /// it to be `core::any::Any` (also known as `std::any::Any`), but there are other choices:
 ///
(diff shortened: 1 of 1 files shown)


compiler-errors:
Question for r-a ppl: How often does r-a do subtree syncs with rust-lang/rust?

I asked for this to be separated in another PR just because it felt better not to modify r-a in a rust-lang/rust PR (and so it could be r+'d by an r-a reviewer), but if subtree sync is gonna block landing rust-lang/rust#120248, I wonder if it's better to just do this in-tree. 😸

Veykril:
We tend to do them on a weekly basis though it has been some while now. @lnicola does them, I don't know when they are planning on doing the next one. I think it should be fine to land this in-tree first though (to prevent a sync from blocking that PR), there are no plans to touch this part of the code on our side for the time being.

lnicola (approved):

compiler-errors:
OK, let's close this PR. I will approve the one in the rustc repo.

PR #17564, opened 2024-07-08
minor: Fix comment in generated file

lnicola:


Veykril:
Did I break something for the releases again? 😅

PR #17565, opened 2024-07-08
Remove version check before using `--keep-going`

mo8it:
See https://github.com/rust-lang/rust-analyzer/pull/17561#issuecomment-2214227971


Veykril:
Thanks!
@bors r+

PR #17571, opened 2024-07-09
feat: do not add new enum if it already exists

winstxnhdw:
## Summary

This PR introduces a check for the existence of another enum within the current scope, and if it exist, we skip `add_enum_def`.

## Why?

Currently, when using the `bool_to_enum` assist more than once, it is possible to add multiple enum definitions. For example, the following snippet,

```rs
#[derive(PartialEq, Eq)]
enum Bool {
    True,
    False,
}

fn main() {
    let a = Bool::True;
    let b = true;
    println!("Hello, world!");
}
```

will be transformed into,

```rs
#[derive(PartialEq, Eq)]
enum Bool {
    True,
    False,
}

#[derive(PartialEq, Eq)]
enum Bool {
    True,
    False,
}

fn main() {
    let a = Bool::True;
    let b = Bool::True;
    println!("Hello, world!");
}
```

This can be annoying for users to clean up.

### DIFF
--- a/crates/ide-assists/src/handlers/bool_to_enum.rs
+++ b/crates/ide-assists/src/handlers/bool_to_enum.rs
@@ -461,7 +461,17 @@ fn add_enum_def(
     usages: &UsageSearchResult,
     target_node: SyntaxNode,
     target_module: &hir::Module,
-) {
+) -> Option<()> {
+    if ctx
+        .find_node_at_offset::<ast::SourceFile>()?
+        .syntax()
+        .children()
+        .filter_map(|node| ast::Enum::cast(node).and_then(|e| ctx.sema.to_def(&e)))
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/bool_to_enum.rs
```diff
@@ -461,7 +461,17 @@ fn add_enum_def(
     usages: &UsageSearchResult,
     target_node: SyntaxNode,
     target_module: &hir::Module,
-) {
+) -> Option<()> {
+    if ctx
```
Veykril, line 465: Instead of having this check by syntactic we ought to do it semantically, we can fetch the desf of the relevant module via `semantics.scope(&insert_before)?.module().scope(None)`, and then search through that for the `Bool` name. `insert_before` being the `let insert_before = node_to_insert_before(target_node);` local a few lines below

Issue #17451, opened 2024-06-19
Add `f16` and `f128` support (add to `FloatTy`)


tgross35:
`FloatTy` in rustc has `f16` and `f128` now. It would probably be pretty easy to update RA's `FloatTy` to match and propagate that out.

tgross35:
Probably pretty easy

@rustbot label +E-easy +A-hir

ChayimFriedman2:
Added now.

PR #17572, opened 2024-07-10
Add `f16` and `f128` support

beetrees:
Adds `f16` and `f128` support, using the `rustc_apfloat` library (also used by `rustc`) for parsing/arithmetic/displaying since the types aren't stable yet so can't be used by rust-analyzer itself.

Issue: #17451


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17580, opened 2024-07-11
feat: show pub visibility for closing_brace inlay hints

Young-Flash:
image

### DIFF
--- a/crates/ide/src/inlay_hints/closing_brace.rs
+++ b/crates/ide/src/inlay_hints/closing_brace.rs
@@ -6,7 +6,7 @@
 use hir::{HirDisplay, Semantics};
 use ide_db::{base_db::FileRange, RootDatabase};
 use syntax::{
-    ast::{self, AstNode, HasName},
+    ast::{self, AstNode, HasName, HasVisibility},
     match_ast, SyntaxKind, SyntaxNode, T,
 };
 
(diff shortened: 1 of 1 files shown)


crates/ide/src/inlay_hints/closing_brace.rs
```diff
@@ -188,6 +201,37 @@ fn f() {
...
   }
 //^ fn f
+"#,
+        );
+    }
+
+    #[test]
+    fn hints_closing_brace_pub() {
+        check_with_config(
+            InlayHintsConfig { closing_brace_hints_min_lines: Some(2), ..DISABLED_CONFIG },
```
Marcosmk12, line 211:
```suggestion
            InlayHintsConfig { closing_brace_hints_min_lines: Some(2), ..DISABLED_CONFIG },
```

Marcosmk12 (approved):

Veykril:
Is that useful? The name is the only relevant information that makes sense to show for the hint imo. The hint is mainly there to show which function you are currently in imo, so whether its public or not doesn't really matter here. (we aren't showing any other parts of the signature either for the same reason)

Young-Flash:
When I see "pub" in the inlay hint at the end of a function, I would know that this function might be used externally(without scroll up to the signature). If it's not there, it will only be used within the same file. Feel free to close it if it's not good for you.

PR #17584, opened 2024-07-12
Implement symbol interning infra

Veykril:
Will fix https://github.com/rust-lang/rust-analyzer/issues/15590

My unsafe-fu is not the best but it does satisfy miri.

There is still some follow up work to do, notably a lot of places using strings instead of symbols/names, most notably the token tree.


crates/intern/src/symbol.rs
```diff
@@ -0,0 +1,2209 @@
...
+            packed: unsafe {
+                NonNull::new_unchecked((r as *const &str).cast::<*const str>().cast_mut())
+            },
+        }
+    }
+
+    fn arc(arc: Arc<Box<str>>) -> Self {
+        Self {
+            packed: Self::pack_arc(
+                // Safety: `Arc::into_raw`` always return a non null pointer
```
lnicola, line 50: Nit:

```suggestion
                // Safety: `Arc::into_raw` always returns a non null pointer
```

crates/intern/src/symbol.rs
```diff
@@ -0,0 +1,293 @@
...
+    const fn non_arc(r: &&str) -> Self {
+        Self {
+            // SAFETY: The pointer is non-null as it is derived from a reference
+            // Ideally we would call out to `pack_arc` but for a `false` tag, unfortunately the
+            // packing stuff requires reading out the pointer to an integer which is not supported
+            // in const contexts, so here we make use of the fact that for the non-arc version the
+            // tag is false (0) and thus does not need touching the actual pointer value.ext)
+            packed: unsafe {
+                NonNull::new_unchecked((r as *const &str).cast::<*const str>().cast_mut())
+            },
```
Veykril, line 51: I am not too fond of this part as I'd really like to make sure to clear out the LSB but I am not aware of a way of doing this that does not involve transmuting the address to a usize which is disallowed in const contexts

crates/intern/src/symbol.rs
```diff
@@ -0,0 +1,293 @@
+//! Attempt at flexible symbol interning, allowing to intern and free strings at runtime while also
+//! supporting
```
lnicola, line 2: supporting?

crates/intern/src/symbol/symbols.rs
```diff
@@ -0,0 +1,237 @@
...
+                    let hash_ = hash_thing_(dashmap_.hasher(), &proxy_);
+                    let shard_idx_ = dashmap_.determine_shard(hash_ as usize);
+                    dashmap_.shards_mut()[shard_idx_].get_mut().raw_entry_mut().from_hash(hash_, |k| k == &proxy_).insert(proxy_, SharedValue::new(()));
+                )*
+            }
+            dashmap_
+        }
+    };
+}
+define_symbols! {
```
lnicola, line 41: Were these taken from somewhere?

Veykril: These are taken over from our `name!` macro which I'm currently replacing in a follow up commit

Veykril:
Oh dear I am getting a `STATUS_HEAP_CORRUPTION` when running analysis-stats

lnicola:
Just wondering, do we really need to support static strings, or is it more for the convenience of using `const`s?

Veykril:
Mostly the convenience I'd guess? Can't say how much of a perf difference it is in having the known symbols not participate in const re-allocation and ref counting

Veykril:
I also just wanted to try myself at a bit of unsafe again as it has been a while :) Though I am quite confused as to why this is failing here.

Veykril:
Okay found out where I was being stupid, now its failing on semantic changes instead

Veykril:
So this saves ~50mb of memory on `analysis-stats . --query-sysroot-metadata` for me, and that is with only `Name` changed to use `Symbol` (shrinking it down to 8 bytes, will be 12 when we get hygiene into this). I expect another chunk of savings by replacing `SmolStr` within `tt::Leaf` with `Symbol` as well

Veykril:
This revealed a few places where we relying on iteration order of hashmaps in tests, which is now very much apparent given the hashes rely on pointer values. (also I just realized I forgot switch branches when creating the PR)

Veykril:
Will tackle tt after https://github.com/rust-lang/rust-analyzer/pull/17559 to reduce conflicts

crates/hir-def/src/nameres/tests/macros.rs
```diff
@@ -1348,6 +1348,7 @@ fn proc_attr(a: TokenStream, b: TokenStream) -> TokenStream { a }
         .keys()
         .map(|name| name.display(&db).to_string())
         .sorted()
+        .sorted()
```
lnicola, line 1351: @Veykril typo?

Veykril: yes

Issue #17579, opened 2024-07-11
Prompt for name when using extract refactoring assists


joshka:
In typescript, extracting a variable / function / etc. allows you to enter a name for the just created extraction. It would be nice to do the same in RA for functions, modules, variables, etc.

<img width="430" alt="image" src="https://github.com/rust-lang/rust-analyzer/assets/381361/31b017af-0688-42de-a5c4-6c73844bb84c">
<img width="358" alt="image" src="https://github.com/rust-lang/rust-analyzer/assets/381361/cc2d8860-981c-4fc1-a738-b4e3a4c5f8e9">

Playing with the extract_variable:

```diff
if let Some(cap) = ctx.config.snippet_cap {
    if let Some(ast::Pat::IdentPat(ident_pat)) = let_stmt.pat() {
        if let Some(name) = ident_pat.name() {
-            edit.add_tabstop_before(cap, name);
+            edit.add_placeholder_snippet_group(
+                cap,
+                vec![name.syntax().clone(), name_expr.syntax().clone()],
            )
        }
    }
}
```

SourceChangeBuilder::add_placeholder_snippet_token doesn't seem quite right for this. Instead of a single edit location which applies to both places when the edit is confirmed, this method makes two edit locations that are edited simultaneously.

<img width="299" alt="image" src="https://github.com/rust-lang/rust-analyzer/assets/381361/b8a8c3fd-a06a-40c5-8f57-497b3d20cdf3">

I'm not quite sure what the right terminology to use to find the functionality here, so seeking some guidance on what to do to make this work.

DropDemBits:
It seems that the Typescript language server is sending a command to the client (in this case, VSCode) to trigger a rename dialog, as the dialog that shows up for renaming a symbol is shown by the client, which then sends the new name as part of a [`textDocument/rename`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_rename) request.

rust-analyzer does have custom commands (see https://github.com/rust-lang/rust-analyzer/blob/5577e4e3b1d87eb3a3e45b649ff73e021a10dba5/editors/code/src/main.ts#L113), but you'll have to look into the VSCode Typescript extension to see how they do it.

joshka:
Thanks for the guidance on where to look.

Looking at tsserver, it seems that every refactoring method returns (possibly empty) information on the file and location to use in presenting the rename command after the refactor finishes.

[...]

PR #17587, opened 2024-07-13
Trigger VSCode to rename after extract variable assist is applied

joshka:
When the user applies the "Extract Variable" assist, the cursor is
positioned at the newly inserted variable. This commit adds a command
to the assist that triggers the rename action in VSCode. This way, the
user can quickly rename the variable after applying the assist.

Fixes part of: #17579

https://github.com/user-attachments/assets/4cf38740-ab22-4b94-b0f1-eddd51c26c29

I haven't yet looked at the module or function extraction assists yet.

### DIFF
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -135,6 +135,7 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_>) -> Op
                             }
                         }
                     }
+                    edit.rename();
                 }
                 Anchor::Replace(stmt) => {
                     cov_mark::hit!(test_extract_var_expr_stmt);
@@ -149,6 +150,7 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_>) -> Op
                             }
                         }
                     }
+                    edit.rename();
                 }
                 Anchor::WrapInBlock(to_wrap) => {
                     let indent_to = to_wrap.indent_level();
@@ -192,6 +194,7 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_>) -> Op
                             }
                         }
                     }
+                    edit.rename();
 
                     // fixup indentation of block
                     block.indent(indent_to);
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -135,7 +135,6 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_>) -> Op
                             }
                         }
                     }
-                    edit.rename();
                 }
                 Anchor::Replace(stmt) => {
                     cov_mark::hit!(test_extract_var_expr_stmt);
@@ -150,7 +149,6 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_>) -> Op
                             }
                         }
                     }
-                    edit.rename();
                 }
                 Anchor::WrapInBlock(to_wrap) => {
                     let indent_to = to_wrap.indent_level();
@@ -194,12 +192,12 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_>) -> Op
                             }
                         }
                     }
-                    edit.rename();
 
                     // fixup indentation of block
                     block.indent(indent_to);
                 }
             }
+            edit.rename();
         },
     )
 }
--- a/crates/ide-db/src/assists.rs
+++ b/crates/ide-db/src/assists.rs
@@ -29,7 +29,16 @@ pub struct Assist {
     /// cumbersome, especially if you want to embed an assist into another data
     /// structure, such as a diagnostic.
     pub source_change: Option<SourceChange>,
-    pub trigger_signature_help: bool,
+    /// The command to execute after the assist is applied.
+    pub command: Option<Command>,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Command {
+    /// Show the parameter hints popup.
+    TriggerSignatureHelp,
+    /// Rename the just inserted item.
+    Rename,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
--- a/crates/rust-analyzer/src/lsp/to_proto.rs
+++ b/crates/rust-analyzer/src/lsp/to_proto.rs
@@ -13,7 +13,7 @@ use ide::{
     NavigationTarget, ReferenceCategory, RenameError, Runnable, Severity, SignatureHelp,
     SnippetEdit, SourceChange, StructureNodeKind, SymbolKind, TextEdit, TextRange, TextSize,
 };
-use ide_db::{rust_doc::format_docs, FxHasher};
+use ide_db::{assists, rust_doc::format_docs, FxHasher};
 use itertools::Itertools;
 use paths::{Utf8Component, Utf8Prefix};
 use semver::VersionReq;
@@ -1336,8 +1336,14 @@ pub(crate) fn code_action(
         command: None,
     };
 
-    if assist.trigger_signature_help && snap.config.client_commands().trigger_parameter_hints {
+    if assist.command == Some(assists::Command::TriggerSignatureHelp)
+        && snap.config.client_commands().trigger_parameter_hints
+    {
         res.command = Some(command::trigger_parameter_hints());
+    } else if assist.command == Some(assists::Command::Rename)
+        && snap.config.client_commands().rename
+    {
+        res.command = Some(command::rename());
     }
 
     match (assist.source_change, resolve_data) {
@@ -1712,6 +1718,14 @@ pub(crate) mod command {
         lsp_types::Command {
             title: "triggerParameterHints".into(),
             command: "rust-analyzer.triggerParameterHints".into(),
+            arguments: None,
+        }
+    }
+
+    pub(crate) fn rename() -> lsp_types::Command {
+        lsp_types::Command {
+            title: "rename".into(),
+            command: "rust-analyzer.rename".into(),
             arguments: None,
         }
     }
--- a/crates/rust-analyzer/src/lsp/to_proto.rs
+++ b/crates/rust-analyzer/src/lsp/to_proto.rs
@@ -1336,15 +1336,14 @@ pub(crate) fn code_action(
         command: None,
     };
 
-    if assist.command == Some(assists::Command::TriggerSignatureHelp)
-        && snap.config.client_commands().trigger_parameter_hints
-    {
-        res.command = Some(command::trigger_parameter_hints());
-    } else if assist.command == Some(assists::Command::Rename)
-        && snap.config.client_commands().rename
-    {
-        res.command = Some(command::rename());
-    }
+    let commands = snap.config.client_commands();
+    res.command = match assist.command {
+        Some(assists::Command::TriggerSignatureHelp) if commands.trigger_parameter_hints => {
+            Some(command::trigger_parameter_hints())
+        }
+        Some(assists::Command::Rename) if commands.rename => Some(command::rename()),
+        _ => None,
+    };
 
     match (assist.source_change, resolve_data) {
         (Some(it), _) => res.edit = Some(snippet_workspace_edit(snap, it)?),
--- a/editors/code/src/client.ts
+++ b/editors/code/src/client.ts
@@ -348,6 +348,7 @@ class ExperimentalFeatures implements lc.StaticFeature {
                     "rust-analyzer.showReferences",
                     "rust-analyzer.gotoLocation",
                     "editor.action.triggerParameterHints",
+                    "editor.action.rename",
                 ],
             },
             ...capabilities.experimental,
--- a/crates/ide-assists/src/assist_context.rs
+++ b/crates/ide-assists/src/assist_context.rs
@@ -185,11 +185,11 @@ impl Assists {
             return None;
         }
 
-        let mut trigger_signature_help = false;
+        let mut command = None;
         let source_change = if self.resolve.should_resolve(&id) {
             let mut builder = SourceChangeBuilder::new(self.file);
             f(&mut builder);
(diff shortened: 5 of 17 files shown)


crates/rust-analyzer/src/lsp/to_proto.rs
```diff
@@ -1336,8 +1336,14 @@ pub(crate) fn code_action(
         command: None,
     };
 
-    if assist.trigger_signature_help && snap.config.client_commands().trigger_parameter_hints {
+    if assist.command == Some(assists::Command::TriggerSignatureHelp)
```
DropDemBits, line 1339: nit: Can turn this into a `match` block

crates/ide-db/src/assists.rs
```diff
@@ -29,7 +29,16 @@ pub struct Assist {
...
+    /// The command to execute after the assist is applied.
+    pub command: Option<Command>,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Command {
+    /// Show the parameter hints popup.
+    TriggerSignatureHelp,
+    /// Rename the just inserted item.
+    Rename,
```
DropDemBits, line 41: bikeshedding: A better name might be something like `TriggerRename`, to match the naming of the trigger parameter signature help command.

crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -135,6 +135,7 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_>) -> Op
                             }
                         }
                     }
+                    edit.rename();
```
DropDemBits, line 138: nit: Can pull this out of the match block to after the match block

editors/code/src/client.ts
```diff
@@ -348,6 +348,7 @@ class ExperimentalFeatures implements lc.StaticFeature {
                     "rust-analyzer.showReferences",
                     "rust-analyzer.gotoLocation",
                     "editor.action.triggerParameterHints",
+                    "editor.action.rename",
```
DropDemBits, line 351: I imagine this would require also adding a comment to #4604
announcing this change, but I'm not 100% sure on the procedure for this.

DropDemBits (changes requested):
Ideally all of the extract & generate assists should still be using the `add_placeholder_snippet_group` for better out-of-the box support for other editor extensions, but that's an entirely separate issue 🙂

crates/ide-db/src/assists.rs
joshka: The trigger verb used here seems like it stemmed from the name of the editor action "editor.action.triggerParameterHints". This runs "editor.action.rename", so I'm leaning towards leaving it as rename (but happy to change it if needed)

crates/ide-assists/src/handlers/extract_variable.rs
joshka: Done

crates/rust-analyzer/src/lsp/to_proto.rs
joshka: done

editors/code/src/client.ts
joshka: done

joshka:
> Ideally all of the extract & generate assists should still be using the `add_placeholder_snippet_group` for better out-of-the box support for other editor extensions, but that's an entirely separate issue 🙂

I experimented with that first and it seemed like it was a less than ideal experience. Agree that this is a separate issue.

DropDemBits (approved):

Veykril:
Thanks!
@bors r+

nemethf:
@joshka, @Veykril, shouldn't RA check whether the LSP client announced its support for the command "rust-analyzer.rename" before sending it in a codeAction/resolve reply?

Also, I'm not sure asking the client to do a rename after the the extract is the best approach since RA can send "disjoint" snippets as well.  I have not tested it, but instead of sending this:
````json
{
  "jsonrpc": "2.0", "id": 7,
  "result": {
    "title": "Extract into variable",
    "kind": "refactor.extract",
    "command": { "title": "rename",  "command": "rust-analyzer.rename" },
    "edit": {
      "documentChanges": [
        {
          "textDocument": { "uri": "file:///tmp/rust-rename-17587/src/main.rs", "version": 0 },
          "edits": [
            {
              "range": {
                "start": { "line": 1, "character": 8 },
                "end": { "line": 1, "character": 9  }
              },
              "newText": "$0var_name",
              "insertTextFormat": 2
            },
            {
              "range": {
                "start": { "line": 1, "character": 18 },
                "end": {  "line": 1,  "character": 18 }
              },
              "newText": "\n    let a = var_name;"
            }
          ]
        }
      ]
    }
  }
}
````
RA could send something like this:
````json
{
  "jsonrpc": "2.0", "id": 7,
  "result": {
    "title": "Extract into variable",
    "kind": "refactor.extract",
    "command": { "title": "rename",  "command": "rust-analyzer.rename" },
    "edit": {
      "documentChanges": [
        {
          "textDocument": { "uri": "file:///tmp/rust-rename-17587/src/main.rs", "version": 0 },
          "edits": [
            {
              "range": {
                "start": { "line": 1, "character": 8 },
                "end": { "line": 1, "character": 9  }
              },
              "newText": "$0${1:var_name}",
              "insertTextFormat": 2
            },
            {
              "range": {
                "start": { "line": 1, "character": 18 },
                "end": {  "line": 1,  "character": 18 }
              },
              "newText": "\n    let a = $1;"
              "insertTextFormat": 2
            }
          ]
        }
      ]
    }
  }
}
````
I.e., the "var_name" is a just default value for the mirrored tab-stop field.  The commit log of #15876 directly says: 
> Explicitly allow a single TextDocumentEdit to have multiple SnippetTextEdits. This allows things like renaming extracted variables and functions without having to go through a separate rename step.

**Comment by DropDemBits** [comment]
@nemethf

> shouldn't RA check whether the LSP client announced its support for the command "rust-analyzer.rename" before sending it in a codeAction/resolve reply?

Hmm, it should be respecting the [client commands](https://github.com/rust-lang/rust-analyzer/blob/db43a5a6e97ceb90920169afef4aaa9aa06e82ac/docs/dev/lsp-extensions.md#client-commands) list that the client declares and only emitting it if `editor.action.rename` is specified there.

> Also, I'm not sure asking the client to do a rename after the the extract is the best approach since RA can send "disjoint" snippets as well.

This would also be better for editors like Helix, where I forsee code action snippet support being available before a plugin system is ready.

It is possible to do it today like is done for `generate_trait_from_impl`, but for `extract_variable` there's one case where it's annoying to grab the nodes to mark as a placeholder group. Once we have something like #9649 it'll be nicer to find these nodes again, so hopefully I can scrounge up enough free time and look into that again.

**Comment by nemethf** [comment]
> > shouldn't RA check whether the LSP client announced its support for the command "rust-analyzer.rename" before sending it in a codeAction/resolve reply?
> 
> Hmm, it should be respecting the [client commands](https://github.com/rust-lang/rust-analyzer/blob/db43a5a6e97ceb90920169afef4aaa9aa06e82ac/docs/dev/lsp-extensions.md#client-commands) list that the client declares and only emitting it if `editor.action.rename` is specified there.  

That is not what happens in my case.   The client only states that it supports the most basic rename action in its RenameClientCapabilities and that's all.  And yet the server sends this in its textDocument/rename reply: `"command": { "title": "rename",  "command": "rust-analyzer.rename" },`

````
$ rust-analyzer --version
rust-analyzer 0.4.2041-standalone
````

<details>
<summary>This is the initialize request my LSP client sends:</summary>
<pre>
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "initialize",
  "params": {
    "processId": 1602814,
    "clientInfo": { "name": "Eglot",  "version": "1.17" },
    "rootPath": "/tmp/rust-rename-17587/",
    "rootUri": "file:///tmp/rust-rename-17587",
    "initializationOptions": {},
    "capabilities": {
      "workspace": {
        "applyEdit": true,
        "executeCommand": { "dynamicRegistration": false },
        "workspaceEdit": { "documentChanges": true },
        "didChangeWatchedFiles": { "dynamicRegistration": true },
        "symbol": { "dynamicRegistration": false },
        "configuration": true,
        "workspaceFolders": true
      },
      "textDocument": {
        "synchronization": {
          "dynamicRegistration": false,
          "willSave": true,
          "willSaveWaitUntil": true,
          "didSave": true
        },
        "completion": {
          "dynamicRegistration": false,
          "completionItem": {
            "snippetSupport": true,
            "deprecatedSupport": true,
            "resolveSupport": {
              "properties": [ "documentation", "details", "additionalTextEdits" ]
            },
            "tagSupport": { "valueSet": [ 1 ] }
          },
          "contextSupport": true
        },
        "hover": {
          "dynamicRegistration": false,
          "contentFormat": [ "plaintext" ]
        },
        "signatureHelp": {
          "dynamicRegistration": false,
          "signatureInformation": {
            "parameterInformation": { "labelOffsetSupport": true },
            "documentationFormat": [ "plaintext" ],
            "activeParameterSupport": true
          }
        },
        "references": { "dynamicRegistration": false },
        "definition": {
          "dynamicRegistration": false,
          "linkSupport": true
        },
        "declaration": {
          "dynamicRegistration": false,
          "linkSupport": true
        },
        "implementation": {
          "dynamicRegistration": false,
          "linkSupport": true
        },
        "typeDefinition": {
          "dynamicRegistration": false,
          "linkSupport": true
        },
        "documentSymbol": {
          "dynamicRegistration": false,
          "hierarchicalDocumentSymbolSupport": true,
          "symbolKind": {
            "valueSet": [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26 ]
          }
        },
        "documentHighlight": { "dynamicRegistration": false },
        "codeAction": {
          "dynamicRegistration": false,
          "resolveSupport": {
            "properties": [ "edit", "command" ]
          },
          "dataSupport": true,
          "codeActionLiteralSupport": {
            "codeActionKind": {
              "valueSet": [
                "quickfix",
                "refactor",
                "refactor.extract",
                "refactor.inline",
                "refactor.rewrite",
                "source",
                "source.organizeImports"
              ]
            }
          },
          "isPreferredSupport": true
        },
        "formatting": { "dynamicRegistration": false },
        "rangeFormatting": { "dynamicRegistration": false },
        "rename": { "dynamicRegistration": false },
        "inlayHint": { "dynamicRegistration": false },
        "publishDiagnostics": { 
          "relatedInformation": false,
          "codeDescriptionSupport": false,
          "tagSupport": {
            "valueSet": [ 1, 2 ]
          }
        }
      },
      "window": {
        "showDocument": { "support": true },
        "workDoneProgress": true
      },
      "general": {
        "positionEncodings": [ "utf-32", "utf-8", "utf-16" ]
      },
      "experimental": {
        "snippetTextEdit": true,
        "serverStatusNotification": true,
        "colorDiagnosticOutput": true,
        "openServerLogs": true,
        "testExplorer": true,
        "localDocs": true
      },
      "offsetEncoding": [ "utf-32", "utf-16" ]
    },
    "workspaceFolders": [
      {
        "uri": "file:///tmp/rust-rename-17587",
        "name": "/tmp/rust-rename-17587/"
      }
    ]
  }
}
</pre>
</details>

**Comment by Veykril** [comment]
There is a weird `lens_forceCustomCommands` config that is incorrectly set to to `true` by default, when that is turned on we ignored the capability 🙃 That would explain this

**Comment by joshka** [comment]
> Also, I'm not sure asking the client to do a rename after the the extract is the best approach since RA can send "disjoint" snippets as well. I have not tested it, but instead of sending this:

The rename UX flow seems (subjectively) nicer than the disjoint snippets flow to me as there's no good intuitive way to finish the disjoint edit (neither Tab or Enter work, only Escape), also there's some editing weirdness:

Changing extract_variable.rs:131 (and removing the rename call):
```rust
                    if let Some(cap) = ctx.config.snippet_cap {
                        if let Some(ast::Pat::IdentPat(ident_pat)) = let_stmt.pat() {
                            if let Some(name) = ident_pat.name() {
                                edit.add_placeholder_snippet_group(
                                    cap,
                                    vec![name.syntax().clone(), name_expr.syntax().clone()],
                                );
                            }
                        }
                    }
```

https://github.com/user-attachments/assets/1dd1c96d-bf4d-4e92-9c25-0b676f48777f

Regarding the client commands, the server side execution of the rename command is a noop if the editor.action.rename command is not enabled (as commands.rename == false in that case). I don't think the assists have access to the any client command config, so there's no way to filter this out.

https://github.com/rust-lang/rust-analyzer/blob/b333f85a9dc06a45d9eb126c6371414c49b46784/crates/rust-analyzer/src/lsp/to_proto.rs#L1339-L1346

> That is not what happens in my case. The client only states that it supports the most basic rename action in its RenameClientCapabilities and that's all. And yet the server sends this in its textDocument/rename reply: "command": { "title": "rename",  "command": "rust-analyzer.rename" },

Is there any impact of this other than an unnecessary round trip? I'm trying to understand what the error / problem is here, but I'm not seeing it.

Veykril:
> Is there any impact of this other than an unnecessary round trip? I'm trying to understand what the error / problem is here, but I'm not seeing it.

That depends on the client, we are violating the LSP (or in this case the lsp extension I suppose), we shouldn't be doing that.

The problem is this config being enabled by default for some reason https://github.com/veykril/rust-analyzer/blob/09ef75c717e9ffb4cbc71ce1cfec7694244742c2/crates/rust-analyzer/src/config.rs#L564 which then force enables all of the commands here https://github.com/veykril/rust-analyzer/blob/09ef75c717e9ffb4cbc71ce1cfec7694244742c2/crates/rust-analyzer/src/config.rs#L1896-L1911
We should just get rid of `lens_forceCustomCommands`, I see no value in it for debugging.

joshka:
Ah got it

nemethf:
There is one additional thing that I don't understand.  The client needs to declare support for `editor.action.rename`, but the server sends `rust-analyzer.rename` instead of `editor.action.rename`.  Is this intentional?

Thanks.

Veykril:
Same goes for the `"editor.action.triggerParameterHints",` capability. Thanks for calling that out I think we should move those over to `rust-analyzer.action.*` or similar. The names are based on the equivalent vscode commands but we should clearly not use the exact same names for those.

PR #17588, opened 2024-07-13
feat: Add incorrect case diagnostics for enum variant fields and all variables/params

CamWass:
Updates the incorrect case diagnostic to check:

1. Fields of enum variants. Example:
```rust
enum Foo {
    Variant { nonSnake: u8 }
}
```
2. All variable bindings, instead of just let bindings and certain match arm patters. Examples:
```rust
match 1 { nonSnake => () }
match 1 { nonSnake @ 1 => () }
match 1 { nonSnake1 @ nonSnake2 => () } // slightly cursed, but these both introduce new
                                        // bindings that are bound to the same value.

const ONE: i32 = 1;
match 1 { nonSnake @ ONE } //  ONE is ignored since it is not a binding

match Some(1) { Some(nonSnake) => () }

struct Foo { field: u8 }
match (Foo { field: 1 } ) {
    Foo { field: nonSnake } => ();
}

struct Foo { nonSnake: u8 } // diagnostic here, at definition
match (Foo { nonSnake: 1 } ) { // no diagnostic here...
    Foo { nonSnake } => ();    // ...or here, since these are not where the name is introduced
}

for nonSnake in [] {}

struct Foo(u8);
for Foo(nonSnake) in [] {}
```
3. All parameter bindings, instead of just top-level binding identifiers. Examples:
```rust
fn func(nonSnake: u8) {} // worked before

struct Foo { field: u8 }
fn func(Foo { field: nonSnake }: Foo) {} // now get diagnostic for nonSnake
```

This is accomplished by changing the way binding identifier patterns are filtered:
- Previously, all binding idents were skipped, except a few classes of "good" binding locations that were checked.
- Now, all binding idents are checked, except field shorthands which are skipped.

Moving from a whitelist to a blacklist potentially makes the analysis more brittle:
If new pattern types are added in the future where ident pats don't introduce new names, then they may incorrectly create diagnostics.

But the benefit of the blacklist approach is simplicity: I think a whitelist approach would need to recursively visit patterns to collect renaming candidates?

### DIFF
--- a/crates/hir-ty/src/diagnostics/decl_check.rs
+++ b/crates/hir-ty/src/diagnostics/decl_check.rs
@@ -17,8 +17,8 @@ use std::fmt;
 
 use hir_def::{
     data::adt::VariantData, db::DefDatabase, hir::Pat, src::HasSource, AdtId, AttrDefId, ConstId,
-    EnumId, FunctionId, ItemContainerId, Lookup, ModuleDefId, ModuleId, StaticId, StructId,
-    TraitId, TypeAliasId,
+    EnumId, EnumVariantId, FunctionId, ItemContainerId, Lookup, ModuleDefId, ModuleId, StaticId,
+    StructId, TraitId, TypeAliasId,
 };
 use hir_expand::{
     name::{AsName, Name},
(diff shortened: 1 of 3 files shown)


crates/hir-ty/src/diagnostics/decl_check.rs
```diff
@@ -551,6 +555,75 @@ impl<'a> DeclValidator<'a> {
         }
     }
 
+    /// Check incorrect names for fields of enum variant.
+    fn validate_enum_variant_fields(&mut self, variant_id: EnumVariantId) {
```
CamWass, line 559: This is copied from `validate_struct_fields`, and similar looking code now exists in a few places in this file. As such, I think it would be useful to clean this up in a follow-up PR, extracting the shared logic.

crates/ide-diagnostics/src/handlers/incorrect_case.rs
```diff
@@ -344,24 +345,6 @@ fn main() {
         );
     }
 
-    #[test]
-    fn non_let_bind() {
```
Veykril, line 348: shouldnt this test stay?

CamWass: I ended up replacing it with [this test](https://github.com/rust-lang/rust-analyzer/pull/17588/files/af30111b1c8af7abdaaa52120e24e088f15ff25b#diff-4fb2429f42867a6c75f039b611c8086c035d6daf099dbdaffb5e6fd9a78927c5R817) inside of `test_match` below:
```rust
enum Foo { V1, V2 }
use Foo::V1;
fn main() {
    match V1 {
        _Bad1 @ V1 => {},
     // ^^^^^ 💡 warn: Variable `_Bad1` should have snake_case name, e.g. `_bad1`
        Foo::V2 => {}
    }
}
```
I did this to group the `match` tests, and because the old test has an issue: it doesn't `use Option::None`, so the `None` in `SOME_VAR @ None` actually referes to a new variable binding, rather than an enum variant.
This is the same issue I fixed in `missing_match_arms.rs` [here](https://github.com/rust-lang/rust-analyzer/pull/17588/files/af30111b1c8af7abdaaa52120e24e088f15ff25b#diff-332f11ebf6f6a14d806c673271b14832507641612034ccbec801318bafff5005R611)

Veykril: Gotcha

Veykril:
Thanks!
@bors r+

PR #17595, opened 2024-07-15
Implement lifetime inferring

dfireBird:


Veykril (approved):

flodiebold:
@bors r+

thanks!

Veykril:
This regressed unknown types from 41 to 62 on self somehow 🤔

dfireBird:
There shouldn't be anything creating new `TyKind::Error`. I'll check on that. Meanwhile, if you want to revert that commit we can.

Veykril:
It's fine, it only regressed on self so its likely not too bad. I imagine unification fails due to some new lifetimes now within chalk

PR #17605, opened 2024-07-16
Set `RUSTC_TOOLCHAIN` for runnables

Veykril:
With this the client doesn't necessarily need to guess the sysroot anymore

### DIFF
--- a/editors/code/src/debug.ts
+++ b/editors/code/src/debug.ts
@@ -3,10 +3,10 @@ import * as vscode from "vscode";
 import * as path from "path";
 import type * as ra from "./lsp_ext";
 
-import { Cargo, getRustcId, getSysroot } from "./toolchain";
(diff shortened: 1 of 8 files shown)


editors/code/src/debug.ts
```diff
@@ -142,18 +142,29 @@ async function getDebugConfiguration(
...
 
     const provider = unwrapUndefinable(knownEngines[debugEngine.id]);
-    const debugConfig = provider(runnable, runnableArgs, simplifyPath(executable), env);
+    const debugConfig = provider(
+        runnable,
+        runnableArgs,
+        simplifyPath(executable),
+        env,
+        sourceFileMap,
+    );
```
Veykril, line 167: We were no longer passing the source map here which I think mightve caused https://github.com/rust-lang/rust-analyzer/issues/17568

PR #17610, opened 2024-07-16
Add method `Impl::all_in_module(…)` for allowing more localized querying

regexident:
This PR is motivated by an [outside use](https://github.com/regexident/cargo-modules) of the `ra_ap_hir` crate that would benefit from being able to more efficiently query for all impls of a given module (instead of having to query for its parent crate's impls and then filtering on `impl_hir.module(db) == module_hir`).

I have the suspicion that the code as is won't quite work for file-level modules, since those don't have a block, afaict, but with all the crate-rename and version shenanigans around the `ra_ap_` release process I haven't yet been able to figure out how to patch a `ra_ap_` dependency from a third-party crate (let alone how to test this from inside of `rust-analyzer`), so haven't been able to actually run and test this. 😔

Any hints on how to make this (i.e. the code itself, as well as the testing) work are more than welcome!

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -3724,6 +3724,10 @@ impl Impl {
         inherent.all_impls().chain(trait_.all_impls()).map(Self::from).collect()
     }
 
+    pub fn all_in_module(db: &dyn HirDatabase, module: Module) -> Vec<Impl> {
+        module.id.def_map(db.upcast())[module.id.local_id].scope.impls().map(Into::into).collect()
+    }
+
     pub fn all_for_type(db: &dyn HirDatabase, Type { ty, env }: Type) -> Vec<Impl> {
         let def_crates = match method_resolution::def_crates(db, &ty, env.krate) {
             Some(def_crates) => def_crates,


Veykril:
block refers to block modules here, not `mod foo {}` (called `inline module`s versus `mod foo;` being `outline module`s). Block modules are modules nested within a block expression like `fn foo { mod bar {} }`. 

The simplest way to fetch all impls within a module is probably to go through its item scope, there already is an `impls` field on it. (see `Module::scope` for how to access the item scope of a module)

regexident:
@Veykril so something like this then? I tried figuring out what `ScopeDef::ImplSelfType` does/is exactly, but couldn't find any inline docs/comments on it.

```rust
pub fn all_in_module(db: &dyn HirDatabase, module: Module) -> Vec<Impl> {
    module
        .scope(db, None)
        .into_iter()
        .filter_map(|(_name, scope)| match scope {
            ScopeDef::ImplSelfType(impl_hir) => Some(impl_hir),
            _ => None,
        })
        .collect()
}
```

---

@Veykril Do you know a way to cargo patch ` ra_ap_hir` dependency based on a local `rust-analyzer` checkout? I've tried a bunch of things a while ago, but couldn't get anything to work, which makes testing PRs like this one [somewhat hard](https://github.com/rust-lang/rust-analyzer/pull/17312#issuecomment-2142857345). (Or even just a suggestion on how to at least be able to write throw-away tests for the HIR API?) I feel bad opening PRs that basically rely almost exclusively on type-checking and code review from you folks (which I'm grateful for). 😕

Veykril:
```rs
pub fn all_in_module(db: &dyn HirDatabase, module: Module) -> Vec<Impl> {
        self.id.def_map(db.upcast())[self.id.local_id]
            .scope
            .impls()
            .map(Into::into)
            .collect()
}
```
is what I was thinking of.

Regarding patching, I guess plain cargo patch section fails because of the lack of a filled out version field?

regexident:
> […] is what I was thinking of.

Ah, now I see. Fixed and pushed!

> Regarding patching, I guess plain cargo patch section fails because of the lack of a filled out version field?

It's either that or the rename (`ra_ap_hir` internally is called `hir`), which apparently [isn't supported](https://github.com/rust-lang/cargo/issues/11392) by cargo's patch. Either way the patch doesn't get picked up. The warning isn't super clear about the why:

```toml
// Cargo.toml

[patch.crates-io]
ra_ap_hir = { version = "0.0.0", package = "hir", path = "../rust-analyzer/crates/hir" }
```

```
$ cargo check

warning: Patch `hir v0.0.0 (<SNIP>/rust-analyzer/crates/hir)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.41s
```

Veykril (approved):

PR #17620, opened 2024-07-17
Edition aware parser

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/16324 by allowing us to properly thread through the edition to the parser


Veykril:
There is a few things still missing, like handling `dyn` as a contextual keyword in 2015 edition and probably some other smaller things, but the important part here is being able to parse `gen` without everything suddenly breaking :)

Veykril:
Well that didn't quite work ... 
image

PR #17626, opened 2024-07-18
internal: Escape `*` in Useful Setup Tips

YDX-2147483647:
Useful Setup Tips was added in #17576 from https://github.com/rust-lang/rust-analyzer/issues/17569#issuecomment-2217407336. And the current version looks confusing in VS Code:

Not sure if this is the right way to escape `*` in json in md in json… Please double check!


lnicola:
See https://github.com/rust-lang/rust-analyzer/issues/17621. That's not exactly Markdown, it would display correctly if it was.

YDX-2147483647:
Sorry I didn't notice that…

PR #17639, opened 2024-07-19
Some more small salsa memory improvements

Veykril:
This does limit our lru limits to 2^16 but if you want to set them higher than that you might as well not set them at all. Also makes `LRU` opt-in per query now, allowing us to drop all the unnecessary LRU stuff for most queries


lnicola:
This was a ~23 MB win on self.

PR #17641, opened 2024-07-19
Avoid ref when using format! in compiler

nyurik:
Clean up a few minor refs in `format!` macro, as it has a performance cost. Apparently the compiler is unable to inline `format!("{}", &variable)`, and does a run-time double-reference instead (format macro already does one level referencing). Inlining format args prevents accidental `&` misuse.

See https://github.com/rust-lang/rust-clippy/issues/10851

### DIFF
--- a/crates/ide/src/rename.rs
+++ b/crates/ide/src/rename.rs
@@ -516,7 +516,7 @@ mod tests {
         match result {
             Ok(RangeInfo { range, info: () }) => {
                 let source = analysis.file_text(position.file_id).unwrap();
-                expect.assert_eq(&format!("{range:?}: {}", &source[range]))
+                expect.assert_eq(&format!("{range:?}: {}", source[range]))
             }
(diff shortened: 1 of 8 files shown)


crates/ide/src/rename.rs
```diff
@@ -516,7 +516,7 @@ mod tests {
         match result {
             Ok(RangeInfo { range, info: () }) => {
                 let source = analysis.file_text(position.file_id).unwrap();
-                expect.assert_eq(&format!("{range:?}: {}", &source[range]))
+                expect.assert_eq(&format!("{range:?}: {}", source[range]))
```
Veykril, line 519:
```suggestion
                expect.assert_eq(&format!("{range:?}: {}", &source[range]))
```
Doesn't seem to work for unsized types

Veykril:
Thanks!
@bors r+

PR #17642, opened 2024-07-20
minor: Fix some typos

lnicola:


crates/rust-analyzer/src/handlers/notification.rs
```diff
@@ -292,7 +292,6 @@ fn run_flycheck(state: &mut GlobalState, vfs_path: VfsPath) -> bool {
                 .into_iter()
                 .flat_map(|id| world.analysis.transitive_rev_deps(id))
                 .flatten()
-                .sorted()
```
lnicola, line 295: We just trigger flycheck here, I don't think the order matters.

PR #17647, opened 2024-07-20
Rename rust-analyzer commands

joshka:
The commands `editor.action.triggerParameterHints` and
`editor.action.rename` are now renamed to
`rust-analyzer.triggerParameterHints` and `rust-analyzer.rename`

This change helps make it clear that these commands are specific to
rust-analyzer and not part of the default set of commands provided by
VSCode.

Fixes: https://github.com/rust-lang/rust-analyzer/issues/17644

Note: This seems like it will be a breaking change for any RA client that previously reacted to `editor.action.triggerParameterHints` - naive search: https://github.com/search?q=editor.action.triggerParameterHints+AND+%28NOT+is%3Afork%29+rust-analyzer&type=code


joshka:
ping @Veykril (due to existing context)

Veykril:
This is a breaking change, but that can't be helped imo, thanks!
@bors r+

PR #17650, opened 2024-07-20
Fix path resolution for child mods of those expanded by `include!`

### LINKED ISSUE #17645: Modules inserted via `include` can't resolve later files
**rust-analyzer version**: 0.4.2042-standalone (b333f85a9 2024-07-19)

**rustc version**: rustc 1.78.0 (9b00956e5 2024-04-29)

A test case to reproduce based on `nested_include`
```rust
#[test]
fn include_with_mod_path() {
    check(
        r#"
//- minicore: include
//- /lib.rs
include!("out_dir/includes.rs");

//- /out_dir/includes.rs
#[path = "."]
pub mod company_name {
    pub mod network {
        pub mod v1;
    }
}
//- /out_dir/network/v1.rs
pub struct IpAddress {
    pub ip_type: &'static str,
}
/// Nested message and enum types in `IpAddress`.
pub mod ip_address {
    pub enum IpType {
        IpV4(u32),
    }
}

"#,
        expect![[r#"
            crate
            company_name: t

            crate::company_name
            network: t

            crate::company_name::network
            v1: t

            crate::company_name::network::v1
            IpAddress: t
            ip_address: t

            crate::company_name::network::v1::ip_address
            IpType: t
        "#]],
    );
}
```
The types in the `v1` module don't resolve, failing the test. Rust has no problem handling modules with `#[path = "."]`. RA also can handle modules with `#[path = "."]` if they aren't `include!`d, so this seems to be a problem exclusive to `include!`.

ObsidianMinor:
Child modules wouldn't use the correct candidate paths due to a branch that doesn't seem to be doing what it's intended to do. Removing the branch fixes the problem and all existing test cases pass.

Having no knowledge of how any of this works, I believe this fixes #17645. Using another test that writes the included mod directly into `lib.rs` instead, I found the difference can be traced to the candidate files we use to look up mods. A separate branch for if the file comes from an `include!` macro doesn't take into account the original mod we're contained within:

```rust
None if file_id.macro_file().map_or(false, |it| it.is_include_macro(db.upcast())) => {
    candidate_files.push(format!("{}.rs", name.display(db.upcast())));
    candidate_files.push(format!("{}/mod.rs", name.display(db.upcast())));
}
```

I'm not sure why this branch exists. Tracing the branch back takes us to 3bb9efb but it doesn't say *why* the branch was added. The test case that was added in this commit passes with the branch removed, so I think it's just superfluous at this point.

### DIFF
--- a/crates/hir-def/src/nameres/mod_resolution.rs
+++ b/crates/hir-def/src/nameres/mod_resolution.rs
@@ -1,7 +1,7 @@
 //! This module resolves `mod foo;` declaration to file.
 use arrayvec::ArrayVec;
 use base_db::AnchoredPath;
-use hir_expand::{name::Name, HirFileIdExt, MacroFileIdExt};
+use hir_expand::{name::Name, HirFileIdExt};
 use limit::Limit;
 use span::EditionedFileId;
 use syntax::ToSmolStr as _;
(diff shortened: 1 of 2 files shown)


Veykril:
That branch mightve gotten superceded by some recent changes to the include logic

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17660, opened 2024-07-22
Fix more path resolution for included submodules

ObsidianMinor:
Now with more comprehensive testing! This adds tests for includes within modules. Previous testing was not comprehensive enough since submodules that use `include!` didn't actually work either! The `ModDir` used for resolving mods relative to included files has to be `ModDir::root()`. The original test just so happened to put the submodules in the root which made this work, but if you put the `include!` inside a `mod` block it didn't work.

With this change, when collecting a macro expansion, if the macro call is an `include!`, we use the `ModDir::root()` instead of the current module we're in.


ObsidianMinor:
I probably should've checked this in the previous PR, but this does fix the issues I had before.

From the given `include!`
image

Before (no inline hints or code completion)
image

After (with type hints and code completion)
image

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17661, opened 2024-07-22
minor: Read workspace root ratomls on startup

alibektas:


crates/load-cargo/src/lib.rs
```diff
@@ -238,15 +240,32 @@ impl ProjectFolders {
...
+            }
 
-            let entry = vfs::loader::Entry::Files(vec![manifest.to_owned()]);
+            // In case of detached files we do **not** look for a rust-analyzer.toml.
+            if !matches!(ws.kind, ProjectWorkspaceKind::DetachedFile { .. }) {
+                let ws_root = ws.workspace_root();
+                file_set_roots.push(VfsPath::from(ws_root.to_owned()));
+                entries.push(ws_root.to_owned());
+                register = true;
+            }
```
Veykril, line 260: We should specifically push the `rust-analyzer.toml` file path here, not the workspace root itself. We only want to watch that one file there

crates/load-cargo/src/lib.rs
```diff
@@ -238,15 +240,32 @@ impl ProjectFolders {
...
 
-            res.watch.push(res.load.len());
-            res.load.push(entry);
-            local_filesets.push(fsc.len() as u64);
-            fsc.add_file_set(file_set_roots)
+            if register {
```
Veykril, line 262: `register` can be replaced by checking whether `entries` and/or `file_set_roots` is empty or not

crates/load-cargo/src/lib.rs
alibektas: sorry ofc I meant to write that. :sad

Veykril (approved):

cormacrelf:
Given https://github.com/rust-lang/rust-analyzer/pull/17246, it's still not being read early enough -- there is a pretty big chicken/egg problem with configs like this:

```toml
[workspace.discoverConfig]
...
```

We don't read rust-analyzer.toml until workspaces have already been discovered (to find the root), and we can't discover a workspace because haven't read it. I think we need to search upwards from the current working directory to find a rust-analyzer.toml to use on startup.

For rust-analyzer.toml users, these configs will never be populated in src/bin/main.rs:

https://github.com/rust-lang/rust-analyzer/blob/f46af90ba83f38cb7fa55ba9e5620b2ceed58f2b/crates/rust-analyzer/src/bin/main.rs#L268-L273

Veykril:
Well, we should actually not allow `workspace.discoverConfig` in workspace associated `rust-analyzer.toml` files given the chicken and egg problem. Having it in there is non-sensical.

We are in general planning a slight restructuring of config grouping as the current structure has some problems.

cormacrelf:
Configuring a repo to discover non-cargo workspaces for anyone who opens it is one of the main difficulties I was hoping rust- analyzer.toml would solve, so I really hope we can break the chicken/egg somehow.

Veykril:
We definitely want to have some solution there, but the current implementation isn't sufficient for that (its really just an MVP atm).

alibektas:
That's a valid point to which I have no answer atm. But I am making a separate issue from this and I will find a way to address this

PR #17667, opened 2024-07-22
Use rustup rust-analyzer component when there is a toolchain file override for the opened workspace

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/17663

### DIFF
--- a/editors/code/src/bootstrap.ts
+++ b/editors/code/src/bootstrap.ts
@@ -42,6 +42,7 @@ async function getServer(
         enableProposedApi: boolean | undefined;
     } = context.extension.packageJSON;
 
+    // check if the server path is configured explicitly
     const explicitPath = process.env["__RA_LSP_SERVER_DEBUG"] ?? config.serverPath;
     if (explicitPath) {
         if (explicitPath.startsWith("~/")) {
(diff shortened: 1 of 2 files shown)


editors/code/src/bootstrap.ts
```diff
@@ -51,12 +52,29 @@ async function getServer(
     }
     if (packageJson.releaseTag === null) return "rust-analyzer";
 
+    if (vscode.workspace.workspaceFolders?.length === 1) {
+        // otherwise check if there is a toolchain override for the current vscode workspace
+        // and if the toolchain of this override has a rust-analyzer component
+        // if so, use the rust-analyzer component
+        const toolchainTomlExists = await fileExists(
+            vscode.Uri.joinPath(vscode.workspace.workspaceFolders[0]!.uri, "rust-toolchain.toml"),
```
lnicola, line 60: Nb. this doesn't handle `rust-toolchain`, but I think it's fine.

davidbarsky:
Wait, why does this one prefer a rustup-provided rust-analyzer over a bundled rust-analyzer? I'd expect that the bundled one would take priority.

Veykril:
See the discussion in https://github.com/rust-lang/rust-analyzer/issues/17663, if the user explicitkly installed the rust-analyzer component in a toolchain override it is a big indicator they meant to use that instead. And a toolchain r-a is usually compatible with the toolchain where as the bundled one only supports the last couple stable releases

davidbarsky:
Ah, I assumed that the presence of any `rust-toolchain.toml` file would pull the rust-analyzer component from there. My bad!

PR #17670, opened 2024-07-22
LRU `body_with_source_map` query

Veykril:
This query is being invalidated all the time anyways (we have an extra query on top of it for the body incrementality that is not source dependent), so there is little reason to keep these around all the time when only some IDE features are interested in them.


lnicola:
This is a 61 MB improvement on `self`.

PR #17671, opened 2024-07-22
internal: Shrink size of `Binding`

Veykril:
This should save a bit of memory now that we LRU the source map


lnicola:
This looks like an extra 3 MB reduction on `self`.

PR #17674, opened 2024-07-22
internal: add more output to the status command

davidbarsky:
Bit of a lazy change, but this is was pretty handy. I think I should clean up the configuration into something a bit more legible (maybe serialize as JSON?), but I think this is a good enough starting point that we might as well start asking people for it in issue reports.


lnicola:
Maybe not in this PR, but I think it would be useful to also report:

 - the (anonymized?) sysroot
 - the toolchain version
 - the last error encountered when running build scripts, if any
 - [your own ideas here]

Unfortunately, this won't really help with non-VS Code users, of course.

flodiebold:
> Unfortunately, this won't really help with non-VS Code users, of course.

emacs-lsp has a `lsp-rust-analyzer-status` command. I don't know about other editors, of course.

PR #17675, opened 2024-07-22
Remove lens.forceCustomCommands config

PaulDotSH:
Closes https://github.com/rust-lang/rust-analyzer/issues/17643
A very simple PR that removes the lens.forceCustomCommands config feature without side effects.

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -655,9 +655,6 @@ config_data! {
         lens_debug_enable: bool            = true,
         /// Whether to show CodeLens in Rust files.
         lens_enable: bool           = true,
-        /// Internal config: use custom client-side commands even when the
-        /// client doesn't set the corresponding capability.
-        lens_forceCustomCommands: bool = true,
         /// Whether to show `Implementations` lens. Only applies when
         /// `#rust-analyzer.lens.enable#` is set.
         lens_implementations_enable: bool  = true,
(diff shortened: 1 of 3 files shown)


crates/rust-analyzer/src/config.rs
```diff
@@ -2032,7 +2029,7 @@ impl Config {
 
     pub fn client_commands(&self) -> ClientCommandsConfig {
         let commands = self.commands();
-        let force = commands.is_none() && *self.lens_forceCustomCommands();
+        let force = commands.is_none();
```
nemethf, line 2032: Bit isn't this worse than the current status quo?  Currently,  the whole thing can be disabled with lens_forceCustomCommands=false.  With this PR, if the client does not support anything, the server will assume it supports everything.  Right?

PaulDotSH: Sorry, I was thinking that this was the intended behavior, then the intended fix would be to remove "force" all together, right?

lnicola (approved):

lnicola:
Thanks!

@bors r+

crates/rust-analyzer/src/config.rs
lnicola: Yeah, I _think_ we should get rid of it. The Code extension does set it properly in https://github.com/rust-lang/rust-analyzer/blob/4afe0d5393cdedde58881295752fe68acb3148ae/editors/code/src/client.ts#L344.

Veykril: We should remove the `force` behavior fully

lnicola, line 2032: It's unrelated to this PR and not a big deal, but I think we call `client_commands` four times per hover action, and it does a bunch of allocations, so we might want to cache it.

PaulDotSH: I forgot to reply, I pushed a new commit that should have the correct code now

Issue #17598, opened 2024-07-15
Support for `+ use<..>` `precise_capturing` syntax


traviscross:
We're planning to stabilize `#![feature(precise_capturing)]` ahead of Rust 2024 (in all editions).  This feature allows one to write:

```rust
fn captures<'a: 'a, 'b: 'b, T>() -> impl Sized + use<'b, T> {}
//                                  ~~~~~~~~~~~~~~~~~~~~~~~
//                          This opaque type does not capture `'a`.

fn outlives<'o, T: 'o>(_: T) {}

fn caller<'o, 'a, 'b: 'o, T: 'o>() {
    //        ~~
    //        ^ Note that we don't need `'a: 'o`.
    outlives::<'o>(captures::<'a, 'b, T>());
}
```

[Playground link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&code=%23%21%5Bfeature%28precise_capturing%29%5D%0A%23%21%5Ballow%28unused%29%5D%0A%0Afn+captures%3C%27a%3A+%27a%2C+%27b%3A+%27b%2C+T%3E%28%29+-%3E+impl+Sized+%2B+use%3C%27b%2C+T%3E+%7B%7D%0A%2F%2F++++++++++++++++++++++++++++++++++~~~~~~~~~~~~~~~~~~~~~~~%0A%2F%2F++++++++++++++++++++++++++This+opaque+type+does+not+capture+%60%27a%60.%0A%0Afn+outlives%3C%27o%2C+T%3A+%27o%3E%28_%3A+T%29+%7B%7D%0A%0Afn+caller%3C%27o%2C+%27a%2C+%27b%3A+%27o%2C+T%3A+%27o%3E%28%29+%7B%0A++++%2F%2F++++++++~~%0A++++%2F%2F++++++++%5E+Note+that+we+don%27t+need+%60%27a%3A+%27o%60.%0A++++outlives%3A%3A%3C%27o%3E%28captures%3A%3A%3C%27a%2C+%27b%2C+T%3E%28%29%29%3B%0A%7D%0A%0Afn+main%28%29+%7B%7D%0A)

[...]

PR #17676, opened 2024-07-22
feat: add preliminary support for `+ use<..>` `precise_capturing` syntax

winstxnhdw:
## Summary

This PR adds basic support for the following syntax.

```rs
fn captures<'a: 'a, 'b: 'b, T>() -> impl Sized + use<'b, T> {}
//                                  ~~~~~~~~~~~~~~~~~~~~~~~
//                          This opaque type does not capture `'a`.

fn outlives<'o, T: 'o>(_: T) {}

fn caller<'o, 'a, 'b: 'o, T: 'o>() {
    //        ~~
    //        ^ Note that we don't need `'a: 'o`.
    outlives::<'o>(captures::<'a, 'b, T>());
}
```

Related to #17598

### DIFF
--- a/crates/parser/test_data/parser/inline/ok/precise_capturing.rast
+++ b/crates/parser/test_data/parser/inline/ok/precise_capturing.rast
@@ -48,18 +48,19 @@ SOURCE_FILE
           WHITESPACE " "
           PLUS "+"
           WHITESPACE " "
-          USE_KW "use"
-          GENERIC_PARAM_LIST
-            L_ANGLE "<"
-            LIFETIME_PARAM
-              LIFETIME
-                LIFETIME_IDENT "'b"
-            COMMA ","
-            WHITESPACE " "
-            TYPE_PARAM
-              NAME
-                IDENT "T"
-            R_ANGLE ">"
+          TYPE_BOUND
+            USE_KW "use"
+            GENERIC_PARAM_LIST
+              L_ANGLE "<"
+              LIFETIME_PARAM
+                LIFETIME
+                  LIFETIME_IDENT "'b"
+              COMMA ","
+              WHITESPACE " "
+              TYPE_PARAM
+                NAME
+                  IDENT "T"
+              R_ANGLE ">"
     WHITESPACE " "
     BLOCK_EXPR
       STMT_LIST
--- a/crates/parser/test_data/parser/inline/ok/precise_capturing.rast
+++ b/crates/parser/test_data/parser/inline/ok/precise_capturing.rast
@@ -1,0 +1,68 @@
+SOURCE_FILE
+  FN
+    FN_KW "fn"
+    WHITESPACE " "
+    NAME
+      IDENT "captures"
+    GENERIC_PARAM_LIST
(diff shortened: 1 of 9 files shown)


crates/parser/test_data/parser/inline/ok/precise_capturing.rast
```diff
@@ -0,0 +1,68 @@
...
+          TYPE_BOUND
+            PATH_TYPE
+              PATH
+                PATH_SEGMENT
+                  NAME_REF
+                    IDENT "Sized"
+          WHITESPACE " "
+          PLUS "+"
+          WHITESPACE " "
+          USE_KW "use"
```
DropDemBits, line 51: This should be parsed into it's own `TYPE_BOUND` rather than being part of the previous bound. I believe this is because it's part of the match handling effects before a path, and as it's not finding a path, it hits the `abandon` and dumps the tokens into the proceeding `TYPE_BOUND`.

Moving the parsing into the outer match should make it parse correctly 🙂

DropDemBits (changes requested):

crates/parser/test_data/parser/inline/ok/precise_capturing.rast
winstxnhdw: I had it on the outer arm at first but it kept panicking :/ Will try again..

Veykril:
https://github.com/rust-lang/rust-analyzer/blob/master/crates/syntax/src/ast/node_ext.rs#L791-L814 needs a change as well

crates/syntax/rust.ungram
```diff
@@ -629,6 +629,7 @@ TypeBoundList =
 TypeBound =
   Lifetime
 | ('~' 'const' | 'const')? 'async'? '?'? Type
+| 'use' GenericParamList?
```
Veykril, line 632:
```suggestion
| 'use' GenericParamList
```

winstxnhdw: Oops. Sorry for this.

winstxnhdw:
> https://github.com/rust-lang/rust-analyzer/blob/eb5da56d839ae0a9e9f50774fa3eb78eb0964550/crates/syntax/src/ast/node_ext.rs#L791-L814
> needs a change as well

It's a lot more complex than I thought. Few things I'd like to discuss. What should this new `use` `TypeBoundKind` be named? I am not sure if we can just reuse the old `ast::Use`. And if we do, I'll have to add the `generic_param_list` method which may add some confusion as it isn't used in `use` declarations.

Veykril:
`Use(ast::GenericParamList)` should work fine, note that we can ignore this when lowering later for now. We just want this here to prevent the panics

Veykril:
https://github.com/rust-lang/rust-analyzer/blob/62a7468d46f88fdf4d078681081c46b2cd17045e/crates/hir-def/src/hir/type_ref.rs#L355-L385
needs to handle the new variant, can just return `TypeBound::Error` in there

Veykril:
CI failure should be unrelated now
@bors r+

PR #17690, opened 2024-07-25
internal: Fix and enable unsafe_op_in_unsafe_fn

### LINKED ISSUE #17689: Enable and fix `unsafe_op_in_unsafe_fn` lint in rust-analyzer
That is add it to our `Cargo.toml` lints as `warn` and then fix the warnings. This is slated to be enabled in 2024 edition by default

lnicola:
Closes #17689

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,7 +15,7 @@ env:
   CARGO_NET_RETRY: 10
   CI: 1
   RUST_BACKTRACE: short
-  RUSTFLAGS: "-D warnings -W unreachable-pub -W bare-trait-objects"
+  RUSTFLAGS: "-D warnings -D elided_lifetimes_in_paths -D explicit_outlives_requirements -D unsafe_op_in_unsafe_fn -D unused_extern_crates -D unused_lifetimes -D unreachable_pub"
   RUSTUP_MAX_RETRIES: 10
 
 jobs:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -15,7 +15,7 @@ env:
   CARGO_NET_RETRY: 10
   CI: 1
   RUST_BACKTRACE: short
-  RUSTFLAGS: "-D warnings -W unreachable-pub -W bare-trait-objects"
```
lnicola, line 18: These are warn by default, I don't think we need them here.

Cargo.toml
```diff
@@ -163,14 +163,14 @@ xshell = "0.2.5"
 dashmap = { version = "=5.5.3", features = ["raw-api"] }
 
 [workspace.lints.rust]
-bare_trait_objects = "warn"
+# remember to update RUSTFLAGS in ci.yml if you add something here
```
Veykril, line 166: Wonder if we can make this work out better for CI 🤔

lnicola: We could just set them to `deny`, I don't think it's going to be _too_ annoying.

I don't think we can customize them for specific profiles: https://users.rust-lang.org/t/changing-lint-configuration-based-on-cargo-profile/107286/2.

Veykril: > I don't think it's going to be too annoying.

That will stop check early won't it? That will absolutely annoy me when refactoring because I don't fix all of those warnings immediately

lnicola: That depends on whether you're using `--keep-going` or not :smile:.

Veykril: Well no, because any dependency that errors (due to this) will prevent dependents from building still 😅 At least that is my understanding of `--keep-going`, it will make cargo no bail out on other independent work.

PR #17705, opened 2024-07-26
feat: Use oldest rustup rust-analyzer when toolchain override is present

### LINKED ISSUE #17663: VSCode should honor toolchain overrides that contain a rust-analyzer component for server spawning
We tend to drop support for stable releases somewhat frequently as its difficult to really support more than a handful of rust versions with a rolling release model. This means it becomes the job of the user to pin their rust-analyzer somehow to some version that fits their current roolchain install, that's not ideal. We should offer a setting that when enabled checks for a toolchain override in the workspace and attempts to use its rust-analyzer component if installed when spawning the server.

Addtionally we should probably attempt to let the server know its MSRV and inform the user if their toolchain looks too old.

huntc:
Selects a rust-toolchain declared RA based on its date. The earliest (oldest) RA wins and becomes the one that the workspace uses as a whole.

In terms of precedence:

nightly > stable-with-version > stable

With stable-with-version, we invoke the RA with a `--version` arg and attempt to extract a date. Given the same date as a nightly, the nightly RA will win.

Fixes #17663

### DIFF
--- a/editors/code/src/bootstrap.ts
+++ b/editors/code/src/bootstrap.ts
@@ -4,6 +4,7 @@ import type { Config } from "./config";
 import { type Env, log } from "./util";
 import type { PersistentState } from "./persistent_state";
 import { exec, spawnSync } from "child_process";
+import { TextDecoder } from "node:util";
 
 export async function bootstrap(
     context: vscode.ExtensionContext,
@@ -50,26 +51,35 @@ async function getServer(
         }
         return explicitPath;
     }
-    if (packageJson.releaseTag === null) return "rust-analyzer";
 
-    if (vscode.workspace.workspaceFolders?.length === 1) {
-        // otherwise check if there is a toolchain override for the current vscode workspace
-        // and if the toolchain of this override has a rust-analyzer component
-        // if so, use the rust-analyzer component
-        const toolchainTomlExists = await fileExists(
-            vscode.Uri.joinPath(vscode.workspace.workspaceFolders[0]!.uri, "rust-toolchain.toml"),
-        );
-        if (toolchainTomlExists) {
-            const res = spawnSync("rustup", ["which", "rust-analyzer"], {
-                encoding: "utf8",
-                env: { ...process.env },
-                cwd: vscode.workspace.workspaceFolders[0]!.uri.fsPath,
-            });
-            if (!res.error && res.status === 0) {
-                return res.stdout.trim();
+    let toolchainServerPath = undefined;
+    if (vscode.workspace.workspaceFolders) {
+        for (const workspaceFolder of vscode.workspace.workspaceFolders) {
+            // otherwise check if there is a toolchain override for the current vscode workspace
+            // and if the toolchain of this override has a rust-analyzer component
+            // if so, use the rust-analyzer component
+            const toolchainUri = vscode.Uri.joinPath(workspaceFolder.uri, "rust-toolchain.toml");
+            if (await hasToolchainFileWithRaDeclared(toolchainUri)) {
+                const res = spawnSync("rustup", ["which", "rust-analyzer"], {
+                    encoding: "utf8",
+                    env: { ...process.env },
+                    cwd: workspaceFolder.uri.fsPath,
+                });
+                if (!res.error && res.status === 0) {
+                    toolchainServerPath = earliestToolchainPath(
+                        toolchainServerPath,
+                        res.stdout.trim(),
+                        raVersionResolver,
+                    );
+                }
             }
         }
     }
+    if (toolchainServerPath) {
+        return toolchainServerPath;
+    }
+
+    if (packageJson.releaseTag === null) return "rust-analyzer";
 
     // finally, use the bundled one
     const ext = process.platform === "win32" ? ".exe" : "";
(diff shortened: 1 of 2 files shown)


editors/code/src/bootstrap.ts
```diff
@@ -52,24 +52,39 @@ async function getServer(
...
+            );
+            if (toolchainTomlExists) {
+                const res = spawnSync("rustup", ["which", "rust-analyzer"], {
+                    encoding: "utf8",
+                    env: { ...process.env },
+                    cwd: workspaceFolder.uri.fsPath,
+                });
+                if (!res.error && res.status === 0) {
+                    toolchainServerPath = earliestToolchainPath(toolchainServerPath, res.stdout.trim(), function (path) {
+                        const res = spawnSync(path, ["version"], {
```
Veykril, line 72: Wouldn't it make more sense to query the rustup version for the workspace here?

huntc: This is a lambda where the `path` comes from within evaluating the toolchain path. It is invoked conditionally depending on the type of path to the RA it is i.e. if the RA isn't pointing to a nightly or stable one, then the path is deemed to be an explicity version of RA.

Also, this is a lambda for two reasons:

* other `spawnSync` calls are made within this `getServer` function; and
* so that I can write unit tests easily for `earliestToolchainPath` and `orderFromPath`.

Veykril: I'm confused, `rustup which rust-analyzer` should only yield a rust toolchain version (if installed in the toolchain) or error no?

huntc: It appears to yield paths with stable, nightly and [some-version] in them. At least in my workspace. I’ll create some unit tests with sample paths to illustrate as I move forward.

huntc: Or may be I don’t understand what you’re asking… if we are talking about the lambda still, that path can be a path to a nightly, stable or versioned RA.

huntc:
@Veykril Can you please direct me to some doco in your to conveniently test our the extension?

editors/code/src/bootstrap.ts
Veykril: Oh I see, we can return a manual (non-toolchain component) rust-analyzer install from this. Okay then this handling does make sense

Veykril:
I don't think we can easily test this. We generally don't have a good testing infra setup for the vscode extension unfortunately.

editors/code/src/bootstrap.ts
Veykril: Or actually, what is being invoked here, `path/to/rust-analyzer version`? That wouldn't make sense as that is not valid command for r-a. The flag is `--version`. Can you add a comment above that explaining roughly whats being executed for visibility? Also for the regex part a comment showing an example input and what we disassemble that to via the regexes(regices?) would be helpful for readability

huntc: Absolutely, I’ll get on to some unit tests tomorrow which should make things a bit more obvious  Also, the test infra is there already (npm test).

I just wanted to make sure I was on the right track etc. Thanks for the initial review.

huntc:
> I don't think we can easily test this. We generally don't have a good testing infra setup for the vscode extension unfortunately.

Sorry, I meant, how do I test this manually and locally? I can see an option to install the VSIX file, but I'm unsure if it updates an existing one.

huntc:
OK, this is working, although I think we need to go a bit further...

At the moment, the code will identify RA's for each rust-toolchain.toml file found. This isn't exactly what we want though. We should only select an RA with a toolchain if it is explicitly set via `components`. I'll look at doing that next.

huntc:
> OK, this is working, although I think we need to go a bit further...
> 
> At the moment, the code will identify RA's for each rust-toolchain.toml file found. This isn't exactly what we want though. We should only select an RA with a toolchain if it is explicitly set via `components`. I'll look at doing that next.

BTW I figured this out by installing the extension from its VSIX. I realised that `npm run package` was also required.

huntc:
This PR has now been tested in my environment that has a couple of projects pinned to old toolchains. Having added RA as a component to the relevant toolchain files causes the corresponding version of RA to be used. I can confirm this by using the "show RA version" command.

I am getting the following warnings reported and don't quite understand why:

```
[ERROR rust_analyzer::reload] Failed to run proc-macro server from path /Users/myuser/.rustup/toolchains/stable-aarch64-apple-darwin/libexec/rust-analyzer-proc-macro-srv, error: Custom { kind: Other, error: "proc-macro server's api version (4) is newer than rust-analyzer's (2)" }
```

Do we need to do something with the proc-macro server as well?

editors/code/src/bootstrap.ts
```diff
@@ -50,26 +51,35 @@ async function getServer(
...
+                    );
+                }
             }
         }
     }
+    if (toolchainServerPath) {
+        return toolchainServerPath;
+    }
+
+    if (packageJson.releaseTag === null) return "rust-analyzer";
```
huntc, line 82: Please note that I relocated this check to here - I think it is just for dev mode, right? The original position would bypass any checking of RA in the toolchain files.

editors/code/src/bootstrap.ts
```diff
@@ -205,3 +291,8 @@ async function patchelf(dest: vscode.Uri): Promise<void> {
         },
     );
 }
+
+export const _private = {
+    earliestToolchainPath,
+    orderFromPath,
+};
```
huntc, line 298: This appears to be the TS way of "hiding" things that need to be exported purely for testing.

Veykril:
> This PR has now been tested in my environment that has a couple of projects pinned to old toolchains. Having added RA as a component to the relevant toolchain files causes the corresponding version of RA to be used. I can confirm this by using the "show RA version" command.
> 
> I am getting the following warnings reported and don't quite understand why:
> 
> ```
> [ERROR rust_analyzer::reload] Failed to run proc-macro server from path /Users/myuser/.rustup/toolchains/stable-aarch64-apple-darwin/libexec/rust-analyzer-proc-macro-srv, error: Custom { kind: Other, error: "proc-macro server's api version (4) is newer than rust-analyzer's (2)" }
> ```
> 
> Do we need to do something with the proc-macro server as well?

That is a problem we can't work around unfortunately. The problem here is that the rust-analyzer is too old for one of the newer toolchains in the project being used, so it doesn't speak the proc-macro servers protocol anymore. We can't use a different proc-macro server though as it is required for the proc-macro server to built with the same toolchain as the proc-macros that are being loaded.

huntc:
> > This PR has now been tested in my environment that has a couple of projects pinned to old toolchains. Having added RA as a component to the relevant toolchain files causes the corresponding version of RA to be used. I can confirm this by using the "show RA version" command.
> > I am getting the following warnings reported and don't quite understand why:
> > ```
> > [ERROR rust_analyzer::reload] Failed to run proc-macro server from path /Users/myuser/.rustup/toolchains/stable-aarch64-apple-darwin/libexec/rust-analyzer-proc-macro-srv, error: Custom { kind: Other, error: "proc-macro server's api version (4) is newer than rust-analyzer's (2)" }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do we need to do something with the proc-macro server as well?
> 
> That is a problem we can't work around unfortunately. The problem here is that the rust-analyzer is too old for one of the newer toolchains in the project being used, so it doesn't speak the proc-macro servers protocol anymore. We can't use a different proc-macro server though as it is required for the proc-macro server to built with the same toolchain as the proc-macros that are being loaded.

Ok. I guess we must live with that then. At least the changes here avoid broken development experiences.

lnicola:
> BTW I figured this out by installing the extension from its VSIX. I realised that npm run package was also required.

`cargo xtask install --client` should do the trick.

> regexes (regices?)

Regexen, of course :upside_down_face:.

editors/code/src/bootstrap.ts
```diff
@@ -102,13 +112,89 @@ async function getServer(
...
+
+// Further to extracting a date for comparison, determine the order of a toolchain as follows:
+//  Highest - nightly
+//  Medium  - versioned
+//  Lowest  - stable
+// Example paths:
+//  nightly   - /Users/myuser/.rustup/toolchains/nightly-2022-11-22-aarch64-apple-darwin/bin/rust-analyzer
+//  versioned - /Users/myuser/.rustup/toolchains/1.72.1-aarch64-apple-darwin/bin/rust-analyzer
+//  stable    - /Users/myuser/.rustup/toolchains/stable-aarch64-apple-darwin/bin/rust-analyzer
+function orderFromPath(
```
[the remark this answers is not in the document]
huntc: Please note that we're not actually checking the path itself, but the RA that resides at that path is being asked for its version.

However, I've now actually enhanced the logic to always ask the RA for its version info, even if is stable, beta etc. The logic now stands that given the same date, nightly will take precedence over anything else. This actually simplifies the code.

Veykril (approved):

Veykril:
Looks great, thank you for implementing this!
@bors r+

huntc:
> Looks great, thank you for implementing this! @bors r+

My pleasure. Thanks for the guidance also.

editors/code/src/bootstrap.ts
lnicola: > The original position would bypass any checking of RA in the toolchain files.

Yeah, when running a dev build, I'd expect it to always use my local r-a, so I'm not sure about this change.

huntc: Indeed, if you've declared that you want to use an explicit version of RA (and only then), it'll get used, whether in dev mode or not. I think that makes sense - why would you want to invoke a local dev build for an old toolchain?

PR #17714, opened 2024-07-27
docs: Reference cov_mark crate instead of test_utils/mark

Throne3d:
Seems this was switched over in https://github.com/rust-lang/rust-analyzer/issues/7922

I'm guessing this should skip the changelog because it's so minor, right? Reading through https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/style.md and trying to get up to speed here =)


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #14930, opened 2023-05-30
fix: let glob imports override other globs' visibility

lowr:
Fixes #11858
Fixes #14902

PR #17715, opened 2024-07-27
fix: let glob imports override other globs' visibility

### LINKED ISSUE #11858: conflict between `use` and `pub use` in crate re-exports
crate `mylib`:
```rust
mod bar {
  pub struct Foo;
}
use bar::*;
pub use bar::*;
```
`main.rs`:
```rust
use mylib::Foo;

pub fn f(_: &Foo) {}
fn main() {
  let foo = mylib::Foo;
  f(&foo); // type mismatch: expected &{unknown}, found &Foo
}
```
Other observations about this example: hovering over the reference to `Foo` in `f` shows that it is not recognized, which is why `&{unknown}` is expected in the type mismatch. `rustc` compiles this without issue. If both glob imports are replaced by imports of `bar::Foo`, then it fails to compile because two `Foo` names are introduced in the scope of `mylib`; if only the private one is replaced by `Foo` then rustc will treat `Foo` as private even though there is a public glob use. (It is not clear whether this behavior is intentional / desirable in rustc: [zulip](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Adding.20private.20imports.20can.20break.20downstream.20crates), rust-lang/rust#95420.)

**rust-analyzer version**: `63c4d6b20 2022-03-30 nightly`

**rustc version**: `rustc 1.61.0-nightly (52b34550a 2022-03-15)`

### LINKED ISSUE #14902: unresolvedReference for trait arrow_array::ArrowPrimitiveType
**rust-analyzer version**: (eg. output of "rust-analyzer: Show RA Version" command, accessible in VSCode via <kbd>Ctrl/⌘</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd>) rust-analyzer version: 0.4.1529-standalone (06d02e0a1 2023-05-25)

**rustc version**: (eg. output of `rustc -V`) rustc 1.69.0 (84c898d65 2023-04-16)

**relevant settings**: (eg. client settings, or environment variables like `CARGO`, `RUSTC`, `RUSTUP_HOME` or `CARGO_HOME`) default settings

image

```rust
use arrow_array::ArrowPrimitiveType;
```

```toml
[dependencies]
arrow-array = "40.0.0" 
```

### LINKED ISSUE #17704: Issue loading type data for arrow_array::ArrowPrimitiveType
**rust-analyzer version**: `rust-analyzer version: 0.3.2045-standalone`

**editor or extension**: VSCode, extension version v0.3.2045

**relevant settings**: I specifically noticed this because it flagged as an error under `"rust-analyzer.diagnostics.experimental.enable": true`, but it seems to present, just more subtly, with all my rust-analyzer configs disabled: rust-analyzer doesn't seem to provide meaningful completions for the type when imported at this path.

I first noticed this with the `hash_one` method call when getting started looking at the Apache Datafusion codebase: rust-analyzer won't provide completion and documentation hovers for this `hash_one` method call (and, with the experimental diagnostics enabled, says the hash_one method doesn't exist, even though it passes `cargo check`): https://github.com/apache/datafusion/blob/fab7e238e8559bdb2e4e0449354f7efeccbc07a7/datafusion/common/src/hash_utils.rs#L118

I think this is a successful (though a bit weird) minimization:
```rust
// [dependencies]
// arrow-array = "52.1.0"
fn func1<T>()
where T: arrow_array::ArrowPrimitiveType<Native = i16>,
{
    println!("{}", T::default_value());
}

fn main() {
    func1::<arrow_array::types::Int16Type>();
}
```

Here, this passes `cargo check` fine, and successfully compiles and outputs `0` when run, but shows no documentation hover for `T::default_value()` and (when enabled) the experimental check shows that the associated item doesn't exist. Based on the output of `analysis-stats`, I think something's going wrong with "MIR failed bodies" or "Body lowering" in rust-analyzer, but I couldn't investigate that in more detail:

```
$ ./rust-analyzer analysis-stats . 2>&1 | grep -v "perf counter"
Database loaded:     264.04ms, 12mb (metadata 159.28ms, 178kb; build 18.25ms, 10kb)
  item trees: 1
Item Tree Collection: 254.94µs, 33kb
  crates: 1, mods: 1, decls: 2, bodies: 2, adts: 0, consts: 0
Item Collection:     2.70s, 162mb
Body lowering:       328.05ms, 51mb
  exprs: 48, ??ty: 2 (4%), ?ty: 2 (4%), !ty: 0
  pats: 0, ??ty: 0 (100%), ?ty: 0 (100%), !ty: 0
Inference:           349.40ms, 11mb
MIR lowering:        920.54µs, 17kb
Mir failed bodies: 1 (50%)
Data layouts:        85.00ns, 0b
Failed data layouts: 0 (100%)
Const evaluation:    57.00ns, 0b
Failed const evals: 0 (100%)
Total:               3.38s, 225mb
```

Referencing the type as `arrow_array::types::ArrowPrimitiveType` instead of `arrow_array::ArrowPrimitiveType` seems to solve completions, documentation hovers, and the experimental diagnostic, so I think there's something about the import path (maybe import depth?) - `arrow_array` seems to `pub use array::*`, which itself has `use crate::types::*` (not public) and `pub use primitive_array::*`, the latter of which has `pub use crate::types::ArrowPrimitiveType` with the final `pub trait ArrowPrimitiveType` definition. Maybe there's some conflict due to the public and non-public imports of the same type here? Here's the analysis-stats output I get with the `::types::` path that RA provides hints for:

```
$ ./rust-analyzer analysis-stats . 2>&1 | grep -v "perf counter"
Database loaded:     208.97ms, 12mb (metadata 103.67ms, 178kb; build 18.08ms, 10kb)
  item trees: 1
Item Tree Collection: 222.25µs, 35kb
  crates: 1, mods: 1, decls: 2, bodies: 2, adts: 0, consts: 0
Item Collection:     2.63s, 162mb
Body lowering:       298.56ms, 51mb
  exprs: 48, ??ty: 0 (0%), ?ty: 0 (0%), !ty: 0
  pats: 0, ??ty: 0 (100%), ?ty: 0 (100%), !ty: 0
Inference:           329.07ms, 11mb
MIR lowering:        9.59ms, 40kb
Mir failed bodies: 0 (0%)
Data layouts:        168.00ns, 0b
Failed data layouts: 0 (100%)
Const evaluation:    55.00ns, 0b
Failed const evals: 0 (100%)
Total:               3.26s, 225mb
```

Please let me know if there's anything else I can do to help reproduce this!

Throne3d:
Follow up to #14930

Fixes #11858
Fixes #14902
Fixes #17704

I haven't reworked the code here at all - I don't feel confident in the codebase to do so - just rebased it onto the current main branch and fixed conflicts.

I'm not _entirely_ sure I understand the structure of the `check` function in `crates/hir-def/src/nameres` tests. I think the change to the test expectation from #14930 is correct, marking the `crate::reexport::inner` imports with `i`, and I understand it to mean there's a specific token in the import that we can match it to (in this case, `Trait`, `function` and `makro` of `pub use crate::defs::{Trait, function, makro};` respectively), but I had some trouble understanding the meaning of the different parts of `PerNs` to be sure.
Does this make sense?

I tested building and using RA locally with `cargo xtask install` and after this change the documentation for `arrow_array::ArrowPrimitiveType` seems to be picked up correctly!

### DIFF
--- a/crates/hir-def/src/nameres/collector.rs
+++ b/crates/hir-def/src/nameres/collector.rs
@@ -1025,7 +1025,7 @@ impl DefCollector<'_> {
 
     fn update_recursive(
         &mut self,
-        // The module for which `resolutions` have been resolve
+        // The module for which `resolutions` have been resolved.
         module_id: LocalModuleId,
         resolutions: &[(Option<Name>, PerNs)],
         // All resolutions are imported with this visibility; the visibilities in
@@ -1043,10 +1043,9 @@ impl DefCollector<'_> {
         for (name, res) in resolutions {
             match name {
                 Some(name) => {
-                    let scope = &mut self.def_map.modules[module_id].scope;
-                    changed |= scope.push_res_with_import(
-                        &mut self.from_glob_import,
-                        (module_id, name.clone()),
+                    changed |= self.push_res_and_update_glob_vis(
+                        module_id,
+                        name,
                         res.with_visibility(vis),
                         import,
                     );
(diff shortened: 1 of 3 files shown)


crates/hir-def/src/nameres/collector.rs
```diff
@@ -1112,6 +1111,84 @@ impl DefCollector<'_> {
...
+        let mut changed = false;
+
+        if let Some(ImportType::Glob(_)) = def_import_type {
+            let prev_defs = self.def_map[module_id].scope.get(name);
+
+            // Multiple globs may import the same item and they may override visibility from
+            // previously resolved globs. Handle overrides here and leave the rest to
+            // `ItemScope::push_res_with_import()`.
+            if let Some((def, def_vis, _)) = defs.types {
+                if let Some((prev_def, prev_vis, _)) = prev_defs.types {
```
Throne3d, line 1130: I believe the third parameter here is the location/token the import occurred at - it seems it wasn't present at the time the code was written, but I don't think there's a condition necessary on it here, so hopefully it's correct to leave it ignored here.

If you can think of any test cases to add to validate if there are other necessary conditions here, I'd appreciate any suggestions!

PR #17735, opened 2024-07-29
feat: Introduce workspace `rust-analyzer.toml`s

alibektas:
In order to globally configure a project it was, prior to this PR, possible to have a `ratoml` at the root path of a project. This is not the case anymore. Instead we now let ratoml files that are placed at the root of any workspace have a new scope called `workspace`. Although there is not a difference between a `workspace` scope and and a `global` scope, future PRs will change that.

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -2552,7 +2552,7 @@ macro_rules! _impl_for_config_data {
                 $($doc)*
                 #[allow(non_snake_case)]
                 $vis fn $field(&self, source_root: Option<SourceRootId>) -> &$ty {
-                    let mut source_root = source_root;
+                    let mut source_root = source_root.as_ref();
                     while let Some(sr) = source_root {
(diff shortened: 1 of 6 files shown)


crates/rust-analyzer/src/config.rs
```diff
@@ -1333,11 +1354,6 @@ impl Config {
         // FIXME @alibektas : Temporary solution. I don't think this is right as at some point we may allow users to specify
         // custom USER_CONFIG_PATHs which may also be relative.
         let user_config_path = VfsPath::from(AbsPathBuf::assert(user_config_path));
```
Veykril, line 1356: On a follow up note, I feel like we should not make this configurable, where would you even configure this? I guess in the lient config but that still feels a bit odd?

alibektas: by setting a command line argument to the server but ok I will remove it

alibektas: I took a note of it and will make a separate PR for this.

Veykril (approved):

PR #17736, opened 2024-07-29
feat(ide-completion): explictly show `async` keyword on `impl trait` methods

hyf0:
OLD:

image

Now:

image

---

This is an preparation for https://github.com/rust-lang/rust-analyzer/issues/17719.

```rust
use std::future::Future;

trait DesugaredAsyncTrait {
    fn foo(&self) -> impl Future<Output = usize> + Send;
    fn bar(&self) -> impl Future<Output = usize> + Send;
}

struct Foo;

impl DesugaredAsyncTrait for Foo {
    fn foo(&self) -> impl Future<Output = usize> + Send {
        async { 1 }
    }

    //
    async fn bar(&self) -> usize {
        1
    }
}

fn main() {
    let fut = Foo.bar();
    fn _assert_send<T: Send>(_: T) {}
    _assert_send(fut);
}
```

If we don't distinguish `async` or not. It would be confusing to generate sugared version `async fn foo ....` and original form `fn foo`  for `async fn in trait` that is defined in desugar form.

### DIFF
--- a/crates/ide-diagnostics/src/handlers/trait_impl_missing_assoc_item.rs
+++ b/crates/ide-diagnostics/src/handlers/trait_impl_missing_assoc_item.rs
@@ -13,7 +13,7 @@ pub(crate) fn trait_impl_missing_assoc_item(
 ) -> Diagnostic {
     let missing = d.missing.iter().format_with(", ", |(name, item), f| {
         f(&match *item {
-            hir::AssocItem::Function(func) if func.is_async(ctx.sema.db)  => "`async fn ",
+            hir::AssocItem::Function(func) if func.is_async(ctx.sema.db) => "`async fn ",
             hir::AssocItem::Function(_) => "`fn ",
             hir::AssocItem::Const(_) => "`const ",
             hir::AssocItem::TypeAlias(_) => "`type ",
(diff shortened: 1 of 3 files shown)


hyf0:
Fixed error message

image

crates/ide-diagnostics/src/handlers/trait_impl_missing_assoc_item.rs
```diff
@@ -13,6 +13,7 @@ pub(crate) fn trait_impl_missing_assoc_item(
 ) -> Diagnostic {
     let missing = d.missing.iter().format_with(", ", |(name, item), f| {
         f(&match *item {
+            hir::AssocItem::Function(func) if func.is_async(ctx.sema.db) => "`async fn ",
```
Veykril, line 16: Let's not do that, the important information in the diagnostic here is what item is missing, we aren't reporting the entire signature in the first place to reduce information bloat.

Veykril (commented):
Adding an async function to the fixture here would be good https://github.com/rust-lang/rust-analyzer/blob/7beac14cba1d1ba85011660e5e85159c6f3a136c/crates/ide-completion/src/tests/item_list.rs#L291-L321

crates/ide-diagnostics/src/handlers/trait_impl_missing_assoc_item.rs
hyf0: Done

Veykril:
Thanks!
@bors r+

Issue #17719, opened 2024-07-27
Auto completion of `async fn xxxx` for desugared `async fn in trait` form


hyf0:
Hi. I'm thinking about improving the dx of writing `impl DesugaredAsyncTrait for Foo`. So far the rust analyzer will only infer auto completion like

![image](https://github.com/user-attachments/assets/525418e3-8561-4921-97ca-c7e71dacf228)

However, I found that the compiler will allow the following code compiles. ~If this is an expected behavior~(It's expected, see the following zulipchat link), maybe it's ok to make rust analyzer to support auto completion that starts with `async fn foo`( the sugar version rather than only desugared version like now) for `DesugaredAsyncTrait `.

This will surely improve the dx of writing `impl DesugaredAsyncTrait for XXX` greatly without the need to write `impl Future<Output=XXXX> + Send` over and over again(Though there are also auto completed by the rust-analyzer :).

```rust
use std::future::Future;

trait DesugaredAsyncTrait {
    fn foo(&self) -> impl Future<Output = usize> + Send;
    fn bar(&self) -> impl Future<Output = usize> + Send;
}

struct Foo;

impl DesugaredAsyncTrait for Foo {
    fn foo(&self) -> impl Future<Output = usize> + Send {
        async { 1 }
    }

    //
    async fn bar(&self) -> usize {
        1
    }
}

fn main() {
    let fut = Foo.bar();
    fn _assert_send<T: Send>(_: T) {}
    _assert_send(fut);
}
```

Related:
- https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Auto.20infered.20.60Send.60.20in.20.60impl.20DesugaredAsync.60.2E.20Expected.3F
- https://x.com/Gankra_/status/1141409682017308672

hyf0:
There's a question on `Quick Fix "Implement missing members"` if we support this PR.

On "Implement missing members", should we fill them in sugar version or original form for `async fn in trait` in desugar form?

Veykril:
I'd say sugared version, and then if we don't have it yet, have a desugar assist that goes from `async` to `impl Future` (I think we already have this though)

lnicola:
Fixed in #17737

PR #17737, opened 2024-07-29
feat(ide-completion): extra sugar auto-completion `async fn ...` in `impl trait` for `async fn in trait` that's defined in desugar form

hyf0:
Solves #17719.

---

Preview

image

image

image

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -2207,51 +2207,33 @@ impl Function {
         db.function_data(self.id).is_async()
     }
 
-    /// Whether this function is a `fn` that returns `impl Future`.
-    pub fn is_desugar_async(self, db: &dyn HirDatabase) -> bool {
-        if self.is_async(db) || self.is_const(db) {
-            return false;
+    pub fn returns_impl_future(self, db: &dyn HirDatabase) -> bool {
+        if self.is_async(db) {
+            return true;
         }
 
         let Some(impl_traits) = self.ret_type(db).as_impl_traits(db) else { return false };
-
         let Some(future_trait_id) =
             db.lang_item(self.ty(db).env.krate, LangItem::Future).and_then(|t| t.as_trait())
         else {
             return false;
         };
-
-        let Some(size_trait_id) =
+        let Some(sized_trait_id) =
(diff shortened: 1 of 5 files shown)


crates/hir/src/lib.rs
```diff
@@ -2181,6 +2181,53 @@ impl Function {
...
+            return false;
+        };
+
+        let Some(sync_trait_id) =
+            db.lang_item(self.ty(db).env.krate, LangItem::Sync).and_then(|t| t.as_trait())
+        else {
+            return false;
+        };
+
+        // TODO: There's no `LangItem::Send`. How do we get the id of `Send` trait?
```
hyf0, line 2210: Help wanted here🙋

crates/ide-completion/src/completions/item_list/trait_impl.rs
```diff
@@ -210,7 +210,7 @@ fn add_function_impl(
                 ast::AssocItem::Fn(func) => func,
                 _ => unreachable!(),
             };
-
+            // TODO: need `function_decl` that unwraps future in the return type
```
hyf0, line 213: Help wanted here🙋

crates/hir/src/lib.rs
Veykril: Huh that is very surprising. Goes to show that the Send/Sync stuff really is more a library level concern than a language one, interesting.

I think we can skip looking for those two all together and instead check whether the impl traits (aside from the principal `Future` one) are auto traits or not. That is more general, as the relevant part about those traits being leaked is them being auto traits, not them being `Send` or `Sync`

crates/ide-completion/src/completions/item_list/trait_impl.rs
Veykril: You'll have to manually traverse the AST here and stitch the edit together, that is something like `transformed_fn.ret_type()` to fetch the return type, then you can match on that to check if its an `ImplTrait` one etc

Veykril:
Finished up the work here, hope you don't mind :)

@bors r+

hyf0:
Not at all. Thank you very much. Finally no more manual async anymore :).

PR #17745, opened 2024-07-30
Improve crate manifests, adding missing `[package.repository]` and `[package.description]` fields

regexident:
As [discussed on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/Could.20we.20add.20repository.20url.20to.20.60ra_ap_.60.20crates.20on.20crates.2Eio.3F/near/455095161).

cc @Veykril @lnicola

### DIFF
--- a/crates/base-db/Cargo.toml
+++ b/crates/base-db/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "base-db"
 version = "0.0.0"
-repository = "https://github.com/rust-lang/rust-analyzer"
+repository.workspace = true
 description = "Basic database traits for rust-analyzer. The concrete DB is defined by `ide` (aka `ra_ap_ide`)."
 
 authors.workspace = true
--- a/crates/base-db/Cargo.toml
+++ b/crates/base-db/Cargo.toml
@@ -2,7 +2,7 @@
 name = "base-db"
 version = "0.0.0"
 repository = "https://github.com/rust-lang/rust-analyzer"
-description = "Basic database traits. The concrete DB is defined by `ide` (aka `ra_ap_ide`)."
+description = "Basic database traits for rust-analyzer. The concrete DB is defined by `ide` (aka `ra_ap_ide`)."
 
 authors.workspace = true
 edition.workspace = true
--- a/crates/base-db/Cargo.toml
+++ b/crates/base-db/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "base-db"
 version = "0.0.0"
-description = "TBD"
+description = "Basic database traits. The concrete DB is defined by `ra_ap_ide`."
 
 authors.workspace = true
 edition.workspace = true
--- a/crates/base-db/Cargo.toml
+++ b/crates/base-db/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "base-db"
 version = "0.0.0"
+repository = "https://github.com/rust-lang/rust-analyzer"
 description = "Basic database traits. The concrete DB is defined by `ra_ap_ide`."
 
 authors.workspace = true
--- a/crates/base-db/Cargo.toml
+++ b/crates/base-db/Cargo.toml
@@ -2,7 +2,7 @@
 name = "base-db"
 version = "0.0.0"
 repository = "https://github.com/rust-lang/rust-analyzer"
-description = "Basic database traits. The concrete DB is defined by `ra_ap_ide`."
+description = "Basic database traits. The concrete DB is defined by `ide` (aka `ra_ap_ide`)."
 
 authors.workspace = true
 edition.workspace = true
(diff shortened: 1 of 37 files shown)


crates/base-db/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "base-db"
 version = "0.0.0"
-description = "TBD"
+repository = "https://github.com/rust-lang/rust-analyzer"
+description = "Basic database traits. The concrete DB is defined by `ra_ap_ide`."
```
lnicola, line 5: Bah, feels like this should say `ide`.

regexident: I changed it to `ra_ap_ide` as that's the name that the crate is currently published as. As such anyone reading the description on crates.io would be tempted to search for `ide` but not find anything.

Happy to change it back to `ide` though, if that's the consensus of what it should be referred to as.

crates/ide-db/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "ide-db"
 version = "0.0.0"
-description = "TBD"
+repository = "https://github.com/rust-lang/rust-analyzer"
+description = "Core data-structure representing IDE state."
```
lnicola, line 5: Nit:

```suggestion
description = "Core data structures representing IDE state."
```

lnicola:
It's a bit confusing how half of these say "for rust-analyzer" and half don't.

regexident:
> It's a bit confusing how half of these say "for rust-analyzer" and half don't.

Agreed. I was leaning towards adding it to give people stumbling upon a `ra_ap_` crate a little bit of context.
Otherwise "The type system." or the like (as taken verbatim from the crate's `lib.rs` docs) isn't much to hold on to.

For others adding it felt a bit verbose though.

I'm happy with whatever you —the maintainers— prefer.

regexident:
> I'm happy with whatever you —the maintainers— prefer.

@Veykril do you have an opinion on this (as well as the `ide` vs `ra_ap_ide` topic above)?

crates/base-db/Cargo.toml
Veykril: Can we just do something like `ide/ra_ap_ide` then?

crates/hir-def/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "hir-def"
 version = "0.0.0"
-description = "TBD"
+repository = "https://github.com/rust-lang/rust-analyzer"
+description = "Everything between macro expansion and type inference for rust-analyzer."
```
Veykril, line 5:
```suggestion
description = "Early name and import resolution for rust-analyzer."
```
might be more fitting

crates/proc-macro-api/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "proc-macro-api"
 version = "0.0.0"
-description = "TBD"
+repository = "https://github.com/rust-lang/rust-analyzer"
+description = "Client-side proc-macros for rust-analyzer."
```
Veykril, line 5:
```suggestion
description = "RPC Api for the `proc-macro-srv`.
```

Veykril:
not sure whats up with CI, it runs on nightly given the changes so there might be something breaking for us on latest nightly

regexident:
@Veykril FYI I've applied your suggested changes and also added any missing mentions of rust-analyzer (cc @lnicola ) with the following reasoning:

With the lack of a README on the individually published library crates and the somewhat cryptic `ra_ap_` prefix it is hard to figure out where those crates belong to, so mentioning "rust-analyzer" feels like a useful hint there. I've tried to make the name-drop fit into the description without sticking out too much, so it's not always a suffix.

This is quite a bit more than your suggested changes though, so …

@bors r=@Veykril

… just to be sure you're good with those changes. The PR can be merged you're happy with it.

crates/base-db/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "base-db"
 version = "0.0.0"
-description = "TBD"
+repository = "https://github.com/rust-lang/rust-analyzer"
```
Veykril, line 4:
```suggestion
repository.workspace = true
```
Nit, this should be workspace inehritable no?

Veykril:
CI should succeed again after https://github.com/rust-lang/rust-analyzer/pull/17794 is merged

crates/base-db/Cargo.toml
regexident: Now that you mention it, yes of course! Fixed.

Veykril (approved):

Veykril:
Thanks!
@bors r+

lnicola:
@regexident there's one more if you want (syntax-bridge).

PR #17750, opened 2024-07-31
fix: remove AbsPath requirement from linkedProjects

davidbarsky:
Should (fingers crossed!) fix https://github.com/rust-lang/rust-analyzer/issues/17664. I opened the `rustc` workspace with the [suggested configuration](https://github.com/rust-lang/rust/blob/e552c168c72c95dc28950a9aae8ed7030199aa0d/src/etc/rust_analyzer_settings.json) and I was able to successfully open some rustc crates (`rustc_incremental`) and have IDE functionality.

@Veykril: can you try these changes and let me know if it fixed rustc?

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -1677,7 +1677,7 @@ impl Config {
         !self.linkedProjects(None).is_empty()
     }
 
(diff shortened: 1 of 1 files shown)


Veykril:
Out right now so can't check but this does feel like the cause 
@bors r+

PR #17757, opened 2024-08-01
assist: Add new assist toggle_macro_delimiter

### LINKED ISSUE #17716: Assist to toggle between delimiters for macro calls
Simple assist that goes from `foo!()` -> `foo!{}` -> `foo![]` -> `foo!()` when on one of the two delimiters.

alibektas:
Closes #17716

### DIFF
--- a/crates/ide-assists/src/handlers/toggle_macro_delimiter.rs
+++ b/crates/ide-assists/src/handlers/toggle_macro_delimiter.rs
@@ -1,0 +1,227 @@
+use ide_db::assists::{AssistId, AssistKind};
+use syntax::{
+    ast::{self, make},
+    ted, AstNode, T,
+};
+
+use crate::{AssistContext, Assists};
+
+// Assist: toggle_macro_delimiter
+//
+// Change macro delimiters in the order of `( -> { -> [ -> (`.
+//
+// ```
+// macro_rules! sth  ();
+// sth! $0( );
+// ```
+// ->
+// ```
+// macro_rules! sth! ();
+// sth! { }
+// ```
+pub(crate) fn toggle_macro_delimiter(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
+    #[derive(Debug)]
+    enum MacroDelims {
+        LPar,
+        RPar,
+        LBra,
+        RBra,
+        LCur,
+        RCur,
+    }
+
+    enum MakroTypes {
+        MacroRules(ast::MacroRules),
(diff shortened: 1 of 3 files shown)


alibektas:
I see that I don't know how to turn this into a draft, sorry about that. 

One question to which I have no answer atm. I want this assist to be applicable in the following cases. I haven't tested them all but let's concentrate on the first to variants.

```rust
enum MakroTypes {
        MacroRules(ast::MacroRules),
        MacroCall(ast::MacroCall),
        MacroStmt(ast::MacroStmts),
        MacroItem(ast::MacroItems),
    }
```
If we look at the grammar for macro rules we see that a semicolon is omitted when there are curly braces. So I need to remove it when this assist is applied. However there is not a `ast::MacroRules::semicolon_token`. I could easily traverse the token tree and do what I want to do but I want the implementation to be cleaner. Any suggestions?

Veykril:
I'd say don't bother with `macro_rules` (I wasn't even aware that you could use a non `{}` braced token tree there). The way we handle these in the syntax tree is generally a bit inconsistent I think.

alibektas:
Now, my most current problem is that the assist's generated test doesn't even work (no such assist can be applied), even though the exact same test is among my test cases. I can't quite wrap my head around this so once again, I am looking for some help.

crates/ide-assists/src/handlers/toggle_macro_delimiter.rs
```diff
@@ -0,0 +1,256 @@
...
+    enum MacroDelims {
+        LPar,
+        RPar,
+        LBra,
+        RBra,
+        LCur,
+        RCur,
+    }
+
+    let makro = ctx.find_node_at_offset_with_descend::<ast::MacroCall>()?.clone_for_update();
```
Veykril, line 39: You likely don't want to descend into macro expansions here

Veykril:
> Now, my most current problem is that the assist's generated test doesn't even work (no such assist can be applied), even though the exact same test is among my test cases. I can't quite wrap my head around this so once again, I am looking for some help.

I'd sprinkle some `dbg!` in the assist and see where it exits out then

Veykril:
It might be that the cursor is considered at the whitespace in the generated one (for some reason) instead of the brace. Might be better to change the assist to pick the token first, then check if the parent is a tt and that tt's parent is a macro call. (that is go up with the checks instead of down)

Veykril:
The assist id was wrong hence the doc failure

PR #17763, opened 2024-08-01
fix: Insert a tail `Ok(())` for expr block instead of wrapping with `Ok`

### LINKED ISSUE #17728: Quick fix for function no return value, `Ok()` out of function body
**rust-analyzer version**: rust-analyzer 1.80.0 (05147895 2024-07-21)

**editor or extension**: Vim 9.1 coc-rust-analyzer

```rust
type Result<T> = std::result::Result<T, ()>;

fn foo() -> Result<()> {}

fn main() { }
```
**quick fix**:
```rust
type Result<T> = std::result::Result<T, ()>;

fn foo() -> Result<()> Ok({})

fn main() { }
```
**expect**:
```rust
type Result<T> = std::result::Result<T, ()>;

fn foo() -> Result<()> { Ok(()) }

fn main() { }
```

ShoyuVanilla:
Fixes #17728

When type mismatch is `Result<(), E>, ()` or `Option<()>, ()` and target expr is a block expression, it is more reasonable to insert a wrapped unit - `Ok(())` or `Some(())` - as the tail expression of that block than wrapping the entire block with `Ok` or `Some`

### DIFF
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -2,8 +2,12 @@ use either::Either;
 use hir::{db::ExpandDatabase, ClosureStyle, HirDisplay, HirFileIdExt, InFile, Type};
 use ide_db::{famous_defs::FamousDefs, source_change::SourceChange};
 use syntax::{
-    ast::{self, BlockExpr, ExprStmt},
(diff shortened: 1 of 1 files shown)


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17775, opened 2024-08-02
perf: Segregate syntax and semantic diagnostics

### LINKED ISSUE #17731: Publish diagnostics in two rounds
Our semantic diagnostics can take a while to compute depending on the project and file size, which unfortunately means that syntax errors are not very responsive right now, as we compute them together with the semantic ones and only publish them once those are done computing. We should handle these two separately, allowing for snappy syntax error reporting.

ShoyuVanilla:
Closes #17731

### DIFF
--- a/crates/ide-diagnostics/src/lib.rs
+++ b/crates/ide-diagnostics/src/lib.rs
@@ -96,6 +96,7 @@ use syntax::{
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
 pub enum DiagnosticCode {
     RustcHardError(&'static str),
+    SyntaxError,
     RustcLint(&'static str),
     Clippy(&'static str),
     Ra(&'static str, Severity),
@@ -106,6 +107,9 @@ impl DiagnosticCode {
         match self {
             DiagnosticCode::RustcHardError(e) => {
                 format!("https://doc.rust-lang.org/stable/error_codes/{e}.html")
+            }
+            DiagnosticCode::SyntaxError => {
+                String::from("https://doc.rust-lang.org/stable/reference/")
             }
             DiagnosticCode::RustcLint(e) => {
                 format!("https://doc.rust-lang.org/rustc/?search={e}")
@@ -125,6 +129,7 @@ impl DiagnosticCode {
             | DiagnosticCode::RustcLint(r)
             | DiagnosticCode::Clippy(r)
             | DiagnosticCode::Ra(r, _) => r,
+            DiagnosticCode::SyntaxError => "syntax-error",
         }
     }
 }
@@ -154,7 +159,7 @@ impl Diagnostic {
             message,
             range: range.into(),
             severity: match code {
-                DiagnosticCode::RustcHardError(_) => Severity::Error,
+                DiagnosticCode::RustcHardError(_) | DiagnosticCode::SyntaxError => Severity::Error,
                 // FIXME: Rustc lints are not always warning, but the ones that are currently implemented are all warnings.
                 DiagnosticCode::RustcLint(_) => Severity::Warning,
                 // FIXME: We can make this configurable, and if the user uses `cargo clippy` on flycheck, we can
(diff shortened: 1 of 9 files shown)


crates/ide-diagnostics/src/lib.rs
```diff
@@ -297,31 +302,54 @@ impl DiagnosticsContext<'_> {
...
+    if config.disabled.contains("syntax-error") {
+        return Vec::new();
+    }
+
+    let sema = Semantics::new(db);
+    let file_id = sema
+        .attach_first_edition(file_id)
+        .unwrap_or_else(|| EditionedFileId::current_edition(file_id));
+
+    // [#3434] Only take first 128 errors to prevent slowing down editor/ide, the number 128 is chosen arbitrarily.
```
ShoyuVanilla, line 322: Previous comment was pointing the wrong issue no.34344

crates/ide-diagnostics/src/lib.rs
```diff
@@ -297,31 +302,54 @@ impl DiagnosticsContext<'_> {
...
+                FileRange { file_id: file_id.into(), range: err.range() },
+            )
+        })
+        .collect()
+}
+
+/// Request semantic diagnostics for the given [`FileId`]. The produced diagnostics may point to other files
 /// due to macros.
-pub fn diagnostics(
+pub fn semantic_diagnostics(
```
ShoyuVanilla, line 340: Calling `db.parse`s once in `fn syntax_diagnotics(..)` and again in this function that previously only called once in `fn diagnotics(..)` is a bit reluctant, but I think that it won't harm perf much due to lru cache here;
https://github.com/rust-lang/rust-analyzer/blob/aa00ddcf654a35ba0eafe17247cf189958d33182/crates/base-db/src/lib.rs#L60-L63

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -549,14 +555,35 @@ impl GlobalState {
             }
             // Diagnostics are triggered by the user typing
             // so we run them on a latency sensitive thread.
+            let slice2 = slice.clone();
             self.task_pool.handle.spawn(ThreadIntent::LatencySensitive, {
                 let snapshot = self.snapshot();
                 let subscriptions = subscriptions.clone();
```
Veykril, line 561:
```suggestion
            self.task_pool.handle.spawn(ThreadIntent::LatencySensitive, {
                let snapshot = self.snapshot();
                let subscriptions = subscriptions.clone();
            	let slice = slice.clone();
```
no need to come up with a new name then

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -549,14 +555,35 @@ impl GlobalState {
...
+                        fetch_native_diagnostics(
+                            snapshot,
+                            subscriptions,
+                            slice,
+                            NativeDiagnosticsFetchKind::Syntax,
+                        ),
+                    ))
+                }
+            });
+            self.task_pool.handle.spawn(ThreadIntent::LatencySensitive, {
```
Veykril, line 574: This duplicates the amount of tasks we do now, it also spawns semantic and parser diagnostics tasks in lockstep which means we might push queue up some syntax diagnostic tasks after semantic ones. I think a better way to model this might be having the `DiagnosticsTaskKind::Syntax` spawn the corresponding semantic one once its finished computing. That way we make sure to first do the syntax ones.

ShoyuVanilla: > it also spawns semantic and parser diagnostics tasks in lockstep which means we might push queue up some syntax diagnostic tasks after semantic ones.

That's terrible I couldn't think about it 😅 
I think that with `TaskPool::spawn_with_sender`, I can publish two tasks in order within a single job

ShoyuVanilla:
Since the syntax and the semantic diagnostics are always called in serial order, I think that it might be better to rollback the segregation
 of original `diagnostics` function and just add a callback as a parameter of it to get result of syntax diagnostics before semantic ones 🤔

Veykril (approved):

Veykril:
Thanks!
@bors r+

Veykril:
This seems to have broken diagnostics 😅

crates/rust-analyzer/src/diagnostics.rs
```diff
@@ -88,43 +90,51 @@ impl DiagnosticCollection {
...
+            diagnostics.sort_by_key(|it| (it.range.start, it.range.end));
+
+            if let Some((old_gen, existing_diagnostics)) = target.get_mut(&file_id) {
+                if existing_diagnostics.len() == diagnostics.len()
+                    && iter_eq_by(&diagnostics, &*existing_diagnostics, |new, existing| {
+                        are_diagnostics_equal(new, existing)
+                    })
+                {
+                    // don't signal an update if the diagnostics are the same
+                    return;
```
Veykril, line 113: This should be a `continue` now

ShoyuVanilla: Oh, yes, indeed 🥲 terrible mistake. I'll fix this one or two hours later once I got home

Veykril:
Alright this works great (aside from that one small issue 😄), syntax errors in `hir/lib.rs` are instantaneous now (where as they took a good second before)

PR #17791, opened 2024-08-04
feat: Implement diagnostic for `await` outside of `async`

### LINKED ISSUE #17781: Diagnose `await` outside `async` context
We aren't reporting errors for this yet even though its a fairly simple pass to do within body lowering.

ShoyuVanilla:
Closes #17781

### DIFF
--- a/crates/hir-def/src/body/lower.rs
+++ b/crates/hir-def/src/body/lower.rs
@@ -72,6 +72,7 @@ pub(super) fn lower(
         is_lowering_coroutine: false,
         label_ribs: Vec::new(),
         current_binding_owner: None,
+        awaitable_context: None,
     }
     .collect(params, body, is_async_fn)
 }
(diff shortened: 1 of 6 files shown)


crates/hir-def/src/body/lower.rs
```diff
@@ -100,6 +101,8 @@ struct ExprCollector<'a> {
     // resolution
     label_ribs: Vec<LabelRib>,
     current_binding_owner: Option<ExprId>,
+
+    awaitable_blocks: Vec<Awaitable>,
```
Veykril, line 105: Hmm, an `Option<Awaitable>` should suffice here, which would be replaced with `coroutine_kind` after https://github.com/rust-lang/rust-analyzer/pull/16173. In fact you can probably take a peek at how that PR sets `coroutine_kind`. Mihgt be simple enough to copy if the changes aren't drastic in there.

Then we just need to swap whenever we enter and exit a context like we do with other `Option`/`bool` things here

Veykril:
Thanks!
@bors r+

PR #17793, opened 2024-08-05
docs: add msvc note to manual

### LINKED ISSUE #4870: Add a note related to usage in Windows environment
I have a Windows 10 Enterprise 1909 version. I installed the latest VS Code, Rust and rust-analyzer plugins. I also have the latest rust gnu and msvc toolchains. However, I have an older VC++ compiler (installed with VS 2013).

After several installation/uninstallation attempts, I just couldn't get the rust-analyzer plugin working and I kept getting this - `The Rust Language Service server crashed 5 times in the last 3 minutes. The server will not be restarted`. I did not know where to look for the logs. So I downloaded the latest `rust-analyzer-windows.exe` from the releases and just double clicked to see if it ran. Unfortunately, I got this error - `The application was unable to start correctly (0xc00007b)` and this is why the rust-analyzer server was crashing.

A quick googling for `0xc00007b` hinted at installing the latest MSVC++ Redistributable package. I had an older version of this installed from the VS 2013 installation. 

After installing the latest redistributable package for VS 2015, 2017 and 2019 (both x86 and x64 versions to be on a safer side), the problem got resolved and rust-analyzer is now working fine.

This should help those who are trying to use rust-analyzer on Windows with an older/missing MSVC++ runtime and can't figure out what on earth is going on.

jjoeldaniel:
Added note for Windows users to have the latest MSVC to minimize setup issues.

Closes #4870

### DIFF
--- a/docs/user/manual.adoc
+++ b/docs/user/manual.adoc
@@ -203,6 +203,12 @@ The `rust-analyzer` binary can be installed via https://brew.sh/[Homebrew].
 $ brew install rust-analyzer
 ----
 
+==== Windows
+
+It is recommended to install the latest Microsoft Visual C++ Redistributable prior to installation.
+Download links can be found
+https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist[here].
+
 === VS Code or VSCodium in Flatpak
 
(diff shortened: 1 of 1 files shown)


Veykril:
Thanks!
@bors r+

PR #17794, opened 2024-08-05
internal: Newtype `ErasedFileAstId`

Veykril:
It wrapping `la_arena::Idx` makes it quite annoying to use

### DIFF
--- a/crates/project-model/src/workspace.rs
+++ b/crates/project-model/src/workspace.rs
@@ -1349,12 +1349,13 @@ fn add_target_crate_root(
     );
     if let TargetKind::Lib { is_proc_macro: true } = kind {
         let proc_macro = match build_data.as_ref().map(|it| it.proc_macro_dylib_path.as_ref()) {
-            Some(it) => it.cloned().map(|path| Ok((cargo_name.to_owned(), path))),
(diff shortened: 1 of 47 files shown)


crates/project-model/src/workspace.rs
```diff
@@ -1349,12 +1349,13 @@ fn add_target_crate_root(
     );
     if let TargetKind::Lib { is_proc_macro: true } = kind {
         let proc_macro = match build_data.as_ref().map(|it| it.proc_macro_dylib_path.as_ref()) {
-            Some(it) => it.cloned().map(|path| Ok((cargo_name.to_owned(), path))),
-            None => Some(Err("proc-macro crate is missing its build data".to_owned())),
+            Some(it) => match it {
+                Some(path) => Ok((cargo_name.to_owned(), path.clone())),
+                None => Err("proc-macro crate build data is missing dylib path".to_owned()),
```
bugadani, line 1354: What may cause me hitting this error on most proc macros?

Veykril: Unclear, somethings going wrong when cargo is reporting the build artifact outputs then (or your build script command is configured to something else that misbehaves)

PR #17795, opened 2024-08-05
feat: Load sysroot library via cargo metadata

Veykril:
See https://github.com/rust-lang/rust/pull/128534, fixes https://github.com/rust-lang/rust-analyzer/issues/7637

Requires a toolchain from 176e54520 2024-08-04 or later to work.

### DIFF
--- a/crates/project-model/src/sysroot.rs
+++ b/crates/project-model/src/sysroot.rs
@@ -123,32 +123,27 @@ impl Sysroot {
 // FIXME: Expose a builder api as loading the sysroot got way too modular and complicated.
 impl Sysroot {
     /// Attempts to discover the toolchain's sysroot from the given `dir`.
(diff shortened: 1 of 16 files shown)


crates/project-model/src/sysroot.rs
```diff
@@ -425,6 +311,92 @@ impl Sysroot {
...
+            &|_| (),
+        ) {
+            Ok(it) => it,
+            Err(e) => {
+                tracing::error!("`cargo metadata` failed on `{library_manifest}` : {e}");
+                return None;
+            }
+        };
+
+        // Patch out `rustc-std-workspace-*` crates to point to the real crates.
```
Veykril, line 344: I don't think we need to do this anymore, though it does reduce noise in the crate graph

PR #17809, opened 2024-08-06
Include vendored crates in StaticIndex

nicolas-guichard:
`StaticIndex::compute` filters out modules from libraries. This makes an exceptions for vendored libraries, ie libraries actually defined inside the workspace being indexed.

This aims to solve https://bugzilla.mozilla.org/show_bug.cgi?id=1846041 In general StaticIndex is meant for code browsers, which likely want to index all visible source files.

### DIFF
--- a/crates/ide/src/static_index.rs
+++ b/crates/ide/src/static_index.rs
@@ -3,9 +3,12 @@
 
 use hir::{db::HirDatabase, Crate, HirFileIdExt, Module, Semantics};
 use ide_db::{
-    base_db::SourceRootDatabase, defs::Definition, documentation::Documentation,
-    famous_defs::FamousDefs, helpers::get_definition, FileId, FileRange, FxHashMap, FxHashSet,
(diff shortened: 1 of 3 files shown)


lnicola:
Why aren't we picking them up right now? Is it because of the `Cargo.toml` [exclusion](https://searchfox.org/mozilla-central/rev/8b0666aff1197e1dd8017de366343de9c21ee437/Cargo.toml#28), or something else?

lnicola (approved):

Veykril:
Hmm, now I wonder is it somehow possible for us to detect (and mark) vendored dependencies? The approach in this PR works but feels a bit odd

Veykril:
I see, vendoring really is just overwriting the source to a directory. So we could figure this out by using cargo config actually to check if a source is a directory source to tell whether something is vendored
I'll merge this for now anyways as it gets the job done
@bors r+

PR #17814, opened 2024-08-06
feat: Implement object safety and its hovering hint

### LINKED ISSUE #17779: Hovering a trait should tell whether it is object safe or not

ShoyuVanilla:
Resolves #17779

- [x] Fill missing implementations
- [x] Hover rendering
- [x] Implement object safety's own test suite, like layout
- [x] Add test cases (from rustc maybe)
- [x] Clean up ugly codes
- [x] Add doc string

### DIFF
--- a/crates/hir-ty/src/db.rs
+++ b/crates/hir-ty/src/db.rs
@@ -11,7 +11,7 @@ use base_db::{
 use hir_def::{
     db::DefDatabase, hir::ExprId, layout::TargetDataLayout, AdtId, BlockId, CallableDefId,
     ConstParamId, DefWithBodyId, EnumVariantId, FunctionId, GeneralConstId, GenericDefId, ImplId,
-    LifetimeParamId, LocalFieldId, StaticId, TypeAliasId, TypeOrConstParamId, VariantId,
+    LifetimeParamId, LocalFieldId, StaticId, TraitId, TypeAliasId, TypeOrConstParamId, VariantId,
 };
 use la_arena::ArenaMap;
 use smallvec::SmallVec;
@@ -24,6 +24,7 @@ use crate::{
     lower::{GenericDefaults, GenericPredicates},
     method_resolution::{InherentImpls, TraitImpls, TyFingerprint},
     mir::{BorrowckResult, MirBody, MirLowerError},
+    object_safety::ObjectSafetyViolation,
     Binders, ClosureId, Const, FnDefId, ImplTraitId, ImplTraits, InferenceResult, Interner,
     PolyFnSig, Substitution, TraitEnvironment, TraitRef, Ty, TyDefId, ValueTyDefId,
 };
@@ -106,6 +107,9 @@ pub trait HirDatabase: DefDatabase + Upcast<dyn DefDatabase> {
 
     #[salsa::invoke(crate::layout::target_data_layout_query)]
     fn target_data_layout(&self, krate: CrateId) -> Result<Arc<TargetDataLayout>, Arc<str>>;
+
+    #[salsa::invoke(crate::object_safety::object_safety_of_trait_query)]
+    fn object_safety_of_trait(&self, trait_: TraitId) -> Option<ObjectSafetyViolation>;
 
     #[salsa::invoke(crate::lower::ty_query)]
     #[salsa::cycle(crate::lower::ty_recover)]
--- a/crates/hir-ty/src/object_safety.rs
+++ b/crates/hir-ty/src/object_safety.rs
@@ -1,0 +1,631 @@
+//! Compute the object-safety of a trait
+
+use std::ops::ControlFlow;
+
+use chalk_ir::{
+    cast::Cast,
+    visit::{TypeSuperVisitable, TypeVisitable, TypeVisitor},
+    DebruijnIndex,
+};
+use chalk_solve::rust_ir::InlineBound;
+use hir_def::{
+    lang_item::LangItem, AssocItemId, ConstId, FunctionId, GenericDefId, HasModule, TraitId,
+    TypeAliasId,
+};
+use rustc_hash::{FxHashMap, FxHashSet};
+use smallvec::SmallVec;
+
+use crate::{
+    all_super_traits,
+    db::HirDatabase,
+    from_assoc_type_id, from_chalk_trait_id,
+    generics::{generics, trait_self_param_idx},
+    lower::callable_item_sig,
+    to_assoc_type_id, to_chalk_trait_id,
+    utils::elaborate_clause_supertraits,
+    AliasEq, AliasTy, Binders, BoundVar, CallableSig, GoalData, ImplTraitId, Interner, OpaqueTyId,
+    ProjectionTyExt, Solution, Substitution, TraitRef, Ty, TyKind, WhereClause,
+};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum ObjectSafetyViolation {
+    SizedSelf,
+    SelfReferential,
+    Method(FunctionId, MethodViolationCode),
+    AssocConst(ConstId),
+    GAT(TypeAliasId),
+    // This doesn't exist in rustc, but added for better visualization
+    HasNonSafeSuperTrait(TraitId),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum MethodViolationCode {
+    StaticMethod,
+    ReferencesSelfInput,
+    ReferencesSelfOutput,
+    ReferencesImplTraitInTrait,
+    AsyncFn,
+    WhereClauseReferencesSelf,
+    Generic,
+    UndispatchableReceiver,
+}
+
+pub fn object_safety(db: &dyn HirDatabase, trait_: TraitId) -> Option<ObjectSafetyViolation> {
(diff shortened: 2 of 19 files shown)


ShoyuVanilla:
It might take some time because I'm having difficulties re-implementing some functionality in rustc into r-a 😅 (I think I would do it eventually)

ShoyuVanilla:
I think that basic implementation is complete. Now it's test time

crates/test-utils/src/minicore.rs
```diff
@@ -822,6 +823,29 @@ pub mod ops {
     }
     pub use self::coroutine::{Coroutine, CoroutineState};
     // endregion:coroutine
+
+    // region:dispatch_from_dyn
+    mod dispatch_from_dyn {
+        use crate::marker::Unsize;
+
+        #[unstable(feature = "dispatch_from_dyn", issue = "none")]
```
Veykril, line 831: fwiw we haven't recorded stability attributes in minicore so far so no need to do that here

crates/hir-ty/src/object_safety.rs
```diff
@@ -0,0 +1,522 @@
...
+    WhereClauseReferencesSelf,
+    Generic,
+    UndispatchableReceiver,
+}
+
+// Basically, this is almost same as `rustc_trait_selection::traits::object_safety`
+// but some difference;
+//
+// 1. While rustc gathers almost every violation, but this only early return on
+//    first violation for perf.
```
Veykril, line 56: Might be nice to make this take a (generic) callback:
```rs
pub fn object_safety_of_trait_query<T>(
    db: &dyn HirDatabase,
    trait_: TraitId,
    cb: &mut dyn FnMut(ObjectSafetyViolation) -> ControlFlow<()>
) {
```
even if we don't yet need to collect all reasons

ShoyuVanilla:
While testing this plugged into here;
https://github.com/rust-lang/rust-analyzer/blob/7106cd3be50b2a43c1d9f2787bf22d4369c2b25b/crates/hir-ty/src/chalk_db.rs#L384-L387
I've found some minor mistakes in my implementation and fixes for them 😅 (All the test are green in my local revision with the above things applied)
I'll push them some hours later

crates/hir-ty/src/object_safety.rs
ShoyuVanilla: Well, callback strategy sounds great, but I think that it can't be used as a salsa query 🤔

Veykril: Oh no it can't, mainly meant we can already architecture it that way, for the query we'd keep the single one that just calls into this with a `ControlFlow::Break` on first issue found. Anything that wants all of them can bypass the query if the query reports there is at least on violation.

ShoyuVanilla: Oh, I see. I couldn't think of adding just one more "wrapper" function for query 😅

crates/test-utils/src/minicore.rs
```diff
@@ -1308,7 +1333,10 @@ pub mod iter {
                     self
                 }
                 // region:iterators
-                fn take(self, n: usize) -> crate::iter::Take<Self> {
+                fn take(self, n: usize) -> crate::iter::Take<Self>
+                where
+                    Self: Sized,
+                {
```
ShoyuVanilla, line 1339: `std` has this bound, too

crates/ide/src/hover/render.rs
```diff
@@ -964,3 +980,75 @@ fn keyword_hints(
         _ => KeywordHint::new(token.text().to_owned(), format!("{}_keyword", token.text())),
     }
 }
+
+fn render_object_safety(
```
ShoyuVanilla, line 984: I'm worried that this might be too verbose or not pretty 😅

crates/hir-ty/src/object_safety/tests.rs
ShoyuVanilla, line 1: I brought every relevant test from https://github.com/rust-lang/rust/tree/ae9f5019f9ce6eb3ecd96206ade4a612efe20fd5/tests/ui/object-safety except one with HRTB - it's still checked as object unsafe as it is in rustc but reason is different because we don't support HRBT yet -

crates/ide/src/hover/render.rs
```diff
@@ -964,3 +980,75 @@ fn keyword_hints(
...
+    db: &RootDatabase,
+    name: Name,
+    safety: Option<ObjectSafetyViolation>,
+) -> SmallVec<[String; 3]> {
```
Veykril, line 988:
```suggestion
fn render_object_safety(
    db: &RootDatabase,
    buf: &mut String,
    name: Name,
    safety: Option<ObjectSafetyViolation>,
) {
```
and then just use `format_to!(buf, ...)` to write into that. Should be a bit easier to read

crates/ide/src/hover/render.rs
```diff
@@ -540,6 +549,13 @@ pub(super) fn definition(
         desc.push_str(" = ");
         desc.push_str(&value);
     }
+    if let Some(object_safety_info) = object_safety_info {
```
Veykril, line 552: We should print this above the trait

crates/hir-ty/src/db.rs
```diff
@@ -107,6 +108,12 @@ pub trait HirDatabase: DefDatabase + Upcast<dyn DefDatabase> {
     #[salsa::invoke(crate::layout::target_data_layout_query)]
     fn target_data_layout(&self, krate: CrateId) -> Result<Arc<TargetDataLayout>, Arc<str>>;
 
+    #[salsa::invoke(crate::object_safety::object_safety_query)]
+    fn object_safety(&self, trait_: TraitId) -> Option<ObjectSafetyViolation>;
+
+    #[salsa::invoke(crate::object_safety::object_safety_of_trait_query)]
+    fn object_safety_of_trait(&self, trait_: TraitId) -> Option<ObjectSafetyViolation>;
```
Veykril, line 115: Why do we have two of these?

ShoyuVanilla: I was trying to avoid supertrait traversals so I've made the first one, but I think that if that might be the bottle neck, making supertraits into query is more reasonable 😅 Anyway, the first one seems redundant

crates/ide/src/hover/render.rs
```diff
@@ -964,3 +977,62 @@ fn keyword_hints(
...
+        }
+        ObjectSafetyViolation::Method(func, mvc) => {
+            let name = hir::Function::from(func).name(db);
+            format_to!(
+                buf,
+                "has a method `{}` that is non dispatchable because of;\n//   - ",
+                name.as_str()
+            );
+            let desc = match mvc {
+                MethodViolationCode::StaticMethod => "static (no receiver parameter)",
```
Veykril, line 1006:
```suggestion
                MethodViolationCode::StaticMethod => "missing a receiver",
```
`static` isn't too descriptive I think (and might confuse people with a `static` item)

crates/ide/src/hover/render.rs
```diff
@@ -964,3 +977,62 @@ fn keyword_hints(
...
+            buf.push_str("has a `Self: Sized` bound");
+        }
+        ObjectSafetyViolation::SelfReferential => {
+            buf.push_str("has a bound that references `Self`");
+        }
+        ObjectSafetyViolation::Method(func, mvc) => {
+            let name = hir::Function::from(func).name(db);
+            format_to!(
+                buf,
+                "has a method `{}` that is non dispatchable because of;\n//   - ",
```
Veykril, line 1002:
```suggestion
                "has a method `{}` that is non dispatchable because of:\n//   - ",
```

crates/ide/src/hover/render.rs
```diff
@@ -964,3 +977,62 @@ fn keyword_hints(
...
+                name.as_str()
+            );
+            let desc = match mvc {
+                MethodViolationCode::StaticMethod => "static (no receiver parameter)",
+                MethodViolationCode::ReferencesSelfInput => "a parameter references `Self`",
+                MethodViolationCode::ReferencesSelfOutput => "the return type references `Self`",
+                MethodViolationCode::ReferencesImplTraitInTrait => {
+                    "the return type contains `impl Trait`"
+                }
+                MethodViolationCode::AsyncFn => "async",
```
Veykril, line 1012:
```suggestion
                MethodViolationCode::AsyncFn => "being async",
```

crates/ide/src/hover/render.rs
```diff
@@ -964,3 +977,62 @@ fn keyword_hints(
...
+                MethodViolationCode::WhereClauseReferencesSelf => {
+                    "a where clause references `Self`"
+                }
+                MethodViolationCode::Generic => "a Generic parameter other than lifetime",
```
Veykril, line 1016:
```suggestion
                MethodViolationCode::Generic => "a non-lifetime generic parameter",
```

crates/ide/src/hover/render.rs
```diff
@@ -964,3 +977,62 @@ fn keyword_hints(
...
+                MethodViolationCode::UndispatchableReceiver => "the non dispatchable receiver type",
```
Veykril, line 1017:
```suggestion
                MethodViolationCode::UndispatchableReceiver => "a non-dispatchable receiver type",
```

Veykril (approved):

ShoyuVanilla:
This caused some regressions only in `diesel`. I'll find out why 🤔

PR #17832, opened 2024-08-08
fix: Panic while rendering function type hint with impl trait arg

### LINKED ISSUE #17811: Building hover information for fn pointers from generic with impl args panics.
While trying out https://github.com/rust-lang/rust-analyzer/pull/17809 on mozilla-central, I noticed a failure in the time crate when building the hover information for the `format_number` function pointer [here](https://github.com/time-rs/time/blob/866c3a0c795445b15bfcb18cb7f005827de0db1c/time/src/formatting/mod.rs#L309).

I managed to write a failing test, see b93994b1356c1264dc6cc197496de9b2ddaddef4.

The issue is we end up in `hir_ty::display::generic_args_sans_defaults` with more binders than parameters. I'm not 100% sure how to fix this, I'm also surprised we don't hit this in more cases since `generic_args_sans_defaults` is called multiple times with different numbers of parameters at least [here](https://github.com/rust-lang/rust-analyzer/blob/b23142209ef632475915e858f5f20901ef7c12da/crates/hir-ty/src/display.rs#L1032) and [here](https://github.com/rust-lang/rust-analyzer/blob/b23142209ef632475915e858f5f20901ef7c12da/crates/hir-ty/src/display.rs#L1034).

**rust-analyzer version**: current `HEAD`

**editor or extension**: Kate / SCIP generator

**relevant settings**: 

**repository link (if public, optional)**: initially found with in [time](https://github.com/time-rs/time/blob/866c3a0c795445b15bfcb18cb7f005827de0db1c/time/src/formatting/mod.rs#L309), minimal test at b93994b1356c1264dc6cc197496de9b2ddaddef4

```rust
fn bar<const WIDTH: u8>(_: impl Copy) {}
fn test() {
    let f = bar::<3>;
    f(4);
}
```
hovering over `f` panics.

ShoyuVanilla:
Fixes #17811

### DIFF
--- a/crates/hir-ty/src/display.rs
+++ b/crates/hir-ty/src/display.rs
@@ -1022,16 +1022,16 @@ impl HirDisplay for Ty {
                     // We print all params except implicit impl Trait params. Still a bit weird; should we leave out parent and self?
                     if parameters.len() - impl_ > 0 {
                         // `parameters` are in the order of fn's params (including impl traits), fn's lifetimes
+                        let without_impl = self_param as usize + type_ + const_ + lifetime;
                         // parent's params (those from enclosing impl or trait, if any).
-                        let (fn_params, other) =
-                            parameters.split_at(self_param as usize + type_ + const_ + lifetime);
-                        let (_impl, parent_params) = other.split_at(impl_);
+                        let (fn_params, parent_params) = parameters.split_at(without_impl + impl_);
                         debug_assert_eq!(parent_params.len(), parent_len);
 
                         let parent_params =
                             generic_args_sans_defaults(f, Some(generic_def_id), parent_params);
                         let fn_params =
-                            generic_args_sans_defaults(f, Some(generic_def_id), fn_params);
+                            &generic_args_sans_defaults(f, Some(generic_def_id), fn_params)
+                                [0..without_impl];
 
                         write!(f, "<")?;
                         hir_fmt_generic_arguments(f, parent_params, None)?;
--- a/crates/ide/src/hover/tests.rs
+++ b/crates/ide/src/hover/tests.rs
@@ -8579,3 +8579,26 @@ fn main(a$0: T) {}
         "#]],
     );
 }
+
+#[test]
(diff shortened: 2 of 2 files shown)


crates/hir-ty/src/display.rs
```diff
@@ -1022,16 +1022,16 @@ impl HirDisplay for Ty {
                     // We print all params except implicit impl Trait params. Still a bit weird; should we leave out parent and self?
                     if parameters.len() - impl_ > 0 {
                         // `parameters` are in the order of fn's params (including impl traits), fn's lifetimes
+                        let without_impl = self_param as usize + type_ + const_ + lifetime;
                         // parent's params (those from enclosing impl or trait, if any).
-                        let (fn_params, other) =
-                            parameters.split_at(self_param as usize + type_ + const_ + lifetime);
-                        let (_impl, parent_params) = other.split_at(impl_);
+                        let (fn_params, parent_params) = parameters.split_at(without_impl + impl_);
```
ShoyuVanilla, line 1027: as `generic_args` is `self + type params + const params + lifetimes + hidden generic with impl trait args`, we should include `impl` to generate `generic_args_sans_defaults`

crates/hir-ty/src/display.rs
```diff
@@ -1022,16 +1022,16 @@ impl HirDisplay for Ty {
...
                         debug_assert_eq!(parent_params.len(), parent_len);
 
                         let parent_params =
                             generic_args_sans_defaults(f, Some(generic_def_id), parent_params);
                         let fn_params =
-                            generic_args_sans_defaults(f, Some(generic_def_id), fn_params);
+                            &generic_args_sans_defaults(f, Some(generic_def_id), fn_params)
+                                [0..without_impl];
```
ShoyuVanilla, line 1034: And we exclude `impl trait` things here

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17833, opened 2024-08-08
Reuse recursion limit as expansion limit

edevil:
A configurable recursion limit was introduced by looking at the recursion_limit crate attribute. Instead of relying on a global constant we will reuse this value for expansion limit as well.

Addresses: https://github.com/rust-lang/rust-analyzer/issues/8640#issuecomment-2271740272


Veykril:
Thanks!
@bors r+

PR #17836, opened 2024-08-08
minor: log error when sysroot can't be discovered

### LINKED ISSUE #17808: No error logged when the sysroot can't be discovered
See https://github.com/rust-lang/rust-analyzer/issues/17759#issuecomment-2271044606. We report this in VS Code, but don't log it.

Run `rustup component remove rust-src` and rename `rustup` to test.

winstxnhdw:
Closes #17808


winstxnhdw:
I wonder if it would be better to print this error on the cursor..

Veykril (approved):

Veykril:
@bors r+

Good call, we might want to generate a general diagnostic in opened files rendered at the top line that the sysroot failed to load

PR #17842, opened 2024-08-09
internal: Optimize the usage of channel senders

mo8it:
Used `Sender` directly instead of a boxed closure. There is no need to use the boxed closure. This also allows the caller to decide to do something other than `unwrap` (not a fan of it BTW).

### DIFF
--- a/crates/vfs-notify/src/lib.rs
+++ b/crates/vfs-notify/src/lib.rs
@@ -119,14 +119,12 @@ impl NotifyActor {
                         self.watched_dir_entries.clear();
                         self.watched_file_entries.clear();
 
-                        self.sender
-                            .send(loader::Message::Progress {
-                                n_total,
-                                n_done: LoadingProgress::Started,
-                                config_version,
-                                dir: None,
-                            })
-                            .unwrap();
+                        self.send(loader::Message::Progress {
+                            n_total,
+                            n_done: LoadingProgress::Started,
+                            config_version,
+                            dir: None,
+                        });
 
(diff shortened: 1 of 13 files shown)


mo8it:
LOL, CI failed after 6 hours? This must be a Github issue, right? Cargo tests have a timeout.

lnicola:
Looks like it's a deadlock, maybe a channel that doesn't get closed in time?

crates/vfs-notify/src/lib.rs
```diff
@@ -119,18 +119,19 @@ impl NotifyActor {
...
+                                config_version,
+                                dir: None,
+                            })
+                            .unwrap();
 
                         let (entry_tx, entry_rx) = unbounded();
                         let (watch_tx, watch_rx) = unbounded();
                         let processed = AtomicUsize::new(0);
-                        config.load.into_par_iter().enumerate().for_each(move |(i, entry)| {
+                        config.load.into_par_iter().enumerate().for_each(|(i, entry)| {
```
Veykril, line 134: This is the cause for the deadlock, the `move` moved `entry_tx` and `watch_tx` before and then dropped them once the `for_each` is done. Now they stick around causing the `watch_rx` and `entry_rx` iteration afterwards to not exit as the corresponding senders are not yet dropped

mo8it: Thanks! Fixed :)

Veykril:
Thanks!
@bors r+

PR #17843, opened 2024-08-10
internal: Performance optimizations

mo8it:
- Use `Command::arg` directly
- Avoid the overhead of the `select!` macro when possible
- Use `select_biased!`

### DIFF
--- a/crates/rust-analyzer/src/flycheck.rs
+++ b/crates/rust-analyzer/src/flycheck.rs
@@ -3,7 +3,7 @@
 
 use std::{fmt, io, process::Command, time::Duration};
 
-use crossbeam_channel::{never, select, unbounded, Receiver, Sender};
+use crossbeam_channel::{select_biased, unbounded, Receiver, Sender};
 use paths::{AbsPath, AbsPathBuf, Utf8PathBuf};
 use rustc_hash::FxHashMap;
 use serde::Deserialize;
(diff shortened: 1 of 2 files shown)


crates/rust-analyzer/src/flycheck.rs
```diff
@@ -419,7 +420,8 @@ impl FlycheckActor {
                 cmd.arg("--keep-going");
 
                 options.apply_on_command(&mut cmd);
-                (cmd, options.extra_args.clone())
+                cmd.args(&options.extra_args);
```
mo8it, line 423: It is weird that `extra_args` aren't applied in `apply_on_command`

crates/vfs-notify/src/lib.rs
```diff
@@ -85,10 +85,13 @@ impl NotifyActor {
...
     fn next_event(&self, receiver: &Receiver<Message>) -> Option<Event> {
-        let watcher_receiver = self.watcher.as_ref().map(|(_, receiver)| receiver);
+        let Some((_, watcher_receiver)) = &self.watcher else {
+            return receiver.recv().ok().map(Event::Message);
+        };
+
         select! {
             recv(receiver) -> it => it.ok().map(Event::Message),
-            recv(watcher_receiver.unwrap_or(&never())) -> it => Some(Event::NotifyEvent(it.unwrap())),
+            recv(watcher_receiver) -> it => Some(Event::NotifyEvent(it.unwrap())),
```
mo8it, line 94: Why are we unwrapping here but not with other messages?

Veykril: Unsure, might've been due to drop order at some point to prevent a panic

mo8it: Well, not relevant for the PR. just noticed it.

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17853, opened 2024-08-11
feat: `min-exhaustive-patterns`

### LINKED ISSUE #17851: Support `min_exhaustive_patterns`
https://github.com/rust-lang/rust/issues/119612 has been stabilized but we throw exhaustiveness errors for those examples
```rust

let x: Result<i32, !>;
match x {
    Ok(y) => (),
}
let Ok(y) = x;
```

ShoyuVanilla:
Resolves #17851

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,11 +84,11 @@ tt = { path = "./crates/tt", version = "0.0.0" }
 vfs-notify = { path = "./crates/vfs-notify", version = "0.0.0" }
 vfs = { path = "./crates/vfs", version = "0.0.0" }
 
-ra-ap-rustc_lexer = { version = "0.53.0", default-features = false }
-ra-ap-rustc_parse_format = { version = "0.53.0", default-features = false }
-ra-ap-rustc_index = { version = "0.53.0", default-features = false }
-ra-ap-rustc_abi = { version = "0.53.0", default-features = false }
-ra-ap-rustc_pattern_analysis = { version = "0.53.0", default-features = false }
+ra-ap-rustc_lexer = { version = "0.63.0", default-features = false }
+ra-ap-rustc_parse_format = { version = "0.63.0", default-features = false }
+ra-ap-rustc_index = { version = "0.63.0", default-features = false }
+ra-ap-rustc_abi = { version = "0.63.0", default-features = false }
+ra-ap-rustc_pattern_analysis = { version = "0.63.0", default-features = false }
 
 # local crates that aren't published to crates.io. These should not have versions.
 test-fixture = { path = "./crates/test-fixture" }
@@ -125,11 +125,11 @@ memmap2 = "0.5.4"
 nohash-hasher = "0.2.0"
 oorandom = "11.1.3"
 object = { version = "0.33.0", default-features = false, features = [
-    "std",
-    "read_core",
-    "elf",
-    "macho",
-    "pe",
+  "std",
+  "read_core",
+  "elf",
+  "macho",
+  "pe",
 ] }
 process-wrap = { version = "8.0.2", features = ["std"] }
 pulldown-cmark-to-cmark = "10.0.4"
@@ -157,7 +157,6 @@ tracing-subscriber = { version = "0.3.18", default-features = false, features = 
 triomphe = { version = "0.1.10", default-features = false, features = ["std"] }
 url = "2.3.1"
 xshell = "0.2.5"
-
 
 
 # We need to freeze the version of the crate, as the raw-api feature is considered unstable
--- a/crates/hir-ty/src/diagnostics/expr.rs
+++ b/crates/hir-ty/src/diagnostics/expr.rs
@@ -4,20 +4,25 @@
 
 use std::fmt;
 
+use chalk_solve::rust_ir::AdtKind;
 use either::Either;
-use hir_def::lang_item::LangItem;
-use hir_def::{resolver::HasResolver, AdtId, AssocItemId, DefWithBodyId, HasModule};
-use hir_def::{ItemContainerId, Lookup};
+use hir_def::{
+    lang_item::LangItem,
+    resolver::{HasResolver, ValueNs},
+    AdtId, AssocItemId, DefWithBodyId, HasModule, ItemContainerId, Lookup,
+};
 use intern::sym;
 use itertools::Itertools;
 use rustc_hash::FxHashSet;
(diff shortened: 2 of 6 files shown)


crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
```diff
@@ -328,7 +325,7 @@ impl<'db> PatCx for MatchCheckCtx<'db> {
         self.exhaustive_patterns
     }
     fn is_min_exhaustive_patterns_feature_on(&self) -> bool {
-        self.min_exhaustive_patterns
+        true
```
ShoyuVanilla, line 328: Should I bump up `ra-ap-rustc_pattern_analysis` to latest instead?

crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -253,6 +263,45 @@ impl ExprValidator {
...
+                }
+                _ => self.is_known_valid_scrutinee(*expr, db),
+            },
+            Expr::Index { base, .. } => self.is_known_valid_scrutinee(*base, db),
+            Expr::Cast { expr, .. } => self.is_known_valid_scrutinee(*expr, db),
+            Expr::Missing => false,
+            _ => true,
+        }
+    }
+
```
ShoyuVanilla, line 304: This dance is the compressed version of
https://github.com/rust-lang/rust/blob/c9bd03cb724e13cca96ad320733046cbdb16fbbe/compiler/rustc_mir_build/src/thir/cx/expr.rs
and then
https://github.com/rust-lang/rust/blob/master/compiler/rustc_mir_build/src/thir/pattern/check_match.rs#L288

crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
```diff
@@ -95,8 +93,7 @@ impl<'db> MatchCheckCtx<'db> {
             }
         }
 
-        // FIXME: Determine place validity correctly. For now, err on the safe side.
-        let place_validity = PlaceValidity::MaybeInvalid;
+        let place_validity = PlaceValidity::from_bool(known_valid_scrutinee.unwrap_or(true));
```
ShoyuVanilla, line 96: This `unwrap_or` thing is from rustc;
https://github.com/rust-lang/rust/blob/c9bd03cb724e13cca96ad320733046cbdb16fbbe/compiler/rustc_mir_build/src/thir/pattern/check_match.rs#L372-L373

crates/ide-diagnostics/src/handlers/non_exhaustive_let.rs
```diff
@@ -41,6 +41,27 @@ fn main() {
...
+    #[test]
+    fn min_exhaustive() {
+        check_diagnostics(
+            r#"
+//- minicore: result
+fn test(x: Result<i32, !>) {
+    let Ok(_y) = x;
+}
+"#,
+        );
```
ShoyuVanilla, line 57: The reason that this test doesn't have a pointer deref case is because the following code;
```rust
fn test(ptr: *const Result<i32, !>) {
    unsafe {
        let Ok(_x) = *ptr;
    }
}
```
is getting a block with no stmts but tail one in here,
https://github.com/rust-lang/rust-analyzer/blob/0daeb5c0b05cfdf2101b0f078c27539099bf38e6/crates/hir-ty/src/diagnostics/expr.rs#L256-L257
while the following is getting a block with a single stmt without tail 🤔 
```rust
fn test(x: Result<i32, &'static !>) {
    let Ok(_y) = x;
}
```
I'll make a more deep inspection and issue this as a new issue

crates/hir-ty/src/diagnostics/match_check/pat_analysis.rs
Veykril: There is a PR upstream enabling the in-tree pattern analysis in the rust repo (to prevent us from going out of sync), so these two PRs will likely collide. Might be best if either PR waits on the other (and if we pick this one here first then yes we should bump the dependency here), cc @Nadrieril

ShoyuVanilla: Since the rust-lang/rust/#128922 is in homu queue, maybe I should wait for it?

Veykril:
No that pr got removed from the queue as it doesn't build, so I guess if
you wanna bump our crates io version first we can swap to in tree in the
rust repo as a follow up after we've synced this back

ShoyuVanilla:
In either case, I'll continue on this about 20 hours later when I get back from work because it's midnight here 😅

Veykril:
No rush

crates/ide-diagnostics/src/handlers/non_exhaustive_let.rs
Veykril: > is getting a block with no stmts but tail one in here(thus, no diagnostic error),

oh no that sounds very wrong

ShoyuVanilla: Yes, it looks very wrong 🤔 I'll check the entire lowering process from AST to hir once I get back home

ShoyuVanilla: This is because #17865

Veykril:
Eh, the autopublish action got frozen by github again due to repo inactivity

Veykril:
Kicked off a new one https://github.com/rust-analyzer/rustc-auto-publish/actions/workflows/main.yml

crates/ide-diagnostics/src/handlers/non_exhaustive_let.rs
ShoyuVanilla: I got into this 😂 rust-lang/rust#129009

Cargo.toml
```diff
@@ -88,7 +88,7 @@ ra-ap-rustc_lexer = { version = "0.53.0", default-features = false }
 ra-ap-rustc_parse_format = { version = "0.53.0", default-features = false }
 ra-ap-rustc_index = { version = "0.53.0", default-features = false }
 ra-ap-rustc_abi = { version = "0.53.0", default-features = false }
```
Veykril, line 90: Do bump the other three as well otherwise we get some duplicate transitive dependencies

Veykril (approved):

Nadrieril:
Just had time to take a look, looks good to me! Thx for taking the time to do this

Issue #17390, opened 2024-06-11
Files included using `include!()` macro in `lib.rs` are still `unlinked-file`


NessajCN:
**rust-analyzer version**: (eg. output of "rust-analyzer: Show RA Version" command, accessible in VSCode via <kbd>Ctrl/⌘</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd>)

rust-analyzer version: 0.4.1994-standalone [/home/nessaj/.vscode/extensions/rust-lang.rust-analyzer-0.4.1994-linux-x64/server/rust-analyzer]

**rustc version**: (eg. output of `rustc -V`)
rustc 1.78.0 (9b00956e5 2024-04-29)

**editor or extension**: (eg. VSCode, Vim, Emacs, etc. For VSCode users, specify your extension version; for users of other editors, provide the distribution if applicable)

VSCode 1.90.0 rust-analyzer  v0.4.1994 (pre-release)

**relevant settings**: (eg. client settings, or environment variables like `CARGO`, `RUSTC`, `RUSTUP_HOME` or `CARGO_HOME`)

**repository link (if public, optional)**: (eg. [rust-analyzer](https://github.com/rust-lang/rust-analyzer))

**code snippet to reproduce**: 
```rust title="src/lib.rs"
// src/lib.rs
include!("xx.rs");
```
```rust title="src/xx.rs"
// src/xx.rs
// Statements defined here all indicates: file not included in module tree rust-analyzer (unlinked-file)
// which worked perfect in previous versions.
pub const YY: u8 = 0xff;
```

Having tested in both Linux (Archlinux+KDE) and MacOS. Reproduced.

Veykril:
Hmm right, this file is in fact not a module, so when we try to resolve it as one we fail emitting the diagnostic. `Semantics::file_to_module_def` should probably have a fallback check for files that don't resolve to one and then check if the file is included treating the callee module of the `include!` as the module maybe.

JindrichSima:
Any progress?

NessajCN:
Hey @Veykril
I am not sure if I should open a new issue or reopen this one.
After your fix has been updated I found r-a did not recognize the following attributes correctly in `lib.rs` that using `include!()` macro:
```rust
// src/lib.rs
#![allow(non_upper_case_globals)]
#![allow(non_camel_case_types)]
#![allow(non_snake_case)]
include!("./xx.rs");
```
[...]

PR #17863, opened 2024-08-12
fix: Resolve included files to their calling modules in IDE layer

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/17390 at the expense of reporting duplicate diagnostics for modules that have includes in them when both the calling and called file are included.


NessajCN:
Hey @Veykril 
I am not sure if I should open a new issue as #17390 has been closed.
After your fix has been updated I found r-a did not recognize the following notation correctly to included file using `include!()` macro
```
#![allow(non_upper_case_globals)]
#![allow(non_camel_case_types)]
#![allow(non_snake_case)]
include!("./xx.rs");
```
Warnings like `Type alias __u_char should have CamelCase name, e.g. UChar` still popped up.
Could you possibly fix this?

PR #17886, opened 2024-08-13
internal: ServerStatusParams should consider 'prime caches' in quiescent status

Wilfred:
Priming caches is a performance win, but it takes a lock on the salsa database and prevents rust-analyzer from responding to e.g. go-to-def requests.

This causes confusion for users, who see the spinner next to rust-analyzer in the VS Code footer stop, so they start attempting to navigate their code.

Instead, set the `quiescent` status in LSP to false during cache priming, so the VS Code spinner persists until we can respond to any LSP request.

### DIFF
--- a/crates/rust-analyzer/src/reload.rs
+++ b/crates/rust-analyzer/src/reload.rs
@@ -61,6 +61,10 @@ pub(crate) enum ProcMacroProgress {
 }
 
 impl GlobalState {
+    /// Is the server quiescent?
+    ///
+    /// This indicates that we've fully loaded the projects and
+    /// are ready to do semantic work.
     pub(crate) fn is_quiescent(&self) -> bool {
         self.vfs_done
             && self.last_reported_status.is_some()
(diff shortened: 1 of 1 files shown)


Wilfred:
Zulip thread: https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/Making.20'prime.20caches'.20state.20explicit.20in.20VS.20Code.20UI

crates/rust-analyzer/src/reload.rs
```diff
@@ -61,6 +61,10 @@ pub(crate) enum ProcMacroProgress {
 }
 
 impl GlobalState {
+    /// Is the server quiescent?
+    ///
+    /// This indicates that we've fully loaded the Cargo project and
+    /// can respond to (at least) some LSP requests.
```
Veykril, line 67:
```suggestion
    /// This indicates that we've fully loaded the projects and
    /// are ready to do semantic work.
```

crates/rust-analyzer/src/reload.rs
```diff
@@ -71,6 +75,15 @@ impl GlobalState {
             && self.vfs_progress_config_version >= self.vfs_config_version
     }
 
+    /// Is the server ready to respond to any LSP request?
```
Veykril, line 78:
```suggestion
    /// Is the server ready to respond to analysis dependent LSP requests?
```

Veykril (approved):
Slight comment change suggestions

PR #17888, opened 2024-08-14
chore(config): remove `invocationLocation` in favor of `invocationStrategy`

### LINKED ISSUE #17848: Collapse invocationLocation and invocationStrategy into a single setting
These options were added to explore how RA can best work with repos like rustc and Miri that need special non-cargo invocations (`./x.py ...`/`./miri ...`). The consensus that generally has emerged is that these scripts should be invoked exactly once, in the root, and they will emit all the diagnostics for all parts of the project with paths relative to the root. There were ideas of having RA pass `--manifest-path` or so, with template magic available in `overrideCommand` to control where to put that flag; I don't think any more that those are promising.

If there are no other usecases being covered by these flags then most of the freedom provided by having two independent settings, invocationLocation and invocationStrategy, isn't actually needed. A possible simplification here might be to only have invocationLocation, and if it is set to "root" then that also implies running it only once (since it seems kind of pointless to run the same command in the same location multiple times). invocationStrategy could then be removed. Or conversely, `invocationStrategy` is preserved and setting it to `once` means "run it once, in the root".

Tyrubias:
These flags were added to help rust-analyzer integrate with repos requiring non-Cargo invocations. The consensus is that having two independent settings are no longer needed. This change removes `invocationLocation` in favor of `invocationStrategy` and changes the internal representation of `InvocationStrategy::Once` to hold the workspace root.

Closes #17848.

### DIFF
--- a/crates/project-model/src/build_dependencies.rs
+++ b/crates/project-model/src/build_dependencies.rs
@@ -19,8 +19,8 @@ use serde::Deserialize;
 use toolchain::Tool;
 
 use crate::{
-    utf8_stdout, CargoConfig, CargoFeatures, CargoWorkspace, InvocationLocation,
-    InvocationStrategy, ManifestPath, Package, Sysroot, TargetKind,
+    utf8_stdout, CargoConfig, CargoFeatures, CargoWorkspace, InvocationStrategy, ManifestPath,
+    Package, Sysroot, TargetKind,
 };
 
 /// Output of the build script and proc-macro building steps for a workspace.
@@ -63,10 +63,7 @@ impl WorkspaceBuildScripts {
         progress: &dyn Fn(String),
         sysroot: &Sysroot,
     ) -> io::Result<WorkspaceBuildScripts> {
-        let current_dir = match &config.invocation_location {
-            InvocationLocation::Root(root) if config.run_build_script_command.is_some() => root,
-            _ => workspace.workspace_root(),
-        };
+        let current_dir = workspace.workspace_root();
 
         let allowed_features = workspace.workspace_features();
         let cmd = Self::build_command(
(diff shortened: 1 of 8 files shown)


Tyrubias:
I changed both `project_model::InvocationStrategy::Once` and `flycheck::InvocationStrategy::Once` to hold the workspace root. I'm not sure if this is actually needed because it appears the workspace root is available from other sources whenever these enum variants are used. Let me know if we actually need it in one or both enums.

Veykril (approved):

Veykril:
Thanks!
@bors r+

crates/rust-analyzer/src/config.rs
```diff
@@ -80,13 +80,6 @@ config_data! {
...
-        /// Specifies the working directory for running build scripts.
-        /// - "workspace": run build scripts for a workspace in the workspace's root directory.
-        ///   This is incompatible with `#rust-analyzer.cargo.buildScripts.invocationStrategy#` set to `once`.
-        /// - "root": run build scripts in the project's root directory.
-        /// This config only has an effect when `#rust-analyzer.cargo.buildScripts.overrideCommand#`
-        /// is set.
-        cargo_buildScripts_invocationLocation: InvocationLocation = InvocationLocation::Workspace,
         /// Specifies the invocation strategy to use when running the build scripts command.
         /// If `per_workspace` is set, the command will be executed for each workspace.
         /// If `once` is set, the command will be executed once.
```
RalfJung, line 85: This should document the working directory that the comment will be executed in, for both cases.

crates/rust-analyzer/src/config.rs
```diff
@@ -3196,14 +3158,6 @@ fn field_props(field: &str, ty: &str, doc: &[&str], default: &str) -> serde_json
                 "The command will be executed once."
```
RalfJung, line 3158: Might be good to also mention the effect on the working directory here.

crates/project-model/src/build_dependencies.rs
```diff
@@ -89,15 +86,7 @@ impl WorkspaceBuildScripts {
...
-        let current_dir = match &config.invocation_location {
-            InvocationLocation::Root(root) => root,
-            InvocationLocation::Workspace => {
-                return Err(io::Error::new(
-                    io::ErrorKind::Other,
-                    "Cannot run build scripts from workspace with invocation strategy `once`",
-                ))
-            }
-        };
+        let current_dir = workspace_root;
```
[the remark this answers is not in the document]
Veykril: The variable name here is a bit misleading (should be renamed into `current_dir`/`working_dir`, the sole caller of this passes the actual root here.

PR #17896, opened 2024-08-15
internal: Properly check the edition for edition dependent syntax kinds

Veykril:
This puts the current edition in a bunch of places, most of which I annoted with FIXMEs asside from the ones in ide-assists because I couldnt bother with those


Veykril:
Handling the `dyn` keyword correctly for our generated syntax stuff is a pain btw because it can be anything due to editions, contextual/weak keyword in 2015, strict keyword in 2018+

PR #17900, opened 2024-08-15
Add scip/lsif flag to exclude vendored libaries

darichey:
#17809 changed StaticIndex to include vendored libraries. This PR adds a flag to disable that behavior.

At work, our monorepo has too many rust targets to index all at once, so we split them up into several shards. Since all of our libraries are vendored, if rust-analyzer includes them, sharding no longer has much benefit, because every shard will have to index the entire transitive dependency graphs of all of its targets. We get around the issue presented in #17809 because some other shard will index the libraries directly.

### DIFF
--- a/crates/ide/src/static_index.rs
+++ b/crates/ide/src/static_index.rs
@@ -119,6 +119,11 @@ fn documentation_for_definition(
     def.docs(sema.db, famous_defs.as_ref())
 }
 
+pub enum VendoredLibrariesConfig<'a> {
+    Included { workspace_root: &'a VfsPath },
+    Excluded,
+}
+
 impl StaticIndex<'_> {
     fn add_file(&mut self, file_id: FileId) {
         let current_crate = crates_for(self.db, file_id).pop().map(Into::into);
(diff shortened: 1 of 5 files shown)


crates/ide/src/static_index.rs
```diff
@@ -230,7 +230,11 @@ impl StaticIndex<'_> {
         self.files.push(result);
     }
 
-    pub fn compute<'a>(analysis: &'a Analysis, workspace_root: &VfsPath) -> StaticIndex<'a> {
+    pub fn compute<'a>(
+        analysis: &'a Analysis,
+        workspace_root: &VfsPath,
+        exclude_vendored_libraries: bool,
```
davidbarsky, line 236: up to you, but rust-analyzer has a lot of booleans that might be better served as enums. what do you think about making this something like:

```rust
enum VendoredLibraries {
    Include,
    Exclude
}
```

davidbarsky (commented):
looks good to me, modulo a single comment

crates/ide/src/static_index.rs
davidbarsky: if `workspace_root` is only used when vendored libraries are included, maybe something like this would work:

```rust
enum VendoredLibraries<'a> {
    Included { workpace_root: &'a VfsPath },
    Excluded,
}
```

davidbarsky (approved):
lgtm; i'll merge this.

PR #17903, opened 2024-08-15
Allow flycheck process to exit gracefully

### LINKED ISSUE #17902: Flycheck subprocess group is forcefully terminated after it has sent EOF

rust-analyzer 0.0.0 (b5626032b 2024-05-15)

VSCode

Using `rust-analyzer.check.overrideCommand`

---

We're using a custom check command on our repo. This command optionally collects duration metrics, so we can improve our developer experience. These metrics are collected in the background after the process completes.

Unfortunately, as soon as rust-analyzer gets an EOF from the check command it sends a SIGKILL to the entire process group. This prevents the metrics from being collected.

I don't think such aggressive killing is warranted. If the check is being rerun then it's fine to kill it as a way of cancelling quickly. But if it completes normally, it should be allowed to exit normally.

tmandry:
Assuming it isn't cancelled. Closes #17902.

The only place CommandHandle::join() is used is when the flycheck command
finishes, so this commit changes the behavior of the method itself.

The only reason I can see for the existing behavior is if the command is somehow holding onto a build lock longer than it should, this would force it to be released. But it would be a pretty heavy-handed way to solve that issue. I'm not aware of this occurring in practice.

### DIFF
--- a/crates/rust-analyzer/src/command.rs
+++ b/crates/rust-analyzer/src/command.rs
@@ -148,7 +148,6 @@ impl<T: ParseFromLine> CommandHandle<T> {
     }
 
     pub(crate) fn join(mut self) -> io::Result<()> {
-        let _ = self.child.0.kill();
         let exit_status = self.child.0.wait()?;
         let (read_at_least_one_message, error) = self.thread.join()?;
         if read_at_least_one_message || exit_status.success() {


Veykril:
Hmm, I am pretty sure we had a reason for adding this. I *believe* we used to sometimes leak cargo processes because we weren't killing the process group (usually when people invoke some wrapper like thing instead of cargo directly).

Veykril:
But I might be wrong and that issue was with something else, we'll see if something crops up 
@bors r+

tmandry:
My impression from looking at the original PR is that using the process group was important when cancelling the job, because you want to be able to start the next one as soon as possible.

Issue #10935, opened 2021-12-05
proc macro attributes considered to be `unresolvedReference`


jhgg:
consider the following code used w/ the serde library:

```rust
use serde::Deserialize;

#[derive(Deserialize)]
struct Foo {
    #[serde(default)]
   // ^^^^^ unresolvedReference 
    bar: u64,
}
```

We should be able to figure out if this is a valid attribute or not, by looking at the things the structs derive from and looking at the attributes for each proc macro to compute a list of identifiers that are valid within this context:

```rust
// in serde_derive/src/lib.rs
#[proc_macro_derive(Deserialize, attributes(serde))]
                                         // ^^^^^ we should find these.
pub fn derive_deserialize(input: TokenStream) -> TokenStream {
    let mut input = parse_macro_input!(input as DeriveInput);
    de::expand_derive_deserialize(&mut input)
        .unwrap_or_else(to_compile_errors)
        .into()
}
```

Veykril:
Ah right, I completely forgot about (derive) helper attributes which we now show as unresolved since they aren't being mapped to anything by macros...

As you said we can do some specific things here to fix that for derive helpers specifically, but the problem persists for helper attributes of proc-macro attributes like with salsa here 
![image](https://user-images.githubusercontent.com/3757771/144743059-b3784edc-fb0d-4cd7-bc2a-f90922a28509.png)


So I guess we'll have to add that attribute highlight heuristics back for now...

I do like the hover idea though 👍

Veykril:
Actually now I wonder if it would make sense to push proc macro authors to expose dummy attributes that resolve to these helpers. That way people could also hover them for more info on the helper itself 🤔

Veykril:
Opened a (zulip thread)[https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/.28derive.29.20helper.20attribute.20resolution/near/263763005] for the latter idea.

PR #17904, opened 2024-08-16
Add command to report unresolved references

darichey:
Adds `rust-analyzer unresolved-references` which reports unresolved references. This is useful for debugging and regression testing for both rust-analyzer and project generators like Buck's rust-project.

As discussed: https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/Command.20to.20report.20unresolved.20references

### DIFF
--- a/crates/rust-analyzer/src/bin/main.rs
+++ b/crates/rust-analyzer/src/bin/main.rs
@@ -82,6 +82,7 @@ fn actual_main() -> anyhow::Result<ExitCode> {
         flags::RustAnalyzerCmd::Highlight(cmd) => cmd.run()?,
         flags::RustAnalyzerCmd::AnalysisStats(cmd) => cmd.run(verbosity)?,
         flags::RustAnalyzerCmd::Diagnostics(cmd) => cmd.run()?,
+        flags::RustAnalyzerCmd::UnresolvedReferences(cmd) => cmd.run()?,
         flags::RustAnalyzerCmd::Ssr(cmd) => cmd.run()?,
         flags::RustAnalyzerCmd::Search(cmd) => cmd.run()?,
         flags::RustAnalyzerCmd::Lsif(cmd) => cmd.run()?,
--- a/crates/rust-analyzer/src/cli.rs
+++ b/crates/rust-analyzer/src/cli.rs
@@ -13,6 +13,7 @@ mod rustc_tests;
 mod scip;
 mod ssr;
 mod symbols;
+mod unresolved_references;
 
 mod progress_report;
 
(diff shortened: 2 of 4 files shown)


darichey:
This isn't quite working totally yet, but opening the PR to get some feedback and ask a couple questions. The approach is based on `NameRefClass`, but there are some cases this doesn't handle properly on its own. I think `highlight.rs` has all of the answers, but I'd appreciate some pointers :) 

This is the output when running on rust-analyzer itself: https://gist.github.com/darichey/55e411f55b722d9cc1602e8f273c1a67

It does find some actual issues! e.g.,
```
rust-analyzer/crates/hir-expand/src/files.rs:199:71: text_range
rust-analyzer/crates/salsa/src/runtime.rs:468:45: LEN
rust-analyzer/crates/syntax/src/ted.rs:71:31: last_child_or_token
rust-analyzer/crates/hir-def/src/resolver.rs:1058:13: module
```

However, there's a lot of false positives. In particular, code within `#[tracing::instrument]` (and other proc macros, like `Self` inside a `#[salsa::database]`) gets reported as unresolved. I'm not sure why that is.

Also, as reported in #10935, proc macro attributes are unresolved. Should we copy the hack for highlighting in #10937 or keep them since they _are_ technically unresolved?

darichey:
I've fixed the handling of macros and decided to ignore attributes. This is the new output when running on rust-analyzer: https://gist.github.com/darichey/2f74b0e45f5c1ff18d4ea78039e2c831

All of those are correctly-identified unresolved references except the last one:
https://github.com/rust-lang/rust-analyzer/blob/9b724459b5900c2424f5069023d6596a8617284b/crates/cfg/src/tests.rs#L248
I assume it has something to do with the conditional deriving of Arbitrary, but I wasn't able to repro in isolation. 

Other than that issue, I think this is ready for review.

Veykril:
Had a quick look but I am also not sure why that is. Either way commited some small cleansups and fixes (maybe that fixes things though I doubt it)

darichey:
Figured it out! We need to set `all_targets: true` in the `CargoConfig`. Without it, we don't get artifact info for dev-dependencies when building build dependencies, so we can't expand proc macros from them. 

For now, I've switched to creating the `CargoConfig` via `Config`, the same way `scip` does. This ensures we get the same defaults rust-analyzer does in normal usage. As a follow up, it would be good to extract the common workspace loading logic from scip, lsif, diagnostics, and unresolved_references to make sure they all do the right thing (it looks like lsif and diagnostics would have the same problem).

With that, there are no more false positives in rust-analyzer: https://gist.github.com/darichey/10a7b13961216e680f0d6b1085f79ab5

I will run it on our monorepo to check for any more false positives, but I think this is a good start we can iterate on!

Veykril (approved):

PR #17912, opened 2024-08-16
fix: run flycheck without rev_deps when target is specified

### LINKED ISSUE #17829: Pass `--bin` to `cargo check` when a binary is changed
I marked this as important because it affects Rust beginners and I don't want them to get a bad expression while learning with Rustlings. It causes Rustlings to freeze for a long time because Rust-Analyzer blocks Cargo.

The issue is described here with a video: https://github.com/rust-lang/rustlings/issues/2071

I was able to reproduce it in VS-Code in a Windows 10 VM (probably not a Windows problem, didn't test on Linux).

You don't even need to have Rustlings running in the background to see how Rust-Analyzer rebuilds everything on every file save:
- Initialize Rustlings with `rustlings inti`
- Open VS-Code in the generated `rustlings/` directory
- Wait for Rust-Analyzer to be done
- Open a file and trigger a save
- See that Rust-Analyzer **rebuilds all 188 binaries!**

~It looks like an issue in the Rust-Analyzer VS-Code plugin because it doesn't happen in Helix in the same VM.~

The effect probably became more dramatic after I added `--keep-going` in https://github.com/rust-lang/rust-analyzer/pull/17561

alibektas:
Since querying for a crate's target is a call to salsa and therefore blocking, flycheck task is now deferred out of main thread by using `GlobalState`s `deferred_task_queue`. Fixes #17829  and https://github.com/rust-lang/rustlings/issues/2071

### DIFF
--- a/crates/rust-analyzer/src/flycheck.rs
+++ b/crates/rust-analyzer/src/flycheck.rs
@@ -34,6 +34,14 @@ pub(crate) struct CargoOptions {
     pub(crate) target_dir: Option<Utf8PathBuf>,
 }
 
+#[derive(Clone)]
+pub(crate) enum Target {
+    Bin(String),
+    Example(String),
+    Benchmark(String),
+    Test(String),
+}
+
 impl CargoOptions {
     pub(crate) fn apply_on_command(&self, cmd: &mut Command) {
         for target in &self.target_triples {
@@ -118,12 +126,12 @@ impl FlycheckHandle {
     }
 
     /// Schedule a re-start of the cargo check worker to do a workspace wide check.
-    pub(crate) fn restart_workspace(&self, saved_file: Option<AbsPathBuf>, target: Option<String>) {
-        self.sender.send(StateChange::Restart { package: None, saved_file, target }).unwrap();
+    pub(crate) fn restart_workspace(&self, saved_file: Option<AbsPathBuf>) {
+        self.sender.send(StateChange::Restart { package: None, saved_file, target: None }).unwrap();
     }
 
     /// Schedule a re-start of the cargo check worker to do a package wide check.
-    pub(crate) fn restart_for_package(&self, package: String, target: Option<String>) {
+    pub(crate) fn restart_for_package(&self, package: String, target: Option<Target>) {
         self.sender
             .send(StateChange::Restart { package: Some(package), saved_file: None, target })
             .unwrap();
@@ -183,7 +191,7 @@ pub(crate) enum Progress {
 }
 
 enum StateChange {
-    Restart { package: Option<String>, saved_file: Option<AbsPathBuf>, target: Option<String> },
+    Restart { package: Option<String>, saved_file: Option<AbsPathBuf>, target: Option<Target> },
     Cancel,
 }
 
@@ -281,14 +289,12 @@ impl FlycheckActor {
                         }
                     }
 
-                    let command = match self.check_command(
-                        package.as_deref(),
-                        saved_file.as_deref(),
-                        target.as_deref(),
-                    ) {
-                        Some(c) => c,
-                        None => continue,
+                    let Some(command) =
+                        self.check_command(package.as_deref(), saved_file.as_deref(), target)
+                    else {
+                        continue;
                     };
+
                     let formatted_command = format!("{command:?}");
 
                     tracing::debug!(?command, "will restart flycheck");
@@ -384,7 +390,7 @@ impl FlycheckActor {
         &self,
         package: Option<&str>,
         saved_file: Option<&AbsPath>,
-        bin_target: Option<&str>,
+        target: Option<Target>,
     ) -> Option<Command> {
         match &self.config {
             FlycheckConfig::CargoCommand { command, options, ansi_color_output } => {
@@ -400,8 +406,13 @@ impl FlycheckActor {
                     None => cmd.arg("--workspace"),
                 };
 
-                if let Some(tgt) = bin_target {
-                    cmd.arg("--bin").arg(tgt);
+                if let Some(tgt) = target {
+                    match tgt {
+                        Target::Bin(tgt) => cmd.arg("--bin").arg(tgt),
+                        Target::Example(tgt) => cmd.arg("--example").arg(tgt),
+                        Target::Test(tgt) => cmd.arg("--test").arg(tgt),
+                        Target::Benchmark(tgt) => cmd.arg("--bench").arg(tgt),
+                    };
                 }
 
                 cmd.arg(if *ansi_color_output {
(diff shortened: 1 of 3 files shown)


alibektas:
It has been mentioned that this particular case is not only limited to binary targets. I would be happy to add also other cases.

Veykril (commented):
Okay two things, I think I was wrong with this requiring a queued task as we have already been doing all the work within the spawned task (and can still continue to do so), so we can undo that part. The other thing is that when we do have the binary case here, we should not only skip the reverse dependency stuff, we should also tell flycheck to pass `--bin <bin-name>` for its check, that will require some more changes in the flycheck restart logic

crates/rust-analyzer/src/flycheck.rs
```diff
@@ -396,6 +400,10 @@ impl FlycheckActor {
                     None => cmd.arg("--workspace"),
                 };
 
+                if let Some(tgt) = bin_target {
+                    cmd.arg("--bin").arg(tgt);
```
mo8it, line 404: We probably should also pass the package here (non-optional). Two packages in a workspace can have the a binary with the same name. In that case, Cargo checks both of them (tested it).

crates/rust-analyzer/src/handlers/notification.rs
```diff
@@ -347,8 +365,11 @@ fn run_flycheck(state: &mut GlobalState, vfs_path: VfsPath) -> bool {
                     if id == flycheck.id() {
                         updated = true;
                         match package.filter(|_| !world.config.flycheck_workspace()) {
-                            Some(package) => flycheck.restart_for_package(package),
-                            None => flycheck.restart_workspace(saved_file.clone()),
+                            Some(package) => {
+                                flycheck.restart_for_package(package, target_is_bin.clone())
```
mo8it, line 369: We could break here if the target is a binary and avoid cloning.

crates/rust-analyzer/src/flycheck.rs
```diff
@@ -118,14 +118,14 @@ impl FlycheckHandle {
     }
 
     /// Schedule a re-start of the cargo check worker to do a workspace wide check.
-    pub(crate) fn restart_workspace(&self, saved_file: Option<AbsPathBuf>) {
-        self.sender.send(StateChange::Restart { package: None, saved_file }).unwrap();
+    pub(crate) fn restart_workspace(&self, saved_file: Option<AbsPathBuf>, target: Option<String>) {
```
mo8it, line 121: Workspace with target binary isn't a good fit (see comment below). So only allow passing target binary in `restart_for_package`.

crates/rust-analyzer/src/flycheck.rs
```diff
@@ -281,11 +281,14 @@ impl FlycheckActor {
                         }
                     }
 
-                    let command =
-                        match self.check_command(package.as_deref(), saved_file.as_deref()) {
-                            Some(c) => c,
-                            None => continue,
-                        };
+                    let command = match self.check_command(
```
Veykril, line 284: This can be rewritte as a `let else` statement now (there should be an assist for that)

crates/rust-analyzer/src/flycheck.rs
Veykril: I think its fine to support either, the package is passed above if a package was set (which should be usually the case)

crates/rust-analyzer/src/flycheck.rs
```diff
@@ -183,7 +183,7 @@ pub(crate) enum Progress {
 }
 
 enum StateChange {
-    Restart { package: Option<String>, saved_file: Option<AbsPathBuf> },
+    Restart { package: Option<String>, saved_file: Option<AbsPathBuf>, target: Option<String> },
```
Veykril, line 186:
```suggestion
    Restart { package: Option<String>, saved_file: Option<AbsPathBuf>, target: Option<Target>},
```
```rs
enum Target {
    Bin(String)
}
```
would be better to change that to an enum up front so we can support other targets more easily (mainly examples come to mind), as those require different flags than bin

crates/rust-analyzer/src/handlers/notification.rs
```diff
@@ -287,16 +288,33 @@ fn run_flycheck(state: &mut GlobalState, vfs_path: VfsPath) -> bool {
...
+                        TargetSpec::ProjectJson(p) => Some(p.label),
+                    };
+                }
+
+                None
+            });
+
+            let crate_ids = if target_is_bin.is_some() {
+                // Trigger flychecks for the only workspace which the binary crate belongs to
+                world.analysis.crates_for(file_id)?.into_iter().unique().collect::<Vec<_>>()
```
Veykril, line 305: We should already have that information in the `TargetSpec`, it should contain the relevant crate id

crates/rust-analyzer/src/handlers/notification.rs
Veykril: The clone is fine, this isn't really performance critical code. I'd rather we don't break here as it just makes this more complicated than necessary

crates/rust-analyzer/src/flycheck.rs
mo8it: Bin, example, test and bench

crates/rust-analyzer/src/flycheck.rs
mo8it: Not for Rustlings. But if we extend this to tests instead of only bins, a workspace could have multiple packages with the test `integration_tests` for example. I don't see why rust-analyzer should check all of them if only one changes.

Veykril: Fair, I'd still leave the logic as is though and just change the restart callers as this invocation isn't technically invalid.

crates/rust-analyzer/src/handlers/notification.rs
```diff
@@ -347,8 +365,11 @@ fn run_flycheck(state: &mut GlobalState, vfs_path: VfsPath) -> bool {
                     if id == flycheck.id() {
                         updated = true;
                         match package.filter(|_| !world.config.flycheck_workspace()) {
```
Veykril, line 367:
```suggestion
                        match package.filter(|_| !world.config.flycheck_workspace() || target_is_bin.is_some()) {
```
As per https://github.com/rust-lang/rust-analyzer/pull/17912#discussion_r1724704318

crates/rust-analyzer/src/handlers/notification.rs
alibektas: this is the only bit I couldn't really come up with a solution because the package information for `RustProjectJson` is not really clear.

Veykril: No worries, https://github.com/rust-lang/rust-analyzer/pull/17946 the data was there but not exposed that's all

PR #17923, opened 2024-08-18
Building before a debugging session was restarted

### LINKED ISSUE #17901: Support for building when a debug session is restarted
**Background**
Above every test that you write, a `Run Test | Debug` header is visible, making it easy to quickly debug one particular test.
When you debug, the project is compiled first, and then the process starts. 

```rust
#[test]
fn test_x() {
    assert_eq!("1", "1.1.0");
}
```
However, when you press the restart button, the compile process is not restarted. When you change code in the mean time, you'd have to end the debugging session and run it again.

**Suggested feature**
- Have an `rust-analyzer.debug.prestartcommand` option to make the users fill in a command to rebuild the process.
- Re-compile the project when the restart button is pressed. If this is an unwanted feature, make it configurable via `rust-analyzer.debug.recompileAfterRestart`

basvandriel:
# Background
Resolves #17901. It adds support for rebuilding after debugging a test was restarted. This means the test doesn't have to be aborted and manually re-ran again.

# How this is tested
First, all the Visual Studio Code extensions are loaded into an Extension Host window. Then, a sample test like below was ran and restarted to see if it was correctly rebuild.

```rust
#[test]
fn test_x() {
    assert_eq!("1.1.1", "1.1.0");
}
```

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -520,6 +520,11 @@
                         "type": "boolean",
                         "default": false
                     },
+                    "rust-analyzer.debug.buildBeforeRestart": {
+                        "markdownDescription": "Whether to rebuild the project modules before debugging the same test again",
+                        "type": "boolean",
+                        "default": false
+                    },
                     "rust-analyzer.debug.engineSettings": {
                         "type": "object",
                         "default": {},
--- a/editors/code/src/debug.ts
+++ b/editors/code/src/debug.ts
@@ -5,11 +5,14 @@ import type * as ra from "./lsp_ext";
 
 import { Cargo } from "./toolchain";
 import type { Ctx } from "./ctx";
-import { prepareEnv } from "./run";
+import { createTaskFromRunnable, prepareEnv } from "./run";
 import { execute, isCargoRunnableArgs, unwrapUndefinable } from "./util";
 import type { Config } from "./config";
 
 const debugOutput = vscode.window.createOutputChannel("Debug");
+
+// Here we want to keep track on everything that's currently running
+const activeDebugSessionIds: string[] = [];
 
 export async function makeDebugConfig(ctx: Ctx, runnable: ra.Runnable): Promise<void> {
     const scope = ctx.activeRustEditor?.document.uri;
@@ -45,6 +48,8 @@ export async function startDebugSession(ctx: Ctx, runnable: ra.Runnable): Promis
     const wsLaunchSection = vscode.workspace.getConfiguration("launch");
     const configurations = wsLaunchSection.get<any[]>("configurations") || [];
 
+    // The runnable label is the name of the test with the "test prefix"
+    // e.g. test test_feature_x
     const index = configurations.findIndex((c) => c.name === runnable.label);
     if (-1 !== index) {
         debugConfig = configurations[index];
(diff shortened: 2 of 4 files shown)


editors/code/package.json
```diff
@@ -520,6 +520,11 @@
                         "type": "boolean",
                         "default": false
                     },
+                    "rust-analyzer.debug.buildBeforeRestart": {
+                        "markdownDescription": "Whether to rebuild the project modules before debugging the same test again",
```
Veykril, line 524: is there ever a reason not wanting to have this behavior?

Cargo.toml
```diff
@@ -11,7 +11,7 @@ authors = ["rust-analyzer team"]
 repository = "https://github.com/rust-lang/rust-analyzer"
 
 [profile.dev]
-debug = 1
+debug = 2
```
Veykril, line 14: Please undo this again

editors/code/src/main.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
Veykril, line 1: We should probably replace that console access with our log class (feel free to put a FIXME on this eslint comment for that)

editors/code/src/debug.ts
```diff
@@ -359,3 +370,56 @@ function quote(xs: string[]) {
...
+        }
+    });
+
+    vscode.debug.onDidTerminateDebugSession(async (session: vscode.DebugSession) => {
+        cleanupDebugSessionIdBuffer();
+        removeActiveSession(session);
+    
+        // The id of the session will be the same when pressing restart the restart button
+        if (debugSessionIdBuffer.find((s) => s === session.id)) {
+            execShell("cargo test --no-run");
```
Veykril, line 409: why are we executing cargo when the session ends?

editors/code/src/debug.ts
```diff
@@ -6,11 +6,20 @@ import type * as ra from "./lsp_ext";
...
 import type { Config } from "./config";
 
 const debugOutput = vscode.window.createOutputChannel("Debug");
 
+
+// Every debug session with the same id (restart-candidate) should be in this list
+let debugSessionIdBuffer: string[] = [];
+
+// Here we want to keep track on everything that's currently running
+const activeDebugSessionIds: string[] = [];
```
Veykril, line 19: This seems a bit odd, to my knowledge there can only ever be one active debug session in vscode

editors/code/src/debug.ts
```diff
@@ -6,11 +6,20 @@ import type * as ra from "./lsp_ext";
...
 import { prepareEnv } from "./run";
-import { execute, isCargoRunnableArgs, unwrapUndefinable } from "./util";
+import { execute, isCargoRunnableArgs, log, unwrapUndefinable } from "./util";
 import type { Config } from "./config";
 
 const debugOutput = vscode.window.createOutputChannel("Debug");
 
+
+// Every debug session with the same id (restart-candidate) should be in this list
+let debugSessionIdBuffer: string[] = [];
```
Veykril, line 16: Likewise this seems unnecessary, we should only need to track the last session and compare that?

editors/code/src/main.ts
basvandriel: Removed the line, this was only there for some debugging purposes.

Cargo.toml
basvandriel: Resolved!

editors/code/src/debug.ts
basvandriel: While this does seem like a complex feature, it is supported to have multiple debug sessions :)

image

editors/code/package.json
basvandriel: This was added to make the behavior more flexible. I can imagine there could be use cases where you want to debug code without it rebuilding! Why should it be fixed?

editors/code/src/debug.ts
basvandriel: This takes care of the building process before the new test begins. Unfortunately, doing that in the `onDidStartDebugSession` hook is too late.

Veykril (commented):
Looking into this problem a bit more myself i am actually confused by this behavior. It feels like this is either a VSCode issue or we are missing some VSCode API for this, given VSCode ought to be the one invoking the building step already when we first debug something. So why does it not do that on restart?

editors/code/src/debug.ts
```diff
@@ -359,3 +366,46 @@ function quote(xs: string[]) {
...
+    vscode.debug.onDidStartDebugSession((session: vscode.DebugSession) => {
+        if (!activeDebugSessionIds.includes(session.id)) {
+            activeDebugSessionIds.push(session.id);
+        }
+    });
+
+    vscode.debug.onDidTerminateDebugSession(async (session: vscode.DebugSession) => {
+        // The id of the session will be the same when pressing restart the restart button
+        if (activeDebugSessionIds.find((s) => s === session.id)) {
+            execShell("cargo test --no-run");
```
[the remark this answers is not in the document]
basvandriel: I'll make this change later today!

basvandriel:
> Looking into this problem a bit more myself i am actually confused by this behavior. It feels like this is either a VSCode issue or we are missing some VSCode API for this, given VSCode ought to be the one invoking the building step already when we first debug something. So why does it not do that on restart?

It seems that the debugging session restarts and not the actual test. The root cause of this isn't clear to me as well.

editors/code/src/debug.ts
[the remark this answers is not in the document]
basvandriel: I need some time understand the code base better - there is already functionality in the analyzer server to resolve the correct testing path; `CargoTestHandle`.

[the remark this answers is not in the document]
basvandriel: Hi @Veykril, there is no need to invoke the language server, as all the information is already present in the current session. Published the PR again!

Veykril:
Not too happy with this honestly but I don't see how we can make vscode actually rebuild on restart either. Can you squash the commits?

basvandriel:
> Not too happy with this honestly but I don't see how we can make vscode actually rebuild on restart either. Can you squash the commits?

I would like a different solution for this as well, but that's why I made it an option in the first place. Feel free to squash when merging the pull request :)

editors/code/package.json
```diff
@@ -520,6 +520,11 @@
...
+                        "type": "boolean",
+                        "default": true
```
Veykril, line 526: Given I don't like the approach here very much this should default to `false`

Veykril:
Fine, one more change, also please squash yourself, I am unable to as merging is handled by a bot

editors/code/package.json
lnicola: > I can imagine there could be use cases where you want to debug code without it rebuilding!

I don't think so. If you changed the source, the debugger is going to be worse than useless (i.e. misleading) unless you only look at the assembly (and never at the source code, watches etc.).

I think this should at least default to `true`.

PR #17930, opened 2024-08-19
Remove the ability to configure the user config path

Veykril:
Being able to do this makes little sense as this is effectively a cyclic dependency (and we do not want to fixpoint this really).

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -3,10 +3,13 @@
 //! Of particular interest is the `feature_flags` hash map: while other fields
 //! configure the server itself, feature flags are passed into analysis, and
 //! tweak things like automatic insertion of `()` in completions.
-use std::{fmt, iter, ops::Not, sync::OnceLock};
+use std::{
+    env, fmt, iter,
+    ops::Not,
+    sync::{LazyLock, OnceLock},
+};
 
 use cfg::{CfgAtom, CfgDiff};
-use dirs::config_dir;
 use hir::Symbol;
 use ide::{
     AssistConfig, CallableSnippets, CompletionConfig, DiagnosticsConfig, ExprFillDefaultMode,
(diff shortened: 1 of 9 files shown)


crates/rust-analyzer/src/config.rs
```diff
@@ -794,8 +786,27 @@ impl std::ops::Deref for Config {
...
+    /// | Linux   | `$XDG_CONFIG_HOME` or `$HOME`/.config | /home/alice/.config                      |
+    /// | macOS   | `$HOME`/Library/Application Support   | /Users/Alice/Library/Application Support |
+    /// | Windows | `{FOLDERID_RoamingAppData}`           | C:\Users\Alice\AppData\Roaming           |
+    pub fn user_config_path() -> &'static AbsPath {
+        static USER_CONFIG_PATH: LazyLock<AbsPathBuf> = LazyLock::new(|| {
+            let user_config_path = if let Some(path) = env::var_os("__TEST_RA_USER_CONFIG_DIR") {
+                std::path::PathBuf::from(path)
+            } else {
+                dirs::config_dir()
+                    .expect("A config dir is expected to existed on all platforms ra supports.")
```
lnicola, line 803: This feels a bit too heavy-handed. I don't think we should crash `~/.config` doesn't exist, it doesn't get created until someone actually saves a file there.

lnicola: Ah, never mind, it doesn't check for existence.

Veykril: The panic message is misleading 😅

lnicola: I'm still not convinced that not having the directory should be a fatal error. r-a can still work fine enough without one. The Rust toolchain might not, but that's not our problem.

lnicola: There's no config dir on WASM, for example.

alibektas (approved):

PR #17948, opened 2024-08-23
fix: Wrong `Self: Sized` predicate for trait assoc items

ShoyuVanilla:
Again while implementing object safety like #17939 😅 

If we call `generic_predicates_query` on `fn foo` in the following code;
```
trait Foo {
    fn foo();
}
```
It returns implicit bound `Self: Sized`, even though `Self` is not appearing as a generic parameter inside angle brackets, but as a parent generic parameter, "trait self".

This PR prevent pushing "implicit" `Self: Sized` predicates in such cases

### DIFF
--- a/crates/hir-ty/src/lower.rs
+++ b/crates/hir-ty/src/lower.rs
@@ -1743,15 +1743,39 @@ fn implicitly_sized_clauses<'a, 'subst: 'a>(
     substitution: &'subst Substitution,
     resolver: &Resolver,
 ) -> Option<impl Iterator<Item = WhereClause> + Captures<'a> + Captures<'subst>> {
-    let is_trait_def = matches!(def, GenericDefId::TraitId(..));
-    let generic_args = &substitution.as_slice(Interner)[is_trait_def as usize..];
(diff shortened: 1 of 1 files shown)


crates/hir-ty/src/lower.rs
```diff
@@ -1743,24 +1743,48 @@ fn implicitly_sized_clauses<'a, 'subst: 'a>(
...
-        .and_then(|lang_item| lang_item.as_trait().map(to_chalk_trait_id));
+        .and_then(|lang_item| lang_item.as_trait().map(to_chalk_trait_id))?;
 
-    sized_trait.map(move |sized_trait| {
-        generic_args
-            .iter()
+    let get_trait_self_idx = |container: ItemContainerId| {
+        if matches!(container, ItemContainerId::TraitId(_)) {
+            let generics = generics(db.upcast(), def);
+            Some(generics.len_self())
```
ShoyuVanilla, line 1753: parent generics are comming after self ones

Veykril (approved):

PR #17954, opened 2024-08-24
Update rustc-hash to version 2

Noratrieb:
This brings in the new optimized algorithm that was shown to have small performance benefits for rustc. I haven't run the rust-analyzer benchmarks.

See https://github.com/rust-lang/rustc-hash/pull/37.


Noratrieb:
oh lord, it looks like something relies on the iteration order or something like that :3.

Veykril:
Yup, the iteration here calls a mutating callback and the state there seems to be iteration dependent https://github.com/rust-lang/rust-analyzer/blob/1271dff610719930bb2cf20206057e57201d2528/crates/hir-ty/src/lib.rs#L249

Veykril:
The funny thing is, that test apparently should succeed given rustc is fine with it?

```rs
        pub enum TagTree {
            Leaf,
            Choice(&'static [TagTree]),
        }
        const GOAL: TagTree = {
            const TAG_TREE: TagTree = TagTree::Choice(&[
                {
                    const VARIANT_TAG_TREE: TagTree = TagTree::Choice(
                        &[
                            TagTree::Leaf,
                        ],
                    );
                    VARIANT_TAG_TREE
                },
            ]);
            TAG_TREE
        };
        ```
        compiles just fine, yet the test errors on master

**Comment by HKalbasi** [comment]
The test looks strange. I guess it meant to do recursive constants, something like this:
```Rust
pub enum TagTree {
    Leaf,
    Choice(&'static [TagTree]),
}
const GOAL: TagTree = {
    const TAG_TREE: TagTree = TagTree::Choice(&[
        {
            const VARIANT_TAG_TREE: TagTree = TagTree::Choice(
                &[
                    TAG_TREE,
                ],
            );
            VARIANT_TAG_TREE
        },
    ]);
    TAG_TREE
};
```

The failure of the original test clearly is a bug. It should work with any order of execution, and sorting or using an stable iteration order does not fix the underlying problem. I would suggest changing the test to use an actual recursive constant like the code above, and opening an issue for that bug for further investigation.

Veykril:
Okay let's see what breaks
@bors r+

Noratrieb:
nice!

PR #17963, opened 2024-08-26
Always show error lifetime arguments as `'_`

### LINKED ISSUE #17947: Lifetime arguments of types are missing when hovering a value
```rust
let tmp = [0];
let _this_has_a_lifetime = tmp.iter();
```

When hovering `_this_has_a_lifetime`, RA shows:

> `let _this_has_a_lifetime: Iter<i32>`

popup

But this is not the full type. Crucially, this isn't a self-contained type, and it is tied a temporary loan. Hiding lifetimes in this case is very unhelpful, because I have to go to the documentation to see if the type has any lifetime arguments. 

I'd like this to be displayed as:

> `let _this_has_a_lifetime: Iter<'_, i32>`

or showing the form used at the point of definition of the `Iter`.

Lack of lifetime information is especially problematic when working with an unfamiliar library, because the code needs to be structured differently if the types are borrowing (to avoid self-referential types etc.), and the types will need to be annotated with lifetime arguments sooner or later. Without RA's help, I only learn about the lifetimes later when getting borrow checker errors.

avrong:
Fixes #17947

Changed error lifetime argument presentation in non-test environment to `'_` and now showing them even if all of args are error lifetimes.

This also influenced some of the other tests like `extract_function.rs`, `predicate.rs` and `type_pos.rs`. Not sure whether I need to refrain from adding lifetimes args there. Happy to fix if needed

### DIFF
--- a/crates/hir-ty/src/display.rs
+++ b/crates/hir-ty/src/display.rs
@@ -1438,17 +1438,7 @@ fn hir_fmt_generics(
 
     let parameters_to_write = generic_args_sans_defaults(f, generic_def, parameters);
 
-    // FIXME: Remote this
-    // most of our lifetimes will be errors as we lack elision and inference
-    // so don't render them for now
-    let only_err_lifetimes = !cfg!(test)
-        && parameters_to_write.iter().all(|arg| {
-            matches!(
-                arg.data(Interner),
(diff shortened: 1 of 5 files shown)


Veykril:
Needs a rustfmt run, thanks
@bors delegate+

PR #17984, opened 2024-08-28
feat: Implement cast typecheck and diagnostics

### LINKED ISSUE #16564: Incorrect "cannot move `&mut ()` out of reference" error
```rs
fn main() {
    let x = (&(&mut (),)).0 as *const ();
}
```
This compiles

### LINKED ISSUE #17897: No diagnostic reported for non-primitive casts
```rust
_ = ((), ()) as ();
```

ShoyuVanilla:
Fixes  #17897 and fixes #16564
Mainly adopted from https://github.com/rust-lang/rust/blob/100fde5246bf56f22fb5cc85374dd841296fce0e/compiler/rustc_hir_typeck/src/cast.rs

### DIFF
--- a/crates/hir-def/src/data/adt.rs
+++ b/crates/hir-def/src/data/adt.rs
@@ -14,6 +14,7 @@ use triomphe::Arc;
 use crate::{
     builtin_type::{BuiltinInt, BuiltinUint},
     db::DefDatabase,
+    hir::Expr,
     item_tree::{
         AttrOwner, Field, FieldParent, FieldsShape, ItemTree, ModItem, RawVisibilityId, TreeId,
     },
@@ -316,6 +317,27 @@ impl EnumData {
             Some(ReprOptions { int: Some(builtin), .. }) => builtin,
             _ => IntegerType::Pointer(true),
         }
+    }
+
+    // [Adopted from rustc](https://github.com/rust-lang/rust/blob/bd53aa3bf7a24a70d763182303bd75e5fc51a9af/compiler/rustc_middle/src/ty/adt.rs#L446-L448)
+    pub fn is_payload_free(&self, db: &dyn DefDatabase) -> bool {
+        self.variants.iter().all(|(v, _)| {
+            // The condition check order is slightly modified from rustc
+            // to improve performance by early returning with relatively fast checks
+            let variant = &db.enum_variant_data(*v).variant_data;
+            if !variant.fields().is_empty() {
+                return false;
+            }
+            // The outer if condition is whether this variant has const ctor or not
+            if !matches!(variant.kind(), StructKind::Unit) {
+                let body = db.body((*v).into());
+                // A variant with explicit discriminant
+                if body.exprs[body.body_expr] != Expr::Missing {
+                    return false;
+                }
+            }
+            true
+        })
     }
 }
 
--- a/crates/hir-ty/src/infer/cast.rs
+++ b/crates/hir-ty/src/infer/cast.rs
@@ -1,47 +1,412 @@
 //! Type cast logic. Basically coercion + additional casts.
 
-use crate::{infer::unify::InferenceTable, Interner, Ty, TyExt, TyKind};
+use chalk_ir::{Mutability, Scalar, TyVariableKind, UintTy};
+use hir_def::{hir::ExprId, AdtId};
+use stdx::never;
+
+use crate::{
+    infer::unify::InferenceTable, Adjustment, Binders, DynTy, InferenceDiagnostic, Interner,
+    PlaceholderIndex, QuantifiedWhereClauses, Ty, TyExt, TyKind, TypeFlags, WhereClause,
+};
+
+#[derive(Debug)]
+pub(crate) enum Int {
+    I,
+    U(UintTy),
+    Bool,
+    Char,
+    CEnum,
+    InferenceVar,
+}
+
+#[derive(Debug)]
+pub(crate) enum CastTy {
+    Int(Int),
+    Float,
+    FnPtr,
+    Ptr(Ty, Mutability),
+    // `DynStar` is Not supported yet in r-a
+}
+
+impl CastTy {
+    pub(crate) fn from_ty(table: &mut InferenceTable<'_>, t: &Ty) -> Option<Self> {
+        match t.kind(Interner) {
+            TyKind::Scalar(Scalar::Bool) => Some(Self::Int(Int::Bool)),
+            TyKind::Scalar(Scalar::Char) => Some(Self::Int(Int::Char)),
+            TyKind::Scalar(Scalar::Int(_)) => Some(Self::Int(Int::I)),
+            TyKind::Scalar(Scalar::Uint(it)) => Some(Self::Int(Int::U(*it))),
+            TyKind::InferenceVar(_, TyVariableKind::Integer) => Some(Self::Int(Int::InferenceVar)),
+            TyKind::InferenceVar(_, TyVariableKind::Float) => Some(Self::Float),
+            TyKind::Scalar(Scalar::Float(_)) => Some(Self::Float),
+            TyKind::Adt(..) => {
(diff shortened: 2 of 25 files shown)


crates/hir-ty/src/infer/cast.rs
```diff
@@ -1,47 +1,392 @@
...
 }
 
+#[derive(Debug)]
+enum Int {
+    I,
+    U(UintTy),
+    Bool,
+    Char,
+    CEnum,
+    InfereceVar,
```
Veykril, line 28:
```suggestion
    InferenceVar,
```

ShoyuVanilla: This somehow evaded typo checks 😂

crates/hir-ty/src/infer/cast.rs
```diff
@@ -1,47 +1,392 @@
...
+}
+
+fn check_ref_cast(
+    t_expr: &Ty,
+    m_expr: Mutability,
+    t_cast: &Ty,
+    m_cast: Mutability,
+    table: &mut InferenceTable<'_>,
+) -> Result<(), CastError> {
+    // Mutability order is opposite tu rustc. `Mut < Not`
```
Veykril, line 217: Might be good to change that on our side then (separately from this), the less we diverge from rustc with things the better we can draw comparisons between the code bases (which imo for type check is quite important)

ShoyuVanilla: I agree. But this type is from `chalk_ir`. Maybe we should change this at there?

ShoyuVanilla: BTW, another typo "tu" here 😅

Veykril: Ah I see, probably no point in changing it then. But we should keep that in mind for the trait solver migration 😬

ShoyuVanilla: Yes, there are many places in our code that tailored for chalk quirks and I can remember some of them. The ones that no one remembers or knows would be covered by our tests, I hope 😨

ShoyuVanilla:
I was blocked on mir evaluation failures when I push cast adjustments into InferenceResult and finally found the reason 🥳 our mir castkind is doing both mir and hir_typeck parts of rustc and thus pushing adjustment makes erroneous duplication.
I'll continue on this tommorrow

ShoyuVanilla:
So, this will also fix #16564 when finished

crates/ide-diagnostics/src/handlers/missing_unsafe.rs
```diff
@@ -278,7 +282,7 @@ fn main() {
 "#,
             r#"
 unsafe fn func() {
-    let x = &5 as *const usize;
+    let x = &5_usize as *const usize;
```
ShoyuVanilla, line 285: Previous cast `&5 as *const usize` is illegal

crates/hir-def/src/data/adt.rs
```diff
@@ -317,6 +318,23 @@ impl EnumData {
...
+    pub fn is_payload_free(&self, db: &dyn DefDatabase) -> bool {
+        self.variants.iter().all(|(v, _)| {
+            let variant = &db.enum_variant_data(*v).variant_data;
+            if !variant.fields().is_empty() {
+                return false;
+            }
+            if !matches!(variant.kind(), StructKind::Unit) {
+                let body = db.body((*v).into());
+                // A variant with explicit discriminant
+                if body.exprs[body.body_expr] != Expr::Missing {
```
Veykril, line 331: shouldn't the kind check suffice? Why query the body

crates/hir-ty/src/infer/unify.rs
```diff
@@ -898,6 +899,33 @@ impl<'a> InferenceTable<'a> {
...
+
+    /// Check if given type is `Sized` or not
+    pub(crate) fn is_sized(&mut self, ty: &Ty) -> bool {
+        if let Some((AdtId::StructId(id), subst)) = ty.as_adt() {
+            let struct_data = self.db.struct_data(id);
+            if let Some((last_field, _)) = struct_data.variant_data.fields().iter().last() {
+                let last_field_ty =
+                    self.db.field_types(id.into())[last_field].clone().substitute(Interner, subst);
+                // Structs can have DST as its last field and such cases are not handled
+                // as unsized by the chalk, so we do this manually
```
Veykril, line 911: (not sure how relevant this is but) this reminds me, we do not populate chalk's `AdtvariantDatum`'s fields  right now as that slows down things by quite a bit. Just letting you know so you are aware (the relevant part for that is stubbed out at the moment even though we have it implemented) https://github.com/rust-lang/rust-analyzer/blob/134e0a4e874e7a739b4e96c987c6b520f8da37ff/crates/hir-ty/src/chalk_db.rs#L768-L783

crates/hir-def/src/data/adt.rs
ShoyuVanilla: Because the condition is
Not having a const constructor && has a explicit discriminant

https://github.com/rust-lang/rust/blob/bd53aa3bf7a24a70d763182303bd75e5fc51a9af/compiler/rustc_middle/src/ty/adt.rs#L446-L448

The kind check is for the former and the body query is for explicit discriminant, though the order is opposite to rustc to mitigate perf issue

crates/hir-ty/src/infer/unify.rs
ShoyuVanilla: Oh, I was wondering why chalk is saying `struct Foo { inner: [u8] }` is `Sized` and that was the answer😄

crates/hir-def/src/data/adt.rs
Veykril: Ah okay, let's copy the comments over then

crates/hir-ty/src/infer/unify.rs
Veykril: iirc it added like 10 seconds to inference on self which is a bit much + at the time we couldnt inspect third party deps of the standard library (which we soon can) so we could reconsider if this turns out to cause troubles

ShoyuVanilla: In that case we might omit it and consider it as sized; that will make some false negative cast errors but we'll not have irritating false positives anyway (though we should change function name into somewhat like `maybe_sized` to avoid possible footguns in future that might be caused by trusting that function will always return the correct value 😅 )

crates/hir-def/src/data/adt.rs
ShoyuVanilla: Maybe we could apply the similar heuristic if the body query causes noticeable perf regression like in my comment below and that would make some false negatives, not false positives errors.
Or it might be possible to specialized function for quickly checking whether the enum variant has explicit discriminant or not

crates/hir-def/src/data/adt.rs
```diff
@@ -317,6 +318,27 @@ impl EnumData {
...
+            if !variant.fields().is_empty() {
+                return false;
+            }
+            // The outer if condition is whether this variant has const ctor or not
+            if !matches!(variant.kind(), StructKind::Unit) {
+                let body = db.body((*v).into());
+                // A variant with explicit discriminant
+                if body.exprs[body.body_expr] != Expr::Missing {
+                    return false;
+                }
```
ShoyuVanilla, line 337:
```suggestion
                let loc = v.lookup(db);
                let src = loc.source(db);
                if src.value.expr().is_some() {
                    return false;
                }
```

Might this be more performant?

Veykril: That makes anything calling this dependent on the parse tree, so that would be generally worse for incrementality. Querying the body should be fine, most enums don't have one anyways

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #17994, opened 2024-08-29
fix: Fix TokenStream::to_string implementation dropping quotation marks

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/17986
We might wanna consider backporting this to beta if that's simple enough to do


lnicola:
This doesn't seem to backport cleanly because the current beta is missing 93024ad4117f1aad6cb587da255049f95332cef5.

Veykril:
Ah, then that should be fine, I think the regression came from that PR (or judging from the beta source, the issue doesnt seem to be present there)

PR #17999, opened 2024-08-29
fix: `std::error::Error` is object unsafe

### LINKED ISSUE #17998: `std::error:Error` is considered as object unsafe
**rust-analyzer version**: `0ae42bd42576566540a84c62e118aa823edcf2ec`

image

While investigating regressions on `diesel` I found this😅

ShoyuVanilla:
Fixes #17998

I tried to get generic predicates of assoc function itself, not inherited from the parent here;

https://github.com/rust-lang/rust-analyzer/blob/0ae42bd42576566540a84c62e118aa823edcf2ec/crates/hir-ty/src/object_safety.rs#L420-L442

But this naive equality check approach doesn't work when the assoc function has one or more generic paramters like;

```rust
trait Foo {}
trait Bar: Foo {
    fn bar(&self);
}
```

because the generic predicates of the parent, `Bar` is `[^1.0 implements Foo]` and the generic predicates of `fn bar` is `[^1.1 implements Foo]`, which are different.

This PR implements a correct logic for filtering out parent generic predicates for this.

### DIFF
--- a/crates/hir-ty/src/db.rs
+++ b/crates/hir-ty/src/db.rs
@@ -154,6 +154,9 @@ pub trait HirDatabase: DefDatabase + Upcast<dyn DefDatabase> {
     #[salsa::invoke(crate::lower::generic_predicates_query)]
     fn generic_predicates(&self, def: GenericDefId) -> GenericPredicates;
 
+    #[salsa::invoke(crate::lower::generic_predicates_without_parent_query)]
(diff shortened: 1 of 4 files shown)


Veykril (approved):

ShoyuVanilla:
Screenshot_20240830-152144 (1)
Screenshot_20240830-152211 (1)
Phew, the metric regressions were resolved by this 😅

PR #18005, opened 2024-08-30
fix(ide-completion): fix handling of `for` in `impl T for A` in function body

### LINKED ISSUE #17787: in function body use `impl for`, expand to `for loop`
**rust-analyzer version**: rust-analyzer 1.80.0 (05147895 2024-07-21)

**editor or extension**: Vim

```rust
fn foo() {
    struct X;
    impl X fo { }
}
```

select completion list...

**output**:
```rust
fn foo() {
    struct X;
    impl X for  in  {
        
    } { }
}
```

**expect**:
```rust
fn foo() {
    struct X;
    impl X for  { }
}
```

rami3l:
Closes #17787.

### DIFF
--- a/crates/ide-completion/src/completions/keyword.rs
+++ b/crates/ide-completion/src/completions/keyword.rs
@@ -151,6 +151,68 @@ fn foo(a: A) { a.$0 }
     }
 
     #[test]
+    fn for_in_impl() {
+        check_edit(
+            "for",
+            r#"
+struct X;
+impl X $0 {}
+"#,
+            r#"
+struct X;
+impl X for $0 {}
+"#,
+        );
+        check_edit(
+            "for",
+            r#"
+fn foo() {
+    struct X;
+    impl X $0 {}
+}
+"#,
+            r#"
+fn foo() {
+    struct X;
+    impl X for $0 {}
+}
+"#,
+        );
+        check_edit(
+            "for",
+            r#"
+fn foo() {
+    struct X;
+    impl X $0
+}
+"#,
+            r#"
+fn foo() {
+    struct X;
+    impl X for $0
+}
+"#,
+        );
+        check_edit(
+            "for",
+            r#"
+fn foo() {
+    struct X;
+    impl X { fn bar() { $0 } }
+}
+"#,
+            r#"
+fn foo() {
+    struct X;
+    impl X { fn bar() { for $1 in $2 {
+    $0
+} } }
+}
+"#,
+        );
+    }
+
+    #[test]
     fn let_semi() {
         cov_mark::check!(let_semi);
         check_edit(
--- a/crates/ide-completion/src/context/analysis.rs
+++ b/crates/ide-completion/src/context/analysis.rs
@@ -1132,10 +1132,18 @@ fn classify_name_ref(
             ast::PathType(it) => make_path_kind_type(it.into()),
             ast::PathExpr(it) => {
                 if let Some(p) = it.syntax().parent() {
-                    if ast::ExprStmt::can_cast(p.kind()) {
-                        if let Some(kind) = inbetween_body_and_decl_check(p) {
-                            return Some(make_res(NameRefKind::Keyword(kind)));
-                        }
+                    let p_kind = p.kind();
+                    // The syntax node of interest, for which we want to check whether
+                    // it is sandwiched between an item decl signature and its body.
+                    let probe = if ast::ExprStmt::can_cast(p_kind) {
+                        Some(p)
+                    } else if ast::StmtList::can_cast(p_kind) {
+                        Some(it.syntax().clone())
+                    } else {
+                        None
+                    };
+                    if let Some(kind) = probe.and_then(inbetween_body_and_decl_check) {
+                        return Some(make_res(NameRefKind::Keyword(kind)));
                     }
                 }
 
@@ -1199,7 +1207,13 @@ fn classify_name_ref(
                     }
                 }
             },
-            ast::RecordExpr(it) => make_path_kind_expr(it.into()),
+            ast::RecordExpr(it) => {
+                // A record expression in this position is usually a result of parsing recovery, so check that
+                if let Some(kind) = inbetween_body_and_decl_check(it.syntax().clone()) {
+                    return Some(make_res(NameRefKind::Keyword(kind)));
+                }
+                make_path_kind_expr(it.into())
+            },
             _ => return None,
         }
     };


lnicola (approved):

lnicola:
The couple of failed tests I checked seem fine, they need a baseline update.

rami3l:
> The couple of failed tests I checked seem fine, they need a baseline update.

@lnicola Thanks! The fix per se looks good to me, but I wonder if this is in line with other stuff, such as:

https://github.com/rust-lang/rust-analyzer/blob/13ac53e73d030f7bdeb93a9183996ada74c7f095/crates/ide-completion/src/tests/special.rs#L1266-L1299

Do we really need all those suggestions? This PR eliminates `for` which seems reasonable to me, but what about other stuff like `loop` and `match`? Maybe `in_block_expr` is not the correct predicate here?

rami3l:
@lnicola The key issue here seems to be that `impl T for A` is local to the function, otherwise I cannot reproduce it. I'll try a smarter way of doing things.

Veykril:
We incorrectly land in a [`PathKind::Expr` here ](https://github.com/rust-lang/rust-analyzer/blob/13ac53e73d030f7bdeb93a9183996ada74c7f095/crates/ide-completion/src/context/analysis.rs#L1066)in completion analysis likely due to parser recovery

Veykril:
Given
```rs
fn foo() {
    struct X;
    impl X fo { }
}
```
we get a 
```
FN@48489..48557
  FN_KW@48489..48491 "fn"
  WHITESPACE@48491..48492 " "
  NAME@48492..48495
    IDENT@48492..48495 "foo"
  PARAM_LIST@48495..48497
    L_PAREN@48495..48496 "("
    R_PAREN@48496..48497 ")"
  WHITESPACE@48497..48498 " "
  BLOCK_EXPR@48498..48557
    STMT_LIST@48498..48557
      L_CURLY@48498..48499 "{"
      WHITESPACE@48499..48512 "\n            "
      STRUCT@48512..48521
        STRUCT_KW@48512..48518 "struct"
        WHITESPACE@48518..48519 " "
        NAME@48519..48520
          IDENT@48519..48520 "X"
        SEMICOLON@48520..48521 ";"
      WHITESPACE@48521..48534 "\n            "
      IMPL@48534..48540
        IMPL_KW@48534..48538 "impl"
        WHITESPACE@48538..48539 " "
        PATH_TYPE@48539..48540
          PATH@48539..48540
            PATH_SEGMENT@48539..48540
              NAME_REF@48539..48540
                IDENT@48539..48540 "X"
      WHITESPACE@48540..48541 " "
      RECORD_EXPR@48541..48547
        PATH@48541..48543
          PATH_SEGMENT@48541..48543
            NAME_REF@48541..48543
              IDENT@48541..48543 "fo"
        WHITESPACE@48543..48544 " "
        RECORD_EXPR_FIELD_LIST@48544..48547
          L_CURLY@48544..48545 "{"
          WHITESPACE@48545..48546 " "
          R_CURLY@48546..48547 "}"
      WHITESPACE@48547..48556 "\n        "
      R_CURLY@48556..48557 "}"
```
(there is a show syntax tree command in vscode to get this view)
as youcan see, we parse the `fo` outside of the impl but within the block expression of the function, so we parse it as an expression. From a recovery perspective this does make sense, so we should probably fix that up in the completion analysis? Not sure

rami3l:
@Veykril Thanks for your pointers! I think I've found a way to be less disruptive.

Veykril:
The proper fix will require changes in `crates\ide-completion\src\context\analysis.rs` as we land in the wrong branch somewhere (due to not accounting for parser recovery)

Veykril:
https://github.com/rust-lang/rust-analyzer/blob/13ac53e73d030f7bdeb93a9183996ada74c7f095/crates/ide-completion/src/completions/keyword.rs#L7-L31 is where we should be landing in this snippet instead of the expression completion

rami3l:
> https://github.com/rust-lang/rust-analyzer/blob/13ac53e73d030f7bdeb93a9183996ada74c7f095/crates/ide-completion/src/completions/keyword.rs#L7-L31
> 
> is where we should be landing in this snippet instead of the expression completion

But isn't it reasonable to expect everything to be "expression-like" in a `fn` context?

> https://github.com/rust-lang/rust-analyzer/blob/13ac53e73d030f7bdeb93a9183996ada74c7f095/crates/ide-completion/src/completions/keyword.rs#L7-L31
> 
> is where we should be landing in this snippet instead of the expression completion

I'm aware of this, it's just that the context of this call is:

https://github.com/rust-lang/rust-analyzer/blob/13ac53e73d030f7bdeb93a9183996ada74c7f095/crates/ide-completion/src/completions.rs#L775-L777

Where clearly it's not the same as in a `fn`.

Veykril:
If the preceding token to the cursor is an `impl Foo` then the only things thats may follow that `Foo` is a `for` token, a `where` token or a curly brace, not an expression. So the completion analysis should compute a `NameRefKind::Keyword` here, not a `NameRefKind::Path`

rami3l:
> If the preceding token to the cursor is an `impl Foo` then the only things thats may follow that `Foo` is a `for` token, a `where` token or a curly brace, not an expression. So the completion analysis should compute a `NameRefKind::Keyword` here, not a `NameRefKind::Path`

I see. I believe it works outside of functions because you special-cased it:

https://github.com/rust-lang/rust-analyzer/blob/13ac53e73d030f7bdeb93a9183996ada74c7f095/crates/ide-completion/src/context/analysis.rs#L1157-L1161

Veykril:
Yes thats the special case handling it outside of bodies, would be neat if that can be re-used though its probably a bit more tricky

rami3l:
> Yes thats the special case handling it outside of bodies, would be neat if that can be re-used though its probably a bit more tricky

So by adding the same hack to records in https://github.com/rust-lang/rust-analyzer/pull/18005/commits/3d13c7fdb0891487f35e66d9c53938a893e55be3, I was able to address the original issue. However, I'm still more interested in the `fn foo() { impl X for $0 }` case.

Veykril:
What exaclty is with `fn foo() { impl X for $0 }`?

rami3l:
> What exaclty is with `fn foo() { impl X for $0 }`?

The `inbetween_body_and_decl_check()` hack on records has worked with `fn foo() { impl X for $0 {} }` because `$0 {}` was resolved to a record, according to your first example in https://github.com/rust-lang/rust-analyzer/pull/18005#issuecomment-2320962941. However I'm not sure how to handle the case without `{}`.

Veykril:
Isn't that this case? https://github.com/rust-lang/rust-analyzer/pull/18005/files#diff-9cef2bb04b7b1bc7d74e0d5822c991876dec30dce47004692a99369f18651442R166-R180
I'm sorry I'm confused, `fn foo() { impl X for $0 {} }` already works fine today for me, your change ought to make `fn foo() { impl X fo$0 {} }` / `fn foo() { impl X $0 {} }` work from what I can tell

rami3l:
> I'm sorry I'm confused, `fn foo() { impl X for $0 {} }` already works fine today for me, your change ought to make `fn foo() { impl X fo$0 {} }` / `fn foo() { impl X $0 {} }` work from what I can tell

@Veykril No worries. I believe now I've fixed both of them.

Veykril:
Thanks!
@bors r+

PR #18031, opened 2024-09-02
feat: Suggest name in completion for let_stmt and fn_param

### LINKED ISSUE #17780: Complete type name as local name in pattern position
```rs
struct Directory {}
let $0 = Directory {};
```
I feel like proposing `directory` would be quite helpful at times.

That is, when completing a (top level) pattern binding and the corresponding type is an ADT (after stripping references), we should propose a lower snake cased version ADTs name.

roife:
fix #17780

1. Refactor: move `ide_assist::utils::suggest_name` to `ide-db::syntax_helpers::suggest_name` for reuse.
2. When completing `IdentPat`, detecte if the current node is a `let_stmt` or `fn_param`, and suggesting a new name based on the context.

### DIFF
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -1,4 +1,5 @@
 use hir::TypeInfo;
+use ide_db::syntax_helpers::suggest_name;
 use syntax::{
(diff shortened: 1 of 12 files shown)


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #18032, opened 2024-09-02
internal: Add preliminary `SyntaxEditor` functionality

### LINKED ISSUE #9649: Investigate syntax tree editing via node tracking
* https://github.com/dotnet/roslyn/blob/28191eef78568088a332a435dcd734fad1bd4fbf/src/Compilers/Core/Portable/Syntax/SyntaxNodeExtensions_Tracking.cs
* https://joshvarty.com/2015/09/18/learn-roslyn-now-part-13-keeping-track-of-syntax-nodes-with-syntax-annotations/
* https://github.com/dotnet/roslyn/issues/28378

I wonder if that's actually better than mutable-immutable syntax trees :thinking:

### LINKED ISSUE #14921: Edits in assists are prone to panic
Immediate problem: In the following code, all but the last either panic or produce bogus code. The last one had also panicked before #12789.

<details>
<summary>Code</summary>

```rust
// inline into all callers (bogus output)
fn inline_me/*$0*/(a: i32) -> i32 { a }
fn test1() {
    inline_me(inline_me(0));
}

// inline type alias (panic)
type Alias/*$0*/<T> = Vec<T>;
fn test2(_: Alias<Alias<i32>>) {}

// convert to named struct (panic)
struct Tuple/*$0*/<T>(T);
fn test3() {
    let _ = Tuple(Tuple(0));
}

// convert to tuple struct (panic)
struct Named/*$0*/<T> { inner: T }
fn test4() {
    let _ = Named {
        inner: Named {
            inner: 0,
        }
    };
}

// remove unused parameter (handles correctly!)
fn unused(a/*$0*/: i32) -> i32 { 0 }
fn test5() {
    unused(unused(5));
}
```
</details>

Problem: [`SourceChangeBuilder`][scb] is easy to use, but also easy to misuse when the assist operates on self-nestable nodes (e.g. record expressions can be embedded into another record expression) because there is an invariant for the edits: [each range of the edits MUST NOT overlap with each other][ref-invariant].

We can avoid panics by:
1. retrieve all nodes/ranges to modify
2. check if one is contained in another, build a forest whose each tree represent disjoint nodes
3. operate on nodes in the order of innermost to outermost
4. turn the modifications into text edits

(This is just an example generic procedure that should work for all cases, there may be better/more efficient procedures depending on the assist)

Currently we should implement this for every assist that may operate on overlappable nodes, but manually doing so is extremely error-prone. There should be another API that wraps `SourceChangeBuilder` (or anything needed) and does things correctly without much care from its users.

Feel free to change the issue title, I'm obviously not the best person to come up with a good title.

[scb]: https://github.com/rust-lang/rust-analyzer/blob/505fd09f9e020b096d014e68b667268e743c2dd6/crates/ide-db/src/source_change.rs#L94
[ref-invariant]: https://github.com/rust-lang/rust-analyzer/blob/505fd09f9e020b096d014e68b667268e743c2dd6/crates/text-edit/src/lib.rs#L15

DropDemBits:
Related to #15710

Implements a `SyntaxEditor` interface to abstract over the details of modifying syntax trees, to both simplify creating new code fixes and code actions, as well as start on the path of getting rid of mutable syntax nodes.

`SyntaxEditor` relies on `SyntaxMappingBuilder`s to feed in the correct information to map AST nodes created by `make` constructors, as `make` constructors do not guarantee that node identity is preserved. This is to paper over the fact that `make` constructors simply re-parse text input instead of building AST nodes from the ground up and re-using the provided syntax nodes.

`SyntaxAnnotation`s are used to find where syntax elements have ended up after edits are applied. This is primarily useful for the `add_{placeholder,tabstop}` set of methods on `SourceChangeBuilder`, as that currently relies on the nodes provided being in the final syntax tree.

Eventually, the goal should be to move this into the `rowan` crate when we move away from mutable syntax nodes, but for now it'll stay in the `syntax` crate.

---

Closes #14921 as `SyntaxEditor` ensures that all replace changes are disjoint
Closes #9649 by implementing `SyntaxAnnotation`s

### DIFF
--- a/crates/syntax/src/lib.rs
+++ b/crates/syntax/src/lib.rs
@@ -40,6 +40,7 @@ pub mod ast;
 #[doc(hidden)]
 pub mod fuzz;
 pub mod hacks;
+pub mod syntax_editor;
 pub mod ted;
 pub mod utils;
 
--- a/crates/syntax/src/syntax_editor.rs
+++ b/crates/syntax/src/syntax_editor.rs
@@ -61,16 +61,28 @@ pub struct SyntaxEdit {
 }
 
 impl SyntaxEdit {
+    /// Root of the modified syntax tree
     pub fn root(&self) -> &SyntaxNode {
         &self.root
     }
 
+    /// Which syntax elements in the modified syntax tree were modified as part
+    /// of the edit.
+    ///
+    /// Note that for syntax nodes, only the upper-most parent of a set of
+    /// changes is included, not any child elements that may have been modified.
     pub fn changed_elements(&self) -> &[SyntaxElement] {
         self.changed_elements.as_slice()
     }
 
-    pub fn find_annotation(&self, annotation: SyntaxAnnotation) -> Option<&[SyntaxElement]> {
-        self.annotations.get(&annotation).as_ref().map(|it| it.as_slice())
+    /// Finds which syntax elements have been annotated with the given
+    /// annotation.
+    ///
+    /// Note that an annotation might not appear in the modified syntax tree if
+    /// the syntax elements that were annotated did not make it into the final
+    /// syntax tree.
+    pub fn find_annotation(&self, annotation: SyntaxAnnotation) -> &[SyntaxElement] {
+        self.annotations.get(&annotation).as_ref().map_or(&[], |it| it.as_slice())
     }
 }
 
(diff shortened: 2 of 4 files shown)


Veykril:
> Eventually, the goal should be to move this into the rowan crate when we move away from mutable syntax nodes, but for now it'll stay in the syntax crate.

We should also consider forking rowan in-tree as we will likely start changing things a bunch once we consider dropping mutable syntax trees. (and at some point feed the changes back to main rowan and adopt it into the monorepo)

Veykril:
> This is to paper over the fact that make constructors simply re-parse text input instead of building AST nodes from the ground up and re-using the provided syntax nodes.

Presumably, fixing this would simplify things here? Would be good to check if if thats a requirement we want for our next rowan and whether it has certain requirements on the implementation then

DropDemBits:
> Presumably, fixing this would simplify things here?
Thinking about it more, I think we'll still need `SyntaxMapping` especially for when we move to immutable syntax trees, since reusing the syntax nodes to preserve node identity that way would imply that we mutate that syntax node.

DropDemBits:
I believe functionality-wise `SyntaxEditor` is complete, so marking as ready for review.

davidbarsky (approved):
Honestly, looks good to me: I'd like to land this and start moving assists over to this.

PR #18036, opened 2024-09-03
Bump `smol_str`

Veykril:


Cargo.lock
```diff
@@ -96,6 +96,15 @@ version = "2.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b048fb63fd8b5923fc5aa7b340d8e156aec7ec02f0c78fa8a6ddc2613f6f71de"
 
+[[package]]
+name = "borsh"
```
Veykril, line 100: very confusing that that pops up in here given its optional (and disabled, I checked this). Guess a quirk of `?` feature flags

PR #18038, opened 2024-09-03
feat: generate names for tuple-struct in add-missing-match-arms

### LINKED ISSUE #18034: Fill match arm assist should prefill binding names for tuple struct patterns
For
```rs
pub enum SmolStrBuilder {
    Inline { len: usize, buf: [u8; INLINE_CAP] },
    Heap(String),
}
```
currently fills as 
```rs

        match value {
            SmolStrBuilder::Inline { len, buf } => todo!(),
            SmolStrBuilder::Heap(_) => todo!(),
        }
```
We can do better for the latter arm, for single field tuple structs/variants we should use the variant name I feel like: `SmolStrBuilder::Heap(heap)`, though that might turn odd with things like `Option`. A better one would probably be to just fill the names by their types, so SmolStrBuilder::Heap(string)` in this case (and for any arity of fields that is).

cc @roife might be a nice follow up for you given you basically did this for completions earlier :^)

roife:
fix #18034.

This PR includes the following enhancement:

- Introduced a `NameGenerator` in `suggest_name`, which implements an automatic renaming algorithm to avoid name conflicts. Here are a few examples:

```rust
let mut generator = NameGenerator::new();
assert_eq!(generator.suggest_name("a"), "a");
assert_eq!(generator.suggest_name("a"), "a1");
assert_eq!(generator.suggest_name("a"), "a2");

assert_eq!(generator.suggest_name("b"), "b");
assert_eq!(generator.suggest_name("b"), "b1");
assert_eq!(generator.suggest_name("b2"), "b2");
assert_eq!(generator.suggest_name("b"), "b3");
assert_eq!(generator.suggest_name("b"), "b4");
assert_eq!(generator.suggest_name("b3"), "b5");
```

- Updated existing testcases in ide-assists for the new `NameGenerator` (only modified generated names).
- Generate names for tuple structs instead of using wildcard patterns in `add-missing-match-arms`.

### DIFF
--- a/crates/ide-assists/src/handlers/generate_delegate_trait.rs
+++ b/crates/ide-assists/src/handlers/generate_delegate_trait.rs
@@ -584,7 +584,7 @@ fn resolve_name_conflicts(
 
             for old_strukt_param in old_strukt_params.generic_params() {
                 // Get old name from `strukt`
-                let mut name = SmolStr::from(match &old_strukt_param {
+                let name = SmolStr::from(match &old_strukt_param {
                     ast::GenericParam::ConstParam(c) => c.name()?.to_string(),
                     ast::GenericParam::LifetimeParam(l) => {
                         l.lifetime()?.lifetime_ident_token()?.to_string()
@@ -593,8 +593,19 @@ fn resolve_name_conflicts(
                 });
 
                 // The new name cannot be conflicted with generics in trait, and the renamed names.
-                name = suggest_name::for_unique_generic_name(&name, old_impl_params);
-                name = suggest_name::for_unique_generic_name(&name, &params);
+                let param_list_to_names = |param_list: &GenericParamList| {
+                    param_list.generic_params().flat_map(|param| match param {
+                        ast::GenericParam::TypeParam(t) => t.name().map(|name| name.to_string()),
+                        p => Some(p.to_string()),
+                    })
+                };
+                let existing_names = param_list_to_names(old_impl_params)
+                    .chain(param_list_to_names(&params))
+                    .collect_vec();
+                let mut name_generator = suggest_name::NameGenerator::new_with_names(
+                    existing_names.iter().map(|s| s.as_str()),
+                );
+                let name = name_generator.suggest_name(&name);
                 match old_strukt_param {
                     ast::GenericParam::ConstParam(c) => {
                         if let Some(const_ty) = c.ty() {
@@ -1213,9 +1224,9 @@ struct S<T> {
     b : B<T>,
 }
 
-impl<T0> Trait<T0> for S<T0> {
-    fn f(&self, a: T0) -> T0 {
-        <B<T0> as Trait<T0>>::f(&self.b, a)
+impl<T1> Trait<T1> for S<T1> {
+    fn f(&self, a: T1) -> T1 {
+        <B<T1> as Trait<T1>>::f(&self.b, a)
     }
 }
 "#,
@@ -1527,12 +1538,12 @@ where
     b : B<T, T1>,
 }
 
-impl<T, T2, T10> Trait<T> for S<T2, T10>
+impl<T, T2, T3> Trait<T> for S<T2, T3>
 where
-    T10: AnotherTrait
+    T3: AnotherTrait
 {
     fn f(&self, a: T) -> T {
-        <B<T2, T10> as Trait<T>>::f(&self.b, a)
+        <B<T2, T3> as Trait<T>>::f(&self.b, a)
     }
 }"#,
         );
@@ -1589,12 +1600,12 @@ where
     b : B<T>,
 }
 
-impl<T, T0> Trait<T> for S<T0>
+impl<T, T2> Trait<T> for S<T2>
 where
-    T0: AnotherTrait
+    T2: AnotherTrait
 {
     fn f(&self, a: T) -> T {
-        <B<T0> as Trait<T>>::f(&self.b, a)
+        <B<T2> as Trait<T>>::f(&self.b, a)
     }
 }"#,
         );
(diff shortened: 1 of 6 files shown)


roife:
There’s another issue: if a position in the tuple is a generic parameter, it currently gets filled with an `_`. In this case, do we have to convert it back to the AST to get the name?

crates/ide-db/src/syntax_helpers/suggest_name.rs
```diff
@@ -62,71 +64,131 @@ const USELESS_METHODS: &[&str] = &[
...
+    /// it will try to resolve the conflict by adding a numeric suffix.
+    pub fn suggest_name(&mut self, name: &str) -> SmolStr {
+        let (prefix, suffix) = Self::split_numeric_suffix(name);
+        let prefix = SmolStr::new(prefix);
+        let suffix = suffix.unwrap_or(0);
+
+        match self.pool.entry(prefix.clone()) {
+            Entry::Vacant(entry) => {
+                entry.insert(suffix);
+                SmolStr::from_str(name).unwrap()
```
Veykril, line 113:
```suggestion
                SmolStr::new(name)
```

crates/ide-db/src/syntax_helpers/suggest_name.rs
```diff
@@ -62,71 +64,131 @@ const USELESS_METHODS: &[&str] = &[
...
+            }
+            Entry::Occupied(mut entry) => {
+                let count = entry.get_mut();
+                *count = (*count + 1).max(suffix);
+
+                let mut new_name = SmolStrBuilder::new();
+                new_name.push_str(&prefix);
+                new_name.push_str(count.to_string().as_str());
+                new_name.finish()
```
Veykril, line 122:
```suggestion
                format_smolstr!("{prefix}{count}")
```

crates/ide-db/src/syntax_helpers/suggest_name.rs
```diff
@@ -62,71 +64,131 @@ const USELESS_METHODS: &[&str] = &[
...
+    /// Remove the numeric suffix from the name
+    ///
+    /// # Examples
+    /// `a1b2c3` -> `a1b2c`
+    fn split_numeric_suffix(name: &str) -> (&str, Option<usize>) {
+        let pos =
+            name.rfind(|c: char| !c.is_numeric()).expect("Name cannot be empty or all-numeric");
+        let (prefix, suffix) = name.split_at(pos + 1);
+        (prefix, suffix.parse().ok())
+    }
```
Veykril, line 191:
```suggestion
    fn split_numeric_suffix(name: &str) -> (SmolStr, Option<usize>) {
        let pos =
            name.rfind(|c: char| !c.is_numeric()).expect("Name cannot be empty or all-numeric");
        let (prefix, suffix) = name.split_at(pos + 1);
        (SmolStr::new(prefix), suffix.parse().ok())
    }
```

all callers transform the return type like that, so we can do that here and re-use the smolstr in the `Vacant` case by cloning it which is rather cheap

crates/ide-assists/src/handlers/generate_delegate_trait.rs
```diff
@@ -1589,12 +1600,12 @@ where
...
-impl<T, T0> Trait<T> for S<T0>
+impl<T, T2> Trait<T> for S<T2>
 where
-    T0: AnotherTrait
+    T2: AnotherTrait
 {
     fn f(&self, a: T) -> T {
-        <B<T0> as Trait<T>>::f(&self.b, a)
+        <B<T2> as Trait<T>>::f(&self.b, a)
     }
```
Veykril, line 1609: These should be `T1` not `T2` no? 🤔

crates/ide-assists/src/handlers/generate_delegate_trait.rs
```diff
@@ -1213,9 +1224,9 @@ struct S<T> {
     b : B<T>,
 }
 
-impl<T0> Trait<T0> for S<T0> {
-    fn f(&self, a: T0) -> T0 {
-        <B<T0> as Trait<T0>>::f(&self.b, a)
+impl<T1> Trait<T1> for S<T1> {
+    fn f(&self, a: T1) -> T1 {
+        <B<T1> as Trait<T1>>::f(&self.b, a)
```
Veykril, line 1229: These should be `T`, not `T1` imo

crates/ide-db/src/syntax_helpers/suggest_name.rs
```diff
@@ -62,71 +64,131 @@ const USELESS_METHODS: &[&str] = &[
...
+    /// Create a new empty generator
+    pub fn new() -> Self {
+        Self { pool: FxHashMap::default() }
+    }
+
+    /// Create a new generator with existing names. When suggesting a name, it will
+    /// avoid conflicts with existing names.
+    pub fn new_with_names<'a>(existing_names: impl Iterator<Item = &'a str>) -> Self {
+        let mut generator = Self::new();
+        existing_names.for_each(|name| generator.insert(name));
```
Veykril, line 99: This will increase the counters if `existing_names` has duplicates, it should only keep the counter to the max suffix it sees in the input

roife: > it should only keep the counter to the max suffix it sees in the input

It is exactly what the code does. Given existing names `("a", "a")`, it would return `a1` instead of `a2`.

crates/ide-assists/src/handlers/generate_delegate_trait.rs
roife: Perhaps I should refactor the logic of renaming in `generate_delegate_trait`. 🙊

crates/ide-db/src/syntax_helpers/suggest_name.rs
Veykril: Ah right, I misunderstood due to the `generate_delegate_trait` bit

PR #18041, opened 2024-09-03
feat: better name suggestions for fn

### LINKED ISSUE #17631: Improve automatic name selection for "Extract into variable" on `from()` calls
```rust
// Stubs for demo
struct Foo {}
struct Bar {}
impl From<Foo> for Bar {
    fn from(_: Foo) -> Self {
        Bar {}
    }
}
fn do_something(_: Bar) {}

// Code of interest
fn main() {
    let foo = Foo {};
    do_something(Bar::from(foo));
    //           ^^^^^^^^^^^^^^ select this and extract into variable
}
```

The result of extraction is:

```rust
    let from = Bar::from(foo);
    do_something(from);
```

A better result would be:

```rust
    let bar = Bar::from(foo);
    do_something(bar);
```

`.into()` calls also currently produce variables named `into`, but that might be more trouble to fix; I don't know.

rust-analyzer version: 0.3.2037-standalone (e9afba57a 2024-07-14)

roife:
fix #17631.

Better name suggestions for fn-calls / method-calls in the form of `from()`, `from_xxx()`, `into()`, etc.

### DIFF
--- a/crates/ide-db/src/syntax_helpers/suggest_name.rs
+++ b/crates/ide-db/src/syntax_helpers/suggest_name.rs
@@ -18,7 +18,9 @@ const USELESS_TRAITS: &[&str] = &["Send", "Sync", "Copy", "Clone", "Eq", "Partia
 ///
 /// **NOTE**: they all must be snake lower case
 const USELESS_NAMES: &[&str] =
-    &["new", "default", "option", "some", "none", "ok", "err", "str", "string"];
+    &["new", "default", "option", "some", "none", "ok", "err", "str", "string", "from", "into"];
+
+const USELESS_NAME_PREFIXES: &[&str] = &["from_", "with_", "into_"];
 
 /// Generic types replaced by their first argument
 ///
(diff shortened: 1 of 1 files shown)


crates/ide-db/src/syntax_helpers/suggest_name.rs
```diff
@@ -18,7 +18,9 @@ const USELESS_TRAITS: &[&str] = &["Send", "Sync", "Copy", "Clone", "Eq", "Partia
 ///
 /// **NOTE**: they all must be snake lower case
 const USELESS_NAMES: &[&str] =
-    &["new", "default", "option", "some", "none", "ok", "err", "str", "string"];
+    &["new", "default", "option", "some", "none", "ok", "err", "str", "string", "from", "into"];
+
+const USELESS_NAME_PREFIX: &[&str] = &["from_", "with_", "into_"];
```
rami3l, line 23:
```suggestion
const USELESS_NAME_PREFIXES: &[&str] = &["from_", "with_", "into_"];
```

PR #18043, opened 2024-09-04
Configure flycheck using discovered using [workspace.discoverConfig]

cormacrelf:
So, quick recap:

1. We have workspace.discoverConfig


crates/rust-analyzer/src/handlers/notification.rs
```diff
@@ -352,20 +353,30 @@ fn run_flycheck(state: &mut GlobalState, vfs_path: VfsPath) -> bool {
...
-                    } => cargo.packages().find_map(|pkg| {
-                        let has_target_with_root = cargo[pkg]
-                            .targets
-                            .iter()
-                            .any(|&it| crate_root_paths.contains(&cargo[it].root.as_path()));
-                        has_target_with_root.then(|| cargo[pkg].name.clone())
-                    }),
+                    } => {
+                        // Iterate crate_root_paths first because it is in topological
+                        // order, and we can therefore find the actual crate your saved
```
cormacrelf, line 358: Ah, I believe this was true when I first wrote it, but this appears to be no longer the case.

cormacrelf: Now the reverse deps function is topological.

crates/rust-analyzer/src/flycheck.rs
```diff
@@ -232,13 +367,62 @@ enum FlycheckStatus {
...
+
+struct Substitutions<'a> {
+    label: Option<&'a str>,
+    saved_file: Option<&'a str>,
+}
+
+impl<'a> Substitutions<'a> {
+    /// If you have a runnable, and it has {label} in it somewhere, treat it as a template that
+    /// may be unsatisfied if you do not provide a label to substitute into it. Returns None in
+    /// that situation. Otherwise performs the requested substitutions.
```
davidbarsky, line 382:
```suggestion
    /// A basic template engine for runnables.
    ///
    /// If a runnable has the string `{label}`, the whole runnable will be treated as a template.
```

crates/rust-analyzer/src/flycheck.rs
```diff
@@ -232,13 +367,62 @@ enum FlycheckStatus {
     Finished,
 }
 
-pub(crate) const SAVED_FILE_PLACEHOLDER: &str = "$saved_file";
+/// This is stable behaviour. Don't change.
+const SAVED_FILE_PLACEHOLDER: &str = "$saved_file";
```
davidbarsky, line 371:
```suggestion
const SAVED_FILE_PLACEHOLDER_BAD: &str = "$saved_file";
const SAVED_FILE_PLACEHOLDER: &str = "{saved_file}";
```

crates/rust-analyzer/src/flycheck.rs
```diff
@@ -232,13 +367,62 @@ enum FlycheckStatus {
     Finished,
 }
 
-pub(crate) const SAVED_FILE_PLACEHOLDER: &str = "$saved_file";
+/// This is stable behaviour. Don't change.
```
davidbarsky, line 370: fwiw, we can remove `$saved_file` with a bit of notice.

davidbarsky (changes requested):
some assorted comments. I'll do another pass later.

crates/rust-analyzer/src/flycheck.rs
davidbarsky: actually, no, I lied we need to keep this for while. the main user of this is Fuchsia and until they're using Bazel, I don't think it's feasible to remove `$saved_file`.

davidbarsky: (cc: @P1n3appl3)

P1n3appl3: thanks for the ping. We'd actually be ok with removing `$saved_file` if we could use `{label}` in the flycheck override. Today we actually do that mapping ourselves (wrapper script checks a global file list to find all labels that include the saved file and runs a check build/clippy on all those in the build system).

I'd be happy to migrate us over to label replacement when this lands.

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -951,8 +951,24 @@ impl GlobalState {
             FlycheckMessage::ClearDiagnostics { id } => self.diagnostics.clear_check(id),
 
             FlycheckMessage::Progress { id, progress } => {
+                let format_with_id = |user_facing_command: String| {
+                    if self.flycheck.len() == 1 {
+                        user_facing_command
+                    } else {
+                        format!("{user_facing_command} (#{})", id + 1)
```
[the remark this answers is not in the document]
cormacrelf: With crates A, dependents B and C, if you open B/lib.rs and then C/lib.rs you will get two separate workspaces, as C isn't a dependency of B, so it does discovery again for C. Then you have two flycheck handles, and saving A/lib.rs will trigger both with the same label A. Buck doesn't care, one will just wait for the other, but that is two flychecks with the same command line.

crates/rust-analyzer/src/flycheck.rs
```diff
@@ -69,6 +119,28 @@ impl CargoOptions {
     }
 }
 
+/// The flycheck config from a rust-project.json file
+#[derive(Debug, Default)]
+pub(crate) struct FlycheckConfigJson {
+    // XXX: unimplemented because not all that important> nobody
+    // doing custom rust-project.json needs it most likely
+    // pub workspace_template: Option<project_json::Runnable>,
```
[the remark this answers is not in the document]
cormacrelf: Handle all the flychecks where we would normally run `cargo check --workspace`. Currently we pass these requests to the flycheck handle, but always end up doing nothing, because the saved file is None and there is no label to interpolate.

rust-project doesn't have anything to run for this, but it is not that hard to add. I have some simple BXL that builds and prints out the diag.txt/json for a target and all its dependencies, and frequently run this on `//...`. The only problem with it is that the BXL is slow at unifying all the deps of all the targets and always stutters for a few seconds doing "graph computations", but I'm probably doing something dumb. Other than that works exactly like cargo check.

crates/rust-analyzer/src/flycheck.rs
```diff
@@ -105,36 +177,60 @@ pub(crate) struct FlycheckHandle {
     sender: Sender<StateChange>,
     _thread: stdx::thread::JoinHandle,
     id: usize,
+
+    /// Bit hacky, but this lets us force the use of restart_for_package when the flycheck
+    /// configuration does not support restart_workspace.
+    cannot_run_workspace: bool,
```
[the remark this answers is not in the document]
cormacrelf: If I didn't add this, then if check.workspace was enabled (as it is by default), then we would never run any ProjectJson.runnables flychecks if they included `{label}`. The overall effect here is that having `{label}` in your flycheck template implies check.workspace = false.

The concept was that the discovery tool should be the only thing you need to configure. I spent a while trying to track down why nothing was working only to find I had deleted my check.workspace config -- and I am working on this code!

crates/rust-analyzer/src/flycheck.rs
```diff
@@ -403,19 +651,26 @@ impl FlycheckActor {
...
+                        if let Some(tgt) = bin_target {
+                            tgt.append_cargo_arg(&mut cmd);
+                        }
+                    }
+                    PackageToRestart::Package(PackageSpecifier::BuildInfo { label: _ }) => {
+                        // No way to flycheck this single package. All we have is a build label.
+                        // There's no way to really say whether this build label happens to be
+                        // a cargo canonical name, so we won't try.
+                        return None;
+                    }
```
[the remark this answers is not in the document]
cormacrelf: We can rename it Automatic. Because it is! In cargo workspaces it does cargo stuff; in json workspaces it does JSON stuff.

crates/base-db/src/input.rs
```diff
@@ -446,15 +451,16 @@ impl CrateGraph {
...
         while let Some(krate) = worklist.pop() {
             if let Some(krate_rev_deps) = inverted_graph.get(&krate) {
-                krate_rev_deps
-                    .iter()
-                    .copied()
-                    .filter(|&rev_dep| rev_deps.insert(rev_dep))
-                    .for_each(|rev_dep| worklist.push(rev_dep));
+                krate_rev_deps.iter().copied().filter(|&rev_dep| visited.insert(rev_dep)).for_each(
+                    |rev_dep| {
+                        rev_deps.push(rev_dep);
```
cormacrelf, line 456: Ayyy, that is a very sus line in retrospect. This needs tests.

davidbarsky:
Just to confirm my understanding: when you say:

> `Allows the `discoverConfig` to describe how to flycheck crates, using a new runnable kind of "flycheck".`

The code changes here seem to imply that it'd be relying on the runnables in a `rust-project.json`; not that you're adding a few field to the configuration object/value. Is that correct? Do you intend to add a new field to `discoverConfig`...?

cormacrelf:
Correct. However now that I think about it, the runnable templating would be a useful thing to expose to end users to override. They can then have custom test commands, debug commands, etc.

(Noting that Meta's internal test harness has different flags to the OSS one. Would we want to have another configuration layer for rust-project to pick up and change its output? I would hope not!)

davidbarsky:
> Correct. 

Sorry, I realized I phrased my question ambiguously: are you planning on adding another field to `workspace.discoverConfig` for flycheck-specific? If the answer is yes, I'd push back on that and ask for a more wholistic design of flycheck's configuration: it's pretty messy and I don't want to make it worse.

> However now that I think about it, the runnable templating would be a useful thing to expose to end users to override. They can then have custom test commands, debug commands, etc.

I agree! I think it's be nice to expose that via the standard rust-analyzer configuration mechanism.

> Noting that Meta's internal test harness has different flags to the OSS one. 

Oi, I didn't realize that fact, but good to know. Lemme see if I can try and unify the two.

> Would we want to have another configuration layer for rust-project to pick up and change its output? I would hope not!

We should probably move this discussion over to the buck2 repo, but I've wanted a `rust-project.toml` for _ages_. There a lot of stuff in rust-project that's `#[cfg(fbcode)]`'d that I'd want to be controlled through configuration. It also means that setting up Buck with rust-analyzer could be as simple as placing a predefined `.rust-project.toml` and `.rust-analyzer.toml` at the root of a Buck workspace. Heck, we could even make the `buck2 init` command do that.

cormacrelf:
Well, I've been using it with my team pretty comfortably for 8 months. Will soon be upgrading a bunch of systems and will have to rebase this anyway. I will keep this PR updated when I do.

cormacrelf:
@Veykril this is rebased onto master. I actually found during rebasing that a number of the things I did have been done independently in other PRs, just slightly differently. So the diff is a little smaller.

To remind you:

- Adds a new runnable kind `"flycheck"` / RunnableKind::Flycheck. Must be distinct from check (check runnable is for human-readable output)
- Substitutes `{label}` or `$saved_file` in it
- Abstracts over cargo vs projectjson flycheck sources
- Uses all this to flycheck

This lets your 3rd party build tool configure how RA will flycheck.

Veykril:
I'll try to take a look this week.

crates/base-db/src/input.rs
cormacrelf: Actually this isn't needed any more, I will remove the commit

crates/rust-analyzer/src/flycheck.rs
cormacrelf: Done, both syntaxes are supported now. Tests added.

crates/rust-analyzer/src/flycheck.rs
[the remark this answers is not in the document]
cormacrelf: Renamed to Automatic, it is a bit clearer.

crates/rust-analyzer/src/flycheck.rs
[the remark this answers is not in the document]
cormacrelf: No longer needed

crates/rust-analyzer/src/flycheck.rs
cormacrelf: I'm keeping it because I think it's still useful -- a build system might know that there are multiple crates that use a particular file and compile both. $saved_file is still there but prefer `{saved_file}` to match all other command interpolations.

cormacrelf:
I also added a pretty big docs overhaul for non-cargo build systems. I recommend reviewing this commit-by-commit, as there is a bit of diff noise now due to some renames etc.

crates/rust-analyzer/src/flycheck.rs
```diff
@@ -348,14 +423,153 @@ enum Event {
...
+            }
+            cmd.arg(arg);
+        }
+        cmd.current_dir(&template.cwd);
+        Some(cmd)
+    }
+}
+
+#[test]
+fn test_substitutions() {
```
Veykril: nit lets move this to the bottom of the file, I don't like tests living in the middle of a file

cormacrelf: Done, now cfg(test) mod tests {}.

Veykril:
Just gotta make clippy happy now and then I am fine with merging this 👍

cormacrelf:
Clippy issues were in hir-ty, done.

cormacrelf:
Ripper, thanks a ton mate

Wilfred:
Nice work! I particularly appreciate the docs polish too :)

We should look at updating rust-project to use this new interpolation syntax too.

PR #18050, opened 2024-09-04
assist: ensure `replace_qualified_name_with_use` applies to the first path segment

davidbarsky:
This change helps a bit with the discoverability of `replace_qualified_name_with_use`. Specifically, it ensures that a cursor on the first path segment (e.g., `$0std::fmt::Debug`, where `$0` is the cursor) would result in an import along the lines of `use std::fmt;` and `fmt::Debug;` at the usage sites.

### DIFF
--- a/crates/ide-assists/src/handlers/replace_qualified_name_with_use.rs
+++ b/crates/ide-assists/src/handlers/replace_qualified_name_with_use.rs
@@ -29,7 +29,7 @@ pub(crate) fn replace_qualified_name_with_use(
     acc: &mut Assists,
     ctx: &AssistContext<'_>,
 ) -> Option<()> {
-    let original_path: ast::Path = ctx.find_node_at_offset()?;
+    let mut original_path: ast::Path = ctx.find_node_at_offset()?;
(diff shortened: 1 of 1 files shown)


Veykril (approved):

Veykril:
Wait what just happened here?

Veykril:
That was a fun ride fixing this whew
@bors r+

lnicola:
Fun thing: the changes here and the bors merge commit show up twice in the git history (9d74b5f2649983c11ad746765cbc18f176cb75f6, 7048534d2d40bdec06f74921d365ae1470d91243).

Veykril:
How 🤔 One of those is also not associated with a PR, do we not have our main branch protected?

lnicola:
Bors glitched? I think it pushed to master twice.

davidbarsky:
I think I pushed a branch to `rust-lang/rust-analyzer`, not my fork of it (`davidbarsky/rust-analyzer`). Sorry, I'll change my jj config to push to `davidbarsky/rust-analyzer` to make this more difficult to happen!

Issue #18047, opened 2024-09-04
False positive `non-primitive cast` error in recent nightly


frxstrem:
When working for a while in a large codebase, I get a lot of errors like this one from `tracing::info!` macro invocations:

> non-primitive cast: `&Arguments<_>` as `&dyn Value`

This seems to come from an expression `&(format_args!(...)) as &dyn Value` in the macro expansion, which should work (and works according to `rustc`). `std::fmt::Arguments<'_>` implements `tracing_core::field::Value`, however rust-analyzer seems to in some cases still produce an error.

I've so far been unable to create a minimal reproducible example, I've only so far seen this in a fairly large (170k LOC) commercial codebase that I'm working with. Restarting rust-analyzer does make the error go away for a while, in my case for about 1.5 minutes after other errors are visible.

```rust
let _ = &format_args!("...") as &dyn std::io::Write; // <-- immediately has error
let _ = &format_args!("...") as &dyn tracing::Value; // <-- has an error after a while
```

The errors started showing up after updating the rust-analyzer extension earlier today, not sure what the version prior to the update was. It seems like it may be related to the recently merged #17984.

---

**rust-analyzer version**: rust-analyzer version: 0.4.2098-standalone

**rustc version**: rustc 1.80.1 (3f5fd8dd4 2024-08-06)

**editor or extension**: VSCode, extension version v0.4.2098 (pre-release)

**relevant settings**: none that I can find

lnicola:
CC #11847. We might have to mark it as experimental :disappointed:.

ShoyuVanilla:
> I've only so far seen this in a fairly large (170k LOC) commercial codebase that I'm working with.

Is your codebase commercial?

ShoyuVanilla:
> CC #11847. We might have to mark it as experimental 😞.

This seems to be caused from coerce_unsize as you said. I'll try with the way rustc handles this 🤔

flodiebold:
> This seems to be caused from coerce_unsize as you said. I'll try with the way rustc handles this 🤔 

Our CoerceUnsized solving sadly isn't quite correct and probably can't be fixed before the new trait solver is integrated :/

ShoyuVanilla:
> Our CoerceUnsized solving sadly isn't quite correct and probably can't be fixed before the new trait solver is integrated :/

It might be turned out otherwise, but I guess it could be.
[...]

PR #18054, opened 2024-09-05
fix: Disable invalid cast diagnostics for casts like `&T as &dyn Trait`

ShoyuVanilla:
Hope this might suppress false positive diagnostics like #18047


ShoyuVanilla:
So many typos 😅 Will fix and squash tomorrow

HKalbasi:
Shouldn't we also suppress it for `Box<dyn T>` as well?

ShoyuVanilla:
> Shouldn't we also suppress it for `Box<dyn T>` as well?

Oh, you're right. I neglected it

ShoyuVanilla:
It seems that I misdiagnosed the cause of the issue. The test I added works fine without any modification to the logic if I just add `minicore: builtin_impls`

ShoyuVanilla:
https://github.com/rust-lang/rust-analyzer/issues/18047#issuecomment-2333017248

Issue #18021, opened 2024-09-01
Custom snippets no longer work


Veykril:
Probably broke with the r-a toml changes

PR #18057, opened 2024-09-05
internal: Better testing infra for ratoml

alibektas:
This PR makes some improvements on how we test configs that come from `rust-analyzer.toml` files.
It was primarily used to solve #18021 but along the way I could not really determine the cause of the said issue which makes me think that it may not be related to the changes that I made earlier to the ratoml infra. In either way `custom_snippets` are now made `global` because we still don't have a tree that maps a `SourceRootId` to a set of `Snippet`s.


Veykril (approved):

Veykril:
bors merge

PR #18065, opened 2024-09-06
fix: Catch panics from diagnostics computation

Veykril:


crates/rust-analyzer/src/main_loop.rs
```diff
@@ -552,23 +553,33 @@ impl GlobalState {
...
-                        );
+                        let Ok(diags) = std::panic::catch_unwind(|| {
+                            fetch_native_diagnostics(
+                                &snapshot,
+                                subscriptions.clone(),
+                                slice.clone(),
+                                NativeDiagnosticsFetchKind::Semantic,
+                            )
+                        }) else {
+                            return;
```
lnicola, line 581: Does this mean we no longer log them?

Veykril: I believe we should still see the panic as `catch_unwind` doesn't inhibit the panic handler from running.

PR #18068, opened 2024-09-06
fix: Always explicitly set `TraitRef` self types when lowering

Veykril:


crates/hir-ty/src/lower.rs
```diff
@@ -516,8 +516,11 @@ impl<'a> TyLoweringContext<'a> {
             TypeNs::TraitId(trait_) => {
                 let ty = match remaining_segments.len() {
                     1 => {
-                        let trait_ref =
-                            self.lower_trait_ref_from_resolved_path(trait_, resolved_segment, None);
+                        let trait_ref = self.lower_trait_ref_from_resolved_path(
+                            trait_,
+                            resolved_segment,
+                            TyKind::Error.intern(Interner),
```
Veykril, line 522: This should also be an IV, but we are in the ty lowering context here without access to an inference table though that is probably fine as they callers of this at some point call `insert_inference_vars` anyways

PR #18075, opened 2024-09-08
feat: render patterns in params for hovering

### LINKED ISSUE #17858: Incorrect parameter names preview when tuple destructuring is involed.
**rust-analyzer version**: `rust-analyzer 1.80.0 (05147895 2024-07-21)`

**editor or extension**: Vim ([coc-rust-analyzer](https://github.com/fannheyward/coc-rust-analyzer))

**code snippet to reproduce**: cited from #4960:
```rust
fn main() {
    let range = (3, 5);
    test_1(range);
}

fn test_1((start_range, end_range): (u32, u32)) {
    println!("{} - {}", start_range, end_range);
}
```

## Actual

The popup documentation window shows the parameter name as `_`. This information is *completely useless* and I'm always forced to jump to the definition just to see the correct parameter name.

The situation becomes far more worse when two or more tuple destructurings are involed:

```rust
pub fn unchecked_move_forward(
    &self,
    (i, j): (usize, usize),
    (h_exclusive, w_exclusive): (usize, usize),
) -> (usize, usize) {
```

In this case, I get confused which is which.

## Expected

The popup documentation window shows the parameter name as `(start_range, end_range)` or anything meaningful.

roife:
Fix #17858 

This PR introduces an option to [hir-def/src/body/pretty.rs](https://github.com/rust-lang/rust-analyzer/blob/08c7bbc2dbe4dcc8968484f1a0e1e6fe7a1d4f6d/crates/hir-def/src/body/pretty.rs) to render the result as a single line, which is then reused for rendering patterns in parameters for hovering.

### DIFF
--- a/crates/hir-def/src/body.rs
+++ b/crates/hir-def/src/body.rs
@@ -227,6 +227,17 @@ impl Body {
         pretty::print_expr_hir(db, self, owner, expr, edition)
     }
 
+    pub fn pretty_print_pat(
+        &self,
+        db: &dyn DefDatabase,
+        owner: DefWithBodyId,
+        pat: PatId,
+        oneline: bool,
+        edition: Edition,
+    ) -> String {
+        pretty::print_pat_hir(db, self, owner, pat, oneline, edition)
+    }
+
     fn new(
         db: &dyn DefDatabase,
         owner: DefWithBodyId,
--- a/crates/hir-def/src/body/pretty.rs
+++ b/crates/hir-def/src/body/pretty.rs
@@ -16,6 +16,13 @@ use crate::{
 
 use super::*;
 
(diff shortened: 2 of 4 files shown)


roife:
The only issue now is that we are rendering `fn test(Point { x, y }: Point) {}` as `fn test(Point { x: x, y: y }: Point) {}`. Can we just modify `pretty.rs`, although this might change the output of `hir_view`? 👀

Veykril:
> The only issue now is that we are rendering `fn test(Point { x, y }: Point) {}` as `fn test(Point { x: x, y: y }: Point) {}`. Can we just modify `pretty.rs`, although this might change the output of `hir_view`? 👀

Yes lets adjust that

Veykril (approved):

PR #18080, opened 2024-09-09
Remove crate graph deduplication logic

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/17748


davidbarsky:
Not opposed to the removal, but what's the thinking, if you don't mind my asking?

Veykril:
Cargo runs proc macros in the workspace directory, not the crate manifest
dir, which we need to replicate and as such all crates are unique wrt to
their workspace. Cargo also will expose this in an env var in the future
(if it doesn't already on stable)

max-sixty:
Thanks @Veykril !

PR #18093, opened 2024-09-10
Skip checks for cast to dyn traits

ShoyuVanilla:
It seems that chalk fails to solve some obvious goals when there are some recursiveness in trait environments.
And it doesn't support trait upcasting yet. rust-lang/chalk#796

This PR just skips for casting into types containing `dyn Trait` to prevent false positive diagnostics like #18047 and #18083


Veykril (approved):

Veykril:
bors merge

Veykril:
ugh
@bors r+

PR #18100, opened 2024-09-11
fix: Don't send empty inlay hints over LSP

Wilfred:
LSP states that inlay hints should never be empty:

    /**
     * The label of this hint. A human readable string or an array of
     * InlayHintLabelPart label parts. * * *Note* that neither the string nor the label part can be empty.
     */
    label: string | InlayHintLabelPart[];

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#inlayHint

rust-analyzer would sometimes return empty inlay label parts, leading to this warning in the VS Code devtools console:

> INVALID inlay hint, empty label part rust-analyzer

### DIFF
--- a/crates/rust-analyzer/src/lsp/to_proto.rs
+++ b/crates/rust-analyzer/src/lsp/to_proto.rs
@@ -557,6 +557,7 @@ fn inlay_hint_label(
             let parts = label
                 .parts
                 .into_iter()
+                .filter(|part| !part.text.is_empty())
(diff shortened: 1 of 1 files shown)


Veykril:
We should rather fix this in our inlay hint generation, an empty text part sounds non-sensical

Veykril:
https://github.com/rust-lang/rust-analyzer/pull/18107 should fix those

PR #18136, opened 2024-09-18
Don't lint names of #[no_mangle] extern fns

valadaptive:
[Rust doesn't run the `non_snake_case_name` lint on `extern fn`s with the `#[no_mangle]` attribute](https://github.com/rust-lang/rust/pull/44966).

The conditions are:
- The function must be `extern` and have a `#[no_mangle]` attribute.
- The function's ABI must not be explicitly set to "Rust".

This PR replicates that logic here.


Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #18164, opened 2024-09-22
fix: Temporary fix for `remove_unused_imports` not handling import aliases correctly

### LINKED ISSUE #18129: "Remove all unused imports" code action incorrectly removes `as` aliased imports.
**rust-analyzer version**: 0.3.2112-standalone

**rustc version**: rustc 1.80.1 (3f5fd8dd4 2024-08-06)

**editor or extension**: VScode - latest version (seems to be `v0.3.2112`)

**relevant settings**: None

**repository link (if public, optional)**: Irrelevant

```rust
// Any aliased import would really work.

use std::io::Error as IoError; // Here (read after code snippet)

fn main() -> Result<(), IoError> { // Any usage of the alias (it's now clearly used).
    Ok(())
}
```

Place your cursor at the marked spot in the code, click <kbd>ctrl + .</kbd> (or your keybind for code actions) and select "Remove all the unused imports".

ShoyuVanilla:
Fixes #18129

### DIFF
--- a/crates/ide-db/src/search.rs
+++ b/crates/ide-db/src/search.rs
@@ -19,7 +19,7 @@ use parser::SyntaxKind;
 use rustc_hash::{FxHashMap, FxHashSet};
 use span::EditionedFileId;
 use syntax::{
-    ast::{self, HasName},
+    ast::{self, HasName, Rename},
     match_ast, AstNode, AstToken, SmolStr, SyntaxElement, SyntaxNode, TextRange, TextSize,
     ToSmolStr,
 };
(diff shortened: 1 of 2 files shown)


crates/ide-db/src/search.rs
```diff
@@ -447,6 +449,11 @@ impl<'a> FindUsages<'a> {
         self
     }
 
+    pub fn with_rename(mut self, rename: Option<&'a Rename>) -> Self {
```
Veykril, line 452: This should be a temporary thing. Ideally this would trivially fix itself once we handle renamed `use`s as proper `Definition`s

ShoyuVanilla: Oh, that's right. This should be handled as TypeAlias, right? I'll add a comment and modify the commit and PR message

PR #18173, opened 2024-09-23
Rename docs/user/manual.adoc to manual.adoc

natalyag236:


lnicola:
What is the point of this?

natalyag236:
im installing rust on vscode

lnicola:
Okay, but the manual is available on rust-analyzer.github.io, and moving it will break that. Plus, there's not much reason to move it.

PR #18175, opened 2024-09-23
internal: Make COMPLETION_MARKER more explicitly r-a

Wilfred:
If a user ever sees the completion marker, it's confusing to see text about IntelliJ. Use a string that's more explicitly about completion for rust-analyzer.


Veykril (approved):

lnicola:
Oh no, not our Easter egg!

matklad:
🥲

PR #18180, opened 2024-09-24
feat: Index workspace symbols at startup rather than on the first symbol search.

kpreid:
This will eliminate potential many-second delays when performing the first search, at the price of making cache priming (“Indexing N/M” in the VS Code status bar) take a little longer in total. Hopefully this additional time is insignificant because a typical session will involve at least one symbol search.

Further improvement would be to do this as a separate parallel task (which will be beneficial if the workspace contains a small number of large crates), but that would require significant additional refactoring of the progress-reporting mechanism to understand multiple tasks per crate. Happy to tackle that in this PR if desired, but I thought I'd propose the minimal change first.

### DIFF
--- a/crates/ide-db/src/prime_caches.rs
+++ b/crates/ide-db/src/prime_caches.rs
@@ -13,6 +13,7 @@ use crate::{
         salsa::{Database, ParallelDatabase, Snapshot},
         Cancelled, CrateId, SourceDatabase, SourceRootDatabase,
     },
+    symbol_index::SymbolsDatabase,
     FxIndexMap, RootDatabase,
 };
 
@@ -54,11 +55,13 @@ pub fn parallel_prime_caches(
         let (progress_sender, progress_receiver) = crossbeam_channel::unbounded();
         let (work_sender, work_receiver) = crossbeam_channel::unbounded();
         let graph = graph.clone();
+        let local_roots = db.local_roots();
         let prime_caches_worker = move |db: Snapshot<RootDatabase>| {
             while let Ok((crate_id, crate_name)) = work_receiver.recv() {
                 progress_sender
                     .send(ParallelPrimeCacheWorkerProgress::BeginCrate { crate_id, crate_name })?;
 
+                // Compute the DefMap and possibly ImportMap
                 let file_id = graph[crate_id].root_file_id;
                 let root_id = db.file_source_root(file_id);
                 if db.source_root(root_id).is_library {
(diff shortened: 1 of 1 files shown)


crates/ide-db/src/prime_caches.rs
```diff
@@ -68,6 +71,23 @@ pub fn parallel_prime_caches(
...
+                // We do this for workspace crates only (members of local_roots), because doing it
+                // for all dependencies could be *very* unnecessarily slow in a large project.
+                //
+                // FIXME: We should do it unconditionally if the configuration is set to default to
+                // searching dependencies (rust-analyzer.workspace.symbol.search.scope), but we
+                // would need to pipe that configuration information down here.
+                //
+                // FIXME: This would complete faster if it were done in parallel as a separate
+                // work item. That will make the progress calculation more complex since there
+                // will be more than one work item per crate.
```
Veykril, line 86: it wouldn't, this depends on the definition maps which is what we are already computing here, so if we were to run this in parallel it would just block on that computation.

Veykril:
This sounds fine to me but @davidbarsky might have opinions about this given this "slows down startup"

davidbarsky:
I'll fix https://github.com/rust-lang/rust-analyzer/pull/18181 and rebase this PR against 18181, but I don't expect this to be PR to be an issue.

---

As to why I care about startup times: at work, my team has some extremely detailed telemetry concerning people's usage of rust-analyzer and we've found that people want to start using rust-analyzer immediately after opening a Rust file, so faster startup times are a good thing. However, prior changes that increased cache priming times to reduce the number of uncached items _also_ helped perceived despite nominally longer startup times.

The general tenor of written/survey feedback from people is that while they wish they didn't need to wait for rust-analyzer to start, they found that the steady-state, warm rust-analyzer is heads-and-shoulders better than any other supported language's LSP.

(I should probably write this down somewhere more permanent. Maybe a blog post?)

crates/ide-db/src/prime_caches.rs
kpreid: OK, I deleted this comment.

kpreid:
FWIW, my personal motivation for this change is broadly like:

1. I open VS Code on a project. I know this will take a while, so I do something else meanwhile. (Or maybe it just reopened after a reboot, or something of that flavor. In any case, RA has started up but has had no user interactions.)
2. I come back after it’s ready and the first thing I want to do on the project is navigate to a symbol (or a file, because I have VS Code set to mix file and symbol search) and I have to wait an eternal ~10 seconds for the indexing.
3. If I go do something else while I wait for *that*, then the VS Code search popup will vanish and I will have no progress feedback and I will have to retype my search :(

And even if the symbol search *isn't* the first interaction overall, the delay on the first search is still a significant factor since jumping to a symbol is often an important task right in the middle of “having a thought”.

I expect that this aligns well with @davidbarsky’s goals.

davidbarsky:
It aligns with my goals, yeah.

I rebased your changes locally against my tracing changes and saw that `parallel_prime_caches` on my preferred test project (buck2) made parallel_prime_caches go from ~28 seconds to ~34 seconds, but symbol search (invoked by `Command + T` in VS Code was *instant*. I think this is a reasonable tradeoff.

PR #18192, opened 2024-09-26
Include buildfiles in VFS

darichey:
We subscribe to `textDocument/didSave` for `filesToWatch`, but the VFS doesn't contain those files. Before https://github.com/rust-lang/rust-analyzer/pull/18105, this would bring down the server. Now, it's only a benign error logged:
```
ERROR notification handler failed handler=textDocument/didSave error=file not found: /foo/bar/TARGETS
```
It's benign, because we will also receive a `workspace/didChangeWatchedFiles` for the file which will invalidate and load it.

Explicitly include the buildfiles in the VFS to prevent the handler from erroring.

### DIFF
--- a/crates/load-cargo/src/lib.rs
+++ b/crates/load-cargo/src/lib.rs
@@ -265,6 +265,11 @@ impl ProjectFolders {
                 entries.push(manifest.to_owned());
             }
 
+            for buildfile in ws.buildfiles() {
+                file_set_roots.push(VfsPath::from(buildfile.to_owned()));
+                entries.push(buildfile.to_owned());
+            }
+
             // In case of detached files we do **not** look for a rust-analyzer.toml.
             if !matches!(ws.kind, ProjectWorkspaceKind::DetachedFile { .. }) {
                 let ws_root = ws.workspace_root();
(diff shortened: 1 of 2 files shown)


crates/load-cargo/src/lib.rs
```diff
@@ -265,6 +265,11 @@ impl ProjectFolders {
                 entries.push(manifest.to_owned());
             }
 
+            for buildfile in ws.buildfiles() {
+                file_set_roots.push(VfsPath::from(buildfile.to_owned()));
+                entries.push(buildfile.to_owned());
```
davidbarsky, line 270: Would be nice to avoid these `.to_owned()` calls if possible.

davidbarsky (changes requested):

crates/load-cargo/src/lib.rs
darichey: `VfsPath` owns the `AbsPathBuf`, so I'm not sure if it can be avoided. fwiw, this is copied from above: https://github.com/rust-lang/rust-analyzer/blob/85ca217765a61db7d72783a18f77db6ef7022637/crates/load-cargo/src/lib.rs#L263-L266

davidbarsky (approved):
LGTM, but I won't bors this for conflict of interest reasons.

PR #18193, opened 2024-09-26
fix: Don't report a startup error when a discover command is configured

Wilfred:
Previously, r-a would show an error if both fetch_workspaces_queue and discover_workspace_queue were empty. We're in this state at startup, so users would see an error if they'd configured
discover_workspace_config.

Instead, allow the fetch_workspaces_queue to have zero items if discover_workspace_config is set.

Whilst we're here, prefer "failed to fetch" over "failed to discover", so the error message better reflects what this function is doing.

### DIFF
--- a/crates/rust-analyzer/src/reload.rs
+++ b/crates/rust-analyzer/src/reload.rs
@@ -743,18 +743,12 @@ impl GlobalState {
             return Ok(());
         };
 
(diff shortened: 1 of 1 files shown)


davidbarsky:
This looks to good to me, but I don't want to merge this because Wilfred and I work together and that feels like a conflict of interest.

lnicola (approved):

PR #18195, opened 2024-09-26
internal: remove `Default` from OpQueue

davidbarsky:
@Wilfred and I were talking about `OpQueue` and we identified one symptom of its (relative) weirdness is that `last_op_result` returns a `T::default()`; not `Option<&T>`, which means it's not possible to distinguish between "the `OpQueue` hasn't run yet" and "the OpQueue ran, but didn't produce a result". This branch fixes that.

### DIFF
--- a/crates/rust-analyzer/src/global_state.rs
+++ b/crates/rust-analyzer/src/global_state.rs
@@ -46,6 +46,11 @@ pub(crate) struct FetchWorkspaceRequest {
     pub(crate) force_crate_graph_reload: bool,
 }
 
+pub(crate) struct FetchWorkspaceResponse {
+    pub(crate) workspaces: Vec<anyhow::Result<ProjectWorkspace>>,
+    pub(crate) force_crate_graph_reload: bool,
+}
+
 // Enforces drop order
 pub(crate) struct Handle<H, C> {
     pub(crate) handle: H,
@@ -146,8 +151,7 @@ pub(crate) struct GlobalState {
     pub(crate) detached_files: FxHashSet<ManifestPath>,
 
     // op queues
-    pub(crate) fetch_workspaces_queue:
-        OpQueue<FetchWorkspaceRequest, Option<(Vec<anyhow::Result<ProjectWorkspace>>, bool)>>,
+    pub(crate) fetch_workspaces_queue: OpQueue<FetchWorkspaceRequest, FetchWorkspaceResponse>,
     pub(crate) fetch_build_data_queue:
         OpQueue<(), (Arc<Vec<ProjectWorkspace>>, Vec<anyhow::Result<WorkspaceBuildScripts>>)>,
     pub(crate) fetch_proc_macros_queue: OpQueue<Vec<ProcMacroPaths>, bool>,
@@ -502,7 +506,7 @@ impl GlobalState {
             mem_docs: self.mem_docs.clone(),
             semantic_tokens_cache: Arc::clone(&self.semantic_tokens_cache),
             proc_macros_loaded: !self.config.expand_proc_macros()
-                || *self.fetch_proc_macros_queue.last_op_result(),
+                || self.fetch_proc_macros_queue.last_op_result().copied().unwrap_or(false),
             flycheck: self.flycheck.clone(),
         }
     }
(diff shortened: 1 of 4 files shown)


crates/rust-analyzer/src/reload.rs
```diff
@@ -769,7 +772,11 @@ impl GlobalState {
     pub(super) fn fetch_build_data_error(&self) -> Result<(), String> {
         let mut buf = String::new();
 
-        for ws in &self.fetch_build_data_queue.last_op_result().1 {
+        let Some((_, ws)) = &self.fetch_build_data_queue.last_op_result() else {
```
Wilfred, line 775: Maybe consider `.unwrap_or_default()` here? That way the code could execute the `for` loop as below, and it's maybe easier to reason about (fewer returns).

Wilfred (commented):
I really like this idea, I think explicitly having a 'OpQueue has never done anything' is much clearer.

crates/rust-analyzer/src/reload.rs
davidbarsky: the unfortunate bit is that `Option` is also an iterator, so that'd lead to some very weird combinator gunk for a nested for loop. I'd prefer to keep this as-is.

davidbarsky:
The test failures are legitimate, fwiw.

crates/rust-analyzer/src/global_state.rs
```diff
@@ -502,7 +502,11 @@ impl GlobalState {
             mem_docs: self.mem_docs.clone(),
             semantic_tokens_cache: Arc::clone(&self.semantic_tokens_cache),
             proc_macros_loaded: !self.config.expand_proc_macros()
-                || *self.fetch_proc_macros_queue.last_op_result(),
+                || if let Some(loaded) = self.fetch_proc_macros_queue.last_op_result() {
+                    *loaded
+                } else {
+                    false
+                },
```
Veykril, line 509:
```suggestion
                || self.fetch_proc_macros_queue.last_op_result().copied().unwrap_or(false),
```

crates/rust-analyzer/src/reload.rs
```diff
@@ -474,7 +474,10 @@ impl GlobalState {
                 .all(|(l, r)| l.eq_ignore_build_data(r));
 
         if same_workspaces {
-            let (workspaces, build_scripts) = self.fetch_build_data_queue.last_op_result();
+            let Some((workspaces, build_scripts)) = self.fetch_build_data_queue.last_op_result()
+            else {
+                return;
+            };
```
Veykril, line 480: This is wrong, this means if we reload the workspace before building build scripts we'll stop updating things entirely. Previously we only early returned inthis case if `!Arc::ptr_eq(workspaces, &self.workspaces)` (and even then we mightve done rebuilt the crate graph still)

crates/rust-analyzer/src/reload.rs
Veykril: I prefer the early return here as it makes the logic more clear imo. an "early return control flow prelude" in a function is not that bad for reasoning if you ask me (unlike returns in the middle of a function).

crates/rust-analyzer/src/reload.rs
davidbarsky: oh, yeah. I see what you mean. fixed.

davidbarsky:
If there are no objections, I'll merge this after Monday's release. This change feels like it has a large blast radius.

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -549,8 +549,11 @@ impl GlobalState {
                 let subscriptions = subscriptions.clone();
                 // Do not fetch semantic diagnostics (and populate query results) if we haven't even
                 // loaded the initial workspace yet.
-                let fetch_semantic =
-                    self.vfs_done && self.fetch_workspaces_queue.last_op_result().is_some();
+                let fetch_semantic = self.vfs_done
+                    && self
+                        .fetch_workspaces_queue
+                        .last_op_result()
+                        .is_some_and(|inner| inner.is_some());
```
Veykril, line 556: We should be able to get rid of the `Option` inside of the `fetch_workspaces_queue` field now. It serves the same purpose as your changes (discerning between initial run) so the double nesting of options is pointless

davidbarsky: Done! I wanted to be conservative with my changes, but I removed the nested option.

(I also switched the "response" of `fetch_workspaces_queue` to a struct from an tuple. I'll switch the remaining OpQueues to use named structs instead of tuples in a subsequent PR.)

Veykril (approved):

PR #18196, opened 2024-09-26
internal: Add `SyntaxFactory` to ease generating nodes with syntax mappings

DropDemBits:
Part of [#​15710](https://github.com/rust-lang/rust-analyzer/issues/15710)

Instead of requiring passing a `&mut SyntaxEditor` to every make constructor to generate mappings, we instead wrap that logic in `SyntaxFactory`, and afterwards add all the mappings to the `SyntaxEditor`.

Includes an example of using `SyntaxEditor` & `SyntaxFactory` in the `extract_variable` assist.

### DIFF
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -1,8 +1,12 @@
 use hir::TypeInfo;
 use ide_db::syntax_helpers::suggest_name;
 use syntax::{
-    ast::{self, edit::IndentLevel, edit_in_place::Indent, make, AstNode, HasName},
-    ted, NodeOrToken,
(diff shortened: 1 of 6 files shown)


Veykril (approved):

Veykril:
Not really the place to discuss but I wonder how this could play into having quasi quoting https://github.com/rust-lang/rust-analyzer/issues/2227. I guess that could take a factory as an extra input or return something that needs to go through the factory to be "registered"

DropDemBits:
@Veykril Having an extra input to the quasi quoting macro(s) would probably be best, as doing something like post-registering with the factory would mean that we'd always need to generate `SyntaxMapping`s, even if we don't need it.

PR #18197, opened 2024-09-27
minor: Stricter requirements for package wide flycheck

### LINKED ISSUE #18194: Diagnostic message disappearing
**rust-analyzer version**: rust-analyzer 0.0.0 (03456c545 2024-08-22)

**rustc version**: rustc 1.82.0-beta.4 (8c27a2ba6 2024-09-21)

**editor or extension**: Neovim v0.10.1

Commit 03456c545f6f8abd507f9ee16b85f17d5d9346ee introduced an issue where diagnostics disappear.  Diagnostics are visible when first opening a file, but saving any file in the project will cause diagnostic messages to disappear from all files **for an affected crate**.  After making some changes to the crate, I also had the issue occur without saving the file.  The diagnostics just flash up on the editor quickly before disappearing.  The changes I made were just commenting out some unused code and fixing a compilation error.

This only seems to affect diagnostic messages from `cargo check`, things such a missing semicolons that are detected on-the-fly will still show up and remain through saves.

Unfortunately I've been unable to produce a minimal reproducible example, and cannot share my codebase.  Hopefully knowing the commit will help though.

The issue is in a workspace with 10 crates, and seemingly only affects one of them.  Nothing about the affected crate stands out as different to the others in any significant way.  The crate has no other errors or warnings from cargo check, other than dead code which I suppressed with `#![allow(dead_code)]` (removing the suppression has no effect on this issue).

My :LspLog is empty other than the following

```
[START][2024-09-25 22:02:32] LSP logging initiated
[WARN][2024-09-25 22:02:32] .../lua/vim/lsp.lua:628	"buf_attach_client called on unloaded buffer (id: 4): "
```

However this is also shown for previous, working versions and doesn't seem to be causing any issues.

All other language server features appear to be normal (go-tos, code actions, auto-complete).

alibektas:
Require the existence of a target and `check_workspace` to be false to restart package-wide flycheck. Fixes #18194 , #18104

### DIFF
--- a/crates/rust-analyzer/src/handlers/notification.rs
+++ b/crates/rust-analyzer/src/handlers/notification.rs
@@ -380,7 +380,7 @@ fn run_flycheck(state: &mut GlobalState, vfs_path: VfsPath) -> bool {
                     if id == flycheck.id() {
                         updated = true;
                         match package.filter(|_| {
-                            !world.config.flycheck_workspace(source_root_id) || target.is_some()
+                            !world.config.flycheck_workspace(source_root_id) && target.is_some()
                         }) {
                             Some(package) => flycheck
                                 .restart_for_package(package, target.clone().map(TupleExt::head)),


alibektas:
Either it is late and I am not thinking right or we should have used `&&` in the first place and it took me two weeks to figure out. I hope it is the latter.

Veykril (approved):

PR #18204, opened 2024-09-27
Update cc to 1.1.22

ChrisDenton:
This version of `cc` contains [a fix](https://github.com/rust-lang/cc-rs/releases/tag/cc-v1.1.22) to prevent spurious rebuilds. Hopefully this should help avoid the CI issues rustc has been having.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -145,9 +145,12 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.1.10"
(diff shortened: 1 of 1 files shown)


lnicola (approved):

lnicola:
Thanks! We've had users complaining about spurious rebuilds (of their project, not of r-a) because PowerShell mangled their PATH. I wonder if some of them are using `cc`.

Issue #18185, opened 2024-09-25
rename "object safety" to "dyn compatibility"


Veykril:
https://github.com/rust-lang/lang-team/issues/286

We have a diagnostic for this now, as well as a hover message that use object safety as the term, we should change that if this goes through.

noahmbright:
Is this as simple as scrubbing the code base for "object_safety", "ObjectSafety", and what not and swapping with dyn_compatibility? I can take care of this if that's the case

alibektas:
Almost. You can check https://github.com/rust-lang/rust-analyzer/tree/master/crates/hir-ty/src/object_safety and see which other parts refer to this part of the codebase change every reference, usage of words "Object Safety" etc. 

Do you want me to assign you to this issue?

noahmbright:
@rustbot claim

Got it. I think between grep, the compiler and a reviewer I can get this taken care of. Hopefully the former 2 make it easy on the latter

PR #18205, opened 2024-09-28
Rename object_safety

noahmbright:
First PR here (yay!), so I read some of the getting started docs. There are a couple references to `handlers.rs`, which as far as I can tell has been refactored into `handlers/*.rs`. I made some tweaks to that in one commit. There is one fixme about a function called `to_lsp_runnable`, which I can't find anywhere at all. I can update that if I get some more info there.

Otherwise I changed references to object safety, is object safe, etc., trying to match case/style as I went. There was one case I found where there's a trait from somewhere else called `is_object_safe`, which I found defined in my cargo registry. I didn't touch that for now, just marked it with a fixme

### DIFF
--- a/crates/hir-ty/src/dyn_compatibility.rs
+++ b/crates/hir-ty/src/dyn_compatibility.rs
@@ -1,4 +1,4 @@
-//! Compute the object-safety of a trait
+//! Compute the dyn-compatibility of a trait
 
 use std::ops::ControlFlow;
 
@@ -28,14 +28,14 @@ use crate::{
 };
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
-pub enum ObjectSafetyViolation {
+pub enum DynCompatibilityViolation {
     SizedSelf,
     SelfReferential,
     Method(FunctionId, MethodViolationCode),
     AssocConst(ConstId),
     GAT(TypeAliasId),
     // This doesn't exist in rustc, but added for better visualization
-    HasNonSafeSuperTrait(TraitId),
+    HasNonCompatibleSuperTrait(TraitId),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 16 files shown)


noahmbright:
Closes #18185

ShoyuVanilla:
> There was one case I found where there's a trait from somewhere else called is_object_safe, which I found defined in my cargo registry.

Do you mean https://doc.rust-lang.org/nightly/nightly-rustc/rustdoc_json_types/struct.Trait.html#structfield.is_object_safe ? It's nothing to do with rust-analyzer. But I think that this should be merged after rust-lang/lang-team#286 because "dyn compatiblity" is not an official term until it's done

HKalbasi:
> BTW I think that this should be merged after https://github.com/rust-lang/lang-team/issues/286 because "dyn compatiblity" is not an official term until it's done

Its FCP is completed so it is effectively done.

crates/hir-ty/src/dyn_compatibility.rs
```diff
@@ -50,70 +50,73 @@ pub enum MethodViolationCode {
...
 ) -> ControlFlow<()>
 where
-    F: FnMut(ObjectSafetyViolation) -> ControlFlow<()>,
+    F: FnMut(DynCompatibilityViolation) -> ControlFlow<()>,
 {
     for super_trait in all_super_traits(db.upcast(), trait_).into_iter().skip(1).rev() {
-        if db.object_safety_of_trait(super_trait).is_some() {
-            cb(ObjectSafetyViolation::HasNonSafeSuperTrait(trait_))?;
+        if db.dyn_compatibility_of_trait(super_trait).is_some() {
+            cb(DynCompatibilityViolation::HasNonSafeSuperTrait(trait_))?;
```
HKalbasi, line 76: I think this also needs to be changed to `HasNonCompatibleSuperTrait`.

HKalbasi (approved):

crates/hir-ty/src/dyn_compatibility.rs
noahmbright: Good catch, fixed

HKalbasi:
Thanks!
@bors r+

PR #18206, opened 2024-09-29
Fix: Handle block exprs as modules when finding their parents

### LINKED ISSUE #18187: Hover path does not correctly account for block scopes
Given
```rs

mod m {
    const _: () = {
        mod m2 {
            const C$0: () = ();
        }
    };
}
```
we get a hover path of `<crate-name>`, it ought to be `<crate-name>::m::m2` though (that is also what `module_path!()` would yield)

ShoyuVanilla:
Fixes #18187


crates/ide/src/hover/render.rs
```diff
@@ -813,7 +813,15 @@ fn definition_mod_path(db: &RootDatabase, def: &Definition, edition: Edition) ->
...
-    def.module(db).map(|module| path(db, module, definition_owner_name(db, def, edition), edition))
+    let container: Option<Definition> =
+        def.as_assoc_item(db).and_then(|assoc| match assoc.container(db) {
+            AssocItemContainer::Trait(trait_) => Some(trait_.into()),
+            AssocItemContainer::Impl(impl_) => impl_.self_ty(db).as_adt().map(|adt| adt.into()),
+        });
+    container
+        .unwrap_or(*def)
+        .module(db)
+        .map(|module| path(db, module, definition_owner_name(db, def, edition), edition))
```
ShoyuVanilla, line 824: We currently have another bug like in;
image
So, checks for the path of the assoc item container whenever `def` is an assoc item instead

crates/ide-completion/src/context.rs
```diff
@@ -743,7 +744,12 @@ impl<'a> CompletionContext<'a> {
             }
         });
 
-        let depth_from_crate_root = iter::successors(module.parent(db), |m| m.parent(db)).count();
+        let depth_from_crate_root = iter::successors(Some(module), |m| m.parent(db))
+            // `BlockExpr` modules are not count as module depth
+            .filter(|m| !matches!(m.definition_source(db).value, ModuleSource::BlockExpr(_)))
+            .count()
+            // exclude `m` itself
+            .saturating_sub(1);
```
ShoyuVanilla, line 752: This is a bit awkward but we need to substract one when `m` itself is a block expr module

Veykril (approved):

PR #18216, opened 2024-09-30
internal: switch remaining OpQueues to use named structs

davidbarsky:
Building atop of https://github.com/rust-lang/rust-analyzer/pull/18195, I switched `GlobalState::fetch_build_data_queue` to use a struct instead of a tuple.

(I didn't switch `fetch_proc_macros_queue` to not return a bool, as the return value is only used in one spot.)


Veykril (approved):

Veykril:
bors merge

PR #18239, opened 2024-10-04
chore: rename `salsa` to `ra_salsa`

davidbarsky:
Laying some groundwork to start before I import the new Salsa crate. Here's why:
1. As part of the migration, @darichey, @Wilfred, and I will create new Salsa equivalents of the existing databases/query groups. We'll get them to compile crate-by-crate.
2. Once we wrote all equivalents of all queries, we'd start to refactor usage sites of the vendored Salsa to use the new Salsa databases.
3. Starting porting usage sites of old Salsa to the new Salsa.
4. Remove the vendored `ra_salsa`; declare victory.


crates/ra-salsa/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "salsa"
+name = "ra-salsa"
```
Veykril, line 2: Can we keep this as salsa? (that way we dont create a new autopublished crate for the new name), given the `[lib].name` section this should be fine I think? If not then thats fine, its just a minor nit (same for the macros crate)

Veykril (approved):

crates/ra-salsa/Cargo.toml
davidbarsky: oh, good point! i added a `package` specifier wherever Salsa is used, so that'll avoid creating new autopublished crates.

ShoyuVanilla: A small confusion caused by this 😅 #18297

davidbarsky: whoops!

PR #18245, opened 2024-10-05
fix: include description in label details when detail field is marked for …

boattime:
Fixes https://github.com/rust-lang/rust-analyzer/issues/18231. 

When omitting the autocomplete detail field, the autocomplete label details can still be returned. Currently the label details are missing the description field if the detail field is included in resolveSupport since it is being overwritten as None and opted to be sent with `completionItem/resolve`.

Example completion capabilities.
```
completion = {          
    completionItem = {            
        commitCharactersSupport = true,            
        deprecatedSupport = true,            
        documentationFormat = { "markdown", "plaintext" },            
        insertReplaceSupport = true,            
        insertTextModeSupport = {              
            valueSet = { 1, 2 }            
        },            
        labelDetailsSupport = true,            
        preselectSupport = true,            
        resolveSupport = {              
            properties = { "documentation", "detail", "additionalTextEdits", "sortText", "filterText", "insertText", "textEdit", "insertTextFormat", "insertTextMode" }            
        },            
        snippetSupport = true,            
        tagSupport = {              
            valueSet = { 1 }            
        }          
}
```

PR #18247, opened 2024-10-05
lsp: fix completion_item something_to_resolve not being a latch to true

jhgg:
while looking at #18245 i noticed that `something_to_resolve` could technically flap between true -> false if some subsequent fields that were requested to be resolved were empty. 

this fixes that by using `|=` instead of `=` when assigning to `something_to_resolve` which will prevent it from going back to false once set. 

although some cases it's simply assigning to `true` i opted to continue to use `|=` there for uniformity sake. but happy to change those back to `=`'s. 

cc @SomeoneToIgnore


SomeoneToIgnore:
Thank you for the catch, looks good to me.

SomeoneToIgnore:
@Veykril  would you mind a review?

PR #18266, opened 2024-10-08
internal: Use macos-13 runners and bump MACOSX_DEPLOYMENT_TARGET to 13.0

lnicola:
As Monterey seems to be EOL.


lnicola:
Test release available on https://github.com/rust-lang/rust-analyzer/releases/tag/nightly.

PR #18275, opened 2024-10-10
Skip #[test_case] expansion

### LINKED ISSUE #18274: rust-analyzer hangs with too many `#[test_case]`s
**rust-analyzer version**: master

```rust
#[test_case]
#[test_case]
#[test_case]
#[test_case]
#[test_case]
#[test_case]
#[test_case]
#[test_case]
fn test() {}
```
rust-analyzer will hang while trying to analyze this file. I think it's exponential, so if it doesn't for you, just add more `#[test_case]` :)

darichey:
Fixes #18274, although I don't fully understand if this is the best fix (it's not clear to me why this didn't cause issues before https://github.com/rust-lang/rust-analyzer/pull/18085).

### DIFF
--- a/crates/hir-def/src/nameres/collector.rs
+++ b/crates/hir-def/src/nameres/collector.rs
@@ -1323,7 +1323,7 @@ impl DefCollector<'_> {
                         _ => return Resolved::No,
                     };
 
-                    // Skip #[test]/#[bench] expansion, which would merely result in more memory usage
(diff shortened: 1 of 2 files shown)


Veykril (approved):

Veykril:
The issue itself seems a bit weird I agree, either way this sounds correct to me to do anyways as we handle both `#[bench]` and `#[test]` this way
@bors r+

PR #18278, opened 2024-10-10
Do not consider match/let/ref of place that evaluates to ! to diverge, disallow coercions from them too

### LINKED ISSUE #18237: Do not consider match/let/ref of place that evaluates to ! to diverge, disallow coercions from them too
We should implement https://github.com/rust-lang/rust/pull/129392

ShoyuVanilla:
Resolves #18237

### DIFF
--- a/crates/hir-ty/src/infer/coerce.rs
+++ b/crates/hir-ty/src/infer/coerce.rs
@@ -139,8 +139,8 @@ impl CoerceMany {
         };
         if let Some(sig) = sig {
             let target_ty = TyKind::Function(sig.to_fn_ptr()).intern(Interner);
-            let result1 = ctx.table.coerce_inner(self.merged_ty(), &target_ty);
-            let result2 = ctx.table.coerce_inner(expr_ty.clone(), &target_ty);
+            let result1 = ctx.table.coerce_inner(self.merged_ty(), &target_ty, CoerceNever::Yes);
+            let result2 = ctx.table.coerce_inner(expr_ty.clone(), &target_ty, CoerceNever::Yes);
             if let (Ok(result1), Ok(result2)) = (result1, result2) {
                 ctx.table.register_infer_ok(InferOk { value: (), goals: result1.goals });
                 for &e in &self.expressions {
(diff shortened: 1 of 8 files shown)


crates/hir-ty/src/tests/never_type.rs
```diff
@@ -539,3 +539,249 @@ fn test() {
...
+            70..71 '0': i32
+            70..76 '0 as _': *const !
+            90..96 'ref _x': &'? ()
+            103..105 '*x': !
+            104..105 'x': *const !
+            103..105: expected (), got !
+        "#]],
+    )
+}
+
```
ShoyuVanilla, line 762: We should have a test that make type mismatch error without this PR, but it doesn't emit the error, either with or without this PR.
https://github.com/rust-lang/rust/pull/129392/files#r1746775782
I'll address it as a new issue

crates/hir-ty/src/tests/never_type.rs
```diff
@@ -539,3 +539,249 @@ fn test() {
 "#,
     );
 }
+
```
ShoyuVanilla, line 542: These tests are all of the tests that included in the rustc's original PR, except one case (commented below) and mir ones

crates/hir-ty/src/infer/coerce.rs
```diff
@@ -227,10 +227,16 @@ impl InferenceContext<'_> {
         expr: Option<ExprId>,
         from_ty: &Ty,
         to_ty: &Ty,
+        // [Comment from rustc](https://github.com/rust-lang/rust/blob/4cc494bbfe9911d24f3ee521f98d5c6bb7e3ffe8/compiler/rustc_hir_typeck/src/coercion.rs#L85-L89)
+        // Whether we allow `NeverToAny` coercions. This is unsound if we're
+        // coercing a place expression without it counting as a read in the MIR.
+        // This is a side-effect of HIR not really having a great distinction
+        // between places and values.
+        coerce_never: bool,
```
Veykril, line 235: Nit but we might want to make this a binary enum `CoerceNever::Yes/No` :^) so that its more clear to callers (we have an assist for that)

crates/hir-ty/src/infer/expr.rs
```diff
@@ -53,8 +54,13 @@ use super::{
 };
 
 impl InferenceContext<'_> {
-    pub(crate) fn infer_expr(&mut self, tgt_expr: ExprId, expected: &Expectation) -> Ty {
-        let ty = self.infer_expr_inner(tgt_expr, expected);
+    pub(crate) fn infer_expr(
+        &mut self,
+        tgt_expr: ExprId,
+        expected: &Expectation,
+        is_read: bool,
```
Veykril, line 61: Likewise for these

crates/hir-ty/src/infer/coerce.rs
ShoyuVanilla: That would be much less confusing 👍

Veykril (approved):

PR #18280, opened 2024-10-10
Wait for discover command on Finished/Error

darichey:
Currently, we [kill](https://github.com/rust-lang/rust-analyzer/blob/0fb804acb375b02a3beeaceeb75b71969ef37b15/crates/rust-analyzer/src/command.rs#L92) the discover process upon receiving Finished/Error which prevents it from completing gracefully and doing any cleanup/logging ([as we want to do in rust-project](https://github.com/facebook/buck2/blob/66be61717ff358cd9a4da4acd2acffc2f52d7052/integrations/rust-project/src/cli/develop.rs#L222)). Wait for it to exit instead.

### DIFF
--- a/crates/rust-analyzer/src/main_loop.rs
+++ b/crates/rust-analyzer/src/main_loop.rs
@@ -890,23 +890,33 @@ impl GlobalState {
             .expect("No title could be found; this is a bug");
         match message {
             DiscoverProjectMessage::Finished { project, buildfile } => {
-                self.discover_handle = None;
                 self.report_progress(&title, Progress::End, None, None, None);
                 self.discover_workspace_queue.op_completed(());
 
                 let mut config = Config::clone(&*self.config);
                 config.add_discovered_project_from_command(project, buildfile);
                 self.update_configuration(config);
+
+                let discover_handle = self
+                    .discover_handle
+                    .take()
+                    .expect("no project discovery command set; this is a bug.");
+                let _ = discover_handle.join();
             }
             DiscoverProjectMessage::Progress { message } => {
                 self.report_progress(&title, Progress::Report, Some(message), None, None)
             }
             DiscoverProjectMessage::Error { error, source } => {
-                self.discover_handle = None;
                 let message = format!("Project discovery failed: {error}");
                 self.discover_workspace_queue.op_completed(());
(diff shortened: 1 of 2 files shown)


crates/rust-analyzer/src/main_loop.rs
```diff
@@ -890,23 +890,27 @@ impl GlobalState {
...
             }
             DiscoverProjectMessage::Error { error, source } => {
-                self.discover_handle = None;
                 let message = format!("Project discovery failed: {error}");
                 self.discover_workspace_queue.op_completed(());
                 self.show_and_log_error(message.clone(), source);
-                self.report_progress(&title, Progress::End, Some(message), None, None)
+                self.report_progress(&title, Progress::End, Some(message), None, None);
+
+                let discover_handle = self.discover_handle.take().unwrap();
```
davidbarsky, line 912:
```suggestion
                let discover_handle = self.discover_handle.take().expect("no project discovery command set; this is a bug.");
```

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -890,23 +890,27 @@ impl GlobalState {
...
             DiscoverProjectMessage::Finished { project, buildfile } => {
-                self.discover_handle = None;
                 self.report_progress(&title, Progress::End, None, None, None);
                 self.discover_workspace_queue.op_completed(());
 
                 let mut config = Config::clone(&*self.config);
                 config.add_discovered_project_from_command(project, buildfile);
                 self.update_configuration(config);
+
+                let discover_handle = self.discover_handle.take().unwrap();
```
davidbarsky, line 900:
```suggestion
                let discover_handle = self.discover_handle.take().expect("no project discovery command set; this is a bug.");
```

davidbarsky (changes requested):
`homer_bushes.gif`

(lgtm modulo suggested nits.)

Veykril (commented):
That does mean the main loop will block on the process exiting which I don't think is a good idea?

darichey:
Closing in favor of https://github.com/facebook/buck2/commit/ae1b8e4bda7e3ebddd2cd628a775d8be81de3bcc

PR #18289, opened 2024-10-12
Fix panic when json project has relative buildfile paths

darichey:
The `build_file` path may be relative to the workspace root.


lnicola (approved):

lnicola:
Thanks!

@bors r+

PR #18292, opened 2024-10-12
feat: handle self-param outside of methods when renaming

### LINKED ISSUE #17427: renaming `self` should work outside of methods
```
pub fn dot(self, rhs: v64) -> f64 {
    self.x * rhs.x + self.y * rhs.y + self.z * rhs.z
}
````

This code is erroneous, as it is outside the impl. It should still be possible to rename `self` here:

```
pub fn dot(lhs: _, rhs: v64) -> f64 {
    lhs.x * rhs.x + lhs.y * rhs.y + lhs.z * rhs.z
}
```

### PR DESCRIPTION
**Pull request by roife**
close #17427

### CODE REVIEW

File: crates/ide/src/rename.rs [hunk 1]
```diff
@@ -1977,6 +1986,24 @@ impl Foo {
...
     }
 
+    #[test]
+    fn test_self_outside_of_methods() {
+        cov_mark::check!(rename_self_outside_of_methods);
+        check(
+            "foo",
+            r#"
+fn f($0self) -> i32 {
+    self.i
```
Veykril, line 1996: can we put a `use self as _;` into the body here? Just to check that we aren't accidentally touching the `self` modules

Veykril (approved):

PR #18294, opened 2024-10-13
Add wrap/unwrap return type in Option

Giga-Bowser:
I pretty much just copied over the code and tests for wrapping/unwrapping return types in `Result` and then did a bunch of find and replace changes. 

I handled unwrapping statements returning `None` by just replacing `None` with the unit type, but I'm open to suggestions for more intuitive behavior here.


Veykril (commented):
Hmm, there shouldn't be a need to fully duplicate the two modules. I'd expect us to be able to re-use a lot of code here and abstract it over Option/Result

Giga-Bowser:
That's a good point, I'll start working on that tonight.

Giga-Bowser:
Let me know if there's anything else you want to see changed. I'm not really sure what the precedent is for when to group assists, so I decided to leave them separate, but I'm open to suggestions for that as well.

crates/ide-assists/src/handlers/wrap_return_type.rs
```diff
@@ -26,6 +42,61 @@ use crate::{AssistContext, AssistId, AssistKind, Assists};
...
+
+    fn wrap_type(&self, type_ref: &ast::Type) -> ast::Type {
+        match self {
+            WrapperKind::Option => make::ext::ty_option(type_ref.clone()),
+            WrapperKind::Result => make::ext::ty_result(type_ref.clone(), make::ty_placeholder()),
+        }
+    }
+}
+
+fn wrap_return_type(acc: &mut Assists, ctx: &AssistContext<'_>, kind: WrapperKind) -> Option<()> {
```
Veykril, line 99: The way this is setup we do this work twice now (always), might be better to only have one entry point, do the filtering logic once and then add both assists at once. We also probably want to group the two assist under one assist group (label being something "Wrap return type in ...").

crates/ide-assists/src/handlers/unwrap_return_type.rs
```diff
@@ -0,0 +1,2159 @@
...
+// ```
+// ->
+// ```
+// fn foo() -> i32 { 42i32 }
+// ```
+pub(crate) fn unwrap_result_return_type(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
+    unwrap_return_type(acc, ctx, UnwrapperKind::Result)
+}
+
+fn unwrap_return_type(
```
Veykril, line 45: In similar fashion we can combine the entry points here to one and then check for either types after `resolve_type` to figure out which one is applicable

Veykril (commented):
Generally good to me, just raising 2 perf things that would be good to fix

Giga-Bowser:
Sounds good, I'll get that wrapped up today.

crates/ide-assists/src/handlers/unwrap_return_type.rs
```diff
@@ -65,11 +55,12 @@ fn unwrap_return_type(
...
-        kind.core_type(FamousDefs(&ctx.sema, ctx.sema.scope(type_ref.syntax())?.krate()))?;
-    if ret_enum != core_enum {
-        return None;
-    }
+
+    let famous_defs = FamousDefs(&ctx.sema, ctx.sema.scope(type_ref.syntax())?.krate());
+
+    let kind = UnwrapperKind::ALL
+        .iter()
+        .find(|k| matches!(k.core_type(&famous_defs), Some(core_type) if ret_enum == core_type))?;
```
Giga-Bowser, line 63: If this seems a bit too dense I could easily rewrite this as an `if` that checks against each `UnwrapperKind` individually, just let me know.

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #18297, opened 2024-10-14
minor: `ra-salsa` in `profile.dev.package`

ShoyuVanilla:
Since `ra-salsa`'s package name is actually `salsa` it makes the following warning in `cargo` commands; 

```
warning: profile package spec `ra-salsa` in profile `dev` did not match any packages
```

and the opt level isn't applied to it.


davidbarsky:
Whoops! @bors r+

PR #18299, opened 2024-10-15
fix: incorrect autofix for missing wrapped unit in return expr

### LINKED ISSUE #18298: autofix for wrapping in Ok() confused by return;
**rust-analyzer version**: `rust-analyzer 1 (2b750da1a1 2024-10-06)` (on Arch Linux)

**editor or extension**: Emacs

```rust
fn foo(b: bool) -> Result<(), String> {
    if b {
        return;
    }

    Err("oh dear".to_owned())
}
```

If I apply the fix offered in rust-analyzer, I get:

```rust
fn foo(b: bool) -> Result<(), String> {
    if b {
        Ok(return);
    }

    Err("oh dear".to_owned())
}
```

rustc gets this right, so I'm not sure why r-a differs in this case.

```
error[E0069]: `return;` in a function whose return type is not `()`
  --> src/main.rs:73:9
   |
71 | fn foo(b: bool) -> Result<(), String> {
   |                    ------------------ expected `Result<(), String>` because of this return type
72 |     if b {
73 |         return;
   |         ^^^^^^ return type is not `()`
   |
help: give the `return` a value of the expected type
   |
73 |         return Ok(());
   |                ++++++

For more information about this error, try `rustc --explain E0069`.
```

roife:
fix #18298.

We should insert `Ok(())` or `Some(())` instead of wrapping `return` with variants.

### DIFF
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -130,6 +130,7 @@ fn add_missing_ok_or_some(
     if d.actual.is_unit() {
         if let Expr::BlockExpr(block) = &expr {
             if block.tail_expr().is_none() {
+                // Fix for forms like `fn foo() -> Result<(), String> {}`
                 let mut builder = TextEdit::builder();
                 let block_indent = block.indent_level();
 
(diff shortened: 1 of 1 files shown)


Veykril (approved):

Wilfred:
Wow, that was fast! Thanks for putting up a fix :)

PR #18312, opened 2024-10-16
Report document symbols of kind `variable` for let statements

LastExceed:
initially i had implemented a bunch more kinds, but eventually realized that `ide_db::SymbolKind` != `lsp_types::SymbolKind` and that many of the former kind are mapped to the same latter kind, which could cause annoyances when e.g. the user wants their outline view in vscode to show `ide_db::SymbolKind::Const` and `ide_db::SymbolKind::Static` but not `ide_db::SymbolKind::ConstParam` (which wouldn't be possible to configure in vscode since all of them are mapped to `lsp_types::SymbolKind::CONSTANT`) so i decided to take a conservative approach and ended up removing all but 1 kind

### DIFF
--- a/crates/ide/src/file_structure.rs
+++ b/crates/ide/src/file_structure.rs
@@ -187,6 +187,24 @@ fn structure_node(node: &SyntaxNode) -> Option<StructureNode> {
                 };
                 Some(node)
             },
+            ast::LetStmt(it) => {
+                let pat = it.pat()?;
+
+                let mut label = String::new();
+                collapse_ws(pat.syntax(), &mut label);
+
+                let node = StructureNode {
+                    parent: None,
+                    label,
+                    navigation_range: pat.syntax().text_range(),
+                    node_range: it.syntax().text_range(),
+                    kind: StructureNodeKind::SymbolKind(SymbolKind::Local),
+                    detail: it.ty().map(|ty| ty.to_string()),
+                    deprecated: false,
+                };
+
+                Some(node)
+            },
             ast::Macro(it) => decl(it, StructureNodeKind::SymbolKind(SymbolKind::Macro)),
             _ => None,
         }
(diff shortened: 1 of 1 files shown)


Veykril:
What is the goal here? We generally don't consider let statements to be part of the file structure

LastExceed:
the goal is to make as much information accessible as possible. The client can always choose to selectively ignore stuff. E.g. if you don't want variable declarations to show up in VSCode's outline view then you can simply uncheck `Settings > Outline > Show Variables`

My personal motivation (which brought me here in the first place) is a language agnostic extension I am developing which selectively fades everything outside the currently selected scope/expression/whatever to reduce visual noise. This extension is currently more capable in e.g. TypeScript than it is in Rust simply because the TS language server provides more information than RA does

Whether or not variable declarations are part of the file structure is IMO a matter of opinion/preference, and I see no good reason to withhold information when configuration options exist

Veykril:
Fair, I think the last time someone proposed adding matches to this which was confusing.

Veykril (commented):
Please add to the `test_file_structure` test

LastExceed:
sry for the delay. added tests and rebased the branch

Veykril (approved):

matklad:
Is there a vs-code side settings to remove variables from the "goto document symbol" popup? The thing at the top left on the screenshot:

image

It looks like there isn't, so I think it's worth adding a server-side settings here. Going to function/type in the same document by name is one of the most useful navigation abilities of an IDE, and polluting namespace with locals make this features significantly less useful (in particular, less useful than just grepping for `fn prefix`)

matklad:
It would be good to rename file_structure to document_symbols as well! The original thinking was for this to only show structure, and for most languages, including Rust, local variables definitely do not belong to the structure (they sort of do in JavaScript and Zig, whey thaey can also play the role of Rust's `mod`).

I agree that returning everything and letting the client filter is a better approach technically, but then the name become wrong :)

PR #18315, opened 2024-10-17
Bump version of scip crate

varungandhi-src:
Follow up to https://github.com/sourcegraph/scip/issues/284

Manually verified that SCIP generation works OK for rust-analyzer itself.

cc @RalfJung


lnicola:
Thanks!

@bors r+

PR #18317, opened 2024-10-17
minor: sync from downstream

lnicola:


bench_data/numerous_macro_rules
```diff
@@ -528,7 +528,7 @@ macro_rules! __ra_macro_fixture526 {($expr :  expr )=>{|| -> _ { Some ($expr )}(
...
...)=>{$crate ::  __quote ! (@  SUBTREE  Brace $($tt )*)}; ([$($tt :  tt )* ])=>{$crate ::  __quote ! (@  SUBTREE  Bracket $($tt )*)}; (($($tt :  tt )* ))=>{$crate ::  __quote ! (@  SUBTREE  Parenthesis $($tt )*)}; ($tt :  literal )=>{ vec ! [$crate ::  quote ::  ToTokenTree ::  to_token ($tt ).  into ()]}; ($tt :  ident )=>{ vec ! [{ tt ::  Leaf ::  Ident ( tt ::  Ident { text :  stringify ! ($tt ).  into (),  id :  tt ::  TokenId ::  unspecified (), }).  into ()}]}; (-> )=>{$crate ::  __quote ! (@  PUNCT  '-' ,  '>' )}; (& )=>{$crate ::  __quote ! (@  PUNCT  '&' )}; (, )=>{$crate ::  __quote ! (@  PUNCT  ',' )}; (: )=>{$crate ::  __quote ! (@  PUNCT  ':' )}; (; )=>{$crate ::  __quote ! (@  PUNCT  ';' )}; (:: )=>{$crate ::  __quote ! (@  PUNCT  ':' ,  ':' )}; (. )=>{$crate ::  __quote ! (@  PUNCT  '.' )}; (< )=>{$crate ::  __quote ! (@  PUNCT  '<' )}; (> )=>{$crate ::  __quote ! (@  PUNCT  '>' )}; ($first :  tt $($tail :  tt )+ )=>{{ let  mut  tokens = $crate ::  quote ::  IntoTt ::  to_tokens ($crate ::  __quote ! ($first ));  let  mut  tail_tokens = $crate ::  quote ::  IntoTt ::  to_tokens ($crate ::  __quote ! ($($tail )*));  tokens .  append (&  mut  tail_tokens );  tokens }}; }
```
lnicola, line 531: I changed `(## $first : ident $($tail : tt )* )` to `(!! $first : ident $($tail : tt )* ) here.

crates/parser/src/lexed_str.rs
```diff
@@ -187,6 +187,12 @@ impl<'a> Converter<'a> {
                 }
 
                 rustc_lexer::TokenKind::RawIdent => IDENT,
+
+                rustc_lexer::TokenKind::GuardedStrPrefix => {
```
Veykril, line 191: Yuck we need to break the token apart here in < Edition2024

lnicola: Yeah, I'm afraid I'm gonna let you handle that :sweat_smile:.

Veykril: Will see when I get to that. I was also confused as to how this didn't break our test upstream but I forgot upstream doesn't run our tests (except for proc-macro-srv) ... We should make them run them :^)

lnicola: I thought they were disabled because of the impact on CI time, but it was probably being built with optimizations: https://github.com/rust-lang/rust/pull/113244.

PR #18320, opened 2024-10-17
internal: fix lldb-dap unconditionally calling rustc

davidbarsky:
Fixes https://github.com/rust-lang/rust-analyzer/issues/18318. I also took the opportunity to refactor how `discoverSourceFileMap` worked—it now returns a type instead of mutating a map in place.

I tested this change using the LLDB DAP extension. I needed to set `"lldb-dap.executable-path": "/Applications/Xcode.app/Contents/Developer/usr/bin/lldb-dap"` for everything to work, however, but once I did, was able to successfully debug a test.

### DIFF
--- a/editors/code/src/debug.ts
+++ b/editors/code/src/debug.ts
@@ -6,7 +6,6 @@ import type * as ra from "./lsp_ext";
 import { Cargo } from "./toolchain";
 import type { Ctx } from "./ctx";
 import { createTaskFromRunnable, prepareEnv } from "./run";
-import { execSync } from "node:child_process";
 import { execute, isCargoRunnableArgs, unwrapUndefinable } from "./util";
 import type { Config } from "./config";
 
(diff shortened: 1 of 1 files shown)


editors/code/src/debug.ts
```diff
@@ -203,9 +221,11 @@ async function discoverSourceFileMap(
         const commitHash = rx.exec(data)?.[1];
         if (commitHash) {
             const rustlib = path.normalize(sysroot + "/lib/rustlib/src/rust");
-            sourceFileMap[`/rustc/${commitHash}/`] = rustlib;
+            return { source: rustlib, destination: rustlib };
         }
     }
+
+    return;
```
lnicola, line 228: Is this `return` needed?

editors/code/src/debug.ts
```diff
@@ -189,11 +203,15 @@ async function getDebugConfiguration(
     return debugConfig;
 }
 
+type SourceFileMap = {
+    source: string;
+    destination: string;
```
lnicola, line 208: Nit: maybe `target`, but don't bother changing it.

lnicola (approved):

editors/code/src/debug.ts
davidbarsky: afraid so, yeah. Without this return, we'd get a "Not all code paths return a value. `ts(7030)`" error.

PR #18323, opened 2024-10-17
Fix link syntax

kevinwuhoo:
There's a small bug in the Markdown syntax for the below dialog:

Screenshot 2024-10-17 at 14 18 58

Removing the backticks should fix this.


lnicola:
The syntax is fine, it's supposed to be monospace, but the dialog probably doesn't accept Markdown.

Veykril:
Yes the bug is likely on the typescript side not telling vscode its markdown

Veykril:
Yes the issue is here https://github.com/rust-lang/rust-analyzer/blob/8dd53a3a46adffdc7928bbfabab90d6348c9a089/editors/code/src/ctx.ts#L449
This ought to be `appendMarkdown`

PR #18328, opened 2024-10-18
fix: Fix CI running analysis-stats incorrectly against the standard libraries

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/18326

### DIFF
--- a/crates/rust-analyzer/src/cli/analysis_stats.rs
+++ b/crates/rust-analyzer/src/cli/analysis_stats.rs
@@ -6,6 +6,7 @@ use std::{
     time::{SystemTime, UNIX_EPOCH},
 };
 
+use cfg::{CfgAtom, CfgDiff};
 use hir::{
     db::{DefDatabase, ExpandDatabase, HirDatabase},
     Adt, AssocItem, Crate, DefWithBody, HasSource, HirDisplay, HirFileIdExt, ImportPathConfig,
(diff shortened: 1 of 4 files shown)


ShoyuVanilla:
```
thread 'main' panicked at crates/hir-ty/src/infer/cast.rs:423:13:
internal error: entered unreachable code: `[Int(I8); Bytes([32, 0, 0, 0, 0, 0, 0, 0], Empty)]` should be sized but is not?
```
It seems that we are not getting `Sized` trait from lang items 🤔

Veykril:
Yes, we set the test cfg which causes core to be empty hence no lang items

Veykril:
Ugh, now its stuck in const eval somewhere

Veykril:
Okay now it should work
@bors r+

crates/rust-analyzer/src/cli/analysis_stats.rs
```diff
@@ -59,13 +60,17 @@ impl flags::AnalysisStats {
             Rand32::new(seed)
         };
 
+        let mut cfg_overrides = CfgOverrides::default();
+        cfg_overrides.global =
+            CfgDiff::new(vec![CfgAtom::Flag(hir::sym::miri.clone())], vec![]).unwrap();
```
Veykril, line 65: This probably should be a default for the cli commands

Veykril:
Clippy always has the last word
@bors r+

Issue #18287, opened 2024-10-12
Private items are shown in completions for modules defined in bodies (e.g. in functions)


Veetaha:
**rust-analyzer version**: `rust-analyzer version: 0.3.2137-standalone`
**rustc version**: `rustc 1.81.0 (eeb90cda1 2024-09-04)`
**editor or extension**:  VSCode v0.3.2137

<img src="https://github.com/user-attachments/assets/6d83f2be-d183-4912-9c51-9f4cca9b4513" width="300"/>

**code snippet to reproduce**: 
```rust
fn main() {
    mod nice_module {
        struct MyPrivateLife;
    }

    nice_module::
}
```

dqkqd:
I'm not sure if I understand correctly, but checking whether `to_module` is a block module and breaking the loop if it isn't might solves this?

https://github.com/rust-lang/rust-analyzer/blob/3ddfb0da474bdf243a655a5d785de9b59c7f1397/crates/hir-def/src/visibility.rs#L140-L150

ShoyuVanilla:
> I'm not sure if I understand correctly, but checking whether `to_module` is a block module and breaking the loop if it isn't might solves this?

@dqkqd I'm not sure cause I haven't started looking into this yet 😅
I just assigned myself to this because I'm dealing another namers/visibility issue now.
So, you can take this issue if you'd like 😄

dqkqd:
>  So, you can take this issue if you'd like 😄

@ShoyuVanilla Sure, I'd love to

PR #18337, opened 2024-10-19
fix: private items are shown in completions for modules in fn body

dqkqd:
Close: #18287

### DIFF
--- a/crates/hir-def/src/visibility.rs
+++ b/crates/hir-def/src/visibility.rs
@@ -139,13 +139,11 @@ impl Visibility {
         let def_map_block = def_map.block_id();
         loop {
             match (to_module.block, def_map_block) {
-                // to_module is not a block, so there is no parent def map to use
+                // `to_module` is not a block, so there is no parent def map to use.
                 (None, _) => (),
+                // `to_module` is at `def_map`'s block, no need to move further.
                 (Some(a), Some(b)) if a == b => {
                     cov_mark::hit!(is_visible_from_same_block_def_map);
-                    if let Some(parent) = def_map.parent() {
-                        to_module = parent;
-                    }
                 }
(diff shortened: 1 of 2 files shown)


crates/hir-def/src/visibility.rs
```diff
@@ -138,6 +138,12 @@ impl Visibility {
...
         loop {
+            // `to_module` is not a block module and should not be moved upward,
+            // as doing so might expose private items.
+            if !to_module.is_block_module() {
+                break;
+            }
+
             match (to_module.block, def_map_block) {
                 // to_module is not a block, so there is no parent def map to use
                 (None, _) => (),
```
ShoyuVanilla, line 149: Wasn't the case for the your above code `if !to_module.is_block_module() { .. }` being handled here?

Veykril: Was wondering the same but actually not quite, `is_block_module` is defined as `self.block.is_some() && self.local_id == DefMap::ROOT`, so we are checking `self.block.is_none()` (the match arm part) or `self.local_id != ROOT` (thats the new thing).

So if I get this right, the problem is that if we are in a module within a block module we incorrectly move upwards out of the local module right now. (this is all so confusing ..,).

Can we encode that branch in the match as well (or clean up the branching in some other way)? This as proposed makes it difficult to read imo

dqkqd: Thanks for pointing out.

Actually, the fix by checking whether `to_module` is a `block_module` was incorrect.

The problem was that we keep moving upwards even though we should stop as soon as we reach the same block as `from_module`'s `def_map`.

Veykril (approved):

Veykril:
Thanks!
@bors r+

PR #18361, opened 2024-10-21
fix: Fix token downmapping failing for include! inputs

Veykril:
Supercedes https://github.com/rust-lang/rust-analyzer/pull/18325

Fixes https://github.com/rust-lang/rust-analyzer/pull/18325
Fixes https://github.com/rust-lang/rust-analyzer/issues/18313
Fixes https://github.com/rust-lang/rust-analyzer/issues/18314

### DIFF
--- a/crates/hir/src/semantics/source_to_def.rs
+++ b/crates/hir/src/semantics/source_to_def.rs
@@ -104,7 +104,7 @@ use hir_expand::{
 };
 use rustc_hash::FxHashMap;
 use smallvec::SmallVec;
-use span::{FileId, MacroFileId};
+use span::{EditionedFileId, FileId, MacroFileId};
 use stdx::impl_from;
 use syntax::{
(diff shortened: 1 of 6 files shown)


crates/hir/src/semantics/source_to_def.rs
```diff
@@ -163,7 +164,10 @@ impl SourceToDefCtx<'_, '_> {
                             .include_macro_invoc(crate_id)
                             .iter()
                             .filter(|&&(_, file_id)| file_id == file)
-                            .flat_map(|(call, _)| {
+                            .flat_map(|(call, file_id)| {
+                                self.cache
+                                    .included_file_cache
+                                    .insert(*file_id, MacroFileId { macro_call_id: *call });
```
Veykril, line 170: This needs to happen somewhere else I think (its luck that the tests pass as is right now)

PR #18362, opened 2024-10-21
feat: goto definition on range operators

### LINKED ISSUE #18342: Go to definition on `..`/`..=` should go to the corresponding `Range*` type
As in `1..2` should go to `Range`, `1..` to `RangeExclusive` etc

duncpro:
Closes #18342

### DIFF
--- a/crates/hir/src/semantics.rs
+++ b/crates/hir/src/semantics.rs
@@ -17,7 +17,7 @@ use hir_def::{
     path::ModPath,
     resolver::{self, HasResolver, Resolver, TypeNs},
     type_ref::Mutability,
-    AsMacroCall, DefWithBodyId, FunctionId, MacroId, TraitId, VariantId,
+    AsMacroCall, DefWithBodyId, FunctionId, MacroId, StructId, TraitId, VariantId,
 };
 use hir_expand::{
     attrs::collect_attrs,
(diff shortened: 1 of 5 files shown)


Veykril (commented):
(shouldve said that in the issue as a mentoring instruction sorry)
I think we should add a couple new `OperatorClass`es (`OperatorClass::Range(Struct)`, etc) here for the range types https://github.com/rust-analyzer/rust-analyzer/blob/fb832ff2ba338105dda1c8172f89d580cb0a570f/crates/ide-db/src/defs.rs#L548-L554 and populate them here https://github.com/rust-analyzer/rust-analyzer/blob/fb832ff2ba338105dda1c8172f89d580cb0a570f/crates/ide-db/src/defs.rs#L309-L330

WIth that I think the go to def code needs no changes in fact (aside from tests) and this should automatically give us support for ranges in other ide features as well (like hover and stuff)

Veykril (approved):

Veykril:
Thanks!
@bors r+

1234Brian (approved):

duncpro:
@Veykril Thank you for helping me along. This was my first real OSS contribution.

Veykril:
Ah, if you want there is a follow up task. We can do the same for patterns I just realized, not just expressions since there are also range patterns :)

duncpro:
I'm down. I'll open an issue for it right now

PR #18382, opened 2024-10-23
fix: auto-complete import for aliased function and module

### LINKED ISSUE #17042: Public re-exports do not appear in code completion
**rust-analyzer version**: v0.3.1916

**rustc version**: 1.76.0

**editor or extension**: VSCodium

**relevant settings**: None

crate-a/lib.rs
```rust
pub fn add(a:i32,b:i32)->i32{
    a+b
}
```

crate-b/lib.rs
```rust
pub use crate_a::add as my_add;
```

crate-b/some.rs
```rust
fn xxx(){
    // type my_add here and there is no completion or auto-import.
}
```

dqkqd:
Close #17042 

Get the aliased name from `import_path` to use for import instead of name defined inside local module.

image

### DIFF
--- a/crates/ide-completion/src/render.rs
+++ b/crates/ide-completion/src/render.rs
@@ -281,13 +281,7 @@ pub(crate) fn render_resolution_with_import(
     import_edit: LocatedImport,
 ) -> Option<Builder> {
     let resolution = ScopeDef::from(import_edit.original_item);
-    // Use the last segment when `item_to_import` matches `original_item`,
-    // as it will take the aliased name into account.
-    let local_name = if import_edit.item_to_import == import_edit.original_item {
-        import_edit.import_path.segments().last()?.clone()
-    } else {
-        scope_def_to_name(resolution, &ctx, &import_edit)?
-    };
+    let local_name = get_import_name(resolution, &ctx, &import_edit)?;
     // This now just renders the alias text, but we need to find the aliases earlier and call this with the alias instead.
     let doc_aliases = ctx.completion.doc_aliases_in_scope(resolution);
     let ctx = ctx.doc_aliases(doc_aliases);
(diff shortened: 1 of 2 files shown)


crates/ide-completion/src/render.rs
```diff
@@ -281,8 +281,14 @@ pub(crate) fn render_resolution_with_import(
     import_edit: LocatedImport,
 ) -> Option<Builder> {
     let resolution = ScopeDef::from(import_edit.original_item);
-    let local_name = scope_def_to_name(resolution, &ctx, &import_edit)?;
-    //this now just renders the alias text, but we need to find the aliases earlier and call this with the alias instead
+    // Use the last segment when `item_to_import` matches `original_item`,
+    // as it will take the aliased name into account.
+    let local_name = if import_edit.item_to_import == import_edit.original_item {
+        import_edit.import_path.segments().last()?.clone()
```
dqkqd, line 287: I'm not sure if this is a right way to get the aliased name.

Veykril:
So, the root issue is due to https://github.com/rust-lang/rust-analyzer/issues/14079, basically we should be treating imports as their own items to refer to. I'd need to look into that issue again to tell where we are with it (iirc only the IDE integration is missing, but that is still quite a bit to work on)

crates/ide-completion/src/render.rs
Veykril: It is a decent workaround for the core issue I mentioned for now I think, could put a FIXME comment mentioning https://github.com/rust-lang/rust-analyzer/issues/14079 as the proper fix for this above?

dqkqd: I've added the FIXME.
There was a similar case with completion in pattern as well, so I added a testcase with the same fix.

Veykril (approved):
Thanks!

PR #18385, opened 2024-10-23
feat: Add assist to generate a type alias for a function

### LINKED ISSUE #18343: Assist to derive `type alias = signature` for a function
Given some function 
```rs
unsafe fn foo(param: u32) -> String { ... }
```
with the cursor on `foo` we should offer an assist that generates a type alias for the function's signature
```
type FooFn = unsafe fn(u32) -> String;
```

Giga-Bowser:
closes #18343
Technically this doesn't support generating `for <...>` lifetimes, but I'm not very familiar with them, and they seem a bit outside of the scope of an assist like this. If somebody has a way to generate them feel free to add it. 

Also, when generating aliases with self parameters, they are left unnamed even in the `generate_fn_type_alias_named` assist. I think this behavior makes the most sense but you could also name them something like `this` and add a placeholder to change the name if that's desired.

### DIFF
--- a/crates/ide-assists/src/handlers/generate_fn_type_alias.rs
+++ b/crates/ide-assists/src/handlers/generate_fn_type_alias.rs
@@ -1,0 +1,442 @@
+use either::Either;
+use hir::HirDisplay;
+use ide_db::assists::{AssistId, AssistKind, GroupLabel};
+use syntax::{
+    ast::{self, edit::IndentLevel, make, HasGenericParams, HasName},
+    syntax_editor, AstNode,
+};
+
+use crate::{AssistContext, Assists};
+
+// Assist: generate_fn_type_alias_named
+//
+// Generate a type alias for the function with named parameters.
+//
+// ```
+// unsafe fn fo$0o(n: i32) -> i32 { 42i32 }
+// ```
+// ->
+// ```
+// type ${0:FooFn} = unsafe fn(n: i32) -> i32;
+//
+// unsafe fn foo(n: i32) -> i32 { 42i32 }
+// ```
+
+// Assist: generate_fn_type_alias_unnamed
+//
+// Generate a type alias for the function with unnamed parameters.
+//
+// ```
+// unsafe fn fo$0o(n: i32) -> i32 { 42i32 }
+// ```
+// ->
+// ```
+// type ${0:FooFn} = unsafe fn(i32) -> i32;
+//
+// unsafe fn foo(n: i32) -> i32 { 42i32 }
+// ```
+
+pub(crate) fn generate_fn_type_alias(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
+    let name = ctx.find_node_at_offset::<ast::Name>()?;
+    let func = &name.syntax().parent()?;
+    let item = func.ancestors().find_map(ast::Item::cast)?;
+    let assoc_owner =
+        item.syntax().ancestors().nth(2).and_then(Either::<ast::Trait, ast::Impl>::cast);
+    let node = assoc_owner.as_ref().map_or_else(
+        || item.syntax(),
+        |impl_| impl_.as_ref().either(AstNode::syntax, AstNode::syntax),
+    );
+    let func_node = ast::Fn::cast(func.clone())?;
+    let param_list = func_node.param_list()?;
+
+    for style in ParamStyle::ALL {
+        let generic_params = func_node.generic_param_list();
+        let module = match ctx.sema.scope(node) {
+            Some(scope) => scope.module(),
+            None => continue,
+        };
+
+        acc.add_group(
+            &GroupLabel("Generate a type alias for function...".into()),
+            style.assist_id(),
+            style.label(),
+            func_node.syntax().text_range(),
+            |builder| {
+                let mut edit = builder.make_editor(func);
+
+                let alias_name = format!("{}Fn", stdx::to_camel_case(&name.to_string()));
+
+                let fn_abi = match func_node.abi() {
+                    Some(abi) => format!("{} ", abi),
+                    None => "".into(),
+                };
+
+                let fn_unsafe = if func_node.unsafe_token().is_some() { "unsafe " } else { "" };
+
+                let fn_qualifiers = format!("{fn_unsafe}{fn_abi}");
+
+                let fn_type = return_type(&func_node);
+
+                let mut fn_params_vec = Vec::new();
+
+                if let Some(self_param) = param_list.self_param() {
+                    if let Some(local) = ctx.sema.to_def(&self_param) {
+                        let ty = local.ty(ctx.db());
+                        if let Ok(s) = ty.display_source_code(ctx.db(), module.into(), false) {
+                            fn_params_vec.push(s)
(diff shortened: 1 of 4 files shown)


crates/ide-assists/src/handlers/generate_fn_type_alias.rs
```diff
@@ -0,0 +1,442 @@
...
+    let name = ctx.find_node_at_offset::<ast::Name>()?;
+    let func = &name.syntax().parent()?;
+    let item = func.ancestors().find_map(ast::Item::cast)?;
+    let assoc_owner =
+        item.syntax().ancestors().nth(2).and_then(Either::<ast::Trait, ast::Impl>::cast);
+    let node = assoc_owner.as_ref().map_or_else(
+        || item.syntax(),
+        |impl_| impl_.as_ref().either(AstNode::syntax, AstNode::syntax),
+    );
+    let func_node = ast::Fn::cast(func.clone())?;
```
Veykril, line 49: Let's move this below the `let fun` line so we filter out non functions as early as possible

crates/ide-assists/src/handlers/generate_fn_type_alias.rs
```diff
@@ -0,0 +1,442 @@
...
+    let param_list = func_node.param_list()?;
+
+    for style in ParamStyle::ALL {
+        let generic_params = func_node.generic_param_list();
+        let module = match ctx.sema.scope(node) {
+            Some(scope) => scope.module(),
+            None => continue,
+        };
```
Veykril, line 57: Can be moved out of the loop

crates/ide-assists/src/handlers/generate_fn_type_alias.rs
```diff
@@ -0,0 +1,442 @@
...
+        item.syntax().ancestors().nth(2).and_then(Either::<ast::Trait, ast::Impl>::cast);
+    let node = assoc_owner.as_ref().map_or_else(
+        || item.syntax(),
+        |impl_| impl_.as_ref().either(AstNode::syntax, AstNode::syntax),
+    );
+    let func_node = ast::Fn::cast(func.clone())?;
+    let param_list = func_node.param_list()?;
+
+    for style in ParamStyle::ALL {
+        let generic_params = func_node.generic_param_list();
```
Veykril, line 53: Can be moved into the `|builder|` closure

crates/ide-assists/src/handlers/generate_fn_type_alias.rs
```diff
@@ -0,0 +1,442 @@
...
+// ```
+// type ${0:FooFn} = unsafe fn(i32) -> i32;
+//
+// unsafe fn foo(n: i32) -> i32 { 42i32 }
+// ```
+
+pub(crate) fn generate_fn_type_alias(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
+    let name = ctx.find_node_at_offset::<ast::Name>()?;
+    let func = &name.syntax().parent()?;
+    let item = func.ancestors().find_map(ast::Item::cast)?;
```
Veykril, line 42: What item is this supposed to be? This and the `node` var are a bit confusing to me

crates/ide-assists/src/handlers/generate_fn_type_alias.rs
```diff
@@ -0,0 +1,442 @@
...
+                let mut edit = builder.make_editor(func);
+
+                let alias_name = format!("{}Fn", stdx::to_camel_case(&name.to_string()));
+
+                let fn_abi = match func_node.abi() {
+                    Some(abi) => format!("{} ", abi),
+                    None => "".into(),
+                };
+
+                let fn_unsafe = if func_node.unsafe_token().is_some() { "unsafe " } else { "" };
```
Veykril, line 74: There are more qualifiers to handle, `const` and `async` (keep ordering in mind, the qualifiers allow only a specific order)

crates/ide-assists/src/handlers/generate_fn_type_alias.rs
```diff
@@ -0,0 +1,442 @@
...
+                        fn_params_vec.extend(param_list.params().map(|p| p.to_string()))
+                    }
+                    ParamStyle::Unnamed => fn_params_vec.extend(
+                        param_list.params().filter_map(|p| p.ty()).map(|ty| ty.to_string()),
+                    ),
+                };
+
+                let fn_params = fn_params_vec.join(", ");
+
+                // FIXME: sometime in the far future when we have `make::ty_func`, we should use that
```
Veykril, line 102: You can create that function if you like, that would probably clean up the implementation here a bit as well

crates/ide-assists/src/handlers/generate_fn_type_alias.rs
```diff
@@ -0,0 +1,442 @@
...
+                let mut fn_params_vec = Vec::new();
+
+                if let Some(self_param) = param_list.self_param() {
+                    if let Some(local) = ctx.sema.to_def(&self_param) {
+                        let ty = local.ty(ctx.db());
+                        if let Ok(s) = ty.display_source_code(ctx.db(), module.into(), false) {
+                            fn_params_vec.push(s)
+                        }
+                    }
+                }
```
Veykril, line 89: Seeing this, it might be good to generally work with the hir version of a function instead of picking apart its syntax node and reconstructing it? That is wrok with the result of `ctx.sema.to_def(&func_node)`

Veykril:
> Technically this doesn't support generating for <...> lifetimes, but I'm not very familiar with them, and they seem a bit outside of the scope of an assist like this. If somebody has a way to generate them feel free to add it.

That's fine. One thing that would be nice to have though is that any generic parameters declared on the function are lifted to the type alias, that is:

`fn f<T>(t: T) {}` becomes `type TFn<T> = fn(T)`.

> Also, when generating aliases with self parameters, they are left unnamed even in the generate_fn_type_alias_named assist. I think this behavior makes the most sense but you could also name them something like this and add a placeholder to change the name if that's desired.

Yes thats fine that way I'd say

crates/ide-assists/src/handlers/generate_fn_type_alias.rs
Giga-Bowser: Function pointers cannot be `const` nor `async`. I suppose I could generate an alias returning `Future` for `async` functions, let me know if this is desirable. I'm not really sure of an equivalent for a `const` functions, but a non-`const` function pointer *should* accept a `const` function as an argument, so I think it's fine to just ignore a `const` qualifier.

Giga-Bowser:
> One thing that would be nice to have though is that any generic parameters declared on the function are lifted to the type alias, that is:
> 
> `fn f<T>(t: T) {}` becomes `type TFn<T> = fn(T)`.

Unless I misunderstand, this code does that. For example, the test `generate_fn_alias_unnamed_generics` ensures the assist turns 
```rust
fn foo<A, B>(a: A, b: B) -> i32 { return 42; }
```
into
```rust
type FooFn<A, B> = fn(A, B) -> i32;
```

Veykril (approved):

Veykril:
Thanks!

PR #18388, opened 2024-10-23
Fix checking for `false` `labelDetailsSupport` value.

jaboatman:
The check for `labelDetailsSupport` only checks for non-`None`, which means setting `labelDetailsSupport: false` incorrectly treats it as being supported.

For context: https://github.com/mrcjkb/rustaceanvim/discussions/414


lnicola:
Thanks! But let's remove the `as_ref` and do `== Some(true)` instead, it's a bit tidier.

jaboatman:
> Thanks! But let's remove the `as_ref` and do `== Some(true)` like above, it's a bit tidier.

Sure, no problem. Done.

PR #18390, opened 2024-10-23
fix: Prevent public re-export of private item

### LINKED ISSUE #18308: Import suggestions do not take into account visibility
I never had problems with this in the past, but suddenly rust-analyzer's import suggestions no longer take into account `pub` visibility. Unfortunately I don't have a minimal reproduction, but I do have a reproduction.

1. Clone `https://github.com/pola-rs/polars` (suggest `--depth 1`).
2. Edit file `crates/polars-expr/src/lib.rs`, add the following function:
  ```rust
fn foo() -> DataType {
    todo!()
}
```
3. Ask rust-analyzer to import `DataType` for you: Image
4. Suggestion is wrong, it's not visibile to us here, it suggests something else:
  Image
5. Suggestion is *still* wrong, it finally gets a public symbol:
  Image

**rust-analyzer version**: rust-analyzer version: 0.4.2148-standalone (418c1365e 2024-10-15) 

**rustc version**: rustc 1.83.0-nightly (ed04567ba 2024-09-28)

**repository link (if public, optional)**: https://github.com/polar-rs/polars

### PR DESCRIPTION
**Pull request by ShoyuVanilla**
Fixes #18308

The original issue is quite complicated.
`polar-rs` monorepo has the following structure;

```rust
// crate polars-core
pub mod prelude {
    pub use crate::datatype::DataType;
}

// crate polars-ops
pub mod frame {
    pub use join::*;

    pub mod join {
        pub use hash_join::*;
        use polars_core::prelude::*;

        mod hash_join {
            pub use super::*;
        }
    }
}
```

1. The glob import `pub use super::*` inside `mod hash_join` reexports the visibility of the glob import `use polars_core::prelude::*` inside the `mod join` as "public", which was previously "module(join)"
2. The glob import `pub use hash_join::*` inside `mod join` overrides visibility of the previous visibility of the glob import `use polars_core::prelude::*` with "public" in 1., which was previously "module(join)"

The problematic step is 1., as rustc doesn't allow such re-export with E0603.
This PR prevent such "more visible" re-export by taking minimum of resolved item visibility and the re-export visibility

### DIFF
--- a/crates/hir-def/src/find_path.rs
+++ b/crates/hir-def/src/find_path.rs
@@ -1025,7 +1025,7 @@ pub mod ast {
         check_found_path(
             r#"
 mod bar {
-    mod foo { pub(super) struct S; }
+    mod foo { pub(crate) struct S; }
     pub(crate) use foo::*;
 }
 $0
@@ -1047,7 +1047,7 @@ $0
         check_found_path(
             r#"
 mod bar {
-    mod foo { pub(super) struct S; }
+    mod foo { pub(crate) struct S; }
     pub(crate) use foo::S as U;
 }
 $0
--- a/crates/hir-def/src/nameres/collector.rs
+++ b/crates/hir-def/src/nameres/collector.rs
@@ -985,12 +985,8 @@ impl DefCollector<'_> {
         for (name, res) in resolutions {
             match name {
                 Some(name) => {
-                    changed |= self.push_res_and_update_glob_vis(
-                        module_id,
-                        name,
-                        res.with_visibility(vis),
-                        import,
-                    );
+                    changed |=
+                        self.push_res_and_update_glob_vis(module_id, name, *res, vis, import);
                 }
                 None => {
                     let tr = match res.take_types() {
@@ -1043,10 +1039,11 @@ impl DefCollector<'_> {
             .collect::<Vec<_>>();
 
         for (glob_importing_module, glob_import_vis, use_) in glob_imports {
+            let vis = glob_import_vis.min(vis, &self.def_map).unwrap_or(glob_import_vis);
             self.update_recursive(
                 glob_importing_module,
                 resolutions,
-                glob_import_vis,
+                vis,
                 Some(ImportType::Glob(use_)),
                 depth + 1,
             );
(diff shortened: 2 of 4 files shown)


crates/hir-def/src/find_path.rs
```diff
@@ -1047,7 +1047,7 @@ $0
         check_found_path(
             r#"
 mod bar {
-    mod foo { pub(super) struct S; }
+    mod foo { pub(crate) struct S; }
     pub(crate) use foo::S as U;
```
ShoyuVanilla, line 1051: These test case is wrong. It emits `E0364` on rustc

```rust
mod bar {
    mod foo { pub(super) struct S; }
    pub(crate) use foo::S as U;
}

fn main() {}
```

```
error[E0364]: `S` is private, and cannot be re-exported
 --> src/main.rs:3:20
  |
3 |     pub(crate) use foo::S as U;
  |                    ^^^^^^^^^^^
  |
note: consider marking `S` as `pub` in the imported module
 --> src/main.rs:3:20
  |
3 |     pub(crate) use foo::S as U;
  |                    ^^^^^^^^^^^
```

ShoyuVanilla:
BTW I spent most time on making a MCVE, as the original issue was quite hard to reproduce outside the `pola-rs` repository😅
Both `pub use super::*` and the importing order mattered

Veykril (approved):

orlp:
Great work! Sorry for the large reproduction, I tried a bit to make it minimal but wasn't successful 😓.

ShoyuVanilla:
> Great work! Sorry for the large reproduction, I tried a bit to make it minimal but wasn't successful 😓.

No worries 😅 The root causes were quite interwoven with each other and I just managed to made a MVCE with custom r-a build with extra tracings.
Thank you for maintaining `polars`, a great library for data manipulations

lnicola:
This seems to have caused a regression in pattern type inferences:

image

ShoyuVanilla:
> This seems to have caused a regression in pattern type inference:

Oh, I'll check about this

ShoyuVanilla:
I ran `cargo run --release -p rust-analyzer -- unresolved-references .` and it seems that r-a isn't resolving the following layout things since this PR 🤔 
https://github.com/rust-lang/rust-analyzer/blob/f4466718b838de706d74e2c13f20a41c034d87a5/crates/hir/src/lib.rs#L1432

ShoyuVanilla:
MCVE:

```rust
//- /main.rs crate:main deps:test_crate
use test_crate::foo; // not resolving this!

use foo::Foo;

//- /lib.rs crate:test_crate deps:test_crate2
extern crate test_crate2;

pub use test_crate2 as foo;

//- /lib2.rs crate:test_crate2
pub struct Foo;
```

It seems that I should exempt min visibility for `extern crate`, as it seems to be exceptional in rustc 🤔

PR #18392, opened 2024-10-23
Swap query call order in `file_item_tree_query`

Veykril:
cc https://github.com/rust-lang/rust-analyzer/issues/18387


davidbarsky (approved):
What Lukas said [here](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Debugging.20.2318387/near/478576539):

> Basically under the assumption that the LRU stuff is causing the problem, here we compute the parse tree, then we compute the ast_id_map (which calls the parse tree query as well). Looking at the panic message, it looks like the ast_id_map is old, where as the tree is new (has one more character in it), so my gut feeling is that the ast_id_map is actually not being invalidated here, which might be because we repopulate the evicted parse tree cache entry prior to it (and not marking thigns as dirty where we should), but in the same query stack state

PR #18394, opened 2024-10-23
internal: Pretty-print Config in status command

Wilfred:
Config can become very big, even for relatively small rust projects, and printing everything on one line makes reading the output in VS Code harder.


Veykril (approved):

lnicola:
Related, `Config::default_config` is huge and a bit confusing, at least when you're looking at the unformatted output. Maybe we should omit it when printing `Config`.

PR #18395, opened 2024-10-23
fix: Add missing cfg flags for `core` crate

Wilfred:
Some types in `core` are conditionally compiled based on `target_has_atomic` or `target_has_atomic_load_store` without an argument, for example `AtomicU64`.

This is less noticeable in Cargo projects, where rust-analyzer adds the output of `RUSTC_BOOTSTRAP=1 cargo rustc --print cfg` so it gets the full set of cfg flags.

This fixes go-to-definition on `std::sync::atomic::AtomicU64` in non-cargo projects.

### DIFF
--- a/crates/project-model/src/rustc_cfg.rs
+++ b/crates/project-model/src/rustc_cfg.rs
@@ -24,14 +24,15 @@ pub(crate) fn get(
     config: RustcCfgConfig<'_>,
 ) -> Vec<CfgAtom> {
     let _p = tracing::info_span!("rustc_cfg::get").entered();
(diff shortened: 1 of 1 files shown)


Veykril (approved):

Veykril:
> This is less noticeable in Cargo projects, where rust-analyzer adds the output of RUSTC_BOOTSTRAP=1 cargo rustc --print cfg so it gets the full set of cfg flags.

Sounds like we should make that work for non cargo projects as well (and ideally remove the hardcoded stuff here entirely)

PR #18404, opened 2024-10-24
feat: Implement diagnostics pull model

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/18375

### DIFF
--- a/crates/rust-analyzer/src/main_loop.rs
+++ b/crates/rust-analyzer/src/main_loop.rs
@@ -417,6 +417,8 @@ impl GlobalState {
                 }
             }
 
+            let supports_diagnostic_pull_model = self.config.text_document_diagnostic();
+
             let client_refresh = became_quiescent || state_changed;
             if client_refresh {
                 // Refresh semantic tokens if the client supports it.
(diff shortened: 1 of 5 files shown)


Veykril:
Okay implementing this is rather straight forward actually, that's nice

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -417,6 +417,8 @@ impl GlobalState {
                 }
             }
 
+            let supports_diagnostic_pull_model = self.config.text_document_diagnostic().is_some();
```
lnicola, line 420: This feels a bit confusing. `text_document_diagnostic` actually checks for `related_document_support`. And the latter should probably return `bool`, to avoid situations like #18388?

rchl:
I know this should be a separate issue but for now I don't have time for that so just gonna point at https://github.com/sublimelsp/LSP/issues/2572

Basically, there appears to be some issues with the implementation - the server still pushes diagnostics even if client supports pull diagnostics and also it returns empty diagnostics for pull request.

Veykril:
> Basically, there appears to be some issues with the implementation - the server still pushes diagnostics even if client supports pull diagnostics 

This is intentional, the native rust-analyzer diagnostics support the push model, but the on save document handler that invokes cargo check remains using the push model as pulling makes little sense in that way (cargo is inherently pushing). The LSP does not forbid (to my knowledge) both to be used.

jwortmann:
According to the comments in https://github.com/microsoft/language-server-protocol/issues/1743#issuecomment-1584057181 servers should not use both pull and push diagnostics at the same time (at least for a particular DocumentUri; what you are allowed to do is to register pull diagnostics only for certain URIs and then use push diagnostics for different URIs):

> If a server opts into providing pull then the client pulls for diagnostics and the server doesn't need to push.

> So I opt to simply add to the spec that a server if opting into pull it shouldn't push diagnostics for the same set of resources.

(though I think it was forgotten to add this last sentence to the specs explicitly)

---

Furthermore, since you don't seem to register an `identifier` for the pull diagnostics (as far as I can tell from the recording provided in https://github.com/sublimelsp/LSP/issues/2572), and push diagnostics per definition don't have any identifier, it would mean that they just override each other for the same URI, i.e. the latest arrived wins.

PR #18407, opened 2024-10-25
Fix formatting on welcome page, read only paths setting example

hackervole:
This attempts to resolve https://github.com/rust-lang/rust-analyzer/issues/17788.

It looks like the description text of the "walkthrough" steps doesn't support Markdown code blocks, but you can provide a Markdown file that will be rendered on the right half of the page in place of the logo image. This file does support code blocks and syntax highlighting, and as an added bonus, the text can be selected/copied.

The final result looks like this:
image

Additionally, it looks like this renderer expects double backticks for inline code snippets, so I updated the descriptions of the other walkthrough steps accordingly.

### DIFF
--- a/editors/code/.vscodeignore
+++ b/editors/code/.vscodeignore
@@ -12,3 +12,4 @@
 !ra_syntax_tree.tmGrammar.json
 !server
 !README.md
+!walkthrough-setup-tips.md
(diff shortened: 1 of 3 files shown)


Veykril (approved):
Neat, thanks for figuring that out!

PR #18412, opened 2024-10-25
internal: Stop producing .gz artifacts for Windows

lnicola:


Veykril (approved):

lnicola:
I'll merge this on Monday, to get a note about it in the release notes.

ia0:
Not that the change matters to me, but I'm curious to know the rationale.

lnicola:
ZIP files are nicer because they preserve the executable flag (relevant on Unix) and original file name (e.g. https://github.com/rust-lang/rust-analyzer/issues/9469, but there were others). And we were already producing Windows ZIP artifacts since forever. And as a bonus, they also contain debug symbols,

ia0:
Oh I see, I missed the fact that it was really just a `.gz` and not a `.tar.gz`. Makes sense. Thanks a lot for the explanation!

lnicola:
Yeah, at some point Windows was shipping a `tar` version (based on `libarchive`) that could read `.tar.gz`, but not `.gz`.

PR #18413, opened 2024-10-25
fix: Allow public re-export of `extern crate` import

ShoyuVanilla:
MCVE:

```rust
//- /main.rs crate:main deps:test_crate
use test_crate::foo; // not resolving this!

use foo::Foo;

//- /lib.rs crate:test_crate deps:test_crate2
extern crate test_crate2;

pub use test_crate2 as foo;

//- /lib2.rs crate:test_crate2
pub struct Foo;
```

It seems that I should exempt min visibility for `extern crate`, as it seems to be exceptional in rustc 🤔

_Originally posted by @ShoyuVanilla in https://github.com/rust-lang/rust-analyzer/issues/18390#issuecomment-2438253336_

Should fix the regression made by #18390 
Thank you @lnicola for pointing out this regression 👍

### DIFF
--- a/crates/hir-def/src/nameres/collector.rs
+++ b/crates/hir-def/src/nameres/collector.rs
@@ -31,7 +31,7 @@ use crate::{
     item_scope::{ImportId, ImportOrExternCrate, ImportType, PerNsGlobImports},
     item_tree::{
         self, AttrOwner, FieldsShape, FileItemTreeId, ImportKind, ItemTree, ItemTreeId,
-        ItemTreeNode, Macro2, MacroCall, MacroRules, Mod, ModItem, ModKind, TreeId,
+        ItemTreeNode, Macro2, MacroCall, MacroRules, Mod, ModItem, ModKind, TreeId, UseTreeKind,
     },
     macro_call_as_call_id, macro_call_as_call_id_with_eager,
     nameres::{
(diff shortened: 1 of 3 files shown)


crates/hir-def/src/nameres/tests.rs
```diff
@@ -385,6 +385,41 @@ pub struct Arc;
...
+//- /extern_crate.rs crate:extern_crate
+pub mod exported {
+    pub struct PublicItem;
+    struct PrivateItem;
+}
+
+pub struct Exported;
+struct NotExported1;
+struct NotExported2;
+struct NotExported3;
```
ShoyuVanilla, line 412: Oh, these three items should be public visibility 😅 
I'll fix this tomorrow

ShoyuVanilla:
Oh, I found another bug - preexisting before #18390 - related to extern crate that should be dealt with this together.
I'll handle it as well with this

crates/hir-def/src/nameres/collector.rs
```diff
@@ -1058,8 +1059,37 @@ impl DefCollector<'_> {
...
+        // https://github.com/rust-lang/rust/issues/127909
+        if let Some((_, v, it)) = defs.types.as_mut() {
+            let is_extern_crate_reimport_without_prefix = || {
+                let Some(ImportOrExternCrate::ExternCrate(_)) = it else {
+                    return false;
+                };
+                let Some(ImportType::Import(id)) = def_import_type else {
+                    return false;
+                };
+                let source = id.import.child_source(self.db);
```
Veykril, line 1076: We can't use that here, this draws a dependency edge between the defmap query and the AST causing any AST change to invalidate the def map. We need to derive this info from the item tree somehow

ShoyuVanilla: Okay, I'll try different approach

PR #18415, opened 2024-10-26
fix: Turn "Remove `dbg!`" into a quick fix for better prioritization

### LINKED ISSUE #15364: `Remove dbg!` should have higher priority than `Inline macro`

lnicola:
Closes #15364
See https://github.com/microsoft/vscode/issues/193355


lnicola:
The LSP examples for `refactor.rewrite` seem more heavy-weight than this one, I think it's fine to have it as a quick fix:

	 * - Convert JavaScript function to class
	 * - Add or remove parameter
	 * - Encapsulate field
	 * - Make method static
	 * - Move method to base class

PR #18436, opened 2024-10-29
Do not render meta info when hovering usages

Veykril:
Closes https://github.com/rust-lang/rust-analyzer/issues/17096


Wilfred:
Huge fan of the ordering changes to the hover information, thank you :)

PR #18441, opened 2024-10-30
internal: Do not cache the config directory path

Veykril:
Closes https://github.com/rust-lang/rust-analyzer/pull/18070


Veykril:
Seems like its still racing

PR #18458, opened 2024-11-01
feat: Add diagnostic fix to remove unnecessary wrapper in type mismatch

### LINKED ISSUE #18383: Add/remove Some() or Ok() or Err() wrapper
I frequently find myself refactoring code and needing to either add or remove a `Some()` wrapper from an expression. Similar need for `Ok()` and `Err()`. It'd be great if Rust Analyzer had an action to add or remove these wrappers.

I can approximate this somewhat with macros of key sequences in my editor, but it'd be neat if RA could do it :)

Giga-Bowser:
closes #18383 
I also reorganized some of the tests for the fix to add a missing `Ok()` or `Some()` wrapper as they were somewhat scattered. The diff for this ended up looking messy but it's just moving functions around.

### DIFF
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -9,9 +9,10 @@ use syntax::{
     ast::{
         self,
         edit::{AstNodeEdit, IndentLevel},
-        make, BlockExpr, Expr, ExprStmt, HasArgList,
+        syntax_factory::SyntaxFactory,
+        BlockExpr, Expr, ExprStmt, HasArgList,
     },
-    ted, AstNode, AstPtr, TextSize,
+    AstNode, AstPtr, TextSize,
 };
 
 use crate::{adjusted_display_range, fix, Assist, Diagnostic, DiagnosticCode, DiagnosticsContext};
@@ -223,8 +224,9 @@ fn remove_unnecessary_wrapper(
 
     let inner_arg = call_expr.arg_list()?.args().next()?;
 
-    let mut builder = SourceChangeBuilder::new(expr_ptr.file_id.original_file(ctx.sema.db));
-
+    let file_id = expr_ptr.file_id.original_file(db);
+    let mut builder = SourceChangeBuilder::new(file_id);
+    let mut editor;
     match inner_arg {
         // We're returning `()`
         Expr::TupleExpr(tup) if tup.fields().next().is_none() => {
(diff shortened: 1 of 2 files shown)


crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -184,6 +189,86 @@ fn add_missing_ok_or_some(
...
+    let expr_range = expr.syntax().text_range();
+    let actual_enum = d.actual.as_adt().and_then(|adt| adt.as_enum())?;
+    let famous_defs = FamousDefs(&ctx.sema, ctx.sema.scope(expr.syntax())?.krate());
+    let file_id = expr_ptr.file_id.original_file(ctx.sema.db);
+
+    let Expr::CallExpr(call_expr) = expr else {
+        return None;
+    };
+
+    let path_str = call_expr.syntax().descendants().find_map(ast::PathExpr::cast)?.to_string();
```
Veykril, line 209: We shouldn't do stringly checks, let's inspect the actual ADT type here instead, `d.actual.as_adt()` and then compare that to the relevant famous defs to check which one it is.

crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -184,6 +189,86 @@ fn add_missing_ok_or_some(
...
 }
 
+fn remove_unnecessary_wrapper(
+    ctx: &DiagnosticsContext<'_>,
+    d: &hir::TypeMismatch,
+    expr_ptr: &InFile<AstPtr<ast::Expr>>,
+    acc: &mut Vec<Assist>,
+) -> Option<()> {
+    let root = ctx.sema.db.parse_or_expand(expr_ptr.file_id);
+    let expr = expr_ptr.value.to_node(&root);
```
Veykril, line 199: This node might be inside a macro expansion which will cause issues later as we aren't accounting for that in our replacements so we should attempt to upmap the node via `ctx.sema.original_ast_node(&expr)?`

crates/ide-diagnostics/src/handlers/type_mismatch.rs
Giga-Bowser: This gets me the relevant enum type, but I can't seem to figure out how to get which specific was variant constructed. Any idea?

crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -184,6 +189,90 @@ fn add_missing_ok_or_some(
...
+                    editor.replace(stmt_list.syntax().parent()?, new_block.syntax());
+                }
+            }
+
+            editor.add_mappings(make.finish_with_mappings());
+        }
+        _ => {
+            editor = builder.make_editor(call_expr.syntax());
+            // FIXME: remove `clone_for_update` when `SyntaxEditor` handles it for us
+            editor.replace(call_expr.syntax(), inner_arg.syntax().clone_for_update());
```
Veykril, line 261: This probably needs adjusting now after your recent PR got merged

Veykril (approved):
Thanks!

Veykril (approved):
Thanks!

PR #18459, opened 2024-11-01
feat: Use snippet placeholders for generated match arms

Giga-Bowser:
When generating missing match arms, we currently use `todo!()`s for the match arm bodies. This is fine, but given they're obviously meant to be replaced, I figure they should also be snippet placeholders.


DropDemBits (approved):

DropDemBits (approved):
All good!

Giga-Bowser:
I know I said I would migrate this to the new `SyntaxEditor` API earlier, but that's actually been tougher than I expected, so it should probably go in it's own PR. Once this PR is merged I'll open another and outline the migration there.

lnicola (approved):

Issue #18347, opened 2024-10-19
function import get replaced by a module import


wiiznokes:
**rust-analyzer version**: (eg. output of "rust-analyzer: Show RA Version" command, accessible in VSCode via <kbd>Ctrl/⌘</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd>)

0.3.2146-standalone

**rustc version**: (eg. output of `rustc -V`)

rustc 1.84.0-nightly (3ed6e3cc6 2024-10-17)
**editor or extension**: (eg. VSCode, Vim, Emacs, etc. For VSCode users, specify your extension version; for users of other editors, provide the distribution if applicable)

**relevant settings**: (eg. client settings, or environment variables like `CARGO`, `RUSTC`, `RUSTUP_HOME` or `CARGO_HOME`)

**repository link (if public, optional)**: (eg. [rust-analyzer](https://github.com/rust-lang/rust-analyzer))

**code snippet to reproduce**: 
```rust
// add your code here

```

rwakulszowa:
I can take a look at this. @rustbot claim

rwakulszowa:
## A simpler repro

1. Start with this
  
  ```rs
  // main.rs
  use lsp_test::foo;
  
  fn main() {
      println!("{:}", foo());
  }
  ```
  
  ```rs
  // lib.rs
  pub fn foo() -> u32 {
      44
  }
  
  pub mod foo {
      pub fn bar() -> u32 {
          88
      }
  }
  ```
  Compiles and works fine.

2. Call `bar` in main.
  ```rs
  // main.rs
  use lsp_test::foo::{self, bar};  // auto imported

  fn main() {
      println!("{:} {:}", foo(), bar());
  }
  ```
  > // expected function, found module `foo`

PR #18464, opened 2024-11-02
Test function imports replaced by module imports.

rwakulszowa:
Test for #18347.
The issue affects all code using import merging. When a function and a module share the same identifier, e.g.

```rust
pub fn foo() {}
pub mod foo { pub fn bar() {} }
```

rust analyzer often confuses the function with the module, generating invalid code.

### DIFF
--- a/crates/ide-assists/src/handlers/auto_import.rs
+++ b/crates/ide-assists/src/handlers/auto_import.rs
@@ -1722,4 +1722,39 @@ mod foo {
             ",
         );
     }
+
+    #[test]
+    fn check_understands_function_and_module_imports() {
+        // FIXME(18347): import of `bar` should not be converted
+        // to `bar::{self}`. The former is a function, the latter is a module.
+        check_assist(
+            auto_import,
+            r"
+            use foo::bar;
+
+            mod foo {
+                pub fn bar() {}
+
+                pub mod bar {
+                    pub fn baz() {}
+                }
+            }
+
+            baz$0();
+            ",
+            r"
+            use foo::bar::{self, baz};
+
+            mod foo {
+                pub fn bar() {}
+
+                pub mod bar {
+                    pub fn baz() {}
+                }
+            }
+
+            baz();
+            ",
+        );
+    }
 }
--- a/crates/ide-assists/src/handlers/merge_imports.rs
+++ b/crates/ide-assists/src/handlers/merge_imports.rs
@@ -802,4 +802,36 @@ use top::{a::A, b::{B as D, B as C}};
 ",
         );
     }
+
+    #[test]
+    fn test_doesnt_convert_function_to_module() {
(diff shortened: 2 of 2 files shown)


rwakulszowa:
I haven't been able to actually find a proper fix yet. I can't find a reliable way to tell the merging function whether an import is a module (and can be used for merging) or a function.

I'm not sure if I manage to find a solution, so I'm sending a small PR with what I have.

Veykril:
The IDE layer currently lacks being able to resolve something to multiple namespaces (`Semantics::resolve_path` only yields one namespace at most), so that needs tackling first and is unfortunately a bigger change.

Veykril:
I'll close this for now due to inactivity, feel free to re-open if you try to tackle it again

rwakulszowa:
I'll give it another try.
____

> The IDE layer currently lacks being able to resolve something to multiple namespaces (`Semantics::resolve_path` only yields one namespace at most), so that needs tackling first and is unfortunately a bigger change.

I'm not sure I get this part. Why would we need `resolve_path` to resolve to multiple namespaces?

Let's look at the `auto_import` case.
https://github.com/rust-lang/rust-analyzer/blob/62dea277cc13031e01483a9400a3d20c888f586d/crates/ide-assists/src/handlers/auto_import.rs#L131-L144
My (limited) understanding is as follows. When the user selects the suggestion, we generate a new `use ...` tree
https://github.com/rust-lang/rust-analyzer/blob/62dea277cc13031e01483a9400a3d20c888f586d/crates/ide-db/src/imports/insert_use.rs#L213
and then try to merge it into existing imports
https://github.com/rust-lang/rust-analyzer/blob/62dea277cc13031e01483a9400a3d20c888f586d/crates/ide-db/src/imports/insert_use.rs#L220-L231
ultimately ending at
https://github.com/rust-lang/rust-analyzer/blob/62dea277cc13031e01483a9400a3d20c888f586d/crates/ide-db/src/imports/merge_imports.rs#L87-L118

IIUC, the root cause is how the common prefix is calculated - the current logic is not `Sema` aware, it finds the common prefix (== merge point) doing simple string comparison - if the same string is a module and a function, we end up with the buggy suggestion. If we add `Sema` to that function, we will be able to make sure that common prefix ends with a module
```
// pseudocode
let mut prefix = /*current logic*/
if Sema::resolve_path(prefix) is module {
  prefix = prefix.parent();
}
```
Would that do the trick?

Veykril:
My point is that `resolve_path` might tell us that the path resolves to a module, but it also reoslves to a macro (because of namespaces). So now we think it only resolves to a module, while it actually resolves to other things as well and thus the check is incorrect. We need to discern between the case of the path *only* resolving to a module or to resolving at least one non-module thing.

rwakulszowa:
Mkay.
Nvm then :/

PR #18480, opened 2024-11-05
Discover config improvements

bobozaur:
This PR adds some small QoL improvements to the `workspace.discoverConfig` workflow:
- Extends file watching beyond just the crate specific build files. Some build systems have files other scattered across the workspace that are yet worth keeping an eye on.
- Adds a `run_at_startup` setting to optionally allow the discovery to take place at startup, before workspace fetching. This means though that no `DiscoverArgument` gets passed, which is why this defaults to `false`.


davidbarsky:
Sorry for the delay! 

> Extends file watching beyond just the crate specific build files. Some build systems have files other scattered across the workspace that are yet worth keeping an eye on.

This part seems immediately and obviously reasonable.

> Adds a run_at_startup setting to optionally allow the discovery to take place at startup, before workspace fetching. This means though that no DiscoverArgument gets passed, which is why this defaults to false.

This *probably* seems reasonable to me, but out of curiosity: what build system is this?

bobozaur:
> This probably seems reasonable to me, but out of curiosity: what build system is this?

We're using Bazel. This is not that important as the discovery would happen when a Rust file is opened anyway, but it's a nice thing to have to get it more on par with how a Cargo workspace behaves.

May I mention that we ended up avoiding the `DiscoverArgument` because I couldn't really get it to work, haha. The JSON projects get generated for each crate as they are accessed and `rust-analyzer` accepts and indexes them, but weirdly nothing changes in the IDE. Might be path related or something as it works though if the `buildfile` in the `DiscoverProject::Finished` is explicilty set to the `BUILD.bazel` file at the root of the workspace, but that means only one crate is indexed at a time which is a no-go.

I do feel like I'm missing something, but not entirely sure what. I looked a lot at the integration that the Buck2 build system has but couldn't find some different logic. Nevertheless, since we generate a big JSON project for all the crates in a workspace and avoid the `DiscoverArgument` entirely, I figured it might be nice to run the discovery at startup as well. Made it configurable and optional because the `DiscoverArgument` thing still evades me and I wouldn't want to mess up someone's working setup.

I can spit the `runAtStartup` thing into a separate PR if you feel like this is not the right way to go about it and it might need some more work. This way the file watching adjustments would not be blocked.

bobozaur:
Closing this as it seems the build file paths were not being constructed correctly. Once that issue was fixed, build file tracking seems to work as expected.

PR #18482, opened 2024-11-06
internal: Add public `direct_supertraits(…)` & `all_supertraits(…)` accessor methods to `hir::Trait`

regexident:
Hi @Veykril,

I need to access a trait's supertraits (as external user of `ra_ap`'s APIs) and while `hir_ty::all_super_traits()` already provides such capabilities to a limited degree I'd like to propose the following accessors on `hir::Trait` itself, it only does provide access to the entire trait hierarchy (while I only want the direct supertraits) and requires a descent into `hir_ty` land. I'd thus like to propose adding the following methods: `hir::Trait::direct_supertraits(db)` & `hir::Trait::all_supertraits(db)`.

---

There seems to be some naming inconsistency with regard to `supertrait` (43 occurrences) vs. `super_trait` (53 occurrences) in the codebase. (I'd be happy to open a follow-up PR unifying the project-wide naming, if so desired.)
I went with `supertraits` in this PR since that's what the already existing `hir::Trait::items_with_supertraits(…)` uses in close proximity to the newly added methods.

### DIFF
--- a/crates/hir-ty/src/utils.rs
+++ b/crates/hir-ty/src/utils.rs
@@ -43,6 +43,17 @@ pub(crate) fn fn_traits(
         .flat_map(|it| it.as_trait())
 }
 
+/// Returns an iterator over the direct super traits (including the trait itself).
+pub fn direct_super_traits(db: &dyn DefDatabase, trait_: TraitId) -> SmallVec<[TraitId; 4]> {
+    let mut result = smallvec![trait_];
+    direct_super_traits_cb(db, trait_, |tt| {
+        if !result.contains(&tt) {
+            result.push(tt);
+        }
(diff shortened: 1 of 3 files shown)


lnicola (approved):

crates/hir-ty/src/utils.rs
```diff
@@ -43,6 +43,17 @@ pub(crate) fn fn_traits(
         .flat_map(|it| it.as_trait())
 }
 
+/// Returns an iterator over the direct super traits (including the trait itself).
```
lnicola, line 46: "Iterator" :laughing:

(I know, `all_super_traits` says the same thing)

Issue #18285, opened 2024-10-11
Migrate assists to `SyntaxEditor`


darichey:
We want to migrate all assists to use the new `SyntaxEditor`/`SyntaxFactory` abstraction. This will unblock moving to immutable syntax trees (#15710). In doing so, we can choose a flat, contiguous representation that should yield significant performance wins (#17491).

@DropDemBits has [outlined](https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer/topic/.60SyntaxEditor.60.20replacement.20for.20Mutable.20Syntax.20trees/near/472962988) the process for migrating most assists:
> * Create a `SyntaxFactory`, optionally named `make` to ease migration
> * Create a `SyntaxEditor` using `builder.make_editor` with a node from the file
> * Replace all `make::` with make
>   * Add any missing methods to `syntax::ast::syntax_factory::constructors` as needed
> * Replace all `ted::` with `editor.`
> * After all changes are complete:
>   * `editor.add_mappings(make.finish_with_mappings())` to add the generated mappings
>   * `builder.add_file_edits(ctx.file_id(), editor)` to add the edits
> For assists that use snippets, check the snippet capability, and then use `editor.add_annotation` to associate the snippets with the elements.

They have also already migrated the `extract_variable` assist in #18196 which should be a useful reference.

The assists are located in [`crates/ide-assists/src/handlers`](https://github.com/rust-lang/rust-analyzer/tree/master/crates/ide-assists/src/handlers). They should all have tests which you can use to verify that your change works properly by running:
```
cargo test --package ide-assists -- handlers::<ASSIST>::tests  
```
replacing `<ASSIST>` with the assist name, e.g., `add_braces`.

darichey:
We have two MLH fellows, @tareknaser, @davidkurilla, who are interested in working on this.

Giga-Bowser:
I'm not sure of the precedent on this sort of thing, but I've got two open PRs, #18458 and #18459, that both have to do with assists, and if it would help I'd gladly migrate those assists as part of those PRs.

Specifically I would be migrating:
- `remove_unnecessary_wrapper` which is added in #18458
- `add_missing_match_arms` which is modified in #18459

Giga-Bowser:
[...]

PR #18483, opened 2024-11-06
Migrate `introduce_named_generic` Assist to Use `SyntaxFactory`

tareknaser:
This PR is part of #15710 and #18285.

**Changes Made**
- Followed the steps in #18285 for migration.
- Added a new `SyntaxFactory` method, `type_param`.
- Replaced `add_generic_param` (originally in `GenericParamList`) with a new method, `syntax_editor_add_generic_param`. This new method is similar but takes a `SyntaxEditor` instance instead of `ted` to avoid the “immutable tree” error.

### DIFF
--- a/crates/ide-assists/src/handlers/introduce_named_generic.rs
+++ b/crates/ide-assists/src/handlers/introduce_named_generic.rs
@@ -42,8 +42,6 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext<'_>
         target,
         |edit| {
             let mut editor = edit.make_editor(&parent_node);
-            let impl_trait_type = edit.make_mut(impl_trait_type);
-            let fn_ = edit.make_mut(fn_);
             let fn_generic_param_list = fn_.get_or_create_generic_param_list();
 
             let existing_names = fn_generic_param_list
--- a/crates/ide-assists/src/handlers/introduce_named_generic.rs
+++ b/crates/ide-assists/src/handlers/introduce_named_generic.rs
@@ -1,8 +1,11 @@
 use ide_db::syntax_helpers::suggest_name;
 use itertools::Itertools;
 use syntax::{
-    ast::{self, edit_in_place::GenericParamsOwnerEdit, make, AstNode, HasGenericParams, HasName},
-    ted,
+    ast::{
+        self, edit_in_place::GenericParamsOwnerEdit, syntax_factory::SyntaxFactory, AstNode,
+        HasGenericParams, HasName,
+    },
+    SyntaxElement,
 };
 
 use crate::{AssistContext, AssistId, AssistKind, Assists};
@@ -25,12 +28,20 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext<'_>
 
     let type_bound_list = impl_trait_type.type_bound_list()?;
 
+    // FIXME: Is this node appropriate to use for SyntaxEditor in this case?
+    let parent_node = match ctx.covering_element() {
+        SyntaxElement::Node(n) => n,
+        SyntaxElement::Token(t) => t.parent()?,
+    };
+    let make = SyntaxFactory::new();
+
     let target = fn_.syntax().text_range();
     acc.add(
         AssistId("introduce_named_generic", AssistKind::RefactorRewrite),
         "Replace impl trait with generic",
         target,
         |edit| {
+            let mut editor = edit.make_editor(&parent_node);
             let impl_trait_type = edit.make_mut(impl_trait_type);
             let fn_ = edit.make_mut(fn_);
             let fn_generic_param_list = fn_.get_or_create_generic_param_list();
@@ -47,11 +58,12 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext<'_>
             )
             .for_impl_trait_as_generic(&impl_trait_type);
 
-            let type_param = make::type_param(make::name(&type_param_name), Some(type_bound_list))
+            let type_param = make
+                .type_param(make.name(&type_param_name), Some(type_bound_list))
                 .clone_for_update();
-            let new_ty = make::ty(&type_param_name).clone_for_update();
+            let new_ty = make.ty(&type_param_name).clone_for_update();
 
-            ted::replace(impl_trait_type.syntax(), new_ty.syntax());
+            editor.replace(impl_trait_type.syntax(), new_ty.syntax());
             fn_generic_param_list.add_generic_param(type_param.into());
 
             if let Some(cap) = ctx.config.snippet_cap {
@@ -61,6 +73,9 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext<'_>
                     edit.add_tabstop_before(cap, generic_param);
                 }
             }
+
+            editor.add_mappings(make.finish_with_mappings());
+            edit.add_file_edits(ctx.file_id(), editor);
         },
     )
 }
--- a/crates/ide-assists/src/handlers/introduce_named_generic.rs
+++ b/crates/ide-assists/src/handlers/introduce_named_generic.rs
@@ -56,13 +56,11 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext<'_>
             )
             .for_impl_trait_as_generic(&impl_trait_type);
 
-            let type_param = make
-                .type_param(make.name(&type_param_name), Some(type_bound_list))
-                .clone_for_update();
-            let new_ty = make.ty(&type_param_name).clone_for_update();
+            let type_param = make.type_param(make.name(&type_param_name), Some(type_bound_list));
+            let new_ty = make.ty(&type_param_name);
 
             editor.replace(impl_trait_type.syntax(), new_ty.syntax());
-            fn_generic_param_list.add_generic_param(type_param.into());
+            fn_generic_param_list.syntax_editor_add_generic_param(&mut editor, type_param.into());
 
             if let Some(cap) = ctx.config.snippet_cap {
                 if let Some(generic_param) =
--- a/crates/ide-assists/src/handlers/introduce_named_generic.rs
+++ b/crates/ide-assists/src/handlers/introduce_named_generic.rs
@@ -39,16 +39,17 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext<'_>
         target,
         |edit| {
             let mut editor = edit.make_editor(&parent_node);
-            let fn_generic_param_list =
-                fn_.syntax_editor_get_or_create_generic_param_list(&mut editor);
 
-            let existing_names = fn_generic_param_list
-                .generic_params()
-                .flat_map(|param| match param {
-                    ast::GenericParam::TypeParam(t) => t.name().map(|name| name.to_string()),
-                    p => Some(p.to_string()),
-                })
-                .collect_vec();
+            let existing_names = match fn_.generic_param_list() {
+                Some(generic_param_list) => generic_param_list
+                    .generic_params()
+                    .flat_map(|param| match param {
+                        ast::GenericParam::TypeParam(t) => t.name().map(|name| name.to_string()),
+                        p => Some(p.to_string()),
+                    })
+                    .collect_vec(),
+                None => Vec::new(),
+            };
             let type_param_name = suggest_name::NameGenerator::new_with_names(
                 existing_names.iter().map(|s| s.as_str()),
             )
@@ -58,13 +59,13 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext<'_>
             let new_ty = make.ty(&type_param_name);
 
             editor.replace(impl_trait_type.syntax(), new_ty.syntax());
-            fn_generic_param_list.syntax_editor_add_generic_param(&mut editor, type_param.into());
+            fn_.syntax_editor_add_generic_param(&mut editor, type_param.into());
 
             if let Some(cap) = ctx.config.snippet_cap {
                 if let Some(generic_param) =
                     fn_.generic_param_list().and_then(|it| it.generic_params().last())
                 {
-                    edit.add_tabstop_before(cap, generic_param);
+                    editor.add_annotation(generic_param.syntax(), edit.make_tabstop_before(cap));
                 }
             }
 
--- a/crates/ide-assists/src/handlers/introduce_named_generic.rs
+++ b/crates/ide-assists/src/handlers/introduce_named_generic.rs
@@ -51,7 +51,7 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext<'_>
 
             editor.replace(impl_trait_type.syntax(), new_ty.syntax());
             let generic_param = syntax::ast::GenericParam::from(type_param);
-            fn_.syntax_editor_add_generic_param(&mut editor, generic_param.clone());
+            editor.syntax_editor_add_generic_param(fn_, generic_param.clone());
 
             if let Some(cap) = ctx.config.snippet_cap {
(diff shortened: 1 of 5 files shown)


tareknaser:
Marking this as a draft for now since I’m running into a couple of errors with this assist

1. The first issue is with cursor positioning:
```sh
---- handlers::introduce_named_generic::tests::replace_two_impl_trait_with_generic_params stdout ----
Left:
fn foo<G, $0B: Bar>(foo: impl Foo, bar: B) {}

Right:
fn foo<$0G, B: Bar>(foo: impl Foo, bar: B) {}

Diff:
fn foo<$0G, $0B: Bar>(foo: impl Foo, bar: B) {}


thread 'handlers::introduce_named_generic::tests::replace_two_impl_trait_with_generic_params' panicked at crates/ide-assists/src/handlers/introduce_named_generic.rs:105:9:
text differs
```
Here, only the cursor position appears to be off.

2. The second error is related to an immutable tree in rowan:
```sh
---- tests::generated::doctest_introduce_named_generic stdout ----
thread 'tests::generated::doctest_introduce_named_generic' panicked at /Users/tareknasser/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rowan-0.15.15/src/cursor.rs:810:9:
immutable tree: fn foo(bar: impl Bar) {}
```

crates/ide-assists/src/handlers/introduce_named_generic.rs
```diff
@@ -25,14 +28,20 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext<'_>
...
     let target = fn_.syntax().text_range();
     acc.add(
         AssistId("introduce_named_generic", AssistKind::RefactorRewrite),
         "Replace impl trait with generic",
         target,
         |edit| {
-            let impl_trait_type = edit.make_mut(impl_trait_type);
-            let fn_ = edit.make_mut(fn_);
+            let mut editor = edit.make_editor(&parent_node);
             let fn_generic_param_list = fn_.get_or_create_generic_param_list();
```
ShoyuVanilla, line 45: This `get_or_create_generic_param_list` mutates syntax tree if there is no generic params - as the function name shows 😄 - and causes immutable tree panics in rowan. (You can check this with `RUST_BACKTRACE=full cargo test <panicking test name>` or `$env:RUST_BACKTRACE="full"; cargo test <panicking test name>` if you are on Windows.)
This should be modified to copying existing generic params or making fresh new generic params rather than inserting into exsisting syntax tree.
#18385 might be a good reference working on `SyntaxFactory` assist things

ShoyuVanilla: The reason for panicking is because L35: `let fn_ = edit.make_mut(fn_)` is being removed. Of course, it should be removed since we are migrating to `SyntaxFactory` from mutable syntax tree and we should modify the L45

tareknaser: Thank you for the review

> This `get_or_create_generic_param_list` mutates syntax tree if there is no generic params ... This should be modified to copying existing generic params 

How can I ‘copy existing generic params’ if there are none to begin with?

> or making fresh new generic params

How would I go about doing that? I looked into initializing a new `GenericParamList`, and the only way I found was using `make::generic_param_list`, which is actually what’s in place now (but fails).


I defined a new `ast::Fn` method similar to `get_or_create_generic_param_list()` that calls `syntax_editor_create_generic_param_list` if there are no existing generic params. This uses `SyntaxEditor` to initialize the list (since `insert_raw()` isn’t available for `SyntaxEditor`, I used `insert` instead).

Unfortunately, I can’t fully test my updates due to a new error:
```sh
---- tests::generated::doctest_introduce_named_generic stdout ----
thread 'tests::generated::doctest_introduce_named_generic' panicked at crates/syntax/src/syntax_editor.rs:55:9:
assertion failed: is_ancestor_or_self(&position.parent(), &self.root)
```
After tracing, this seems to be related to my `syntax_editor_add_generic_param` implementation, though it’s identical to `add_generic_param` except it uses `SyntaxEditor` instead of `ted`.

tareknaser: By the way, I know these new functions would ideally go into a separate file, but right now I’m just focused on getting them working first

ShoyuVanilla: Sorry for the really late reply. I've been quite busy 'cause I moved to a new job 😅

Your code that creating fresh new generic param list seems good but as it still panics, I'll add some comments.

crates/syntax/src/ast/edit_in_place.rs
```diff
@@ -257,6 +290,28 @@ impl ast::GenericParamList {
...
+                    self.generic_params().map(|param| param.clone()).collect::<Vec<_>>();
+                params.push(new_param.into());
+                let new_param_list = make::generic_param_list(params);
+
+                editor.replace(self.syntax(), new_param_list.syntax());
+            }
+            None => {
+                let position = crate::syntax_editor::Position::after(self.l_angle_token().unwrap());
+                let new_param_list = make::generic_param_list(once(new_param.clone()));
+                editor.insert(position, new_param_list.syntax());
```
ShoyuVanilla, line 310: This panics on the following assertion;

https://github.com/rust-lang/rust-analyzer/blob/a2bb0149144e0e6330e60abe4683b4e1a7ab3582/crates/syntax/src/syntax_editor.rs#L54-L55

I think that the panic happens when we are here in the following process;

1. First, we insert a new empty generic arg list because the function doesn't have one
2. And then we add a new type parameter for impl Trait thing

In 2. we don't have the new empty generic arg list made in 1. in our syntax tree yet (it will be made when the syntax edit finishes) so the position for new type parameter doesn't exist in our syntax tree yet either.
So, I think that we should directly insert a new generic arg list with new type param in 1. without separated step 2.

(Also, the current logic seems problematic beside the above because we are inserting `<T>` into `<>`, instead of `T`)

crates/syntax/src/ast/edit_in_place.rs
```diff
@@ -257,6 +290,28 @@ impl ast::GenericParamList {
...
+        new_param: ast::GenericParam,
+    ) {
+        match self.generic_params().last() {
+            Some(_) => {
+                let mut params =
+                    self.generic_params().map(|param| param.clone()).collect::<Vec<_>>();
+                params.push(new_param.into());
+                let new_param_list = make::generic_param_list(params);
+
+                editor.replace(self.syntax(), new_param_list.syntax());
```
ShoyuVanilla, line 305: It also panics here with rowan's mutable syntax tree thing that the existing generic params list is not mutable.
As the current `SyntaxEditor` is transitional thing, we might have to mark it with something like `clone_for_update` or something I guess.
But I think that an in-place inserting like the following would also work
```suggestion
                let position = self.r_angle_token().map_or_else(
                    || crate::syntax_editor::Position::last_child_of(self.syntax()),
                    crate::syntax_editor::Position::before,
                );
                let elements = vec![
                    make::token(SyntaxKind::COMMA).into(),
                    make::token(SyntaxKind::WHITESPACE).into(),
                    new_param.syntax().clone().into(),
                ];
                editor.insert_all(position, elements);
```

crates/ide-assists/src/handlers/introduce_named_generic.rs
```diff
@@ -61,6 +67,9 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext<'_>
                     edit.add_tabstop_before(cap, generic_param);
```
ShoyuVanilla, line 67: The weird snippet cursor position(`$0`) thing is caused in here.
As I commented on the first panic comment, our code is changed when the `SyntaxEditor::finish` is called.
So here, the last element of generic params is pre-existing one, not our new one. (Sorry for my misleading comment about this weeks ago 😅)
I think that we should change `add_tabstop_before` as entried like `add_file_edits` and triggered in order

ShoyuVanilla: Oh, I was wrong again 😅 Sorry. I think that we should use `SyntaxEditor::add_annotation` here (Find its usages for examples)

crates/ide-assists/src/handlers/introduce_named_generic.rs
tareknaser: > Sorry for the really late reply. I've been quite busy 'cause I moved to a new job 😅


No worries at all. Best of luck with your new job

crates/syntax/src/ast/edit_in_place.rs
tareknaser: So just for context, there’s been some [discussion about this in Zulip](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/.60SyntaxEditor.60.20replacement.20for.20Mutable.20Syntax.20trees/near/483142487).

> In 2. we don't have the new empty generic arg list made in 1. in our syntax tree yet (it will be made when the syntax edit finishes) so the position for new type parameter doesn't exist in our syntax tree yet either.

That was exactly the issue. I updated the code to add a method, `ast::Fn::syntax_editor_add_generic_param`, that handles the logic for us. Now, there’s a clear separation between the three cases in this method:
- If there is an existing generic param list and it’s not empty, we replace the entire list with a new one (existing params + the new one).
- If there’s an existing but empty generic param list, we place the new param after the left angle token.
- If there is no generic param list, we create one from scratch and place it in the right position.

This is working fine for all these cases, but the only issue left is the cursor position—it’s off in all cases.

I updated the code to use `SyntaxEditor::add_annotation()` as suggested https://github.com/rust-lang/rust-analyzer/pull/18483#discussion_r1856919754 , but I'm still facing the cursor issue. After tracing it a bit, it seems that this might be happening because `fn_.generic_param_list()` isn’t being updated. If that’s the case, how can I update it?

crates/syntax/src/ast/edit_in_place.rs
```diff
@@ -54,6 +55,80 @@ impl GenericParamsOwnerEdit for ast::Fn {
...
+                                make::token(SyntaxKind::COMMA),
+                            );
+                        } else {
+                            let elements = vec![
+                                make::token(SyntaxKind::COMMA).into(),
+                                make::token(SyntaxKind::WHITESPACE).into(),
+                                new_param.syntax().clone().into(),
+                            ];
+                            editor.insert_all(position, elements);
+                        }
```
tareknaser, line 99: For some reason, inserting this as a vector of elements like:
```rust
let elements = vec![
	make::token(SyntaxKind::COMMA).into(),
	make::token(SyntaxKind::WHITESPACE).into(),
	new_param.syntax().clone().into(),
];
editor.insert_all(position, elements);
```

causes only one test to fail (`replace_impl_trait_with_multiline_generic_params`). I’m not sure why.

crates/ide-assists/src/handlers/introduce_named_generic.rs
```diff
@@ -25,42 +25,49 @@ pub(crate) fn introduce_named_generic(acc: &mut Assists, ctx: &AssistContext<'_>
...
-                .clone_for_update();
-            let new_ty = make::ty(&type_param_name).clone_for_update();
+            let type_param = make.type_param(make.name(&type_param_name), Some(type_bound_list));
+            let new_ty = make.ty(&type_param_name);
 
-            ted::replace(impl_trait_type.syntax(), new_ty.syntax());
-            fn_generic_param_list.add_generic_param(type_param.into());
+            editor.replace(impl_trait_type.syntax(), new_ty.syntax());
+            let generic_param = syntax::ast::GenericParam::from(type_param);
+            fn_.syntax_editor_add_generic_param(&mut editor, generic_param.clone());
```
darichey, line 63:
```suggestion
            fn_.syntax_editor_add_generic_param(&mut editor, type_param.clone().into());
```

darichey (changes requested):

crates/ide-assists/src/handlers/introduce_named_generic.rs
tareknaser: Then we would need to clone `type_param` on the previous line, so it would be:
```rust
let generic_param = syntax::ast::GenericParam::from(type_param.clone());
fn_.syntax_editor_add_generic_param(&mut editor, type_param.clone().into());
```
Do you prefer this over the current code?

crates/syntax/src/ast/edit_in_place.rs
```diff
@@ -54,6 +55,80 @@ impl GenericParamsOwnerEdit for ast::Fn {
...
+        match self.generic_param_list() {
+            Some(generic_param_list) => match generic_param_list.generic_params().last() {
+                Some(_last_param) => {
+                    // There exists a generic param list and it's not empty
+                    let position = generic_param_list.r_angle_token().map_or_else(
+                        || crate::syntax_editor::Position::last_child_of(self.syntax()),
+                        crate::syntax_editor::Position::before,
+                    );
+
+                    if let Some(last_param) = generic_param_list.generic_params().last() {
```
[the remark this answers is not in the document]
tareknaser: Of course! I’m not sure how I missed that 😃
Thanks for catching it

crates/syntax/src/ast/edit_in_place.rs
```diff
@@ -54,6 +55,80 @@ impl GenericParamsOwnerEdit for ast::Fn {
     }
 }
 
+impl ast::Fn {
+    /// Adds a new generic param to the function using `SyntaxEditor`
+    pub fn syntax_editor_add_generic_param(
```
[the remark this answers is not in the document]
tareknaser: Thank you for the review

> but I do think `edit_in_place.rs` is probably the wrong place.

I completely agree. I actually mentioned this earlier here: https://github.com/rust-lang/rust-analyzer/pull/18483#discussion_r1838496721

I got a bit carried away focusing on getting the assist to work and forgot about this.

So, I’ve now moved it to a method of `SyntaxEditor`, but I created a new file for these edits, similar to `crates/syntax/src/ast/edit_in_place.rs`. I thought this might be a better place since `crates/syntax/src/syntax_editor.rs` mainly contains editing operations like `new`, `merge`, `insert`, `replace`, `delete`, etc.

But let me know if you’d prefer the method to be in `crates/syntax/src/syntax_editor.rs`, and I can update the code.

crates/ide-assists/src/handlers/introduce_named_generic.rs
darichey: I think we can just remove `generic_param` and replace with `type_param` below

tareknaser: We actually use it later to add annotation

darichey: Right, but can we add the annotation using `type_param` instead?

crates/syntax/src/ast/edit_in_place.rs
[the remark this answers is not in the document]
darichey: That seems fine to me! We can always move things around later.

crates/ide-assists/src/handlers/introduce_named_generic.rs
tareknaser: We can. should be fixed now

crates/syntax/src/syntax_editor/edits.rs
```diff
@@ -0,0 +1,72 @@
+//! Structural editing for ast using `SyntaxEditor`
+
+use crate::{
+    ast::make, ast::AstNode, ast::Fn, ast::GenericParam, ast::HasGenericParams, ast::HasName,
+    syntax_editor::Position, syntax_editor::SyntaxEditor, SyntaxKind,
+};
+
+impl SyntaxEditor {
+    /// Adds a new generic param to the function using `SyntaxEditor`
+    pub fn syntax_editor_add_generic_param(&mut self, function: Fn, new_param: GenericParam) {
```
darichey, line 10: last nit :) 
```suggestion
    pub fn add_generic_param(&mut self, function: Fn, new_param: GenericParam) {
```

you can probably also take `&Fn`

tareknaser: > you can probably also take &Fn

Yes

PR #18486, opened 2024-11-06
editors/code: Change minimum VS Code from 1.78 to 1.83

Wilfred:
It's been a year since we last bumped this (see #15904), and VS Code 1.83 is the first version that supports LSP 3.17.5 (via vscode-languageclient 9.0.1).

https://code.visualstudio.com/updates/v1_83#_language-server-protocol

(@davidbarsky and I work together, so I'm sure he'll be happy with this version bump too.)


lnicola (approved):
Thanks!

PR #18495, opened 2024-11-08
Migrate `reorder_fields` Assist to Use `SyntaxFactory`

tareknaser:
This PR is part of https://github.com/rust-lang/rust-analyzer/issues/15710 and https://github.com/rust-lang/rust-analyzer/issues/18285

### DIFF
--- a/crates/ide-assists/src/handlers/reorder_fields.rs
+++ b/crates/ide-assists/src/handlers/reorder_fields.rs
@@ -1,7 +1,7 @@
 use either::Either;
 use ide_db::FxHashMap;
 use itertools::Itertools;
-use syntax::{ast, ted, AstNode, SmolStr, ToSmolStr};
+use syntax::{ast, syntax_editor::SyntaxEditor, AstNode, SmolStr, SyntaxElement, ToSmolStr};
 
 use crate::{AssistContext, AssistId, AssistKind, Assists};
 
@@ -23,6 +23,11 @@ pub(crate) fn reorder_fields(acc: &mut Assists, ctx: &AssistContext<'_>) -> Opti
     let path = ctx.find_node_at_offset::<ast::Path>()?;
     let record =
         path.syntax().parent().and_then(<Either<ast::RecordExpr, ast::RecordPat>>::cast)?;
+
+    let parent_node = match ctx.covering_element() {
+        SyntaxElement::Node(n) => n,
+        SyntaxElement::Token(t) => t.parent()?,
+    };
 
     let ranks = compute_fields_ranks(&path, ctx)?;
     let get_rank_of_field = |of: Option<SmolStr>| {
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/reorder_fields.rs
```diff
@@ -65,23 +74,31 @@ pub(crate) fn reorder_fields(acc: &mut Assists, ctx: &AssistContext<'_>) -> Opti
...
 }
 
 fn replace<T: AstNode + PartialEq>(
+    editor: &mut SyntaxEditor,
     fields: impl Iterator<Item = T>,
     sorted_fields: impl IntoIterator<Item = T>,
 ) {
     fields.zip(sorted_fields).for_each(|(field, sorted_field)| {
-        ted::replace(field.syntax(), sorted_field.syntax().clone_for_update())
+        editor.replace(field.syntax(), sorted_field.syntax().clone_for_update())
```
darichey, line 101: We need to eliminate this `clone_for_update` using `SyntaxFactory`. It looks like we will need to add constructors for `record_expr_field` and `record_pat_field`. For now, I would inline this `replace` function so you don't need to worry about how to abstract it over the two cases. We can try to clean it up afterwards.

darichey (changes requested):

crates/ide-assists/src/handlers/reorder_fields.rs
DropDemBits: @darichey Generally we'd like to hide the details of creating new nodes under `SyntaxFactory`. In this case though, we're picking nodes from the original syntax tree so we wouldn't need to go through a `SyntaxFactory`.

`SyntaxEditor` should handle transform any node in an edit that reference the original syntax tree into their mutable equivalents, but seems like I forgot this case for the new nodes in replace edits 😅. For now it should be okay to use `clone_for_update` and add a fixme, and then we can remove those fixmes once `SyntaxEditor` is fixed.

darichey: Got it, thank you! Sorry about that, @tareknaser, you can revert inlining `replace` and just add a FIXME for the clone_for_update :)

crates/ide-assists/src/handlers/reorder_fields.rs
```diff
@@ -65,26 +71,28 @@ pub(crate) fn reorder_fields(acc: &mut Assists, ctx: &AssistContext<'_>) -> Opti
...
 
-fn replace<T: AstNode + PartialEq>(
-    fields: impl Iterator<Item = T>,
-    sorted_fields: impl IntoIterator<Item = T>,
-) {
-    fields.zip(sorted_fields).for_each(|(field, sorted_field)| {
-        ted::replace(field.syntax(), sorted_field.syntax().clone_for_update())
-    });
-}
-
```
darichey, line 87: So just bring this back and add like
```rust
// FIXME: remove clone_for_update when SyntaxEditor handles it for us
editor.replace(field.syntax(), sorted_field.syntax().clone_for_update());
```

crates/ide-assists/src/handlers/reorder_fields.rs
```diff
@@ -24,6 +24,12 @@ pub(crate) fn reorder_fields(acc: &mut Assists, ctx: &AssistContext<'_>) -> Opti
     let record =
         path.syntax().parent().and_then(<Either<ast::RecordExpr, ast::RecordPat>>::cast)?;
 
+    let parent_node = match ctx.covering_element() {
+        SyntaxElement::Node(n) => n,
+        SyntaxElement::Token(t) => t.parent()?,
+    };
+    let make = SyntaxFactory::new();
```
darichey, line 31: And in that case, we don't need a SyntaxFactory
```suggestion
```

crates/ide-assists/src/handlers/reorder_fields.rs
```diff
@@ -65,26 +71,28 @@ pub(crate) fn reorder_fields(acc: &mut Assists, ctx: &AssistContext<'_>) -> Opti
...
+                    });
+                }
+                Either::Right((sorted, field_list)) => {
+                    field_list.fields().zip(sorted).for_each(|(field, sorted_field)| {
+                        editor.replace(field.syntax(), sorted_field.syntax().clone_for_update());
+                    });
+                }
             }
+
+            editor.add_mappings(make.finish_with_mappings());
```
darichey, line 90:
```suggestion
```

darichey (changes requested):
Sorry for the churn, this should be good after this!

crates/ide-assists/src/handlers/reorder_fields.rs
tareknaser: All set 5c41c20

darichey (approved):
Awesome!

PR #18501, opened 2024-11-11
Revert "Resolve completion items"

### LINKED ISSUE #18363: Auto-imports not working in Kate
**rust-analyzer version**: rust-analyzer 1 (9323b53858 2024-10-20)

**rustc version**: rustc 1.84.0-nightly (662180b34 2024-10-20)

**editor or extension**: Kate 24.08.2 on Arch linux

Auto-imports not working in Kate

Downgrading to rust-analyzer 2024-09-30 fixes it. In versions after that nothing happens.

traviscross:
Reverting this commit fixes automatic importing in many editors.

Certainly there's probably a better way to fix this than a straight revert.  Feel free to close this, of course, if that path is taken.

This fixes #18363.

This reverts commit 950bb83f818342d9f9ed472c1d30492db324f5a6.

### DIFF
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -11,9 +11,9 @@ use std::{
 use anyhow::Context;
 
 use ide::{
-    AnnotationConfig, AssistKind, AssistResolveStrategy, Cancellable, CompletionFieldsToResolve,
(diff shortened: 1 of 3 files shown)


SomeoneToIgnore:
> Certainly there's probably a better way to fix this than a straight revert.

I would want to discuss that first: https://github.com/rust-lang/rust-analyzer/issues/18363#issuecomment-2467491145

to understand which properties are missing — then, at worst, we can always send back that particular subset always instead of the entire field list (as the revert will force us to do).

PR #18511, opened 2024-11-14
Re-add `rust-analyzer.cargo.sysrootQueryMetadata`

darichey:
#17795 removed this option, turning it on for everyone. This commit adds it back, still on by default.

At work, users are effectively offline, so the initial `cargo metadata` will always fail. The `--no-deps` retry will succeed, but the experience with this path is very poor, because a ton of std stuff fails to resolve (some still works, I can't tell the pattern). Something like this is probably happening in #17759.
Screenshot 2024-11-14 at 12 32 23 PM

We want this config option so we can turn it off by default for our users and always use the stitched sysroot library.

### DIFF
--- a/crates/project-model/src/cargo_workspace.rs
+++ b/crates/project-model/src/cargo_workspace.rs
@@ -13,7 +13,7 @@ use serde_json::from_value;
 use span::Edition;
 use toolchain::Tool;
 
-use crate::{utf8_stdout, ManifestPath, Sysroot};
+use crate::{utf8_stdout, ManifestPath, Sysroot, SysrootQueryMetadata};
 use crate::{CfgOverrides, InvocationStrategy};
 
 /// [`CargoWorkspace`] represents the logical structure of, well, a Cargo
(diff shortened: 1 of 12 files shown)


darichey:
> The `--no-deps` retry will succeed, but the experience with this path is very poor, because a ton of std stuff fails to resolve

In fact, this experience is so poor, I wonder if it's even worth supporting?

darichey:
If we'd like to get rid of the stitched sysroot stuff, a longer term solution could be teaching rust-analyzer to use something other than `cargo metadata` for this. At work, the sysroot builds with Buck, so presumably we could use it to get the info we need.

davidbarsky (changes requested):
I'm a plus one on this, but I'd prefer someone

lnicola (approved):

davidbarsky:
> I know, but I wasn't able to merge it because of your pending "changes requested" review.

Wild, okay! I thought it was overridable.

lnicola:
I know, check out my struggles in the history above 😕.

davidbarsky:
> I know, check out my struggles in the history above 😕.

I didn't see those in email and I didn't stroll up. I'm sorry about that.

Veykril:
> At work, users are effectively offline, so the initial `cargo metadata` will always fail. The `--no-deps` retry will succeed, but the experience with this path is very poor, because a ton of std stuff fails to resolve (some still works, I can't tell the pattern). 

Hmm, that seems very odd to me? The no-deps retry should yield the same result as the stitched workspace

Veykril:
In fact, I can't reproduce this. If I run r-a offline I still get resolved standard library stuff. I'd really prefer to get rid of this config as I'd like to drop the stitched workspace hack in the near future

PR #18514, opened 2024-11-15
minor: Remove impossible function pointer qualifiers from grammar

Giga-Bowser:
[Rust doesn't support `const` or `async` function pointers](https://github.com/rust-lang/rust/blob/76fd47124b8f8037b6187169b2cdf39139466952/compiler/rustc_parse/src/parser/ty.rs#L604-L609), nor does [our parser](https://github.com/rust-lang/rust-analyzer/blob/fc98e0657abf3ce07eed513e38274c89bbb2f8ad/crates/parser/src/grammar/types.rs#L217-L244).


lnicola:
We might want to keep these so we can offer diagnostics like `rustc`: "error: an `fn` pointer type cannot be `async`".

Giga-Bowser:
That diagnostic is offered as it stands, with an accompanying assist, although with a lot of cruft afterwards. We could probably just eat the tokens to fix this. Is there a reason to emit a proper error from the parser rather than relying on the diagnostic from `rustc`?
image

lnicola:
I like to disable `check` and rely on the native r-a diagnostics. They're (arguably) faster, and get displayed on-the-fly, not just when saving the file.

Ideally, we would be able to reuse the rustc diagnostic. Not going to happen, but we do have a couple of them implemented.

Giga-Bowser:
Whoops, of course I pushed that right before seeing your reply. I'll eventually make a new PR to generate a proper rust-analyzer diagnostic for this. For now, I'll just close this and handle this stuff there.

Issue #16084, opened 2023-12-11
Erroneous right paren when annotating closure with tuple return type


Lunderberg:
**Reproduction Steps**

1. Create a new project with `cargo new temp`.
2. In `bin.rs`, define the following function:
```rust
fn func() {
    let closure = |a: i64| {
        let b = 2 * a;
        (a, b)
    };
}
```
3. Place the cursor after `|a: i64|`, just before the opening `{`.
4. Type `-> (i64, i64) `

**Expected behavior**
After adding the return type annotation to the closure, the editor should have the following text:
```rust
fn func() {
    let closure = |a: i64| -> (i64, i64) {
        let b = 2 * a;
        (a, b)
    };
}
```

**Observed behavior**
After adding the return type annotation to the closure, the editor has the following text.
```rust
fn func() {
    let closure = |a: i64| -> (i64, i64) {
        let b = 2 * a;
        (a, b)
    });
}
```
The extra closing `)` occurs at the end of the closure definition, causing a compilation error due to unmatched parentheses.  Because the insertion of the extra `)` may be several lines away from the point where the user is typing, this is difficult to spot as it occurs.

**Suspected Mechanism**

1. The opening `(` in `(i64, i64)` triggers the [`on_char_typed`](https://github.com/rust-lang/rust-analyzer/blob/master/crates/ide/src/typing.rs#L63), which then delegates to [`on_opening_bracket_typed`](https://github.com/rust-lang/rust-analyzer/blob/master/crates/ide/src/typing.rs#L98)
2. `on_opening_bracket_typed` calls [`bracket_expr`](https://github.com/rust-lang/rust-analyzer/blob/master/crates/ide/src/typing.rs#L158), to determine if the `(` occurred just prior to an expression that should be parenthesized.
3. `bracket_expr` recognizes the body of the lambda as a block expression, and inserts a closing `)` between the closing `}` of the lambda and the `;`

**rust-analyzer version**:
* First noticed in `rust-analyzer 1.74.1 (a28077b 2023-12-04)`
* Last nightly build without the error: `rust-analyzer 1.74.0-nightly (203c57d 2023-09-17)`
* First nightly build with the error: `rust-analyzer 1.74.0-nightly (65ea825 2023-09-18)`

[...]

PR #18522, opened 2024-11-17
Add Configurable Option to Exclude Trigger Characters for Typing Assists

tareknaser:
This PR addresses #16084 by adding a new configuration option: `rust-analyzer.typing.excludeChars`.

cc: @darichey 
### Changes
- Added `rust-analyzer.typing.excludeChars` configuration.
    - Defaults to `None`, meaning all trigger characters (`.=<>{(`) are active by default.
    - Users can specify a string of characters to exclude from triggering assists.
- Updated documentation and configuration schema to reflect this new option.

### DIFF
--- a/crates/ide/src/lib.rs
+++ b/crates/ide/src/lib.rs
@@ -410,14 +410,10 @@ impl Analysis {
         &self,
         position: FilePosition,
         char_typed: char,
-        autoclose: bool,
         chars_to_exclude: Option<String>,
     ) -> Cancellable<Option<SourceChange>> {
         // Fast path to not even parse the file.
         if !typing::TRIGGER_CHARS.contains(char_typed) {
-            return Ok(None);
-        }
-        if char_typed == '<' && !autoclose {
             return Ok(None);
         }
         if let Some(chars_to_exclude) = chars_to_exclude {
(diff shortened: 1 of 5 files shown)


darichey:
I think this is a good workaround for people who want some functionality of format-on-type but find the limitations/bugs (like the one presented in the linked issue) annoying. 

I would reword "typing assists" to "format-on-type"

tareknaser:
> I would reword "typing assists" to "format-on-type"

Thanks for the review.

I initially thought `rust-analyzer.typing.excludeChars` made sense because it's similar to `rust-analyzer.typing.autoClosingAngleBrackets.enable`.

Do you think it should be renamed to `rust-analyzer.formatOnType.excludeChars` instead? If so, should we also rename the other configurable option to `rust-analyzer.formatOnType.autoClosingAngleBrackets.enable`?

Veykril:
This setting should replace `rust-analyzer.typing.autoClosingAngleBrackets.enable` imo (the main reason that setting exists is because the typing for angle brackets is bugged atm)

tareknaser:
> This setting should replace `rust-analyzer.typing.autoClosingAngleBrackets.enable` imo

Sure. I’ve just pushed a commit to remove it from the code. I’m not sure if there’s a specific process for deprecation like this. If there is, please let me know and I can update the code

Veykril (approved):

PR #18531, opened 2024-11-18
fix: Map new replacement nodes to their mutable equivalents in `SyntaxEditor`

### LINKED ISSUE #18525: Map new nodes in replacement edits to their mutable equivalents in SyntaxEditor
`SyntaxEditor` should handle transform any node in an edit that reference the original syntax tree into their mutable equivalents, but seems like I forgot this case for the new nodes in replace edits 😅. For now it should be okay to use `clone_for_update` and add a fixme, and then we can remove those fixmes once `SyntaxEditor` is fixed.

_Originally posted by @DropDemBits in https://github.com/rust-lang/rust-analyzer/pull/18495#discussion_r1838996205_

Giga-Bowser:
closes #18525

Obviously there are lots of in-progress PRs right now which need updating as a result of this


Veykril (approved):
Thanks!

PR #18538, opened 2024-11-20
Migrate `sort_items` Assist to Use `SyntaxFactory`

tareknaser:
part of #15710 and #18285
For this one, I had to pass `AssistContext` to other methods as well in order to get the file ID.

### DIFF
--- a/crates/ide-assists/src/handlers/sort_items.rs
+++ b/crates/ide-assists/src/handlers/sort_items.rs
@@ -4,7 +4,7 @@ use itertools::Itertools;
 
 use syntax::{
     ast::{self, HasName},
-    ted, AstNode, TextRange,
+    AstNode, SyntaxNode,
 };
 
 use crate::{utils::get_methods, AssistContext, AssistId, AssistKind, Assists};
@@ -114,7 +114,7 @@ trait AddRewrite {
         label: &str,
         old: Vec<T>,
         new: Vec<T>,
-        target: TextRange,
+        target: SyntaxNode,
     ) -> Option<()>;
 }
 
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/sort_items.rs
```diff
@@ -121,24 +122,33 @@ trait AddRewrite {
 impl AddRewrite for Assists {
     fn add_rewrite<T: AstNode>(
         &mut self,
+        ctx: &AssistContext<'_>,
         label: &str,
         old: Vec<T>,
         new: Vec<T>,
         target: TextRange,
     ) -> Option<()> {
+        let node = old.first().unwrap().syntax().parent().unwrap();
```
darichey, line 131: Try instead modifying `add_rewrite` to accept `target: SyntaxNode`

crates/ide-assists/src/handlers/sort_items.rs
```diff
@@ -121,24 +122,33 @@ trait AddRewrite {
...
-                .zip(new)
-                .for_each(|(old, new)| ted::replace(old.syntax(), new.clone_for_update().syntax()));
+            let mut editor = builder.make_editor(&node);
+
+            old.into_iter().zip(new).for_each(|(old, new)| {
+                // FIXME: remove `clone_for_update` when `SyntaxEditor` handles it for us
+                editor.replace(old.syntax(), new.clone_for_update().syntax())
+            });
+
+            builder.add_file_edits(ctx.file_id(), editor)
```
darichey, line 140: I think we can get around passing the `ctx` everywhere but just using the `file_id` from the builder:
```suggestion
            builder.add_file_edits(builder.fild_id, editor)
```

darichey (changes requested):
Mostly looks good, a couple opportunities for simplification.

Not for this PR, but after doing a few of these, it looks like there may be potential for a slightly nicer api that takes care of some things for us. Something like:
```rust
fn add_with_editor(
    &mut self,
    id: AssistId,
    label: impl Into<String>,
    target: SyntaxNode,
    f: impl FnOnce(&mut SyntaxEditor, &SyntaxFactory),
) -> Option<()>
```

crates/ide-assists/src/handlers/sort_items.rs
tareknaser: Right, updated

crates/ide-assists/src/handlers/sort_items.rs
tareknaser: Updated.
But, `Assists::add` actually accepts `TextRange` as a parameter, so I had to use `target.text_range()` for that. Let me know if this works better.

darichey: Awesome, and now you can call `make_editor` with `target` instead!

Also, you can probably take `target: &SyntaxNode` instead to avoid the clones at the callers

darichey (changes requested):

crates/ide-assists/src/handlers/sort_items.rs
tareknaser: Right. updated

darichey (approved):
lgtm! cc @davidbarsky

davidbarsky (approved):

PR #18539, opened 2024-11-20
feat: Add `remove_braces` assist to remove braces from closures and match arms

Giga-Bowser:
The changes made to the handling of match arm commas resulted in broader changes than I would've liked, but I think it's worthwhile in this case.


Veykril:
Closing due to inactivity, feel free to come back to this whenever you want

PR #18548, opened 2024-11-23
Update file path for request.rs entry point in architecture.md

markmurphydev:
The path to the request handler file is currently `crates/rust-analyzer/src/handlers/request.rs`. Updated `docs/dev/architecture.md` to reflect that.


lnicola (approved):
Thanks!

PR #18575, opened 2024-11-29
minor: Migrate `flip_*` assists to `SyntaxEditor`

Giga-Bowser:
Token tree handling is a bit annoying but I don't think there's anything you can do about it, that's just the nature of raw token trees.


Veykril (approved):
Thanks!

PR #18589, opened 2024-12-03
Advertise completions and inlay hints resolve server capabilities based on the client capabilities

SomeoneToIgnore:


Veykril (approved):

Veykril (approved):

bstaletic:
@SomeoneToIgnore Thank you! I have just tested with my client and this pull request resolves the regression I have seen.

Veykril:
I'm not up to speed with what happened the past weeks wrt to this but does this also need a beta backport?

SomeoneToIgnore:
So far it seems that most of the completion-related issues were in "rich" editors that were always requiring the `additionalTextEdits`, so this fix changes nothing for them and is indeed, somewhat separate.

I am trying not to do anything extra on this front without the confirmation now, so let me know if you think it's a good idea to backport.
For that, I need to open this PR against the `rustc:beta` branch, don't I?

lnicola:
Wasn't this disabled on beta in https://github.com/rust-lang/rust/pull/133546?

SomeoneToIgnore:
Oh, sorry, I'm sick and not thinking straight.
It was, indeed, so we're "good" for now modulo inlays' resolve (which no one seem to care about given how much time it's there)

lnicola:
I'd say not backport if it's not strictly needed, it's fine if it hits stable in the next version.

Veykril:
Ah okay, then I agree. No need to backport this 👍

bstaletic:
May I ask if there's any policy regarding when this change will be available through rustup-installed rust-anayzer?
I'd like to get these changes into the default config of my client. To be honest, I'm not in a hurry.

lnicola:
I think the problematic code was disabled in 1.83. I'll try to do another subtree sync tomorrow, so this should hit nightly in a couple of days.

PR #18592, opened 2024-12-03
vscode: Only show status bar item in relevant files

darichey:
The status bar can become a bit crowded with many extensions adding status bar items. rust-analyzer can be a good citizen and hide itself unless the user is looking at relevant files.

I've set the default to show on Rust files and Cargo.toml. It could be reasonable to instead keep the current always-show behavior as the default, just giving users the option to configure it. I'm open to discussion!

This could also be a workaround for #15068.

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -425,6 +425,41 @@
                         ],
                         "default": "openLogs",
                         "markdownDescription": "Action to run when clicking the extension status bar item."
+                    },
+                    "rust-analyzer.statusBar.documentSelector": {
+                        "type": [
+                            "array",
+                            "null"
+                        ],
+                        "items": {
+                            "type": "object",
+                            "properties": {
+                                "language": {
+                                    "type": [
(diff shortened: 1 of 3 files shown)


editors/code/package.json
```diff
@@ -425,6 +425,38 @@
...
+                                    ]
+                                }
+                            }
+                        },
+                        "default": [
+                            {
+                                "language": "rust"
+                            },
+                            {
+                                "pattern": "**/Cargo.toml"
```
Veykril, line 456: `**/Cargo.lock` might also be a good fit

bjorn3:
So this is why I couldn't find the status bar item for rust-analyzer in some cases even when rust-analyzer is busy running. This makes it harder to stop rust-analyzer when it is interfering with things.

Veykril:
I am honestly tempted to change the default to always, I've seen a lot of people confused by this behavior (rightfully so)

PR #18593, opened 2024-12-03
Fix parsing of parenthesized type args and RTN

Veykril:
Big diff due to test changes


Veykril:
Hmm, this regressed type inference on diesel

PR #18608, opened 2024-12-04
Better parser recovery for paths

Veykril:


Veykril:
Looking at the test changes here, our path parsing recovery is really really bad, we shouldn't be recovering into empty nodes either

PR #18630, opened 2024-12-06
fix: Temporarily disable completion resolve support for helix and neovim

Veykril:
cc https://github.com/rust-lang/rust-analyzer/issues/18547#issuecomment-2523527080

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -730,6 +730,12 @@ enum RatomlFile {
     Crate(LocalConfigInput),
 }
 
+#[derive(Clone, Debug)]
+struct ClientInfo {
+    name: String,
+    version: Option<Version>,
+}
+
 #[derive(Clone)]
 pub struct Config {
     /// Projects that have a Cargo.toml or a rust-project.json in a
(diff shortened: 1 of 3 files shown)


crates/rust-analyzer/src/config.rs
```diff
@@ -2163,7 +2174,18 @@ impl Config {
...
+            .filter(|it| it.name.starts_with("Visual Studio Code"))
+            .and_then(|it| it.version.as_ref())
+    }
+
+    pub fn client_is_helix(&self) -> bool {
+        self.client_info.as_ref().map(|it| it.name == "helix").unwrap_or_default()
+    }
+
+    pub fn client_is_neovim(&self) -> bool {
+        self.client_info.as_ref().map(|it| it.name == "Neovim").unwrap_or_default()
```
Veykril, line 2188: Found this string in one issue in https://github.com/hrsh7th/cmp-nvim-lsp/issues/36, so I'll just hope that that works

crates/rust-analyzer/src/lsp/capabilities.rs
```diff
@@ -41,7 +41,11 @@ pub fn server_capabilities(config: &Config) -> ServerCapabilities {
         })),
         hover_provider: Some(HoverProviderCapability::Simple(true)),
         completion_provider: Some(CompletionOptions {
-            resolve_provider: Some(config.caps().completions_resolve_provider()),
+            resolve_provider: if config.client_is_helix() || config.client_is_neovim() {
+                None
```
SomeoneToIgnore, line 45: This has to be `self.completion_item_edit_resolve().then_some(true)` or similar as it was before https://github.com/rust-lang/rust-analyzer/pull/18589 due to r-a actually always resolving the `additionalTextEdits`.

SomeoneToIgnore (approved):
Thank you for coming up with this, I would have been waiting for the spec clarifications still otherwise 😢

PR #18649, opened 2024-12-09
minor: enhance name suggestion for `Arc<T>` and `Rc<T>`

### LINKED ISSUE #18648: Enhance name suggestion for `Arc<T>` and `Rc<T>`
Currently we would suggest `arc` for `Arc<T>` and `rc` for `Rc<T>` in `add missing match arms`. 

It would be preferable to suggest names based on `T`.

roife:
fix #18648

### DIFF
--- a/crates/ide-db/src/syntax_helpers/suggest_name.rs
+++ b/crates/ide-db/src/syntax_helpers/suggest_name.rs
@@ -29,7 +29,7 @@ const USELESS_NAME_PREFIXES: &[&str] = &["from_", "with_", "into_"];
 /// # Examples
 /// `Option<Name>` -> `Name`
 /// `Result<User, Error>` -> `User`
-const WRAPPER_TYPES: &[&str] = &["Box", "Option", "Result"];
+const WRAPPER_TYPES: &[&str] = &["Box", "Arc", "Rc", "Option", "Result"];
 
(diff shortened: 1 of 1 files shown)


lnicola (approved):

crates/ide-db/src/syntax_helpers/suggest_name.rs
```diff
@@ -29,7 +29,7 @@ const USELESS_NAME_PREFIXES: &[&str] = &["from_", "with_", "into_"];
 /// # Examples
 /// `Option<Name>` -> `Name`
 /// `Result<User, Error>` -> `User`
-const WRAPPER_TYPES: &[&str] = &["Box", "Option", "Result"];
+const WRAPPER_TYPES: &[&str] = &["Box", "Arc", "Rc", "Option", "Result"];
```
Veykril, line 32: I wonder if there is some heuristic we can reliably apply here 🤔

lnicola: `AsRef` would cover the first three, but..

Veykril: `AsRef` is probably too general, likewise for `Deref`. I don't think there are any useful traits for this right now, so a list might be the best we have

PR #18652, opened 2024-12-09
feat: Add an assist to extract an expression into a constant

### LINKED ISSUE #7692: ‘Extract into constant’ assist
rust-analyzer currently has ‘Extract into variable’ and ‘Extract into function’ assists; I sometimes find myself using ‘Extract into variable’, renaming it to use a `SCREAMING_SNAKE_CASE` name, and adjusting the definition to use `const`. It would be nice to have an assist that does all this. Bonus: by default put the constant at the top of the file, after any `use` statements.

Giga-Bowser:
closes #7692

Technically, this will permit some invalid const definitions since, as far as I know, we don't have the ability to properly ensure that an expression can be const evaluated, but I believe that would require a larger overhaul of RA's integration of MIR stuff. Either way, this should be fine for now and does a good job rejecting most of the obviously incorrect definitions. 

I could also group these assists if that's desirable, just let me know.

### DIFF
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -1,14 +1,12 @@
 use hir::{HirDisplay, TypeInfo};
-use ide_db::syntax_helpers::suggest_name;
+use ide_db::{assists::GroupLabel, syntax_helpers::suggest_name};
 use syntax::{
     ast::{
         self, edit::IndentLevel, edit_in_place::Indent, make, syntax_factory::SyntaxFactory,
         AstNode,
     },
     syntax_editor::Position,
-    NodeOrToken,
-    SyntaxKind::{self},
-    SyntaxNode, T,
+    NodeOrToken, SyntaxKind, SyntaxNode, T,
 };
 
 use crate::{utils::is_body_const, AssistContext, AssistId, AssistKind, Assists};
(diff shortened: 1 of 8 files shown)


crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -7,11 +7,11 @@ use syntax::{
     },
     syntax_editor::Position,
     NodeOrToken,
-    SyntaxKind::{BLOCK_EXPR, BREAK_EXPR, COMMENT, LOOP_EXPR, MATCH_GUARD, PATH_EXPR, RETURN_EXPR},
+    SyntaxKind::{self},
```
Veykril, line 10:
```suggestion
    SyntaxKind,
```

crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -211,26 +238,76 @@ fn peel_parens(mut expr: ast::Expr) -> ast::Expr {
...
+        SyntaxKind::RETURN_EXPR => ast::ReturnExpr::cast(node).and_then(|e| e.expr()),
+        SyntaxKind::BLOCK_EXPR => {
             ast::BlockExpr::cast(node).filter(|it| it.is_standalone()).map(ast::Expr::from)
         }
         _ => ast::Expr::cast(node),
     }
 }
 
-#[derive(Debug)]
+enum ExtractionKind {
```
Veykril, line 251: Might as well cover `Static`s while we are at it

Veykril:
> Technically, this will permit some invalid const definitions since, as far as I know, we don't have the ability to properly ensure that an expression can be const evaluated, but I believe that would require a larger overhaul of RA's integration of MIR stuff. Either way, this should be fine for now and does a good job rejecting most of the obviously incorrect definitions.

That's fine, in that case we do want the assist to be more permissive I think. Its okay for assists to generate semantically invalid code (to a certain degree), its not okay for them to generate broken syntax.

Veykril:
> I could also group these assists if that's desirable, just let me know.

This I think we might want to do (grouped together with function extraction even as i believe those can trigger on the same selection snippets at times)

Giga-Bowser:
This force-push was just fixing merge conflicts, I will begin working on those suggestions now!

Veykril (approved):

PR #18653, opened 2024-12-09
Hash completion items to properly match them during /resolve

SomeoneToIgnore:
Closes https://github.com/rust-lang/rust-analyzer/issues/18547
Closes https://github.com/rust-lang/rust-analyzer/issues/18640

Follows https://github.com/rust-lang/rust-analyzer/issues/18547#issuecomment-2525373488 proposal and adds completion items hashing, using all their data possible, that does not change between the edits.

### DIFF
--- a/crates/ide-completion/src/item.rs
+++ b/crates/ide-completion/src/item.rs
@@ -346,8 +346,7 @@ pub enum CompletionItemKind {
 impl_from!(SymbolKind for CompletionItemKind);
 
 impl CompletionItemKind {
-    #[cfg(test)]
-    pub(crate) fn tag(self) -> &'static str {
+    pub fn tag(self) -> &'static str {
         match self {
             CompletionItemKind::SymbolKind(kind) => match kind {
                 SymbolKind::Attribute => "at",
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -36,6 +36,7 @@ use triomphe::Arc;
 use vfs::{AbsPath, AbsPathBuf, FileId, VfsPath};
 
 use crate::{
+    completion_item_hash,
     config::{Config, RustfmtConfig, WorkspaceSymbolConfig},
     diagnostics::convert_diagnostic,
     global_state::{FetchWorkspaceRequest, GlobalState, GlobalStateSnapshot},
@@ -1122,12 +1123,15 @@ pub(crate) fn handle_completion_resolve(
         return Ok(original_completion);
     };
     let source_root = snap.analysis.source_root_id(file_id)?;
+    let Some(completion_hash_for_resolve) = &resolve_data.completion_item_hash else {
+        return Ok(original_completion);
+    };
 
     let mut forced_resolve_completions_config = snap.config.completion(Some(source_root));
     forced_resolve_completions_config.fields_to_resolve = CompletionFieldsToResolve::empty();
 
     let position = FilePosition { file_id, offset };
-    let Some(resolved_completions) = snap.analysis.completions(
+    let Some(completions) = snap.analysis.completions(
         &forced_resolve_completions_config,
         position,
         resolve_data.trigger_character,
@@ -1135,6 +1139,14 @@ pub(crate) fn handle_completion_resolve(
     else {
         return Ok(original_completion);
     };
+
+    let Some(corresponding_completion) = completions.into_iter().find(|completion_item| {
+        let hash = completion_item_hash(&completion_item, resolve_data.for_ref);
+        &hash == completion_hash_for_resolve
+    }) else {
+        return Ok(original_completion);
+    };
+
     let mut resolved_completions = to_proto::completion_items(
         &snap.config,
         &forced_resolve_completions_config.fields_to_resolve,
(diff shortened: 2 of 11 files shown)


crates/rust-analyzer/src/lsp/to_proto.rs
```diff
@@ -393,16 +395,15 @@ fn completion_item(
...
     let (ref_resolve_data, resolve_data) = if something_to_resolve || !imports.is_empty() {
-        let mut item_index = acc.len();
         let ref_resolve_data = if ref_match.is_some() {
             let ref_resolve_data = lsp_ext::CompletionResolveData {
                 position: tdpp.clone(),
                 imports: Vec::new(),
                 version,
                 trigger_character: completion_trigger_character,
-                completion_item_index: item_index,
+                for_ref: true,
```
SomeoneToIgnore, line 404: Out of the same item, we can create two completion items, so keep another property in the resolve data to have the hashing code without any complex conditions but still distinct for both cases.

SomeoneToIgnore (commented):
My only concern is how relatively fragile this is, if another field is added or some fields start to alter between text edits, but otherwise it seems to work on VSCode and Helix during my manual testing.

cc @Veykril as it would be good to use this for some while to see if that sporadic error is gone with this build.

crates/rust-analyzer/src/lib.rs
```diff
@@ -61,3 +63,79 @@ pub fn from_json<T: DeserializeOwned>(
...
+    // NB: do not hash edits or source range, as those may change between the time the client sends the resolve request
+    // and the time it receives it: some editors do allow changing the buffer between that, leading to ranges being different.
+    hasher.update(item.kind.tag());
+    hasher.update(&item.lookup);
+    if let Some(detail) = &item.detail {
+        hasher.update(detail);
+    }
+    if let Some(documentation) = &item.documentation {
+        hasher.update(documentation.as_str());
+    }
```
Veykril, line 130: hashing this seems unnecessary to me, it doesn't really make completions more unique among another and is just a huge blob to hash

crates/rust-analyzer/src/lib.rs
```diff
@@ -61,3 +63,79 @@ pub fn from_json<T: DeserializeOwned>(
...
+    hash_completion_relevance(&mut hasher, &item.relevance);
+    if let Some((mutability, text_size)) = &item.ref_match {
```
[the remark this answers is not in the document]
Veykril: `text_size` is the offset to insert the borrow prefix in, so I think its safe here as that should always be in front of the cursor. Given completions don't popup by removing characters (shifting the cursor to the left of its identifier) it ought to be fine

crates/ide-completion/src/item.rs
```diff
@@ -346,8 +346,7 @@ pub enum CompletionItemKind {
 impl_from!(SymbolKind for CompletionItemKind);
 
 impl CompletionItemKind {
-    #[cfg(test)]
```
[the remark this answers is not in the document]
Veykril: Can't we just make `CompletionItemKind ` and `SymbolKind` `Hash`?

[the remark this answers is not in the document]
Veykril: Oh I see, `tenthash` doesn't accept hash things, only byte slices...

crates/rust-analyzer/src/lsp/to_proto.rs
Veykril: isn't most of the moved fields `Copy` here? I'd expect partial cloning to not be that bad

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -1127,30 +1128,34 @@ pub(crate) fn handle_completion_resolve(
...
         position,
         resolve_data.trigger_character,
     )?
     else {
         return Ok(original_completion);
     };
+
+    let Some(corresponding_completion) = completions.into_iter().find(|completion_item| {
+        let hash = completion_item_hash(completion_item, resolve_data.for_ref);
+        hash == resolve_data.hash
```
pascalkuthe, line 1142: I think may be a good idea to first check if the labels are identical and only compute and compare the hash if they are:

* this will reduce computational requirements a bit since hash is only computed for any realistic candidates
* as a bonus it makes collisons even more unlikely

crates/rust-analyzer/src/lsp/ext.rs
```diff
@@ -826,7 +826,8 @@ pub struct CompletionResolveData {
     pub imports: Vec<CompletionImport>,
     pub version: Option<i32>,
     pub trigger_character: Option<char>,
-    pub completion_item_index: usize,
+    pub for_ref: bool,
+    pub hash: [u8; 20],
```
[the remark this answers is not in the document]
pascalkuthe: I would do something like base64, that encoding is just a hex string which is pretty standard to use for hashes but not necessairily the most information dense (compared to base64)

pascalkuthe (commented):
this is a much more robust approach, thanks for the fast turnaround!

crates/rust-analyzer/src/lsp/to_proto.rs
SomeoneToIgnore: Thank you, indeed, it's the `documentation` and `import_to_add` fields partially moved that I had to clone, this way is definitely better.

crates/rust-analyzer/src/lsp/ext.rs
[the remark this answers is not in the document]
SomeoneToIgnore: Thanks, had applied Base64 and it does trim the character number by half:

Before:
"hash":[215,92,116,28,211,148,173,32,58,55,147,15,75,30,169,165,93,8,60,63]

After:
"hash":"11x0HNOUrSA6N5MPSx6ppV0IPD8="

Seems like a good thing to do.

pascalkuthe (approved):

crates/rust-analyzer/src/handlers/request.rs
SomeoneToIgnore: Nice idea.

It's a bit tricky because LSP completion items get their labels changed compared to the original ones, but luckily only by adding a suffix, so we can still reduce the computations with a `starts_with` check.

tgross35:
Would it be possible to do a sync soonish? I don't think either of the fixes have made it to r-l/rust so nightly is still buggy.

PR #18654, opened 2024-12-10
feat: Implement import ide layer handling

Veykril:
Only implements the type driven changes so far for threading things through. It doesn't actually resolve to imports yet as I am currently thinking about whether to put `Import` (and `ExternCrate`) as `ModuleDef` variants or not
Closes https://github.com/rust-lang/rust-analyzer/issues/14079

### DIFF
--- a/crates/hir-def/src/item_scope.rs
+++ b/crates/hir-def/src/item_scope.rs
@@ -16,10 +16,10 @@ use syntax::ast;
 
 use crate::{
     db::DefDatabase,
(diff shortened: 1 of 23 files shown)


Veykril:
Alright, after having a more thorough look, implementing this properly will be quite painful! We'd need to keep track of the import information throughout body lowering and type inference as we'd otherwise lose the link from where something has come from. I'll stash this away for now as I am not comfortable doing this right before the salsa changes.

PR #18660, opened 2024-12-11
fix: copied proc-macros not being cleaned up on exit

Veykril:


Veykril (commented):
Can anyone inform me about what's unsafe about https://docs.rs/memmap/latest/memmap/struct.Mmap.html#method.map, given the docs don't state it?

lnicola:
Is that the right crate? I suspect it's because another process can truncate or change the file while you're reading from it. See also https://docs.rs/memmap2/latest/memmap2/struct.Mmap.html#safety.

Veykril:
Ah okay, so I was right. That means technically our usages are unsound ...

lnicola:
I'm not to worried about the memory maps causing a crash. I think there's a very small window when we have the library mapped, but not loaded.

PR #18670, opened 2024-12-12
internal: Drop proc-macro server support for ~1.66.0 and older toolchains

Veykril:
That is toolchains older than ~20 months from today on where this version was released. Reason is to simplify the codebase, and we likely don't even support version that old due to other reasons anymore either

### DIFF
--- a/crates/proc-macro-api/src/process.rs
+++ b/crates/proc-macro-api/src/process.rs
@@ -56,8 +56,25 @@ impl ProcMacroProcessSrv {
         match srv.version_check() {
             Ok(v) if v > CURRENT_API_VERSION => Err(io::Error::new(
                 io::ErrorKind::Other,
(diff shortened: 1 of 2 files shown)


crates/proc-macro-api/src/process.rs
```diff
@@ -56,8 +56,25 @@ impl ProcMacroProcessSrv {
...
+            )),
+            Ok(v) if v < RUST_ANALYZER_SPAN_SUPPORT => Err(io::Error::new(
+                io::ErrorKind::Other,
+                format!(
+                    "The version of the proc-macro server ({v}) in your Rust toolchain \
+                is too old and no longer supported by your rust-analyzer which requires\
+                version {RUST_ANALYZER_SPAN_SUPPORT} or higher.
+\
+                This will prevent proc-macro expansion from working. \
+                Please consider updating your rust-analyzer to ensure compatibility with your \
```
lnicola, line 76: Should this say "downgrading your rust-analyzer"?

PR #18672, opened 2024-12-12
internal: Do not require a special env var to be set for the proc-macro-srv

Veykril:
It feels a bit silly to require this at this point


lnicola:
The variable was added because of concerns about the server not being as "stable" as the stable toolchain. Should we get this approved by the compiler or another team?

Veykril:
Hmm, right. I'll repurpose this for the other cleanup then only

Veykril:
i forgot to change the commit message whoops

PR #18676, opened 2024-12-12
fix: Fix parser trying to parse let exprs in invalid places

Veykril:
cc https://github.com/rust-lang/rust-analyzer/issues/18673


Veykril:
Hmm, I guess we do have the validation pass for now. That isn't quite correct as these ought to be parse errors but implementing them properly is annoying with our parser

PR #18715, opened 2024-12-18
Fix AsmOption rule in rust.ungram

aibaars:
@Veykril I think you forgot to move the `*` operator when translating `options := "options(" option *("," option) [","] ")"` to an ungrammar rule.


Veykril (approved):
I did, thanks :)

PR #18718, opened 2024-12-18
internal: Standardize how we take iterator parameters in `SyntaxFactory`

Giga-Bowser:
In my PR for migrating wrap/unwrap return type, [I remarked about two different ways of taking many elements as a parameter](https://github.com/rust-lang/rust-analyzer/pull/18524#issuecomment-2523440023), and it seemed like a matter of style at the time, but it turns out it actually matters a lot! Because if you provide a parameter taking `IntoIterator + Clone` using something a `.map()`, in which you construct additional elements using the same `SyntaxFactory` you're providing the parameter to, then by the lazy nature of iterators, there will be a looming mutable borrow on the `RefCell` for `mappings`, and the code will panic. So we need to be `collect`ing the iterator to force any internal references be dropped.

### DIFF
--- a/crates/syntax/src/ast/syntax_factory/constructors.rs
+++ b/crates/syntax/src/ast/syntax_factory/constructors.rs
@@ -1,6 +1,4 @@
 //! Wrappers over [`make`] constructors
-use itertools::Itertools;
-
 use crate::{
     ast::{self, make, HasGenericParams, HasName, HasTypeBounds, HasVisibility},
     syntax_editor::SyntaxMappingBuilder,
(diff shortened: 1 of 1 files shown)


crates/syntax/src/ast/syntax_factory/constructors.rs
```diff
@@ -512,3 +507,13 @@ impl SyntaxFactory {
         make::tokens::whitespace(text)
     }
 }
+
+fn iterator_input<N: AstNode>(input: impl IntoIterator<Item = N>) -> (Vec<N>, Vec<SyntaxNode>) {
```
Veykril, line 511: That sounds like there should be a comment on this function then explaining the rational like the PR does

PR #18722, opened 2024-12-19
Rename `rust-analyzer.statusBar.documentSelector` to `showStatusBar`, add "always" and "never" options.

### LINKED ISSUE #18665: Add setting to always show status bar item in vscode
Hello. Recent pull #18592 has changed the default behavior of the vscode's rust-lang.rust-analyzer extension to hide the status bar item unless certain types of files are open.

I believe it should have a setting to always show the status bar item. There might be multiple use cases to that, one of which is to know whether the server is running and start/stop it when needed (e.g. I don't need it all the time while vscode is open, but keeping it running wastes a significant portion of RAM (more than 1GiB even for a small project)), but opening a .rs/cargo.toml file every time to check it's status seems unnecessary.

Setting `rust-analyzer.statusBar.documentSelector` to `{"pattern": "**"}` doesn't work when no editors are open.

markmurphydev:
Closes #18665.

`rust-analyzer.statusBar.showStatusBar` can be `"always"`, `"never"` or `{ "documentSelector": <DocumentSelector>[] }`.
I added `DocumentSelector.notebookType` and `.scheme` as possible properties. The array is passed directly to `vscode.languages.match(documentSelector, editor.document)`, so the properties should Just Work.

Manually  testing, `"always"`, `"never"`, and `{ "documentSelector": ... }` seemed to work correctly.

I think the renamed setting is clearer, but it will break previous `statusBar.documentSelector` configs.

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -426,40 +426,55 @@
                         "default": "openLogs",
                         "markdownDescription": "Action to run when clicking the extension status bar item."
                     },
-                    "rust-analyzer.statusBar.documentSelector": {
-                        "type": [
-                            "array",
-                            "null"
-                        ],
-                        "items": {
-                            "type": "object",
-                            "properties": {
-                                "language": {
-                                    "type": [
-                                        "string",
-                                        "null"
-                                    ]
-                                },
-                                "pattern": {
-                                    "type": [
-                                        "string",
-                                        "null"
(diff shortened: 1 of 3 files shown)


Veykril (approved):

Veykril:
CI is red

PR #18727, opened 2024-12-20
fix: remove `always!` check for file_id in `runnables`

### LINKED ISSUE #18704: codeLens panic when working on a include!(source)
Here is a MRE for this bug:

```rust
// main.rs
fn main() {
    println!("Hello, world!");
}

include!("./out.rs");
```

```rust
// out.rs
#[test]
fn test_out(){
}
```
When you go to the `out.rs` file, A panic is triggered:

```
thread 'Worker' panicked at crates/ide/src/runnables.rs:134:13:
tried adding a runnable pointing to a different file: TestMod { path: "" } for FileId(2014)
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   3: ide::runnables::runnables
   4: ide::annotations::annotations
   5: ra_salsa::Cancelled::catch
   6: ide::Analysis::annotations
   7: rust_analyzer::handlers::request::handle_code_lens
   8: std::panic::catch_unwind
   9: core::ops::function::FnOnce::call_once{{vtable.shim}}
[Error - 2:19:45 PM] Request textDocument/codeLens failed.
  Message: request handler panicked: tried adding a runnable pointing to a different file: TestMod { path: "" } for FileId(2014)
  Code: -32603 
Panic context:
```

roife:
fix #18704.

### DIFF
--- a/crates/ide/src/runnables.rs
+++ b/crates/ide/src/runnables.rs
@@ -16,7 +16,7 @@ use ide_db::{
 };
 use itertools::Itertools;
 use span::{Edition, TextSize};
-use stdx::{always, format_to};
+use stdx::format_to;
 use syntax::{
     ast::{self, AstNode},
     SmolStr, SyntaxNode, ToSmolStr,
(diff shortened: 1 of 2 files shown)


crates/ide/src/runnables.rs
```diff
@@ -183,8 +183,9 @@ pub(crate) fn runnables(db: &RootDatabase, file_id: FileId) -> Vec<Runnable> {
     });
 
     sema.file_to_module_defs(file_id)
-        .map(|it| runnable_mod_outline_definition(&sema, it))
-        .for_each(|it| add_opt(it, None));
+        .filter_map(|it| runnable_mod_outline_definition(&sema, it))
+        .filter(|it| it.nav.file_id == file_id)
```
Veykril, line 187: I think we need to just discard the assert actually? The problem here is the upmapping of `include` yielding back a real file id isntead of the usual macro call, and so I think we can still hit this with some silly stuff like
```rs
impl Type {
    include!("path/of/a-runnable-assoc-thing)
}
```

The assert used to be right when it was written but since we've changed how token upmapping works for includes so thats no longer the case I think (we had a similar issue with diagnostics)

roife: Agree. I also find the assertion here a bit strange.

Veykril: The reason for it was mainly if we ended up generating a runnable for a differing file that indicates a bug, since we are supposed to generate them for a specific file, but with `include` this can happen either way so plain filtering is the way to go now

Veykril (approved):

PR #18729, opened 2024-12-20
Clear flycheck diagnostics more granularly

Veykril:
Closes https://github.com/rust-lang/rust-analyzer/pull/18332


Strackeror:
Alright, yeah, that should fix my problem, if I'm understanding correctly ! Interesting to see the different approach :)

PR #18731, opened 2024-12-20
Revert "internal: Drop proc-macro server support for ~1.66.0 and older toolchains"

Veykril:
Reverts rust-lang/rust-analyzer#18670

Reverting for now as this did not go as far back as I initially anticipated, see https://github.com/rust-lang/rust-analyzer/issues/18705#issuecomment-2557404602


Veykril:
Once https://github.com/rust-lang/rust-analyzer/issues/18730 is implemented I'll propose setting that minimum version to 1.77 and then removing this in a couple months after as a headsup.

PR #18740, opened 2024-12-22
fix: Fix empty check diagnostics not marking files as changed

Veykril:
Closes https://github.com/rust-lang/rust-analyzer/issues/17300

### DIFF
--- a/crates/hir-ty/src/chalk_db.rs
+++ b/crates/hir-ty/src/chalk_db.rs
@@ -22,7 +22,6 @@ use hir_def::{
 
 use crate::{
     db::{HirDatabase, InternedCoroutine},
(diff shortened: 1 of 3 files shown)


crates/hir-ty/src/chalk_db.rs
```diff
@@ -850,13 +848,6 @@ fn impl_def_datum(
     };
     let where_clauses = convert_where_clauses(db, impl_id.into(), &bound_vars);
     let negative = impl_data.is_negative;
-    debug!(
```
Veykril, line 853: This keeps popping up in logs when you set debug/trace for any specific module. I am not sure why (likely a bug in our logging infra somehow), but it is rather annoying

PR #18746, opened 2024-12-23
minor: Sync from downstream

lnicola:


crates/hir-ty/src/mir/eval.rs
```diff
@@ -1638,6 +1638,7 @@ impl Evaluator<'_> {
             return Ok(0);
         };
         match &layout.variants {
+            Variants::Empty => unreachable!(),
```
lnicola, line 1641: New panic here, we should keep an eye out for it.

PR #18757, opened 2024-12-25
feat: support updating snapshot tests with codelens/hovering/runnables

### LINKED ISSUE #17812: [Feature Request] Add `Update Expect` CodeLens for `expect-test`
First, let's create a new project and add the `expect-test` dependency to demonstrate:

```bash
cargo new demo
cd demo
cargo add expect-test
```

add the following code to `src/main.rs`:
```rust
fn main() {
    println!("Hello, world!");
}

#[test]
fn test() {
    expect_test::expect![].assert_eq("hi");
}
```

For a test block, CodeLens will be shown:

```rust
#[test]
▶ Run Test | Debug
fn test() {
    expect_test::expect![].assert_eq("hi");
}
```

But to update **an** expect_test block(not update all by `UPDATE_EXPECT=1 cargo test`), we currently need to:

- Move the cursor into the test block.
- Use `Ctrl+Shift+P`.
- Select rust-analyzer: Run.
- Choose test test + expect.

## Proposed Feature

It would be more convenient if we could directly update the expect_test block with a single click by adding an `Update Expect` button.

```rust
#[test]
▶ Run Test | Debug | Update Expect
fn test() {
    expect_test::expect![].assert_eq("hi");
}
```

After clicking Update Expect, the content will be updated to: 

```rust
#[test]
▶ Run Test | Debug | Update Expect
fn test() {
    expect_test::expect!["hi"].assert_eq("hi");
}
```

This feature would greatly enhance the user experience when working with `expect-test` in Rust projects. Thank you for considering this request.

roife:
Fix #17812

This PR introduces functionality to update snapshot tests.

It detect dependencies of snapshot testing crates first (similar to how `FamousDefs`). If any dependencies are found, it uses `FindUsages` to determine whether the fn/impl/mod interacts with snapshot testing APIs.

Currently, the supported crates include `expect-test`, `insta` and `snapbox`.

This PR adds the following configurations:
- `lens.UpdateTest.Enable` (default: true).
- `hover.Actions.UpdateTest.Enable` (default: true).
- `runnables.AskBeforeUpdateTest` (default: true, client-side).

### DIFF
--- a/crates/ide/src/runnables.rs
+++ b/crates/ide/src/runnables.rs
@@ -657,62 +657,58 @@ impl<'a, 'b> TestDefs<'a, 'b> {
     }
 
     fn expect_test(&self) -> bool {
-        self.find_macro("expect_test:expect") || self.find_macro("expect_test::expect_file")
+        self.find_macro("expect_test", &["expect", "expect_file"])
     }
 
     fn insta(&self) -> bool {
-        self.find_macro("insta:assert_snapshot")
-            || self.find_macro("insta:assert_debug_snapshot")
-            || self.find_macro("insta:assert_display_snapshot")
-            || self.find_macro("insta:assert_json_snapshot")
-            || self.find_macro("insta:assert_yaml_snapshot")
-            || self.find_macro("insta:assert_ron_snapshot")
-            || self.find_macro("insta:assert_toml_snapshot")
-            || self.find_macro("insta:assert_csv_snapshot")
-            || self.find_macro("insta:assert_compact_json_snapshot")
-            || self.find_macro("insta:assert_compact_debug_snapshot")
-            || self.find_macro("insta:assert_binary_snapshot")
+        self.find_macro(
+            "insta",
+            &[
+                "assert_snapshot",
+                "assert_debug_snapshot",
+                "assert_display_snapshot",
+                "assert_json_snapshot",
+                "assert_yaml_snapshot",
+                "assert_ron_snapshot",
+                "assert_toml_snapshot",
+                "assert_csv_snapshot",
+                "assert_compact_json_snapshot",
+                "assert_compact_debug_snapshot",
+                "assert_binary_snapshot",
+            ],
+        )
     }
 
     fn snapbox(&self) -> bool {
-        self.find_macro("snapbox:assert_data_eq")
-            || self.find_macro("snapbox:file")
-            || self.find_macro("snapbox:str")
+        self.find_macro("snapbox", &["assert_data_eq", "file", "str"])
     }
 
-    fn find_macro(&self, path: &str) -> bool {
-        let Some(hir::ScopeDef::ModuleDef(hir::ModuleDef::Macro(it))) = self.find_def(path) else {
-            return false;
-        };
-
-        Definition::Macro(it)
-            .usages(self.0)
-            .in_scope(&SearchScope::file_range(self.2))
-            .at_least_one()
-    }
-
-    fn find_def(&self, path: &str) -> Option<hir::ScopeDef> {
+    fn find_macro(&self, crate_name: &str, paths: &[&str]) -> bool {
         let db = self.0.db;
 
-        let mut path = path.split(':');
-        let item = path.next_back()?;
-        let krate = path.next()?;
-        let dep = self.1.dependencies(db).into_iter().find(|dep| dep.name.eq_ident(krate))?;
+        let Some(dep) =
+            self.1.dependencies(db).into_iter().find(|dep| dep.name.eq_ident(crate_name))
+        else {
+            return false;
+        };
+        let module = dep.krate.root_module();
+        let scope = module.scope(db, None);
 
-        let mut module = dep.krate.root_module();
-        for segment in path {
-            module = module.children(db).find_map(|child| {
-                let name = child.name(db)?;
-                if name.eq_ident(segment) {
-                    Some(child)
-                } else {
-                    None
+        paths
+            .iter()
+            .filter_map(|path| {
+                let (_, def) = scope.iter().find(|(name, _)| name.eq_ident(path))?;
+                match def {
+                    hir::ScopeDef::ModuleDef(hir::ModuleDef::Macro(it)) => Some(it),
+                    _ => None,
                 }
-            })?;
-        }
-
-        let (_, def) = module.scope(db, None).into_iter().find(|(name, _)| name.eq_ident(item))?;
-        Some(def)
+            })
+            .any(|makro| {
+                Definition::Macro(*makro)
+                    .usages(self.0)
+                    .in_scope(&SearchScope::file_range(self.2))
+                    .at_least_one()
+            })
     }
 }
 
--- a/crates/ide/src/runnables.rs
+++ b/crates/ide/src/runnables.rs
@@ -1,4 +1,4 @@
-use std::{fmt, ops::Not};
+use std::fmt;
 
 use ast::HasName;
 use cfg::{CfgAtom, CfgExpr};
@@ -20,7 +20,7 @@ use span::{Edition, TextSize};
 use stdx::format_to;
 use syntax::{
     ast::{self, AstNode},
-    SmolStr, SyntaxNode, ToSmolStr,
+    format_smolstr, SmolStr, SyntaxNode, ToSmolStr,
 };
 
 use crate::{references, FileId, NavigationTarget, ToNav, TryToNav};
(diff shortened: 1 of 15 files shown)


roife:
Emmm, I can pass the Prettier check locally, but it fails on CI. 🤔

roife:
It seems to be a bug in Prettier. 🫤
prettier --write . && git diff --exit-code passes, but prettier --check . fails. 🫨

roife:
Here's a demo.
Screen Recording 2024-12-25 at 16 01 40

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -933,39 +933,31 @@ pub(crate) fn handle_runnables(
...
 
-    let expect_test = match offset {
-        Some(offset) => {
-            let source_file = snap.analysis.parse(file_id)?;
-            algo::find_node_at_offset::<ast::MacroCall>(source_file.syntax(), offset)
-                .and_then(|it| it.path()?.segment()?.name_ref())
-                .map_or(false, |it| it.text() == "expect" || it.text() == "expect_file")
-        }
-        None => false,
-    };
```
Veykril, line 944: 🤨 oh I never realized we were doing the check here of all places, that seems very out of place to me

crates/rust-analyzer/src/lsp/to_proto.rs
```diff
@@ -1786,12 +1800,24 @@ pub(crate) mod command {
...
     }
 
+    pub(crate) fn update_single(
+        runnable: &lsp_ext::Runnable,
+        label: &Option<SmolStr>,
+    ) -> Option<lsp_types::Command> {
+        let runnable = make_update_runnable(runnable, label)?;
+        Some(lsp_types::Command {
+            title: format!("↺\u{fe0e} Update Tests ({})", label.as_ref()?),
+            command: "rust-analyzer.updateSingle".into(),
```
Veykril, line 1816: Why do we need a new command for this? That requires this change to be implemented by each editor where as before we were merely using the env variables which would work out of the box for any client.

Veykril: If this is solely for the prompt that asks the user whether you want to update, we should make that VSCode specific in the sense that the vscode extension should check for the env vars in the runnable payload and act accordingly instead that way we do not need to expose a client implemented command here

crates/ide/src/hover/render.rs
```diff
@@ -378,7 +378,18 @@ pub(super) fn process_markup(
 
 fn definition_owner_name(db: &RootDatabase, def: &Definition, edition: Edition) -> Option<String> {
     match def {
-        Definition::Field(f) => Some(f.parent_def(db).name(db)),
+        Definition::Field(f) => {
```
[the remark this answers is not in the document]
roife: Ah, I didn’t realize that. 🙊 It was accidentally merged in.

crates/rust-analyzer/src/lsp/to_proto.rs
```diff
@@ -1838,6 +1864,29 @@ pub(crate) mod command {
...
+
+    let mut runnable = runnable.clone();
+    runnable.label = format!("{} + {}", runnable.label, label);
+
+    let lsp_ext::RunnableArgs::Cargo(r) = &mut runnable.args else {
+        unreachable!();
+    };
+
+    let environment_vars =
+        [("UPDATE_EXPECT", "1"), ("INSTA_UPDATE", "always"), ("SNAPSHOTS", "overwrite")];
```
Veykril, line 1884: I f eel like these should be provided by the `UpdateTest` instance (that way we have all that crate specific stuff in one place).

crates/ide/src/runnables.rs
```diff
@@ -575,6 +616,102 @@ fn has_test_function_or_multiple_test_submodules(
...
+                "assert_binary_snapshot",
+            ],
+        )
+    }
+
+    fn snapbox(&self) -> bool {
+        self.find_macro("snapbox", &["assert_data_eq", "file", "str"])
+    }
+
+    fn find_macro(&self, crate_name: &str, paths: &[&str]) -> bool {
```
Veykril, line 686: This seems a bit weird to me the way this works, a better idea would be to just check the macro calls of the corresponding body I think and match them up with the expected macros. That is fetch the `BodySourceMap` for the corresponding body (we might have that via semantics already here, not sure), and iterate the `expansions` in there to find the macros. Before doing so we should probably first try to resolve all the known testing macros and cache them to just match up the IDs against another

roife:
I’ve been struggling to pass the `prettier --check .`, even after running `npm run format` locally.

I just realized that I’m using Node 18.20.5 (installed with nvm by default), while the CI is on Node 18.4.0 (See https://github.com/actions/setup-node). The `prettier` installed via `npm install` in these two environments use different linting rules. 😧 Therefore, it could lead to a scenario where the code runs successfully locally but fails in the CI environment...

crates/ide/src/runnables.rs
roife: I use FindUsage because it allows us to quickly filter out irrelevant macro calls by name. Since we need to search within (potentially large) mod/impl blocks, this approach is more efficient.

Veykril: > Resolving the testing macros and caching their IDs is reasonable, but I haven’t figured out where to store them yet

The caching here is good enough to be done at the start of the runnables computation (that is we don't cache across requests)

> I use FindUsage because it allows us to quickly filter out irrelevant macro calls by name.

I see, I was thinking of what if the macro call is hidden behind another macro call, but I think that basicalyl breaks all of those frameworks anyways so little point in trrying to do that. I see your approach then. The way this is done here still feels a bit brittle to me though. Will add another comment regarding that in a bit

Veykril:
I think it should be fine for us to bump the default node version on our CI then (separate PR)

crates/ide/src/runnables.rs
```diff
@@ -575,6 +616,120 @@ fn has_test_function_or_multiple_test_submodules(
...
+
+        let res: SmolStr = builder.join(" + ").into();
+        if res.is_empty() {
+            None
+        } else {
+            Some(format_smolstr!("↺\u{fe0e} Update Tests ({res})"))
+        }
+    }
+
+    pub fn env(&self) -> SmallVec<[(&str, &str); 3]> {
```
Veykril, line 649: Assuming we have it in the deps of this crate here already (we have it in the tree for sure), these could also just be `ArrayVec`s

crates/ide/src/runnables.rs
```diff
@@ -575,6 +616,120 @@ fn has_test_function_or_multiple_test_submodules(
     number_of_test_submodules > 1
 }
 
+struct TestDefs<'a, 'b>(&'a Semantics<'b, RootDatabase>, hir::Crate, hir::FileRange);
```
Veykril, line 619: Let's make these named fields

crates/ide/src/runnables.rs
```diff
@@ -575,6 +616,102 @@ fn has_test_function_or_multiple_test_submodules(
...
+    fn find_macro(&self, crate_name: &str, paths: &[&str]) -> bool {
+        let db = self.0.db;
+
+        let Some(dep) =
+            self.1.dependencies(db).into_iter().find(|dep| dep.name.eq_ident(crate_name))
+        else {
+            return false;
+        };
+        let module = dep.krate.root_module();
+        let scope = module.scope(db, None);
```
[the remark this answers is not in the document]
Veykril: Yes, would be good to cache this at the start of the runnables computation

crates/ide/src/runnables.rs
```diff
@@ -575,6 +616,120 @@ fn has_test_function_or_multiple_test_submodules(
...
+                    _ => None,
+                }
+            })
+            .any(|makro| {
+                Definition::Macro(*makro)
+                    .usages(self.0)
+                    .in_scope(&SearchScope::file_range(self.2))
+                    .at_least_one()
+            })
+    }
```
Veykril, line 730: instead of `scope.iter()` checking the parts we should just use `Semantics::resolve_mod_path` and construct `ModPath`s from the string paths, that way we re-use our normal resolution logic. And again this can be cached like the other part at the start of the runnables stuff.

roife: Currently we are unable to create `ModPath` directly from string paths because we are unable to construct `Name` from a String. An alternative option is to use `new_symbol_root`, but I am unsure if it is the correct approach.

Veykril: That's fine, you can use `new_symbol_root` here. We don't care about hygiene for this.

Veykril (approved):

PR #18758, opened 2024-12-26
Improve SCIP symbols

mgsloan:
In particular, the symbol generation before this change creates a lot of symbols with the same name for different definitions. This change makes progress on symbol uniqueness, but does not fix a couple cases where it was unclear to me how to fix (see TODOs in `scip.rs`)

Before this change `SymbolInformation` provided by a document was the
info for all encountered symbols that have not yet been emitted. So,
the symbol information on a `Document` was a mishmash of symbols
defined in the documents, symbols from other documents, and external
symbols.

After this change, the `SymbolInformation` on documents is just the
locals and defined symbols from the document.  All symbols referenced
and not from emitted documents are included in `external_symbols`.

Other behavior changes:

* `scip` command now reports symbol information omitted due to symbol collisions. Iterating with this on a large codebase (Zed!) resulted in the other improvements in this change.

* Generally fixes providing the path to nested definitions in symbols. Instead of having special cases for a couple limited cases of nesting, implements `Definition::enclosing_definition` and uses this to walk definitions.

* Parameter variables are now treated like locals.

    - This fixes a bug where closure captures also received symbols scoped to the containing function.  To bring back parameter symbols I would want a way to filter these out, since they can cause symbol collisions.

    - Having symbols for them seems to be intentional in 27e2eea54fbd1edeefa2b47ddd4f552a04b86582, but no particular use is specified there. For the typical indexing purposes of SCIP I don't see why parameter symbols are useful or sensible, as function parameters are not referencable by anything but position. I can imagine they might be useful in representing diagnostics or something.

* Inherent impls are now represented as `impl#[SelfType]` - a type named `impl` which takes a single type parameter.

* Trait impls are now represented as `impl#[SelfType][TraitType]` - a type named `impl` which takes two type parameters.

* Associated types in traits and impls are now treated like types instead of type parameters, and so are now suffixed with `#` instead of wrapped with `[]`.  Treating them as type parameters seems to have been intentional in 73d9c77f2aeed394cf131dce55807be2d3f54064 but it doesn't make sense to me, so changing it.

* Static variables are now treated as terms instead of `Meta`, and so receive `.` suffix instead of `:`.

* Attributes are now treated as `Meta` instead of `Macro`, and so receive `:` suffix instead of `!`.

* `enclosing_symbol` is now provided for labels and generic params, which are local symbols.

* Fixes a bug where presence of `'` causes a descriptor name to get double wrapped in backticks, since both `fn new_descriptor` and `scip::symbol::format_symbol` have logic for wrapping in backticks. Solution is to simply delete the redundant logic.

### DIFF
--- a/crates/hir-ty/src/display.rs
+++ b/crates/hir-ty/src/display.rs
@@ -474,7 +474,7 @@ impl HirDisplay for ProjectionTy {
 
         let trait_ref = self.trait_ref(f.db);
         write!(f, "<")?;
-        fmt_trait_ref(f, &trait_ref, true)?;
+        fmt_trait_ref(f, &trait_ref, TraitRefFormat::SelfAsTrait)?;
         write!(
             f,
             ">::{}",
@@ -1775,21 +1775,34 @@ fn write_bounds_like_dyn_trait(
     Ok(())
 }
 
+#[derive(Clone, Copy)]
+pub enum TraitRefFormat {
+    SelfAsTrait,
+    SelfImplementsTrait,
+    OnlyTrait,
+}
+
 fn fmt_trait_ref(
     f: &mut HirFormatter<'_>,
     tr: &TraitRef,
-    use_as: bool,
+    format: TraitRefFormat,
 ) -> Result<(), HirDisplayError> {
     if f.should_truncate() {
         return write!(f, "{TYPE_HINT_TRUNCATION}");
     }
 
-    tr.self_type_parameter(Interner).hir_fmt(f)?;
-    if use_as {
-        write!(f, " as ")?;
-    } else {
-        write!(f, ": ")?;
+    match format {
+        TraitRefFormat::SelfAsTrait => {
+            tr.self_type_parameter(Interner).hir_fmt(f)?;
+            write!(f, " as ")?;
+        }
+        TraitRefFormat::SelfImplementsTrait => {
+            tr.self_type_parameter(Interner).hir_fmt(f)?;
+            write!(f, ": ")?;
+        }
+        TraitRefFormat::OnlyTrait => {}
     }
+
     let trait_ = tr.hir_trait_id();
     f.start_location_link(trait_.into());
     write!(f, "{}", f.db.trait_data(trait_).name.display(f.db.upcast(), f.edition()))?;
@@ -1798,9 +1811,14 @@ fn fmt_trait_ref(
     hir_fmt_generics(f, &substs[1..], None, substs[0].ty(Interner))
 }
 
-impl HirDisplay for TraitRef {
+pub struct TraitRefDisplayWrapper {
+    pub trait_ref: TraitRef,
+    pub format: TraitRefFormat,
+}
+
+impl HirDisplay for TraitRefDisplayWrapper {
     fn hir_fmt(&self, f: &mut HirFormatter<'_>) -> Result<(), HirDisplayError> {
-        fmt_trait_ref(f, self, false)
+        fmt_trait_ref(f, &self.trait_ref, self.format)
     }
 }
 
@@ -1811,10 +1829,12 @@ impl HirDisplay for WhereClause {
         }
 
         match self {
-            WhereClause::Implemented(trait_ref) => trait_ref.hir_fmt(f)?,
+            WhereClause::Implemented(trait_ref) => {
+                fmt_trait_ref(f, trait_ref, TraitRefFormat::SelfImplementsTrait)?;
+            }
             WhereClause::AliasEq(AliasEq { alias: AliasTy::Projection(projection_ty), ty }) => {
                 write!(f, "<")?;
-                fmt_trait_ref(f, &projection_ty.trait_ref(f.db), true)?;
+                fmt_trait_ref(f, &projection_ty.trait_ref(f.db), TraitRefFormat::SelfAsTrait)?;
                 write!(f, ">::",)?;
                 let type_alias = from_assoc_type_id(projection_ty.associated_ty_id);
                 f.start_location_link(type_alias.into());
--- a/crates/hir-ty/src/display.rs
+++ b/crates/hir-ty/src/display.rs
@@ -474,7 +474,9 @@ impl HirDisplay for ProjectionTy {
 
         let trait_ref = self.trait_ref(f.db);
         write!(f, "<")?;
-        fmt_trait_ref(f, &trait_ref, TraitRefFormat::SelfAsTrait)?;
+        trait_ref.self_type_parameter(Interner).hir_fmt(f)?;
+        write!(f, " as ")?;
+        trait_ref.hir_fmt(f)?;
         write!(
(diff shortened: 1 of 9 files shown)


crates/rust-analyzer/src/cli/scip.rs
```diff
@@ -224,6 +265,53 @@ impl flags::Scip {
...
+included in the output, but this causes information lookup to be ambiguous and so information about
+these symbols presented by downstream tools may be incorrect.
+
+Known cases that can cause this:
+
+  * Definitions in crate example binaries which have the same symbol as definitions in the library
+  or some other example.
+
+  * When a struct/enum/const/static/impl is defined with a function, it erroneously appears to be
+  defined at the same level as the function.
```
davidbarsky, line 280: is this a rust-analyzer bug or a limitation of the SCIP format?

davidbarsky (commented):
i'll probably have a few additional comments later, but I didn't want to block anything beyond what is strictly necessary.

crates/ide/src/moniker.rs
```diff
@@ -232,157 +247,129 @@ pub(crate) fn def_to_kind(db: &RootDatabase, def: Definition) -> SymbolInformati
...
+/// Computes a `MonikerResult` for a definition. Result cases:
+///
+/// `Some(MonikerResult::Moniker(_))` provides a unique `Moniker` which refers to a definition.
+///
+/// `Some(MonikerResult::Local { .. })` provides a `Moniker` for the definition enclosing a local.
+///
+/// `None` is returned in the following cases:
+///
+/// * Inherent impl definitions, as they cannot be uniquely identified (multiple are allowed for the
+///   same type).
```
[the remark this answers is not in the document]
Veykril: Can't we just add a disambiguating counter if there are multiple impls? (that's how rustdoc does it, which we fwiw, don't yet mimic for the open docs stuff either)

crates/rust-analyzer/src/cli/scip.rs
mgsloan: I would consider both of these to be rust-analyzer bugs.  The challenge for this one is that the `hir` representation only appears to know what module they are in, and not whether they are contained in a function.  I am not familiar enough with rust-analyzer internals to provide such information.

crates/ide/src/moniker.rs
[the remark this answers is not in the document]
mgsloan: Numbering makes sense, but it seems tricky as these definitions are populated when visiting tokens in other files that refer to them.  Tracking this in https://github.com/rust-lang/rust-analyzer/issues/18772

For now I like the idea of just using the first impl.  That's easy to do, I've updated the change with it.

crates/hir-ty/src/display.rs
```diff
@@ -1775,21 +1775,34 @@ fn write_bounds_like_dyn_trait(
     Ok(())
 }
 
+#[derive(Clone, Copy)]
+pub enum TraitRefFormat {
+    SelfAsTrait,
+    SelfImplementsTrait,
+    OnlyTrait,
+}
```
Veykril, line 1783: I think it would make more sense to implement HirDisplay directly for `hir::TraitRef` and have that by default omit the self type. Then we don't need to expose this enum beyond the `hir` crate nor do we need the formatting wrapper struct

crates/ide-db/src/defs.rs
```diff
@@ -932,3 +954,29 @@ impl TryFrom<DefWithBody> for Definition {
...
+    type Error = ();
+    fn try_from(container: ItemContainer) -> Result<Self, Self::Error> {
+        match container {
+            ItemContainer::Trait(it) => Ok(it.into()),
+            ItemContainer::Impl(it) => Ok(it.into()),
+            ItemContainer::Module(it) => Ok(it.into()),
+            ItemContainer::ExternBlock() | ItemContainer::Crate(_) => Err(()),
+        }
+    }
+}
```
Veykril, line 968: Can we put this as a local closure inside `Definition::enclosing_definition`? The impl itself does not make too much sense to me (exposition wise)

crates/ide/src/moniker.rs
```diff
@@ -232,157 +247,117 @@ pub(crate) fn def_to_kind(db: &RootDatabase, def: Definition) -> SymbolInformati
...
-            desc: def_to_kind(db, it.parent_def(db).into()).into(),
-        });
+fn def_to_non_local_moniker(
+    db: &RootDatabase,
+    definition: Definition,
+    from_crate: Crate,
+) -> Option<Moniker> {
+    match definition {
+        // Not possible to give sensible unique symbols for inherent impls, as multiple can be
+        // defined for the same type.
```
Veykril, line 297: Let's adjust the comment as it is possible in theory (by numbering) though right now it may be non-trivial to do so

crates/ide/src/moniker.rs
```diff
@@ -232,157 +247,117 @@ pub(crate) fn def_to_kind(db: &RootDatabase, def: Definition) -> SymbolInformati
...
+                        desc: def_to_kind(db, def).into(),
+                    });
+                } else if reverse_description.is_empty() {
+                    // Don't allow the last descriptor to be absent.
+                    return None;
+                } else {
+                    match def {
+                        Definition::Module(module) if module.is_crate_root() => {}
+                        _ => {
+                            tracing::error!(?def, "Encountered enclosing definition with no name");
```
Veykril, line 346: I think this can be hit with consts:
```rs
const _: () = { let x; };
```
Here `x`'s enclosing definition is a const with no name

crates/rust-analyzer/src/cli/scip.rs
```diff
@@ -619,6 +802,70 @@ pub mod example_mod {
         );
     }
 
+    // TODO: This test represents current misbehavior.
```
Veykril, line 805: nit we prefer FIXME over TODO

crates/hir-ty/src/display.rs
mgsloan: That makes sense!  Certainly much cleaner, I've made that change.  Thanks for identifying it

crates/ide/src/moniker.rs
mgsloan: Ah good catch!  I should have removed this condition in 34dc94bb2d05def8b88739a9382b127feb1a35a0 as now scip.rs is handling yielding SymbolInformation for only the first impl.

crates/rust-analyzer/src/cli/scip.rs
mgsloan: Done

crates/ide-db/src/defs.rs
mgsloan: Done

mgsloan:
@Veykril Thanks for the detailed review!  I've updated this with the changes you identified

Veykril (approved):

PR #18760, opened 2024-12-26
internal: Workaround salsa cycles leaking

Veykril:
cc https://github.com/rust-lang/rust-analyzer/issues/18745

### DIFF
--- a/crates/rust-analyzer/src/handlers/dispatch.rs
+++ b/crates/rust-analyzer/src/handlers/dispatch.rs
@@ -5,6 +5,7 @@ use std::{
 };
 
 use ide::Cancelled;
+use ide_db::base_db::ra_salsa::Cycle;
 use lsp_server::{ExtractError, Response, ResponseError};
 use serde::{de::DeserializeOwned, Serialize};
 use stdx::thread::ThreadIntent;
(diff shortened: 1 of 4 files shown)


crates/rust-analyzer/src/handlers/dispatch.rs
```diff
@@ -328,7 +329,13 @@ where
             if let Some(panic_message) = panic_message {
                 message.push_str(": ");
                 message.push_str(panic_message)
-            };
+            } else if let Some(cycle) = panic.downcast_ref::<Cycle>() {
+                tracing::error!("Cycle propagated out of salsa! This is a bug: {cycle:?}");
+                return Err(Cancelled::PropagatedPanic);
```
davidbarsky, line 334: I'm rebasing my fork against this PR now, and I'm debating whether this the removal of `#[non_exhaustive]` in Salsa is necessary—could we return a different error here or is `salsa::Cancelled` load-bearing?

Veykril: At this point the value of cancelled doesn't really matter anymore I think so we could also return some other unit error type here (the caller merely wants to know whether cancellation occured)

davidbarsky: Done: https://github.com/rust-lang/rust-analyzer/pull/18762

PR #18788, opened 2024-12-29
Remove `rust-analyzer.cargo.sysrootQueryMetadata` config again

Veykril:
Closes https://github.com/rust-lang/rust-analyzer/issues/18590


Veykril:
Ideally the rust-project.json will actually list the sysroot workspace (that is, it provides a stitched one), so that will need some adjustments to the json format.

Veykril:
A thought for a follow up to this PR, we should be able to cache the fetched sysroot and re-use it between multiple workspaces.

darichey (commented):
Thanks, this looks good for our use case :)

PR #18791, opened 2024-12-30
fix: avoid generating colliding names in extract_variable

### LINKED ISSUE #18786: Extract variable assist creates colliding names
This can cause the snippet-rename it invokes to immediately rename differing usages, we should make sure the prefilled name is not colliding with other locals. cc @roife I think you've been doing a bunch of the name collision stuff in assists recently :^)

roife:
fix #18786.

This PR:

- Move the `for_variable` into the `NameGenerator` and change return types from `String` to `SmolStr`
- Introduces a new method `NameGenerator::new_from_scope_locals`, which is initialized with names of local variables from the current semantics scope, and refactor the following assists:
   - `destructure_tuple_binding`
   - `extract_variable`
   - `replace_is_method_with_if_let_method`

### DIFF
--- a/crates/ide-assists/src/handlers/destructure_tuple_binding.rs
+++ b/crates/ide-assists/src/handlers/destructure_tuple_binding.rs
@@ -6,7 +6,6 @@ use ide_db::{
     text_edit::TextRange,
 };
 use itertools::Itertools;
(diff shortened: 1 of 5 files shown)


Veykril (approved):
Thanks!

PR #18795, opened 2024-12-30
Fix render of literal to be rendered in codeblock

### LINKED ISSUE #18765: Quote or escape the raw string correctly in popup
Currently, the popup for raw string syntax uses the following code:

https://github.com/rust-lang/rust-analyzer/blob/1ce4de2585cb9802c4d5a99d1a904b11fdba9501/crates/ide/src/hover/render.rs#L709-L713

So `r#"A "quoted" string"#` will be rendered like this:

> value of literal: A "quoted" string

Image

However, when the code contains special characters in Markdown like `*` (which is common in regex), the popup is messed up.

For example, `r".*regex.*"` became:

> value of literal: .*regex.*

Image

With a proper quotation or escaping, this could be rendered as:

> value of literal: `.*regex.*`

> value of literal: .\*regex.\*  (Less preferable, since leading and trailing spaces are hard to distinguish.)

dfireBird:
Fixes #18765.

Chose this approach to simplify finding the needed n back ticks, instead of finding the consecutive back ticks. 
Tested for about 100 consecutive back ticks, it was working fine. 

Regarding tests, I'll add some soon. Just wanted to review the implementation.

### DIFF
--- a/crates/ide/src/hover/render.rs
+++ b/crates/ide/src/hover/render.rs
@@ -715,10 +715,17 @@ pub(super) fn literal(
     let mut s = format!("```rust\n{ty}\n```\n___\n\n");
     match value {
         Ok(value) => {
+            let backtick_len = value.chars().filter(|c| *c == '`').count();
+            let backticks = "`".repeat(backtick_len + 1);
+
             if let Some(newline) = value.find('\n') {
-                format_to!(s, "value of literal (truncated up to newline): {}", &value[..newline])
+                format_to!(
+                    s,
+                    "value of literal (truncated up to newline): {backticks} {} {backticks}",
+                    &value[..newline]
+                )
             } else {
-                format_to!(s, "value of literal: {value}")
+                format_to!(s, "value of literal: {backticks} {value} {backticks}")
             }
         }
         Err(error) => format_to!(s, "invalid literal: {error}"),
--- a/crates/ide/src/hover/tests.rs
+++ b/crates/ide/src/hover/tests.rs
@@ -8157,7 +8157,7 @@ fn main() {
             ```
             ___
 
-            value of literal: 🦀🦀\A
+            value of literal: ` 🦀🦀\A `
         "#]],
     );
     check(
@@ -8173,7 +8173,7 @@ fn main() {
             ```
             ___
 
-            value of literal: 🦀\u{1f980}\\\x41
+            value of literal: ` 🦀\u{1f980}\\\x41 `
         "#]],
     );
     check(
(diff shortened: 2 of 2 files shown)


crates/ide/src/hover/render.rs
```diff
@@ -715,10 +715,17 @@ pub(super) fn literal(
...
             if let Some(newline) = value.find('\n') {
-                format_to!(s, "value of literal (truncated up to newline): {}", &value[..newline])
+                format_to!(
+                    s,
+                    "value of literal (truncated up to newline): {backticks} {} {backticks}",
+                    &value[..newline]
+                )
             } else {
-                format_to!(s, "value of literal: {value}")
+                format_to!(s, "value of literal: {backticks} {value} {backticks}")
```
Veykril, line 728:
```suggestion
                format_to!(s, "value of literal: {backticks}{value}{backticks}")
```

Veykril: Ah wait I just realized this is allowed by the spec (and required to quote a backtick) nvm

Veykril (approved):

crates/ide/src/hover/render.rs
dfireBird: I forgot to mention in the PR initial comment, the space is needed, otherwise single back tick or back tick at last position will mess the formatting

dfireBird:
Added tests as part of same commit, they basically test the number of back ticks that is before and after. Nothing much can be done apart from that I guess.

Jisu-Woniu:
What if the string consists entirely of space characters?

> If the resulting string both begins and ends with a [space](https://spec.commonmark.org/0.31.2/#space) character, but does not consist entirely of [space](https://spec.commonmark.org/0.31.2/#space) characters, a single [space](https://spec.commonmark.org/0.31.2/#space) character is removed from the front and back.

e.g.

```markdown
`  `   (the rendered form of an empty string)
`   `  (the rendered form of string " ", or 1 space)
```

will become 

`  ` (2 spaces) and `   ` (3 spaces)

Veykril:
Ah indeed, fully whitespace strings we need to special case thats a good point!

dfireBird:
We should do that, but fun thing in vscode hover renderer trims all leading and trailing spaces inside codeblock. Like even with current implementation hovering over something like `r" hello "` will only render `hello`

Veykril:
That is unfortunate but not much we can do about then I think, ultimately markdown rendering differs from implementation to implementation. We could note down the number of leading / trailing whitespace in parenthesis after I suppose (if there are any)

Jisu-Woniu:
Is it possible to use `<code>` `</code>` to wrap the values instead?

dfireBird:
From my research about it, VSC only allows `<span></span` by default and others are sanitized. There is user config that can control it, but I'm not sure if that can be used here.

PR #18806, opened 2024-12-31
fix: Deduplicate crate graph

LHolten:
See the discussion in https://github.com/rust-lang/rust-analyzer/issues/18804


LHolten:
Maybe the deduplication test should be brought back too?

Veykril:
Yes they should

LHolten:
Brought back the tests

crates/project-model/src/tests.rs
```diff
@@ -46,6 +46,39 @@ fn load_cargo_with_overrides(
     to_crate_graph(project_workspace)
 }
 
+fn load_cargo_with_fake_sysroot(
```
Veykril, line 49: do we still need this function or can we just re-use `load_rust_project`

LHolten: I cleaned up this test file, the tests didn't need the fake sysroot, but I had to split up and use `load_cargo_with_overrides` to make it work.

LHolten: oops, it seems i broke something

LHolten:
@Veykril What do you think of just making the `CrateGraph::extend` method sort the dependencies automatically?
That way it is hard to misuse the `CrateGraph` api and thus the tests in the rust-analyzer crate are not really needed, so we can then remove those.

Veykril:
Yes that should be fine. I just realized these tests will be gone soon enough anyways as we are currently doing a big rewrite that includes splitting the crate graph up into smaller parts.

LHolten:
Done

LHolten:
Done

filipwiech:
@rustbot ready

Drive-by question: back when the crate deduplication was originally introduced, I think it promised potential memory usage improvements in r-a. Is it still the case, or has the design changed since then? Sorry for a layman's inquiry. :slightly_smiling_face:

Veykril:
That is the case (for multi cargo workspace projects)

Veykril (approved):

filipwiech:
> That is the case
Nice, thanks! :+1: 
> (for multi cargo workspace projects)
Ah, yes, now I remember, thank you. :slightly_smiling_face:

PR #18813, opened 2025-01-01
feat: Add a new and improved syntax tree view

Giga-Bowser:
Lots of cool features!

- Reveal button to highlight a given syntax element
- Tree selection matches document selection
- Hover to see an element's text
- Supports trees for rust code in string literals

I'm still pretty new to the VS Code side of things so there's probably a better way to do some of the stuff I did. If someone more familiar with this sort of thing wants to try to record a GIF or something of this in use, feel free. Otherwise I'll try to get one up in a bit,


Giga-Bowser:
Also, would people be fine with this replacing the old textual syntax tree view? I could try to move some of the functionality to integrate with this feature if it's still desired. Maybe an action on each syntax element to copy the old format view to the clipboard?

Giga-Bowser:
Done!

crates/ide/src/view_syntax_tree.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    // SAFETY: This is actually just a modified version of str::replace that
+    // tracks replacements, so I defer to the standard library on safety here
+    let mut trim_result = String::new();
+    let mut marker_positions = Vec::new();
+    let mut skipped = 0;
+    let mut last_end = 0;
+    for (start, part) in text.match_indices("$0") {
+        marker_positions.push((start - skipped) as u32);
+        trim_result.push_str(unsafe { text.get_unchecked(last_end..start) });
```
Veykril, line 114: Imo unsafe code in this feature isn't really necessary, the extra bounds checks are irrelevant wrt to perf as this is not such a performance sensitive thing

editors/code/package.json
```diff
@@ -288,6 +283,29 @@
...
+                "title": "Hide Whitespace",
+                "icon": "$(filter)",
+                "category": "rust-analyzer"
+            },
+            {
+                "command": "rust-analyzer.syntaxTreeShowWhitespace",
+                "title": "Show Whitespace",
+                "icon": "$(filter-filled)",
+                "category": "rust-analyzer"
+            },
```
Veykril, line 308: The category of these should be distinct, otherwise all you see in the command palette is `rust-analyzer: Copy` which seems a bit confusing. So maybe `rust-analyzer (syntax tree)` or something along those lines

editors/code/package.json
```diff
@@ -3362,6 +3370,22 @@
                 },
                 {
                     "command": "rust-analyzer.openWalkthrough"
+                },
+                {
+                    "command": "rust-analyzer.syntaxTreeReveal",
+                    "when": "false"
```
Veykril, line 3376: Ah I guess this perma hides the command (separate category would still be good I think)

Veykril:
Couple things but otherwise this seems good to me!

lnicola:
I guess we can't put a "Copy" icon next to :mag:, can we?

Giga-Bowser:
> I guess we can't put a "Copy" icon next to 🔍, can we?

We can! Should be fixed.

Veykril (approved):

Veykril:
Thanks!

lnicola:
I'm playing with this, is there a difference between these two? :sweat_smile: 

image

Giga-Bowser:
I actually have no clue. VS Code registers both of these commands for us so I'm not sure why there's two, I'll take a look.

Giga-Bowser:
It seems like since we're registering a custom view container, `View: Show Rust Syntax Tree` shows that container, while `Rust Syntax Tree: Focus on Rust Syntax Tree View` focuses on the tree view within said container. You can see the same thing with builtin views, for example, `View: Show Explorer` and `Explorer: Focus on Folders View`.

PR #18822, opened 2025-01-02
allow targetDir to be an absolute path

bryceberger:
Using an absolute path for cargo's build dir is allowed. I would like to have a global build dir for rust-analyzer as well, but _not_ use the same one as cargo (to avoid waiting for locks to release).

This was previously allowed, but was changed by @Veykril in c8fdcea85c273c386d1024815b955af6edae3fa2. Unsure of the reason for the change, using an absolute path seems to work fine.


Veykril:
Hmm, good question I am not so sure myself anymore why I enforced this (I might've misinterpreted the comment that was on the function back then)

Veykril (approved):

PR #18832, opened 2025-01-03
fix: do not offer completions within macro strings

### LINKED ISSUE #17479: Macro string literal completion edits not replacing prior text
**rust-analyzer version**: rust-analyzer 0.0.0 (498b94b730 2024-06-17)

**rustc version**: rustc 1.74.0 (79e9716c9 2023-11-13)

**editor or extension**: helix

```rust
// add your code here
use chrono::serde::ts_milliseconds;
use chrono::{DateTime, Utc};
use serde::Deserialize;

#[derive(Deserialize)]
struct A {
    #[serde(with = "ts_mill")]
    time: DateTime<Utc>,
}

fn main() {
    println!("Hello, world!");
}
```
helix_auto_complete_in_string

Originally thought it was a helix issue, but was directed here after some discussion.
https://github.com/helix-editor/helix/discussions/11020

vishruth-thimmaiah:
Fixes #17479

This PR disables completions within strings in proc macros.

### DIFF
--- a/crates/ide-completion/src/context/analysis.rs
+++ b/crates/ide-completion/src/context/analysis.rs
@@ -417,6 +417,15 @@ fn analyze(
         derive_ctx,
     } = expansion_result;
 
+    if original_token.kind() != self_token.kind()
+        // FIXME: This check can be removed once we use speculative database forking for completions
+        && !(original_token.kind().is_punct() || original_token.kind().is_trivia())
+        && !(SyntaxKind::is_any_identifier(original_token.kind())
+            && SyntaxKind::is_any_identifier(self_token.kind()))
+    {
+        return None;
+    }
+
     // Overwrite the path kind for derives
     if let Some((original_file, file_with_fake_ident, offset, origin_attr)) = derive_ctx {
         if let Some(ast::NameLike::NameRef(name_ref)) =
--- a/crates/ide-completion/src/tests/attribute.rs
+++ b/crates/ide-completion/src/tests/attribute.rs
@@ -713,6 +713,28 @@ struct Foo;
     );
 }
 
+#[test]
+fn issue_17479() {
+    check(
+        r#"
+//- proc_macros: issue_17479
+fn main() {
+    proc_macros::issue_17479!("te$0");
+}
+"#,
+        expect![""],
+    );
+    check(
+        r#"
+//- proc_macros: issue_17479
+fn main() {
+    proc_macros::issue_17479!("$0");
+}
+"#,
+        expect![""],
+    )
+}
+
 mod cfg {
     use super::*;
 
--- a/crates/test-fixture/src/lib.rs
+++ b/crates/test-fixture/src/lib.rs
@@ -376,7 +376,7 @@ impl ChangeFixture {
     }
 }
 
-fn default_test_proc_macros() -> [(String, ProcMacro); 7] {
+fn default_test_proc_macros() -> [(String, ProcMacro); 8] {
     [
         (
             r#"
@@ -480,6 +480,21 @@ pub fn issue_18840(_attr: TokenStream, _item: TokenStream) -> TokenStream {
                 name: Symbol::intern("issue_18840"),
                 kind: ProcMacroKind::Attr,
                 expander: sync::Arc::new(Issue18840ProcMacroExpander),
+                disabled: false,
+            },
+        ),
+        (
+            r#"
+#[proc_macro]
+pub fn issue_17479(input: TokenStream) -> TokenStream {
+    input
+}
+"#
+            .into(),
+            ProcMacro {
+                name: Symbol::intern("issue_17479"),
+                kind: ProcMacroKind::Bang,
+                expander: sync::Arc::new(Issue17479ProcMacroExpander),
                 disabled: false,
             },
         ),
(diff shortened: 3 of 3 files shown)


ChayimFriedman2:
Please add a test. You can look at existing tests in https://github.com/rust-lang/rust-analyzer/tree/master/crates/ide-completion/src/tests.

crates/ide-completion/src/tests/attribute.rs
```diff
@@ -713,6 +713,39 @@ struct Foo;
...
+fn no_completion_in_attr_strings() {
+    check(
+        r#"
+fn test() -> u32 {
+    return 0;
+}
+
+#[derive(serde::Deserialize)]
+struct A {
+    #[serde(default = "t$0")]
```
ChayimFriedman2, line 726: That test isn't really testing anything because we don't have `serde` in scope, so the string is just a normal string token.

To get a proper test you will need to write a custom proc macro expander (this is not like writing real proc macros), which is a bit involved. If you want to take the challenge, [here](https://github.com/rust-lang/rust-analyzer/blob/6725e046df7493c1047e115ebc7180fb06416038/crates/test-fixture/src/lib.rs#L580) are some proc macros for reference. If not, that's also fine; I'll complete the test later. Thank you anyway!

vishruth-thimmaiah: I'd like to work on this, I have one question: how do I let rust analyzer know that the value in a proc macro wrapped in double quotes is a function? Something like:
```rust
#[proc_macro_attribute]
pub fn function_attr(args: TokenStream, input: TokenStream) -> TokenStream {
    let args = parse_macro_input!(args as FunctionAttrArgs);
    let input = parse_macro_input!(input as syn::ItemFn);
    
    let function = args.function.value();
    let function_ident = syn::Ident::new(&function, proc_macro2::Span::call_site());
    let fn_name = &input.sig.ident;

    let expanded = quote! {
        fn #fn_name() {
            #function_ident();
        }
    };

    expanded.into()
}

```
will give a diagnostic error, but completions do not show up like they would have with serde(default = "$0") in previous versions of r-a.

ChayimFriedman2: You need to set the span. But working with spans in rust-analyzer tests is not the same as with syn.
Because this is just a test, make yourself an easy life and find the offset of the path string inside the macro. Say it's the 4th token (it can be nested inside subtrees). Then, you can use the `quote!()` macro (of rust-analyzer, not of the `quote` crate, whose syntax is `quote! { span=> tokens }. Just put a call to the function inside `tokens` (of course, as a valid item).

Veykril: It would likely be even easier to make it a function-like proc-macro that is just called as `foo!("bar")`, then you only need to pick the first token, strip the quotes and turn it into an identifier.

crates/ide-completion/src/context/analysis.rs
```diff
@@ -417,6 +417,14 @@ fn analyze(
         derive_ctx,
     } = expansion_result;
 
+    if original_token.kind() != self_token.kind()
+        && !(original_token.kind().is_punct() || original_token.kind().is_trivia())
```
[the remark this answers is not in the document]
vishruth-thimmaiah: Without those checks, the following test cases fail:
- render::macro_::tests::complete_missing_macro_arg
- tests::expression::in_macro_expr_frag

original_token is a WHITESPACE and LPAREN respectively.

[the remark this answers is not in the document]
Veykril: Ah I see, thanks for pointing to the tests. Yes we do indeed need this right now, though only due to some limitations atm. Can you add a comment above this specific line in the condition the following comment
`// FIXME: This check can be removed once we use speculative database forking for completions`

(its a limitation with our incremental database engine called salsa)

crates/ide-completion/src/tests/attribute.rs
vishruth-thimmaiah: I tried the following:
```rust
#[derive(Debug)]
struct Issue17479ProcMacroExpander;
impl ProcMacroExpander for Issue17479ProcMacroExpander {
    fn expand(
        &self,
        _: &TopSubtree,
        attrs: Option<&TopSubtree>,
        _: &Env,
        _: Span,
        call_site: Span,
        _: Span,
        _: Option<String>,
    ) -> Result<TopSubtree, ProcMacroExpansionError> {
        let TokenTree::Leaf(Leaf::Literal(lit)) = &attrs.unwrap().0[1] else {
            return Err(ProcMacroExpansionError::Panic("incorrect Input".into()));
        };
        let symbol = lit.symbol.clone();

        Ok(quote! { call_site =>
            #symbol()
        })
    }
}
```

The test case:
```rust
#[test]
fn issue_17479() {
    check(
        r#"
//- proc_macros: issue_17479
fn test() {
    return;
}

#[proc_macros::issue_17479("t$0")] struct Test;
"#,
        expect![[r#""#]],
    )
}
```
From what I can tell, the macro is generated properly, but the test does not fail with `master`. Could you help me if I'm missing something in the test setup?

ChayimFriedman2: I believe (not sure) string literal symbols include the quotes (you can print them and see).

crates/ide-completion/src/tests/attribute.rs
```diff
@@ -713,6 +713,28 @@ struct Foo;
...
+    check(
+        r#"
+//- proc_macros: issue_17479
+fn main() {
+    proc_macros::issue_17479!("te$0");
+}
+"#,
+        expect![""],
+    );
+    check(
```
ChayimFriedman2, line 727: Why is the check duplicated?

crates/test-fixture/src/lib.rs
```diff
@@ -761,3 +776,28 @@ impl ProcMacroExpander for ShortenProcMacroExpander {
...
+        _: Span,
+        _: Option<String>,
+    ) -> Result<TopSubtree, ProcMacroExpansionError> {
+        let TokenTree::Leaf(Leaf::Literal(lit)) = &subtree.0[1] else {
+            return Err(ProcMacroExpansionError::Panic("incorrect Input".into()));
+        };
+        let symbol = &lit.symbol;
+        let span = lit.span;
+        Ok(quote! { span =>
+            #symbol()
```
ChayimFriedman2, line 800: Did you check that the test fail without your changes? Because I think `lit.symbol` contains the quote.

ChayimFriedman2 (changes requested):

crates/ide-completion/src/tests/attribute.rs
vishruth-thimmaiah: I saw noticed some tests check for both empty and partially completed fields, that is why I added two checks.

ChayimFriedman2 (approved):

PR #18845, opened 2025-01-06
fix: Fix flycheck getting confused which package to check

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/18844, the code here is now a lot clearer as well, no longer mixing two clear branches of logic


Veykril:
@mhnap thanks for the on-going investigations regarding this, I believe this should finally fix it for good (and least from my testing it works correctly now)

PR #18848, opened 2025-01-06
fix: Fix non-cargo flychecks immediately clearing received diagnostics

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/18847#issuecomment-2573022201


crates/rust-analyzer/src/flycheck.rs
```diff
@@ -419,6 +420,7 @@ impl FlycheckActor {
             self.command_receiver.take();
             self.report_progress(Progress::DidCancel);
             self.diagnostics_cleared_for.clear();
+            self.diagnostics_received = false;
```
Veykril, line 423: The bug for things not clearing is here, this line and the above should not be gated by the surrounding if let

PR #18878, opened 2025-01-08
manual: Document all rust-project.json fields

Wilfred:
Ensure that all the fields that rust-analyzer understands are in the manual, they all have doc comments, and they use consistent punctuation (`;` rather than mixing `,` and `;`).

Whilst we're here, fix the `sysroot_src` example and add 2024 as a legal value for Rust edition.

### DIFF
--- a/docs/user/manual.adoc
+++ b/docs/user/manual.adoc
@@ -716,6 +716,32 @@ interface JsonProject {
     /// dependencies as well as sysroot crate (libstd,
     /// libcore and such).
     crates: Crate[];
+    /// Configuration for CLI commands.
+    ///
+    /// These are used for running and debugging binaries
+    /// and tests without encoding build system-specific
+    /// knowledge into rust-analyzer.
+    ///
+    /// # Example
+    ///
+    /// Below is an example of a test runnable. `{label}` and `{test_id}`
+    /// are explained in `Runnable::args`'s documentation below.
+    ///
+    /// ```json
+    /// {
+    ///     "program": "buck",
+    ///     "args": [
+    ///         "test",
+    ///          "{label}",
+    ///          "--",
+    ///          "{test_id}",
+    ///          "--print-passing-details"
+    ///     ],
+    ///     "cwd": "/home/user/repo-root/",
+    ///     "kind": "testOne"
+    /// }
+    /// ```
+    runnables?: Runnable[];
 }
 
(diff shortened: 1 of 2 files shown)


docs/user/manual.adoc
```diff
@@ -776,21 +805,63 @@ interface Crate {
...
+    program: string;
+    /// The arguments passed to `program`.
+    ///
+    /// The args can contain two template strings: `{label}` and `{test_id}`.
+    /// rust-analyzer will find and replace `{label}` with `Build::label` and
+    /// `{test_id}` with the test name.
+    args: string[];
+    /// The current working directory of the runnable.
+    cwd: string;
+    kind: 'check' | 'run' | 'testOne';
```
lnicola, line 864: Let's give `kind` a comment too.

Wilfred: I think `testOne` is the only value that's used when you have a rust-project.json, so I've documented that and removed the other mentions.

docs/user/manual.adoc
```diff
@@ -776,21 +805,71 @@ interface Crate {
...
+    program: string;
+    /// The arguments passed to `program`.
+    ///
+    /// The args can contain two template strings: `{label}` and `{test_id}`.
+    /// rust-analyzer will find and replace `{label}` with `Build::label` and
+    /// `{test_id}` with the test name.
+    args: string[];
+    /// The current working directory of the runnable.
+    cwd: string;
+    /// Used to decide what CodeLens to offer.
```
lnicola, line 864:
```suggestion
    /// Used to decide what code lens to offer.
```

docs/user/manual.adoc
```diff
@@ -776,21 +805,71 @@ interface Crate {
...
+    ///
+    /// 'testOne': This runnable will be used when the user clicks the 'Run Test'
+    /// CodeLens above a test.
+    ///
+    /// The args can contain two template strings: `{label}` and `{test_id}`.
```
lnicola, line 869: This seems copy-pasted from above, by mistake.

docs/user/manual.adoc
Wilfred: Updated. FWIW VS Code uses this capitalisation, but I agree that 'code lens' reads more nicely here.

lnicola (approved):

PR #18880, opened 2025-01-08
Add config setting which allows adding additional include paths to the VFS.

PrototypeNM1:
The motivation for this is resolving unresolved symbols for Firefox. We build Rust code with cargo but generate some code via our primary build system mach. Additionally, we bundle the windows crate in an out-of-source and out-of-build location. Essentially the same idea as discussed [here](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/include!.20failing.20for.20some.20paths), Firefox issue tracked [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1890431).

I investigated instead populating the vfs via rustc's `--emit=dep-info` as cargo already emits .d files but at this time cargo [does not relay .d file changes](https://github.com/rust-lang/cargo/issues/13672#issuecomment-2147772916) as it hazards the unspecified cargo directory layout becoming defacto specified.

This change therefore instead adds a config option to manually specify extra includes for the VFS as suggested in the above Zulip conversation and in this [bug](https://github.com/rust-lang/rust-analyzer/issues/17040#issuecomment-2105778865) (similarly motivated by Firefox's needs).

I don't believe this change ready to land but it's in a place where it's sufficient to anchor a conversation on what's necessary beyond a minimal implementation, and I've verified this works with both Firefox's XPCOM generated functions and windows includes.

### DIFF
--- a/crates/project-model/src/cargo_workspace.rs
+++ b/crates/project-model/src/cargo_workspace.rs
@@ -92,6 +92,8 @@ pub struct CargoConfig {
     pub sysroot_src: Option<AbsPathBuf>,
     /// rustc private crate source
     pub rustc_source: Option<RustLibSource>,
+    /// Extra includes to add to the VFS.
+    pub extra_includes: Vec<AbsPathBuf>,
     pub cfg_overrides: CfgOverrides,
     /// Invoke `cargo check` through the RUSTC_WRAPPER.
     pub wrap_rustc_in_build_scripts: bool,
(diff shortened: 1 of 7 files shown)


Veykril (approved):
I am not sure whether adding this to all local roots will work (I fear it might break the sourceroot setup), but generally this seems fine to me to temporarily paper over the issues people have until we have it properly fixed (or a proper way to add these things on a finer basis).

Veykril:
You'll need to run the tests in the rust-analyzer crate (that will regenerate some files) `cargo test -p rust-analyzer` as well as move the config so its in lexicographic order (that one should be done first as it affects the generated file).

PrototypeNM1:
Is there an issue I can link to for tracking the proper fix?

Veykril:
https://github.com/rust-lang/rust-analyzer/issues/9173

eronnen:
@Veykril Is it possible to use this extra include paths config in order to fix https://github.com/rust-lang/rust-analyzer/issues/9173? If so, what paths to configure?

PR #18885, opened 2025-01-08
refactor: struct holding cargo cfgs settings

qjerome:
# Change to `rust-analyzer.cargo.cfgs` Setting

This PR replaces the current structure (a `HashMap`) for the `rust-analyzer.cargo.cfgs` setting with a `Vec<String>`. The use of a vector seems more appropriate for the following reasons:

- Repeated same `cfg` keys can be used (currently impossible due to the use of a `HashMap`).
- In the end, the original `HashMap` is turned into a vector of (key, value) tuples, so it seems there is no benefit to using a `HashMap` to hold these settings.

A beneficial side effect of this change is that `rust-analyzer` can have partial support for multi-target analysis through this setting (see the following configuration example).

## New Format Example

```json
{
  "rust-analyzer.cargo.cfgs": [
    "debug_assertion",
    "miri",
    "target_arch=x86_64",
    "target_arch=aarch64",
  ]
}

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -571,12 +571,10 @@ config_data! {
         /// avoid checking unnecessary things.
         cargo_buildScripts_useRustcWrapper: bool = true,
         /// List of cfg options to enable with the given values.
-        cargo_cfgs: FxHashMap<String, Option<String>> = {
-            let mut m = FxHashMap::default();
-            m.insert("debug_assertions".to_owned(), None);
-            m.insert("miri".to_owned(), None);
-            m
-        },
+        cargo_cfgs: Vec<String> = {
+            vec!["debug_assertion".into(), "miri".into()]
+        }
+        ,
         /// Extra arguments that are passed to every cargo invocation.
         cargo_extraArgs: Vec<String> = vec![],
         /// Extra environment variables that will be set when running cargo, rustc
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: crates/rust-analyzer/src/config.rs [hunk 1]
```diff
@@ -1944,6 +1942,17 @@ impl Config {
...
                         .iter()
+                        // parse any cfg setting formatted as key=value
+                        .map(|s| {
+                            let mut sp = s.splitn(2, "=");
+                            let key = sp.next();
+                            let val = sp.next();
+                            (key, val)
+                        })
+                        // we filter out anything with a None key
+                        .filter(|(key, _)| key.is_some())
```
lnicola, line 1953: You can avoid the `unwrap` and do this in one step using `filter_map`.

crates/rust-analyzer/src/config.rs
```diff
@@ -571,12 +571,10 @@ config_data! {
...
         /// List of cfg options to enable with the given values.
-        cargo_cfgs: FxHashMap<String, Option<String>> = {
-            let mut m = FxHashMap::default();
-            m.insert("debug_assertions".to_owned(), None);
-            m.insert("miri".to_owned(), None);
-            m
-        },
+        cargo_cfgs: Vec<String> = {
+            vec!["debug_assertion".into(), "miri".into()]
+        }
```
lnicola, line 576:
```suggestion
        },
```

lnicola:
`cargo test` should fix the CI issue.

crates/rust-analyzer/src/config.rs
qjerome: done

Veykril (approved):

qjerome:
> No it still doesn't. What you are using it for merely enables CFGs that are assumed by the ecosystem to be disjoint. It may improve your specific situation here but it doesn't change the fact that r-a cannot handle multiple targets in a crate at once.

Yep, I fully agree with this statement. Yet I believe it provides a working alternative to opening several IDE with different r-a settings which becomes heavier when the number of targets increases. I actually hope it is gonna help others having the same issues as I have (being lazy enough to open several IDE instances to develop each of the targets).

Thank you @lnicola for the review and @Veykril for the merging

PR #18908, opened 2025-01-10
Fix: Detect missing errors for } braces before else in let...else statements

### LINKED ISSUE #18345: Missing errors for `}` braces before `else` in `let...else` statement
The following are snippets that are denied by rustc, but rust-analyzer's parser does not emit an error.
```rs


struct X {a: i32}
fn f() {
    let foo = X {
        a: 1
    } else {
        //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        return;
    };
}

fn i() {
    let foo = &{
        1
    } else {
        //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        return;
    };
}

fn j() {
    let bar = 0;
    let foo = bar = {
        1
    } else {
        //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        return;
    };
}

fn k() {
    let foo = 1 + {
        1
    } else {
        //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        return;
    };
}

fn l() {
    let foo = 1..{
        1
    } else {
        //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        return;
    };
}

fn m() {
    let foo = return {
        ()
    } else {
        //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        return;
    };
}

fn n() {
    let foo = -{
        1
    } else {
        //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        return;
    };
}

fn o() -> Result<(), ()> {
    let foo = do yeet {
        ()
    } else {
        //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        return Ok(());
    };
}

fn p() {
    let foo = become {
        ()
    } else {
        //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        return;
    };
}

fn q() {
    let foo = |x: i32| {
        x
    } else {
        //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        return;
    };
}

fn r() {
    let ok = format_args!("") else { return; };

    let bad = format_args! {""} else { return; };
    //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
}

fn s() {
    macro_rules! a {
        () => { {} }
    }

    macro_rules! b {
        (1) => {
            let x = a!() else { return; };
        };
        (2) => {
            let x = a! {} else { return; };
            //~^ ERROR right curly brace `}` before `else` in a `let...else` statement not allowed
        };
    }

    b!(1); b!(2);
}

```

These are taken from rustc's test suite

jnyfah:
fixes #18345 - rust-analyzer fails to emit errors for let...else statements when the initializer expression ends in a block-like construct

- Updated `CompletedMarker` to store the position where the node was completed.
- Introduced a new `last_token` function that walks back through the parsing events from the completion position to find the last emitted token.

### DIFF
--- a/crates/parser/src/grammar/expressions.rs
+++ b/crates/parser/src/grammar/expressions.rs
@@ -134,10 +134,12 @@ pub(super) fn let_stmt(p: &mut Parser<'_>, with_semi: Semicolon) {
         // test_err let_else_right_curly_brace
         // fn func() { let Some(_) = {Some(1)} else { panic!("h") };}
         if let Some(expr) = expr_after_eq {
-            if BlockLike::is_blocklike(&expr, p) {
-                p.error(
-                    "right curly brace `}` before `else` in a `let...else` statement not allowed",
-                )
+            if let Some(token) = expr.last_token(p) {
+                if token == T!['}'] {
+                    p.error(
+                        "right curly brace `}` before `else` in a `let...else` statement not allowed"
+                    )
+                }
             }
         }
 
--- a/crates/parser/src/grammar/expressions.rs
+++ b/crates/parser/src/grammar/expressions.rs
@@ -134,10 +134,12 @@ pub(super) fn let_stmt(p: &mut Parser<'_>, with_semi: Semicolon) {
         // test_err let_else_right_curly_brace
         // fn func() { let Some(_) = {Some(1)} else { panic!("h") };}
         if let Some(expr) = expr_after_eq {
-            if BlockLike::is_blocklike(expr.kind()) {
-                p.error(
-                    "right curly brace `}` before `else` in a `let...else` statement not allowed",
-                )
+            if let Some(token) = expr.last_token(p) {
+                if token == T!['}'] {
+                    p.error(
+                        "right curly brace `}` before `else` in a `let...else` statement not allowed"
+                    )
+                }
             }
         }
 
--- a/crates/parser/src/grammar/expressions.rs
+++ b/crates/parser/src/grammar/expressions.rs
@@ -134,12 +134,10 @@ pub(super) fn let_stmt(p: &mut Parser<'_>, with_semi: Semicolon) {
         // test_err let_else_right_curly_brace
         // fn func() { let Some(_) = {Some(1)} else { panic!("h") };}
         if let Some(expr) = expr_after_eq {
-            if let Some(token) = expr.last_token(p) {
-                if token == T!['}'] {
-                    p.error(
-                        "right curly brace `}` before `else` in a `let...else` statement not allowed"
-                    )
-                }
+            if BlockLike::is_blocklike(&expr, p) {
+                p.error(
+                    "right curly brace `}` before `else` in a `let...else` statement not allowed",
+                )
             }
         }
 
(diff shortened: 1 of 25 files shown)


crates/parser/src/grammar/expressions.rs
```diff
@@ -134,10 +134,12 @@ pub(super) fn let_stmt(p: &mut Parser<'_>, with_semi: Semicolon) {
         // test_err let_else_right_curly_brace
         // fn func() { let Some(_) = {Some(1)} else { panic!("h") };}
         if let Some(expr) = expr_after_eq {
-            if BlockLike::is_blocklike(expr.kind()) {
```
Veykril, line 137: I think we should actually change how `BlockLike` works, it should be checking the syntax kind and the last token (for the kinds where the node is conditionally a block). That way we likely fix a bunch of bugs in other places where `BlockLike` is used

crates/parser/src/parser.rs
```diff
@@ -386,4 +388,15 @@ impl CompletedMarker {
     pub(crate) fn kind(&self) -> SyntaxKind {
         self.kind
     }
+
+    pub(crate) fn last_token(&self, p: &Parser<'_>) -> Option<SyntaxKind> {
+        let end_pos = self.end_pos as usize;
+        if end_pos > p.events.len() {
+            return None;
+        }
```
Veykril, line 396:
```suggestion
        debug_assert_eq!(p.events[end_pos - 1], Event::Finish);
```
The early return should never happen here

crates/parser/src/parser.rs
```diff
@@ -386,4 +388,15 @@ impl CompletedMarker {
...
+        p.events[..end_pos].iter().rev().find_map(|event| match event {
```
Veykril, line 397: I think this should be the following (to properly account for empty nodes)
```suggestion
        p.events[self.start_pos..end_pos].iter().rev().find_map(|event| match event {
```
BUT, the `forward_parent` stuff might cause issues for this and empty nodes aren't really a thing in r-a so keeping it as is ought to be fine

crates/parser/src/grammar/expressions.rs
jnyfah: checking the last token in `BlockLike` breaks some closure parsing because closures with block bodies will being considered block-like 
eg: `fn main() { || -> i32 { 92 }(); }` the closure is considered blocklike even though the `{ 92 }` part is the body of the closure, not a standalone block expression.

one way to handle this tho would be explicitly excluding `CLOSURE_EXPR` when checking for trailing braces

Veykril: Ah I see, let's keep it as is for now then actually. There might be other problems like that then.

Veykril (commented):
Ah, forgot to mention something, adding more tests would be nice :) (the fixture posted in the issue ideally). Just throw that in https://github.com/rust-lang/rust-analyzer/tree/master/crates/parser/test_data/parser/ok as a new .rs file, then running the tests with env var `UPDATE_EXPECT=1` set should generate the corresponding syntax tree file

jnyfah:
> Ah, forgot to mention something, adding more tests would be nice :) (the fixture posted in the issue ideally). Just throw that in https://github.com/rust-lang/rust-analyzer/tree/master/crates/parser/test_data/parser/ok as a new .rs file, then running the tests with env var `UPDATE_EXPECT=1` set should generate the corresponding syntax tree file

@Veykril All test passes apart from `macros` probably because of how `macros` are being parsed? not so sure

Veykril (approved):
Thanks!

PR #18909, opened 2025-01-10
Use `strict_provenance`

Veykril:
Given its stable now, though we likely wanna wait some time until we bump our crates


Veykril:
Ah I forgot about that discussion 😅

PR #18912, opened 2025-01-10
fix: upmap ranges in convert_tuple_struct_to_named_struct assist

### LINKED ISSUE #18766: `convert_tuple_struct_to_named_struct` produces invalid code
**rust-analyzer version**: 2024-12-23

**editor or extension**: Reproduced under both Neovim and VSCode

```rust
// add your code here
use std::{borrow::Borrow, debug_assert, rc::Rc};

#[derive(Clone, PartialEq)]
pub enum Expr {
    NontailCommand(Ustr, usize, ChicSpan),
}

#[cfg(test)]
pub mod tests {
    use super::*;
    use Expr::*;

    #[test]
    fn parses_nontail_command() {
        assert!(matches!(e, NontailCommand(s, 0, ..) if s == "foo"));
    }
}
```

Reproduction:
1. Position cursor at `Expr::NontailCommand` and select "convert to named struct" code action:
Image
2. Code action inserts unexpected text in the first line of the buffer:Image

Same thing under VSCode:
Image

vishruth-thimmaiah:
Fixes #18766

This PR upmaps ranges for the `convert_tuple_struct_to_named_struct` ide-assist, fixing broken substitutions within macros.

### DIFF
--- a/crates/ide-assists/src/handlers/convert_tuple_struct_to_named_struct.rs
+++ b/crates/ide-assists/src/handlers/convert_tuple_struct_to_named_struct.rs
@@ -140,8 +140,10 @@ fn edit_struct_references(
         match_ast! {
             match node {
                 ast::TupleStructPat(tuple_struct_pat) => {
+                    let file_range = ctx.sema.original_range_opt(&node)?;
+                    edit.edit_file(file_range.file_id);
                     edit.replace(
-                        tuple_struct_pat.syntax().text_range(),
+                        file_range.range,
                         ast::make::record_pat_with_fields(
                             tuple_struct_pat.path()?,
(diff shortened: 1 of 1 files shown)


Veykril (changes requested):
Actual upmapping is done via the `Semantics::original_range*` functions which returns a file range (opposed to a text range), the important part there is to account for the file id that may be different than the one that is currently assumed by the code

Veykril:
> The existing functionality already handles file ids properly, I don't think additional changes are needed.

The file id we get from here might be different, additional `original_range` will fall back to the entire macro call if it fails upmapping which is not what we want. So this should make proper use of the file id and we also likely wanna use `original_range_opt`

Veykril (approved):

PR #18917, opened 2025-01-11
feat: Add dereferencing autocomplete

boattime:
This PR adds dereferencing suggestions for Copy types. When a function expects a non-reference type T and we have a &T where T: Copy, we now suggest the dereferenced value in completions.

https://github.com/rust-lang/rust-analyzer/issues/18873

### DIFF
--- a/crates/ide-completion/src/item.rs
+++ b/crates/ide-completion/src/item.rs
@@ -79,7 +79,7 @@ pub struct CompletionItem {
     // FIXME: We shouldn't expose Mutability here (that is HIR types at all), its fine for now though
     // until we have more splitting completions in which case we should think about
     // generalizing this. See https://github.com/rust-lang/rust-analyzer/issues/12571
-    pub ref_match: Option<(Mutability, TextSize)>,
+    pub ref_match: Option<(CompletionItemRefMode, TextSize)>,
 
     /// The import data to add to completion's edits.
     /// (ImportPath, LastSegment)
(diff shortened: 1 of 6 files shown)


crates/ide-completion/src/item.rs
```diff
@@ -441,15 +454,18 @@ impl CompletionItem {
...
+                CompletionItemRefMode::Reference(Mutability::Shared) => {
+                    (format!("&{}", self.label.primary), "&")
+                }
+                CompletionItemRefMode::Reference(Mutability::Mut) => {
+                    (format!("&mut {}", self.label.primary), "&mut ")
+                }
+                CompletionItemRefMode::Dereference => (format!("*{}", self.label.primary), "*"),
+            };
+
+            (label, ide_db::text_edit::Indel::insert(offset, String::from(edit_text)), relevance)
```
Veykril, line 468:
```suggestion
            let prefix= match mode {
                CompletionItemRefMode::Reference(Mutability::Shared) => "&",
                CompletionItemRefMode::Reference(Mutability::Mut) => "&mut ",
                CompletionItemRefMode::Dereference =>"*",
            };
            let label = format!("{prefix}{}", self.label.primary),

            (label, ide_db::text_edit::Indel::insert(offset, String::from(prefix)), relevance)
```

crates/ide-completion/src/render.rs
```diff
@@ -638,20 +639,31 @@ fn compute_exact_name_match(ctx: &CompletionContext<'_>, completion_name: &str)
...
+                    hir::Mutability::Mut
+                } else {
+                    hir::Mutability::Shared
+                };
+                return Some(CompletionItemRefMode::Reference(mutability));
+            }
+        }
+    }
+
+    if let Some(completion_ty_without_ref) = completion_ty.remove_ref() {
```
Veykril, line 659: We can move this to the start of the function as `let expected_type_without_ref = expected_type.remove_ref()?;` since the branch above makes use of it as well either way to deduplicate a bit

crates/rust-analyzer/src/lib.rs
```diff
@@ -132,8 +133,16 @@ fn completion_item_hash(item: &CompletionItem, is_ref_completion: bool) -> [u8;
...
+    if let Some((ref_mode, text_size)) = &item.ref_match {
+        match ref_mode {
+            CompletionItemRefMode::Reference(mutability) => match mutability {
+                Mutability::Shared => hasher.update("&"),
+                Mutability::Mut => hasher.update("&mut"),
+            },
+            CompletionItemRefMode::Dereference => {
+                hasher.update("*");
+            }
+        }
```
Veykril, line 145: Similar to above, we can move just return the strings in the match, then call `hasher.update` on that

crates/ide-completion/src/render.rs
boattime: Agreed I'll make the change

boattime: @Veykril I pushed the update to move both of these methods to the top of the function. Let me know if this is not what you were looking for.

Veykril (approved):
Thanks!

PR #18933, opened 2025-01-14
fix: Update `notify` to `8.0.0`

### LINKED ISSUE #18932: Server process exited with code 3221225501 after 0.4.2216
**rust-analyzer version**: `0.4.2216-standalone (9b2e72c404 2024-12-12)`

**rustc version**: `rustc 1.85.0-nightly (4ba4ac612 2024-12-18)`

**editor or extension**: VS Code (1.96.3 / extension 0.4.2263)

**repository link (if public, optional)**: private :(

Since `0.4.2216-standalone (9b2e72c404 2024-12-12)`, one of my workspaces (inconveniently the largest and private-est) has seen `rust-analyzer` crashing whenever a file is saved. This workspace seems to be the only one of mine with this issue. R-A versions prior to this, like `0.4.2215-standalone (b707b5a117 2024-12-11)`, do not have this issue.

`0.4.2216` will always spit out this error:

```
thread 'VfsLoader' panicked at crates\vfs-notify\src\lib.rs:94:72:
called `Result::unwrap()` on an `Err` value: RecvError
stack backtrace:
```

before exiting with code `3221225501` (`0xC000001D`, seemingly illegal instruction?).

Versions after about `0.4.2227` will forego the unwrap panic and instead just crash with the illegal instruction. Versions after about `0.4.2251` will trigger the crash on *any* interaction, like a hover. Most are able to start up without crashing, though. (I didn't have time to do a more precise bisect, sorry)

I was only able to get `info` logs from the server, as `debug` was too verbose and VS Code would clear the output itself. Unfortunately this didn't reveal much as there didn't appear to be any indicators of failure before `Connection to server got closed. Sever will restart.`; nonetheless they are available [here](https://spaces.pyke.io/deca/ra-log.txt) (crash @ 87455, status code @ 89425)

decahedron1:
Fixes #18932 by updating the `notify` crate to v8.0, which fixed UB on Windows that resulted in a `STATUS_ILLEGAL_INSTRUCTION` in certain workspaces when `rust-analyzer` was compiled in release mode.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -98,9 +98,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.6.0"
+version = "2.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b048fb63fd8b5923fc5aa7b340d8e156aec7ec02f0c78fa8a6ddc2613f6f71de"
+checksum = "1be3f42a67d6d345ecd59f675f3f012d6974981560836e938c22b424b85ce1be"
 
 [[package]]
 name = "borsh"
@@ -210,7 +210,7 @@ version = "0.98.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d5f2eb1cd6054da221bd1ac0197fb2fe5e2caf3dcb93619398fc1433f8f09093"
 dependencies = [
- "bitflags 2.6.0",
+ "bitflags 2.7.0",
  "chalk-derive",
 ]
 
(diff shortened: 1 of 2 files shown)


lnicola:
Thanks! The CI fails because of [this](https://github.com/rust-lang/rust-analyzer/blob/8364ef299790cb6ec22b9e09e873c97dbe9f2cb5/xtask/src/tidy.rs#L137) license check.

But I think we should go directly to 8.0 if we're upgrading.

lnicola:
Oh, did they change the license again in 8.0?

lnicola (approved):

decahedron1:
v7 had a dependency on the [`instant`](https://crates.io/crates/instant) crate (BSD-3) on WASM, which was replaced with [`web-time`](https://crates.io/crates/web-time) (MIT/Apache-2.0) in v8.

lnicola:
@decahedron1 just to be sure, did this fix your crash in #18932?

decahedron1:
Yep! (also, for that issue, I did have `"rust-analyzer.files.watcher": "server"` set)

PR #18946, opened 2025-01-16
feat: show go-to-type-def actions for func param and trait bound when hovering

lh123:


crates/ide/src/hover.rs
```diff
@@ -557,12 +557,28 @@ fn goto_type_action_for_def(
             .into_iter()
             .filter(|&it| Some(it.into()) != sized_trait)
             .for_each(|it| push_new_def(it.into()));
+    } else if let Definition::Function(function) = def {
+        walk_and_push_ty(db, &function.ret_type(db), &mut push_new_def);
+
+        let krate = function.module(db).krate();
+        let sized_trait =
+            db.lang_item(krate.into(), LangItem::Sized).and_then(|lang_item| lang_item.as_trait());
+        for param in function.params_without_self(db) {
```
Veykril, line 566: We might as well include the self parameter here as well

PR #18950, opened 2025-01-16
Generalize some type walking in hover type actions

Veykril:
cc https://github.com/rust-lang/rust-analyzer/pull/18946

### DIFF
--- a/crates/ide/src/hover.rs
+++ b/crates/ide/src/hover.rs
@@ -6,7 +6,9 @@ mod tests;
 use std::{iter, ops::Not};
 
 use either::Either;
-use hir::{db::DefDatabase, GenericSubstitution, HasCrate, HasSource, LangItem, Semantics};
+use hir::{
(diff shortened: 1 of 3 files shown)


crates/ide/src/hover.rs
```diff
@@ -599,7 +590,9 @@ fn walk_and_push_ty(
     ty: &hir::Type,
     push_new_def: &mut dyn FnMut(hir::ModuleDef),
 ) {
+    dbg!(&ty);
     ty.walk(db, |t| {
+        dbg!(&t);
```
lh123, line 595: remoe `dbg!`?

crates/ide/src/hover.rs
```diff
@@ -599,7 +590,9 @@ fn walk_and_push_ty(
     ty: &hir::Type,
     push_new_def: &mut dyn FnMut(hir::ModuleDef),
 ) {
+    dbg!(&ty);
```
lh123, line 593: remoe `dbg!`?

PR #18952, opened 2025-01-16
feat: complete raw, const keyword

lh123:
1. complete `raw`, `const` keyword
2. parse `let _ = &raw foo` better


Veykril (commented):
One test caseto add would be `&mut $0` (to verify we don't complete `raw` here)

Veykril (approved):

Veykril:
Thanks!

Issue #18955, opened 2025-01-16
VSCode Test Explorer - bugs running tests for packages with multiple targets


duncanawoods:
There are a number of bugs running tests for packages that define multiple targets:

1.

Running tests for a target can run unrelated tests in other packages / targets.

RA tries to run a test from a fully qualified name like:

    root::module::test_name

a) RA assumes the root is a package name but it can be the name of a target inside a package.

b) RA does not supply `cargo test` with target parameters e.g. ` --bin name` to run in the correct test

c) if test code is shared between multiple targets:

- RA will be asked to run a test for each target
- RA will try to combine the name of all tests into a single pattern with unpredictable results

e.g. if there are targets named "target1", "target2", "target3", RA will run every test in the workspace containing the word "target".

2. 
 
The results of a test are often assigned against the wrong target or crate

This is because the `cargo test` json output does not contain the package or target of the test it has run. RA assigns the result using `hack_recover_crate_name` which guesses at an answer.

duncanawoods:
# Suggested fixes

**1. Running Tests**

1 a) and 1 b) can be solved by extending `test_runner::TestTarget` and `request::find_package_name` to find a matchingtarget not just the package.

c) When invoking a test from source code shared by multiple targets, there are two main options:

i) run the test for all the targets => this is what vscode is trying to do right now and would be fixed by replacing RA's effort to combine tests into a single path and running each requested tests with separate invocations of `cargo test`.

ii) run the test for one target only => when developing a test, it's often undesirable to run it for every target. It's likely slow and redundant e.g. a common reason for multiple targets is a library with a cli wrapper. Identifying a default target to use would likely match the developer's intention.

**2. Test Results**

The current RA design tries to parse cargo command output without the context of the original command. This is insufficient to match test output to targets. The required fix is to extend the cargo command running infrastructure to make command context available when the output is being interpreted.

duncanawoods:
### Background Info

- the default target in a package can be a `lib` or a `bin`
- there can be multiple bin targets but only one lib
- tests can be specific to a target or in shared code used by multiple targets

I am testing against a workspace with two members that define additional targets:
 
- test_bin
  - bin target
  - contains
    - test_bin_lib : a lib target
    - test_bin2 : am extra bin target
- test_lib
  - lib target
  - contains
    - test_lib_bin1 : a bin target
    - test_lib_bin2 : a second bin target
 
Each with shared and target specific tests.

duncanawoods:
[...]

PR #18956, opened 2025-01-16
fix: correctly identify the target inside packages and match output

duncanawoods:
This is a PR to fix most of #18955

The two main changes:

1. update `request::handle_run_test` and `test_runner` to identify the target inside a package that matches a test request. This fixes RA failing to find the right package because the test uses the target name.

2. extend `CargoActor` to take a parser object rather than just a generic type with static parsing functions. This allows context about the invocation of cargo to be used when interpreting the cargo output so it can be properly integrated with the GUI. 

In this case, it's information about the test that has been run which isn't contained in cargo output. The current behaviour is very flaky so that if you have tests with the same name in multiple crates or targets, the results are often assigned to the wrong target.

The `StatelessParser` wrapper ensures existing static parsers can be used without any changes.

### DIFF
--- a/crates/project-model/src/project_json.rs
+++ b/crates/project-model/src/project_json.rs
@@ -63,7 +63,7 @@ use crate::{ManifestPath, TargetKind};
 pub struct ProjectJson {
     /// e.g. `path/to/sysroot`
     pub(crate) sysroot: Option<AbsPathBuf>,
-    /// e.g. `path/to/sysroot/lib/rustlib/src/rust`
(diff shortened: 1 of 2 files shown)


duncanawoods:
This fixes 1 a), 1 b) and 2. from the issue.

1 c) requires more thought / discussion:

- changing the test_runner to launch multiple cargo test processes would be significant behaviour change
- introducing the notion of a `default target` or setting deserves some thought

PR #18964, opened 2025-01-17
internal: port rust-analyzer to new Salsa

davidbarsky:
This PR ports rust-analyzer to the [new Salsa](https://github.com/salsa-rs/salsa/). It is big, partially incomplete, and best reviewed commit-by-commit. I will be rebasing often.

(Why new Salsa? It unblocks a lot of work described in https://github.com/rust-lang/rust-analyzer/issues/17491.)

## Migration Approach

To get rust-analyzer compiling and passing tests, this PR did the following:
- With @nikomatsakis's help, I created [`db-ext-macro`](https://github.com/davidbarsky/db-ext-macro/) to mimic rust-analyzer's usage of old [Salsa's query groups](https://github.com/salsa-rs/salsa-rfcs/blob/master/RFC0001-Query-Group-Traits.md). Over time, rust-analyzer will migrate to new Salsa's idioms, moving away from the query group pattern. For additional details on the differences between old Salsa and new Salsa, please refer to [this document](https://hackmd.io/z5M6WYHFQ6qKxYF8Wwjv_w).
- @Veykril ported `crates/span` to new Salsa, which makes use of [self-referential interning](https://github.com/salsa-rs/salsa/pull/633). This is needed to support macro hygiene.
- New Salsa's structs typically have a `'db` lifetime. However, the presence of the `'db` lifetime made migrating a large number of interned structs challenging, hence https://github.com/salsa-rs/salsa/pull/632.

With these changes, _all tests were passing_ (no correctness issues!), `analysis-stats` produced the same results as rust-analyzer on the old Salsa, and—critically—the new Salsa-powered rust-analyzer was usable as my daily driver. Unfortunately, memory usage was too high. I'm really grateful to the following folks who helped fix that regression:

- @ShoyuVanilla reduced rust-analyzer's memory usage on `analysis-stats` from 8 gigabytes to 4 gigabytes by through https://lgithub.com/salsa-rs/salsa/pull/614/.
- @ChayimFriedman2 and @Veykril made further improvements to Salsa's memory usage, which reduced usage by around [575 megabytes](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Porting.20to.20Salsa.203.2E0/near/491981774).
- Chayim implemented [`#[salsa::enum]`](https://github.com/salsa-rs/salsa/issues/578), which reduced memory usage by about 150 megabytes and made rust-analyzer's usage of new Salsa more idiomatic.
- I reenabled LRU, which reduced memory usage by an additional gigabyte.

## Performance

- On `rust-analyzer analysis-stats`, new Salsa is slightly slower and uses slightly more memory. 
    - On bfb81275, new Salsa uses 2201mb of memory, whereas old Salsa uses ~2000 megabytes.
- On `crates/rust-analyzer/src/integrated_benchmarks.rs`, new Salsa is slightly faster on both cold and warm updates. These benchmarks should probably be more rigorous, but new Salsa was _consistently_ faster.

## Things to Fix Before to Merging

There's a few things to do before this change is merged:
1. Reintroduce debug item printing. However, this is blocked on https://github.com/salsa-rs/salsa/pull/640 and some additional changes inside of rust-analyzer, but they're not complicated changes.
2. Figure out why the changes in the `enum`/[`#[salsa::supertype]`](https://github.com/salsa-rs/salsa/issues/578) commit cause tests to fail to type confusion in [the underlying Salsa pages](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Porting.20to.20Salsa.203.2E0/near/494209802). Chayim is working on root-causing this, but if he's unable to determine the cause, I can make the enum changes a separate pull request.
3. Release `db-ext-macro`.
    1. I'd like to rename `db-ext-macro` to something else, or have it live in Salsa. I'm unsure which option I'd prefer right now.
4. Release the new Salsa to crates.io.

## Acknowledgements

This wouldn't have been possible without the assistance of many folks including @ChayimFriedman2, @ShoyuVanilla, @nikomatsakis, and @Veykril. I'm sure I'm forgetting some people, to whom I apologize.


davidbarsky:
Chayim fixed the test failures: this was due us mixing and matching Salsa interning with rust-analyzer's own, home-grown interning. He put it succinctly [here](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Porting.20to.20Salsa.203.2E0/near/494538002):

> My macro also implements traits like `Eq` and `Hash`. And it implements them by hashing the ID only, because I thought it will be more efficient.
> But have we global, unrelated to salsa, interning. For example, that's how we intern `Ty`s. And this global interning will _persist_ even after we switch dbs. But in different dbs, different IDs can map to different kinds!
> So we might switch the type of an `AdtId`, for example, because we mapped it to the incorrect interned key!

As a short-term fix, he's required interned values to implement `PartialEq` and friends. Longer-term, we'll move this interner into new Salsa.

the8472:
> reduced rust-analyzer's memory usage on analysis-stats from 8 gigabytes to 4 gigabytes by through https://lgithub.com/salsa-rs/salsa/pull/614/.

contains a typo in the link

Veykril:
Regarding `db-ext-macro`'s home, we could also consider placing it in-tree here. That might be easiest tbh.

Cargo.toml
```diff
@@ -11,7 +11,7 @@ authors = ["rust-analyzer team"]
 repository = "https://github.com/rust-lang/rust-analyzer"
 
 [profile.dev]
-debug = 1
+debug = 2
```
Veykril, line 14:
```suggestion
debug = 1
```
(reminder)

Cargo.toml
```diff
@@ -27,7 +27,7 @@ miniz_oxide.opt-level = 3
 [profile.release]
 incremental = true
 # Set this to 1 or 2 to get more useful backtraces in debugger.
-debug = 0
+debug = 2
```
Veykril, line 30:
```suggestion
debug = 0
```
(reminder)
also for this you could use the `dev-rel` profile

crates/base-db/src/input.rs
Veykril, line 1: how come that there are formatting changes in here (without CI ever complaining about them) 🤨

crates/profile/src/lib.rs
```diff
@@ -55,13 +55,13 @@ pub fn cpu_span() -> CpuSpan {
...
-            r#"cpu profiling is disabled, uncomment `default = [ "cpu_profiler" ]` in Cargo.toml to enable."#
-        );
-    }
+    // #[cfg(not(feature = "cpu_profiler"))]
+    // #[allow(clippy::print_stderr)]
+    // {
+    //     eprintln!(
+    //         r#"cpu profiling is disabled, uncomment `default = [ "cpu_profiler" ]` in Cargo.toml to enable."#
+    //     );
+    // }
```
Veykril, line 64: Whats up with this

crates/span/src/lib.rs
```diff
@@ -1,19 +1,19 @@
 //! File and span related types.
 use std::fmt::{self, Write};
 
-#[cfg(feature = "ra-salsa")]
-use ra_salsa::InternId;
-
```
Veykril, line 6: Can we re-introduce this cfg'ing for the new salsa? The point of that is that the proc-macro server has no need to depend on salsa

crates/rust-analyzer/src/cli/rustc_tests.rs
Veykril, line 1: Whats up with the changes here?

crates/base-db/src/lib.rs
```diff
@@ -47,159 +51,255 @@ pub const DEFAULT_FILE_TEXT_LRU_CAP: u16 = 16;
...
+                    FileSourceRootInput::builder(source_root_id).durability(durability).new(db);
+                vacant.insert(file_source_root);
+            }
+        };
+    }
+}
+
+#[salsa::interned(no_lifetime)]
+pub struct EditionedFileId {
+    pub file_id: FileText,
```
Veykril, line 172:
```suggestion
    pub file_id: vfs::FileId,
```

Veykril (commented):
(partial review)

davidbarsky:
@Veykril  Regarding `crates/rust-analyzer/src/cli/rustc_tests.rs`, it looks like some parts of new Salsa—namely, `(dyn salsa::accumulator::accumulated::AnyAccumulated + 'static)`—don't implement `UnwindSafe`. Here's the error:

<details>

```rust
error[E0277]: the type `(dyn salsa::accumulator::accumulated::AnyAccumulated + 'static)` may not be safely transferred across an unwind boundary
    --> crates/rust-analyzer/src/cli/rustc_tests.rs:163:60
     |
163  |                           let res = std::panic::catch_unwind(move || {
     |  ___________________________________------------------------_^
     | |                                   |
     | |                                   required by a bound introduced by this call
164  | |                             analysis.full_diagnostics(
165  | |                                 diagnostic_config,
166  | |                                 ide::AssistResolveStrategy::None,
...    |
169  | |                         });
     | |_________________________^ `(dyn salsa::accumulator::accumulated::AnyAccumulated + 'static)` may not be safely transferred across an unwind boundary
     |
     = help: the trait `UnwindSafe` is not implemented for `(dyn salsa::accumulator::accumulated::AnyAccumulated + 'static)`
     = note: required for `std::ptr::Unique<(dyn salsa::accumulator::accumulated::AnyAccumulated + 'static)>` to implement `UnwindSafe`
note: required because it appears within the type `std::boxed::Box<(dyn salsa::accumulator::accumulated::AnyAccumulated + 'static)>`
    --> /Users/dbarsky/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/alloc/src/boxed.rs:231:12
     |
231  | pub struct Box<
     |            ^^^
     = note: required for `HashMap<IngredientIndex, std::boxed::Box<(dyn salsa::accumulator::accumulated::AnyAccumulated + 'static)>, FxBuildHasher>` to implement `UnwindSafe`
note: required because it appears within the type `salsa::accumulator::accumulated_map::AccumulatedMap`
    --> /Users/dbarsky/.cargo/git/checkouts/salsa-deab236162935b01/592552b/src/accumulator/accumulated_map.rs:10:12
     |
10   | pub struct AccumulatedMap {
     |            ^^^^^^^^^^^^^^
note: required because it appears within the type `salsa::active_query::ActiveQuery`
    --> /Users/dbarsky/.cargo/git/checkouts/salsa-deab236162935b01/592552b/src/active_query.rs:17:19
     |
17   | pub(crate) struct ActiveQuery {
     |                   ^^^^^^^^^^^
note: required because it appears within the type `std::marker::PhantomData<salsa::active_query::ActiveQuery>`
    --> /Users/dbarsky/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/marker.rs:757:12
     |
757  | pub struct PhantomData<T: ?Sized>;
     |            ^^^^^^^^^^^
note: required because it appears within the type `alloc::raw_vec::RawVec<salsa::active_query::ActiveQuery>`
    --> /Users/dbarsky/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/alloc/src/raw_vec.rs:70:19
     |
70   | pub(crate) struct RawVec<T, A: Allocator = Global> {
     |                   ^^^^^^
note: required because it appears within the type `Vec<salsa::active_query::ActiveQuery>`
    --> /Users/dbarsky/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:397:12
     |
397  | pub struct Vec<T, #[unstable(feature = "allocator_api", issue = "32838")] A: Allocator = Global> {
     |            ^^^
note: required because it appears within the type `std::cell::UnsafeCell<Vec<salsa::active_query::ActiveQuery>>`
    --> /Users/dbarsky/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/cell.rs:2076:12
     |
2076 | pub struct UnsafeCell<T: ?Sized> {
     |            ^^^^^^^^^^
note: required because it appears within the type `RefCell<Vec<salsa::active_query::ActiveQuery>>`
    --> /Users/dbarsky/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/cell.rs:730:12
     |
730  | pub struct RefCell<T: ?Sized> {
     |            ^^^^^^^
note: required because it appears within the type `ZalsaLocal`
    --> /Users/dbarsky/.cargo/git/checkouts/salsa-deab236162935b01/592552b/src/zalsa_local.rs:30:12
     |
30   | pub struct ZalsaLocal {
     |            ^^^^^^^^^^
note: required because it appears within the type `Storage<RootDatabase>`
    --> /Users/dbarsky/.cargo/git/checkouts/salsa-deab236162935b01/592552b/src/storage.rs:25:12
     |
25   | pub struct Storage<Db: Database> {
     |            ^^^^^^^
note: required because it appears within the type `std::mem::ManuallyDrop<Storage<RootDatabase>>`
    --> /Users/dbarsky/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/mem/manually_drop.rs:157:12
     |
157  | pub struct ManuallyDrop<T: ?Sized> {
     |            ^^^^^^^^^^^^
note: required because it appears within the type `RootDatabase`
    --> /Users/dbarsky/Developer/rust-analyzer/crates/ide-db/src/lib.rs:79:12
     |
79   | pub struct RootDatabase {
     |            ^^^^^^^^^^^^
note: required because it appears within the type `Analysis`
    --> /Users/dbarsky/Developer/rust-analyzer/crates/ide/src/lib.rs:219:12
     |
219  | pub struct Analysis {
     |            ^^^^^^^^
note: required because it's used within this closure
    --> crates/rust-analyzer/src/cli/rustc_tests.rs:163:60
     |
163  |                         let res = std::panic::catch_unwind(move || {
     |                                                            ^^^^^^^
note: required by a bound in `std::panic::catch_unwind`
    --> /Users/dbarsky/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/std/src/panic.rs:357:40
     |
357  | pub fn catch_unwind<F: FnOnce() -> R + UnwindSafe, R>(f: F) -> Result<R> {
     |                                        ^^^^^^^^^^ required by this bound in `catch_unwind`
```
</details>


I think that https://github.com/rust-lang/rust-analyzer/pull/19093/ is also slightly problematic from a "get rust-analyzer to compile" standpoint, but I don't know enough about unwind safety to say whether I should wrap analysis— and therefore, accumulators—in a `AssertUnwindSafe` as the [documentation points out](https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html#when-should-unwindsafe-be-used). Put differently, if rust-analyzer was to make more pervasive use of accumulators, would we risk getting incorrect results, or...?

Veykril:
That's fine, the problematic type only appears within `ActiveQuery`, the query stack of the runtime which is always empty at this point. So `AssertUnwindSafe` is okay to be used here

davidbarsky:
Alright, few updates:

## Debug Printing

I put up https://github.com/salsa-rs/salsa/pull/676, which makes it possible for us to access the underlying data _without_ involving a `salsa::Database`'s storage (which may not be accessible!). I've used this to port over `dump_syntax_contexts` in `crates/hir-expand/src/hygiene.rs` locally, but I haven't pushed yet—I will, though.

Unfortunately, the approach is not applicable to debug printing in `crates/ide/src/status.rs`, as we'd need to access a tracked function's memos and it's not immediately obvious to me how to do that. Given that it's a single-digit number of queries[^1], I think I'd prefer to accept a regression in the quality of debug output and prioritize porting those queries over to idiomatic Salsa (c.f., https://github.com/rust-lang/rust-analyzer/issues/19098), especially since we've blocked landing any  database changes until this PR lands.

## Enums

I've _also_ opened https://github.com/salsa-rs/salsa/pull/677, which _mostly_ just publishes @ChayimFriedman2's `#[salsa::enum]`/`#[salsa::supertype]` work.

[^1]: The queries in question are `SourceDatabase::file_text`, `RootQueryDb::parse`, `ExpandDatabase::parse_macro_expansion`, `SymbolsDatabase::module_symbols`, `SymbolsDatabase::library_symbols`, `ExpandDatabase::ast_id_map`, and `DefDatabase::block_def_map`.

crates/span/src/hygiene.rs
```diff
@@ -21,44 +21,295 @@
...
+        ) -> Self
+        where
+            Db: ?Sized + salsa::Database,
+            Option<MacroCallId>: zalsa_::interned::HashEqLike<T0>,
+            Transparency: zalsa_::interned::HashEqLike<T1>,
+            Edition: zalsa_::interned::HashEqLike<T2>,
+            SyntaxContext: zalsa_::interned::HashEqLike<T3>,
+        {
+            // FIXME: do we need this?
+            let current_revision = zalsa_::current_revision(db);
```
Veykril, line 184: No,  the macro generates that but it doesn't seem to be needed

Veykril (commented):
Another partial review, still need to go through the ide crates and the main rust-analyzer crate

Veykril:
Also feel free to squash the history here to make your rebase life easier, I don't really care for my commit(s) in there

crates/ide/src/inlay_hints/bind_pat.rs
```diff
@@ -818,6 +818,7 @@ fn fallible() -> ControlFlow<()> {
     }
 
     #[test]
+    #[ignore]
```
Veykril, line 821:
```suggestion
    #[ignore] // FIXME(salsa-transition): bring this back
```

crates/ide/src/lib.rs
```diff
@@ -276,13 +276,13 @@ impl Analysis {
     }
 
     pub fn source_root_id(&self, file_id: FileId) -> Cancellable<SourceRootId> {
-        self.with_db(|db| db.file_source_root(file_id))
+        self.with_db(|db| db.file_source_root(file_id).source_root_id(db))
     }
 
-    pub fn is_local_source_root(&self, source_root_id: SourceRootId) -> Cancellable<bool> {
+    pub fn is_local_source_root(&self, file_id: FileId) -> Cancellable<bool> {
```
Veykril, line 282: Odd change

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -492,6 +492,7 @@ pub(crate) fn handle_document_diagnostics(
     if !snap.analysis.is_local_source_root(source_root)? {
         return Ok(empty_diagnostic_report());
     }
+    let source_root = snap.analysis.source_root_id(file_id)?;
```
Veykril, line 495:
```suggestion
```

Veykril (commented):
Okay I've done a full pass now

crates/hir-ty/src/tests/incremental.rs
```diff
@@ -103,6 +103,7 @@ fn baz() -> i32 {
                 }
             });
         });
-        assert!(format!("{events:?}").matches("infer").count() == 1, "{events:#?}")
+        // todo: figire this out
```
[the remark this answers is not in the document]
davidbarsky: oh, this was slightly broken at the start, but i've fixed it since I started to actually mutate inputs. i should get rid of this comment.

davidbarsky:
Alrighty, I rebased yesterday. Here's my plan for today:
1. I'll go address to Lukas' comments.
2. I'll squash down/absorb most of the commits after the one titled `fix: ensure that all SourceDatabase implementations mutate inputs`. If it's too annoying to absorb properly, I'll squash this down into a single commit.
3. Later today, I'll put up a PR vendoring `db-ext-macro` into rust-analyzer and rebase this PR atop of it.

crates/base-db/src/lib.rs
davidbarsky: wait, sorry: why do you think this change is necessary?

Veykril: Isn't `FileId` the key for `FileText`? `EditionedFileId` should point to its edition and its actual file id, not the file text in my eyes. This definition here seems wrong to me.

davidbarsky: I'll try that now, but if I recall correctly, there were a few cases where the call patterns made `FileText` preferable. However, I'd prefer to be wrong!

davidbarsky: Well, you're right: `vfs::FileId` was sufficient here. This change saves bunch of database lookups!

crates/profile/src/lib.rs
davidbarsky: was accidental, fixed.

crates/rust-analyzer/src/cli/rustc_tests.rs
davidbarsky: Not sure what you're referring to, but I think the only changes here now is wrapping the database in an `AssertUnwindSafe`.

crates/hir-def/Cargo.toml
```diff
@@ -45,6 +47,7 @@ cfg.workspace = true
 tt.workspace = true
 limit.workspace = true
 span.workspace = true
+vfs.workspace = true
```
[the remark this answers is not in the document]
davidbarsky: Generally speaking, I agree, but we need `vfs` to be a dev dep of both `hir-def` and `hir-ty` in order to implement  `base_db::SourceDatabase` in their respective tests: the `file_text` methods take a `vfs::FileId` and we can't rely on a default implementation like we could in old Salsa. Are you doing to treat this as a blocker, or...?

crates/rust-analyzer/src/handlers/request.rs
davidbarsky: wait, why this change? we need the source root in order to to call `Config::diagnostics`. Do you want me to inline this, or...?

crates/base-db/src/input.rs
davidbarsky: i have honestly no idea. i'll do a big format before landing this change; I assume the formatting changes came from resolving conflicts when rebasing.

crates/hir-def/Cargo.toml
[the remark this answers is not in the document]
Veykril: Shouldn't we be re-exporting `FileId` from base-db already? We have that as a dependency so that should suffice

crates/rust-analyzer/src/handlers/request.rs
davidbarsky: oh, no worries.

crates/hir-def/Cargo.toml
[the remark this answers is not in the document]
davidbarsky: oh, you're right. using it as `base_db::FileId` is sufficient.

crates/ide/src/inlay_hints/bind_pat.rs
davidbarsky: decided to unignore this and accept the weird IDs for now. they seem mostly stable?

crates/hir-expand/src/hygiene.rs
```diff
@@ -219,57 +202,57 @@ pub fn marks_rev(
 }
 
 pub(crate) fn dump_syntax_contexts(db: &dyn ExpandDatabase) -> String {
```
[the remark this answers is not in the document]
davidbarsky: well, this thingy works, so i won't remove it for now.

crates/hir-def/src/nameres.rs
```diff
@@ -295,9 +295,13 @@ impl ModuleOrigin {
...
         match self {
-            &ModuleOrigin::File { definition, .. } | &ModuleOrigin::CrateRoot { definition } => {
+            &ModuleOrigin::File { definition: editioned_file_id, .. }
+            | &ModuleOrigin::CrateRoot { definition: editioned_file_id } => {
+                let (file_id, _) = editioned_file_id.unpack();
+                let definition = base_db::EditionedFileId::new(db, file_id, editioned_file_id);
+
                 let sf = db.parse(definition).tree();
-                InFile::new(definition.into(), ModuleSource::SourceFile(sf))
+                InFile::new(editioned_file_id.into(), ModuleSource::SourceFile(sf))
```
Veykril, line 304: This is weird, why are we unpacking it to just repack it?

crates/hir-def/src/test_db.rs
```diff
@@ -203,10 +267,14 @@ impl TestDB {
...
         let def_with_body = fn_def?.into();
-        let (_, source_map) = self.body_with_source_map(def_with_body);
-        let scopes = self.expr_scopes(def_with_body);
-        let root = self.parse(position.file_id);
+        let source_map = db.body_with_source_map(def_with_body).1;
+        let scopes = db.expr_scopes(def_with_body);
+
+        let (file_id, _) = position.file_id.unpack();
+        let editioned_file_id_wrapper =
+            base_db::EditionedFileId::new(db.as_dyn_database(), file_id, position.file_id);
```
Veykril, line 275: same here

crates/hir-expand/src/builtin/fn_macro.rs
```diff
@@ -725,16 +728,20 @@ fn include_expand(
     tt: &tt::TopSubtree,
     span: Span,
 ) -> ExpandResult<tt::TopSubtree> {
-    let file_id = match include_input_to_file_id(db, arg_id, tt) {
-        Ok(it) => it,
+    let (file_id, editioned_file_id) = match include_input_to_file_id(db, arg_id, tt) {
+        Ok(editioned_file_id) => {
+            let (file_id, _) = editioned_file_id.unpack();
+            let file_id = base_db::EditionedFileId::new(db, file_id, editioned_file_id);
+            (file_id, editioned_file_id)
```
Veykril, line 735: and here

crates/hir-expand/src/db.rs
```diff
@@ -322,9 +349,16 @@ fn ast_id_map(db: &dyn ExpandDatabase, file_id: span::HirFileId) -> triomphe::Ar
...
 }
 
+/// Main public API -- parses a hir file, not caring whether it's a real
+/// file or a macro expansion.
 fn parse_or_expand(db: &dyn ExpandDatabase, file_id: HirFileId) -> SyntaxNode {
     match file_id.repr() {
-        HirFileIdRepr::FileId(file_id) => db.parse(file_id).syntax_node(),
+        HirFileIdRepr::FileId(editioned_file_id) => {
+            let (file_id, _) = editioned_file_id.unpack();
+            let file_id = base_db::EditionedFileId::new(db, file_id, editioned_file_id);
```
Veykril, line 358: and here

crates/hir-expand/src/files.rs
```diff
@@ -284,12 +287,15 @@ impl<SN: Borrow<SyntaxNode>> InFile<SN> {
...
 
-        let FileRange { file_id, range } = map_node_range_up_rooted(
+        let FileRange { file_id: editioned_file_id, range } = map_node_range_up_rooted(
             db,
             &db.expansion_span_map(file_id),
             self.value.borrow().text_range(),
         )?;
 
+        let (file_id, _) = editioned_file_id.unpack();
+        let file_id = base_db::EditionedFileId::new(db, file_id, editioned_file_id);
```
Veykril, line 297: and here

crates/hir-expand/src/files.rs
```diff
@@ -453,24 +459,27 @@ impl<N: AstNode> InFile<N> {
...
 
-        let FileRange { file_id, range } = map_node_range_up_rooted(
+        let FileRange { file_id: editioned_file_id, range } = map_node_range_up_rooted(
             db,
             &db.expansion_span_map(file_id),
             self.value.syntax().text_range(),
         )?;
 
+        let (file_id, _) = editioned_file_id.unpack();
+        let file_id = base_db::EditionedFileId::new(db, file_id, editioned_file_id);
```
Veykril, line 477: and here

crates/hir-expand/src/span_map.rs
```diff
@@ -79,10 +79,17 @@ impl SpanMapRef<'_> {
...
+    db: &dyn ExpandDatabase,
+    editioned_file_id: EditionedFileId,
+) -> Arc<RealSpanMap> {
     use syntax::ast::HasModuleItem;
     let mut pairs = vec![(syntax::TextSize::new(0), span::ROOT_ERASED_FILE_AST_ID)];
-    let ast_id_map = db.ast_id_map(file_id.into());
+    let ast_id_map = db.ast_id_map(editioned_file_id.into());
+
+    let (file_id, _) = editioned_file_id.unpack();
+    let file_id = base_db::EditionedFileId::new(db, file_id, editioned_file_id);
```
Veykril, line 91: and here

crates/hir/src/semantics.rs
```diff
@@ -1931,7 +1936,10 @@ fn macro_call_to_macro_id(
     match loc.def.ast_id() {
         Either::Left(it) => {
             let node = match it.file_id.repr() {
-                HirFileIdRepr::FileId(file_id) => {
+                HirFileIdRepr::FileId(editioned_file_id) => {
+                    let (file_id, _) = editioned_file_id.unpack();
+                    let file_id = base_db::EditionedFileId::new(db, file_id, editioned_file_id);
```
Veykril, line 1941: You get the idea (I'll stop now, there is more of these than I expected 😬)

Veykril (commented):
Judging by the ide test failure, the weird editioned file handling is likely the cause (we are scrambling the editions around somewhere)

crates/base-db/src/lib.rs
```diff
@@ -47,39 +50,202 @@ pub const DEFAULT_FILE_TEXT_LRU_CAP: u16 = 16;
...
+                vacant.insert(file_source_root);
+            }
+        };
+    }
+}
+
+#[salsa::interned(no_lifetime)]
+pub struct EditionedFileId {
+    pub file_id: vfs::FileId,
+    pub editioned_file_id: span::EditionedFileId,
```
Veykril, line 161: Oh ye that would do it, this definition is recursive :) That should be just an edition

Veykril: Hm okay i see, thats because the source database is already salsafied so we need a different input here. That is quite annoying. Either way the definition of this should be
```rust
pub struct EditionedFileId {
    pub file_id: vfs::FileId,
    pub edition: span::Edition
}
```
or

```rust
pub struct EditionedFileId {
    pub file_id: span::EditionedFileId,
}
```

Right now it encodes the file id twice resulting in very weird transformation code. We also should introduce a help func that transforms between the span and base-db versions

crates/base-db/src/lib.rs
```diff
@@ -47,39 +50,202 @@ pub const DEFAULT_FILE_TEXT_LRU_CAP: u16 = 16;
...
+}
+
+#[salsa::input]
+pub struct FileText {
+    pub text: Arc<str>,
+    pub file_id: vfs::FileId,
```
Veykril, line 167: This field is unnecessary (and unused, i checked), file text should be the text, not the text and its key

davidbarsky:
> Judging by the ide test failure, the weird editioned file handling is likely the cause (we are scrambling the editions around somewhere)

the keyword highlighting one? Yeah, good point. Will dig around, must be something kinda inadvertently silly. 

I'll fix the weird construction/deconstruction of `EditionedFileId`, but I'm doing is that it's the detritus left from repeated refactors + bit of brain fog. Thanks for catching it.

crates/base-db/src/lib.rs
davidbarsky: I think I'd prefer

```rust
pub struct EditionedFileId {
    pub file_id: span::EditionedFileId,
}
```

...because I'm pretty sure `span::EditionedFileId` is used in crates without a Salsa dep: I remember this was the case about 3 months ago and stumbling over it. Would rather just have the ID here than reconstruct it each time.

I'll add some methods which make deconstructing the value nicer.

davidbarsky: done.

davidbarsky:
Updates:
1. Fixed the weird File ID refactoring detritus stuff that Lukas pointed out.
2. The cross-compilation tests are failing on `powerpc-unknown-linux-gnu`, as new Salsa is directly using a `AtomicU64` to reduce memory usage. We'll probably `cfg` out the 64-bit specific optimizations on 32-bit targets, as discussed on the [Salsa Zulip](https://salsa.zulipchat.com/#narrow/channel/145099-Using-Salsa/topic/Use.20of.20AtomicU64.20in.20Salsa), but if it's too annoying, we'll use https://github.com/taiki-e/portable-atomic in Salsa instead.
3. The syntax highlighting test failures (`syntax_highlighting::tests::{test_keyword_highlighting, test_keyword_macro_edition_highlighting, test_rainbow_highlighting}`) are failing. I've just pushed a commit that include the "actual" changes suffixed by `wat`; I'm pretty sure it's related to me messing up default editions _somewhere_, but maybe based on the nature of the failures, the cause will be blindingly obvious to _someone_.
4. This is no longer marked as a draft!

We'll still need to:
1. Vendor https://github.com/davidbarsky/db-ext-macro/ into rust-analyzer. I'll _actually_ do that today.
2. Cut a `Salsa 1.0-alpha-1` release to unblock @carljm from landing https://github.com/salsa-rs/salsa/pull/603.

crates/hir-def/src/macro_expansion_tests/mod.rs
```diff
@@ -63,9 +64,13 @@ fn check_errors(#[rust_analyzer::rust_fixture] ra_fixture: &str, expect: Expect)
                 MacroCallKind::Derive { ast_id, .. } => ast_id.map(|it| it.erase()),
                 MacroCallKind::Attr { ast_id, .. } => ast_id.map(|it| it.erase()),
             };
-            let ast = db
-                .parse(ast_id.file_id.file_id().expect("macros inside macros are not supported"))
-                .syntax_node();
+
+            let editioned_file_id =
+                ast_id.file_id.file_id().expect("unable to get FileId; this is a bug");
```
Veykril, line 69:
```suggestion
                ast_id.file_id.file_id().expect("macros inside macros are not supported");
```

crates/span/src/hygiene.rs
```diff
@@ -21,44 +21,288 @@
...
+        }
+
+        pub fn edition<Db>(self, db: &'db Db) -> Edition
+        where
+            Db: ?Sized + zalsa_::Database,
+        {
+            // FIXME: is this right?
+            if self.is_root() {
+                return Edition::LATEST;
+            }
```
Veykril, line 239: This might be the cause for the syntax highlighting test failures

Veykril:
```suggestion
            if self.is_root() {
                return Edition::from_u32(self.0.as_u32() - (SyntaxContext::MAX_ID - Edition::LATEST as u32));
            }
```
Note `Edition::from_u32` isn't a thing, gotta create that as well

davidbarsky: Good catch! Applying both makes `test_keyword_macro_edition_highlighting` pass, but `test_keyword_highlighting` and `test_rainbow_highlighting` still fail. Not sure why, but working on vendoring the query group macro (and updating it to the latest Salsa...).

Veykril: rainbow i think isnt a problem, we likely are hashing different data now, as long as the hashes are stable and equal as before (wrt to positions they appear in). The other keyword one is still edition related though hmm

Veykril: Actually it is weird that the hashes change, they shouldn't nothing there goes through salsa ...

Veykril:
`handlers::convert_bool_to_enum::tests::field_enum_cross_file` might be due to `Ord` changes, `render::tests::score_fn_type_and_name_match` definitely due to `Ord` changes. Neither are problematic right now

davidbarsky:
okay! I will squash and add this to the merge queue (!)

editors/code/package.json
Veykril, line 1: This should not have changed

docs/book/src/configuration_generated.md
Veykril, line 1: likewise

editors/code/package.json
davidbarsky: whoops, undoing

Veykril (approved):

lnicola:
I'm a bit late to the party, but did y'all see this?

image

Veykril:
Yes it is expected, we aren't triggering the LRU in analysis-stats anymore.  New salsa only triggers it when a new revision bumps which doesn't happen in analysis-stats anymore

lnicola:
Hmm, can we do that to keep the metrics comparable with the previous ones?

Veykril:
We can trigger it at the end manually yes `https://docs.rs/salsa/latest/salsa/trait.Database.html#method.trigger_lru_eviction`

PR #18967, opened 2025-01-18
Properly record meaningful imports as re-exports in symbol index

Veykril:
I somehow didn't realize that https://github.com/rust-lang/rust-analyzer/pull/18699 never fixed what it was supposed to fix. Probably because I accidentally expect updated the test.

The gist of this PR is that we now record all imports that have an explicit visibility set (or that exist in the crate-root), treating them as if they are re-exports (opposed to only treating rename imports as such). This in turn fixes the completion issue as the relevant import is now recorded.

I do fear this might have an impact on flyimport completion times as we have a lot more symbols to go through now, will investigate that before merging.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -523,6 +523,7 @@ dependencies = [
  "hir-def",
  "hir-expand",
  "hir-ty",
+ "indexmap",
  "intern",
  "itertools",
  "rustc-hash 2.0.0",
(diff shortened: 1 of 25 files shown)


Veykril:
A bit noisy of a commit history as there were a couple of bugs I encountered.

Veykril:
This seems to have some unintended side effects, symbol search now hangs for me in the salsa repo 😬

PR #18973, opened 2025-01-18
Make niches into nices

ThouCheese:
Format niche information like so:
image

### DIFF
--- a/crates/ide/src/hover/render.rs
+++ b/crates/ide/src/hover/render.rs
@@ -1082,7 +1082,19 @@ fn render_memory_layout(
 
     if config.niches {
         if let Some(niches) = layout.niches() {
-            format_to!(label, "niches = {niches}, ");
+            if niches > 1024 {
+                if is_pwr2(niches) {
+                    format_to!(label, "niches = 2{}, ", pwr2_to_exponent(niches));
+                } else if is_pwr2plus1(niches) {
+                    format_to!(label, "niches = 2{} + 1, ", pwr2_to_exponent(niches - 1));
+                } else if is_pwr2minus1(niches) {
+                    format_to!(label, "niches = 2{} - 1, ", pwr2_to_exponent(niches + 1));
+                } else {
+                    format_to!(label, "niches = really rather quite large, ");
+                }
+            } else {
+                format_to!(label, "niches = {niches}, ");
+            }
         }
     }
     label.pop(); // ' '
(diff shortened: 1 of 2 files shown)


crates/ide/src/hover/render.rs
```diff
@@ -1210,3 +1222,85 @@ fn render_dyn_compatibility(
...
+    val == u128::MAX || (val + 1).count_ones() == 1
+}
+
+fn is_pwr2plus1(val: u128) -> bool {
+    val != 0 && (val - 1).count_ones() == 1
+}
+
+fn pwr2_to_exponent(num: u128) -> String {
+    const DIGITS: [char; 10] = ['⁰', '¹', '²', '³', '⁴', '⁵', '⁶', '⁷', '⁸', '⁹'];
+    (127 - num.leading_zeros())
```
lnicola, line 1240: Wouldn't `num.trailing_zeros()` work here?

crates/ide/src/hover/render.rs
```diff
@@ -1082,7 +1082,19 @@ fn render_memory_layout(
...
-            format_to!(label, "niches = {niches}, ");
+            if niches > 1024 {
+                if is_pwr2(niches) {
+                    format_to!(label, "niches = 2{}, ", pwr2_to_exponent(niches));
+                } else if is_pwr2plus1(niches) {
+                    format_to!(label, "niches = 2{} + 1, ", pwr2_to_exponent(niches - 1));
+                } else if is_pwr2minus1(niches) {
+                    format_to!(label, "niches = 2{} - 1, ", pwr2_to_exponent(niches + 1));
+                } else {
+                    format_to!(label, "niches = really rather quite large, ");
```
lnicola, line 1093: :bike::house: This seems a bit excessive, so maybe


```suggestion
                    format_to!(label, "niches = rather many, ");
```

or something like that?

crates/ide/src/hover/render.rs
```diff
@@ -1210,3 +1222,85 @@ fn render_dyn_compatibility(
         }
     }
 }
+
+fn is_pwr2(val: u128) -> bool {
```
lnicola, line 1226: Let's inline (remove) this.

crates/ide/src/hover/render.rs
lnicola: There's a bunch of ways around that, like restructuring the code by splitting off "number to superscript digits" from "checking if it's a power of 2, more or less" so it's clearly correct, or by using an `assert!`, but if we have a bug here, it's better to make it more obvious anyway.

lnicola (approved):
Never mind, we can fix this at any time.

crates/ide/src/hover/render.rs
ThouCheese: Can do!

ThouCheese: Sorted in https://github.com/rust-lang/rust-analyzer/pull/18973/commits/705eb84941c2d2eb3daa741fb6a99216572737f4

crates/ide/src/hover/render.rs
ThouCheese: Sorted in https://github.com/rust-lang/rust-analyzer/pull/18973/commits/5ad3de8b29840c5975ae04b4d0d5f5d85dafc179

Veykril (approved):
Thanks!

lnicola:
image

Will this work, I wonder?

PR #18994, opened 2025-01-21
internal: Treat cfg fetching failures as a warning

Wilfred:
If the user doesn't have rustc on $PATH, rust-analyzer won't be able to run `rustc --print cfg`. This isn't really an error, as rust-analyzer can still proceed without it.

This is particularly noticeable when loading crates defined in a rust-project.json. Until the configuration is loaded, the opened files are briefly treated as detached files and users see this error.

Environments with rust-project.json generally have a sysroot and rustc elsewhere, so the error confuses users.


davidbarsky (approved):
Approved, but will ask someone else to merge.

PR #18995, opened 2025-01-21
fix: Lower range pattern bounds to expressions

### LINKED ISSUE #12210: Lower range patterns from AST to HIR
Hm, to me it seems correct to use `pat..pat` in the syntax, but `expr..expr` in the hir.

Syntactically, they *are* patterns, as `..` is an infix binary operator. We *could* add some LR-style re-structuring of the tree, where we don't create a pat until we haven't seen `..`, but that seems complicated. Feels more natural to parse a pat, than notice an `..`, and parse another `pat`. 

But yeah, in semantics we want to keep those as expressions, so during *lowering* we need to figure out that, what was a pattern syntactically, actually is an expression semantically. 

This feel dual to recently stabilized destructive assignment, where we parse stuff like `(a, b) = (b, a)` as `expr = expr`, but, during lowering, lower the LHS as a pattern. 

Though, seeing the two example side-by-side, maybe we should bite the bullet and just *don't* distinguish patterns and expressions at the level of syntax? That is, we'd use `ast::Expr` for both patterns and expressions. We'd then have an API like

```rust
impl ast::Expr {
    fn classify(&self) -> IsPattern | IsExpression | Ambiguous
}

impl Semnatics {
    fn classify_expr(&self, expr: &ast::Expr) -> IsPattern | IsExpression
}
```

_Originally posted by @matklad in https://github.com/rust-lang/rust-analyzer/pull/12158#discussion_r867470092_

alibektas:
fixes #12210

### DIFF
--- a/crates/hir/src/diagnostics.rs
+++ b/crates/hir/src/diagnostics.rs
@@ -6,6 +6,7 @@
 use cfg::{CfgExpr, CfgOptions};
 use either::Either;
 use hir_def::{
+    body::ExprOrPatPtr,
     hir::ExprOrPatId,
     path::{hir_segment_to_ast_segment, ModPath},
     type_ref::TypesSourceMap,
@@ -115,14 +116,14 @@ diagnostics![
 
 #[derive(Debug)]
 pub struct BreakOutsideOfLoop {
-    pub expr: InFile<AstPtr<ast::Expr>>,
+    pub expr: InFile<AstPtr<Either<ast::Expr, ast::Pat>>>,
     pub is_break: bool,
     pub bad_value_break: bool,
 }
 
 #[derive(Debug)]
 pub struct TypedHole {
-    pub expr: InFile<AstPtr<ast::Expr>>,
+    pub expr: InFile<AstPtr<Either<ast::Expr, ast::Pat>>>,
     pub expected: Type,
 }
 
@@ -234,13 +235,13 @@ pub struct MismatchedTupleStructPatArgCount {
 
 #[derive(Debug)]
 pub struct ExpectedFunction {
-    pub call: InFile<AstPtr<ast::Expr>>,
+    pub call: InFile<AstPtr<Either<ast::Expr, ast::Pat>>>,
     pub found: Type,
 }
 
 #[derive(Debug)]
 pub struct UnresolvedField {
-    pub expr: InFile<AstPtr<ast::Expr>>,
+    pub expr: InFile<AstPtr<Either<ast::Expr, ast::Pat>>>,
     pub receiver: Type,
     pub name: Name,
     pub method_with_same_name_exists: bool,
@@ -248,7 +249,7 @@ pub struct UnresolvedField {
 
 #[derive(Debug)]
 pub struct UnresolvedMethodCall {
-    pub expr: InFile<AstPtr<ast::Expr>>,
+    pub expr: InFile<AstPtr<Either<ast::Expr, ast::Pat>>>,
     pub receiver: Type,
     pub name: Name,
     pub field_with_same_name: Option<Type>,
@@ -267,7 +268,7 @@ pub struct UnresolvedIdent {
 
 #[derive(Debug)]
 pub struct PrivateField {
-    pub expr: InFile<AstPtr<ast::Expr>>,
+    pub expr: InFile<AstPtr<Either<ast::Expr, ast::Pat>>>,
     pub field: Field,
 }
 
@@ -302,7 +303,7 @@ pub struct ReplaceFilterMapNextWithFindMap {
 
 #[derive(Debug)]
 pub struct MismatchedArgCount {
-    pub call_expr: InFile<AstPtr<ast::Expr>>,
+    pub call_expr: InFile<ExprOrPatPtr>,
     pub expected: usize,
     pub found: usize,
 }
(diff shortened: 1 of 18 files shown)


crates/hir-def/src/body/lower.rs
```diff
@@ -1716,23 +1716,39 @@ impl ExprCollector<'_> {
...
+                                        });
+                                }
+
+                                None
+                            }
+                            ast::Pat::PathPat(p) => p
+                                .path()
+                                .and_then(|path| self.parse_path(path))
+                                .map(|parsed| self.alloc_expr_from_pat(Expr::Path(parsed), ptr)),
+                            _ => None,
```
Veykril, line 1750: Let's maybe add a comment here (or on top of the match) regarding the unhandled variants.
> We only need to handle literal, ident (if bare) and path patterns here, as any other pattern as a range pattern operand is semantically invalid.

crates/hir-def/src/body/lower.rs
```diff
@@ -1716,23 +1716,39 @@ impl ExprCollector<'_> {
...
+                let mut range_part_lower = |p: Option<ast::Pat>| -> Option<ExprId> {
+                    p.and_then(|it| {
+                        let ptr = PatPtr::new(&it);
+                        match &it {
+                            ast::Pat::LiteralPat(it) => Some(self.alloc_expr_from_pat(
+                                Expr::Literal(pat_literal_to_hir(it)?.0),
+                                ptr,
+                            )),
+                            ast::Pat::IdentPat(ident) => {
+                                if ident.is_simple_ident() {
```
Veykril, line 1734:
```suggestion
                            // treat a bare ident as a path
                            ast::Pat::IdentPat(ident) if ident.is_simple_ident() => {
```

Then we could drop the `return`

crates/hir/src/diagnostics.rs
```diff
@@ -115,14 +116,14 @@ diagnostics![
 
 #[derive(Debug)]
 pub struct BreakOutsideOfLoop {
-    pub expr: InFile<AstPtr<ast::Expr>>,
+    pub expr: InFile<AstPtr<Either<ast::Expr, ast::Pat>>>,
```
Veykril, line 119:
```suggestion
    pub expr: InFile<ExprOrPatPtr>,
```
should work right? (and elsewhere in the file)

crates/hir/src/source_analyzer.rs
```diff
@@ -212,8 +212,10 @@ impl SourceAnalyzer {
     ) -> Option<(Type, Option<Type>)> {
         let pat_id = self.pat_id(pat)?;
         let infer = self.infer.as_ref()?;
-        let coerced =
-            infer.pat_adjustments.get(&pat_id).and_then(|adjusts| adjusts.last().cloned());
+        let coerced = infer
+            .pat_adjustments
+            .get(&pat_id.as_pat()?)
```
Veykril, line 217: This might need to handle the expression case (and lookup in the expression adjustments etc) but I am not entirely sure. Otherwise trying to lookup the type of `FOO` in pattern `FOO..` will fail right? Either way would need to confirm whether thats relevant or not (by checking some IDE feature that makes use of this `type_of_pat` here).

crates/hir/src/source_analyzer.rs
```diff
@@ -684,10 +686,19 @@ impl SourceAnalyzer {
     ) -> Option<ModuleDef> {
         let pat_id = self.pat_id(&pat.clone().into())?;
         let body = self.body()?;
-        let path = match &body[pat_id] {
-            Pat::Path(path) => path,
-            _ => return None,
+
+        let path = if pat_id.is_pat() {
```
Veykril, line 690: This can be a match on `pat_id`

crates/hir/src/source_analyzer.rs
```diff
@@ -783,7 +794,7 @@ impl SourceAnalyzer {
                 prefer_value_ns = true;
             } else if let Some(path_pat) = parent().and_then(ast::PathPat::cast) {
                 let pat_id = self.pat_id(&path_pat.into())?;
-                if let Some((assoc, subs)) = infer.assoc_resolutions_for_pat(pat_id) {
+                if let Some((assoc, subs)) = infer.assoc_resolutions_for_pat(pat_id.as_pat()?) {
```
Veykril, line 797: Similar thing here, might need to handle the expression case due to having a path pattern

crates/hir/src/source_analyzer.rs
```diff
@@ -807,7 +818,7 @@ impl SourceAnalyzer {
                     return Some((PathResolution::Def(AssocItem::from(assoc).into()), Some(subst)));
                 }
                 if let Some(VariantId::EnumVariantId(variant)) =
-                    infer.variant_resolution_for_pat(pat_id)
+                    infer.variant_resolution_for_pat(pat_id.as_pat()?)
```
Veykril, line 821: Similar thing here, might need to handle the expression case due to having a path pattern

Veykril:
Last commit message should be more descriptive, https://github.com/rust-lang/rust-analyzer/pull/18995/commits/6f5e24ce286aac6765fc6602be7336ad769e4209 as well

Veykril (approved):

Veykril:
Thanks!

PR #19001, opened 2025-01-22
feat: Implement `default-field-values`

### LINKED ISSUE #18938: default_field_values grammar
Allow struct definitions to provide default values for individual fields and thereby allowing those to be omitted from initializers.

https://rust-lang.github.io/rfcs/3681-default-field-values.html

https://github.com/dtolnay/syn/issues/1774

ShoyuVanilla:
Resolves #18938


crates/hir-def/src/hir.rs
```diff
@@ -255,6 +255,7 @@ pub enum Expr {
         path: Option<Box<Path>>,
         fields: Box<[RecordLitField]>,
         spread: Option<ExprId>,
+        ellipsis: bool,
```
Veykril, line 258: Can we encode this with a ternary enum? Like
```
enum Spread {
    No,
    Yes,
    Base(ExprId)
}
```

Reason being that `ellipsis: false` `spread: Some` is a bogus state.

Name ideas (for variants/enum) welcome

crates/hir-def/src/lib.rs
```diff
@@ -1333,6 +1385,11 @@ impl HasModule for DefWithBodyId {
             DefWithBodyId::ConstId(it) => it.module(db),
             DefWithBodyId::VariantId(it) => it.module(db),
             DefWithBodyId::InTypeConstId(it) => it.lookup(db).owner.module(db),
+            DefWithBodyId::FieldId(it) => match it.parent {
```
Veykril, line 1388: Might be best to implement `HasModule` for `VariantId`?

crates/hir-def/src/resolver.rs
```diff
@@ -1239,6 +1244,11 @@ impl HasResolver for DefWithBodyId {
             DefWithBodyId::StaticId(s) => s.resolver(db),
             DefWithBodyId::VariantId(v) => v.resolver(db),
             DefWithBodyId::InTypeConstId(c) => c.lookup(db).owner.resolver(db),
+            DefWithBodyId::FieldId(f) => match f.parent {
```
Veykril, line 1247: and `HasResolver`

crates/hir-ty/src/infer.rs
```diff
@@ -910,6 +913,19 @@ impl<'a> InferenceContext<'a> {
...
     }
 
+    fn collect_field(&mut self, field: FieldId) {
+        let variant_data = field.parent.variant_data(self.db.upcast());
+        let field_data = &variant_data.fields()[field.local_id];
+        let types_map = variant_data.types_map();
+        let return_ty =
+            self.make_ty(field_data.type_ref, types_map, InferenceTyDiagnosticSource::Signature);
+
+        // Field default value exprs might be defining usage sites of TAITs.
```
Veykril, line 923: Can we have a little test for that?

crates/hir/src/lib.rs
```diff
@@ -415,6 +415,19 @@ impl ModuleDef {
...
+        let fields = match self {
+            ModuleDef::Adt(Adt::Struct(it)) => Some(it.fields(db)),
+            ModuleDef::Adt(Adt::Union(it)) => Some(it.fields(db)),
+            ModuleDef::Variant(it) => Some(it.fields(db)),
+            _ => None,
+        };
+        if let Some(fields) = fields {
+            for field in fields {
+                let def: DefWithBody = field.into();
+                def.diagnostics(db, &mut acc, style_lints);
```
Veykril, line 427: I do worry about the perf impact this PR might have wrt to memory (and revision refreshing due to so many new query entries) as now every field becomes a body, and no one will really be using this feature yet so basically all fields will be default-less. So I'd propose we e ither do not compute diagnostics for these at all for now, OR, give variants a flag for whether they contain any defaulted-fields or not as a small optimization that allows us to skip this if we know that none of their fields have any bodies.

crates/hir-def/src/lib.rs
```diff
@@ -516,6 +516,7 @@ pub type LocalModuleId = Idx<nameres::ModuleData>;
 pub struct FieldId {
     pub parent: VariantId,
     pub local_id: LocalFieldId,
+    pub has_default: bool,
```
Veykril, line 519: I just realized we have t his in `FieldData` already, so we can't we reuse that instead? If we have the ID we can trivially index into the `variant_data` where required.

crates/hir/src/lib.rs
```diff
@@ -1414,7 +1418,7 @@ impl Struct {
             .variant_data
             .fields()
             .iter()
-            .map(|(id, _)| Field { parent: self.into(), id })
+            .map(|(id, d)| Field { parent: self.into(), id, has_default: d.has_default })
```
Veykril, line 1421: That is here we can fetch it via `variant_data[id]` since we already have ` db.struct_data(self.id).variant_data` anyways

Veykril (approved):
One small thing but otherwise lgtm!

ShoyuVanilla:
I'm rebasing to resolve conflicts 😄

Veykril:
Ah, sorry :)

ShoyuVanilla:
Nevermind. They were quite easy to resolve

PR #19005, opened 2025-01-22
fix testing packages with multiple targets

### LINKED ISSUE #18955: VSCode Test Explorer - bugs running tests for packages with multiple targets

duncanawoods:
This PR fixes #18955 and other bugs related to RA executing the wrong test and/or assigning results incorrectly.

It makes test execution aware of the difference between packages and targets.

It removes the need to use `hack_recover_crate_name` to match test output to tests.

- `request::handle_run_test` 
    - no longer tries (and fails) to reduce a list of tests to a single pattern
- `test_runner::CargoTestHandle` 
    - no longer runs patterns against the workspace and now always runs tests in the correct target
- `CargoActor` 
    - is extended to take a parser that has the required context for parsing tool output correctly
- `main_loop`
    - every `CargoTestMessage` in the event queue now contains the required information to assign to correct test


HKalbasi:
Ah sorry I forgot this. Now that the `hack_recover_crate_name` is gone, this actually reduces the amount of workarounds so I'm happy to merge this if you remove the merge commits.

HKalbasi:
If you want to add some end to end tests, look at https://github.com/rust-lang/rust-analyzer/blob/master/crates/rust-analyzer/tests/slow-tests/main.rs which creates a cargo project and communicate with rust-analyzer using lsp. But those tests slowdown CI so I would suggest having at most one test about test-explorer there (which can test many operations, most of the time is spent on the loading) and test other corner cases using unit tests that takes some hard coded json of cargo and/or assert text of the cargo commands instead of executing them.

crates/project-model/src/cargo_workspace.rs
```diff
@@ -228,6 +228,25 @@ pub enum TargetKind {
...
+            f,
+            "{}",
+            match self {
+                TargetKind::Bin => "bin",
+                TargetKind::Lib { is_proc_macro: true } => "proc-macro",
+                TargetKind::Lib { is_proc_macro: false } => "lib",
+                TargetKind::Example => "example",
+                TargetKind::Test => "test",
+                TargetKind::Bench => "bench",
+                TargetKind::BuildScript => "custom-build",
```
Veykril, line 243:
```suggestion
                TargetKind::BuildScript => "build-script",
```

crates/rust-analyzer/src/command.rs
```diff
@@ -14,23 +14,60 @@ use process_wrap::std::{StdChildWrapper, StdCommandWrap};
 use stdx::process::streaming_output;
 
 /// Cargo output is structured as a one JSON per line. This trait abstracts parsing one line of
-/// cargo output into a Rust data type.
+/// cargo output into a Rust data type where the data is self-describing.
 pub(crate) trait ParseFromLine: Sized + Send + 'static {
```
Veykril, line 18: Can we just discard this trait in favor of `CargoParser`? This code is already fairly convoluted and this isn't really helping that

crates/project-model/src/cargo_workspace.rs
duncanawoods: The actual name cargo uses for this target kind is `custom-build` which is what needs to be used in commands and interpreting output.

https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Target.html

I'm using the display trait for formatting cargo arguments (maybe this is wrong?) so this change would require a new function `to_cargo_str` or something. If your your desire is to fix the code-smell of the display not matching the enum, a more correct change might be to change the enum name from `BuildScript` to `CustomBuild`.

(possibly a little moot for this bug fix since I'm not sure cargo test accepts tests in custom-build targets and I haven't tested it)

crates/rust-analyzer/src/command.rs
Veykril: I guess its fine for now

crates/project-model/src/cargo_workspace.rs
duncanawoods: Done. 

Much clearer intention and can now return an option to make it clear that not all RA TargetKinds can be converted to a Cargo TargetKinds.

Veykril:
Looks good to me now, though I don't know anything about the test runner code so thats up to @HKalbasi to decide

duncanawoods:
@HKalbasi good to go?

duncanawoods:
@HKalbasi @Veykril 

Is there something I can do to help close this off?

Should I keep pestering or is it blocked/unwanted? Thanks!

Veykril:
I am not familiar with the text explorer so I leave that up to @HKalbasi for the final call

HKalbasi:
Ah sorry I missed the mentions. I assumed you wanted to add some tests, but it's fine to add it in another PR later. So let's merge this one.

duncanawoods:
Thanks @HKalbasi.

The e2e test I had in mind wasn't suitable for CI because my main uncertainty is the assumptions made about `vscode` and `cargo test` rather than the correctness of the glue in between. For a reasonable exhaustive exercise of all the permutations, it would have to be a manually run test or a "very slow test suite" that doesn't block CI.

In my codebases I would add a folder of test files and a unit test module with an `[#ignore]` attribute so it's excluded from automatic suite runs but can be invoked explicitly with `--include-ignored`.

When I see a bug it would have caught then I know I have to write this!

PR #19019, opened 2025-01-24
Explicitly add buildfiles when constructing ProjectFolders

darichey:
Partially reverts #18668 which changed this such that buildfiles are implicitly included by including their parent directory. This behavior is potentially problematic, because it also includes everything else in that directory. At work, we vendor third-party libraries under a common directory with a single shared buck buildfile. With this approach, the entire third-party directory is included in the PackageRoot.


Veykril (approved):

Veykril:
Thanks and sorry!

PR #19020, opened 2025-01-24
fix: Prevent infinite recursion of bounds formatting

### LINKED ISSUE #19007: [scip] "thread 'main' has overflowed its stack" since nightly-2025-01-21
Steps to reproduce:
- Clone https://hg.mozilla.org/mozilla-central (mercurial) or https://github.com/mozilla/gecko-dev (git)
- Presumably, this would be reproductible with what you get out of the clone, but just in case, checkout 4d679baa3086ad410eb22618e50aea045424c0de (mercurial) or 3ee2e63c1b9cff116ff7bec1189e99fab6e1c15c (git)
- Copy `.cargo/config.toml.in` to `.cargo/config.toml`.
- Run `rust-analyzer scip .`

This is a serious regression in nightly and it's important to fix it before the next release.

Bad: rust-analyzer 1.86.0-nightly (f3d1d47 2025-01-20)
Good: rust-analyzer 1.86.0-nightly (9a1d156 2025-01-19)
... but of course the commits shown there are not from rust-analyzer itself... and it's not a submodule.

I /think/ the good one is 7b368b97bab2384c661dc3a1f3945b1a5c59c969 and the bad one is 93a58467841b3233c15c6a0019edc7f75e2b5337.

ShoyuVanilla:
Fixes #19007

I've inserted extra tracings in scip and found out the exact position causing the issue:

https://docs.rs/futures-util/latest/src/futures_util/stream/try_stream/try_concat.rs.html#40

```rust
impl<St> Future for TryConcat<St>
where
    St: TryStream,
    St::Ok: Extend<<St::Ok as IntoIterator>::Item> + IntoIterator + Default,
{
    type Output = Result<St::Ok, St::Error>;

    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        let mut this = self.project();

        Poll::Ready(Ok(loop {
            if let Some(x) = ready!(this.stream.as_mut().try_poll_next(cx)?) {
                      //^ This token
                if let Some(a) = this.accum {
                    a.extend(x)
                } else {
                    *this.accum = Some(x)
                }
            } else {
                break this.accum.take().unwrap_or_default();
            }
        }))
    }
}
```

the `x` on the above code has the type `St::Ok` and the bounds `St::Ok: Extend<<**St::Ok** as IntoIterator>::Item> + IntoIterator + Default` and this is causing the infinite recursion, [as @glandium said](https://github.com/rust-lang/rust-analyzer/issues/19007#issuecomment-2611119108)


crates/hir-ty/src/display.rs
```diff
@@ -479,42 +527,54 @@ impl HirDisplay for ProjectionTy {
         // `<Param as Trait>::Assoc`
         if !f.display_target.is_source_code() {
             if let TyKind::Placeholder(idx) = self_ty.kind(Interner) {
-                let db = f.db;
-                let id = from_placeholder_idx(db, *idx);
-                let generics = generics(db.upcast(), id.parent);
+                let res = f.format_bounds_with(|f| {
```
Veykril, line 530: I feel like we might wanna move this down into the ` if !bounds.is_empty() {` branch body? Recursive bounds like these will be the minority and so we prevent triggering a heap alloc in the case where we aren't actually doing the different rendering here (but only realize so within this closure.

Veykril (approved):

PR #19023, opened 2025-01-24
minor: Suggest better names when a type is a sequence

Wilfred:
Previously, we'd suggest a type of `vec` for a value of type `Vec<T>`, which is rarely what the user wants. We also had no suggestions for values of type `&[T]`.

Instead, try to pluralise the inner type name, and fall back to `items`.

### DIFF
--- a/crates/ide-db/src/syntax_helpers/suggest_name.rs
+++ b/crates/ide-db/src/syntax_helpers/suggest_name.rs
@@ -31,6 +31,12 @@ const USELESS_NAME_PREFIXES: &[&str] = &["from_", "with_", "into_"];
 /// `Result<User, Error>` -> `User`
 const WRAPPER_TYPES: &[&str] = &["Box", "Arc", "Rc", "Option", "Result"];
 
+/// Generic types replaced by a plural of their first argument.
+///
+/// # Examples
+/// `Vec<Name>` -> "names"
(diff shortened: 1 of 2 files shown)


crates/ide-db/src/syntax_helpers/suggest_name.rs
```diff
@@ -390,12 +401,32 @@ fn name_of_type(ty: &hir::Type, db: &RootDatabase, edition: Edition) -> Option<S
...
+    let items_str = SmolStr::new_static("items");
+    let Some(inner_ty) = inner_ty else {
+        return items_str;
+    };
+    let Some(name) = name_of_type(inner_ty, db, edition) else {
+        return items_str;
+    };
+
+    if name.ends_with(['s', 'x', 'y']) {
+        // Given a type called e.g. "Boss", "Fox" or "Story", don't try to
```
Veykril, line 422: Is english regular enough with regards to plurals that we could be smart here? `s` -> `ses`, `x` -> `xes`, `y` -> `ies`. I don't think it is is it 😔

Veykril (approved):
Thanks!

crates/ide-db/src/syntax_helpers/suggest_name.rs
lnicola: Not under those exact rules, but I think it is.

PR #19029, opened 2025-01-25
internal: Skip redundant path search in `resolve_completion_edits`

Veykril:


crates/ide-completion/src/lib.rs
```diff
@@ -289,27 +285,12 @@ pub fn resolve_completion_edits(
...
-                current_module.find_use_path(db, candidate, config.insert_use.prefix_kind, cfg)
-            })
-            .find(|mod_path| mod_path.display(db, current_edition).to_string() == full_import_path);
-        if let Some(import_path) = import {
-            insert_use::insert_use(
-                &new_ast,
-                mod_path_to_ast(&import_path, current_edition),
-                &config.insert_use,
-            );
-        }
```
Veykril, line 312: @SomeoneToIgnore to you remember why we did this back then? It seems entirely unnecessary to me

SomeoneToIgnore: I have merely moved this piece around, so not sure.

For myself, I've explained this as there are certain imports that take overly long time to do `find_use_path`, so we are not doing it during completion list returning, and doing during the resolve part.

Veykril: Oh okay, I only recalled you doing a bunch of completion resolve stuff in the past.

> For myself, I've explained this as there are certain imports that take overly long time to do find_use_path, so we are not doing it during completion list returning, and doing during the resolve part.

Right, I don't think we are doing this right now, but we might very well in the future. The `find` certainly wouldn't really allow it to either.

PR #19036, opened 2025-01-26
Split out `ExpressionStore` from `Body`

Veykril:
Simple refactor that can be split out for the https://github.com/rust-lang/rust-analyzer/issues/7434 work


crates/hir-def/src/lib.rs
```diff
@@ -927,18 +929,23 @@ pub enum GenericDefId {
     ConstId(ConstId),
     FunctionId(FunctionId),
     ImplId(ImplId),
+    // can't actually have generics currently, but they might in the future
+    // More importantly, this completes the set of items that contain type references
+    // which is to be used by the signature expression store in the future.
+    StaticId(StaticId),
```
Veykril, line 935: Everything in this PR is basically just moving and renaming parts except for this. If we don't put static in here we'd need to create an entirely new id enum that is effectively `GenericDefId` but with `Static`. I don't feel like we gain anything from such a split and it should simplify the coming changes a bunch.

PR #19062, opened 2025-01-28
Fix scip indexing of module names

darichey:
* Fix an overflow in computing `is_inherent_impl` (`moniker.identifier.description.len() - 2`). No behavior change, just annoying when trying to debug.
* Fix missing symbol name for occurrences of `crate`. Where we previously emitted, e.g., `{ "range": [ 5, 4, 9 ] }`, we will now emit `{ "range": [ 5, 4, 9 ], "symbol": "rust-analyzer cargo foo . crate/" }`.
* Fix missing symbols for implicit module names. We previously emitted occurrences of module names but not the actual symbol definitions for modules implicitly defined by files.

### DIFF
--- a/crates/ide/src/moniker.rs
+++ b/crates/ide/src/moniker.rs
@@ -289,7 +289,10 @@ fn def_to_non_local_moniker(
     definition: Definition,
     from_crate: Crate,
 ) -> Option<Moniker> {
-    let module = definition.module(db)?;
+    let module = match definition {
+        Definition::Module(module) if module.is_crate_root() => module,
+        _ => definition.module(db)?,
+    };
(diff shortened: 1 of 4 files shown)


Veykril (approved):

darichey:
This also affects LSIF output. The changes look reasonable, but I don't know how to verify they are correct. cc @Urhengulas

darichey:
or @Veykril since you added the final version of the [lsif_contains_generated_constant](https://github.com/rust-lang/rust-analyzer/pull/19062/commits/a81712b9a648b62d0d35c1e28edadd5aad538c7c) test.

Veykril (approved):

Issue #19059, opened 2025-01-28
No type hint and no formatting


rcitsz:
rust-analyzer v0.3.2282  
rustc 1.83.0 (90b35a623 2024-11-26)  
vscode version 1.94  

no custom relevant settings, all vscode config is default.

my code  as follows:

```rust
 // Cannot recognize join, and there are no type hints for vals, but rust-analyzer v0.3.2273 is normal.
 let vals = [“xxx”,"yyy"].iter().map(|v| v.to_string()).collect::<Vec<_>>().join(",");

// Formating not work
async fn fx() {
   let res = async {
       // here is my code,  but formating not work
   }.await;
}

// if do this, formating is ok
async fn fx() {
   let res = async {
       // here is my code,  formating work
   };
  res.awiat;
}

```

ikatson:
Yeah, 2025-01-27 is broken, I needed to revert too. Simple repro that broke for me

```
use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver};
use std::sync::Mutex;

fn broken_types() {
    let (tx, rx) = unbounded_channel::<()>();
    let mutex = Mutex::new(rx);
    let g = mutex.lock().unwrap();

    /// This "poll_recv" is unrecognized by typing and autocomplete
    g.poll_recv(..);
}
```

dnlsndr:
Every. Single. Update. Stuff breaks...

Veykril:
@rcitsz
Your code snippet contains invalid rust characters `”`, with those fixed stuff formats for me as normal

Veykril:
https://github.com/rust-lang/rust-analyzer/pull/19012 probably broke things for <1.84 toolchains because the receiver impls are missing

Veykril:
That confirms my suspicion then, I have kicked off a new release with the fix so it should be available in an hour or so

PR #19064, opened 2025-01-28
fix: Make changes in `arbitrary-self-types` work with rustc < 1.84

ShoyuVanilla:
This makes `arbitrary-self-types` work again, without backward compatibility issues like #19059

I'm afraid that this might interrupt migration to next-gen solver as this uses chalk types.
So, please consider this as a PoC rather than a final PR if it does.


Veykril (commented):
That's a lot of extra work we'd be doing for every method call ... I'd say let's wait how fast `Receiver` is being stabilized, if it takes a while we can just accept the MSRV bump

PR #19067, opened 2025-01-29
Update TentHash to version 1.0

cessen:
The TentHash spec was frozen Jan 1st 2025, and release 1.0 of the Rust crate is a minor cleanup as a follow-up to that, representing a commitment to API stability as well.

The hash output remains the same as version 0.4, which rust-analyzer was previously using. The only API change was a struct rename.


lnicola (approved):
Thanks!

PR #19072, opened 2025-01-29
Fix #19071: ensure `completion_item_hash` serializes items uniquely

### LINKED ISSUE #19071: Potential for data collisions in `completion_item_hash()` implementation
I'm not sure if this qualifies as a bug, but it's some code that makes me *nervous*.  I ran across it when putting together #19067, and want to bring it to the project's attention.

In the implementation of [`completion_item_hash()`](https://github.com/rust-lang/rust-analyzer/blob/13cde501e00441cd1f5f42a522aaf1bfa4f7088c/crates/rust-analyzer/src/lib.rs#L68C4-L68C24), data about a completion item is concatenated and hashed to create a unique hash identifier for that item.  The hash being used, [TentHash](https://github.com/cessen/tenthash), is a robust fingerprinting hash, and therefore is not a concern (full disclosure: I developed TentHash).

However, concatenating data has the potential to create *data collisions* if not done carefully, where the actual concatenated data ends up being identical between two differing items.  This potential for data collisions is present whenever the data being concatenated is variable-length (i.e. some data fields can change in length or are optional).  As a trivial example: `a` + `bc`, `ab` + `c`, and nothing + `abc` will all end up the same.  And that is what is making me nervous.

To get more specific, here is an example of such a variable-length/optional field in `completion_item_hash()`'s code:

https://github.com/rust-lang/rust-analyzer/blob/13cde501e00441cd1f5f42a522aaf1bfa4f7088c/crates/rust-analyzer/src/lib.rs#L82-L88

(Note that [`TentHash::update()`](https://docs.rs/tenthash/latest/tenthash/struct.TentHash.html#method.update) is equivalent to appending data to a continuous stream being hashed, so these calls are equivalent to concatenating the data being hashed.)

It's possible that due to other specifics of the data, these kinds of concatenation collisions aren't actually possible in this case.  But unless someone has *actually verified that that's the case*, it would be a good idea to use something like length prefixing to ensure that it's impossible by construction.

I would be happy to put together a PR if that would be helpful.

cessen:
Previously it may have been possible for different completion items to produce colliding hashes, not because of the hash but because of how the items were serialized into byte streams for hashing. See #19071 for details.

The chances of that happening were low, if it was actually possible at all. Nevertheless, this commit ensures that it definitely can't happen.

This commit uses a handful of techniques used to fix this, but they all boil down to "ensure this could be re-parsed". If it's possible to parse to recreate the original item, then by construction there is no chance of two different items getting serialized to identical byte streams.

### DIFF
--- a/crates/rust-analyzer/src/lib.rs
+++ b/crates/rust-analyzer/src/lib.rs
@@ -118,18 +118,18 @@ fn completion_item_hash(item: &CompletionItem, is_ref_completion: bool) -> [u8; 
         u8::from(item.trigger_call_info),
     ]);
 
-    hasher.update(item.label.primary.len().to_le_bytes());
+    hasher.update(item.label.primary.len().to_ne_bytes());
     hasher.update(&item.label.primary);
 
     hasher.update([u8::from(item.label.detail_left.is_some())]);
     if let Some(label_detail) = &item.label.detail_left {
-        hasher.update(label_detail.len().to_le_bytes());
+        hasher.update(label_detail.len().to_ne_bytes());
         hasher.update(label_detail);
     }
 
     hasher.update([u8::from(item.label.detail_right.is_some())]);
     if let Some(label_detail) = &item.label.detail_right {
-        hasher.update(label_detail.len().to_le_bytes());
+        hasher.update(label_detail.len().to_ne_bytes());
         hasher.update(label_detail);
     }
 
@@ -140,15 +140,15 @@ fn completion_item_hash(item: &CompletionItem, is_ref_completion: bool) -> [u8; 
     // while not really making completion properties more unique as they are already.
 
     let kind_tag = item.kind.tag();
-    hasher.update(kind_tag.len().to_le_bytes());
+    hasher.update(kind_tag.len().to_ne_bytes());
     hasher.update(kind_tag);
 
-    hasher.update(item.lookup.len().to_le_bytes());
+    hasher.update(item.lookup.len().to_ne_bytes());
     hasher.update(&item.lookup);
 
     hasher.update([u8::from(item.detail.is_some())]);
     if let Some(detail) = &item.detail {
-        hasher.update(detail.len().to_le_bytes());
+        hasher.update(detail.len().to_ne_bytes());
         hasher.update(detail);
     }
 
@@ -162,12 +162,12 @@ fn completion_item_hash(item: &CompletionItem, is_ref_completion: bool) -> [u8; 
             CompletionItemRefMode::Dereference => 2u8,
         };
         hasher.update([discriminant]);
-        hasher.update(u32::from(*text_size).to_le_bytes());
+        hasher.update(u32::from(*text_size).to_ne_bytes());
     }
 
-    hasher.update(item.import_to_add.len().to_le_bytes());
+    hasher.update(item.import_to_add.len().to_ne_bytes());
     for import_path in &item.import_to_add {
-        hasher.update(import_path.len().to_le_bytes());
+        hasher.update(import_path.len().to_ne_bytes());
         hasher.update(import_path);
     }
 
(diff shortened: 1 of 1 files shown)


crates/rust-analyzer/src/lib.rs
```diff
@@ -115,35 +117,59 @@ fn completion_item_hash(item: &CompletionItem, is_ref_completion: bool) -> [u8;
...
-    hasher.update(item.kind.tag());
+
+    let kind_tag = item.kind.tag();
+    hasher.update(kind_tag.len().to_le_bytes());
+    hasher.update(kind_tag);
+
+    hasher.update(item.lookup.len().to_le_bytes());
     hasher.update(&item.lookup);
+
+    hasher.update([u8::from(item.detail.is_some())]);
```
lnicola, line 149: Not that it matters, but `None` and `Some("")` are likely to be the same for practical purposes.

cessen: Yeah, I considered that as well, and I'm sure you're right.  But I think it's simplest to just treat everything as a "serialize the struct" problem, and not worry about special cases like this and whether other code actually treats those cases as the same or not.

crates/rust-analyzer/src/lib.rs
```diff
@@ -115,35 +117,59 @@ fn completion_item_hash(item: &CompletionItem, is_ref_completion: bool) -> [u8;
         u8::from(item.deprecated),
         u8::from(item.trigger_call_info),
     ]);
+
+    hasher.update(item.label.primary.len().to_le_bytes());
```
lnicola, line 121: `to_ne_bytes` would be enough, we don't persist these for a long time, and if CI passes, we're not checking for specific hash values in tests.

lnicola (approved):

crates/rust-analyzer/src/lib.rs
cessen: There was one place that was already using `to_le_bytes`, so I decided to follow suite.  But I agree `to_ne_bytes` makes more sense in this case.

Do you want me to change the one existing case of `to_le_bytes` as well, or leave it alone?

crates/rust-analyzer/src/lib.rs
```diff
@@ -115,35 +117,59 @@ fn completion_item_hash(item: &CompletionItem, is_ref_completion: bool) -> [u8;
...
+
+    hasher.update([u8::from(item.detail.is_some())]);
     if let Some(detail) = &item.detail {
+        hasher.update(detail.len().to_le_bytes());
         hasher.update(detail);
     }
+
     hash_completion_relevance(&mut hasher, &item.relevance);
+
+    hasher.update([u8::from(item.ref_match.is_some())]);
```
lnicola, line 157: Here (and above, for `relevance.function` we could renumber the variants from 1 and encode `None` as 0, saving one whole (:grimacing:) byte.

SomeoneToIgnore:
Looks nice to me, thank you for improving this.

The only question I have, after reading the issue: are `0u8`, `1u8` and `2u8` really better than their string counterparts?
The issue mentions `a + bc` and `ab + c` uniqueness concern, but here I see that multiple different strings are mapped to the same digits (`could_unify`, `non_exact` and `"&mut "` were different but now are all `1u8`) — is this really an improvement?

crates/rust-analyzer/src/lib.rs
lnicola: Sure, let's use `to_ne_bytes`.

lnicola:
I think it's an improvement in the sense we're hashing slightly less stuff. We're not going to get collisions for using the same discriminants (because we keep track of each field), but also I don't think it we could have gotten collisions before.

crates/rust-analyzer/src/lib.rs
cessen: We could, but in this case since there's also another data item in the `Option` I feel like that makes it less obvious why the serialization is correct.  But if you think it would be better, let me know and I'll be happy to change it.

PR #19080, opened 2025-02-02
More rustfmt hang investigations

Veykril:
We desperately need to build better perf monitoring infra, the `RA_PROFILE` setup is honestly quite useless

### DIFF
--- a/crates/rust-analyzer/src/reload.rs
+++ b/crates/rust-analyzer/src/reload.rs
@@ -701,8 +701,7 @@ impl GlobalState {
 
         let (crate_graph, proc_macro_paths, ws_data) = {
             // Create crate graph from all the workspaces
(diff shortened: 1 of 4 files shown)


crates/rust-analyzer/src/reload.rs
```diff
@@ -701,8 +701,7 @@ impl GlobalState {
 
         let (crate_graph, proc_macro_paths, ws_data) = {
             // Create crate graph from all the workspaces
-            let vfs = &mut self.vfs.write().0;
-
+            let vfs = &self.vfs.read().0;
```
Veykril, line 704: This could've caused some requests (including formatting) to wait on acquiring their read lock when we were recreating the crate graph. This was needed to be a write lock before, but we've since changed the loading behavior here, so this is no longer the case.

PR #19096, opened 2025-02-05
Allow rust-project.json to specify sysroot workspace

darichey:
Following up on https://github.com/rust-lang/rust-analyzer/issues/18590#issuecomment-2515320962 and https://github.com/rust-lang/rust-analyzer/pull/18788 to add a `sysroot_project` key to rust-project.json which allows specifying a nested rust-project for the sysroot. This gives control over the sysroot layout to the rust-project generator (e.g., Buck). At work, we can build the sysroot with Buck, so this change allows us to make use of that instead of the stitched sysroot. Most users of Buck and Bazel will still want to use Cargo for the sysroot.


darichey:
There is still some work to do here, but I wanted to get feedback on the overall approach before moving forward and making the corresponding changes in Buck's [rust-project](https://github.com/facebook/buck2/tree/7aa4e26277869340889a331e0de7c056cc345dcc/integrations/rust-project).

We can also delete stitched sysroot here if we change Cargo to be the fallback.

crates/project-model/src/workspace.rs
```diff
@@ -566,7 +573,9 @@ impl ProjectWorkspace {
                         Some(PackageRoot { is_local: false, include, exclude })
                     })
                     .collect(),
-                SysrootWorkspace::Stitched(_) | SysrootWorkspace::Empty => vec![],
+                SysrootWorkspace::Json(_)
```
Veykril, line 576: Do we not need to handle this as well?

crates/project-model/src/project_json.rs
```diff
@@ -65,6 +65,7 @@ pub struct ProjectJson {
     pub(crate) sysroot: Option<AbsPathBuf>,
     /// e.g. `path/to/sysroot/lib/rustlib/src/rust/library`
     pub(crate) sysroot_src: Option<AbsPathBuf>,
+    pub(crate) sysroot_project: Option<ProjectJsonData>,
```
Veykril, line 68: Shouldn't this be `Option<ProjectJson>?`

Veykril, line 68: Comment above would be nice

Veykril (commented):
Seems reasonable to me, I assume (should we rid of the stitched variant), setting `sysroot`, `sysroot_src` and `sysroot_project` as `None` would imply not to load a sysroot? Or should we have a separate flag to opt-into not loading. Maybe just `sysroot_src` and `sysroot_project` being `None` should opt-out (as a `sysroot` dir is still useful for tools). I don't know the use-cases of alternate build systems too well here

darichey:
I've added another commit that deletes stitched sysroot. The current behavior is...
* For cargo projects, use cargo
* For json projects...
    1. If we have both `sysroot_src` and `sysroot_project`, use that
    2. Otherwise, try cargo
    3. If cargo fails, so sysroot is loaded

I'm not aware of a use case for explicitly opting out of the sysroot entirely. I think if it's not needed for cargo projects, it shouldn't be needed for json projects either.

Veykril (commented):
Will need a rebase over my latest changes which likely conflict a fair bit, sorry. I don't think cargo projects ever want to specify not having a sysroot rust source so removing that should be fine. I do now wonder though whether rust-project.json even wants a sysroot source field at all, or if instead we want to expose the lang origin thing for its specified crate data. That is have it embed the sysroot into its crate graph itself and tag the relevant crates accordingly?

Veykril:
@davidbarsky  regarding my comment there, specifically
> That is have it embed the sysroot into its crate graph itself and tag the relevant crates accordingly?

I believe that is something you've been thinking about this week basically (in the space of cargo metadata, but here it applies more directly wrt to the json format)

davidbarsky:
> @davidbarsky  regarding my comment there, specifically
> 
> > That is have it embed the sysroot into its crate graph itself and tag the relevant crates accordingly?
> 
> 
> 
> I believe that is something you've been thinking about this week basically (in the space of cargo metadata, but here it applies more directly wrt to the json format)

Yeah, I have been thinking about this because I was trying to see if there's a way to unify the paths for handling a `rust-project.json` with a rustup-managed rustc source, a `rust-project.json` with a build system-managed rustc source, `cargo-metadata` as it exists today, and `cargo-metadata` with a `-Zbuild-std`. This was _mostly_ a dead-end because it doesn't seem feasible to do that today, but directly tagging crates in the crate graph with an attribute along the lines of "this is a `rust-src`/prelude crate" is one possible direction for a future design of `rust-project.json`. Regardless, I think that the approach @darichey took _here_ is probably the most optimal under the current circumstances.

I also chatted with @darichey and @capickett this week, where I learned that Buck can use a Buckified `rust-src` (this component isn't open-source, but it should be!) where the overall build looks _extremely_ similar to where `-Zbuild-std` will probably end up: namely, there's only one `cargo-metadata`-shaped call, so the `rust-src` crates are included inline. However, I imagine most non-Facebook/Google users of Buck and Bazel would prefer to use a rustup-managed sysroot because they don't want to touch `RUST_BOOTSTRAP=1`.

Veykril (approved):

davidbarsky:
I just realized that I actually don't know what a "stitched workspace" refers to: I can't really find it in my limited scrollback!

Veykril:
> I just realized that I actually don't know what a "stitched workspace" refers to: I can't really find it in my limited scrollback!

Oh I thought I replied to this by mail, sorry. Stitched just means we manually picked the crates in the sysroot from a known list. That logic predates the sysroot now shipping a proper cargo workspace (which is why we can drop support for it now, its been long enough)

crates/project-model/src/tests.rs
```diff
@@ -225,18 +223,6 @@ fn rust_project_cfg_groups() {
     check_crate_graph(crate_graph, expect_file!["../test_data/output/rust_project_cfg_groups.txt"]);
 }
 
-#[test]
```
darichey, line 228: I deleted this test because it's really testing a behavior of the stitched sysroot.

Veykril (approved):

Veykril:
Thanks!

Issue #15661, opened 2023-09-24
An option to disable inlay Type hints for Closure parameters


not-holar:
I can't be the only one for whom the Types of closure parameters in chains get pretty long, the line often overflows the screen and it starts to look very messy

I also like to have a low opacity for inlay hints, which works great for other hints, but makes the closures specifically look broken up and harder to follow

Since anonymous closures are usually very small and limited in scope, it's often obvious what the types are anyway, so i think letting the user decide where the Type hints appear more granularly would be very helpful

butuzov:
You not the only one. Please add this configuration.

Veykril:
Relevant config is defined here https://github.com/rust-lang/rust-analyzer/blob/32b86a837805a4162a01b2356c65ef9dd53dbae6/crates/rust-analyzer/src/config.rs#L204-L210

Probably just needs a new `hide*` variant that gets threaded through like the others

PR #19104, opened 2025-02-06
option to disable inlay Type hints for Closure parameters

jnyfah:
Fix: #15661

Add a parameter to hide inlay parameter type hints for closures only


crates/ide/src/inlay_hints/bind_pat.rs
```diff
@@ -36,6 +36,9 @@ pub(super) fn hints(
                 if it.ty().is_some() {
                     return None;
                 }
+                if config.hide_closure_parameter_hints && it.syntax().ancestors().any(|n| matches!(ast::Expr::cast(n), Some(ast::Expr::ClosureExpr(_)))) {
```
Veykril, line 39:
```suggestion
                if config.hide_closure_parameter_hints && it.syntax().ancestors().nth(2).is_none_or(|n| matches!(ast::Expr::cast(n), Some(ast::Expr::ClosureExpr(_)))) {
```

PR #19105, opened 2025-02-06
fix: Don't emit empty scip occurrence for builtins

darichey:
We currently emit occurrences without a symbol name (e.g., `{ "range": [ 1, 7, 10 ] }`) for builtins (e.g., `i64`). This happens because `ide::moniker::def_to_moniker` returns `None` for builtins:
https://github.com/rust-lang/rust-analyzer/blob/ca47cddc31ae76a05e8709ed4aec805c5ef741d3/crates/ide/src/moniker.rs#L253-L257

It never makes sense to emit an occurrence without a symbol name, so just skip in that case.

### DIFF
--- a/crates/rust-analyzer/src/cli/scip.rs
+++ b/crates/rust-analyzer/src/cli/scip.rs
@@ -139,45 +139,42 @@ impl flags::Scip {
             let mut occurrences = Vec::new();
             let mut symbols = Vec::new();
 
-            tokens.into_iter().for_each(|(text_range, id)| {
+            for (text_range, id) in tokens.into_iter() {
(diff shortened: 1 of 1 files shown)


Veykril (approved):

Veykril:
We could maybe point the primitive types to their core re-exports https://doc.rust-lang.org/src/core/primitive.rs.html or rustdoc module shims https://doc.rust-lang.org/src/core/primitive_docs.rs.html

davidbarsky:
I feel like the rustdoc module shims would be more useful for users, but I don't feel _too_ strongly.

PR #19110, opened 2025-02-07
Simplify panic_context

eagr:
Housekeeping. `PanicContext::init` doesn't seem necessary.


eagr:
Came across this when browsing the codebase. Thought it'd nice to simplify it a bit. If you think this is not worth it, feel free to close :))

Bikeshed:
* could also get rid of `with_context`: const `CTX` is evaluated at compile time anyways and initialized at the first `with()` call, maybe could place the `thread_local!` at mod level?
* `PanicContext` -> `ThreadContext`?

Veykril (approved):

Veykril:
Thanks!

PR #19111, opened 2025-02-07
fix: Apply adjustments to proper expr when invoking `CoerceMany`

### LINKED ISSUE #19021: Overloaded deref on type Foo is not a projection
This can probably be minimized further, but I don't understand the code well enough...
```rust
struct Foo {}

fn foo() {
    let x: Box<Foo> = todo!();
    let y: &Foo = todo!();

    || match x {
        ref x => x,
        _ => y,
    }
}
```
```
internal error: entered unreachable code: Overloaded deref on type Foo is not a projection
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/panicking.rs:74:14
   2: hir_ty::mir::ProjectionElem<V,T>::projected_ty
             at /Users/davidrichey/Desktop/test/rust-analyzer/crates/hir-ty/src/mir.rs:162:21
   3: hir_ty::infer::closure::<impl hir_ty::infer::InferenceContext>::restrict_precision_for_unsafe
             at /Users/davidrichey/Desktop/test/rust-analyzer/crates/hir-ty/src/infer/closure.rs:1050:22
   4: hir_ty::infer::closure::<impl hir_ty::infer::InferenceContext>::analyze_closure
             at /Users/davidrichey/Desktop/test/rust-analyzer/crates/hir-ty/src/infer/closure.rs:1299:9
   5: hir_ty::infer::closure::<impl hir_ty::infer::InferenceContext>::infer_closures
             at /Users/davidrichey/Desktop/test/rust-analyzer/crates/hir-ty/src/infer/closure.rs:1339:24
   6: hir_ty::infer::infer_query
             at /Users/davidrichey/Desktop/test/rust-analyzer/crates/hir-ty/src/infer.rs:143:5
```
https://github.com/rust-lang/rust-analyzer/blob/6e4c29f7ce18cea7d3d31237a4661ab932eab636/crates/hir-ty/src/mir.rs#L156-L168

ShoyuVanilla:
Fixes #19021

I think that this issue was quite hard to find (I wonder how David found this out 😄 )

For instance, removing lang item annotation `#[lang = "owned_box"]`, making the last expression of the `fn main()` into a plain match expression instead of a closure, or annotating the closure return type to the following code won't trigger the issue.

```rust
use core::ops::Deref;

#[lang = "owned_box"]
struct Box<T>(T);

impl<T> Deref for Box<T> {
    type Target = T;
    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

struct Foo;

fn main() {
    let x = Box(Foo);
    let y = &Foo;
    || match x {
        ref x => x,
        _ => y,
    };
}
```

The path in which above code triggers the issue is quite interweaven, but the root cause is that we are applying resulting adjustments of the `CoreceMany` to the wrong expression; we are applying it always to the very expression that we are coercing from.

If the next arm/branch of the `CoerceMany` can coerce into the previous `merge_ty`, this is not a problem.
But if the previous `merge_ty` has to be coerced into the type of next arm/branch, the adjustments must be applied to the previous `CoerceMany` expressions.

So, in the above code, when calling `CoerceMany::coerce` with the second arm, the match expression's type is inferred as `&Foo` and the first branch must be coerced into it.
But adjustments for this coercion `[Deref, Deref, Borrow]` is applied to the next arm, `y`, and since `y = &Foo` cannot be dereferenced twice, and this causes the issue.

### DIFF
--- a/crates/hir-ty/src/tests/coercion.rs
+++ b/crates/hir-ty/src/tests/coercion.rs
@@ -185,11 +185,10 @@ fn test() {
     let t = &mut 1;
     let x = match 1 {
         1 => t as *mut i32,
+           //^^^^^^^^^^^^^ adjustments: Pointer(MutToConstPointer)
         2 => t as &i32,
            //^^^^^^^^^ expected *mut i32, got &'? i32
         _ => t as *const i32,
-          // ^^^^^^^^^^^^^^^ adjustments: Pointer(MutToConstPointer)
-
     };
     x;
   //^ type: *const i32
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -1239,7 +1239,29 @@ impl<'a> InferenceContext<'a> {
     }
 
     fn write_expr_adj(&mut self, expr: ExprId, adjustments: Vec<Adjustment>) {
-        self.result.expr_adjustments.insert(expr, adjustments);
+        if adjustments.is_empty() {
+            return;
+        }
+        match self.result.expr_adjustments.entry(expr) {
+            std::collections::hash_map::Entry::Occupied(mut entry) => {
+                match (&mut entry.get_mut()[..], &adjustments[..]) {
+                    (
+                        [Adjustment { kind: Adjust::NeverToAny, target }],
+                        [.., Adjustment { target: new_target, .. }],
+                    ) => {
+                        // NeverToAny coercion can target any type, so instead of adding a new
+                        // adjustment on top we can change the target.
+                        *target = new_target.clone();
+                    }
+                    _ => {
+                        *entry.get_mut() = adjustments;
+                    }
+                }
+            }
+            std::collections::hash_map::Entry::Vacant(entry) => {
+                entry.insert(adjustments);
+            }
+        }
     }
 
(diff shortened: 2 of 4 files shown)


crates/hir-ty/src/tests/coercion.rs
```diff
@@ -185,11 +185,10 @@ fn test() {
     let t = &mut 1;
     let x = match 1 {
         1 => t as *mut i32,
+           //^^^^^^^^^^^^^ adjustments: Pointer(MutToConstPointer)
         2 => t as &i32,
            //^^^^^^^^^ expected *mut i32, got &'? i32
         _ => t as *const i32,
-          // ^^^^^^^^^^^^^^^ adjustments: Pointer(MutToConstPointer)
-
```
ShoyuVanilla, line 192: This test also illustrates the issue; as the `x = match 1 { ... }` is inferred as `*const i32`, the `MutToconstPointer` adjustment must be applied to `t as *mut i32`, not `t as *const i32`

Veykril (approved):
Thanks!

PR #19131, opened 2025-02-10
doc: ignore generated configuration docs as well

joshrotenberg:
I believe in the original mdbook PR there was some unresolved discussion around whether or not the generated documentation should be committed or ignored. It looks like we decided to ignore three of the four, so this PR ignores the `configuration_generated.md` as well for consistency. If there is a reason we don't want to do that, though, feel free to close.


joshrotenberg:
Ok, now I'm remembering some kind of chicken/egg issue with that file.

Wilfred:
Yeah, some of the generated files are updated with tests, others by `xtask codegen`. I think the obvious next step is to merge dev/docs/ into the manual, so we at least have a clear place for instructions about each of these files.

joshrotenberg:
This needs to be thought through and probably reworked so I'm going to close this iteration and do some of that thinking.

PR #19132, opened 2025-02-10
doc: add xtask codegen command in development README as well

joshrotenberg:
We suggest running both `cargo test` and `cargo xtask codgen` to generate the docs in the book's top level README.md but not in `docs/book/README.md`. Add it there too.


Veykril (approved):

Veykril:
Thanks!

PR #19133, opened 2025-02-10
doc: fix mdbook repository paths

joshrotenberg:
The paths that point to the `Repository` and `Suggest an edit` links are incorrect (based on the original PR's assumption of `/manual`).


Wilfred (approved):
Nice catch. I don't have power to merge this, but LGTM.

ChayimFriedman2 (approved):
Thanks!

PR #19137, opened 2025-02-11
fix: highlight `extern crate` in doc comments

### LINKED ISSUE #19041: Import aliases (crates) in doc comment injection retain their unresolvedReference
```rs
/// ```
/// use thing as Unresolved;
/// extern crate Krate;
/// ```
struct S;
```

here `Unresolved` is tagged `unresolvedReference` (and `Krate` too), even though we strip this tag usually. Probably not hjere as these two identifiers are considered names (and usually names do not turn `unresolvedReference`!)

asuto15:
I'm new to this project and still getting familiar with the codebase.
I've added a branch to ensure that `extern crate Krate;` is highlighted as a module in `highlight_name_ref_by_syntax`, and I've updated the corresponding doctest expectations accordingly. If I've misunderstood the intended behavior or made any mistakes in these changes, please let me know. I'm happy to adjust the patch as needed.

Fixes #19041.


lnicola (approved):
I'm pretty sure I tried this and it made no difference. Thanks!

lnicola:
Hm, in Rust code we mark these as namespaces with `crateRoot` and `library` modifiers, we should probably match that.

Veykril:
Given it could be `extern crate self` the `library` mod won't necessarily be applicable, but we should tag it with `CrateRoot`

asuto15:
I have modified the code and confirmed by both the "Inspect Editor Tokens and Scopes" command and tests, the `crateRoot` and `library` mods are applied to extern crate. 

I also attempted to apply the default_library modifier as follows:
```rust
let is_from_builtin_crate = krate.is_builtin(sema.db);
if is_from_builtin_crate {
    h |= HlMod::DefaultLibrary;
}
```
However, for some reason the `is_builtin` function did not behave as expected, and consequently the `defaultLibrary` modifier was not applied. So, I have omitted this code from the commit for now.

If possible, should I also apply the `defaultLibrary` mod? If you have any insights about this issue, please let me know.

crates/ide/src/syntax_highlighting/highlight.rs
```diff
@@ -703,6 +703,24 @@ fn highlight_name_ref_by_syntax(
...
+                }
+            } else {
+                false
+            };
+
+            if is_crate_root {
+                h |= HlMod::CrateRoot;
+            }
+
+            h | HlMod::Library
```
Veykril, line 722:
```suggestion
            HlTag::Symbol(SymbolKind::Module) | HlMod::CrateRoot
```
this should suffice, it might not be a library (without semantic info we don't know), but its always going to be a crate root either way.

asuto15:
Thank you for the suggestion. I've updated the patch.

lnicola:
Thanks!

Veykril (approved):

PR #19139, opened 2025-02-11
manual: Fix URLs to rustdoc pages

Wilfred:
Now that the manual lives at /manual/, we need to use absolute URLs to link to rustdoc content.


joshrotenberg:
Nice catch, right back 'atcha!

PR #19142, opened 2025-02-12
Spawn toolchain querying processes in parallel

Veykril:


crates/project-model/src/workspace.rs
```diff
@@ -224,134 +238,207 @@ impl ProjectWorkspace {
...
+                toolchain.join()?,
+                rustc_cfg.join()?,
+                data_layout.join()?,
+                rustc_dir.join()?,
+                loaded_sysroot.join()?,
+                cargo_metadata.join()?,
+                cargo_config_extra_env.join()?,
+            ))
+        })
+        .unwrap();
```
davidbarsky, line 367:
```suggestion
        .expect("Unable to join threads");
```

crates/project-model/src/sysroot.rs
```diff
@@ -34,14 +34,14 @@ pub struct Sysroot {
 }
 
 #[derive(Debug, Clone, Eq, PartialEq)]
-pub(crate) enum SysrootWorkspace {
+pub enum SysrootWorkspace {
```
davidbarsky, line 37: While we're here, should this be renamed to `RustSrc`, since this isn't the sysroot _per-se_, but rather a component in a sysroot called `rust-src`?

davidbarsky (approved):
lgtm, modulo nits

PR #19151, opened 2025-02-12
(revert c2258d8) Use cargo env instead of crate_workspace_data for procmacro expansion

infiniteregrets:
I noticed that there is a bug in the rust-analyzer where on hover/cmd+click on types that are decorated with custom macros being used from a different crate doesn't work.

This happens specifically between commit 08c7bbc2d which is the good version and commit 94b526fc8 which is the bad version - these commits are also corresponding releases.

After running git bisect I found that c2258d88 is the bad commit.

For more context, I discovered this issue based on the usage of https://github.com/s2-streamstore/sync_docs in https://github.com/s2-streamstore/s2-sdk-rust which when used in a separate cargo project fails to show docs/go to definition for a type using `#[sync_docs]`

This commit (c2258d88) uses `db.crate_workspace_data()[&calling_crate].proc_macro_cwd` to provide the current_dir - which is always the current workspace's `current_dir` as seen here: https://github.com/rust-lang/rust-analyzer/blob/c661e5e14e2b3f2467438a49ec8d2bdfcbfd9a79/crates/rust-analyzer/src/reload.rs#L888

Based on my understanding, this seems problematic as it can't resolve these macros where the `proc_macro_cwd` should be the actual `current_dir` of the crate where the macro is used.

For the good commit after adding some logs I see the following `current_dir` being passed:

```
...
2025-02-12T16:17:24.611516Z INFO expanding macro: "sync_docs"
2025-02-12T16:17:24.611527Z INFO current_dir: Some("/Users/infiniteregrets/.cargo/registry/src/index.crates.io-6f17d22bba15001f/streamstore-0.6.0")
2025-02-12T16:17:24.838798Z INFO expanding macro: "Message"
2025-02-12T16:17:24.838807Z INFO current_dir: Some("/Users/infiniteregrets/.cargo/registry/src/index.crates.io-6f17d22bba15001f/google-cloud-googleapis-0.16.0")
```

For the bad commit:

```
...
2025-02-12T16:27:53.141436Z INFO expanding macro: "Message"
2025-02-12T16:27:53.141439Z INFO current_dir: Some("/Users/infiniteregrets/Documents/s2")
2025-02-12T16:27:53.141547Z INFO expanding macro: "sync_docs"
2025-02-12T16:27:53.141552Z INFO current_dir: Some("/Users/infiniteregrets/Documents/s2")
```

I have tried playing around with setting the `ws_data` to be the actual package in question's directory, however that seems like there isn't really an API to do so. I tried getting data from `CargoWorkspace.packages` as it has manifest path whose parent should work here in 

https://github.com/rust-lang/rust-analyzer/blob/c661e5e14e2b3f2467438a49ec8d2bdfcbfd9a79/crates/rust-analyzer/src/reload.rs#L888

```rs
ws_data.extend(mapping.values().copied().zip(iter::repeat(Arc::new(CrateWorkspaceData {
            toolchain: toolchain.clone(),
            data_layout: target_layout.clone(),
            proc_macro_cwd: Some(ws.workspace_root().to_owned()),
        }))));
```

But I don't know this is the right direction, just reverting c2258d8 fixes things as a temporary solution - I would be happy to work on this in a followup if someone can provide a direction where I should be looking to change things/also happy if someone who has a better idea about this repo and wants to take this over and fix this!


infiniteregrets:
Hi @bjorn3 Isnt this the case when we use

```
let current_dir = env.get("CARGO_RUSTC_CURRENT_DIR").or_else(|| env.get("CARGO_MANIFEST_DIR"));
```

Unless I am missing something please correct me! Additonally, I was trying to figure how I can change the `proc_macro_cwd` to be exactly reflect what you just mentioned

Veykril:
`CARGO_RUSTC_CURRENT_DIR` is no longer being set in rust-analyzer (as the env var is also likely not going to be stabilized within cargo anymore). `CARGO_MANIFEST_DIR` I believe won't be right in the workspace case. Either way we shouldn't be extracting the working directory from the env vars.

infiniteregrets:
@Veykril I agree with you for not using env, I was playing around with the snippet here [rust-analyzer/crates/rust-analyzer/src/reload.rs](https://github.com/rust-lang/rust-analyzer/blob/c661e5e14e2b3f2467438a49ec8d2bdfcbfd9a79/crates/rust-analyzer/src/reload.rs#L888) however I haven't had any luck extracting the current_dir for the crates when being set in the mapping based on the current implementation, what do you think should be done here?

crates/project-model/src/workspace.rs
```diff
@@ -1448,8 +1451,9 @@ fn add_target_crate_root(
         Arc::new(cfg_options),
         potential_cfg_options.map(Arc::new),
         env,
-        matches!(kind, TargetKind::Lib { is_proc_macro: true }),
         origin,
+        matches!(kind, TargetKind::Lib { is_proc_macro: true }),
+        Some(pkg.manifest.parent().to_path_buf()),
```
infiniteregrets, line 1456: @Veykril is this ok to do here?

crates/base-db/src/input.rs
```diff
@@ -296,6 +296,7 @@ pub struct CrateData {
     pub dependencies: Vec<Dependency>,
     pub origin: CrateOrigin,
     pub is_proc_macro: bool,
+    pub cwd: Option<AbsPathBuf>,
```
Veykril, line 299:
```suggestion
    /// The working directory to run proc-macros in. This is the workspace root of the cargo workspace for workspace members, the crate manifest dir otherwise.
    pub proc_macro_cwd: Option<AbsPathBuf>,
```

crates/project-model/src/project_json.rs
```diff
@@ -240,6 +241,7 @@ pub struct Crate {
     pub(crate) include: Vec<AbsPathBuf>,
     pub(crate) exclude: Vec<AbsPathBuf>,
     pub(crate) is_proc_macro: bool,
+    pub(crate) cwd: Option<AbsPathBuf>,
```
Veykril, line 244:
```suggestion
    /// The working directory to run proc-macros in. This is usually the workspace root of cargo workspaces.
    pub(crate) proc_macro_cwd: Option<AbsPathBuf>,
```

crates/project-model/src/project_json.rs
```diff
@@ -362,6 +364,8 @@ struct CrateData {
     repository: Option<String>,
     #[serde(default)]
     build: Option<BuildData>,
+    #[serde(default)]
+    cwd: Option<Utf8PathBuf>,
```
Veykril, line 368:
```suggestion
    proc_macro_cwd: Option<Utf8PathBuf>,
```
@davidbarsky for input here, unsure if we need to model this in the project json payload or not but I don't think we can derive the correct one from the data.

crates/project-model/src/workspace.rs
```diff
@@ -1283,11 +1285,12 @@ fn detached_file_to_crate_graph(
         cfg_options.clone(),
         None,
         Env::default(),
-        false,
         CrateOrigin::Local {
             repo: None,
             name: display_name.map(|n| n.canonical_name().to_owned()),
         },
+        false,
+        Some(detached_file.parent().to_path_buf()),
```
Veykril, line 1293:
```suggestion
        None
```

crates/project-model/src/workspace.rs
Veykril: That unconditionally sets it to the manifest dir which is incorrect, as was  discussed, for workspace members this needs to be the workspace root

infiniteregrets: sorry about that, fixing

crates/project-model/src/project_json.rs
davidbarsky: I *think* for rust-project.json payloads, we should be conservative and mimic what Cargo does, but I'm not 100% sure.

infiniteregrets:
@Veykril updated, lmk if theres anything else missing!

PR #19158, opened 2025-02-15
Implement expand_glob_reexport assist

### LINKED ISSUE #13511: `expand glob import` should expand to all possible items without filtering for `pub use` reexports.
**rust-analyzer version**:  `0.4.1255-standalone`
**rustc version**: `rustc 1.66.0-nightly (3f83906b3 2022-09-24)`

```rust
mod qqq {
    use super::*;

    pub trait Tr1{}
    pub trait Tr2{}
    pub struct St1{}
    pub struct St2{}
    pub enum En1{}
}

pub use qqq::┃*;
```

After "expand glob import", I expect.

```rust
mod qqq {
    use super::*;

    pub trait Tr1{}
    pub trait Tr2{}
    pub struct St1{}
    pub struct St2{}
    pub enum En1{}
}

pub use qqq::┃{Tr1, Tr2, St1, St2, En1};
```

but I get:

```rust
mod qqq {
    use super::*;

    pub trait Tr1{}
    pub trait Tr2{}
    pub struct St1{}
    pub struct St2{}
    pub enum En1{}
}

pub use qqq::┃{};
```

Unless RA is willing to analyse the whole workspace for which reexports are actually needed, I think it should just expand to everything. This would allow for easy gradual refactor by moving chunks of code to separate files without also having multiple glob imports one after another.

Alternatively, a second, simpler code assist "Expand glob import (all items)" can be added.

Workaround: Use Ctrl+Space, Enter, Comma, Ctrl+Space, Enter, Comma, ... repeatedly until it offers `self.`

PoignardAzur:
Fixes #13511 .

The code ends up pretty messy. To handle this more cleanly, I'd want some equivalent of `hir_def::visibility::Visibility::is_visible_from` that takes a `syntax::ast::Visibility` as the origin, or at least some type that can be easily derived from `syntax::ast::Visibility`.

The general problem (from my naive viewpoint as someone unfamiliar with the codebase) is that as far as I can tell the HIR data types you can get from `ctx.sema` already lost all knowledge about `use xxx` items; and those data types are the ones that can do fancy visibility computation. (`hir_def::item_tree::Use` would be useful, but I haven't found a way to get one from a `syntax::ast::Use`)

Anyway, 99% of re-exports are `pub` or `pub(crate)` anyway, so this should already be a large improvement.

### DIFF
--- a/crates/ide-assists/src/handlers/expand_glob_import.rs
+++ b/crates/ide-assists/src/handlers/expand_glob_import.rs
@@ -3,10 +3,11 @@ use hir::{AssocItem, Enum, HasVisibility, Module, ModuleDef, Name, PathResolutio
 use ide_db::{
     defs::{Definition, NameRefClass},
     search::SearchScope,
+    source_change::SourceChangeBuilder,
 };
 use stdx::never;
 use syntax::{
-    ast::{self, make},
+    ast::{self, make, Use, UseTree},
     ted, AstNode, Direction, SyntaxNode, SyntaxToken, T,
 };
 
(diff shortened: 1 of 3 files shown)


Veykril (approved):
Thanks!

PoignardAzur:
Well that was fast.

PR #19159, opened 2025-02-16
fix: Set `RUSTUP_TOOLCHAIN` when loading sysroot workspace

Veykril:


Veykril:
I believe the issue here is that we run `cargo metadata` on the sysroot library source but prior to this PR we did not set the `RUSTUP_TOOLCHAIN` env var for that invocation, and as such, if the `cargo` we invoke here is the rustup proxy then that might pick a different toolchain cargo than the one we want to use for current workspace being analyzed.

PR #19180, opened 2025-02-18
doc: use fully qualified url for source path

joshrotenberg:
I just realized that the links to source files in three of the four generated doc files are wrong because they assume relative paths but they point to the doc site url instead of the repository itself (configuration is correct because its a single static link in `configuration.md`). 

This fix just points to the full URL to this repo instead, not sure if there is a better option.


Veykril (approved):
Thanks!

PR #19191, opened 2025-02-20
Remove `limit` crate in favor `usize`

Veykril:
The crate doesn't pull its weight imo, I've never used the tracking feature (nor know of anyone who did)


lnicola:
I think I used it at one point, but I it's probably not worth it.

PR #19197, opened 2025-02-21
feat: update insta inline snapshot when clicking 'Update Test' runnable

andylokandy:
The 'Update Test' runnable was implemented in [PR #18757](https://github.com/rust-lang/rust-analyzer/pull/18757).

It was found that the `insta` snapshot library does not update inline snapshots using the 'Update Test' runnable because it only sets the environment variable `INSTA_UPDATE=always` which does not affect inline snapshots.

`insta` supports updating snapshots using `cargo-nextest`, which run each test in a single process, requiring the external tool `cargo-insta` to collect diffs and apply changes. Thus, updating inline snapshots with plain `cargo test` is not feasible([insta #490](https://github.com/mitsuhiko/insta/issues/490)).

This PR changes the 'Update Test' runnables to use `cargo insta test` instead of `cargo test ...` when the test involves `insta`. This change is justified since `cargo insta test` is compatible with other snapshot libraries and it is assumed that `insta` users have `cargo-insta` installed.

### DIFF
--- a/crates/ide/src/lib.rs
+++ b/crates/ide/src/lib.rs
@@ -104,7 +104,7 @@ pub use crate::{
     navigation_target::{NavigationTarget, TryToNav, UpmappingResult},
     references::ReferenceSearchResult,
     rename::RenameError,
-    runnables::{Runnable, RunnableKind, TestId},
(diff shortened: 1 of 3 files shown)


Veykril (approved):
Sounds reasonable, thanks!

PR #19200, opened 2025-02-21
Include private items in completions for local crates

### LINKED ISSUE #9850: Feature request: allow linking to private methods, or convert to public if in crate.
I'd love to add this if given proper guidance.

I frequently run into making a method, trying to call it, realizing it's not public, then not being able to jump to symbol to change it to pub. 

It would be great to enable jumping to private methods as well as being able to make a private method public from the callsite if it's already in my project.

Timmmm:
When building an import map for a module, don't skip private items if they are from a local (workspace) crate. This is explained in the comment.

Fixes #9850 (hopefully anyway; not tested because [I can't debug rust-analyzer](https://github.com/rust-lang/rust-analyzer/discussions/19199))

### DIFF
--- a/crates/hir-def/src/import_map.rs
+++ b/crates/hir-def/src/import_map.rs
@@ -132,6 +132,8 @@ impl ImportMap {
     fn collect_import_map(db: &dyn DefDatabase, krate: CrateId) -> ImportMapIndex {
         let _p = tracing::info_span!("collect_import_map").entered();
 
+        let is_local = db.crate_graph()[krate].origin.is_local();
+
         let def_map = db.crate_def_map(krate);
         let mut map = FxIndexMap::default();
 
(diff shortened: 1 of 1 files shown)


Timmmm:
Ok I managed to test this and it doesn't work. Seems like this code path is only used for third party crates or something. I can't find the relevant code for the local crates. Might need a hint...

Veykril:
The relevant logic for computing (importing) paths is in https://github.com/rust-lang/rust-analyzer/blob/master/crates/hir-def/src/find_path.rs

Timmmm:
Thanks @Veykril ! I found that there was already a flag for this behaviour, but unfortunately it is disabled by default (fixed in #19210 ) and didn't work for path completions (fixed in #19211 hopefully).

PR #19207, opened 2025-02-23
WIP: Add additional closing brace inlay hints

markmurphydev:
Working on additional inlay hints for #18833, and looking for some guidance.

Currently we traverse the syntax tree, checking if the current node is a block-like expression, then figuring out the type of the parent.
- It seems possible to go parent node -> child block, which might be more readable. Any opinion one way or the other?

Currently I include almost every child of the parent syntax element (ex. function's visibility, params, types, ...) in the type hint, then join newlines and truncate the string to `InlayHintsConfig.max_length`.

スクリーンショット 2025-02-22 午後6 48 36

- `HirFormatter` has a `max_size` field, but most of the implementations don't currently respect it. If I updated them to do so, and also added a configuration to exclude specific children, I think the text for each inlay hint could be gotten with a single call like `parent.display_truncate_exclude(max_size: 40, exclude: vec![block_child])`. Does that seem like a good way to go?
- I'm using `format!` on the AST nodes, because I had them already. If hir-format is not respecting `max_size`, is there still a reason to prefer HIR display functions?
- Closing hints look pretty bad when the original definition spanned multiple lines. I join the newlines, but line-beginning whitespace from child tokens (ex. function parameters) stays in. To me, the most obvious solution is to do a regex match for whitespace, but I think we don't want to pull in the `regex` crate? 

スクリーンショット 2025-02-22 午後7 35 07

- I added "// " before closing-brace inlay hints, to make clear that they don't represent code. Does it look ok?


Veykril:
Sorry for having let this sit for so long, I tend to ignore draft PRs 😅 

In my eyes, we should just render the ast nodes as is, then trim them by line endings. The HIR representation can look vastly different which may be confusing.

As for your last point, ye that is kind of a tricky problem. I think we can want to compromise here, for items keep the previous behavior that is just rendering the signature (so `fn name` for functions etc), where as for the expression like things we want to render the first line with `...` if truncated on a newline. (so no trying to render multiple lines into one line here).

Veykril:
Closing due to inactivity, feel free to re-open if you come back to it!

PR #19218, opened 2025-02-24
Disable incremental on release builds

Veykril:
cc https://github.com/rust-lang/rust/issues/137351

I do not know what intention behind this was, the introducing PR doesn't state why


lnicola:
I always assumed it was so that it would compile faster. We disable it back for releases.

Veykril:
it would explain why the install xtask actions tends to give me gibberish linker errors every now and then though

lnicola:
You can really feel the difference here :sweat_smile:.

PR #19221, opened 2025-02-24
minor: Add tabstop to impl body in `generate_trait_impl` assist

Giga-Bowser:
This matches the behavior of the `generate_impl` assist.

I have no idea what this assist-doc demorgan cruft getting pulled along is but if it's just something #19171 forgot to do I'm fine having it done here.


lnicola (approved):

lnicola:
Yeah, don't worry about the doc changes, `cargo test` does that.

ShoyuVanilla:
> I have no idea what this assist-doc demorgan cruft getting pulled along is but if it's just something #19171 forgot to do I'm fine having it done here.

I forgot to generate docs 😅 Thanks!

lnicola:
I'm not sure if we want to have those under git, but if we do, we probably want a test for it.

PR #19228, opened 2025-02-26
Split some queries

Veykril:
Small step needed for https://github.com/rust-lang/rust-analyzer/issues/7434
Blocked on https://github.com/rust-lang/rust-analyzer/pull/18964

### DIFF
--- a/crates/hir-def/src/attr.rs
+++ b/crates/hir-def/src/attr.rs
@@ -84,7 +84,7 @@ impl Attrs {
                 let krate = loc.parent.lookup(db).container.krate;
                 item_tree = loc.id.item_tree(db);
                 let variant = &item_tree[loc.id.value];
-                (FieldParent::Variant(loc.id.value), &variant.fields, krate)
+                (FieldParent::EnumVariant(loc.id.value), &variant.fields, krate)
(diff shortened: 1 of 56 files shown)


Veykril:
See https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/https.3A.2F.2Fgithub.2Ecom.2Frust-lang.2Frust-analyzer.2Fissues.2F7434/near/501477376

The gist is the data queries will turn into the "signatures" of the things (generic params, where clauses, generic args, etc) which for impls means generics, target trait (and its args which mat contain expressions) and the target type (again may contain expressions). All of that is information that isnt necessary when querying the assoc items they are separate concerns. Assoc item lowering is still just part of early name res, we just do it lazily because it doesn't quite participate in the fixpoint (yet, traits will be soon as rust will allow importing trait functions in the future).

The item tree is not really related to this concrete change here.

Veykril:
A follow up is to split out the `VariantData` (fields) from data queries of variants (`struct_data`, `union_data`, `enum_variant_data`), for similar reasons, they aren't part of the signature.

PR #19230, opened 2025-02-26
internal: set up Zig on CI and start using it in rust-analyzer

lnicola:
:upside_down_face:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,8 +17,8 @@ env:
   RUSTUP_MAX_RETRIES: 10
   FETCH_DEPTH: 0 # pull in the tags for the version string
   MACOSX_DEPLOYMENT_TARGET: 13.0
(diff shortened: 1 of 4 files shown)


Veykril (approved):

BenjaminBrienen:
What does this do and what motivated it? Is this just to avoid using C?

lnicola:
When you cross-compile Rust, you still need a target-specific linker, and those aren't always easy to find. In addition, if you build e.g. on Ubuntu 24.04, your binary won't work on older distros. So every couple of years we used to bump the minimum required GLIBC, which caused a number of people to complain that they stopped working on unsupported distros like Debian oldoldstable, or CentOS/EL, which are even older.

`cargo-zigbuild` works as a linker and knows stuff about what's in each version of GLIBC, so we can build every Linux binary on `ubuntu-latest` while targeting the lovely EL8 vintage.

PR #19233, opened 2025-02-26
rust 2024, rust version 1.85

BenjaminBrienen:
- Update workspace to all rust 2024
- Update workspace to all rust-version 1.85
- Fix `cargo test --doc` (mostly ignore things that aren't valid code)
- `cargo fmt`


ChayimFriedman2:
We usually wait a bit before we transition to a new version, as some users are building rust-analyzer with non-bleeding-edge Rust.

BenjaminBrienen:
Ok, no worries. It can just sit open until you decide to take action. I can fix merge conflicts if it says open for a while :)

ChayimFriedman2:
Some notes:
 1. `cargo test` already runs doctests, no need to run them separately (and also this is better done as a separate PR, this one is large enough).
 2. "2024" isn't a very descriptive commit message.
 3. Put the formatting changes in a commit for themselves, they're huge and not really meaningful.
 4. The same for `cargo fix --edition`, then put manual changes as a commit on top.
 5. And another commit to change the edition and MSRV.
 6. Please avoid formatting all TOML files by the way.

BenjaminBrienen:
The doc tests were failing, so I'm surprised to hear that they're already being run

BenjaminBrienen:
I'll redo this

BenjaminBrienen:
I don't understand why you don't use squash and merge practice.

ChayimFriedman2:
Squashing is fine, but we still like meaningful commits. Reviewing huge mostly machine-generated commits with a bit of manual code is very hard.

lnicola:
@BenjaminBrienen we used to use the two borses, for permission management and to serialize the tests, but we switched to merge queues a while ago. These don't let you squash:

image

BenjaminBrienen:
I think bevy uses both merge queues and squashing, but I could be wrong.

Veykril:
Please wait with this until https://github.com/rust-lang/rust-analyzer/pull/18964 has landed

Veykril:
Merge queues allow you to do either merges ,rebases or squashing yes. Why would you want to squash a PR by default though, then you lose the history. Its much better to have the author squash themselves if the history is deemed unnecessary / convoluted

BenjaminBrienen:
I have arguments in favor of squashing all PRs, but I don't want to address it here

BenjaminBrienen:
`cargo test` definitely doesn't catch all the same things that `cargo test --doc` does.

ChayimFriedman2:
https://doc.rust-lang.org/stable/cargo/commands/cargo-test.html#target-selection

> When no target selection options are given, cargo test will build the following targets of the selected packages:
>
>  - ...
>  - doc tests for the lib target
>
> ...
>
> --doc
>        Test only the library’s documentation. This cannot be mixed with other target options.

BenjaminBrienen:
Idk what to tell you. `cargo test` passes, but `cargo test --doc` finds a bunch of errors.

ChayimFriedman2:
Well that sounds like a bug in Cargo then.

PR #19241, opened 2025-02-27
Fix sysroot crate-graph construction not mapping crate-ids for proc-macros

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/19214

Needs backporting to beta


Veykril:
gonna wait on https://github.com/rust-lang/rust-analyzer/pull/19096 and then rebase on top of it

PR #19253, opened 2025-02-28
internal: Migrate `convert_bool_then` to `SyntaxEditor`

ShoyuVanilla:
Part of #18285

### DIFF
--- a/crates/ide-assists/src/handlers/convert_bool_then.rs
+++ b/crates/ide-assists/src/handlers/convert_bool_then.rs
@@ -8,12 +8,13 @@ use ide_db::{
 };
 use itertools::Itertools;
 use syntax::{
-    ast::{self, edit::AstNodeEdit, make, HasArgList},
-    ted, AstNode, SyntaxNode,
+    ast::{self, edit::AstNodeEdit, syntax_factory::SyntaxFactory, HasArgList},
+    syntax_editor::SyntaxEditor,
+    AstNode, SyntaxNode,
 };
 
 use crate::{
-    utils::{invert_boolean_expression_legacy, unwrap_trivial_block},
+    utils::{invert_boolean_expression, unwrap_trivial_block},
     AssistContext, AssistId, AssistKind, Assists,
 };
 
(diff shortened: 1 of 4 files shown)


ShoyuVanilla:
Oh, I accidentally added part of another migration - `convert_closure_to_fn` - I was working on 😅 
I'll continue working on it in a follow-up PR or this PR, if this isn't merged by then

DropDemBits (approved):
I've only looked at the `convert_bool_then`-related changes, but looks good to me & approving those changes.

Waiting on `convert_closure_to_fn` changes to be split out. Surprised that it's using `splice_children` in the first place, but shouldn't be too hard to replace it with the `SyntaxEditor` equivalent.

ShoyuVanilla:
Dropped unfinished `convert_closure_to_fn` and will address it in a new PR

Veykril (approved):

crates/ide-assists/src/handlers/convert_bool_then.rs
```diff
@@ -8,12 +8,13 @@ use ide_db::{
 };
 use itertools::Itertools;
 use syntax::{
-    ast::{self, edit::AstNodeEdit, make, HasArgList},
-    ted, AstNode, SyntaxNode,
+    ast::{self, edit::AstNodeEdit, syntax_factory::SyntaxFactory, HasArgList},
```
Veykril, line 11: This does still use `AstNodeEdit` for indentation, I assume that's mainly to not regress indentation for now? (merging either way)

ShoyuVanilla: Well, currently we have two ways to indent things (except manually indenting all lines in a syntax node):

https://github.com/rust-lang/rust-analyzer/blob/81ff38f53b9a14ac608feb30b21ed42a41d016c6/crates/syntax/src/ast/edit.rs#L119-L133

https://github.com/rust-lang/rust-analyzer/blob/81ff38f53b9a14ac608feb30b21ed42a41d016c6/crates/syntax/src/ast/edit_in_place.rs#L964-L970

and both of them utilizes mutable syntax trees.l

I'll add a non-mutable syntax tree alternative and replace previous usages with it in next PR 😄

Issue #17401, opened 2024-06-12
"Locate child modules", inverse of the "Locate parent modules" feature


Veykril:
We currently have a feature that allows running a command to open the parent module(s) of a file's corresponding modules. We should have the inverse of this, opening a popup with the selection of all direct children modules of the currently open file's module(s).

See https://github.com/rust-lang/rust-analyzer/blob/4af21ffb026c7ec3a97a484ca27b36f703eb5fb1/crates/ide/src/parent_module.rs for the parent module implementation, you should be able to trace how this is used to know what to edit to add a similar feature.

This will need an lsp-extension like https://github.com/veykril/rust-analyzer/blob/4af21ffb026c7ec3a97a484ca27b36f703eb5fb1/crates/rust-analyzer/src/lsp/ext.rs#L389-L395 and vscode glue like https://github.com/rust-lang/rust-analyzer/blob/4af21ffb026c7ec3a97a484ca27b36f703eb5fb1/editors/code/src/lsp_ext.ts#L188-L192

randomicon00:
Hello, I would like to work on it!

geetanshjuneja:
Hi I would like to work on this.

geetanshjuneja:
After reading 'parent_module' code I found that we need nearest ancestor of type `Module` which contains cursor position and `find_node_at_offset` returns these modules. For the Locate child modules feature we want to get the children which are of type `Module`. Am I correct here. If yes then I just have to check `sourcefile.descendants` and filter out modules from this list?

geetanshjuneja:
what would be the output in the following cases?

1. 

//- /lib.rs
$0
mod foo;

is the ans `foo` here?

2. 

//- /lib.rs
mod foo {
    mod bar {
        mod $0baz {}
    }  
}

is the ans `baz` here?

3. 

//- /lib.rs
$0
mod foo {
    mod bar {
        mod baz {}
    }   
}

is the ans `foo` here?

Veykril:
Once you have the `hir::Module` of the current file you can iterate its children via its `children` method

Veykril:
> what would be the output in the following cases?
> 
> 
> //- /lib.rs $0 mod foo;
> 
> is the ans `foo` here?

Yes sounds reasonable

> //- /lib.rs mod foo { mod bar { mod $0baz {} } }
> 
> is the ans `baz` here?
Yes sounds reasonable

> //- /lib.rs $0 mod foo { mod bar { mod baz {} } }
[...]

PR #19255, opened 2025-03-01
Add children modules feature

geetanshjuneja:
Links to #17401

### DIFF
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -943,6 +943,18 @@ pub(crate) fn handle_parent_module(
     Ok(Some(res))
 }
 
+pub(crate) fn handle_children_modules(
+    snap: GlobalStateSnapshot,
+    params: lsp_types::TextDocumentPositionParams,
+) -> anyhow::Result<Option<lsp_types::GotoDefinitionResponse>> {
+    let _p = tracing::info_span!("handle_children_module").entered();
+    // locate children module by semantics
+    let position = try_default!(from_proto::file_position(&snap, params)?);
+    let navs = snap.analysis.children_modules(position)?;
+    let res = to_proto::goto_definition_response(&snap, None, navs)?;
+    Ok(Some(res))
+}
+
 pub(crate) fn handle_runnables(
     snap: GlobalStateSnapshot,
     params: lsp_ext::RunnablesParams,
--- a/crates/ide/src/children_modules.rs
+++ b/crates/ide/src/children_modules.rs
@@ -1,0 +1,116 @@
+use hir::Semantics;
+use ide_db::{FilePosition, RootDatabase};
+use syntax::{
+    algo::find_node_at_offset,
+    ast::{self, AstNode},
+};
+
+use crate::NavigationTarget;
+
+/// This returns `Vec` because a module may be included from several places.
+pub(crate) fn children_modules(db: &RootDatabase, position: FilePosition) -> Vec<NavigationTarget> {
+    let sema = Semantics::new(db);
+    let source_file = sema.parse_guess_edition(position.file_id);
+    // First go to the parent module which contains the cursor
+    let module = find_node_at_offset::<ast::Module>(source_file.syntax(), position.offset);
+
+    match module {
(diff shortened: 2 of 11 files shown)


geetanshjuneja:
@Veykril I have implemented whatever was stated in issue. Looks like I also need need to add `childrenModule` function in [commands.ts](https://github.com/rust-lang/rust-analyzer/blob/master/editors/code/src/commands.ts) and `children_module` handler in [request.rs](https://github.com/rust-lang/rust-analyzer/blob/master/crates/rust-analyzer/src/handlers/request.rs) similar to `parentModule`.

.github/actions/github-release/package-lock.json
Veykril, line 1: Whats up with this file being added

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -920,6 +920,79 @@ pub(crate) fn handle_parent_module(
...
+            let res = vec![LocationLink {
+                origin_selection_range: None,
+                target_uri: cargo_toml_url,
+                target_range: Range::default(),
+                target_selection_range: Range::default(),
+            }]
+            .into();
+            return Ok(Some(res));
+        }
+    }
```
Veykril, line 987: This is the same logic as the parent lookup which isn't what we want here. We'd want to go from Cargo.toml to the respective crates root rust file instead. For simplicity of the PR we can just remove this part for now though

crates/ide/src/children_module.rs
```diff
@@ -0,0 +1,150 @@
...
+    // If cursor is literally on `mod foo`, go to the grandpa.
+    if let Some(m) = &module {
+        if !m
+            .item_list()
+            .is_some_and(|it| it.syntax().text_range().contains_inclusive(position.offset))
+        {
+            cov_mark::hit!(test_resolve_parent_module_on_module_decl);
+            module = m.syntax().ancestors().skip(1).find_map(ast::Module::cast);
+        }
+    }
```
Veykril, line 36: I think we want to show the children of the `mod foo` in this case, so this here should not be necessary

crates/ide/src/children_module.rs
```diff
@@ -0,0 +1,150 @@
...
+                .children()
+                .filter_map(ast::ItemList::cast)
+                .flat_map(|it| it.syntax().children())
+                .filter_map(ast::Module::cast)
+                .flat_map(|m| {
+                    sema.to_def(&m)
+                        .into_iter()
+                        .flat_map(|module| NavigationTarget::from_module_to_decl(db, module))
+                })
+                .collect()
```
Veykril, line 52:
```suggestion
 sema
            .to_def(&module)
            .into_iter()
            .flat_map(|module| module.children(db))
            .map(|module| NavigationTarget::from_module_to_decl(db, module))
            .collect(),
```

crates/ide/src/children_module.rs
```diff
@@ -0,0 +1,150 @@
...
+            source_file
+                .syntax()
+                .children()
+                .filter_map(ast::Module::cast)
+                .flat_map(|m| {
+                    sema.to_def(&m)
+                        .into_iter()
+                        .flat_map(|module| NavigationTarget::from_module_to_decl(db, module))
+                })
+                .collect()
```
Veykril, line 65:
```suggestion
sema
            .file_to_module_defs(position.file_id)
            .flat_map(|module| module.children(db))
            .map(|module| NavigationTarget::from_module_to_decl(db, module))
            .collect(),
```

Veykril (changes requested):
Let's make this plural, that is all `children_module` occurences should be `children_modules`

Veykril (approved):
Thanks!

Veykril:
Will need a rebase + updating the hash in  docs/dev/lsp-extensions.md.

davidbarsky:
> Let's make this plural, that is all `children_module` occurences should be `children_modules`

This might be autocorrect, but I think this should be called `child_modules`?

lnicola:
Hmm, I think a slightly better UX would be to navigate to the submodule implementation (not declaration) when double-clicking on it here on the right side:

image

PR #19267, opened 2025-03-02
Migrate `remove_dbgs` from `ted` to `SyntaxEditor`

ThouCheese:
Hi! I am trying to get started on migrating a couple of assists to the SyntaxEditor machinery. I am struggling ever so slightly with converting things however. One of the problems is that because I prepare all the changes into the SyntaxEditor, and only apply it later, I don't really know where to look when the apply-step results in an error. 

Specifically, the tests are currently failing with `assertion failed: is_ancestor_or_self_of_element(&old, &self.root)`, and with that I am leaving trailing semicolons in place when I am removing dbgs. Any help on this would me greatly appreciated :)


DropDemBits:
`assertion failed: is_ancestor_or_self_of_element` is raised whenever you're trying to replace/delete a node from a different syntax tree. I assume this is happening inside of `replace_nested_dbgs` because `expanded` eventually comes from `input_expressions`, which are newly created syntax trees that aren't part of the original source syntax tree.
https://github.com/ThouCheese/rust-analyzer/blob/2ba82fdd2401d5dc32b90af703394c663525fc42/crates/ide-assists/src/handlers/remove_dbg.rs#L78-L82
Instead, you'll likely need to create a separate `SyntaxEditor` inside of `replace_nested_dbgs`, and then `finish()` it to get the `new_root()` which contains the applied changes.

Regarding the semicolon, it seems that this is supposed to be deleting the entire `ast::ExprStmt` node instead of just the macro expr:
https://github.com/ThouCheese/rust-analyzer/blob/2ba82fdd2401d5dc32b90af703394c663525fc42/crates/ide-assists/src/handlers/remove_dbg.rs#L99-L107

ThouCheese:
I am afraid I am extremely stuck on how to properly handle the SyntaxEditor here. I tried making various granularities of sub-syntax editors that produce and expr, and then feeding that into the next SyntaxEditor, but it keeps crashing with variations on `can't resolve SyntaxNodePtr`.

Issue #19256, opened 2025-03-01
Replace `for` loop with `while let` assist


Veykril:
As the title says, would be nice to have an assist that turns a for loop into the corresponding `while let` construct

PR #19271, opened 2025-03-03
feat: for loop to while let assist

vishruth-thimmaiah:
Closes: #19256

adds an assist that converts a for loop into a while let loop.

### DIFF
--- a/crates/ide-assists/src/handlers/convert_for_to_while_let.rs
+++ b/crates/ide-assists/src/handlers/convert_for_to_while_let.rs
@@ -1,0 +1,422 @@
+use hir::{
+    sym::{self},
+    Name,
+};
+use ide_db::{famous_defs::FamousDefs, syntax_helpers::suggest_name};
+use syntax::{
+    ast::{self, edit::IndentLevel, make, syntax_factory::SyntaxFactory, HasLoopBody},
+    syntax_editor::Position,
+    AstNode,
+};
+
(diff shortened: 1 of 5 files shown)


crates/ide-assists/src/handlers/convert_for_to_while_let.rs
```diff
@@ -0,0 +1,383 @@
...
+                // range expressions need to be parenthesized for the syntax to be correct
+                make::expr_paren(iterable)
+            } else if impls_core_iter(&ctx.sema, &iterable) {
+                // We can't use `iterable` if its immutable, so we create a tmp mutable variable
+                let mut_expr = make::let_stmt(
+                    make::ident_pat(false, true, make::name("tmp")).into(),
+                    None,
+                    Some(iterable),
+                );
+                builder.insert(
```
DropDemBits, line 73: Less of a priority but would you be willing to use `SyntaxEditor`-based edits instead?
[`introduce_named_type_parameter.rs`](https://github.com/rust-lang/rust-analyzer/blob/8d01570b5e812a49daa1f08404269f6ea5dd73a1/crates/ide-assists/src/handlers/introduce_named_type_parameter.rs) has a small example of using `SyntaxEditor` along with a `SourceChangeBuilder`, but there are also other files you can look at for more examples.

DropDemBits (changes requested):

DropDemBits (approved):
Looks good to me, thanks!

PR #19288, opened 2025-03-05
Use `size_of` from the prelude instead of imported

thaliaarchi:
Use `std::mem::{size_of, size_of_val, align_of, align_of_val}` from the prelude instead of importing or qualifying them.

These functions were added to all preludes in Rust 1.80.

### DIFF
--- a/crates/hir-ty/src/consteval/tests.rs
+++ b/crates/hir-ty/src/consteval/tests.rs
@@ -2713,12 +2713,11 @@ fn const_trait_assoc() {
         r#"
     //- minicore: size_of, fn
     //- /a/lib.rs crate:a
-    use core::mem::size_of;
     pub struct S<T>(T);
     impl<T> S<T> {
         pub const X: usize = {
             let k: T;
-            let f = || core::mem::size_of::<T>();
+            let f = || size_of::<T>();
             f()
         };
     }
--- a/crates/test-utils/src/minicore.rs
+++ b/crates/test-utils/src/minicore.rs
@@ -1874,6 +1874,7 @@ pub mod prelude {
             marker::Sized,                           // :sized
             marker::Sync,                            // :sync
             mem::drop,                               // :drop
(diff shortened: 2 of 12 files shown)


crates/hir-ty/src/consteval/tests.rs
```diff
@@ -2713,12 +2713,11 @@ fn const_trait_assoc() {
         r#"
     //- minicore: size_of, fn
     //- /a/lib.rs crate:a
-    use core::mem::size_of;
     pub struct S<T>(T);
     impl<T> S<T> {
         pub const X: usize = {
             let k: T;
-            let f = || core::mem::size_of::<T>();
+            let f = || size_of::<T>();
```
Veykril, line 2720: This is why CI fails. This either needs to be undone or the corresponding prelude changes need to be done to the minicore.rs file.

crates/test-utils/src/minicore.rs
```diff
@@ -1825,7 +1825,7 @@ macro_rules! impl_int {
     ($($t:ty)*) => {
         $(
             impl $t {
-                pub const fn from_ne_bytes(bytes: [u8; mem::size_of::<Self>()]) -> Self {
+                pub const fn from_ne_bytes(bytes: [u8; size_of::<Self>()]) -> Self {
```
Veykril, line 1828: This needs to be undone, this is effectively a core library subset so it doesnt have its own prelude

crates/hir-ty/src/consteval/tests.rs
thaliaarchi: I opted to add `size_of` to the prelude (and add `prelude::rust_2024`). Let me know if you'd rather I'd have kept these uses qualified.

crates/test-utils/src/minicore.rs
thaliaarchi: With `size_of` added to the prelude, this is fine.

PR #19290, opened 2025-03-05
Add warning and debug information when `cargo metadata` fails

aibaars:
The errors are silently dropped elsewhere, which make it really hard to debug issues due to dependency download failures.

### DIFF
--- a/crates/project-model/src/cargo_workspace.rs
+++ b/crates/project-model/src/cargo_workspace.rs
@@ -300,10 +300,9 @@ impl CargoWorkspace {
         if let Ok((_, Some(ref e))) = res {
             tracing::warn!(
                 %cargo_toml,
-                %e,
+                ?e,
                 "`cargo metadata` failed, but retry with `--no-deps` succeeded"
             );
-            tracing::debug!("{e:?}");
         }
         res
     }
--- a/crates/project-model/src/cargo_workspace.rs
+++ b/crates/project-model/src/cargo_workspace.rs
@@ -299,7 +299,9 @@ impl CargoWorkspace {
         );
         if let Ok((_, Some(ref e))) = res {
             tracing::warn!(
-                "`cargo metadata` failed on `{cargo_toml}`, but retry with `--no-deps` succeeded"
+                %cargo_toml,
+                %e,
+                "`cargo metadata` failed, but retry with `--no-deps` succeeded"
             );
             tracing::debug!("{e:?}");
         }
--- a/crates/project-model/src/cargo_workspace.rs
+++ b/crates/project-model/src/cargo_workspace.rs
@@ -288,7 +288,22 @@ impl CargoWorkspace {
         locked: bool,
         progress: &dyn Fn(String),
     ) -> anyhow::Result<(cargo_metadata::Metadata, Option<anyhow::Error>)> {
-        Self::fetch_metadata_(cargo_toml, current_dir, config, sysroot, locked, false, progress)
+        let res = Self::fetch_metadata_(
+            cargo_toml,
+            current_dir,
+            config,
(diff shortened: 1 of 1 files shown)


crates/project-model/src/cargo_workspace.rs
```diff
@@ -288,7 +288,22 @@ impl CargoWorkspace {
...
+            sysroot,
+            locked,
+            false,
+            progress,
+        );
+        if let Ok((_, Some(ref e))) = res {
+            tracing::warn!(
+                "`cargo metadata` failed on `{cargo_toml}`, but retry with `--no-deps` succeeded"
+            );
+            tracing::debug!("{e:?}");
```
Veykril, line 304:
```suggestion
            tracing::warn!(
                %cargo_toml,
                %e,
                "`cargo metadata` failed on `{}`, but retry with `--no-deps` succeeded"
            );
```

aibaars: That looks better, but doesn't print the `stderr` though, while `tracing::debug!("{e:?}");` does. I'd really like to see a short warning with a default log level and a detailed error message explaining the cause at debug/trace level. Not sure that the best way is to achieve that.

crates/project-model/src/cargo_workspace.rs
```diff
@@ -288,7 +288,24 @@ impl CargoWorkspace {
...
+            false,
+            progress,
+        );
+        if let Ok((_, Some(ref e))) = res {
+            tracing::warn!(
+                %cargo_toml,
+                %e,
+                "`cargo metadata` failed, but retry with `--no-deps` succeeded"
+            );
+            tracing::debug!("{e:?}");
```
aibaars, line 306: @Veykril I left the `tracing::debug!("{e:?}");` line in for now. Do you have an idea how to achieve detailed logging in debug mode without having two `tracing::` lines?

crates/project-model/src/cargo_workspace.rs
davidbarsky: this should include the command's `stderr`:

```rust
tracing::warn!(
    %cargo_toml,
    ?e,
    "`cargo metadata` failed, but retry with `--no-deps` succeeded"
);
```

personally, I think including `stderr` is worth the noise, but.

crates/project-model/src/cargo_workspace.rs
davidbarsky: I commented above, but see `tracing`'s documentation on `?` and `%` sigils: https://docs.rs/tracing/latest/tracing/#recording-fields. `tracing` doesn't have per-field levels, so you'll need to have two events or accept the higher verbosity.

crates/project-model/src/cargo_workspace.rs
aibaars: Ah yes, that makes sense, I can change to `?e` . @Veykril what do you think?

crates/project-model/src/cargo_workspace.rs
```diff
@@ -288,7 +288,24 @@ impl CargoWorkspace {
...
+            config,
+            sysroot,
+            locked,
+            false,
+            progress,
+        );
+        if let Ok((_, Some(ref e))) = res {
+            tracing::warn!(
+                %cargo_toml,
+                %e,
```
aibaars, line 303:
```suggestion
                ?e,
```

crates/project-model/src/cargo_workspace.rs
aibaars: Thanks both! I applied @davidbarsky's suggestion and replaced `%e` with `?e` .

Veykril:
Thanks!

PR #19293, opened 2025-03-05
Correct manual installation command for terminal usage

theplasticbrick:
The previous command installed the rust-src component, not rust-analyzer. Changed to match the correct usage on https://rust-analyzer.github.io/book/rust_analyzer_binary.html


lnicola:
Thanks, but that part is actually about the:

## Rust Standard Library

rust-analyzer will attempt to install the standard library source code
automatically. You can also install it manually with `rustup`.

    $ rustup component add rust-src

theplasticbrick:
Ah, that makes more sense. I'll try check twice before submitting next time, sorry about that.

PR #19294, opened 2025-03-05
minor: Show build scripts errors in server status 

### LINKED ISSUE #19184: Error message if build.rs panics
I happen to have an issue, where I would like to check static things in the build.rs such as license validity and so on (I know cargo deny exists, I have further objectives with licenses, it's part of the integration). So when a license can't be validated in my case I simply `panic!`, and there are many more cases for checks to be made in `build.rs`.

The point is that there was no direct indicator that the `build.rs` panicked, I simply noticed because the editor wasn't showing any error, and to actually notice what failed I had to run `cargo build` manually.

The feature request in this case is to add an indicator maybe tied in the `build.rs` itself that says something along the lines of "Project build panicked: <panic reason>", so there is an actual indicative on why there is no completion.

This could also be helpful with `Cargo.toml` errors when moving workspaces and doing that kinds of things

alibektas:
Not the prettiest PR so I want to suggest a few things : ( fixes #19184 )

1. While I was tinkering with our build script runners, I saw that we ignore any Warn level diagnostics but if there are any errors, warnings would also be included in the "problems". Since the PR simply adds showing what `cargo check` outputs as a server status message, users will now also see these warning when there are errors and not see them where there are not. This tells me that we can add a `DiagnosticLevel` guard to filter warnings out.
2. I could better format error messages, since they are not very pleasant to look at.

### DIFF
--- a/crates/rust-analyzer/src/reload.rs
+++ b/crates/rust-analyzer/src/reload.rs
@@ -139,10 +139,10 @@ impl GlobalState {
                 "Proc-macros and/or build scripts have changed and need to be rebuilt.\n\n",
             );
         }
-        if let Err(build_data_err) = self.fetch_build_data_error() {
-            status.health |= lsp_ext::Health::Error;
+        if self.fetch_build_data_error().is_err() {
+            status.health |= lsp_ext::Health::Warning;
             message.push_str("Failed to run build scripts of some packages.\n\n");
-            format_to!(message, "{build_data_err}\n");
+            message.push_str("Please refer to the logs for more details on the errors.");
         }
         if let Some(err) = &self.config_errors {
             status.health |= lsp_ext::Health::Warning;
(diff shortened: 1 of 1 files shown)


crates/rust-analyzer/src/reload.rs
```diff
@@ -139,9 +139,10 @@ impl GlobalState {
                 "Proc-macros and/or build scripts have changed and need to be rebuilt.\n\n",
             );
         }
-        if self.fetch_build_data_error().is_err() {
-            status.health |= lsp_ext::Health::Warning;
+        if let Err(build_data_err) = self.fetch_build_data_error() {
+            status.health |= lsp_ext::Health::Error;
```
Veykril, line 143: This should stay a warning, even if build scripts fail we can still load the project.

crates/rust-analyzer/src/reload.rs
```diff
@@ -139,9 +139,10 @@ impl GlobalState {
...
             message.push_str("Failed to run build scripts of some packages.\n\n");
+            format_to!(message, "{build_data_err}\n");
```
Veykril, line 145: The issue with this is I believe that we have limited real estate in the hover box for the status message and so depending on the error it might not fit. We are logging the errror though so maybe we should instruct the user to look at the logs

Veykril (approved):

Issue #19296, opened 2025-03-06
Modify the score threshold for relevance judgment in the is_relevant() function


Natural-selection1:
Six months ago, we modified the scoring threshold from `0` to `!0 / 2` to address the issue where `u32` cannot be negative:  
https://github.com/rust-lang/rust-analyzer/blob/d426cbee769d759b3ba5ec175e7fc7cc1f4dd900/crates/ide-completion/src/item.rs#L261-L263  

---  
However, we did not update the checking criterion in `is_relevant()` ~~(even though the original `> 0` check was incorrect for `u32` types)~~:  
https://github.com/rust-lang/rust-analyzer/blob/d426cbee769d759b3ba5ec175e7fc7cc1f4dd900/crates/ide-completion/src/item.rs#L352-L354  

---  
Should we make the following change?  
```diff  
pub fn is_relevant(&self) -> bool {  
-    self.score() > 0  
+    self.score() > (!0 / 2)  
}  
```  

---  
Otherwise, checks like the one below will forever remain unrelated to the truth value of `is_relevant()`:  
https://github.com/rust-lang/rust-analyzer/blob/d426cbee769d759b3ba5ec175e7fc7cc1f4dd900/crates/rust-analyzer/src/lsp/to_proto.rs#L443-L460  

---

PR #19297, opened 2025-03-06
Adjust relevance scoring threshold to consistent with existing implem…

Natural-selection1:
…entations

close: #19296


Veykril (approved):
Thanks!

PR #19305, opened 2025-03-06
internal: vendor `query-group-macro` in rust-analyzer

davidbarsky:
Dependency of https://github.com/rust-lang/rust-analyzer/pull/18964. This shouldn't land until we get a Salsa release out, which depends on fixing the `powerpc` build failures.

### DIFF
--- a/crates/query-group-macro/src/lib.rs
+++ b/crates/query-group-macro/src/lib.rs
@@ -1,0 +1,437 @@
+//! A macro that mimics the old Salsa-style `#[query_group]` macro.
+
+use core::fmt;
+use std::vec;
+
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use queries::{
+    GeneratedInputStruct, InputQuery, InputSetter, InputSetterWithDurability, Intern, Lookup,
+    Queries, SetterKind, TrackedQuery, Transparent,
+};
+use quote::{format_ident, quote, ToTokens};
+use syn::spanned::Spanned;
+use syn::visit_mut::VisitMut;
+use syn::{parse_quote, Attribute, FnArg, ItemTrait, Path, TraitItem, TraitItemFn};
+
+mod queries;
+
+#[proc_macro_attribute]
+pub fn query_group(args: TokenStream, input: TokenStream) -> TokenStream {
+    match query_group_impl(args, input.clone()) {
+        Ok(tokens) => tokens,
+        Err(e) => token_stream_with_error(input, e),
+    }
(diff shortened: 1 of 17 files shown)


davidbarsky:
I'll fix that unused warning tomorrow, but I feel like we should hold off on landing this until on Monday, at the easliest.

crates/query-group-macro/src/lib.rs
```diff
@@ -0,0 +1,492 @@
...
+                            interned_struct_path,
+                        };
+                        method.prepare_signature();
+
+                        lookup_signatures
+                            .push(TraitItem::Fn(make_trait_method(method.signature.clone())));
+                        lookup_methods.push(method);
+                    }
+                    // tracked function without *any* invoke.
+                    (QueryKind::Tracked, None) => {
```
lnicola, line 294: Can we fold the two cases for `Tracked`, `TrackedWithSalsaStruct` and `Transparent` like this?

```suggestion
                    (QueryKind::Tracked, invoke) => {
```

davidbarsky:
I realized (and fixed, locally) why we were getting a dead code warning: the `crates/query-group-macro/tests/logger_db.rs` was now being compiled as its own test rather than its own module. I assume this has to do with this crate being migrated from its own, standalone thing to being part of a larger Cargo workspace.

crates/query-group-macro/src/lib.rs
davidbarsky: We can on `QueryKind::Tracked` and `QueryKind::Transparent`. It might be possible to extend this to `TrackedWithSalsaStruct`, but we don't know _at the macro usage site_ whether the parameter is a Salsa struct or not. Could be possible if we attached an annotation to the parameter itself, but I don't think it's a particularly high priority for what is largely a migration aide.

davidbarsky:
Alright, flattened the match expression and fixed the dead code warning. I'd like to squash those 3 commits into one before landing this, however.

xtask/src/tidy.rs
```diff
@@ -147,6 +147,7 @@ MIT/Apache-2.0
 MPL-2.0
 Unlicense OR MIT
 Unlicense/MIT
+Zlib
```
davidbarsky, line 150: This is coming from `foldhash`, which was introduced in [hashbrown 0.15](https://github.com/rust-lang/hashbrown/blob/master/CHANGELOG.md#v0150---2024-10-01).

Veykril (approved):

PR #19306, opened 2025-03-06
fix: Renaming both struct and function with the same name

### LINKED ISSUE #18892: Renaming a struct that has a function with the same name will break code:

asuto15:
fixes #18892.

## Note
Although the original issue did not specify whether to rename both definitions or only one, since the original code exports both under the same name, it is preferable for rust-analyzer’s behavior to rename both.
This implementation therefore renames both the struct and the function, ensuring consistency.

## Important
This implementation is limited to handling struct/function pairs. It does not cover other combinations (e.g. enum variants with associated functions, constants paired with functions, or trait methods).
If needed, separate issues should be created to address those other cases individually.

### DIFF
--- a/crates/ide/src/rename.rs
+++ b/crates/ide/src/rename.rs
@@ -3276,7 +3276,7 @@ mod foo {
         size: usize,
     }
 
-    pub fn bar() -> bar {
+    pub fn bar(size: usize) -> bar {
         bar { size: 0 }
     }
 }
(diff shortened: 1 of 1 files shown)


Veykril (changes requested):
This seems like the wrong way to tackle this. Specifically we need to verify that the references we want to change are not ambigious, and in the case they are, disambiguate them. That is for imports at least, we need to split out the ambigious import into something unambigious (so for example instead of changing the identifier, just add an import for the new name). This probably requires our semantics path resolver to be namespace aware which is a still a big task that has to be done in general. So this might not be that actionable yet

PR #19307, opened 2025-03-07
Fix logical error in PR #19297

Natural-selection1:
In PR #19297, we overlooked the type inconsistency caused by the new scope for `!0/2` (which then led to logical errors)

---
The rationale is as follows:  
image

---

Although we modified the threshold expression in PR #19297:
```diff
    pub fn is_relevant(&self) -> bool {
-        self.score() > 0
+        self.score() > (!0 / 2)
    }
```
We neglected some type-related issues within a new function scope, which resulted in only modifying the "apparent value" while the actual value remained 0 (as shown in the image above)

---
The reason I chose the first fix option is that it expresses the intent more clearly and avoids ambiguity

### DIFF
--- a/crates/ide-completion/src/item.rs
+++ b/crates/ide-completion/src/item.rs
@@ -252,14 +252,14 @@ impl CompletionRelevance {
     /// Provides a relevance score. Higher values are more relevant.
     ///
     /// The absolute value of the relevance score is not meaningful, for
-    /// example a value of 0 doesn't mean "not relevant", rather
+    /// example a value of (!(0 as u32) / 2) doesn't mean "not relevant", rather
     /// it means "least relevant". The score value should only be used
     /// for relative ordering.
     ///
     /// See is_relevant if you need to make some judgement about score
     /// in an absolute sense.
     pub fn score(self) -> u32 {
-        let mut score = !0 / 2;
+        let mut score = !(0 as u32) / 2;
         let CompletionRelevance {
             exact_name_match,
             type_match,
@@ -350,7 +350,7 @@ impl CompletionRelevance {
     /// some threshold such that we think it is especially likely
     /// to be relevant.
     pub fn is_relevant(&self) -> bool {
-        self.score() > (!0 / 2)
+        self.score() > !(0 as u32) / 2
     }
 }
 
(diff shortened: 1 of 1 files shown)


lnicola:
Bah. You can do `0u32` but let's use `u32::MAX / 2`.

Natural-selection1:
> Bah. You can do `0u32` but let's use `u32::MAX / 2`.呃。你可以做 `0u32` 但让我们用 `u32::MAX / 2` 。

Thanks for your advice, I'm still new to rust

crates/ide-completion/src/item.rs
```diff
@@ -252,14 +252,14 @@ impl CompletionRelevance {
...
-    /// example a value of 0 doesn't mean "not relevant", rather
+    /// example a value of (!(0 as u32) / 2) doesn't mean "not relevant", rather
     /// it means "least relevant". The score value should only be used
     /// for relative ordering.
     ///
     /// See is_relevant if you need to make some judgement about score
     /// in an absolute sense.
     pub fn score(self) -> u32 {
-        let mut score = !0 / 2;
+        let mut score = !(0 as u32) / 2;
```
Veykril, line 262: Let's move this starting score into a constant to prevent the same issue (that you recently fixed) from happening again
```rust
const BASE_SCORE: u32 = !0u32 / 2;
```

lnicola: `const BASE_SCORE: u32 = u32::MAX / 2;` :upside_down_face:

Natural-selection1: > `const BASE_SCORE: u32 = u32::MAX / 2;` 🙃

@Veykril I think this maybe better?

Natural-selection1: 我不常在rust中声明常量, 我该声明在哪里?
```diff
impl CompletionRelevance {
    /// ...
+   const BASE_SCORE: u32 = u32::MAX / 2;
+
    pub fn score(self) -> u32 {
```
这样可以么?

Veykril: Yes an associated const should be good here

Veykril (approved):
Thanks!

Natural-selection1:
Thank all of you very much. I have learned a lot from the recent issues and PRs. Thank you for your patience and suggestions
❤️

PR #19308, opened 2025-03-07
Move loaded project MSRV back to 1.78, show notification for the warning

Veykril:
We shouldve probably rolled that back from the start when we noticed this


lnicola:
I think it's fine either way, but most (?) of the reports were from users who didn't mean to use an old toolchain.

Veykril:
Yes but I don't think we have to be too aggressive with this here (likewise for some decisions in the past), the code here is not really a maintenance burden.

Veykril:
I also want to pop a notification for our MSRV check, as people do not seem to notice the warning otherwise and then bumping aggressively isn't that great.

PR #19311, opened 2025-03-07
Log build script error output in `load_cargo::load_workspace_at`

aibaars:
@Veykril This pull request adds some logging for build script error output. I'm not sure this is the best place for the log statement. I picked the `load_workspace_at` because it's the place where the `ProjectWorkspace` and the error messages it may contain is dropped. The other functions have access to the `ProjectWorkspace` so calling application code can handle the errors itself if it wishes to do so. If you want to always log the errors then perhaps the `ProjectWorkspace::set_build_scripts()` method would be a better place for the log statement.

This log output would have been really helpful for figuring out some  undefined `OUT_DIR` error messages I was debugging earlier, so I was hoping it can be logged to make things easier in the future ;-)


lnicola:
It feels like this is useful enough to bump to a warning?

Veykril:
Agree

PR #19315, opened 2025-03-07
fix(codegen): do not generate docs with `--check`

snprajwal:
`cargo test -p xtask` regenerates bindings with xtask. If any of these changes are relevant to a PR, they should a part of the PR. This change causes the CI to fail if a commit does not include the corresponding changes to generated files.

EDIT: The command already panics if any non-doc files are modified. This patch disables doc file generation with `--check`.

### DIFF
--- a/xtask/src/codegen/diagnostics_docs.rs
+++ b/xtask/src/codegen/diagnostics_docs.rs
@@ -10,13 +10,15 @@ use crate::{
 
 pub(crate) fn generate(check: bool) {
     let diagnostics = Diagnostic::collect().unwrap();
-    if !check {
-        let contents =
-            diagnostics.into_iter().map(|it| it.to_string()).collect::<Vec<_>>().join("\n\n");
-        let contents = add_preamble(crate::flags::CodegenType::DiagnosticsDocs, contents);
-        let dst = project_root().join("docs/book/src/diagnostics_generated.md");
-        fs::write(dst, contents).unwrap();
+    // Do not generate docs when run with `--check`
+    if check {
+        return;
     }
+    let contents =
+        diagnostics.into_iter().map(|it| it.to_string()).collect::<Vec<_>>().join("\n\n");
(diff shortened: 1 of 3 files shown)


snprajwal:
Just saw #19334 - I've made the changes to not generate the ignored files when running with `--check`.

xtask/src/codegen/diagnostics_docs.rs
```diff
@@ -9,14 +9,16 @@ use crate::{
...
+    if check {
+        return;
     }
+
+    let diagnostics = Diagnostic::collect().unwrap();
+    let contents =
+        diagnostics.into_iter().map(|it| it.to_string()).collect::<Vec<_>>().join("\n\n");
+    let contents = add_preamble(crate::flags::CodegenType::DiagnosticsDocs, contents);
+    let dst = project_root().join("docs/book/src/diagnostics_generated.md");
+    fs::write(dst, contents).unwrap();
```
Veykril, line 21: Why this change? `let diagnostics = Diagnostic::collect().unwrap();` should run even if a check is requested

xtask/src/codegen/feature_docs.rs
```diff
@@ -8,7 +8,11 @@ use crate::{
     util::list_rust_files,
 };
 
-pub(crate) fn generate(_check: bool) {
+pub(crate) fn generate(check: bool) {
+    if check {
+        return;
+    }
```
Veykril, line 14: likewise `Feature::collect().unwrap()` should at least run with check enabled as it does some assertions

xtask/src/codegen/diagnostics_docs.rs
snprajwal: Since the value was used only for the generation, I assumed we can ignore it when returning early. Fixed.

Veykril (approved):

Veykril:
Thanks!

PR #19324, opened 2025-03-09
fix: Prevent wrong invocations of `needs_parens_in` with non-ancestral "parent"s

### LINKED ISSUE #19323: inline veriable assist changes semantics (forgets to add parantheses)
**rust-analyzer version**: rust-analyzer version: 0.3.2330-standalone (02862f5d52 2025-03-03)

**rustc version**: rustc 1.85.0 (4d91de4e4 2025-02-17)

**editor or extension**: Code OSS (vscode), extension version: 0.3.2330

**relevant settings**: None

```rust
let a = 123 < 456;
let b = !a;
```

Right click `a` (does not matter which one) and choose "Inline variable".
The result should be `!(123 < 456)`, but actually is `!123 < 456`, which is equivalent to `(!123) < 456`.

ShoyuVanilla:
Fixes #19323

This also migrates `inline_local_variable` to `SyntaxEditor` as a part of #18285

### DIFF
--- a/crates/ide-assists/src/handlers/apply_demorgan.rs
+++ b/crates/ide-assists/src/handlers/apply_demorgan.rs
@@ -128,7 +128,9 @@ pub(crate) fn apply_demorgan(acc: &mut Assists, ctx: &AssistContext<'_>) -> Opti
                     let parent = neg_expr.syntax().parent();
                     editor = builder.make_editor(neg_expr.syntax());
 
-                    if parent.is_some_and(|parent| demorganed.needs_parens_in(&parent)) {
+                    if parent.is_some_and(|parent| {
+                        demorganed.needs_parens_in_place_of(&parent, neg_expr.syntax())
+                    }) {
                         cov_mark::hit!(demorgan_keep_parens_for_op_precedence2);
                         editor.replace(neg_expr.syntax(), make.expr_paren(demorganed).syntax());
                     } else {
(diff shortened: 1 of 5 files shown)


crates/ide-assists/src/handlers/apply_demorgan.rs
```diff
@@ -392,15 +394,19 @@ fn f() { !(S <= S || S < S) }
...
 
     #[test]
-    fn demorgan_removes_pars_in_eq_precedence() {
+    fn demorgan_keeps_pars_in_eq_precedence() {
         check_assist(
             apply_demorgan,
             "fn() { let x = a && !(!b |$0| !c); }",
-            "fn() { let x = a && b && c; }",
+            "fn() { let x = a && (b && c); }",
         )
```
ShoyuVanilla, line 410: These test chages might be a regression, but I think that the changed ones are more consistent with other test cases

crates/ide-assists/src/handlers/inline_local_variable.rs
```diff
@@ -73,40 +57,60 @@ pub(crate) fn inline_local_variable(acc: &mut Assists, ctx: &AssistContext<'_>)
...
-            let usage_parent_option = usage_node.and_then(|it| it.parent());
+            let usage_parent_option = usage_node.as_ref().and_then(|it| it.parent());
             let usage_parent = match usage_parent_option {
                 Some(u) => u,
-                None => return Some((range, name_ref, false)),
+                None => return Some((name_ref, false)),
             };
-            Some((range, name_ref, initializer_expr.needs_parens_in(&usage_parent)))
+            let should_wrap = initializer_expr
+                .needs_parens_in_place_of(&usage_parent, usage_node.as_ref().unwrap());
```
ShoyuVanilla, line 66:
```rust
let a = 123 < 456;
let b = !a;
```
In the above code, `initializer_expr` is `123 < 456` and `usage_parent` is `let b = !a`. So, inside `needs_parens_in`, it assumes that `initializer_expr` "comes before parent" because its text range is earlier, and decides that it needs no parentheses

Veykril (approved):

PR #19328, opened 2025-03-10
internal: Run proc-macro server tests as separate CI job

Veykril:


.github/workflows/ci.yaml
```diff
@@ -86,16 +113,16 @@ jobs:
...
       # It's faster to `test` before `build` ¯\_(ツ)_/¯
       - name: Compile (rust-analyzer)
         if: matrix.os == 'ubuntu-latest'
-        run: cargo build --quiet ${{ env.USE_SYSROOT_ABI }}
+        run: cargo build --quiet
 
       - name: Test
         if: matrix.os == 'ubuntu-latest' || matrix.os == 'windows-latest' || github.event_name == 'push'
-        run: cargo test ${{ env.USE_SYSROOT_ABI }} -- --nocapture --quiet
+        run: cargo test -- --quiet
```
Veykril, line 125: I dropped the `--nocapture` here, I am not sure what  the benefit for us here is to run CI with it but now that our doc tests are being tested this adds noise for each compilefail test

BenjaminBrienen (approved):

PR #19333, opened 2025-03-10
fix: Fix diagnostics being cleared right after being received

alexkirsz:
cc @Veykril [this predicate](https://github.com/rust-lang/rust-analyzer/commit/5e18ad0#diff-8028b5acd2856fbdc72fafd05032eb259143fc51f4eecc6d1d87b1022475f857R417-R418) `self.diagnostics_received != DiagnosticsReceived::YesAndClearedForAll` will always evaluate to `true` as you're setting `self.diagnostics_received` to `DiagnosticsReceived::Yes` just before.

This breaks most Clippy diagnostics from displaying in our codebase.

### DIFF
--- a/crates/rust-analyzer/src/flycheck.rs
+++ b/crates/rust-analyzer/src/flycheck.rs
@@ -401,7 +401,9 @@ impl FlycheckActor {
                             package_id = package_id.as_ref().map(|it| &it.repr),
                             "diagnostic received"
                         );
-                        self.diagnostics_received = DiagnosticsReceived::Yes;
+                        if self.diagnostics_received == DiagnosticsReceived::No {
+                            self.diagnostics_received = DiagnosticsReceived::Yes;
+                        }
                         if let Some(package_id) = &package_id {
                             if self.diagnostics_cleared_for.insert(package_id.clone()) {
                                 tracing::trace!(


crates/rust-analyzer/src/flycheck.rs
```diff
@@ -401,7 +396,7 @@ impl FlycheckActor {
                             package_id = package_id.as_ref().map(|it| &it.repr),
                             "diagnostic received"
                         );
-                        self.diagnostics_received = DiagnosticsReceived::Yes;
```
Veykril, line 404: Oh, this line is wrong, that should be
```rust
if self.diagnostics_received == DiagnosticsReceived::No {
    self.diagnostics_received = DiagnosticsReceived::Yes;
}
```

alexkirsz: If we want to preserve the previous behavior, I believe this line should actually be:

```rust
                        self.diagnostics_received |= DiagnosticsReceived::Yes;
```

Where:

```rust
DiagnosticsReceived::Yes | DiagnosticsReceived::YesAndClearedForAll == DiagnosticsReceived::YesAndClearedForAll
DiagnosticsReceived::No  | DiagnosticsReceived::Yes == DiagnosticsReceived::Yes
DiagnosticsReceived::No  | DiagnosticsReceived::YesAndClearedForAll == DiagnosticsReceived::YesAndClearedForAll
```
 
 WDYT?

Veykril (approved):
Thanks!

PR #19334, opened 2025-03-10
Remove generated md book files

Veykril:


davidbarsky:
Wait, why this change?

Veykril:
CI does not require them to be up to date anymore, so they tend to go stale / some PRs randomly update them even when they are unrelated due to the codegen having been run. Additionally these are already in our `.gitignore`, the files themselves were just never removed.

Issue #19289, opened 2025-03-05
Missing ambiguity error for impl trait 


Veykril:
While we are here, this one can also be triggered with `impl` syntax
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d43383743ffef4b512c74809eb38bbca
```rs
fn f(_: &impl 'a + Sized) {}
```

_Originally posted by @Veykril in https://github.com/rust-lang/rust-analyzer/pull/19265#discussion_r1980895651_

PR #19347, opened 2025-03-13
Add diagnostic for missing ambiguity error for impl trait

Shourya742:
closes: #19289


Veykril (dismissed):
We can deduplicate this a bit between `validate_trait_object_ty` and `validate_impl_object_ty` The only difference here really is the error message

crates/syntax/src/validation.rs
```diff
@@ -339,35 +337,73 @@ fn validate_trait_object_fn_ptr_ret_ty(ty: ast::FnPtrType, errors: &mut Vec<Synt
...
+    check_for_and_token: bool,
+) -> Option<SyntaxError> {
+    let bounds: Vec<_> = bounds.collect();
+    let bounds_count = bounds.len();
+
+    if bounds.is_empty()
+        || bounds.iter().all(|b| matches!(b.kind(), ast::TypeBoundKind::Lifetime(_)))
+    {
+        return Some(SyntaxError::new(error_no_trait, ty_syntax_range));
+    }
```
Veykril, line 354: we definitely don't want to allocate here so
```suggestion
    if !bounds.iter().any(|b| !matches!(b.kind(), ast::TypeBoundKind::Lifetime(_)))
    {
        return Some(SyntaxError::new(error_no_trait, ty_syntax_range));
    }
    let bounds_count = bounds.len();
```

Veykril (commented):
The new code does look rather unwieldy and illegible. Let's just duplicate this after all to make it more readable, its not that much code anyways.

PR #19348, opened 2025-03-13
Add text edit support for return type hints on non-block body closures

jnyfah:
This PR fixes the FIXME in `closure_ret` by adding text edit functionality for closures with non-block bodies. When a user activates the return type hint for a closure without a block body (e.g., `|x| x + 1`), the edit now properly inserts both the return type annotation and the necessary braces around the body expression. This makes the return type hint feature work consistently for all types of closures.

### DIFF
--- a/crates/ide/src/inlay_hints/closure_ret.rs
+++ b/crates/ide/src/inlay_hints/closure_ret.rs
@@ -61,10 +61,9 @@ pub(super) fn hints(
             if arrow.is_none() { " -> " } else { "" },
         )
     } else {
-        let body = closure.body()?;
-        let body_range = body.syntax().text_range();
-
         Some(config.lazy_text_edit(|| {
+            let body = closure.body();
+            let body_range = body.expect("Closure must have a body").syntax().text_range();
             let mut builder = TextEdit::builder();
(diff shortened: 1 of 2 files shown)


crates/ide/src/inlay_hints/closure_ret.rs
```diff
@@ -62,7 +61,27 @@ pub(super) fn hints(
             if arrow.is_none() { " -> " } else { "" },
         )
     } else {
-        None
+        Some(config.lazy_text_edit(|| {
+            let body = closure.body();
+            let body_range = body.expect("Closure must have a body").syntax().text_range();
```
Veykril, line 66: Sorry I didn't mean to unwrap it, more that we should not generate a text edit if the body is missing 😅

I also just realized the `closure` in scope here is the one that is descended into macros. We need to use the one before `descend_node_into_attributes` or the offsets will be wrong (due to us considering the macro expanded note)

jnyfah: Ohh okay i get it now

Veykril (approved):

Veykril:
Thanks!

PR #19352, opened 2025-03-13
generate missing param docs

newcomertv:
Generates doc comments for function parameters if not already present.

This follows the way how most of rustc does it since there is no official style guideline but doc comments aren't allowed directly above parameters.

Definitely can be improved upon. Just wondering if there is general interest in such a feature and what would be the requirements.


Veykril:
Closing this as discussed in #19353

PR #19362, opened 2025-03-15
Speed up resolving a "Generate delegate method" assist

### LINKED ISSUE #19322: Resolving a "Generate delegate method" takes too long
**rust-analyzer version**:  rust-analyzer 0.4.2261-standalone

**rustc version**:  rustc 1.84.1 (e71f9a9a9 2025-01-27)

**editor or extension**:  Eglot

**relevant settings**: 

```rust
struct Foo {
    bar:  Vec<bool>,
}

fn main() {
    println!("hello");
}
```

Requesting a CodeAction for the word `Vec` in the example above results in 185 assists in the "Generate delegate methods" action group.  After the user selects one from the list, the LSP client requests to resolve the selected CodeAction.  However, rust-analyzer generates edits for the all the 185 assists, which in my case takes more than 4 seconds. 

I work on a feature outline [here](https://github.com/rust-lang/rust-analyzer/issues/19090#issuecomment-2644742013), so in reality rust-analyzer might answer the resolve request faster, it is sub-optimal nevertheless. 

The inefficiency caused by the fact that the `AssistId`s of all the 185 assists are the same, therefore all of them are resolved during a resolve request.

I am able to bring the answer time down to around 200 milliseconds by generating a different index for each "generate delegate" assists.  This requires extending `struct AssisId` like this:
```patch
diff --git a/crates/ide-db/src/assists.rs b/crates/ide-db/src/assists.rs
index 1c40685ebb..14f3280bb5 100644
--- a/crates/ide-db/src/assists.rs
+++ b/crates/ide-db/src/assists.rs
@@ -104,8 +104,8 @@ impl FromStr for AssistKind {
 
 /// Unique identifier of the assist, should not be shown to the user
 /// directly.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct AssistId(pub &'static str, pub AssistKind);
+pub struct AssistId(pub &'static str, pub AssistKind, pub Option<usize>);
  
 /// A way to control how many assist to resolve during the assist resolution.
 /// When an assist is resolved, its edits are calculated that might be costly to always do by default.
```
Unfortunately, the complete patch quite large because a `None` must be appended to most of the places where an AssistId is created.  Nevertheless, I can create a pull request if you think this is a good approach to fix this performance bug.

Thank you.

nemethf:
To advance the discussion of #19322 this pull request implements the solution outlined there.  The PR is split into two parts.  The first one contains the core idea, while the boring second part makes sure rust-analyzer compiles after the extension of struct `AssistId`.

---
Part 1

Fix #19322

Sometimes there are 185 "Generate delegate" assists with the same
assist_id and asssist_kind.  This commit introduces and additional
differentiator: assist_subtype.  Therefore, when the LSP client sends
an assist resolve request, rust-analyzer only need to compute edits
for a single assist instead of 185.

---
Part 2

Make it compile by adding a `None` subtype to rest of the AssistId
instantiations.

### DIFF
--- a/crates/ide-assists/src/handlers/add_braces.rs
+++ b/crates/ide-assists/src/handlers/add_braces.rs
@@ -3,7 +3,7 @@ use syntax::{
     ast::{self, edit_in_place::Indent, syntax_factory::SyntaxFactory},
 };
 
-use crate::{AssistContext, AssistId, AssistKind, Assists};
+use crate::{AssistContext, AssistId, Assists};
 
 // Assist: add_braces
 //
@@ -32,7 +32,7 @@ pub(crate) fn add_braces(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<(
     let (expr_type, expr) = get_replacement_node(ctx)?;
 
     acc.add(
-        AssistId("add_braces", AssistKind::RefactorRewrite),
+        AssistId::refactor_rewrite("add_braces"),
         match expr_type {
             ParentType::ClosureExpr => "Add braces to closure body",
             ParentType::MatchArmExpr => "Add braces to arm expression",
--- a/crates/ide-assists/src/handlers/add_explicit_enum_discriminant.rs
+++ b/crates/ide-assists/src/handlers/add_explicit_enum_discriminant.rs
@@ -1,9 +1,5 @@
 use hir::Semantics;
-use ide_db::{
-    RootDatabase,
-    assists::{AssistId, AssistKind},
-    source_change::SourceChangeBuilder,
-};
+use ide_db::{RootDatabase, assists::AssistId, source_change::SourceChangeBuilder};
 use syntax::{AstNode, ast};
 
 use crate::{AssistContext, Assists};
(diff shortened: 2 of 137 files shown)


crates/ide-assists/src/handlers/add_braces.rs
```diff
@@ -32,7 +32,7 @@ pub(crate) fn add_braces(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<(
     let (expr_type, expr) = get_replacement_node(ctx)?;
 
     acc.add(
-        AssistId("add_braces", AssistKind::RefactorRewrite),
+        AssistId("add_braces", AssistKind::RefactorRewrite, None),
```
Veykril, line 35: It would likely be nicer to have a constructor function for the `None` case here, something like
```suggestion
        assist_id("add_braces", AssistKind::RefactorRewrite),
```
or 
```suggestion
        AssistId::refactor_rewrite(add_braces"),
```
To keep the group one long but the common case short?

Veykril (commented):
Overall this sounds reasonable to me

crates/ide-assists/src/handlers/add_braces.rs
nemethf: I force-pushed a new version with the second option.  I chose it over the first one because it resulted in more concise code.  Thanks for the suggestion.

Veykril (approved):

nemethf:
I've resolved the merge conflict.   Is there anything more that I need to do here?  Thanks.

Veykril:
Nope, should be good to go
Thanks!

PR #19363, opened 2025-03-15
display varargs in completion detail

euclio:
Currently the varargs are ignored, making the function signatures misleading.

### DIFF
--- a/crates/ide-completion/src/render/function.rs
+++ b/crates/ide-completion/src/render/function.rs
@@ -347,7 +347,7 @@ fn detail_full(ctx: &CompletionContext<'_>, func: hir::Function) -> String {
 }
 
 fn params_display(ctx: &CompletionContext<'_>, func: hir::Function) -> String {
-    if let Some(self_param) = func.self_param(ctx.db) {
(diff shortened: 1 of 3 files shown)


crates/ide-completion/src/render/function.rs
```diff
@@ -324,7 +324,12 @@ fn detail(ctx: &CompletionContext<'_>, func: hir::Function) -> String {
         format_to!(detail, "unsafe ");
     }
 
-    format_to!(detail, "fn({})", params_display(ctx, func));
+    format_to!(
+        detail,
+        "fn({}{})",
+        params_display(ctx, func),
+        if func.is_varargs(ctx.db) { ", ..." } else { "" }
```
Veykril, line 331: Let's add that change to `params_display`. Might even be better to change this to not have `params_display` allocate a separate string here in the first place, that is use `format_to` for `params_display` (instead of `format!`).

euclio: Done. Passing a mutable string to `params_display` seemed a bit messy to me compared to the gains, so I left as-is.

Veykril (approved):
Thanks!

PR #19374, opened 2025-03-16
minor: Use cargo nextest on CI

### LINKED ISSUE #19365: Look into using `cargo nextest` for CI
Used it earlier today in https://github.com/salsa-rs/salsa/pull/758 which sped things up vastly

lnicola:
Closes #19365

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -105,6 +105,8 @@ jobs:
       - name: Cache Dependencies
         uses: Swatinem/rust-cache@9bdad043e88c75890e36ad3bbc8d27f0090dd609
 
+      - uses: taiki-e/install-action@nextest
+
       - name: Bump opt-level
         if: matrix.os == 'ubuntu-latest'
         run: sed -i '/\[profile.dev]/a opt-level=1' Cargo.toml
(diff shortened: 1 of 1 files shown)


lnicola:
This takes about 28 vs. 39 seconds on my computer, but I'd expect the difference to be smaller on CI.

lnicola:
Right, there's practically no difference between https://github.com/rust-lang/rust-analyzer/actions/runs/13882176057/job/38842137079?pr=19374 and https://github.com/rust-lang/rust-analyzer/actions/runs/13876440422/job/38829335936.

Veykril:
test time did go down from 37 to 18 but it took a minute more to compile

lnicola:
No? I think went from 37 to 36 or 35? It gets a couple of seconds faster, and I suspect it's not worth bumping the optimization level.

Veykril:
Oh mustve misread a column, ye i guess the gains here aren't really there. Unfortunate

Veykril:
It does seem to save a minute on windows though if I see this right?

lnicola:
Not sure, the compilation was also faster (didn't check if it was before or after removing salsa). But sure, let's give it a try.

lnicola:
It won't help directly with the CI times, though, because Linux is the slowest. We could move `analysis-stats` to the MacOS runner instead, it seems to be faster. Windows runs Clippy at the moment.

PR #19383, opened 2025-03-17
minor: sync from downstream

lnicola:


lnicola:
Oh no.

Veykril:
Oh, that thing is insta-stable and was immediately used in there ... that is not good

PR #19391, opened 2025-03-18
chore: Clean up proc-macro-srv CI job

Veykril:
Noticed our CI does not restore caches .. ever? So I hope this "just" fixes it magically but I doubt it


lnicola:
No, it actually does in some cases, see https://github.com/rust-lang/rust-analyzer/actions/runs/13905452610/job/38907327239. I thought it busts the cache after a failed build, but maybe that's not the case.

Veykril:
😕

Issue #19357, opened 2025-03-14
`&raw const`/`&raw mut` is not syntax-highlighted in all files


RalfJung:
When opening a file that is not in the current project (i.e. RA cannot do its full semantic analysis), syntax highlighting of `&raw` is broken:

![Image](https://github.com/user-attachments/assets/717c72f8-0690-48e0-9d2c-1e2882aaee86)

Note that `raw` is not in the same style as `mut`.
`raw` is not a keyword, but its role as a contextual keyword in these expressions is entirely syntactic so this does not seem to me like it should require full type information to be properly highlighted.

**rust-analyzer version**:  rust-analyzer version: 0.3.2337-standalone (5e7dd31c80 2025-03-09) [/home/r/.vscode-oss/extensions/rust-lang.rust-analyzer-0.3.2337-linux-x64/server/rust-analyzer]

**editor or extension**: VSCodium, extension v0.3.2337

**code snippet to reproduce**: 
```rust
fn main() {
    let x = &raw mut 5;
}
```

Veykril:
Technically that's an issue with the textmate grammar vscode ships. Though our semantic highlighting infra should handle this either way (which would fix it here)

Veykril:
Oh actually this should already work but does not because of https://github.com/rust-lang/rust-analyzer/blob/185f9deb452760f3abc2fde0500398e3198678cd/crates/ide/src/syntax_highlighting.rs#L227
We should make that `krate` an `Option`

PR #19400, opened 2025-03-20
Fix missing syntax highlighting for `&raw const` / `&raw mut` in all files.

Shourya742:
closes: #19357

### DIFF
--- a/crates/ide/src/syntax_highlighting.rs
+++ b/crates/ide/src/syntax_highlighting.rs
@@ -222,10 +222,7 @@ pub(crate) fn highlight(
     };
 
     let mut hl = highlights::Highlights::new(root.text_range());
-    let krate = match sema.scope(&root) {
-        Some(it) => it.krate(),
-        None => return hl.to_vec(),
-    };
+    let krate = sema.scope(&root).map(|it| it.krate());
     traverse(&mut hl, &sema, config, InRealFile::new(file_id, &root), krate, range_to_highlight);
     hl.to_vec()
 }
@@ -235,7 +232,7 @@ fn traverse(
     sema: &Semantics<'_, RootDatabase>,
     config: HighlightConfig,
     InRealFile { file_id, value: root }: InRealFile<&SyntaxNode>,
-    krate: hir::Crate,
+    krate: Option<hir::Crate>,
     range_to_highlight: TextRange,
 ) {
     let is_unlinked = sema.file_to_module_def(file_id).is_none();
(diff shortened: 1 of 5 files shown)


crates/hir/src/semantics.rs
```diff
@@ -2002,7 +2002,7 @@ fn find_root(node: &SyntaxNode) -> SyntaxNode {
 ///
 /// Note that if you are wondering "what does this specific existing name mean?",
 /// you'd better use the `resolve_` family of methods.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
Veykril, line 2005: This should not be necessary

crates/ide/src/syntax_highlighting.rs
```diff
@@ -222,10 +223,7 @@ pub(crate) fn highlight(
     };
 
     let mut hl = highlights::Highlights::new(root.text_range());
-    let krate = match sema.scope(&root) {
-        Some(it) => it.krate(),
-        None => return hl.to_vec(),
-    };
+    let krate = sema.scope(&root);
```
Veykril, line 226:
```suggestion
    let krate = sema.scope(&root).map(|it| it.krate());
```

crates/ide/src/syntax_highlighting.rs
```diff
@@ -235,7 +233,7 @@ fn traverse(
     sema: &Semantics<'_, RootDatabase>,
     config: HighlightConfig,
     InRealFile { file_id, value: root }: InRealFile<&SyntaxNode>,
-    krate: hir::Crate,
+    krate: Option<SemanticsScope<'_>>,
```
Veykril, line 236:
```suggestion
    krate: Option<hir::Crate>,
```
and so on

Veykril:
> @Veykril, with our current test fixture infrastructure, we can't write tests for non-crate files. The test expects the error: "missing default crate root, specify a main.rs or lib.rs."
> 
> If I want to test something like this, it won't work:
> 
> ```rust
>     check_highlighting(
>         r#"
> //- /foo.rs
> fn main() {
>     let x = &raw mut 5;
> }
> "#,
>         expect_file!["./test_data/highlight_issue_19357.html"],
>         false,
>     );
> ```
> 
> Should we make changes in the same PR to support this test?

I think


```rust
check_highlighting(
    r#"
//- /foo.rs
fn main() {
    let x = &raw mut 5;
}
//- /main.rs
"#,
```
might work?

Veykril (approved):
Thanks!

PR #19416, opened 2025-03-22
fix: Handle multiple `#[repr(..)]` attrs correctly

### LINKED ISSUE #15037: Layout info of `#[repr(C)] #[repr(packed)]` struct is wrong
```Rust
#[repr(C)]
#[repr(packed)]
struct X {
    id: i32,
}

fn main() {
    println!("Hello, world! {}", std::mem::align_of::<X>());
}
```
Hover on `X` shows alignment is 4, but the output of program is 1.

ShoyuVanilla:
Fixes #15037

In codes like the following cases:

```rust
#[repr(X)]
#[repr(Y)]
#[repr(Z)]
struct Foo { .. }
```

we've been using only the first `#[repr(X)]` attribute on struct, but we should handle them like `#[repr(X, Y, Z)]`

### DIFF
--- a/crates/hir-def/src/attr.rs
+++ b/crates/hir-def/src/attr.rs
@@ -233,7 +233,12 @@ impl Attrs {
     }
 
     pub fn repr(&self) -> Option<ReprOptions> {
-        self.by_key(&sym::repr).tt_values().find_map(parse_repr_tt)
+        self.by_key(&sym::repr).tt_values().filter_map(parse_repr_tt).fold(None, |acc, repr| {
+            acc.map_or(Some(repr), |mut acc| {
+                merge_repr(&mut acc, repr);
+                Some(acc)
+            })
+        })
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/hir-def/src/attr.rs
```diff
@@ -260,6 +265,19 @@ fn parse_rustc_legacy_const_generics(tt: &crate::tt::TopSubtree) -> Box<[u32]> {
...
+    *pack = match (*pack, other.pack) {
+        (Some(pack), None) | (None, Some(pack)) => Some(pack),
+        _ => (*pack).min(other.pack),
+    };
+    if other.int.is_some() {
+        *int = other.int;
+    }
+}
+
 fn parse_repr_tt(tt: &crate::tt::TopSubtree) -> Option<ReprOptions> {
```
ShoyuVanilla, line 281: Diffs inside this function is not necessary but I think that using the above `fn merge_repr` would be better in consistency.

crates/hir-def/src/lib.rs
```diff
@@ -24,9 +24,6 @@ extern crate rustc_hashes;
 #[cfg(not(feature = "in-rust-tree"))]
 extern crate ra_ap_rustc_abi as rustc_abi;
 
-#[cfg(not(feature = "in-rust-tree"))]
-extern crate ra_ap_rustc_hashes as rustc_hashes;
-
```
ShoyuVanilla, line 29: We've been using this for `rustc_hashes::Hash64::ZERO` but it is replaced with `Default::default()` in this PR

Veykril (approved):

Veykril:
Thanks!

PR #19419, opened 2025-03-22
Render layout and other extra information on hovering `Self`

### LINKED ISSUE #15132: Hovering `Self` of an impl does not show memory layout
```rs
struct Foo(i32) where Self:; // This `Self` works, as its an ADT Self type

// these two do not as they are impl self types, but we can totally calculate the layout of these 
impl Foo {
    fn f() -> Self 
		where Self: 
	{loop {}}
}
```

ShoyuVanilla:
Resolves #15132


Veykril (approved):
Thanks!

PR #19429, opened 2025-03-23
chore: Remove unused dependencies

Veykril:


lnicola:
We did lose source text compression in the Salsa migration, didn't we?

Veykril:
> We did lose source text compression in the Salsa migration, didn't we?

Yes I was going to point that out, we likely wanna bring that back.

PR #19430, opened 2025-03-23
chore: Bump dependencies

Veykril:


lnicola:
Hmm, so we have `zip -> zopfli -> lockfree-object-pool` which is BSL-1.0 (the Boost Software License, not the MariaDB/Hashicorp one). But the `rustc` check only allows for BSL crates dual-licensed as Apache-2.0: "BSL is not acceptable, but we use it under Apache-2.0".

So if BSL and Apache-2.0 are incompatible, maybe we should raise this with `zopfli`.

lnicola:
I pushed a commit to disable `zopfli`. We only use `zip` in `xtask dist`, it's not going to make a difference.

Cargo.toml
```diff
@@ -159,12 +156,16 @@ tracing-subscriber = { version = "0.3.18", default-features = false, features =
...
 ] }
 triomphe = { version = "0.1.14", default-features = false, features = ["std"] }
-url = "2.3.1"
-xshell = "0.2.5"
+url = "2.5.4"
+xshell = "0.2.7"
 
 
 # We need to freeze the version of the crate, as the raw-api feature is considered unstable
 dashmap = { version = "=5.5.3", features = ["raw-api"] }
```
davidbarsky, line 164: It'd be good to bump dashmap to `7.0.0-rc2` (here, and in Salsa), as it'd allow us to get rid of the `raw-api` feature entirely: it's replaced by hashbrown's newly-stable `HashTable`.

Unfortunately, when I tried doing so, the changes to the interning macro in `hir` was kinda difficult  to wrap my head around and I meant to open an issue to ask for help/clarification, so... I did it on this PR.

davidbarsky (approved):

Cargo.toml
Veykril: Let's do that separately

Veykril: Looking into this, what makes you think we can drop the `raw-api` feature? We still need that

PR #19432, opened 2025-03-23
fix: Yet another false positive invalid cast diagnostic

### LINKED ISSUE #19431: False positive cast diagnostics with DSTs and `zerocopy`
> There's still a problem with `KnownLayout` when the DST is wrapped:
> 
> ```rust
> use zerocopy::KnownLayout;
> 
> #[derive(KnownLayout)]
> #[repr(C)]
> pub struct Dst([u8]);
> 
> #[derive(KnownLayout)]
> #[repr(C)]
> pub struct Struct {
>     pub dst: Dst,
> }
> ```
> 
> ```
> cannot cast thin pointer `*mut __ZerocopyKnownLayoutMaybeUninit` to fat pointer `*mut _` rust-analyzer[E0607]
> ```
> 
> `rust-analyzer 1.85.1 (4eb16125 2025-03-15)`
> 
> No problem with rustc. 

 _Originally posted by @cloneable in [#18682](https://github.com/rust-lang/rust-analyzer/issues/18682#issuecomment-2746185008)_

ShoyuVanilla:
Fixes #19431

### DIFF
--- a/crates/ide-diagnostics/src/handlers/invalid_cast.rs
+++ b/crates/ide-diagnostics/src/handlers/invalid_cast.rs
@@ -440,8 +440,9 @@ fn main() {
     q as *const [i32];
   //^^^^^^^^^^^^^^^^^ error: cannot cast thin pointer `*const i32` to fat pointer `*const [i32]`
 
+    // FIXME: This should emit diagnostics but disabled to prevent many false positives
     let t: *mut (dyn Trait + 'static) = 0 as *mut _;
-                                      //^^^^^^^^^^^ error: cannot cast `usize` to a fat pointer `*mut _`
+
     let mut fail: *const str = 0 as *const str;
                              //^^^^^^^^^^^^^^^ error: cannot cast `usize` to a fat pointer `*const str`
     let mut fail2: *const str = 0isize as *const str;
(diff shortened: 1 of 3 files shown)


crates/ide-diagnostics/src/handlers/invalid_cast.rs
```diff
@@ -440,8 +440,9 @@ fn main() {
     q as *const [i32];
   //^^^^^^^^^^^^^^^^^ error: cannot cast thin pointer `*const i32` to fat pointer `*const [i32]`
 
+    // FIXME: This should emit diagnostics but disabled to prevent many false positives
     let t: *mut (dyn Trait + 'static) = 0 as *mut _;
-                                      //^^^^^^^^^^^ error: cannot cast `usize` to a fat pointer `*mut _`
+
```
ShoyuVanilla, line 445: We have disabled casts into `*const dyn Trait` here

https://github.com/rust-lang/rust-analyzer/blob/37acea8052d5d27e7f1312d9e6e743a9da176c21/crates/hir-ty/src/infer/cast.rs#L125-L130

but this somehow survived it due to hiding behinde inference var 😅

crates/ide-diagnostics/src/handlers/invalid_cast.rs
```diff
@@ -1161,6 +1162,49 @@ struct ZerocopyKnownLayoutMaybeUninit(<<Flexible as Field>::Type as KnownLayout>
...
+impl KnownLayout for Dst {
+    type MaybeUninit = Dst;
+    type PointerMetadata = <[u8] as KnownLayout>::PointerMetadata;
+}
+
+struct ZerocopyKnownLayoutMaybeUninit(<<Struct as Field>::Type as KnownLayout>::MaybeUninit);
+
+fn test(ptr: *mut ZerocopyKnownLayoutMaybeUninit) -> *mut <<Struct as Field>::Type as KnownLayout>::MaybeUninit {
+    ptr as *mut _
```
ShoyuVanilla, line 1206: In `zerocopy`'s proc macro expansion, there are so many nested projection types and inference vars and because of that the last field extracting stops along the way in the following function:

https://github.com/rust-lang/rust-analyzer/blob/37acea8052d5d27e7f1312d9e6e743a9da176c21/crates/hir-ty/src/infer/unify.rs#L919-L944

Ideally, this should anyway return `!Sized` as the function body before trait solving is just a performance-related logic, but the chalk returns wrong unique solution to a goal `struct Dst([u8]): Sized` 🤔 

So, I added a function eagerly resolves such nested inference vars and projection types.

Veykril (approved):

Issue #19417, opened 2025-03-22
Wrong `unexpected token` diagnostic when passing negative numbers to `concat!`


vdrn:
This code compiles fine with rustc, but rust analyzer reports error: `unexpected token`
```rust
 pub const T: &str = concat!(-42.0); 
```

**rust-analyzer version**: 1.87.0-nightly (be73c1f 2025-03-21)

**rustc version**: rustc 1.87.0-nightly (be73c1f46 2025-03-21)

pedrocarlo:
Having this same issue where the rustc can correctly expand the macro with negative integers and floats, but RA cannot. I use extensively this macro, so I am getting red squiggles everywhere

macro:
```rust
sqlite_values!(-1)
```

RA
```rust
// Recursive expansion of sqlite_values! macro
// ============================================

<[_]>::into_vec(
    #[rustc_box]
    alloc::boxed::Box::new([(<[_]>::into_vec(
        #[rustc_box]
        alloc::boxed::Box::new([(::rusqlite::types::Value::Integer())]),
    ))]),
)
```

rustc
```rust
<[_]>::into_vec(
            #[rustc_box]
            ::alloc::boxed::Box::new([
                <[_]>::into_vec(
                    #[rustc_box]
                    ::alloc::boxed::Box::new([::rusqlite::types::Value::Integer(-1i64)]),
                ),
            ]),
        ),
```

PR #19434, opened 2025-03-23
fix: negative nums in `concat!` expansion

vishruth-thimmaiah:
Fixes: #19417

This PR provides a fix to the expansion of the `concat!` macro, where negative numbers give an `unexpected token` error.

```rust
 pub const T: &str = concat!(-42.0);
```
expands to:

```rust
pub const T: &str = "-42.0"
```

### DIFF
--- a/crates/hir-def/src/macro_expansion_tests/builtin_fn_macro.rs
+++ b/crates/hir-def/src/macro_expansion_tests/builtin_fn_macro.rs
@@ -454,13 +454,13 @@ fn test_concat_expand() {
 #[rustc_builtin_macro]
 macro_rules! concat {}
 
-fn main() { concat!("fo", "o", 0, r#""bar""#, "\n", false, '"', '\0'); }
+fn main() { concat!("fo", "o", 0, r#""bar""#, "\n", false, '"', -4, - 4, '\0'); }
 "##,
         expect![[r##"
 #[rustc_builtin_macro]
 macro_rules! concat {}
 
-fn main() { "foo0\"bar\"\nfalse\"\u{0}"; }
+fn main() { "foo0\"bar\"\nfalse\"-4-4\u{0}"; }
 "##]],
     );
 }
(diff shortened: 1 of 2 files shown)


pedrocarlo:
Hey @Veykril, sorry about the ping. Unfortunately, I'm suffering immensely from this bug. The fix here seems to be pretty simple to me. How hard would it be for this to be merged? Thanks for the great work you are putting into Rust Analyzer!

crates/hir-def/src/macro_expansion_tests/builtin_fn_macro.rs
```diff
@@ -456,13 +456,13 @@ fn test_concat_expand() {
 #[rustc_builtin_macro]
 macro_rules! concat {}
 
-fn main() { concat!("fo", "o", 0, r#""bar""#, "\n", false, '"', '\0'); }
+fn main() { concat!("fo", "o", 0, r#""bar""#, "\n", false, '"', -4, '\0'); }
```
Veykril, line 459:
```suggestion
fn main() { concat!("fo", "o", 0, r#""bar""#, "\n", false, '"', -4, - 4, '\0'); }
```
lets also throw in a space for good measure

crates/hir-expand/src/builtin/fn_macro.rs
```diff
@@ -508,10 +511,40 @@ fn concat_expand(
...
+                    TtElement::Leaf(tt::Leaf::Literal(it))
+                        if matches!(it.kind, tt::LitKind::Integer | tt::LitKind::Float) =>
+                    {
+                        format_to!(text, "-{}", it.symbol.as_str());
+                        record_span(punct.span.cover(it.span));
+                    }
+                    _ => {
+                        err.get_or_insert(ExpandError::other(
+                            call_site,
+                            "expected number after '-'",
```
Veykril, line 537:
```suggestion
                            "expected integer or floating pointer number after '-'",
```

Veykril (approved):
small nits, otherwise looks good to me, thanks!

crates/hir-expand/src/builtin/fn_macro.rs
vishruth-thimmaiah: @rustbot ready

PR #19437, opened 2025-03-24
minor: Sync from downstream

lnicola:


lnicola:
Oh, that change didn't get reverted.

PR #14430, opened 2023-03-29
fix: Canonicalize rust-project.json manifest path

Veykril:
Looked a bit more into this and I think we can do this after all, I don't see any place where this should break things 
cc https://github.com/rust-lang/rust-analyzer/pull/14168 https://github.com/rust-lang/rust-analyzer/pull/14402#issuecomment-1487257246

PR #19438, opened 2025-03-24
fix(load-cargo): load workspace from absolute path

### LINKED ISSUE #19386: [ra_ap_load_cargo] Use `std::fs::canonicalize()` instead of `std::env::current_dir()?.join(..)`
In `ra_ap_load_cargo::load_workspace_at()`:

https://github.com/rust-lang/rust-analyzer/blob/facaf8b870e0dc509b34d5de2a0762aa09e50f62/crates/load-cargo/src/lib.rs#L47-L50

We are doing `std::env::current_dir()?.join(root)` assuming that the path being supplied is relative to the current directory. `root.as_ref().canonicalize()?` achieves the same thing, but also allows absolute paths to be passed into the function by external consumers of the library API. I think this is a nice improvement that does not break behaviour (as far as I can see). Happy to send a patch if acceptable.

snprajwal:
Fixes #19386 

The change does not break the API, or affect crate users with working code.

### DIFF
--- a/crates/load-cargo/src/lib.rs
+++ b/crates/load-cargo/src/lib.rs
@@ -44,7 +44,13 @@ pub fn load_workspace_at(
     load_config: &LoadCargoConfig,
     progress: &dyn Fn(String),
 ) -> anyhow::Result<(RootDatabase, vfs::Vfs, Option<ProcMacroClient>)> {
-    let root = AbsPathBuf::assert_utf8(std::env::current_dir()?.join(root));
+    // Use the provided path as is if it is absolute.
+    // Otherwise, join it with the current dir path.
+    let root = AbsPathBuf::assert_utf8(if root.is_absolute() {
+        root.to_path_buf()
+    } else {
+        std::env::current_dir()?.join(root)
+    });
     let root = ProjectManifest::discover_single(&root)?;
     let manifest_path = root.manifest_path().clone();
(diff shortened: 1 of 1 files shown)


snprajwal:
_sigh_ Windows strikes again... saw #14430, RIP

crates/load-cargo/src/lib.rs
```diff
@@ -44,7 +44,13 @@ pub fn load_workspace_at(
...
     progress: &dyn Fn(String),
 ) -> anyhow::Result<(RootDatabase, vfs::Vfs, Option<ProcMacroClient>)> {
-    let root = AbsPathBuf::assert_utf8(std::env::current_dir()?.join(root));
+    // Use the provided path as is if it is absolute.
+    // Otherwise, join it with the current dir path.
+    let root = AbsPathBuf::assert_utf8(if root.is_absolute() {
+        root.to_path_buf()
+    } else {
+        std::env::current_dir()?.join(root)
+    });
```
Veykril, line 53: This seems unnecessary? 
Quoting the docs of `Path::join`

> If `path` is absolute, it replaces the current path.

`path` being the argument to `join` here, that is in our case `root`. So this change shouldn't do anything here

snprajwal: Ah, right. In that case, something else locally was breaking the workspace discovery. Apologies for the noise.

Issue #19442, opened 2025-03-24
Add more completion items for `impl` to handle common scenarios


Natural-selection1:
Currently, the completion for "impl" only provides the basic "impl" snippet. While this is universally applicable, I believe it overlooks adaptation for common usage scenarios - which is precisely where code completion should excel. Therefore, I propose adding more relevant completion items.

![Image](https://github.com/user-attachments/assets/fa602662-80b8-4a1b-a3db-81fec5435e4f)

According to the Rust Reference, I suggest adding these two items would be appropriate:

```rust
add_keyword("impl for", "impl $1 for $2 {\n    $0\n}");
add_keyword("impl<> for <>", "impl<$4> $1 for $2<$3> {\n    $0\n}");
```

However, I would like to gather opinions regarding:
1. The naming conventions for these two items
2. The cursor navigation order in the second snippet (the current version reflects what I consider to be a reasonable sequence)

The current proposals represent my initial thoughts on optimal implementation.

PR #19447, opened 2025-03-25
add more completion about "impl"

Natural-selection1:
Completed: #19442

### DIFF
--- a/crates/ide-completion/src/completions/item_list.rs
+++ b/crates/ide-completion/src/completions/item_list.rs
@@ -115,7 +115,6 @@ fn add_keywords(acc: &mut Completions, ctx: &CompletionContext<'_>, kind: Option
             if no_vis_qualifiers {
                 add_keyword("impl", "impl $1 {\n    $0\n}");
                 add_keyword("impl for", "impl $1 for $2 {\n    $0\n}");
-                add_keyword("impl<> for <>", "impl<$4> $1 for $2<$3> {\n    $0\n}");
             }
         }
 
@@ -147,7 +146,6 @@ fn add_keywords(acc: &mut Completions, ctx: &CompletionContext<'_>, kind: Option
         if no_vis_qualifiers {
             add_keyword("impl", "impl $1 {\n    $0\n}");
             add_keyword("impl for", "impl $1 for $2 {\n    $0\n}");
-            add_keyword("impl<> for <>", "impl<$4> $1 for $2<$3> {\n    $0\n}");
         }
     }
 
--- a/crates/ide-completion/src/completions/item_list.rs
+++ b/crates/ide-completion/src/completions/item_list.rs
@@ -114,6 +114,8 @@ fn add_keywords(acc: &mut Completions, ctx: &CompletionContext<'_>, kind: Option
             add_keyword("trait", "trait $1 {\n    $0\n}");
             if no_vis_qualifiers {
                 add_keyword("impl", "impl $1 {\n    $0\n}");
+                add_keyword("impl for", "impl $1 for $2 {\n    $0\n}");
+                add_keyword("impl<> for <>", "impl<$4> $1 for $2<$3> {\n    $0\n}");
             }
         }
 
@@ -144,6 +146,8 @@ fn add_keywords(acc: &mut Completions, ctx: &CompletionContext<'_>, kind: Option
         add_keyword("use", "use $0");
         if no_vis_qualifiers {
             add_keyword("impl", "impl $1 {\n    $0\n}");
+            add_keyword("impl for", "impl $1 for $2 {\n    $0\n}");
+            add_keyword("impl<> for <>", "impl<$4> $1 for $2<$3> {\n    $0\n}");
         }
     }
 
--- a/crates/ide-completion/src/completions/keyword.rs
+++ b/crates/ide-completion/src/completions/keyword.rs
@@ -56,6 +56,8 @@ mod tests {
                 kw extern
                 kw fn
                 kw impl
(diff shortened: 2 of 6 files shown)


ChayimFriedman2:
I wonder if these should be without spaces/angle brackets, since users cannot type them to filter the completions list?

Natural-selection1:
> I wonder if these should be without spaces/angle brackets, since users cannot type them to filter the completions list? 我怀疑这些是否应该没有空格 / 尖括号，因为用户无法输入它们来过滤完成列表？


I don't have strong opinions about spaces, since we already have completion items like "while let" (unless we consider it outdated design).  
image  

I think `impl` is an extremely special block. When you want to write it, just typing "im" should be enough to lock it in.  
image  

As for angle brackets, I do think it requires careful consideration. But when I realized we also have completion items like "pub(crate)", I believe the reasoning is similar. When we write "pub"/"impl" (and their variants), just typing "pu"/"im" should be sufficient to lock them in. Perhaps the only thing to consider is "how to distinguish between 'impl<> for <>' and 'impl for' without requiring the use of ↑↓ arrows."

Veykril:
> I wonder if these should be without spaces/angle brackets, since users cannot type them to filter the completions list?

That;s a good thing to check, I would not be surprised if editors have differing behavior for spaces here (though its also relevant what our capabilities register for completion characters iirc)

flodiebold:
I'm not convinced it's worth adding more of these. Especially for the `<>` one, I'd rather have an assist that automatically adds a type parameter? Like you type `impl Foo<T|>`, trigger the assist and it adds the `<T>` after `impl`.

Natural-selection1:
> I'm not convinced it's worth adding more of these. Especially for the `<>` one, I'd rather have an assist that automatically adds a type parameter? Like you type `impl Foo<T|>`, trigger the assist and it adds the `<T>` after `impl`.

Although not strictly enforced (e.g., `impl<T> fmt::Debug for IterMut<'_, T>`),
the `impl<>` tends to be a superset of the generic parameters and lifetime parameters in `Foo<>`.
```rust
impl<T, U> Foo<T> {
    fn convert<U>(&self) -> Foo<U> {}
}
```
Moreover, if we don't use `where` clauses to specify bounds, 
we would annotate them in `impl<>` instead. 
```rust
impl<T: PartialEq, A: Allocator> PartialEq for LinkedList<T, A>
```
If we do need to specify bounds in `impl<>` (which is not uncommon), 
such auxiliary functionality wouldn't become much better than "impl<$4> $1 for $2<$3>"

Natural-selection1:
This PR is based on the following observations:

* We typically write `impl Foo` very few times but write `impl Some for Foo` many times.  
  Therefore, I added `"impl for" (impl $1 for $2)`.

---

* For non-generic data containers, there are also implementations like `impl<'a, T> for Foo`.  
* Deleting a "shifted char" is easier than writing it.  
* For generic data containers, the generic and lifetime parameters in `impl<>` tend to always be a superset of the non-concrete generic and lifetime parameters in `Foo<>`.  
* If `where` is not used, we annotate bounds in `impl<>` (which is extremely common for non-closure generic parameters).  

Thus, what I chose to add is `"impl<> for <>"` `(impl<$4> $1 for $2<$3>)` instead of `(impl<$3> $1 for $2<$3>)`.  

---

I don’t think `blank` `<>` is difficult to accept,
since we already have completions like `while let`/`if let`/`pub(crate)`/`pub(super)`,
which contain `()` as shifted characters and `blank`.
The only problem, in my opinion, is **how to distinguish between `impl<> for <>` and `impl for` without requiring the use of ↑↓ arrows and blank**.  

With the current approach:  
- `im Tab` → `impl`  
- `im Blank Tab` → `impl for`  
- `im Blank Blank Tab` → `impl<> for <>`  

Maybe change `impl<> for <>` to `impl for with param`?


Additionally, adjusting the cursor jump order? for example, changing:  
- `impl $1 for $2` to `impl $2 for $1`  
- `impl<$4> $1 for $2<$3>` to `impl<$4> $3 for $1<$2>` (or another solution)  
—to make it more aligned with the writing logic.  

--- 

I wonder if my argument is persuasive?

Veykril:
I feel like a better workflow (that we should enable) would be the following maybe?:

- write `impl ` / get a `impl $0 {}` completion
- start writing a trait name -> get a `Trait<$1, $2, ...> for $0` completion (with each required trait arg as a tabstop), that is we now have the following state `impl Trait<$1, $2, ...> for $0 {}`
- then offer assists (diagnostic quickfix) for adding unresolved name segments as a new generic parameter to the surrounding impl (this one we should really have either way, its so useful)

That is, I feel like the user shouldn't even need to jump back to the generic parameter listing in the first place. If you create a generic parameter there, you will have to use it anyways and so typing out the reference to it first and then have r-a just create the definition for you sounds more ergonomic to me?

To be clear, I do agree there is an ergonomics gap here. Its just tabstop snippet completions can be a lot more annoying than helpful depending on how they are designed, and duplicating completions to accomodate for a variety of snippet setups can be more confusing than helpful which is why I (and presumably Florian) are a bit wary of this here.

Natural-selection1:
>  which is why I (and presumably Florian) are a bit wary of this here.

I fully understand the caution from you and other members—those would be new perspectives for me to consider (and they are usually better and more comprehensive).

So, at this stage, would just adding the "impl for" be acceptable?

Veykril:
Just the `impl for` is probably fine for now

Veykril (approved):

Veykril:
Those probably want to use snippets as well yes

PR #18565, opened 2024-11-26
feat: convert unmerge_use to SyntaxFactory SyntaxEditor abstraction

davidkurilla:
Contributes to issue #18285 

I think I fully converted this but tests are failing with the following output:

```
failures:

---- handlers::unmerge_use::tests::unmerge_glob_import stdout ----
thread 'handlers::unmerge_use::tests::unmerge_glob_import' panicked at /home/codespace/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rowan-0.15.15/src/cursor.rs:548:9:
assertion failed: !self.data().mutable
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- handlers::unmerge_use::tests::unmerge_indented_use_item stdout ----
thread 'handlers::unmerge_use::tests::unmerge_indented_use_item' panicked at /home/codespace/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rowan-0.15.15/src/cursor.rs:548:9:
assertion failed: !self.data().mutable

---- handlers::unmerge_use::tests::unmerge_nested_use_item stdout ----
thread 'handlers::unmerge_use::tests::unmerge_nested_use_item' panicked at /home/codespace/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rowan-0.15.15/src/cursor.rs:548:9:
assertion failed: !self.data().mutable

---- handlers::unmerge_use::tests::unmerge_renamed_use_item stdout ----
thread 'handlers::unmerge_use::tests::unmerge_renamed_use_item' panicked at /home/codespace/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rowan-0.15.15/src/cursor.rs:548:9:
assertion failed: !self.data().mutable

---- handlers::unmerge_use::tests::unmerge_use_item_on_self stdout ----
thread 'handlers::unmerge_use::tests::unmerge_use_item_on_self' panicked at /home/codespace/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rowan-0.15.15/src/cursor.rs:548:9:
assertion failed: !self.data().mutable

---- handlers::unmerge_use::tests::unmerge_use_item stdout ----
thread 'handlers::unmerge_use::tests::unmerge_use_item' panicked at /home/codespace/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rowan-0.15.15/src/cursor.rs:548:9:
assertion failed: !self.data().mutable

[...]

PR #18484, opened 2024-11-06
Migrate `merge_imports` Assist to Use `SyntaxFactory`

tareknaser:
This PR is part of #15710 and #18285.

**Changes Made**
- Followed the steps in #18285 for migration.
- Removed the `edits_mut` vector from `crates/ide-assists/src/handlers/merge_imports.rs`
- Added a new `SyntaxFactory` method `whitespace`
	- There are no mappings needed as far as I am concerned
	- Currently named `whitespace` as it mirrors `make::tokens::whitespace` but `tokens_whitespace` might be a more fitting name.
- Replaced the `edit_in_place::Removable` trait with `edit_in_place::EditorRemovable` ([Relevant conversation here](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/.60SyntaxEditor.60.20replacement.20for.20Mutable.20Syntax.20trees/near/479370360))
	- This trait requires a `SyntaxEditor` instance in the `remove` method to use instead of `ted`.
	- Implemented for `ast::Use` and `ast::UseTree`
	- Most implementations remain the same, except that I used `SyntaxEditor::delete` instead of `ted::remove` and` ted::remove_all_iter`. Since `SyntaxEditor::remove` doesn’t exist, I assumed delete would perform the same function—though this might be causing issues.

PR #19465, opened 2025-03-27
refactor: migrate `unmerge_use` -> `unmerge_imports` to syntax editor

### LINKED ISSUE #19441: Unmerge use assist ignores attributes
When using the 'unmerge use' assist on part of a `use` statement that has attributes attached, the new extracted `use` statement should use the same attributes, but it doesn't.

**rust-analyzer version**: 0.4.2354-standalone (dd2354c985 2025-03-24)

**rustc version**: 1.87.0-nightly (f4a216d28 2025-03-02)

**editor or extension**: VSCode (arch build, RA 0.4.2354)

```rust
#[cfg(any())]
#[allow(deprecated)]
use foo::{Bar, Baz}; // unmerge use of Baz
```

snprajwal:
Also ensures that attributes on the use item are applied to the new use item when unmerging.

Fixes #19441, contributes to #18285, and supersedes #18565. This patch also includes some changes from #18484, specifically the addition of the `Removable` trait.

cc @darichey, thanks for all the reviews on the other PR :) made things much easier for me

### DIFF
--- a/crates/ide-assists/src/handlers/unmerge_imports.rs
+++ b/crates/ide-assists/src/handlers/unmerge_imports.rs
@@ -12,9 +12,9 @@ use crate::{
     assist_context::{AssistContext, Assists},
 };
 
-// Assist: unmerge_use
+// Assist: unmerge_imports
 //
-// Extracts single use item from use list.
+// Extracts a use item from a use list into a standalone use list.
 //
 // ```
 // use std::fmt::{Debug, Display$0};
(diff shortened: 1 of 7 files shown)


snprajwal:
This is just me nitpicking, but would it make more sense to rename this assist to `unmerge_imports`, given that it has a sister assist called `merge_imports`?

Veykril:
Those should have matching names for sure I agree

snprajwal:
Gentle ping on this and #19469 :)

snprajwal:
Merged as a part of #19469

PR #19471, opened 2025-03-28
fix: resolve doc path for module

Hmikihiro:
This PR fixes an issue where links to modules in doc comments don't match the generated documentation.
スクリーンショット 2025-03-28 23 59 58

スクリーンショット 2025-03-28 23 54 02

I think links should be resolved starting from the parent module.
スクリーンショット 2025-03-28 23 53 09


Veykril:
This probably depends on whether the doc comments are inline (!//) or outline (//)?

PR #19490, opened 2025-04-01
fix: Fix panic in progress due to splitting unicode incorrectly

rossdylan:
In the rare occasion that someone has added unicode characters that are multiple code points, the previous slicing operation could incorrectly split that character, and panic. The fix is to use the `chars()` iterator and just skip the prefix.


rossdylan:
I can add a reproduction / test case tomorrow morning.

ChayimFriedman2 (approved):
This code is currently quadratic (unrelated to your change), it could use a perf improvement, but it isn't hot so I guess this is fine.

Thanks!

PR #19507, opened 2025-04-02
fix: resolve doc path from parent module if outer comments exist on module

Hmikihiro:
fix: https://github.com/rust-lang/rust-analyzer/issues/19506
https://github.com/rust-lang/rust-analyzer/pull/19471
スクリーンショット 2025-04-03 1 40 38

I set placement information about inner or outer in attribute.  
When I check link on module, I search from parent module if attribute of doc item have a outer comment.

### DIFF
--- a/crates/hir-expand/src/attrs.rs
+++ b/crates/hir-expand/src/attrs.rs
@@ -466,7 +466,7 @@ pub fn collect_attrs(
             Either::Left(attr) => attr.kind().is_outer(),
             Either::Right(comment) => comment.is_outer(),
         })
-        .map(|attr| (attr, false));
+        .zip(iter::repeat(false));
     outer_attrs
         .chain(inner_attrs)
         .enumerate()
--- a/crates/hir-expand/src/attrs.rs
+++ b/crates/hir-expand/src/attrs.rs
@@ -460,7 +460,7 @@ pub fn collect_attrs(
     owner: &dyn ast::HasAttrs,
 ) -> impl Iterator<Item = (AttrId, Either<ast::Attr, ast::Comment>)> {
     let inner_attrs =
-        inner_attributes(owner.syntax()).into_iter().flatten().map(|attr| (attr, true));
+        inner_attributes(owner.syntax()).into_iter().flatten().zip(iter::repeat(true));
     let outer_attrs = ast::AttrDocCommentIter::from_syntax_node(owner.syntax())
         .filter(|el| match el {
             Either::Left(attr) => attr.kind().is_outer(),
--- a/crates/hir-expand/src/attrs.rs
+++ b/crates/hir-expand/src/attrs.rs
@@ -123,15 +123,15 @@ impl RawAttrs {
             (None, entries @ Some(_)) => Self { entries },
             (Some(entries), None) => Self { entries: Some(entries.clone()) },
             (Some(a), Some(b)) => {
-                let last_ast_index = a.slice.last().map_or(0, |it| it.id.ast_index() + 1) as u32;
+                let last_ast_index = a.slice.last().map_or(0, |it| it.id.ast_index() + 1);
                 let items = a
                     .slice
                     .iter()
                     .cloned()
                     .chain(b.slice.iter().map(|it| {
                         let mut it = it.clone();
-                        let id = it.id.ast_index() as u32 + last_ast_index;
-                        it.id = AttrId::new(id as usize, it.id.is_inner_attr());
+                        let id = it.id.ast_index() + last_ast_index;
+                        it.id = AttrId::new(id, it.id.is_inner_attr());
                         it
                     }))
                     .collect::<Vec<_>>();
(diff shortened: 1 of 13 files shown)


crates/hir-expand/src/attrs.rs
```diff
@@ -211,6 +215,15 @@ pub struct Attr {
     pub path: Interned<ModPath>,
     pub input: Option<Box<AttrInput>>,
     pub ctxt: SyntaxContext,
+    pub place: AttrPlacement,
```
Veykril, line 218: This grows attributes by `8` bytes which will be noticable, we'll need to avoid that somehow if we can. Note quite sure how yet. Probably by using one of the `AttrId` bits (we already do something there for cfg_attr but that is not really used so we can replace that with this likely). Will write oiut better instructions later

Hmikihiro: https://github.com/rust-lang/rust-analyzer/blob/588948f2676312152122a4caefccd4062b569823/crates/hir-expand/src/attrs.rs#L185-L189

Right, that means AttrId currently uses bits like this:
- Bit 31 → is_cfg_attr
- Bits 24–30 → cfg_attr_index (7 bits)
- Bits 0–23 → ast_index (24 bits)

So, we could either remove or reduce the size of `cfg_attr_index`, and reuse the freed bit for placement information.
- Bit 31 → inner_or_outer
- Bits 24–30 → (unused)or  (is_cfg_attr & cfg_attr_index(6bit))
- Bits 0–23 → ast_index

Hmikihiro: https://github.com/rust-lang/rust-analyzer/blob/588948f2676312152122a4caefccd4062b569823/crates/hir-expand/src/attrs.rs#L108-L110

And `cfg_attr` was accessed through `cfg_attr_index()` but this function was only used to copy it when creating a new `[Attr]`. So would we remove cfg_attr?

Veykril: IF the cfg attr stuff isnt actually used then lets remove it yes.

Hmikihiro: I see. The inner-ness uses the most significant bit, and bits 24–30 are currently unused.

crates/hir-expand/src/attrs.rs
```diff
@@ -183,25 +180,21 @@ pub struct AttrId {
 // FIXME: This only handles a single level of cfg_attr nesting
 // that is `#[cfg_attr(all(), cfg_attr(all(), cfg(any())))]` breaks again
 impl AttrId {
-    const CFG_ATTR_BITS: usize = 7;
     const AST_INDEX_MASK: usize = 0x00FF_FFFF;
```
Veykril, line 183: We don't need the mask anymore, we just need to make sure to strip the inner attribute bit when using the index as an actual index

crates/hir-expand/src/attrs.rs
```diff
@@ -439,13 +432,18 @@ fn unescape(s: &str) -> Option<Cow<'_, str>> {
 pub fn collect_attrs(
     owner: &dyn ast::HasAttrs,
 ) -> impl Iterator<Item = (AttrId, Either<ast::Attr, ast::Comment>)> {
-    let inner_attrs = inner_attributes(owner.syntax()).into_iter().flatten();
-    let outer_attrs =
-        ast::AttrDocCommentIter::from_syntax_node(owner.syntax()).filter(|el| match el {
+    let inner_attrs =
+        inner_attributes(owner.syntax()).into_iter().flatten().map(|attr| (attr, true));
```
Veykril, line 436:
```suggestion
        inner_attributes(owner.syntax()).into_iter().flatten().zip(iter::repeat(true));
```
style nit

crates/hir/src/attrs.rs
```diff
@@ -105,21 +105,32 @@ impl HasAttrs for crate::Crate {
...
 pub fn resolve_doc_path_on(
     db: &dyn HirDatabase,
-    def: impl HasAttrs,
+    def: impl HasAttrs + Copy,
     link: &str,
     ns: Option<Namespace>,
 ) -> Option<DocLinkDef> {
-    resolve_doc_path_on_(db, link, def.attr_id(), ns)
+    let is_inner =
+        def.attrs(db).by_key(&intern::sym::doc).attrs().all(|attr| attr.id.is_inner_attr());
```
Veykril, line 113: This should be a parameter of  `resolve_doc_path_on` as it depends on the attribute the link was usually fetched from. A module for example can very likely have both inner and outer attributes mixed

crates/hir-expand/src/attrs.rs
Hmikihiro: I see. We don't need to clear 24-30bit. AST_INDEX uses 0-30bit.

78cce17cb2b566c4818be1f1aae357504edd42f3

crates/hir/src/attrs.rs
Hmikihiro: スクリーンショット 2025-05-07 1 05 48
fix commit 778322eb31b00ce239781cd29b685f681edaac79

I see. In old commits, if both inner and outer attributes mixed, those attributes was resolved as outer attributes.

Instead to resolve inner-ness in `resolve_doc_path_on`, this resolves inner-ness each attributes. 
https://github.com/rust-lang/rust-analyzer/blob/778322eb31b00ce239781cd29b685f681edaac79/crates/ide-db/src/documentation.rs#L58

https://github.com/rust-lang/rust-analyzer/blob/778322eb31b00ce239781cd29b685f681edaac79/crates/ide/src/doc_links.rs#L333-L342

crates/hir-expand/src/attrs.rs
```diff
@@ -175,25 +175,24 @@ pub struct AttrId {
...
-        self.id as usize & Self::AST_INDEX_MASK
+    const INNER_ATTR_SET_BIT: usize = 1 << 31;
+
+    pub fn new(id: usize, is_inner: bool) -> Self {
+        Self {
+            id: if is_inner {
+                id | Self::INNER_ATTR_SET_BIT
+            } else {
+                id & !Self::INNER_ATTR_SET_BIT
+            } as u32,
```
Veykril, line 186:
```suggestion
        assert!(id <= !Self::INNER_ATTR_SET_BIT);
        let id = as u32;
        Self {
            id: if is_inner {
                id | Self::INNER_ATTR_SET_BIT
            } else {
                id
            }
```

Veykril (approved):
One small thing, otherwise this looks good to me!

crates/hir-expand/src/attrs.rs
Hmikihiro: fix commit 7c7d4401884bd4e614923210897e639c9eada692

https://github.com/rust-lang/rust-analyzer/blob/7c7d4401884bd4e614923210897e639c9eada692/crates/hir-expand/src/attrs.rs#L180-L184

Veykril:
Thanks!

PR #19509, opened 2025-04-03
fix(ide-assists): remove `AssistKind::None`

snprajwal:
This was being used by a single assist, which qualifies under the "refactor" kind. The variant has been removed, and all usages updated accordingly.


snprajwal:
The metadata version bump seems trivial so I included it here. Let me know if you want me to split it into a separate PR.

lnicola (approved):
`Refactor` sounds good for this one. Dropping the attribution on the version commit is a bit unfortunate, but that's fine, it's trivial anyway.

PR #19512, opened 2025-04-03
Upstream stdx changes

BenjaminBrienen:
I have some changes on my version of stdx that I figured I might as well PR back to r-a.

Let me know if there are things that you want me to undo.
Almost all of the changes come from following clippy lints which are not currently enabled in this workspace.

### DIFF
--- a/crates/stdx/src/lib.rs
+++ b/crates/stdx/src/lib.rs
@@ -17,7 +17,7 @@ pub mod thread;
 pub use itertools;
 
 #[inline(always)]
-pub fn is_ci() -> bool {
+pub const fn is_ci() -> bool {
     option_env!("CI").is_some()
 }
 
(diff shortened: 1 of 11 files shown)


BenjaminBrienen:
proc-macro-srv failing unit tests are unrelated

crates/stdx/src/thread/intent.rs
```diff
@@ -265,6 +266,7 @@ mod imp {
 
 // FIXME: Windows has QoS APIs, we should use them!
 #[cfg(not(target_vendor = "apple"))]
+#[expect(clippy::missing_const_for_fn, reason = "const trait impl is not stable")]
```
Veykril, line 269:
```suggestion
``` 
No need for this, we don't have that lint enabled anyways

crates/stdx/src/thread/pool.rs
```diff
@@ -73,13 +78,18 @@ impl Pool {
...
 
+    /// # Panics
+    ///
+    /// Panics if job panics
+    #[inline]
     pub fn spawn<F>(&self, intent: ThreadIntent, f: F)
     where
         F: FnOnce() + Send + UnwindSafe + 'static,
     {
+        #[expect(clippy::semicolon_if_nothing_returned, reason = "thin wrapper")]
```
Veykril, line 92: Let's just add the semicolon here instead

Veykril (commented):
Why all the random inlines? And most of the must use also seems unnecessary to me (aside from the constructors/builders).

Veykril:
I find putting those on a bunch of functions to be more noise than helpful honestly. We don't run clippy with all pedantic lints precisely because a lot of those lints are subjectively questionable.

As for inline specifically, none of that code is perf critical so we can just rely on the compiler doing the right thing (what it considers right). In fact, a lint telling you about when to use inline makes no sense to me. If the lint knows, the compiler should know either way. So I'd at least like to see the `inline` attributes removed here. `must_use` I don't care too much about (as those are reasonable here either way).

BenjaminBrienen:
Let me know if there are any other changes you want me to undo.

crates/stdx/src/lib.rs
```diff
@@ -259,12 +265,11 @@ impl JodChild {
         command.spawn().map(Self)
     }
 
+    #[must_use]
+    #[cfg(target_arch = "wasm32")]
```
Veykril, line 269:
```suggestion
    #[cfg(not(target_arch = "wasm32")]
```

Veykril:
Thanks!

PR #19515, opened 2025-04-04
fix: don't drop references with more than one definition.

### LINKED ISSUE #19393: scip output doesn't include implicit field references during struct initialization
**rust-analyzer version**: rust-analyzer 1.85.0 (4d91de4e 2025-02-17)

```rust
struct Point {
    x: f64,
    y: f64,
}

fn main() {
    let x = 5.;
    let y = 10.;


    let mut p = Point {x, y };


    p.x = 9.;
    p.y = 10.;
    println!("Hello, world! {} {}", p.x, p.y);  
}

```

`scip print index.scip` gives:

```
        &scip.Occurrence{
          Range: []int32{
            10,
            12,
            13,
          },
          Symbol:                "local 3",
          SymbolRoles:           1,
          OverrideDocumentation: []string(nil),
          SyntaxKind:            0,
          Diagnostics:           []*scip.Diagnostic(nil),
          EnclosingRange:        []int32(nil),
        },
        &scip.Occurrence{
          Range: []int32{
            10,
            16,
            21,
          },
          Symbol:                "rust-analyzer cargo ra-test 0.1.0 Point#",
          SymbolRoles:           0,
          OverrideDocumentation: []string(nil),
          SyntaxKind:            0,
          Diagnostics:           []*scip.Diagnostic(nil),
          EnclosingRange:        []int32(nil),
        },
```

for the line `    let mut p = Point {x, y };`

We should include field references for `Point#x` and `Point#y` like is done if it was `Point{x:x, y:y}`

jrmuizel:
Implicit field references during struct initialization were being dropped because get_definition was returning None because there were multiple definitions.

This adds a new helper, `get_defintions`, that supports returning more than one definition for a given token and hooks it up.

Fixes #19393


crates/ide-db/src/helpers.rs
```diff
@@ -126,3 +127,19 @@ pub fn get_definition(
     }
     None
 }
+
+// FIXME: This is a weird function
+pub fn get_definitions(
```
Veykril, line 132: Let's move this to `static_index`

Veykril (approved):
thanks!

PR #19519, opened 2025-04-04
feat(project-model): provide flag for no deps

### LINKED ISSUE #19518: [project-model] Provide a config option to enable `--no-deps` for Cargo projects
Ref: [#t-compiler/rust-analyzer > ✔ Disable cargo metadata fetch when loading workspace](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/.E2.9C.94.20Disable.20cargo.20metadata.20fetch.20when.20loading.20workspace/with/510153024)

snprajwal:
A Cargo project can now be built without any dependency metadata being fetched. This classifies as a breaking change for the `ra_ap_project_model` public API, since we introduce a new field into the config struct, and a new configuration option for the analyzer.

Fixes #19518

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -609,6 +609,9 @@ config_data! {
         cargo_features: CargoFeaturesDef      = CargoFeaturesDef::Selected(vec![]),
         /// Whether to pass `--no-default-features` to cargo.
         cargo_noDefaultFeatures: bool    = false,
+        /// Whether to skip fetching dependencies. If set to "true", the analysis is performed
+        /// entirely offline, and Cargo metadata for dependencies is not fetched.
+        cargo_noDeps: bool = false,
         /// Relative path to the sysroot, or "discover" to try to automatically find it via
         /// "rustc --print sysroot".
         ///
@@ -2027,6 +2030,7 @@ impl Config {
             extra_env: self.cargo_extraEnv(source_root).clone(),
             target_dir: self.target_dir_from_config(source_root),
             set_test: *self.cfg_setTest(source_root),
+            no_deps: *self.cargo_noDeps(source_root),
         }
     }
 
--- a/crates/project-model/src/cargo_workspace.rs
+++ b/crates/project-model/src/cargo_workspace.rs
@@ -108,7 +108,10 @@ pub struct CargoConfig {
     pub invocation_strategy: InvocationStrategy,
     /// Optional path to use instead of `target` when building
     pub target_dir: Option<Utf8PathBuf>,
+    /// Gate `#[test]` behind `#[cfg(test)]`
     pub set_test: bool,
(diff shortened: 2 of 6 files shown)


BenjaminBrienen (approved):

crates/rust-analyzer/src/config.rs
```diff
@@ -609,6 +609,9 @@ config_data! {
         cargo_features: CargoFeaturesDef      = CargoFeaturesDef::Selected(vec![]),
         /// Whether to pass `--no-default-features` to cargo.
         cargo_noDefaultFeatures: bool    = false,
+        /// Whether to skip fetching dependencies. If set to "true", the analysis is performed
+        /// entirely offline, and Cargo metadata for dependencies is not fetched.
+        cargo_noDeps: bool = false,
```
Veykril, line 614: There is little reason to expose this to users of the IDE. This config mainly seems useful for people building tools on top of r-a / using it as a library.

Veykril (changes requested):

crates/rust-analyzer/src/config.rs
snprajwal: It might be useful if someone wants to explicitly use r-a offline in their editor without disconnecting their internet (particularly for bigger projects)

Veykril: That sounds like a very odd workflow to me

BenjaminBrienen: metered connections

Veykril (approved):

crates/rust-analyzer/src/config.rs
davidbarsky: nit: maybe let's call this `fetch_Deps` and default it to true?

davidbarsky (changes requested):
have a naming nit about the configuration option.

crates/rust-analyzer/src/config.rs
Veykril: `fetchDeps` (config keys are camelCase), I'd prefer the `noDeps` variant I think as that mirrors the actual cargo flag name

Veykril:
Welp

davidbarsky:
A manual `Default` implementation is fine. I'd prefer a slightly better user experience over a manual implementation. If we want to be _really_ future-proof, this should probably be an enum

Regardless, I can flip naming in a followup PR.

BenjaminBrienen:
You might call it `offline` if you want the feature to cover other use cases intended for an offline workflow. The fact that it is `--no-deps` could be an implementation detail of offline mode.

crates/rust-analyzer/src/config.rs
davidbarsky: well, if it mirrors cargo, then that's fine.

PR #19520, opened 2025-04-04
Align usage of `default` and `new` with style guide

BenjaminBrienen:
https://github.com/rust-lang/rust-analyzer/blob/master/docs/book/src/contributing/style.md?plain=1#L329-L332

### DIFF
--- a/crates/hir-def/src/dyn_map.rs
+++ b/crates/hir-def/src/dyn_map.rs
@@ -112,6 +112,10 @@ pub struct Key<K, V, P = (K, V)> {
 }
 
 impl<K, V, P> Key<K, V, P> {
+    #[allow(
+        clippy::new_without_default,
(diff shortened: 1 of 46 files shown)


crates/hir-def/src/dyn_map.rs
```diff
@@ -112,6 +112,10 @@ pub struct Key<K, V, P = (K, V)> {
 }
 
 impl<K, V, P> Key<K, V, P> {
+    #[allow(
+        clippy::new_without_default,
+        reason = "this a const fn, so it can't be default yet. See <https://github.com/rust-lang/rust/issues/63065>"
+    )]
```
Veykril, line 118: Seeing this PR I was wondering about whether they finally fixed this or not. Guess they haven't ... (as in that the lint is still firing incorrectly here)

Veykril (approved):
Thanks!

PR #19526, opened 2025-04-05
The rustdoc file prefix for constants is "constant" not "const"

SOF3:
equivalent to https://github.com/rust-lang/rust/pull/139405, not sure which one is the source of truth


lnicola (approved):
Thanks!

PR #19532, opened 2025-04-06
chore: Run analysis-stats CI step in release

Veykril:
The build takes a bit longer but overall the analysis speeds up more from it


lnicola:
Interesting, last time I checked `debug` with `-Copt-level=1` was still faster overall.

Veykril:
Currently the step takes 11-12minutes overall, where as for metrics, build takes ~5 and the run takes 4 minutes. But I just realized I only looked at the `self` one. Forgot we are also running it against the sysroot. So it actually doesn't really save much I think. (maybe a minute)

PR #19534, opened 2025-04-06
chore: Cancel analysis-stats job when rust job fails

Veykril:
Tries to do what https://github.com/rust-lang/rust-analyzer/pull/19521/ does but more fine-grained and without an action dependency


Veykril:
Nevermind, turns out you can only cancel the entire workflow, not a job specifically

PR #19546, opened 2025-04-09
feat: highlighting of return values while the cursor is on `match` / `if` / `=>`

roife:
Inspired by #18924, follow up to #17542.

This PR adds a new feature that highlights the related return values when the cursor is on `match`, `if`, or a match arm arrow (`=>`). This helps users quickly identify the different possible outcomes of control flow expressions.

### For `match` keyword:
- The `match` keyword itself
- All return expressions in each arm of the match

### For `=>` (match arm arrow):
- The arrow itself
- The return expression for that specific arm

### For `if` keyword:
- All `if` keywords in the `if-else` chain
- All return expressions in the if-else branches

This feature works with nested expressions and macros.

### DIFF
--- a/crates/ide/src/goto_definition.rs
+++ b/crates/ide/src/goto_definition.rs
@@ -295,6 +295,7 @@ fn handle_control_flow_keywords(
         T![for] if token.parent().and_then(ast::ForExpr::cast).is_some() => {
             nav_for_break_points(sema, token)
         }
+        T![match] | T![=>] | T![if] => nav_for_branches(sema, token),
         _ => None,
     }
 }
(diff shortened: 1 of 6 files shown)


roife:
Here's a demo.

https://github.com/user-attachments/assets/42048fe1-7ffc-469b-bf27-c58ce0759236

crates/rust-analyzer/src/config.rs
```diff
@@ -93,6 +93,8 @@ config_data! {
 
 
 
+        /// Enables highlighting of related return values while the cursor is on any `match`, `if`, or match arm arrow (`=>`).
+        highlightRelated_branches_enable: bool = true,
```
Veykril, line 97: Maybe lets call all these configs for branches `branch_exit_points`?

crates/ide/src/references.rs
```diff
@@ -309,7 +309,10 @@ fn handle_control_flow_keywords(
     let file = sema.parse_guess_edition(file_id);
     let edition =
         sema.attach_first_edition(file_id).map(|it| it.edition()).unwrap_or(Edition::CURRENT);
-    let token = file.syntax().token_at_offset(offset).find(|t| t.kind().is_keyword(edition))?;
+    let token = file
+        .syntax()
+        .token_at_offset(offset)
+        .find(|t| t.kind().is_keyword(edition) || t.kind() == T![=>])?;
```
Veykril, line 315: Lets use the pick function here https://github.com/rust-lang/rust-analyzer/blob/aded72f8e6048b243e20e0b77a3870b1b76ef62d/crates/ide-db/src/helpers.rs#L28

crates/ide/src/goto_definition.rs
```diff
@@ -404,6 +405,70 @@ fn nav_for_exit_points(
...
+            })
+            .flatten()
+            .collect_vec(),
+
+        _ => return Some(Vec::new()),
+    };
+
+    Some(navs)
+}
+
```
Veykril, line 471: judging by the tests, this doesn't really do anything signifcant.  Can we drop this?

crates/ide/src/highlight_related.rs
```diff
@@ -285,11 +290,129 @@ fn highlight_references(
     if res.is_empty() { None } else { Some(res.into_iter().collect()) }
 }
 
+pub(crate) fn highlight_branches(
```
Veykril, line 293: why is hte implementation of this so different from the test. I'd expect this to look fairly similar to `hl_exit_points`

crates/rust-analyzer/src/config.rs
roife: Yes, `branch_exit_points` is much better! 😄

crates/ide/src/goto_definition.rs
roife: The purpose here is "when a user performs a goto-def on the `match` kw, server should navigate to the `match` kw itself."

This approach is more from a user experience perspective. In VS Code, if the user perform a goto-def on an identifier and the response is the identifier itself, VS Code would understand that this location is a definition. Therefore, VS Code will automatically trigger find-references.

This is very intuitive for the user: when I perform `ctrl + click` on `match` kw, it will automatically trigger find-references to locate all branch exit points.

The current logic here is indeed quite complicated. I will try to simplify it to make it look more elegant.

crates/ide/src/highlight_related.rs
```diff
@@ -302,11 +307,93 @@ fn highlight_references(
...
+                            push_tail_expr(Some(else_block.into()), &mut highlights);
+                            if_to_process = None;
+                        }
+                        Some(ast::ElseBranch::IfExpr(nested_if)) => if_to_process = Some(nested_if),
+                        None => if_to_process = None,
+                    }
+                }
+            }
+        }
+        _ => unreachable!(),
```
Veykril, line 383:
```suggestion
        _ => return Default::default(),
```

Veykril (approved):

roife:
I've been busy with work lately... Apologize for the delayed response.

Veykril (approved):
Thanks

PR #19550, opened 2025-04-09
refactor: Turn `LifetimeRef` into an enum

Veykril:
This makes things more structured

### DIFF
--- a/crates/hir-def/src/hir/type_ref.rs
+++ b/crates/hir-def/src/hir/type_ref.rs
@@ -6,7 +6,6 @@ use std::fmt::Write;
 use hir_expand::name::Name;
 use intern::Symbol;
 use la_arena::Idx;
-use syntax::ast;
 use thin_vec::ThinVec;
 
 use crate::{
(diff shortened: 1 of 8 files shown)


crates/hir-def/src/hir/type_ref.rs
```diff
@@ -145,18 +144,11 @@ const _: () = assert!(size_of::<TypeRef>() == 16);
...
-        LifetimeRef { name: Name::new_lifetime(lifetime) }
-    }
-
-    pub fn missing() -> LifetimeRef {
-        LifetimeRef { name: Name::missing() }
-    }
+pub enum LifetimeRef {
+    Named(Name),
+    Static,
+    Infer,
```
davidbarsky, line 150: For consistency with `Named`, I think this should be `Inferred`?

```suggestion
    Inferred
```

davidbarsky (approved):
lgtm, small nit

crates/hir-def/src/hir/type_ref.rs
Veykril: We used `Infer` for `TypeRef` as well

davidbarsky: Oh, well. I mean, up to you.

Veykril: Actually I'll name it `Placeholder` to mirror `TypeRef`

PR #19566, opened 2025-04-10
Fix dyn compatibility code bypassing callable_item_signature query

flodiebold:


ShoyuVanilla (commented):
I was too late for this - I've been busy these days fiddling with my new computer 😅  - but thank you for fixing my mistake!

PR #19568, opened 2025-04-10
refactor: editor for `destructure_struct_binding`

snprajwal:
Another one bites the dust

### DIFF
--- a/crates/syntax/src/ast/make.rs
+++ b/crates/syntax/src/ast/make.rs
@@ -643,8 +643,8 @@ pub fn expr_method_call(
 ) -> ast::Expr {
     expr_from_text(&format!("{receiver}.{method}{arg_list}"))
 }
-pub fn expr_macro_call(f: ast::Expr, arg_list: ast::ArgList) -> ast::Expr {
-    expr_from_text(&format!("{f}!{arg_list}"))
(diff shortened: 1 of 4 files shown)


crates/syntax/src/ast/make.rs
```diff
@@ -643,8 +643,8 @@ pub fn expr_method_call(
 ) -> ast::Expr {
     expr_from_text(&format!("{receiver}.{method}{arg_list}"))
 }
-pub fn expr_macro_call(f: ast::Expr, arg_list: ast::ArgList) -> ast::Expr {
-    expr_from_text(&format!("{f}!{arg_list}"))
+pub fn expr_macro(path: ast::Path, arg_list: ast::ArgList) -> ast::MacroExpr {
```
Veykril, line 646: Hmm this sig was wrong before already
```suggestion
pub fn expr_macro(path: ast::Path, tt: ast::TokenTree) -> ast::MacroExpr {
```

Veykril (approved):
Lgtm except that the signature fix, which isn't your mistake, but might as well fix it wile we are at it.

crates/syntax/src/ast/make.rs
snprajwal: This breaks `ide-assists/src/utils/gen_trait_fn_body.rs` in a pretty nasty way. Could we do it in a follow-up PR? I'll create an issue to make sure we don't forget.

PR #19569, opened 2025-04-10
fix(ast): return correct types for `make::expr_*` methods

snprajwal:
Much nicer to work with now, less `unreachable!()`s in the syntax factory.


Veykril (approved):
Thanks!

Issue #19464, opened 2025-03-27
Regression: vs code test explorer skips unit tests since v0.3.2353


soerenkoehler:
**steps to reproduce**:

*   in VSCode install current version v0.3.2353 (bug still exists in v0.3.2370)
*   in test explorer click "Run Tests"
*   result:
    *   test explorer / rust-analyzer seem to test dependencies instead of my code  
        [ra-strange-test-output.txt](https://github.com/user-attachments/files/19493978/ra-strange-test-output.txt)
    *   integration tests run, but not the unit tests  
        ![Image](https://github.com/user-attachments/assets/553dbda9-c9d2-4220-92af-c7d288686223)

**counter check for verification**:

*   install v0.3.2345
*   in test explorer click "Run Tests"
*   result:
    *   test output looks normal  
        [ra-correct-test-output.txt](https://github.com/user-attachments/files/19494100/ra-correct-test-output.txt)
    *   tests run as expected  
        ![Image](https://github.com/user-attachments/assets/f8c2269e-a90d-4df3-ba1f-b4f38ed9c080)

**rust-analyzer version**:
Identifier: rust-lang.rust-analyzer
Version: **0.3.2353**
Last Updated: 2025-03-27, 21:29:57

**rustc version**: rustc **1.85.1** (4eb161250 2025-03-15)

**editor or extension**: VSCode (Win11, WSL2)
Version: 1.98.2 (user setup)
Commit: ddc367ed5c8936efe395cffeec279b04ffd7db78
Date: 2025-03-12T13:32:45.399Z
Electron: 34.2.0
ElectronBuildId: 11161602
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Windows_NT x64 10.0.22631

**relevant settings**: ---

**repository link (if public, optional)**: https://github.com/soerenkoehler/rs-chdiff/tree/bug-test-explorer-rust-analyzer

play-tank:
Also seeing this same issue. Same text messages as above. It appears that the code is trying to test all of the dependencies now, and not just the package.

The regression is possibly related to this:
```
error: unexpected argument '--custom-build' found

  tip: to pass '--custom-build' as a value, use '-- --custom-build'
```

I'm currently on VSCode Version: 1.99.2 (Universal), macOs Sequoia 15.3.2 (24D81)
[...]

PR #19571, opened 2025-04-11
Fix missing test update notifications when there are hyphens in the target name and exclude dependencies from `Run All`

duncanawoods:
#19464

Test notifications with targets containing hyphens are failing with the error:

    [error] [Window] [Extension Host] Notification handler 'experimental/changeTestState' failed with message: Cannot read properties of undefined (reading 'id')

The test id expects the namespace form with underscores `tokio_stream` instead of `tokio-stream`.

### DIFF
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -203,7 +203,7 @@ pub(crate) fn handle_view_item_tree(
 //   underscores. cargo test requires the real name.
 // - the target kind e.g. bin or lib
 fn find_test_target(namespace_root: &str, cargo: &CargoWorkspace) -> Option<TestTarget> {
-    cargo.packages().find_map(|p| {
+    cargo.packages().filter(|p| cargo[*p].is_member).find_map(|p| {
         let package_name = &cargo[p].name;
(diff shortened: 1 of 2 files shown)


HKalbasi:
Thanks for the quick fix!

Why checking for `is_member` is needed now?

duncanawoods:
Hi @HKalbasi 

> Why checking for is_member is needed now?

Previously Test Explorer 'Run All'  would execute `cargo test` with a `--workspace` target. This had the drawbacks that:

- RA had to guess at which package each result was for using `hack_recover_crate_name`
- you get no test results if any target in the workspace cannot be compiled

The previous change fixed these by executing tests for each target in the workspace seperately:

- no hacks required to match test results up to targets
- tests results are generated for all packages that compile even if some don't

My mistake was assuming `CargoWorkspace::packages()` was just the members. I didn't realise it included dependencies so "running every target in the workspace" became every dependency too.

I managed to miss this because:

- I had tested `Run All` in the workspace I was using as a test case which had a variety of package and target types but... did not include any external dependencies
- If I want to `Run All` tests, I use the command line so I didn't see it during my normal use. I use the vscode test runner for specific modules/tests during development

PR #19576, opened 2025-04-14
Ignore errors from rustfmt which may trigger error notification

### LINKED ISSUE #19574: A missing closing double quote will trigger error notification
For a scenario like this:

Image

There is a missing right quote on the cursor position(simple source code can not reproduce, we need input a single quote before some other code), it will trigger a notification like this:

Image

Click the setting and "Turn off all notifications from rust-analyzer" does not work for this scenario.

Click "Go to output" shows some errors from `rustfmt`:

Image

It's a distraction when inputting, because if it's shows up I must need to move mouse to close the dialog, is there any way to mute this kind of error notification?

Disable this option also does not work:

Image

Disable "Format on save" will resolve it, but I'd like to enable "Format on save".

chenyukang:
Fixes #19574

### DIFF
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -2418,7 +2418,11 @@ fn run_rustfmt(
                 Ok(None)
             }
             // rustfmt panicked at lexing/parsing the file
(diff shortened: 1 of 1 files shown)


chenyukang:
link #18663

Veykril (approved):
Thanks!

PR #19582, opened 2025-04-14
Distribute x64 and aarch64 Linux builds with PGO optimizations

Kobzol:
Empirically this seems to make the resulting `analysis-stats` command ~20% faster.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -38,10 +38,12 @@ jobs:
             target: x86_64-unknown-linux-gnu
             zig_target: x86_64-unknown-linux-gnu.2.28
             code-target: linux-x64
+            pgo: true
           - os: ubuntu-latest
             target: aarch64-unknown-linux-gnu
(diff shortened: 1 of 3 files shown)


Veykril (approved):
Nice, thanks!

Kobzol:
I haven't really tested this on CI (only locally), so maybe it would also be a good idea to test the `release` workflow using manual execution (`workflow_dispatch`) to test that it works properly.

ofek:
Does this not work on Windows?

FilipAndersson245:
I believe there should be nothing limiting the use of pgo on windows? but this only enables it for linux.

ofek:
Is there a reason why Windows was not enabled?

Kobzol:
It doesn't seem to work with `cargo zigbuild` :( (https://github.com/rust-lang/rust-analyzer/actions/runs/14448767429/job/40515878851) I guess we should revert it for now.

PR #19583, opened 2025-04-14
Do not perform PGO on Linux CI

Kobzol:
Reverts a part of rust-lang/rust-analyzer#19582. PGO does not seem to work with `cargo zigbuild` (https://github.com/rust-lang/rust-analyzer/actions/runs/14448767429/job/40515878851).


Veykril (commented):
can we just undo the ci stuff but keep the xtask changes? That way people can still build a pgo server if they desire themselves if I see this right

Kobzol:
Good point, I only reverted the Linux PGO change, and kept the rest. I'll try to enable it on Windows in a separate PR.

Veykril (approved):

PR #19585, opened 2025-04-14
Allow training PGO on a custom crate and enable it Windows on CI

Kobzol:
This PR adds functionality to train PGO on a custom crate downloaded from GitHub (e.g. `--pgo clap-rs/clap`), and enables PGO for Windows on CI.

I added the functionality because training on itself produced OOMs on Windows CI.

I didn't enable PGO for `aarch64-pc-windows-msvc`, because there it's more complicated because of the cross-compilation.

If someone wants to test the performance effect, they could compare the following builds:
- PGO: https://github.com/rust-lang/rust-analyzer/actions/runs/14465225022
- No PGO: https://github.com/rust-lang/rust-analyzer/actions/runs/14458498119

Download and extract the `dist-x86_64-pc-windows-msvc` artifact, and then run `rust-analyzer.exe analysis-stats --run-all-ide-things <rust-analyzer-checkout-directory>`.

Related issue: https://github.com/rust-lang/rust-analyzer/issues/9412

### DIFF
--- a/xtask/src/dist.rs
+++ b/xtask/src/dist.rs
@@ -189,10 +189,8 @@ fn gather_pgo_profile<'a>(
     eprintln!("Training RA on {label}...");
     cmd!(
         sh,
-        "target/{target}/release/rust-analyzer analysis-stats --run-all-ide-things {train_path}"
+        "target/{target}/release/rust-analyzer analysis-stats -q --run-all-ide-things {train_path}"
     )
-    // analysis-stats produces an enormous amount of output on stdout
-    .ignore_stdout()
     .run()
     .context("cannot generate PGO profiles")?;
 
--- a/xtask/src/dist.rs
+++ b/xtask/src/dist.rs
@@ -168,7 +168,7 @@ fn gather_pgo_profile<'a>(
         .read()
         .context("cannot resolve target-libdir from rustc")?;
     let target_bindir = PathBuf::from(target_libdir).parent().unwrap().join("bin");
-    let llvm_profdata = target_bindir.join(format!("llvm-profdata{}", EXE_EXTENSION));
+    let llvm_profdata = target_bindir.join("llvm-profdata").with_extension(EXE_EXTENSION);
 
     // Build RA with PGO instrumentation
     let cmd_gather =
--- a/xtask/src/dist.rs
+++ b/xtask/src/dist.rs
@@ -12,6 +12,7 @@ use time::OffsetDateTime;
 use xshell::{Cmd, Shell, cmd};
 use zip::{DateTime, ZipWriter, write::SimpleFileOptions};
 
+use crate::flags::PgoTrainingCrate;
 use crate::{
     date_iso,
     flags::{self, Malloc},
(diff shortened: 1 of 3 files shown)


Kobzol:
Tested that now with `clap` PGO works for x64 and i686 Windows.

r? @Veykril

xtask/src/dist.rs
```diff
@@ -168,19 +171,28 @@ fn gather_pgo_profile<'a>(
...
+            (download_crate_for_training(sh, &pgo_dir, url)?, url.as_str())
+        }
+    };
+
+    // Run RA either on itself or on a downloaded crate
+    eprintln!("Training RA on {label}...");
     cmd!(
         sh,
-        "target/{target}/release/rust-analyzer analysis-stats {train_crate} --run-all-ide-things"
+        "target/{target}/release/rust-analyzer analysis-stats --run-all-ide-things {train_path}"
```
lnicola, line 192:
```suggestion
        "target/{target}/release/rust-analyzer -q analysis-stats --run-all-ide-things {train_path}"
```

Then we don't need `ignore_stdout()`.

lnicola (approved):
I wonder if training on the standard library wouldn't be sufficient, but this looks reasonable too.

Kobzol:
That would require looking up the sources, which is IMO similar complexity to downloading a crate from github. Also "normal" crates might be a teeny tiny bit more representative than stdlib, although here it probably doesn't matter that much.

zamazan4ik:
I've made measurements on my Windows machine.

Environment:

* OS: Windows 11, the latest at the moment stable build
* CPU: AMD Ryzen 9 5900x
* Disk: SSD
* Rust-Analyzer versions: from https://github.com/rust-lang/rust-analyzer/pull/19585#issue-2993481774
* Latest at the moment Rust Analyzer sources

For benchmark purposes, I used `rust-analyzer.exe analysis-stats --run-all-ide-things <rust-analyzer-checkout-directory>` command. Measurements are done multiple times, in different orders, etc. First runs are skipped due to "Antimalware Service Executable" thing (I was too lazy to add the binaries to the exceptions). The results are stable across runs.

Results (the `Total` number at the of the execution. Providing the full log will be too lengthy):

* No PGO: 200s
* PGO: 170s

Nice improvement!

Kobzol:
@lnicola Are you fine with running on a custom crate? It seems like it takes ~20 minutes, but since the `release` workflow runs once per day, that shouldn't really be an issue IMO.

lnicola:
Yeah, just me give an hour or so, I want to test it, just in case.

Kobzol:
> Given the xtask dist time difference (386 s vs. 182 s)

Could you please post these timings, to get a better estimation of how long did `xtask` run? Or did you take this from CI?

lnicola:
`xtask dist --pgo clap-rs/clap`  took 182 s on my computer, while `xtask dist --pgo rust-analyzer` took 386 s.

Kobzol:
Oh, I see. Ok, I will use `clap` in a follow-up PR where I will enable PGO also for Linux then.

PR #19586, opened 2025-04-14
[WIP] Use a Docker container instead of Zig for building with old(er) glibc on Linux

Kobzol:
Test to see if it works.


lnicola:
Does `pypa/manylinux_2_28_x86_64` even bring us anything besides `almalinux:8`?

Kobzol:
It's a just a placeholder image that I use on some other crate to get glibc 2.28. We could use a different image for sure :)

Kobzol:
Great, PGO works (with 15% perf. win), and I checked that at most 2.28 GLIBC symbols are being used (https://github.com/rust-lang/rust-analyzer/actions/runs/14469656707?pr=19586):
```
objdump -T rust-analyzer-pgo/rust-analyzer-x86_64-unknown-linux-gnu | grep GLIBC | sed 's/.*GLIBC_\([.0-9]*\).*/\1/g' | sort -Vu
```

I didn't yet enable PGO in this PR, I want to do that separately.

r? @lnicola

.github/workflows/release.yaml
```diff
@@ -71,6 +73,12 @@ jobs:
         with:
           node-version: 22
 
+      - name: Install rustup
+        if: ${{ matrix.container }}
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
lnicola, line 79: Maybe let try `--profile minimal` so we don't get the docs?

lnicola:
We could also try `quay.io/pypa/manylinux_2_34_aarch64`? But it doesn't have to be the same PR. These images are really confusing anyway:


```sh
$ podman run --rm -it quay.io/pypa/manylinux_2_34_aarch64
[snip]
WARNING: image platform (linux/arm64) does not match the expected platform (linux/amd64)
[root@fa86d82151b1 /]#
```

And it's Alma 9, not 8 like the x86-64 one.

Kobzol:
Ok, let's try!

Kobzol:
Hmm, looks like with cross-compilation it's not so easy. Let's just do x64 for now then.

lnicola:
> Error response from daemon: container 6d9810bb76206a2890d94b6fe1d59f91ce996906f62b50fddb2f024051f5671d is not running

What?

lnicola:
It's really weird, I get the same error (container is not running) locally even with the x86-64 image:

```
$ docker create --entrypoint tail quay.io/pypa/manylinux_2_34_x86_64 -f /dev/null
7f757bf3eb98a92723b0435457b9768ce5ca9ba92f950333f5b0f1d9aa2c5a85
$ docker exec 7f757bf3eb98a92723b0435457b9768ce5ca9ba92f950333f5b0f1d9aa2c5a85 sh -c "cat /etc/*release | grep ^ID"
Error response from daemon: container 7f757bf3eb98a92723b0435457b9768ce5ca9ba92f950333f5b0f1d9aa2c5a85 is not running
```

Kobzol:
I can run the container locally just fine. I wouldn't worry about it, tbh, the cross-compilation part is clearly what's the issue, and I don't want to deal with that right now, x64 is much more important anyway for the optimization, as it's used by many more people.

lnicola:
I can run it too, but not like they do on GHA. Anyway, I was just hoping we could drop the Zig stuff at one point.

PR #19587, opened 2025-04-15
Remove incorrect doc comment

MatrixFrog:
I just happened to come across this while investigating an error.

It looks like this comment was mistakenly copied from the function right above it. I'm happy to add a better comment if you like, but I think it's fairly clear from the function name and type signature.


Veykril (approved):
Thanks!

PR #19589, opened 2025-04-15
fix: highlighting for tail expr in labelled blocks

### LINKED ISSUE #19138: Missing exit point highlighting for labeled block tail
```rs
        'a: {
            if true { break 'a None}
            None // this exit point is not being highlighted
        }
```

When the cursor is on `'a` or the `break` we do not highlighting the tail `None` even though it is also an exit point for the block.

roife:
fix rust-lang/rust-analyzer#19138.

This PR adds highlighting for tail expressions of labeled blocks.

### DIFF
--- a/crates/ide/src/highlight_related.rs
+++ b/crates/ide/src/highlight_related.rs
@@ -446,6 +446,18 @@ pub(crate) fn highlight_break_points(
                 push_to_highlights(file_id, text_range);
             });
 
+        if matches!(expr, ast::Expr::BlockExpr(_)) {
+            for_each_tail_expr(&expr, &mut |tail| {
+                if matches!(tail, ast::Expr::BreakExpr(_)) {
+                    return;
+                }
+
+                let file_id = sema.hir_file_for(tail.syntax());
+                let range = tail.syntax().text_range();
+                push_to_highlights(file_id, Some(range));
+            });
+        }
+
         Some(highlights)
     }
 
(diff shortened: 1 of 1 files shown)


crates/ide/src/highlight_related.rs
```diff
@@ -232,6 +232,22 @@ fn highlight_references(
             }
         }
 
+        if matches!(def, Definition::Label(_)) {
+            let label = token.parent_ancestors().find_map(ast::Label::cast);
```
Veykril, line 236:
```suggestion
            let label = token.parent().and_then(ast::Label::cast);
```

crates/ide/src/highlight_related.rs
```diff
@@ -232,6 +232,22 @@ fn highlight_references(
             }
         }
 
+        if matches!(def, Definition::Label(_)) {
```
Veykril, line 235: We should start commenting this file a bit more (I am partially at fault for not doing so here :)

So lets put a comment above this stating what this is doing, highlighting the tail expr of labelled blocks

crates/ide/src/highlight_related.rs
roife: The hierarchy of an `ast::Label` is structured as `(Label (Lifetime ident) colon)`, which suggests that the `token` here may refer to the `ident` within `Lifetime`.

If we don't use `token.parent_ancestors`, we should write:

```rust
let label = token.parent().and_then(|token| {
    ast::Label::cast(token.clone())
        .or_else(|| token.parent().and_then(ast::Label::cast))
});
```

... which is more complicated 🤔

Veykril: Ah right, lifetime is a node as well, then lets do
```suggestion
let label = token.parent_ancestors().nth(1).and_then(ast::Label::cast);
```

Veykril (approved):

PR #19591, opened 2025-04-15
fix: use `ast::TokenTree` in `make::expr_macro`

### LINKED ISSUE #19584: [syntax] `make::expr_macro_call` should receive a token tree instead of arg list
https://github.com/rust-lang/rust-analyzer/blob/aded72f8e6048b243e20e0b77a3870b1b76ef62d/crates/syntax/src/ast/make.rs#L646

The correct way to build a macro call is with the token tree - it need not be an arg  list.

snprajwal:
Fixes rust-lang/rust-analyzer#19584

### DIFF
--- a/crates/syntax/src/ast/make.rs
+++ b/crates/syntax/src/ast/make.rs
@@ -32,12 +32,9 @@ pub mod ext {
     use super::*;
 
     pub fn simple_ident_pat(name: ast::Name) -> ast::IdentPat {
-        return from_text(&name.text());
-
-        fn from_text(text: &str) -> ast::IdentPat {
-            ast_from_text(&format!("fn f({text}: ())"))
-        }
+        ast_from_text(&format!("fn f({}: ())", name.text()))
     }
+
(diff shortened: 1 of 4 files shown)


crates/syntax/src/ast/make.rs
```diff
@@ -116,6 +112,10 @@ pub mod ext {
     pub fn ty_result(t: ast::Type, e: ast::Type) -> ast::Type {
         ty_from_text(&format!("Result<{t}, {e}>"))
     }
+
+    pub fn token_tree_from_arg_list(arg_list: ast::ArgList) -> ast::TokenTree {
```
Veykril, line 116: We can generalize this I think
```suggestion
    pub fn token_tree_from_node(node: &SyntaxNode) -> ast::TokenTree {
```
we effectively just want to turn any nodes tokens into a `TokenTree` so might as well make this API "untyped" in that regard

snprajwal: Nice catch, updated!

Veykril (approved):
thanks!

PR #19595, opened 2025-04-15
Use PGO on Linux x64 builds

Kobzol:
After https://github.com/rust-lang/rust-analyzer/pull/19586 it should be possible to do PGO on Linux.


lnicola (approved):

berkus:
> Thank you so much for setting this up @Kobzol! Up to 20% is quite significant.

It's a big [drop in the ocean](https://github.com/rust-lang/rust-analyzer/issues/9412#issuecomment-1298277389), indeed. Thank you @Kobzol!

PR #19596, opened 2025-04-15
feat: enhance signature help to display generic parameters for callables and default values for generic args

### LINKED ISSUE #12729: Generics signature help should include type parameter defaults
For example, in `std::ops::ControlFlow::<$0`, the signature help is rendered as `enum ControlFlow<B, C>`, which does not tell the user that `C` is optional and only `B` needs to be provided. Rendering the default type would fix that, and result in `enum ControlFlow<B, C = ()>`.

It's possible that a default type would be too long when fully rendered, so the length should probably be limited (the `HirDisplay` machinery can already do that).

roife:
fix rust-lang/rust-analyzer#12729, rust-lang/rust-analyzer#19373.

### DIFF
--- a/crates/ide/src/signature_help.rs
+++ b/crates/ide/src/signature_help.rs
@@ -5,13 +5,15 @@ use std::collections::BTreeSet;
 
 use either::Either;
 use hir::{
-    AssocItem, DisplayTarget, GenericParam, HirDisplay, ModuleDef, PathResolution, Semantics, Trait,
+    AssocItem, DisplayTarget, GenericDef, GenericParam, HirDisplay, ModuleDef, PathResolution,
+    Semantics, Trait,
 };
 use ide_db::{
     FilePosition, FxIndexMap,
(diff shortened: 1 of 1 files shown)


crates/ide/src/signature_help.rs
```diff
@@ -2394,4 +2443,96 @@ fn main() {
...
+    Some(T),
+    None,
+}
+
+fn main() {
+    let opt = Option::Some($0);
+}
+            "#,
+            expect![[r#"
+                enum Option<T>::Some({unknown})
```
Veykril, line 2498: Would be nice if this said `enum Option<T>::Some(T)`, we should open a follow up issue for that

Veykril (approved):
On a related note, it would be nice if we'd show monomorphized signatures, that is, drop the generic params that are inferred from context. Sounds difficult to do (and we still dont do this properly for hovers even).

PR #19597, opened 2025-04-15
Build aarch64 builds on CI with PGO

Kobzol:
Uses the manylinux aarch64 image + a native ARM GitHub runner to compile aarch64 builds with PGO.


Kobzol:
I haven't found a manylinux image for the last arm runner, but at least aarch64 seems to work.

lnicola (approved):

Kobzol:
Looks like a rustc sync is needed after https://github.com/rust-lang/rust/pull/139671.

Kobzol:
This can be requeued now that the proc macro error has been fixed, I think.

PR #19600, opened 2025-04-16
Pin rustc used for the `proc-macro-src` CI job

Kobzol:
Fixes: https://github.com/rust-lang/rust-analyzer/issues/19599


Veykril (approved):
Thanks!

PR #19602, opened 2025-04-16
Use PGO for 32-bit ARM builds

Kobzol:


lnicola:
TBH I don't even know which armhf distros people use these days.

Kobzol:
Bah, I tried to find the corresponding packages on EL, but wasn't successful. I was able to get this working on Ubuntu 18.04 with glibc 2.27. Let me know if you want to go that route, or keep zigbuild. Up to you :)

Kobzol:
Actually, I realized now that we can't actually do PGO for 32-bit ARM on x64, because we won't be able to execute the compiled binary :man_facepalming: I guess it might be possible on 64-bit ARM though? Not sure if aarch64 can execute arm-eabihf though.

Kobzol:
Ubuntu 18.10 has glibc 2.28, so that would solve this, but there is still the small issue of not being able to execute the compiled code xD I'll try it on aarch64.

Kobzol:
Nope, Ubuntu has exhausted my patience.

PR #19611, opened 2025-04-16
Use PGO on macOS x64 and arm64 builds

zamazan4ik:
- related to https://github.com/rust-lang/rust-analyzer/issues/9412#issuecomment-2810772246

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -55,9 +55,11 @@ jobs:
           - os: macos-13
             target: x86_64-apple-darwin
             code-target: darwin-x64
+            pgo: clap-rs/clap
           - os: macos-13
             target: aarch64-apple-darwin
             code-target: darwin-arm64
+            pgo: clap-rs/clap
 
     name: dist (${{ matrix.target }})
     runs-on: ${{ matrix.os }}


lnicola:
You can test this by adding `on: pull_request` and making it fail like in https://github.com/rust-lang/rust-analyzer/pull/19602/files.

lnicola:
I guess it's easier to merge and revert if necessary.

zamazan4ik:
So, it doesn't work: https://github.com/rust-lang/rust-analyzer/actions/runs/14510906031/job/40709149737

As far as I understand, this due to that we use x86-64 macOS image instead of ARM image: https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md . Are we allowed to use ARM macOS machines for RA? If yes, we can try to change the image and try once again.

lnicola:
#19614

You can already test it in the nightly.

zamazan4ik:
PGO world is not so big nowadays, yeah :D

lnicola:
And look:

image

zamazan4ik:
As I did previously in https://github.com/rust-lang/rust-analyzer/pull/19585#issuecomment-2804752547 for Windows - now we test PGO vs non-PGO RA for macOS!

Environment:

* OS: macOS 15.4 with the latest stable updates
* Hardware: Macbook M1 Pro (ARM CPU)
* Rust-Analyzer versions: PGO version from https://github.com/rust-lang/rust-analyzer/actions/runs/14511602298 , non-PGO version from https://github.com/rust-lang/rust-analyzer/actions/runs/14505288557 (the latest build before PGO enable for macOS, no other performance-related changes were committed since this so we can use this artifact)
* Latest at the moment Rust Analyzer sources

For benchmark purposes, I used `rust-analyzer analysis-stats --run-all-ide-things <rust-analyzer-checkout-directory>` command. Measurements are done multiple times, in different orders, etc. First runs are skipped to avoid any first-time-run fluctuations (like "XProtectService" or other "magic" stuff). The results are pretty stable across runs.

Results (the `Total` number at the end of the execution. Providing the full log will be too lengthy):

* No PGO: 140s
* PGO: 110s

So we can conclude that PGO for RA works well for macOS platform too.

I didn't test x86 macOS version (I am a bit lazy to run it under Rosetta2) but I expect exactly the same performance wins for that platform too. As far as I understand, RA in CI uses x86 macOS images to build x86 macOS RA version, so no PGO issues due to cross-compilation should appear. If anyone wants to test it and report the results back - it would be nice :)

PR #19618, opened 2025-04-18
internal: Support specific revisions for PGO

lnicola:
This should make the builds more reproducible.

### DIFF
--- a/xtask/src/dist.rs
+++ b/xtask/src/dist.rs
@@ -180,8 +180,8 @@ fn gather_pgo_profile<'a>(
 
     let (train_path, label) = match &train_crate {
         PgoTrainingCrate::RustAnalyzer => (PathBuf::from("."), "itself"),
-        PgoTrainingCrate::GitHub(url) => {
-            (download_crate_for_training(sh, &pgo_dir, url)?, url.as_str())
+        PgoTrainingCrate::GitHub(repo) => {
+            (download_crate_for_training(sh, &pgo_dir, repo)?, repo.as_str())
         }
     };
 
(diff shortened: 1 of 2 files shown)


xtask/src/dist.rs
```diff
@@ -212,12 +212,18 @@ fn gather_pgo_profile<'a>(
...
 
 /// Downloads a crate from GitHub, stores it into `pgo_dir` and returns a path to it.
-fn download_crate_for_training(sh: &Shell, pgo_dir: &Path, url: &str) -> anyhow::Result<PathBuf> {
-    let normalized_path = url.replace("/", "-");
+fn download_crate_for_training(sh: &Shell, pgo_dir: &Path, repo: &str) -> anyhow::Result<PathBuf> {
+    let mut it = repo.splitn(2, '@');
+    let repo = it.next().unwrap();
+    let revision = it.next();
+    let revision =
+        if let Some(revision) = revision { &["--revision", revision] as &[&str] } else { &[] };
```
Kobzol, line 220: I don't think that `git clone` supports a `--revision` parameter :thinking: I think that `--branch` should work though.

Kobzol (commented):
Great idea!

xtask/src/dist.rs
lnicola: Mine does, but you make a good point: https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-code--revisionltrevgtcode. It's a bit better than `--branch` because it also supports commits, not just branches and tags.

I've changed it to `--branch` and added a FIXME.

Kobzol: Ah, probably it was added in a later version. I have `2.43.0` and it doesn't support the flag.

Kobzol (approved):

xtask/src/dist.rs
lnicola: Yeah, looks like it's from `2.49`, released.. last month :sweat_smile:.

PR #19622, opened 2025-04-18
Fix ide-assists raw_string suffix fail

A4-Tacks:
Fixes rust-lang/rust-analyzer#19621 
Fixes rust-lang/rust-analyzer#19625

### DIFF
--- a/crates/ide-assists/src/handlers/raw_string.rs
+++ b/crates/ide-assists/src/handlers/raw_string.rs
@@ -2,7 +2,10 @@ use std::borrow::Cow;
 
 use syntax::{AstToken, TextRange, TextSize, ast, ast::IsString};
 
-use crate::{AssistContext, AssistId, Assists, utils::required_hashes};
+use crate::{
+    AssistContext, AssistId, Assists,
+    utils::{required_hashes, string_suffix},
+};
 
 // Assist: make_raw_string
 //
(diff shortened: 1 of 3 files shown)


crates/ide-assists/src/handlers/raw_string.rs
```diff
@@ -33,12 +36,15 @@ pub(crate) fn make_raw_string(acc: &mut Assists, ctx: &AssistContext<'_>) -> Opt
         target,
         |edit| {
             let hashes = "#".repeat(required_hashes(&value).max(1));
+            let range = token.syntax().text_range();
+            let suffix = string_suffix(token.text()).unwrap_or_default();
+            let range = TextRange::new(range.start(), range.end() - TextSize::of(suffix));
```
Veykril, line 41: Can we add this stuff as a method on `IsString`? `text_range_with_quotes` or something like that

Veykril: No there isn't, we should probably implement `IsString` for `Char` as well though (and then consider renaming the trait)

A4-Tacks: It seems that `IsString::quote_offsets` is not just extracting suffixes, but extracting quotation marks, which can be difficult to handle for `#`

A4-Tacks: `string_suffix(r###"r##"x"##i32"###)` -> `"i32"`
`QuoteOffsets::new(r###"r##"x"##i32"###)` -> range `"r##\""`, range `"\"##i32"`, and range `"x"`

Veykril: Ah okay, well 'lets merge as is then

Veykril:
Thanks!

PR #19624, opened 2025-04-19
Update chalk

jackh726:
The majority of the changes here are switching the order of generics (to be closer to rustc with self generics first).


davidbarsky (approved):

crates/hir-ty/src/generics.rs
Veykril, line 1: There is a comment at the top of this file that needs to be adjusted now I think

Veykril (changes requested):

PR #19629, opened 2025-04-19
allow using `null` to unset an environment variable

jyn514:
this makes three changes:
- all callsites of `toolchain::command` are changed to use `command(path, extra_env)`, instead of manually adding the env after the fact.
- all `map<str, str>` are changed to `map<str, option<str>>`.
- `command` checks for None and calls `env_remove` if so.

this caught several places where environment variables weren't being propagated:
- when running `rustc --print=target-libdir`
- when running `cargo rustc -- --print=target-spec-json`
- when running the custom DiscoverLinkedProjects config. I *think* this is for use with non-cargo build systems, so I didn't change it.

i verified locally that this actually works.

fixes https://github.com/rust-lang/rust-analyzer/issues/19626

### DIFF
--- a/crates/load-cargo/src/lib.rs
+++ b/crates/load-cargo/src/lib.rs
@@ -66,7 +66,7 @@ pub fn load_workspace_at(
 
 pub fn load_workspace(
     ws: ProjectWorkspace,
-    extra_env: &FxHashMap<String, String>,
+    extra_env: &FxHashMap<String, Option<String>>,
     load_config: &LoadCargoConfig,
 ) -> anyhow::Result<(RootDatabase, vfs::Vfs, Option<ProcMacroClient>)> {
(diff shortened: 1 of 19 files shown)


jyn514:
cc @davidbarsky for the DiscoverCommand changes; should cargo.extraEnv or check.extraEnv apply there? right now they don’t.

davidbarsky:
@jyn514 Sorry for the delay. I—weakly—don’t think that environment variables should apply to `DiscoverCommand` because that functionality was designed to be used with hermetic build systems like Bazel and Buck, where changing environment variables shouldn’t really have an impact on the first place.

I think if we do want to add environment variable support to `DiscoverCommand` (and that’s a big if!), it shouldn’t be through `Cargo`-scoped configuration.

Veykril (approved):
Thanks!

PR #19634, opened 2025-04-20
feat: Allow unsetting env vars in `server.extraEnv` config

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/19627


editors/code/src/config.ts
```diff
@@ -213,12 +213,14 @@ export class Config {
...
-            this.get<{ [key: string]: string | number } | null>("server.extraEnv") ?? {};
+            this.get<{ [key: string]: { toString(): string } | undefined | null } | null>(
+                "server.extraEnv",
+            ) ?? {};
         return substituteVariablesInEnv(
             Object.fromEntries(
                 Object.entries(extraEnv).map(([k, v]) => [
                     k,
-                    typeof v !== "string" ? v.toString() : v,
+                    v ? (typeof v !== "string" ? v.toString() : v) : null,
```
jyn514, line 223: i don’t think this is right, consider for example `”RUST_TEST_THREADS”: 0`

Veykril: That should set it `RUST_TEST_THREADS` to `0` which it still does no?

Veykril: Ah, nevermind I see the issue ugh. I hate these languages

Veykril:
@BenjaminBrienen the typescript CI seems to fail now on the `--experimental-strip-types` you added recently, any idea how to fix this? (I don't know the node ecosystem much so)

BenjaminBrienen:
> @BenjaminBrienen the typescript CI seems to fail now on the `--experimental-strip-types` you added recently, any idea how to fix this? (I don't know the node ecosystem much so)

It just wants you to run prettier. `npm run format` will run the script called format from package.json :)

Veykril:
Well no, it is formatted. `format:check` chokes on the warning generated by the experimental flag

Veykril:
Oh, nevermind. It only reports the files that have formatting erorrs on them, but not the formatting issues themselves ...

PR #19636, opened 2025-04-20
fix: Fix completion_snippets_custom config always erroring

Veykril:


Veykril:
@lnicola I see I was too slow with realizing this but I think we need to re-release this as another stable once merged. It will be really annoying for people otherwise

cujomalainey:
> @lnicola I see I was too slow with realizing this but I think we need to re-release this as another stable once merged. It will be really annoying for people otherwise

This is also broken on Homebrew right now

PR #19640, opened 2025-04-20
feat: show full type in tooltips for hints

### LINKED ISSUE #19615: Show the collapsed section in tooltips for inlay hints.
> How about adding a tooltip to `...` that displays the collapsed part or the entire type? Currently, the only way to view hidden parts is to select the expression and hover over it. 

 _Originally posted by @roife in [#11702](https://github.com/rust-lang/rust-analyzer/issues/11702#issuecomment-2813403981)_

roife:
fix rust-lang/rust-analyzer#19615.

### DIFF
--- a/crates/hir-ty/src/display.rs
+++ b/crates/hir-ty/src/display.rs
@@ -105,6 +105,13 @@ pub trait HirWrite: fmt::Write {
     fn start_location_link(&mut self, _location: ModuleDefId) {}
     fn start_location_link_generic(&mut self, _location: GenericParamId) {}
     fn end_location_link(&mut self) {}
+    /// Returns whether a new truncated part is created during this call. If so, the caller should
+    /// call `end_truncated` at the end. This ensures that the truncated part is not nested, i.e.
+    /// we should not use [`TYPE_HINT_TRUNCATION`] inside a tooltip (a truncated part).
+    fn start_truncated(&mut self) -> bool {
+        false
+    }
+    fn end_truncated(&mut self, _placeholder: &str) {}
 }
 
 // String will ignore link metadata
@@ -174,6 +181,18 @@ impl<'db> HirFormatter<'_, 'db> {
 
     pub fn end_location_link(&mut self) {
         self.fmt.end_location_link();
+    }
+
+    fn maybe_truncated<F: FnOnce(&mut Self) -> Result<(), HirDisplayError>>(
+        &mut self,
+        render_content_to: F,
+    ) -> Result<(), HirDisplayError> {
+        let start_truncated_part = self.should_truncate() && self.fmt.start_truncated();
+        let res = render_content_to(self);
+        if start_truncated_part {
+            self.fmt.end_truncated(TYPE_HINT_TRUNCATION);
+        }
+        res
     }
 
     fn format_bounds_with<F: FnOnce(&mut Self) -> Result>(
@@ -444,6 +463,7 @@ impl<'db> HirFormatter<'_, 'db> {
         sep: &str,
     ) -> Result {
         let mut first = true;
+        let mut in_truncated = false;
         for e in iter {
             if !first {
                 write!(self, "{sep}")?;
@@ -451,12 +471,17 @@ impl<'db> HirFormatter<'_, 'db> {
             first = false;
 
             // Abbreviate multiple omitted types with a single ellipsis.
-            if self.should_truncate() {
-                return write!(self, "{TYPE_HINT_TRUNCATION}");
+            if !in_truncated && self.should_truncate() && self.fmt.start_truncated() {
+                in_truncated = true;
             }
 
             e.hir_fmt(self)?;
         }
+
+        if in_truncated {
+            self.fmt.end_truncated(TYPE_HINT_TRUNCATION);
+        }
+
         Ok(())
     }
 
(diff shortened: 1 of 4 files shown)


roife:
image

crates/ide/src/inlay_hints.rs
```diff
@@ -524,9 +524,17 @@ pub enum InlayTooltip {
     Markdown(String),
 }
 
-#[derive(Default, Hash)]
+#[derive(Default)]
 pub struct InlayHintLabel {
     pub parts: SmallVec<[InlayHintLabelPart; 1]>,
+    pub tooltip: Option<LazyProperty<String>>,
```
Veykril, line 530: This should not be needed here, we should instead record the hover tooltip for the `...` parts (to make that work, we probably need to add a function to `HirWrite` that is called when the truncation string gets emitted).

Veykril (changes requested):

crates/ide/src/inlay_hints.rs
roife: If so, should we concatenate the types at the end? 👀 I tried adding a tooltip only for `...`, but it was quite inconvenient. Displaying the tooltip for the entire type directly is a better choice.

Veykril: The feature itself shouldn't require changes to our inlay hint structure here (and thus also no changes in `to_proto`). Associating a tooltip with all parts doesnt make sense when every part can track tooltips themselves.

I would expect that if `HirWrite` had a function like `fn start_truncated(&mut self)` (and end) that gets called when the formatting infra emits a `...` we could have that implemented as:

```rust
    fn start_truncated(&mut self) {
        self.make_new_part();
        // render full type into the new part tooltip here / (lazily)
    }

    fn end_truncated(&mut self) {
        self.make_new_part();
    }
```
and probably needs `InlayHintLabelBuilder` to get the full type as an additional field or so

roife: I implemented a prototype (the unit tests haven't been updated yet and the code needs refactoring), but I find it somewhat tricky. I would like to discuss potential improvements or alternative implementation ideas. 🤔

crates/ide/src/inlay_hints.rs
```diff
@@ -689,6 +691,25 @@ impl HirWrite for InlayHintLabelBuilder<'_> {
...
+        never!(self.tooltip.is_some(), "tooltip is already started");
+        self.make_new_part();
+    }
+
+    fn end_truncated(&mut self) {
+        self.tooltip = Some(if self.resolve {
+            self.last_part = "…".to_owned();
+            LazyProperty::Lazy
+        } else {
+            // TODO: tricky
```
Veykril: What is tricky here, swapping the buffer out seems correct to me?

crates/ide/src/inlay_hints.rs
Veykril: I think generally this feels right, couple improvements I could think of:

`start_truncated()` should probably return a bool / boolish type to mark whether the writer is interested in receiving an end marker or not (allows us to short circuit as before). If it returns true, we continue rendering and emit the `end_truncated` marker, if it returns false, we bail out and write the `...` to the writer (so we don't need to fixup other uses besides inlay hints of the formatting API).

THen to make the api a bit simpler, we could approach the truncation stuff with closures, `f.maybe_truncate(|f| { ... })` that only gets invoked if it won't get truncated, then we can pair the start end calls around the closure invocation in `maybe_truncate`.

roife: Previously, I was concerned about the issue of nested truncated parts. And I realized that adding a return value for `start_truncated()` resolve this problem.

roife:
Previously, hovering over any part of a hint would display the complete type.

Now, the hidden content is only showed when the mouse hovers over its corresponding `...`.

image

crates/ide/src/inlay_hints.rs
```diff
@@ -778,6 +818,8 @@ fn label_of_ty(
         db: sema.db,
         last_part: String::new(),
         location: None,
+        tooltip: None,
+        in_truncated_part: false,
         result: InlayHintLabel::default(),
         resolve: config.fields_to_resolve.resolve_label_location,
```
Veykril: The resolve field is now also used for tooltips by this PR which means this line here is incorrect. We should rename this field to `resolve_location` and add another `resolve_tooltip` field for this PRs needs instead.

crates/ide/src/inlay_hints.rs
```diff
@@ -903,6 +945,15 @@ mod tests {
...
     };
+    pub(super) const TEST_CONFIG_WITH_TRUNCATION: InlayHintsConfig = InlayHintsConfig {
+        type_hints: true,
+        parameter_hints: true,
+        chaining_hints: true,
+        closure_return_type_hints: ClosureReturnTypeHints::WithBlock,
+        binding_mode_hints: true,
+        max_length: Some(10),
+        ..DISABLED_CONFIG
+    };
```
Veykril:
```suggestion
    pub(super) const TEST_CONFIG_WITH_TRUNCATION: InlayHintsConfig = InlayHintsConfig {
        max_length: Some(10),
        ..TEST_CONFIG
    };
```

crates/hir-ty/src/display.rs
```diff
@@ -147,6 +155,15 @@ impl HirFormatter<'_> {
...
     }
 
+    fn maybe_truncated<T, F: FnOnce(&mut Self) -> T>(&mut self, f: F) -> T {
+        let truncated = self.should_truncate() && self.fmt.start_truncated();
+        let res = f(self);
+        if truncated {
+            self.fmt.end_truncated();
+        }
+        res
+    }
```
Veykril: I'd expect something like the following instead
```suggestion
    fn maybe_truncated<T, F: FnOnce(&mut Self) -> T>(&mut self, f: F) -> Option<T> {
        if self.should_truncate() {
             let end_truncate = self.fmt.start_truncated();
             if self.fmt.start_truncated() {
                 let res = f(self);
                 self.end_truncated();
                 Some(res)
             } else {
                 write!(self, "{TYPE_HINT_TRUNCATION}")
                 None
             }
        } else {
            Some(f(self))
        }
    }
```
This way any writers that don't care about the truncation behavior will have the previous behavior, that is they write the truncation dots as before.

crates/hir-ty/src/display.rs
```diff
@@ -392,19 +409,23 @@ impl HirFormatter<'_> {
...
             if !first {
                 write!(self, "{sep}")?;
             }
             first = false;
 
             // Abbreviate multiple omitted types with a single ellipsis.
-            if self.should_truncate() {
-                return write!(self, "{TYPE_HINT_TRUNCATION}");
-            }
+            in_truncated = !in_truncated && self.should_truncate() && self.fmt.start_truncated();
```
Veykril:
```suggestion
            if self.should_truncate() {
            	in_truncated |= !in_truncated && self.fmt.start_truncated();
            	if !in_truncated {
            	     return write!(self, "{TYPE_HINT_TRUNCATION}");
            	}
            }
```
same reason as the review above

crates/ide/src/inlay_hints.rs
roife: Yes... I didn't notice that `resolve` here represents `resolve_label_location`, which is confusing.

crates/hir-ty/src/display.rs
roife: So we should limit the behavior of "write dots" within HirFormatter and not leak it into the specific writer, right? I will modify this part.

roife: Here we should do three things in `end_truncated`:

1. Move the contents of `last_part` to `tooltip`
2. Write `TYPE_HINT_TRUNCATION` into `last_part`
3. Create a new `part`

Since the second step must be completed after 1 and before 3, we may need to let writer handle it, but I can have `end_truncated` take a parameter as a placeholder, so that `TYPE_HINT_TRUNCATION` is not leaked.

If the writer has more complex behavior in the future, we can consider using a closure.

(resolved in bed7a75435c77b2181ea374b8338e86f7528b1a2)

roife:
GitHub notifies me with "1 review requesting changes by reviewers," but I can't find where it is :(

crates/hir-ty/src/display.rs
```diff
@@ -1050,599 +1076,593 @@ fn render_variant_after_name<'db>(
...
-                write!(f, "]")?;
-            }
-            kind @ (TyKind::RawPtr(t, m) | TyKind::Ref(_, t, m)) => {
-                if let TyKind::Ref(l, _, _) = kind {
-                    f.write_char('&')?;
-                    if f.render_region(l) {
-                        l.hir_fmt(f)?;
-                        f.write_char(' ')?;
-                    }
+fn write_ty<'db>(
```
roife: Should we consider splitting this function into multiple functions or a separate file, as it is too large?

ChayimFriedman2:
@Veykril requested changes. As long as he won't send an approval it will stay with the same status (which doesn't prevent merging).

ChayimFriedman2:
Since there was no response from OP, I'm closing this. Please feel free to reopen or open a new PR if you want to resume work on this.

roife:
Sorry for the delay. I was tied up with my master’s thesis, but I’m able to focus on this again now. I’d be happy to resume the work by reopening this PR :)

PR #19649, opened 2025-04-21
feat(ci): Allow dispatching CI for stable releases

Veykril:


Veykril:
unsure about whether all those if guards are correct

BenjaminBrienen:
Why have both a stable channel and a release branch?

Veykril:
That's how it currently works. You can push to the release branch for CI to kick off a stable release. I figured being able to workflow dispatch via the github actions interface might be nice. Though the benefit of solely using the branch is that you can always see the last commit that we pushedo ut I suppose

PR #19659, opened 2025-04-22
feat: add support for folding ranges for chained expressions

### LINKED ISSUE #19620: Collapsible chained expressions (Method and Field Chains)
## Summary

Add support for collapsing/folding chained expressions in rust-analyzer to improve code readability and navigation in editors.

## Problem

Currently, rust-analyzer does not provide a folding option for chained expressions, which are sequences of method calls or field accesses connected by the dot operator (`.`). For example:

Image

In the editor, there is no folding arrow (or equivalent UI element) before the let statement to collapse this method chain into a single line, unlike other foldable constructs such as blocks or match expressions. This limitation makes it harder to navigate and read code with long chained expressions, especially in complex expressions.

## Proposal

Introduce support for collapsing chained expressions in rust-analyzer. When a method chain spans multiple lines, rust-analyzer should:

1. Display a folding arrow in the editor gutter at the start of the chain (e.g., before the `let` statement in the example above).
2. Allow collapsing the entire chain into a single line, such as:
   ```rust
   let a = b.c.d.e();
   ```
3. Preserve the original formatting when expanding the chain back to its multi-line form.

This feature would align with existing folding capabilities for other Rust constructs and improve the editing experience for codebases with frequent method chaining, such as those using iterator patterns.

## Use Case

Consider the following code:

```rust
let result = collection
    .iter()
    .filter(|x| x > 0)
    .map(|x| x * 2)
    .collect::<Vec<_>>();
```

With the proposed feature, users could collapse this chain to:

```rust
let result = collection.iter().filter(...).map(...).collect::<Vec<_>>();
```

This would reduce visual clutter and make it easier to focus on surrounding code, especially in larger functions or modules.

## Alternatives

To simplify implementation, rust-analyzer could support alternative collapsing formats that are less complex than fully inlining the chain. These include:

1. **Truncated Prefix with Ellipsis**: Collapse the chain to the base expression followed by an ellipsis (`...`).
   - Example: `let result = collection...;`
   - Benefit: Minimizes visual clutter and is easy to implement by identifying the base expression.

2. **Base Expression with Chain Indicator**: Collapse the chain to the base expression with a placeholder like `<chain>` to indicate a chain exists.
   - Example: `let result = collection.<chain>;`
   - Benefit: Signals the presence of a chain without parsing its contents.

3. **First Method/Field with Ellipsis**: Collapse the chain to the base expression plus the first method call or field access, followed by an ellipsis.
   - Example: `let result = collection.iter()...;`
   - Benefit: Preserves context about the chain's starting point while remaining simple.

4. **Type-Informed Collapse**: Collapse the chain to the base expression and the resulting type (if available), indicating the chain's outcome.
   - Example: `let result = collection -> Vec<_>;`
   - Benefit: Provides meaningful context about the chain's purpose, leveraging rust-analyzer's type inference.
   - Note: This alternative may not be so useful if type hints are enabled.

These alternatives reduce the complexity of handling complex method arguments (e.g., closures or generics) and provide flexible options for users to customize folding behavior.

roife:
fix #19620.

### DIFF
--- a/crates/ide/src/folding_ranges.rs
+++ b/crates/ide/src/folding_ranges.rs
@@ -26,6 +26,8 @@ pub enum FoldKind {
     WhereClause,
     ReturnType,
     MatchArm,
+    Stmt,
+    TailExpr,
 }
 
 #[derive(Debug)]
@@ -124,6 +126,11 @@ pub(crate) fn folding_ranges(file: &SourceFile) -> Vec<Fold> {
                                 res.push(Fold {range, kind: FoldKind::MatchArm})
                             }
                         },
+                        ast::Expr(expr) => {
+                            if let Some(range) = fold_range_for_multiline_tail_expr(expr) {
+                                res.push(Fold {range, kind: FoldKind::TailExpr})
+                            }
+                        },
                         _ => (),
                     }
                 }
@@ -151,6 +158,7 @@ fn fold_kind(kind: SyntaxKind) -> Option<FoldKind> {
         | MATCH_ARM_LIST
         | VARIANT_LIST
         | TOKEN_TREE => Some(FoldKind::Block),
+        EXPR_STMT | LET_STMT => Some(FoldKind::Stmt),
         _ => None,
     }
 }
@@ -281,6 +289,18 @@ fn fold_range_for_multiline_match_arm(match_arm: ast::MatchArm) -> Option<TextRa
     }
 }
 
+fn fold_range_for_multiline_tail_expr(expr: ast::Expr) -> Option<TextRange> {
+    let stmt_list = ast::StmtList::cast(expr.syntax().parent()?)?;
+    let tail_expr = stmt_list.tail_expr()?;
+
+    // Only fold if the tail expression spans multiple lines
+    if tail_expr.syntax().text().contains_char('\n') {
+        Some(tail_expr.syntax().text_range())
+    } else {
+        None
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use test_utils::extract_tags;
@@ -317,6 +337,8 @@ mod tests {
                 FoldKind::WhereClause => "whereclause",
                 FoldKind::ReturnType => "returntype",
                 FoldKind::MatchArm => "matcharm",
+                FoldKind::Stmt => "stmt",
+                FoldKind::TailExpr => "tailexpr",
             };
             assert_eq!(kind, &attr.unwrap());
         }
@@ -465,10 +487,10 @@ macro_rules! foo <fold block>{
         check(
             r#"
 fn main() <fold block>{
-    match 0 <fold block>{
+    <fold tailexpr>match 0 <fold block>{
         0 => 0,
         _ => 1,
-    }</fold>
+    }</fold></fold>
 }</fold>
 "#,
         );
@@ -479,7 +501,7 @@ fn main() <fold block>{
         check(
             r#"
             fn main() <fold block>{
-                match foo <fold block>{
+                <fold tailexpr>match foo <fold block>{
                     block => <fold block>{
                     }</fold>,
                     matcharm => <fold matcharm>some.
@@ -498,7 +520,7 @@ fn main() <fold block>{
                     structS => <fold matcharm>StructS <fold block>{
                         a: 31,
                     }</fold></fold>,
-                }</fold>
+                }</fold></fold>
             }</fold>
             "#,
         )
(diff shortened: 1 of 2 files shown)


roife:
I'm interested in the support for `collapsed_text`, so I will convert this to a draft PR for now.

roife:
I have finished a prototype:

image

If we fold the line `let navs`...

image

----------------

image

image

roife:
Currently, this feature only works on nvim, as it is the only editor that supports `collapsed_text`.

One of the nvim maintainers mentioned that syntax highlighting will be applied to collapsed text in the future, which enhancing its appearance.

crates/ide/src/folding_ranges.rs
```diff
@@ -146,8 +163,93 @@ pub(crate) fn folding_ranges(file: &SourceFile) -> Vec<Fold> {
...
+/// If `collapsed_text` is enabled, it generates a preview text for certain fold kinds that
+/// shows a summarized version of the folded content.
+fn build_fold(element: &SyntaxElement, kind: FoldKind, collapsed_text: bool) -> Fold {
+    if !collapsed_text {
+        return Fold::new(element.text_range(), kind);
+    }
+
+    let fold_with_collapsed_text = match kind {
+        FoldKind::TailExpr => {
+            let expr = ast::Expr::cast(element.as_node().unwrap().clone()).unwrap();
```
Veykril: We should be able to get rid of this unwrap dance by making `FoldKind` contain the appropriate syntax element, that is
```rust
enum FoldKind {
    TailExpr(ast::Expr),
    Stmt(ast::Stmt),
}
```

crates/ide/src/folding_ranges.rs
```diff
@@ -146,8 +163,93 @@ pub(crate) fn folding_ranges(file: &SourceFile) -> Vec<Fold> {
...
+
+            match_ast! {
+                match node {
+                    ast::ExprStmt(expr) => {
+                        let Some(expr) = expr.expr() else {
+                            break 'blk None;
+                        };
+
+                        let indent_level = expr.indent_level().0;
+                        let indents = "    ".repeat(indent_level as usize);
```
Veykril: Seemds odd to me that we need to add the leading whitespace, given the fold range starts at the statement start, not at the line start 🤔

crates/ide/src/folding_ranges.rs
```diff
@@ -164,10 +266,105 @@ fn fold_kind(kind: SyntaxKind) -> Option<FoldKind> {
...
+///
+/// # Returns
+/// - `Some(String)`: A dot-separated chain representation if the expression is chainable
+/// - `None`: If the expression is not suitable for collapsing (e.g., simple literals)
+///
+/// # Examples
+/// - `foo.bar().baz?` -> `"foo.bar().baz?"`
+/// - `obj.method(arg1, arg2)` -> `"obj.method(…)"`
+/// - `value?.field` -> `"value?.field"`
+fn collapsed_text_from_expr(mut expr: ast::Expr) -> Option<String> {
```
Veykril: This looks fairly brittle given how many postfix / prefix expression this ignores. Why would we handle `?` specially, but then render prefix operators as `<expr>` for example. I feel like this might be better implemented with a pre/post-order walk (`SyntaxNode::preorder_with_tokens`)?

Veykril (commented):
I don't think this needs a config. This feels like something the client can expose as a config itself (as it is free to discard the `collapsedText` field instead of using it)

roife:
> as it is free to discard the `collapsedText` field instead of using it

While it does not impact functionality, it may decrease both the computational load and the size of responses.

crates/ide/src/folding_ranges.rs
roife: Yes, this should be an issue with nvim, I will report this problem to nvim...

UPDATE: I found that this should be related to the `line_folding_only`, not an issue with nvim.

crates/ide/src/folding_ranges.rs
roife: Are you suggesting using `SyntaxNode::preorder_with_tokens` to traverse the tokens and then ignore the contents inside `()`s?

Veykril: You can `preorder_with_tokens.skip_subtree()` (if you store the iterator) when encouintering `ArgList` to have it not iterate the args further. 

But yes, I would assume if you can dynamically decide on whether to preorder or postorder walk, it might help with the expressions having differing fixivity.

roife:
I refactored the code in the new commit based on the previous review and simplified the implementation of `build_fold`.

Young-Flash:
any update on this?

roife:
> any update on this?

Sorry, I've been busy finishing my master's thesis looking for a job recently.

I have resolved the conflicts. @Veykril Do you have any suggestions?

ofseed:
This feature looks promising, and I'm adding highlighting support for this content to nvim. https://github.com/neovim/neovim/pull/38789
image

PR #19684, opened 2025-04-25
Always error when failed to parse DiscoverProjectMessage

darichey:
Currently we silently fail if the line can be parsed as json but not as `DiscoverProjectData`. That means if the discover command emits a malformed `DiscoverProjectData`, project discovery will never finish, and rust-analyzer becomes unusable. Instead, return an error if the line isn't a valid `DiscoverProjectData` so the discovery process fails and the error is correctly reported.


davidbarsky (approved):
approved, but will let someone else merge.

PR #19685, opened 2025-04-25
Add PGO support to install

fbernier:
Enable PGO optimizations when installing rust-analyzer with the --pgo flag. This mirrors functionality recently merged in rust-lang/rust-analyzer#19582 in dist command, allowing developers to create optimized local builds.

Example:
cargo xtask install --server --pgo clap-rs/clap

PGO code has been extracted to a dedicated module for reuse.


Veykril (approved):
Thanks!

PR #19713, opened 2025-04-29
refactor: Cleanup cfg check handling in expressiont store lowering

Veykril:


Veykril:
Oh, one commit got squashed into this, but that only forced threads to be named.

PR #19731, opened 2025-05-02
refactor: Simplify macro call id construction

Veykril:


Cargo.toml
```diff
@@ -131,6 +131,7 @@ process-wrap = { version = "8.2.0", features = ["std"] }
 pulldown-cmark-to-cmark = "10.0.4"
 pulldown-cmark = { version = "0.9.6", default-features = false }
 rayon = "1.10.0"
+rowan = "=0.15.15"
```
davidbarsky, line 134: I thought we bumped rowan?

Veykril: No we did not, in fact we can't given the new assertion that will fail on our code base still

PR #19733, opened 2025-05-02
fix: Implement mut to const ptr cast for method resolution

ShoyuVanilla:
Fixes rust-lang/rust-analyzer#19724 

Mimicked part of the method resolution logic of rustc and I think that we should move r-a's closer to rustc's (Maybe I'll do it later)


Veykril (approved):

Veykril:
> Mimicked part of the method resolution logic of rustc and I think that we should move r-a's closer to rustc's

We should I agree

PR #19735, opened 2025-05-03
Improve the let code snippet

A4-Tacks:
Improve the `let` and `letm` snippets

Make it work more like a postfix snippets

### DIFF
--- a/crates/ide-completion/src/completions/expr.rs
+++ b/crates/ide-completion/src/completions/expr.rs
@@ -363,9 +363,14 @@ pub(crate) fn complete_expr_path(
                     add_keyword("true", "true");
                     add_keyword("false", "false");
 
-                    if in_condition || in_block_expr {
-                        add_keyword("letm", "let mut $0");
-                        add_keyword("let", "let $0");
+                    if in_condition {
+                        add_keyword("letm", "let mut $1 = $0");
+                        add_keyword("let", "let $1 = $0");
+                    }
+
+                    if in_block_expr {
+                        add_keyword("letm", "let mut $1 = $2;");
+                        add_keyword("let", "let $1 = $2;");
                     }
 
                     if after_if_expr {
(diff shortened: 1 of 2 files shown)


crates/ide-completion/src/completions/expr.rs
```diff
@@ -363,9 +363,14 @@ pub(crate) fn complete_expr_path(
...
-                        add_keyword("letm", "let mut $0");
-                        add_keyword("let", "let $0");
+                    if in_condition {
+                        add_keyword("letm", "let mut $1 = $0");
+                        add_keyword("let", "let $1 = $0");
+                    }
+
+                    if in_block_expr {
+                        add_keyword("letm", "let mut $1 = $2;");
+                        add_keyword("let", "let $1 = $2;");
```
Veykril, line 373:
```suggestion
                        add_keyword("letm", "let mut $1 = $0;");
                        add_keyword("let", "let $1 = $0;");
```

crates/ide-completion/src/completions/expr.rs
```diff
@@ -363,9 +363,14 @@ pub(crate) fn complete_expr_path(
...
 
-                    if in_condition || in_block_expr {
-                        add_keyword("letm", "let mut $0");
-                        add_keyword("let", "let $0");
+                    if in_condition {
+                        add_keyword("letm", "let mut $1 = $0");
+                        add_keyword("let", "let $1 = $0");
+                    }
+
+                    if in_block_expr {
```
Veykril, line 371: why was this block duplicated though

A4-Tacks: The difference is that if let does not require semicolons

crates/ide-completion/src/completions/expr.rs
A4-Tacks: If someone uses let in an in line code block, they may need to quickly jump out and continue writing

`{ let x = Foo; }` -> `{ let x = Foo; x.bar() }`

A4-Tacks: It seems easy to move the cursor behind the semicolon

Issue #19722, opened 2025-04-30
unwrap type assist


Veykril:
An assist that "unwraps" a type into its generic type argument, that is `Option<T>` becomes `T`.

Requirements should be that the type is a path type which exactly one type paramater (and maybe 0+ const and lifetime parameters which are then discarded).

snprajwal:
Interesting, looks like "addresses" [isn't a recognised keyword for GitHub](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) to auto-close linked issues.

PR #19740, opened 2025-05-04
feat: add an assist to unwrap a type with a generic arg

vishruth-thimmaiah:
Addresses #19722

This assist unwraps a type into its generic type argument, ignoring const and lifetime arguments.


crates/ide-assists/src/handlers/unwrap_path_type.rs
```diff
@@ -0,0 +1,156 @@
...
+            }
+            _ => return None,
+        }
+    }
+
+    let generic_arg = generic_arg?;
+
+    acc.add(
+        AssistId::refactor_extract("unwrap_type_to_generic_arg"),
+        "Unwrap type to generic argument",
```
Veykril, line 46:
```suggestion
        format!("Unwrap type to type argument {generic_arg}"),
```
That makes it a bit more clearer what is being transformed

Veykril (commented):
small nit but otherwise looks good to me!

Veykril (approved):
Thanks!

PR #19760, opened 2025-05-08
Use `NonMaxU32` for arena indices

Veykril:
Wanna see what the memory impact of this is


Veykril:
8-9 mb on self and the std libraries, so probably not worth it.

PR #19799, opened 2025-05-15
fix: rename and add import

Hmikihiro:
fix: https://github.com/rust-lang/rust-analyzer/issues/18892


crates/ide-db/src/search.rs
```diff
@@ -1180,7 +1196,8 @@ impl<'a> FindUsages<'a> {
                 if self.def == def
                     // is our def a trait assoc item? then we want to find all assoc items from trait impls of our trait
                     || matches!(self.assoc_item_container, Some(hir::AssocItemContainer::Trait(_)))
-                        && convert_to_def_in_trait(self.sema.db, def) == self.def =>
+                        && convert_to_def_in_trait(self.sema.db, def) == self.def
+                    || self.is_related_import(name_ref) =>
```
Hmikihiro, line 1200: When I find some definition item, we use NameRefClass::classify but it only return single definition. So, I check use block have some definition.

crates/ide-db/src/rename.rs
```diff
@@ -525,6 +528,31 @@ fn source_edit_from_name_ref(
...
+        let range = name_ref.syntax().text_range();
+        if res.iter().any(|res| res == &def) {
+            if res.len() == 1 {
+                edit.replace(range, new_name.to_owned());
+            } else {
+                edit.replace(range, format!("{{{}, {}}}", new_name, name_ref.text()));
+            }
+        }
+
+        return true;
```
Hmikihiro, line 555: When we rename import item, We check that the path is ambiguous or not. if It is ambiguous, we add old import name and new name.

crates/ide-db/src/search.rs
Hmikihiro: this change give affect to this test
https://github.com/rust-lang/rust-analyzer/blob/d2a05af1d30ddd3695008efd7231c631adec207d/crates/ide/src/references.rs#L1765-L1780

PR #19807, opened 2025-05-16
fix: Don't overwrite `RUSTUP_TOOLCHAIN` if it is already set

Veykril:


Veykril:
cc @weihanglo, this might help with your setup slightly as discussed. Either way this seems like the right behavior for rust-analyzer here

crates/project-model/src/sysroot.rs
weihanglo, line 1: Thank you! This looks the correct behavior regardless it fixes issues in our environment or not.

Do you want me to first test against our environment before proceeding?

PR #19876, opened 2025-05-28
feat: Render padding information when hovering on structs

### LINKED ISSUE #19865: Render tail/largest consecutive padding bytes on hover
```rust
struct S {
    field: u64,
    field2: bool,
}
```
Hovering `S` here currently renders `size = 16 (0x10), align = 0x8, niches = 254` which is pretty nice, but there is more layout info hidden here, the fact that 3 bytes of this struct are hidden in padding which means fields of up to 3 bytes can still be added without fear of increasing the struct size.

We should render the largest consecutive padding bytes for `repr(Rust)`, or for `repr(C)` the tail padding (if any)

ShoyuVanilla:
Resolves #19865


crates/hir/src/lib.rs
```diff
@@ -5972,6 +5972,63 @@ impl Layout {
...
+            layout::FieldsShape::Union(_) => None,
+            layout::FieldsShape::Array { stride, count } => (0..count)
+                .map(|i| (i, stride.bytes() * i))
+                .chain(std::iter::once((count, self.0.size.bytes())))
+                .tuple_windows()
+                .filter_map(|((i, start), (_, end))| {
+                    let size = field_size(i as usize)?;
+                    end.checked_sub(start)?.checked_sub(size)
+                })
+                .max(),
```
Veykril, line 6008: No need to loop here, the elements should all have the same padding right?
```suggestion
            layout::FieldsShape::Array { stride: _, count: 0..=1 }  => None,
            layout::FieldsShape::Array { stride, count: _ }  => {
                    let size = field_size(0 as usize)?;
                    stride.bytes().checked_sub(size)
            },
```

crates/rust-analyzer/src/config.rs
```diff
@@ -149,6 +149,8 @@ config_data! {
         hover_memoryLayout_niches: Option<bool> = Some(false),
         /// How to render the offset information in a memory layout hover.
         hover_memoryLayout_offset: Option<MemoryLayoutHoverRenderKindDef> = Some(MemoryLayoutHoverRenderKindDef::Hexadecimal),
+        /// How to render the padding information in a memory layout hover.
+        hover_memoryLayout_padding: Option<MemoryLayoutHoverRenderKindDef> = Some(MemoryLayoutHoverRenderKindDef::Hexadecimal),
```
Veykril, line 153: We should disable this by default like niches are

crates/hir/src/lib.rs
ShoyuVanilla: Yes, I mindlessly wrote those lines with just muscle memory 😅

PR #19879, opened 2025-05-28
fix: Fix IDE layer not resolving some macro calls

Veykril:
Closes (for the most part) https://github.com/rust-lang/rust-analyzer/issues/19873


Veykril:
There seems to be some unnecessary duplication here, I also now uqestion the need for https://github.com/rust-lang/rust-analyzer/pull/19862. I don't think that was necessary after all (also the test doesn't even exercise the blocks I think).

PR #19890, opened 2025-05-29
fix: Fix import insertion not being fully cfg aware

Veykril:
The logic is basically now:
- if we are in a `cfg`'d context and within a block expression, insert the import into the block expression
- otherwise insert the import as usual but with the `cfg`s attached (this does try to merge imports if the attributes match)


Veykril:
I guess this also kind of qualifies as a feature?

Issue #19666, opened 2025-04-23
Code action to convert between struct types


oriongonza:
```rs
struct S {} // Convert into unit struct
struct S; // Convert into normal struct

struct S { field: String }; // Convert into tuple struct
struct S(String); // Convert into normal struct

struct S(String, u64); // Convert into normal struct
struct S { 
  field_1: String, // These can be renamed easily later on
  field_2: u64
}
```
I think that this would aid in refactoring in a big way, especially for tuple -> named structs.

It's a pain to fix them manually.

roife:
We already have this assist. Note that you need to place the cursor on the name of the `struct` to activate it. It would be better to expand the trigger area to allow more people to discover this feature.

Veykril:
I'd expect this assist to trigger  from the struct/visibility token up to the end of the name. The entire definition is way too annoying. Stuff that affects the entire definition of something imo should be triggerable only on the "signature" of the thing.

ada4a:
> I'd expect this assist to trigger from the struct/visibility token up to the end of the name

AFAICT this was implemented in #19901, so this can be closed I think?

PR #19901, opened 2025-05-31
Enable Assist edit for tuple<->named struct for the struct and visiblity keywords

BazookaMusic:
I was looking for an issue to do my first contribution to this repo.  In the discussion for the issue #19666 , a suggestion was made to expand when the code action for converting between named/tuple structs is triggered to include the struct and visibility keywords, so I've added these too.

Maybe the approach of targetting specific keywords is not that nice and it should be any kind of node who parent is a struct but not the inner part of a struct?

PS. I'm new to rust and this repo, so please feel free to nitpick the code

### DIFF
--- a/crates/ide-assists/src/utils.rs
+++ b/crates/ide-assists/src/utils.rs
@@ -1,5 +1,7 @@
 //! Assorted functions shared by several assists.
 
+use either::Either;
+
 pub(crate) use gen_trait_fn_body::gen_trait_fn_body;
 use hir::{
     DisplayTarget, HasAttrs as HirHasAttrs, HirDisplay, InFile, ModuleDef, PathResolution,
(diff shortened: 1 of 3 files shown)


crates/ide-assists/src/utils.rs
```diff
@@ -1146,3 +1148,19 @@ pub fn is_body_const(sema: &Semantics<'_, RootDatabase>, expr: &ast::Expr) -> bo
...
+    ctx.find_node_at_offset::<ast::Name>()
+        .and_then(|name| name.syntax().parent())
+        .or(ctx
+            .token_at_offset()
+            .find(|leaf| matches!(leaf.kind(), STRUCT_KW))
+            .and_then(|kw| kw.parent()))
+        .or(ctx
+            .find_node_at_offset::<ast::Visibility>()
+            .and_then(|visibility| visibility.syntax().parent()))
+        .and_then(<Either<ast::Struct, ast::Variant>>::cast)
```
Veykril: A simpler approach would be to do `ctx.find_node_at_offset::<ast::Struct>()`, then fetch the `.field_list()` and check if the cursor position is to the left of the field list, that is `ctx.offset() < field_list.syntax().text_range().start()`

Same for `ast::Variant`. Then we can also get rid of the helper method and just do this in the corresponding assists directly.

BazookaMusic: I thought about that too, but wasn't sure if we wanted to use the offsets for this. But it is much easier to read, so I'll change to that approach.

BazookaMusic: I've removed the utility method and did it via the offset

Veykril (approved):
Thanks!

PR #19902, opened 2025-06-02
feat: ide-assists, generate binary ops impl

A4-Tacks:
Generate binary operation from assign operation.

# Example

```
struct Int(i32);

impl<T> core::ops::AddAssign$0<i32> for Int {
    fn add_assign(&mut self, rhs: i32) {
        self.0 += rhs;
        todo!()
    }
}
```
->
```
struct Int(i32);

impl<T> core::ops::Add<i32> for Int {
    type Output = Self;

    fn add(mut self, rhs: i32) -> Self::Output {
        self += rhs;
        self
    }
}

impl<T> core::ops::AddAssign<i32> for Int {
    fn add_assign(&mut self, rhs: i32) {
        self.0 += rhs;
        todo!()
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/generate_binary_ops_impl.rs
+++ b/crates/ide-assists/src/handlers/generate_binary_ops_impl.rs
@@ -1,0 +1,698 @@
+use syntax::{
+    AstNode,
+    ast::{self, ArithOp, BinaryOp, HasGenericParams, edit, edit_in_place::Indent, make},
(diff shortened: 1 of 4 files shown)


ShoyuVanilla:
Will we really need this? I think the use cases would be rare and wouldn't take much work to implement them manually

A4-Tacks:

PR #19908, opened 2025-06-02
feat: implement attribute completions for diagnostics module

rmehri01:
Implements completions for `diagnostic::on_unimplemented` and `diagnostic::do_not_recommend`. Unless I missed it, I don't think there are other attributes here that deal with qualifiers like `diagnostic::` so I added somewhat naive splitting on `::` and then made it so that it is stripped before being used in the label and snippet. Similarly, in `complete_known_attribute_input` we first need to check if the qualifier is present instead of assuming that `as_single_name_ref` will work.

I also noticed that `must_use` is currently duplicated in functions and traits since it already shows up in `item` :smile: 

[asciicast](https://asciinema.org/a/xe2w6QoiVbH4Vsv1gPreZGnbx)

Closes rust-lang/rust-analyzer#19892

### DIFF
--- a/crates/ide-completion/src/completions/attribute.rs
+++ b/crates/ide-completion/src/completions/attribute.rs
@@ -152,17 +152,22 @@ pub(crate) fn complete_attribute_path(
     });
 
     let add_completion = |attr_completion: &AttrCompletion| {
-        // if we already have the qualifier of the completion, then trim it from the label and the snippet
-        let mut label = attr_completion.label;
-        let mut snippet = attr_completion.snippet;
-        if let Some(name_ref) = qualifier_path.and_then(|q| q.as_single_name_ref()) {
-            if let Some((label_qual, label_seg)) = attr_completion.label.split_once("::") {
-                if name_ref.text() == label_qual {
-                    label = label_seg;
-                    snippet = snippet.map(|snippet| {
-                        snippet.trim_start_matches(label_qual).trim_start_matches("::")
-                    });
-                }
+        // if we don't already have the qualifiers of the completion, then
+        // add the missing parts to the label and snippet
+        let mut label = attr_completion.label.to_owned();
+        let mut snippet = attr_completion.snippet.map(|s| s.to_owned());
+        let segments = qualifier_path.iter().flat_map(|q| q.segments()).collect::<Vec<_>>();
+        let qualifiers = attr_completion.qualifiers;
+        let matching_qualifiers = segments
+            .iter()
+            .zip(qualifiers)
+            .take_while(|(s, q)| s.name_ref().is_some_and(|t| t.text() == **q))
+            .count();
+        if matching_qualifiers != qualifiers.len() {
+            let prefix = qualifiers[matching_qualifiers..].join("::");
+            label = format!("{prefix}::{label}");
+            if let Some(s) = snippet.as_mut() {
+                *s = format!("{prefix}::{s}");
             }
         }
 
@@ -197,12 +202,17 @@ struct AttrCompletion {
     label: &'static str,
     lookup: Option<&'static str>,
     snippet: Option<&'static str>,
+    qualifiers: &'static [&'static str],
     prefer_inner: bool,
 }
 
 impl AttrCompletion {
     fn key(&self) -> &'static str {
         self.lookup.unwrap_or(self.label)
+    }
+
+    const fn qualifiers(self, qualifiers: &'static [&'static str]) -> AttrCompletion {
+        AttrCompletion { qualifiers, ..self }
     }
 
     const fn prefer_inner(self) -> AttrCompletion {
@@ -215,7 +225,7 @@ const fn attr(
     lookup: Option<&'static str>,
     snippet: Option<&'static str>,
 ) -> AttrCompletion {
-    AttrCompletion { label, lookup, snippet, prefer_inner: false }
+    AttrCompletion { label, lookup, snippet, qualifiers: &[], prefer_inner: false }
 }
 
 macro_rules! attrs {
@@ -324,8 +334,14 @@ const ATTRIBUTES: &[AttrCompletion] = &[
     attr("deny(…)", Some("deny"), Some("deny(${0:lint})")),
     attr(r#"deprecated"#, Some("deprecated"), Some(r#"deprecated"#)),
     attr("derive(…)", Some("derive"), Some(r#"derive(${0:Debug})"#)),
-    attr("diagnostic::do_not_recommend", None, None),
-    attr("diagnostic::on_unimplemented", None, Some(r#"diagnostic::on_unimplemented(${0:keys})"#)),
+    attr("do_not_recommend", Some("diagnostic::do_not_recommend"), None)
+        .qualifiers(&["diagnostic"]),
+    attr(
+        "on_unimplemented",
+        Some("diagnostic::on_unimplemented"),
+        Some(r#"on_unimplemented(${0:keys})"#),
+    )
+    .qualifiers(&["diagnostic"]),
     attr(r#"doc = "…""#, Some("doc"), Some(r#"doc = "${0:docs}""#)),
     attr(r#"doc(alias = "…")"#, Some("docalias"), Some(r#"doc(alias = "${0:docs}")"#)),
     attr(r#"doc(hidden)"#, Some("dochidden"), Some(r#"doc(hidden)"#)),


crates/ide-completion/src/completions/attribute.rs
```diff
@@ -40,14 +41,23 @@ pub(crate) fn complete_known_attribute_input(
...
+    if let Some(qualifier) = path.qualifier() {
+        let qualifier_name_ref = qualifier.as_single_name_ref()?;
+        match (qualifier_name_ref.text().as_str(), name_ref.text().as_str()) {
+            ("diagnostic", "on_unimplemented") => {
+                diagnostic::complete_on_unimplemented(acc, ctx, tt)
+            }
+            _ => (),
+        }
+        return Some(());
+    }
```
rmehri01: This can be combined with the below `match` but is a little annoying because of having to extract local variables to use `as_ref`. Not sure if there is a better way to handle that

crates/ide-completion/src/completions/attribute.rs
```diff
@@ -40,14 +41,23 @@ pub(crate) fn complete_known_attribute_input(
...
 ) -> Option<()> {
     let attribute = fake_attribute_under_caret;
-    let name_ref = match attribute.path() {
-        Some(p) => Some(p.as_single_name_ref()?),
-        None => None,
-    };
-    let (path, tt) = name_ref.zip(attribute.token_tree())?;
+    let path = attribute.path()?;
+    let name_ref = path.segment()?.name_ref();
+    let (name_ref, tt) = name_ref.zip(attribute.token_tree())?;
```
Veykril: A nicer way to handle this I think would be something along the lines of 
```suggestion
    let path = attribute.path()?;
    let segments = path.segments().map(|s| s.name_ref()).collect::<Option<Vec<_>>>()?;
    let segments = segments.iter().map(|n| n.text().as_str()).collect::<ec<_>>();
    let tt = attribute.token_tree()?;
```

Then we can just match on `&*segments` (`&[&str]`), that is `["diagnostic", "on_unimplemented"] => ....`, etc which allows to easily combine the two `match`es here

crates/ide-completion/src/completions/attribute.rs
```diff
@@ -149,18 +161,28 @@ pub(crate) fn complete_attribute_path(
...
+                    label = label_seg;
+                    snippet = snippet.map(|snippet| {
+                        snippet.trim_start_matches(label_qual).trim_start_matches("::")
+                    });
+                }
+            }
+        }
+
+        let mut item =
+            CompletionItem::new(SymbolKind::Attribute, ctx.source_range(), label, ctx.edition);
```
Veykril: I think I'd prefer if we tried to fix up the edit range instead, that is replace `ctx.source_range()` with the appropriate range to cover the entire path here

crates/ide-completion/src/completions/attribute/diagnostic.rs
```diff
@@ -0,0 +1,60 @@
+//! Completion for diagnostic attributes.
+
+use ide_db::SymbolKind;
+use syntax::ast::{self};
```
Veykril:
```suggestion
use syntax::ast;
```

crates/ide-completion/src/completions/attribute/diagnostic.rs
```diff
@@ -0,0 +1,60 @@
...
+            item.add_to(acc, ctx.db);
+        }
+    }
+}
+
+const ATTRIBUTES: &[AttrCompletion] = &[
+    super::attr(r#"label = "…""#, Some("label"), Some(r#"label = "${0:label}""#)),
+    super::attr(r#"message = "…""#, Some("message"), Some(r#"message = "${0:message}""#)),
+    super::attr(r#"note = "…""#, Some("note"), Some(r#"note = "${0:note}""#)),
+];
```
Veykril: nit, these aren't ATTRIBUTES, `ATTRIBUTE_ARGS` is probably slightly better

crates/ide-completion/src/completions/attribute.rs
rmehri01: Oh nice yeah that's better, I think we need to do 3 vecs because of the `as_str` though

crates/ide-completion/src/completions/attribute.rs
rmehri01: Do we still want to change the label for the completion? I noticed when you do `clippy::` for example the completions are shortened to not include `clippy`

crates/ide-completion/src/completions/attribute.rs
Veykril: Ah right, `text()` may own

crates/ide-completion/src/completions/attribute.rs
Veykril: Ah, right, well I guess we can keep this then though I don't like how its done. (doing it any otherway probably requires changing the macro though)

rmehri01: Yeah I don't really like how this is done either, I'll see if I can find a cleaner way

rmehri01: I think something like this might be a bit nicer, it adds a `qualifiers` field to `AttrCompletion` and then if the qualifiers aren't matching, we add on the parts that are needed to the label and snippet. What do you think? 

```diff
diff --git a/crates/ide-completion/src/completions/attribute.rs b/crates/ide-completion/src/completions/attribute.rs
index 0ca90fae4e..3babffc82d 100644
--- a/crates/ide-completion/src/completions/attribute.rs
+++ b/crates/ide-completion/src/completions/attribute.rs
@@ -152,17 +152,22 @@ pub(crate) fn complete_attribute_path(
     });
 
     let add_completion = |attr_completion: &AttrCompletion| {
-        // if we already have the qualifier of the completion, then trim it from the label and the snippet
-        let mut label = attr_completion.label;
-        let mut snippet = attr_completion.snippet;
-        if let Some(name_ref) = qualifier_path.and_then(|q| q.as_single_name_ref()) {
-            if let Some((label_qual, label_seg)) = attr_completion.label.split_once("::") {
-                if name_ref.text() == label_qual {
-                    label = label_seg;
-                    snippet = snippet.map(|snippet| {
-                        snippet.trim_start_matches(label_qual).trim_start_matches("::")
-                    });
-                }
+        // if we don't already have the qualifiers of the completion, then
+        // add the missing parts to the label and snippet
+        let mut label = attr_completion.label.to_owned();
+        let mut snippet = attr_completion.snippet.map(|s| s.to_owned());
+        let segments = qualifier_path.iter().flat_map(|q| q.segments()).collect::<Vec<_>>();
+        let qualifiers = attr_completion.qualifiers;
+        let matching_qualifiers = segments
+            .iter()
+            .zip(qualifiers)
+            .take_while(|(s, q)| s.name_ref().is_some_and(|t| t.text() == **q))
+            .count();
+        if matching_qualifiers != qualifiers.len() {
+            let prefix = qualifiers[matching_qualifiers..].join("::");
+            label = format!("{prefix}::{label}");
+            if let Some(s) = snippet.as_mut() {
+                *s = format!("{prefix}::{s}");
             }
         }
 
@@ -197,6 +202,7 @@ struct AttrCompletion {
     label: &'static str,
     lookup: Option<&'static str>,
     snippet: Option<&'static str>,
+    qualifiers: &'static [&'static str],
     prefer_inner: bool,
 }
 
@@ -205,6 +211,10 @@ impl AttrCompletion {
         self.lookup.unwrap_or(self.label)
     }
 
+    const fn qualifiers(self, qualifiers: &'static [&'static str]) -> AttrCompletion {
+        AttrCompletion { qualifiers, ..self }
+    }
+
     const fn prefer_inner(self) -> AttrCompletion {
         AttrCompletion { prefer_inner: true, ..self }
     }
@@ -215,7 +225,7 @@ const fn attr(
     lookup: Option<&'static str>,
     snippet: Option<&'static str>,
 ) -> AttrCompletion {
-    AttrCompletion { label, lookup, snippet, prefer_inner: false }
+    AttrCompletion { label, lookup, snippet, qualifiers: &[], prefer_inner: false }
 }
 
 macro_rules! attrs {
@@ -324,8 +334,14 @@ const ATTRIBUTES: &[AttrCompletion] = &[
     attr("deny(…)", Some("deny"), Some("deny(${0:lint})")),
     attr(r#"deprecated"#, Some("deprecated"), Some(r#"deprecated"#)),
     attr("derive(…)", Some("derive"), Some(r#"derive(${0:Debug})"#)),
-    attr("diagnostic::do_not_recommend", None, None),
-    attr("diagnostic::on_unimplemented", None, Some(r#"diagnostic::on_unimplemented(${0:keys})"#)),
+    attr("do_not_recommend", Some("diagnostic::do_not_recommend"), None)
+        .qualifiers(&["diagnostic"]),
+    attr(
+        "on_unimplemented",
+        Some("diagnostic::on_unimplemented"),
+        Some(r#"on_unimplemented(${0:keys})"#),
+    )
+    .qualifiers(&["diagnostic"]),
     attr(r#"doc = "…""#, Some("doc"), Some(r#"doc = "${0:docs}""#)),
     attr(r#"doc(alias = "…")"#, Some("docalias"), Some(r#"doc(alias = "${0:docs}")"#)),
     attr(r#"doc(hidden)"#, Some("dochidden"), Some(r#"doc(hidden)"#)),
```

Veykril (approved):
Thanks! I'll do a small follow up change to this

PR #19914, opened 2025-06-03
hir-ty: add incremental tests checking for `infer` invalidation

davidbarsky:
This PR adds some tests that check that `infer` is invalidated when a DefMap is invalidated. These new tests will need to fixed and renamed https://github.com/rust-lang/rust-analyzer/pull/19837 by:
- renaming tests like `adding_use_statement_invalidates_infer` to `adding_use_statement_should_not_invalidate_infer`.
- Negating the post-edit `events` assertion to `assert!(!format!("{events:?}").contains("infer_shim"), "{events:#?}")`.

(Full disclosure: I used Claude Sonnet 4 to write these tests. I am, of course, responsible for any code I submit for review, but it wasn't half bad at the job. I had to rename a few tests/example code in the tests, however.)

### DIFF
--- a/crates/hir-ty/src/tests/incremental.rs
+++ b/crates/hir-ty/src/tests/incremental.rs
@@ -106,3 +106,256 @@ fn baz() -> i32 {
         assert_eq!(format!("{events:?}").matches("infer_shim").count(), 1, "{events:#?}")
     }
 }
+
+#[test]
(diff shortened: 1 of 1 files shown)


davidbarsky:
Updated the tests to assert which queries were run.

Veykril (approved):

PR #19917, opened 2025-06-04
Add AsRef and Borrow for generate_mut_trait_impl

A4-Tacks:
- AsRef -> AsMut
- Borrow -> BorrowMut

Example
====================

```rust
//- minicore: as_ref
struct Foo(i32);

impl core::convert::AsRef$0<i32> for Foo {
    fn as_ref(&self) -> &i32 {
        &self.0
    }
}
```
->
```rust
struct Foo(i32);

$0impl core::convert::AsMut<i32> for Foo {
    fn as_mut(&mut self) -> &mut i32 {
        &self.0
    }
}

impl core::convert::AsRef<i32> for Foo {
    fn as_ref(&self) -> &i32 {
        &self.0
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/generate_mut_trait_impl.rs
+++ b/crates/ide-assists/src/handlers/generate_mut_trait_impl.rs
@@ -48,36 +48,42 @@ pub(crate) fn generate_mut_trait_impl(acc: &mut Assists, ctx: &AssistContext<'_>
     let impl_def = ctx.find_node_at_offset::<ast::Impl>()?.clone_for_update();
     let indent = impl_def.indent_level();
 
+    let (apply_trait, new_apply_trait) = impl_def
+        .syntax()
+        .descendants()
+        .filter_map(ast::NameRef::cast)
+        .find_map(process_trait_name)?;
+
     let trait_ = impl_def.trait_()?;
     if let ast::Type::PathType(trait_path) = trait_ {
         let trait_type = ctx.sema.resolve_trait(&trait_path.path()?)?;
         let scope = ctx.sema.scope(trait_path.syntax())?;
-        if trait_type != FamousDefs(&ctx.sema, scope.krate()).core_convert_Index()? {
+        let famous_defs = FamousDefs(&ctx.sema, scope.krate());
+        if trait_type != get_famous(&apply_trait.text(), famous_defs)? {
             return None;
         }
     }
 
     // Index -> IndexMut
-    let index_trait = impl_def
-        .syntax()
-        .descendants()
-        .filter_map(ast::NameRef::cast)
-        .find(|it| it.text() == "Index")?;
     ted::replace(
-        index_trait.syntax(),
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/generate_mut_trait_impl.rs
```diff
@@ -186,6 +227,35 @@ impl<T> core::ops::Index<Axis> for [T; 3] where T: Copy {
...
+}
+"#,
+            r#"
+struct Foo(i32);
+
+$0impl<T> core::convert::AsMut<i32> for Foo {
+    fn as_mut(&mut self) -> &mut i32 {
+        &self.0
+    }
+}
```
ShoyuVanilla, line 252: This code is not compiles not just because of redundant type parameter `T`, but we also need `mut` keyword after `&`.
Would this assist be really handy? I doubt as it just copies the whole impl block and just changes trait, function names and signatures, which can be easily done with copy-paste.

crates/ide-assists/src/handlers/generate_mut_trait_impl.rs
```diff
@@ -104,14 +108,51 @@ pub(crate) fn generate_mut_trait_impl(acc: &mut Assists, ctx: &AssistContext<'_>
...
-        "Generate `IndexMut` impl from this `Index` trait",
+        format!("Generate `{new_apply_trait}` impl from this `{apply_trait}` trait"),
         target,
         |edit| {
             edit.insert(target.start(), format!("$0{impl_def}\n\n{indent}"));
         },
     )
 }
 
+fn get_famous(apply_trait: &str, famous: FamousDefs<'_, '_>) -> Option<hir::Trait> {
```
ShoyuVanilla: I think this should be receive `hir::Trait` though it's currently using `&str` before this PR.
Like here:
https://github.com/rust-lang/rust-analyzer/blob/5b852da4c3852d030ebca67746980bc0288a85ec/crates/ide-assists/src/handlers/convert_into_to_from.rs#L35-L44

crates/ide-assists/src/handlers/generate_mut_trait_impl.rs
```diff
@@ -186,6 +227,35 @@ impl<T> core::ops::Index<Axis> for [T; 3] where T: Copy {
...
+"#,
+        );
+
+        check_assist(
+            generate_mut_trait_impl,
+            r#"
+//- minicore: as_ref
+struct Foo(i32);
+
+impl<T> core::convert::AsRef$0<i32> for Foo {
```
ShoyuVanilla: Those tests have redundant type parameter `T`

ShoyuVanilla (approved):
I think there's still room for improvement here like:

- We want to replace `ted::*` API with `SyntaxEditor` ones
- When making implementations for `BorrowMut`, it might emit nameres error because we don't import it automatically. In fact, this is same for pre-existing `IndexMut` 😅 

But anyway, I think this is an improvement over the current assist. Thanks!

PR #19918, opened 2025-06-04
Add ide-assist: remove else branches

A4-Tacks:


crates/ide-assists/src/handlers/remove_else_branches.rs
```diff
@@ -0,0 +1,90 @@
...
+// }
+// ```
+pub(crate) fn remove_else_branches(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
+    let else_token = ctx.find_token_syntax_at_offset(T![else])?;
+    let else_branches = ctx
+        .find_node_at_range::<ast::IfExpr>()
+        .and_then(|if_expr| if_expr.else_branch()?.syntax().clone().into())
+        .or_else(|| {
+            ctx.find_node_at_range::<ast::LetStmt>()?
+                .let_else()?
```
ShoyuVanilla, line 45: Should this be triggered on `let else` statements? I think most `let else` statements' else branch is not unreachable and removing them will result in compilation errors

A4-Tacks: Can be used for operations after refactoring into irrefutable pattern

A4-Tacks:
> I think we can combine this with the unwrap block assist which has a similar purpose?

I don't quite understand, what should I do?

Veykril:
We have this unwrap block assist here https://github.com/rust-analyzer/rust-analyzer/blob/c5181dbbe33af6f21b9d83e02fdb6fda298a3b65/crates/ide-assists/src/handlers/unwrap_block.rs

which I just realized already does what this assist does I believe? If you put the cursor on the else block's braces it pops up there

Though it does not trigger for let else yet

A4-Tacks:
> We have this unwrap block assist here

If it is applicable to `let-else`, it may be strange:

For example:

```rust
let Some(x) = val else {$0 return false; };
```

**Possible output one**:

```rust
return false;
```

This seems niche enough

**Possible output two**:

```rust
let Some(x) = val;
```

This does not match the name of the unwrap block

> Though it does not trigger for let else yet

Do I need to trigger `remove_else_branches` on the `l_curly` of let-else as well?

Veykril:
Ah right, the assist does have a different intent than this one, nevermind then.

A4-Tacks:
What do I need to do?

- Move the assist of 'apply for let-else' to 'Unwrap block'?
  - Implement `Possible output one`?
  - Implement `Possible output two`?

- Apply this assist to `l_curly`?

Veykril (approved):

Veykril:
Nothing, disregard my earlier comment its not applicable to this after all.

PR #19919, opened 2025-06-04
ci: When autopublishing, remove `xtask` from workspace

ShoyuVanilla:
c.c rust-lang/rust-analyzer#19903

### DIFF
--- a/.github/workflows/autopublish.yaml
+++ b/.github/workflows/autopublish.yaml
@@ -54,8 +54,8 @@ jobs:
           cargo workspaces rename --from project-model project_model
           cargo workspaces rename --from test-fixture test_fixture
           cargo workspaces rename --from test-utils test_utils
-          # Remove library crates from the workspaces so we don't auto-publish them as well
-          sed -i 's/ "lib\/\*",//' ./Cargo.toml
+          # Remove library crates and xtask from the workspaces so we don't auto-publish them as well
+          sed -i 's/"xtask\\/",//; s/ "lib\/\*",//' ./Cargo.toml
           cargo workspaces rename ra_ap_%n
           find crates/rust-analyzer -type f -name '*.rs' -exec sed -i 's/rust_analyzer/ra_ap_rust_analyzer/g' {} +
           cargo workspaces publish --yes --force '*' --exact --no-git-commit --allow-dirty --skip-published custom 0.0.$(($RUN_NUMBER + 133))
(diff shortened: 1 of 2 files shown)


lnicola:
Should we also revert the other change, then?

ShoyuVanilla:
> Should we also revert the other change, then?

Yes, I think that's reasonable

lnicola (approved):

.github/workflows/autopublish.yaml
```diff
@@ -54,8 +54,8 @@ jobs:
           cargo workspaces rename --from project-model project_model
           cargo workspaces rename --from test-fixture test_fixture
           cargo workspaces rename --from test-utils test_utils
-          # Remove library crates from the workspaces so we don't auto-publish them as well
-          sed -i 's/ "lib\/\*",//' ./Cargo.toml
+          # Remove library crates and xtask from the workspaces so we don't auto-publish them as well
+          sed -i 's/ "lib\/\*",//; s/ "xtask\/",//' ./Cargo.toml
```
lnicola: I need to do it this way because there's no space before `xtask`. The order doesn't matter though.

```suggestion
          sed -i 's/"xtask\\/",//; s/ "lib\/\*",//' ./Cargo.toml
```

ShoyuVanilla: Oh, good catch!

ShoyuVanilla:
```
sed: -e expression #1, char 15: unknown option to `s'
```

~Argh, `sed` in CI is not receiving `;` 😢~
The escape character was wrong. Sorry

lnicola:
`sed -i 's|"xtask/", "lib/\*", ||' Cargo.toml` works, but we can also take this from the other workflow:

```yaml
          # Only publish the crates under lib/
          sed -i 's|^members = .*$|members = ["lib/*"]|' Cargo.toml
```

I was sure I made this change before, that's why.

lnicola:
I see now that `sed -i 's/"xtask\\/",//; s/ "lib\/\*",//' ./Cargo.toml` is missing a `\`, but it actually worked for me in `fish`.

EDIT: no, not missing, an extra one.

ShoyuVanilla:
Yes, that seems better.
```
sed -i 's/"xtask\\/",//; s/ "lib\/\*",//' ./Cargo.toml
                ^ This redundant backslash is the problem
```

ShoyuVanilla:
Opened #19920 😅

PR #19938, opened 2025-06-06
Add ide-assist: generate_impl_trait for generate_impl

A4-Tacks:
Adds this trait impl for a type.

```rust
trait $0Foo {
    fn foo(&self) -> i32;
}
```
->
```rust
trait Foo {
    fn foo(&self) -> i32;
}

impl Foo for ${1:_} {
    $0fn foo(&self) -> i32 {
        todo!()
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/generate_impl.rs
+++ b/crates/ide-assists/src/handlers/generate_impl.rs
@@ -1,12 +1,17 @@
 use syntax::{
-    ast::{self, AstNode, HasName, edit_in_place::Indent, make},
+    ast::{self, AstNode, HasGenericParams, HasName, edit_in_place::Indent, make},
     syntax_editor::{Position, SyntaxEditor},
 };
 
-use crate::{AssistContext, AssistId, Assists, utils};
+use crate::{
+    AssistContext, AssistId, Assists,
+    utils::{self, DefaultMethods, IgnoreAssocItems},
+};
 
 fn insert_impl(editor: &mut SyntaxEditor, impl_: &ast::Impl, nominal: &ast::Adt) {
     let indent = nominal.indent_level();
+
+    impl_.indent(indent);
     editor.insert_all(
         Position::after(nominal.syntax()),
         vec![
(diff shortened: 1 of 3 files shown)


crates/ide-assists/src/handlers/generate_impl.rs
```diff
@@ -485,4 +585,172 @@ mod tests {
...
+                trait Foo<T> {
+                    fn foo(&self) -> T;
+
+                    fn bar(&self) -> T {
+                        self.foo()
+                    }
+                }
+
+                impl Foo<${1:_}> for ${2:_} {
+                    $0fn foo(&self) -> _ {
```
ShoyuVanilla: Do we need a tabstop here?

ShoyuVanilla:
Besides tabstop, other things look good

crates/ide-assists/src/handlers/generate_impl.rs
A4-Tacks: Used to fill in generic parameters?

ShoyuVanilla: I mean, before the `fn` keyword

crates/ide-assists/src/handlers/generate_impl.rs
```diff
@@ -113,6 +118,101 @@ pub(crate) fn generate_trait_impl(acc: &mut Assists, ctx: &AssistContext<'_>) ->
...
+                        edit.add_placeholder_snippet(cap, generic);
+                    }
+                }
+
+                if let Some(ty) = impl_.self_ty() {
+                    edit.add_placeholder_snippet(cap, ty);
+                }
+
+                if let Some(item) = impl_.assoc_item_list().and_then(|it| it.assoc_items().next()) {
+                    edit.add_tabstop_before(cap, item);
```
ShoyuVanilla: I was confused because I'm in my phone outside 😅 

I mean, do we need this tabstop before the first assoc item of a trait?

A4-Tacks: When implementing traits, positioning the cursor nearby is more convenient for editing `todo!()`

Veykril: Then the tabstop should probably be on the `todo!()`

A4-Tacks: > Then the tabstop should probably be on the `todo!()`

ok

ShoyuVanilla (approved):
Thanks!

PR #19966, opened 2025-06-10
document a diff method for flatpak vscodium in vs_code.md

fluoriteByte:
this adds different method for setting up for vscodium since the codium_wrapper allows for setting up for cargo's path automatically (this also avoids the long path that vscodium for some weird reason constructs)


fluoriteByte:
nvm this is broken jvejinjinjenjvfjn im stupid ,sorry

PR #19975, opened 2025-06-11
hir-ty: test incremental trait solving

davidbarsky:
I think this is mostly what we want for testing the changes in https://github.com/rust-lang/rust-analyzer/pull/19837, as rebasing this commit atop of Chayim's changes causes the second assertion to fail.


ChayimFriedman2 (approved):
Yes, this is what I meant.

PR #20001, opened 2025-06-14
Make `ModuleId` a tracked struct

Veykril:
This is in part an experiment. ~~I am partially hoping for this to have a positive memory impact (as it shrinks `ModuleId` from 12 to 4 bytes)~~ it did not sadge. Funnily enough I also revealed 2 bugs with this (and fixed them).

### DIFF
--- a/crates/hir-def/src/expr_store/tests/body/block.rs
+++ b/crates/hir-def/src/expr_store/tests/body/block.rs
@@ -176,7 +176,6 @@ fn outer() {
 
 #[test]
 fn nested_module_scoping() {
-    cov_mark::check!(nested_module_scoping);
     check_block_scopes_at(
         r#"
 fn f() {
@@ -445,7 +444,6 @@ fn main() {
 fn underscore_import() {
     // This used to panic, because the default (private) visibility inside block expressions would
     // point into the containing `DefMap`, which visibilities should never be able to do.
-    cov_mark::check!(adjust_vis_in_block_def_map);
     check_at(
         r#"
 mod m {
--- a/crates/hir-def/src/expr_store/pretty.rs
+++ b/crates/hir-def/src/expr_store/pretty.rs
@@ -141,10 +141,11 @@ pub fn print_variant_body_hir(db: &dyn DefDatabase, owner: VariantId, edition: E
         let FieldData { name, type_ref, visibility, is_unsafe } = data;
         match visibility {
             crate::item_tree::RawVisibility::Module(interned, _visibility_explicitness) => {
-                w!(p, "{}", interned.display(db, p.edition))
+                w!(p, "pub(in {})", interned.display(db, p.edition))
             }
             crate::item_tree::RawVisibility::Public => w!(p, "pub "),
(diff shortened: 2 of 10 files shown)


crates/hir-def/src/expr_store/tests/body/block.rs
```diff
@@ -474,7 +521,6 @@ fn outer() {
             tests: t
 
             block scope::tests
-            name: _
```
Veykril, line 475: This is one of the bugs. `name` is not visible from here

flodiebold:
What's the second bug?

crates/hir/src/lib.rs
```diff
@@ -4437,7 +4436,7 @@ impl Impl {
             MacroCallKind::Derive { ast_id, derive_attr_index, derive_index, .. } => {
                 let module_id = self.id.lookup(db).container;
                 (
-                    crate_def_map(db, module_id.krate())[module_id.local_id]
```
Veykril, line 4496: Ah sorry. This is the second one. We were potentially using a local id of a block def map for the crate def map here.

crates/ide/src/syntax_highlighting/test_data/highlight_block_mod_items.html
```diff
@@ -53,7 +53,7 @@
...
+                <span class="keyword">struct</span> <span class="struct declaration">Innerest</span><span class="angle">&lt;</span><span class="keyword const">const</span> <span class="const_param const declaration">C</span><span class="colon">:</span> <span class="unresolved_reference">usize</span><span class="angle">&gt;</span> <span class="brace">{</span> <span class="field declaration">field</span><span class="colon">:</span> <span class="bracket">[</span><span class="unresolved_reference">u32</span><span class="semicolon">;</span> <span class="brace">{</span><span class="const_param const">C</span><span class="brace">}</span><span class="bracket">]</span><span class="comma">,</span> <span class="field declaration">field2</span><span class="colon">:</span> <span class="punctuation">&</span><span class="struct">Innerest</span> <span class="brace">}</span>
```
Veykril, line 56: There is a bug here, primtiives don't resolve correctly in modules within blocks

Veykril:
So this does increase memory usage by ~30mb 😕 I am not entirely sure why though I guess we have a lot of block def map queries, and us recording tracked structs in those now makes their memo's allocate the extra slot which is what is observable here? IO would've expected the `ModuleId` shrinking to balance that out at least. It might be better with the new salsa.

ShoyuVanilla:
> So this does increase memory usage by ~30mb 😕 I am not entirely sure why though I guess we have a lot of block def map queries, and us recording tracked structs in those now makes their memo's allocate the extra slot which is what is observable here? IO would've expected the `ModuleId` shrinking to balance that out at least. It might be better with the new salsa.

Could working on #16368 mitigate this memory usage regression?

Veykril:
Likely? Though that might also be orthogonal to this. Block def maps do have a lot of overhear iirespective of the changes here I'd say.

PR #20006, opened 2025-06-15
Cleanup incremental tests and verify query executions

Veykril:


crates/hir-def/src/nameres/tests/incremental.rs
```diff
@@ -399,53 +595,54 @@ pub static ST: u8 = 0;
...
-                    ModuleDefId::TypeAliasId(it) => _ = db.type_alias_signature(it),
-                    ModuleDefId::EnumVariantId(_)
-                    | ModuleDefId::ModuleId(_)
-                    | ModuleDefId::MacroId(_)
-                    | ModuleDefId::BuiltinType(_) => unreachable!(),
-                }
-            }
-        });
-        let n_reparsed_files = events.iter().filter(|it| it.contains("parse(")).count();
-        assert_eq!(n_reparsed_files, 0, "{events:?}");
```
Veykril, line 449: this didn't make sense anymore as we have decoupled everything but the crate-def map from the item tree now

PR #20012, opened 2025-06-16
Update to literal-escaper 0.0.4

lnicola:
Port of https://github.com/rust-lang/rust/pull/140999.


hkBst:
Thanks, @lnicola, maybe next time I'll try it myself.

There is one semantic change, that you may want to be aware of, which is that instead of only preserving the last unescape error, now all unescape errors are collected in a Vec before being copied to the enclosing error Vec.

Veykril (approved):

PR #20018, opened 2025-06-17
fix: Copy lockfiles into target directory before invoking `cargo metadata`

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/19729


crates/project-model/src/cargo_workspace.rs
```diff
@@ -427,8 +480,8 @@ impl CargoWorkspace {
                         current_dir,
                         config,
                         sysroot,
-                        locked,
                         true,
+                        locked,
```
Veykril, line 482: bools considered evil strikes again

PR #20025, opened 2025-06-17
Hide imported privates if private editable is disabled

SoxPopuli:
Hi, I noticed that #19211 introduced a (possible) regression where private imports for a module are showing up in the completion list even with `privateEditable` disabled.

Given that `privateEditable` was the reason for the change in the first place, I thought it made sense to make it conditional


lnicola:
Thanks!

PR #20031, opened 2025-06-18
Fix: Resolve HIR display length issues and improve adjustment tooltips

jnyfah:
- Added detailed, tooltips for adjustment hints 
- Implemented custom `format_function_hint` function that includes function parameters for `closing_braces` and respects maximum length constraints by truncating parameter lists when needed


crates/ide/src/inlay_hints/adjustment.rs
```diff
@@ -162,9 +202,12 @@ pub(super) fn hints(
             linked_location: None,
             tooltip: Some(config.lazy_tooltip(|| {
                 InlayTooltip::Markdown(format!(
-                    "`{}` → `{}` ({coercion} coercion)",
+                    "`{}` → `{}`\n\n**{}**\n\n{}",
                     source.display(sema.db, display_target),
                     target.display(sema.db, display_target),
+                    coercion.chars().next().unwrap().to_uppercase().collect::<String>()
+                        + &coercion[1..],
+                    detailed_tooltip
```
Veykril: let's just manually uppercase the string literal instead of doing this

crates/ide/src/inlay_hints/closing_brace.rs
```diff
@@ -156,6 +156,47 @@ pub(super) fn hints(
     None
 }
 
+fn format_function_hint(func: &ast::Fn, max_length: usize) -> Option<String> {
```
Veykril: Imo we should do this as described in my comment in https://github.com/rust-lang/rust-analyzer/pull/19207#issuecomment-2850163723 instead

Veykril:
Thanks!

PR #20044, opened 2025-06-19
internal: add probe-rs debug provider

Young-Flash:
close https://github.com/rust-lang/rust-analyzer/issues/19948

### DIFF
--- a/editors/code/src/debug.ts
+++ b/editors/code/src/debug.ts
@@ -105,11 +105,13 @@ async function getDebugConfiguration(
         const commandCodeLLDB: string = createCommandLink("vadimcn.vscode-lldb");
         const commandNativeDebug: string = createCommandLink("webfreak.debug");
         const commandLLDBDap: string = createCommandLink("llvm-vs-code-extensions.lldb-dap");
+        const commandProbeRS: string = createCommandLink("probe-rs.probe-rs-debugger");
 
         await vscode.window.showErrorMessage(
             `Install [CodeLLDB](command:${commandCodeLLDB} "Open CodeLLDB")` +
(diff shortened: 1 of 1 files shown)


lnicola:
Should we add it to `knownEngines` too?

Young-Flash:
> Should we add it to `knownEngines` too?

yeah, we need to figure out `ProbeRsDebugConfig`

Young-Flash:
`probe-rs` needs to [configure the target chip](https://probe.rs/docs/tools/debugger/), which seems unsuitable for providing out-of-the-box support in rust-analyzer.

cc @mfp20

mfp20:
> `probe-rs` needs to [configure the target chip](https://probe.rs/docs/tools/debugger/), which seems unsuitable for providing out-of-the-box support in rust-analyzer.
> 
> cc @mfp20

In ${workspaceFolder}/.cargo/config.toml there's probe-rs config:

```
[target.'cfg(all(target_arch = "arm", target_os = "none"))']
runner = "probe-rs run --chip rp2040 --protocol swd --speed 20000"
```

All projects must have it. AFAIK it is not used by the vscode extension, but it is used by vscode as run config. Maybe you can't support the vscode extension but you can hook to that for basic support. Just an idea.

ChayimFriedman2:
What's the status of this @Young-Flash?

Young-Flash:
This stalled because `probe-rs` is not really plug-and-play here, it needs explicit target-specific configuration such as `chip`, which rust-analyzer can't infer generically. So I don't think it fits rust-analyzer's out-of-the-box debug config generation. I'm going to close this for now

PR #20046, opened 2025-06-19
Add `fn parent(self, db) -> GenericDef` to `hir::TypeParam`

regexident:
Currently you have to do `param.merge().parent(db)` for `TypeParam`, which slightly less ergonomic, but most importantly it requires you to be aware of the available detour via `TypeOrConstParam`.

With its sibling `ConstParam` you can already do `param.parent(db)`. Now with `TypeParam` you can, too.


flodiebold:
Thanks!

PR #20047, opened 2025-06-19
internal: Utilize `cargo check --compile-time-deps`

ShoyuVanilla:
rust-lang/cargo#15674 has been merged and I guess it would be available in toolchain `1.90.0`

I've tested this on rust-analyzer itself with cargo binary built from current master with the following my rust-analyzer branch
https://github.com/ShoyuVanilla/rust-analyzer/tree/comp-time-deps-test
and these are the results (I had run `cargo clean` before each run and modified the code a bit to use `--compile-time-deps` with injected cargo binary path iff env var `$CARGO_BIN_FOR_TEST` is set)

## With `--compile-time-deps`

image
 
## Without `--compile-time-deps`

image

## Comparison

|   | **`analysis-stat` results** | **Time** | **`target` dir size** |
| - | - | - | - |
| **`--compile-time-deps`**  | Identical | 2m 25s | 347 M |
| **default**  | Identical | 2m 55s | 1.4 G |

## System Information

image

### DIFF
--- a/crates/project-model/src/build_dependencies.rs
+++ b/crates/project-model/src/build_dependencies.rs
@@ -20,7 +20,9 @@ use toolchain::Tool;
 
 use crate::{
     CargoConfig, CargoFeatures, CargoWorkspace, InvocationStrategy, ManifestPath, Package, Sysroot,
-    TargetKind, utf8_stdout,
+    TargetKind,
+    toolchain_info::{QueryConfig, version},
+    utf8_stdout,
 };
 
 /// Output of the build script and proc-macro building steps for a workspace.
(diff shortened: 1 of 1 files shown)


lnicola:
:heart_eyes:

edwin0cheng:
Do we want to disable `RUSTC_WRAPPER` too ?

Veykril:
We should not set the wrapper if we use the flag

ShoyuVanilla:
> Do we want to disable `RUSTC_WRAPPER` too ?

Oh, I missed it. Thanks!

Veykril (approved):

Veykril:
This is amazing, can't wait to get rid of the wrapper in the future

bjorn3:
Does this work in the rust-lang/rust workspace too?

ShoyuVanilla:
> Does this work in the rust-lang/rust workspace too?

Haven't tested in it yet but I guess so?

bjorn3:
We set `rust-analyzer.cargo.buildScripts.overrideCommand` to `./x.py check --json-output`. `./x.py check --json-output -Zunstable-options --compile-time-deps` is not a valid command, which makes it seem to me like `cargo_comp_time_deps_available` is false (maybe `toolchain` is even `None`). We should probably add support for `--compile-time-deps` to `x.py` and then we could set it in the rust-analyzer config, but at the same time rust-analyzer should avoid setting `RUSTC_WRAPPER`.

ShoyuVanilla:
Oh, I forgot about those `x` scripts in rust. Currently rust-analyzer is not appending `--compile-time-deps` when `overrideCommands` is set. #20121 (And setting that also ignores `RUSTC_WRAPPER`, so I guess currently in rust-lang/rust, neither `--compile-time-deps` nor `RUSTC_WRAPPER` is executed and thus full `cargo check` is happening 🙃 )

So, adding supports for `--compile-time-deps` to `./x.py` sounds just enough to me.

bjorn3:
Opened https://github.com/rust-lang/rust/pull/143785. Saves a lot of time!

ShoyuVanilla:
That's way better than my naive expectations 🎉

PR #20049, opened 2025-06-20
Fix: Handle Self replacement contextually in inline assists

### LINKED ISSUE #19827: Inline function assist produces invalid code when inlining across generic impl blocks
I hit this with a generic `From` impl, but it can be reduced to two inherent generic impls:

```rust
struct Generic<T>(T);

impl<T> Generic<T> {
    fn new() -> Self {
        // can also use Self::new2()
        Generic::new2()
    }
}

impl<T> Generic<T> {
    fn new2() -> Self {
        Self(todo!())
    }
}
```

inlining `new2` currently changes the first `impl` block to:

```rust
impl<T> Generic<T> {
    fn new() -> Self {
        Generic<T>(todo!())
    }
}
```

but this is invalid syntax. `Generic<T>` in the function body should be one of

- `Self`
- `Generic`
- `Generic::<T>`

**rust-analyzer version**: 0.4.2467-standalone
**rustc version**: 1.89.0-nightly (414482f6a 2025-05-13)
**editor or extension**: VSCode (ext version 0.4.2467)

@rustbot label +A-assists

phyBrackets:
Only replace Self when inlining across different impl contexts, and use
bare type names for expression contexts to avoid turbofish issues.

Fixes #19827

### DIFF
--- a/crates/ide-assists/src/handlers/inline_call.rs
+++ b/crates/ide-assists/src/handlers/inline_call.rs
@@ -19,7 +19,7 @@ use ide_db::{
 };
 use itertools::{Itertools, izip};
 use syntax::{
-    AstNode, NodeOrToken, SyntaxKind,
+    AstNode, NodeOrToken, SyntaxKind, SyntaxToken,
     ast::{
         self, HasArgList, HasGenericArgs, Pat, PathExpr, edit::IndentLevel, edit_in_place::Indent,
     },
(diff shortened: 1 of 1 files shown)


Veykril (changes requested):
So, we should make `PathTransform` do this transformation (it might already have parts of it implemented), and then use that in the other places in this assist (so far its only used for generic argument rewriting). That way we don't special case things to this assist when its actually a more general applicable transformation

phyBrackets:
Thanks @Veykril for the feedback, I think it does make sense that we should leverage and extend the existing `PathTransform` mechanism that's already used for generic argument rewriting. I have updated the patch based on your review.

crates/ide-db/src/path_transform.rs
```diff
@@ -248,6 +248,47 @@ fn preorder_rev(item: &SyntaxNode) -> impl Iterator<Item = SyntaxNode> {
...
+        if ast::Type::cast(node.clone()).is_some()
+            || ast::Pat::cast(node.clone()).is_some()
+            || ast::RetType::cast(node.clone()).is_some()
+        {
+            return false;
+        }
+        current = node.parent();
+    }
+    false
+}
```
Veykril, line 282: This can be simplified (and also made more robust by that)
```suggestion
fn is_path_in_expression_context(path: &ast::Path) -> bool {
	let path = path.top_path();
    let Some(parent) = path.syntax().parent() else { return false; };
    parent.kind() == SyntaxKind::PATH_EXPR
}
```
I don't think we need to consider record expression constructors, as their path is actually in type namespace if I am not mistaken.

crates/ide-db/src/path_transform.rs
```diff
@@ -413,10 +454,17 @@ impl Ctx<'_> {
                         true,
                     )
                     .ok()?;
-                let ast_ty = make::ty(ty_str).clone_for_update();
+
+                // Context-aware replacement
+                let replacement = if is_path_in_expression_context(&path) {
+                    let bare_name = get_bare_type_name(ty_str);
```
Veykril, line 460: I think the more correct approach would be to replace the `Self` with the fully qualified path in turbofished form instead of dropping the generics entirely.

crates/ide-db/src/path_transform.rs
phyBrackets: Okay, my original concern was about struct constructor calls like `Foo { field: value }` where we might want Foo instead of `Self::Foo`, but thinking about it more, record expressions are indeed type namespace references and even if we wanted bare names there, `PathTransform` should handle the namespace correctly, so this complexity isn't justified indeed, thanks.

Veykril: We can always add back heuristics later that can scan for cases where we can simplify the path, but for now that's not really necessary I think

ChayimFriedman2:
OP didn't respond, closing. Feel free to reopen/open a new PR if you want to resume work on this.

PR #20050, opened 2025-06-20
Add better documentation for excluding imports from symbol search

LHolten:
This should make the option less confusing.

@joshka do you think this documentation is good?

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -762,7 +762,11 @@ config_data! {
         /// though Cargo might be the eventual consumer.
         vfs_extraIncludes: Vec<String> = vec![],
 
-        /// Exclude imports from symbol search.
+        /// Exclude all imports from workspace symbol search.
+        ///
+        /// In addition to regular imports (which are always excluded),
+        /// this option removes public imports (better known as re-exports)
+        /// and removes imports that rename the imported symbol.
         workspace_symbol_search_excludeImports: bool = false,
         /// Workspace symbol search kind.
         workspace_symbol_search_kind: WorkspaceSymbolSearchKindDef = WorkspaceSymbolSearchKindDef::OnlyTypes,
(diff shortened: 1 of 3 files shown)


crates/rust-analyzer/src/config.rs
```diff
@@ -762,7 +762,11 @@ config_data! {
         /// though Cargo might be the eventual consumer.
         vfs_extraIncludes: Vec<String> = vec![],
 
-        /// Exclude imports from symbol search.
+        /// Exclude all imports from workspace symbol search.
```
joshka, line 765: It may be worth defining what symbol search is (eg specifically where this ends up in the ui)

joshka (approved):
LGTM.

crates/rust-analyzer/src/config.rs
LHolten: This doesn't seem like the right location to explain symbol search to me. In my opinion that should be the job of the editor that exposes the feature.

joshka: What I'm saying is that VSCode doesn't have a feature named "symbol search". This is the internal LSP name that's not shared with anything that appears in the VSCode UI.

VSCode has "Go To Symbol" action, and it has various completion behavior that triggers searches of symbols. It also has "Find all references" (to a symbol), and "Find all Implementations". I wouldn't be surprised if there were other UIs that had some form of searching capability over symbols. So it's unclear which of these these settings control.

Digging into the source code to work it out suggests that this is probably only the "Go To Symbol" action and that it's documented (sort of) in https://rust-analyzer.github.io/book/features.html#workspace-symbol (which I found by looking for all instances of "workspace.symbol" in the source). Is that the only place that this behavior surfaces, or does it affect other places?

As an aside, I'd intuitively thought that this actually applied to the symbol search that appears when completing symbols, so my complaint about confusion is grounded in actual confusion, not a hypothetical one (I've used VSCode every day for many years - I'm not unfamiliar with it).

I suspect that the symbol search action was possibly renamed at some point, or maybe we're just used to referring to it by the internal name rather than the UI name, but the settings are configured with respect to the UI names, so they should reference those names. I'm guessing a setting rename probably wasn't anticipated here, but it might be appropriate.

PR #20065, opened 2025-06-23
Add ide-assist; Generate AsRef impl from Borrow

A4-Tacks:
Generate `AsRef` implement from `Borrow`.

```rust
struct Foo<T>(T);

impl<T> $0Borrow<T> for Foo<T> {
    fn borrow(&self) -> &T {
        &self.0
    }
}
```
->
```rust
struct Foo<T>(T);

$0impl<T> AsRef<T> for Foo<T> {
    fn as_ref(&self) -> &T {
        &self.0
    }
}

impl<T> Borrow<T> for Foo<T> {
    fn borrow(&self) -> &T {
        &self.0
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/generate_asref_impl_from_borrow.rs
+++ b/crates/ide-assists/src/handlers/generate_asref_impl_from_borrow.rs
@@ -1,0 +1,251 @@
+use syntax::ast::edit_in_place::Indent;
+use syntax::ast::make;
+use syntax::ast::{self, AstNode, HasName};
+use syntax::ted;
+
+use crate::{AssistContext, AssistId, Assists};
+
+// Assist: generate_asref_impl_from_borrow
+//
+// Generate `AsRef` implement from `Borrow`.
+//
+// ```
+// struct Foo<T>(T);
+//
+// impl<T> $0Borrow<T> for Foo<T> {
+//     fn borrow(&self) -> &T {
+//         &self.0
+//     }
+// }
+// ```
+// ->
+// ```
+// struct Foo<T>(T);
+//
+// $0impl<T> AsRef<T> for Foo<T> {
+//     fn as_ref(&self) -> &T {
+//         &self.0
+//     }
+// }
+//
+// impl<T> Borrow<T> for Foo<T> {
+//     fn borrow(&self) -> &T {
+//         &self.0
+//     }
+// }
+// ```
+//
+// ---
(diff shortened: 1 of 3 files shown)


crates/ide-assists/src/handlers/generate_asref_impl_from_borrow.rs
```diff
@@ -0,0 +1,251 @@
+use syntax::ast::edit_in_place::Indent;
+use syntax::ast::make;
+use syntax::ast::{self, AstNode, HasName};
+use syntax::ted;
```
ShoyuVanilla: As this is a new assist and relatively simple, could this be [implemented with new `SyntaxEditor` infra](https://github.com/rust-lang/rust-analyzer/pull/19789#discussion_r2120642713)?

crates/ide-assists/src/handlers/generate_asref_impl_from_borrow.rs
```diff
@@ -0,0 +1,251 @@
...
+    let path = ast::PathType::cast(impl_.trait_()?.syntax().clone())?;
+    let indent = impl_.indent_level();
+
+    let name = path.path()?.segment()?.name_ref()?;
+
+    let (target_name, target_method_name) = match &*name.text() {
+        "Borrow" => ("AsRef", "as_ref"),
+        "BorrowMut" => ("AsMut", "as_mut"),
+        _ => return None,
+    };
```
ShoyuVanilla: This would be better semantically like:
https://github.com/rust-lang/rust-analyzer/blob/5b852da4c3852d030ebca67746980bc0288a85ec/crates/ide-assists/src/handlers/convert_into_to_from.rs#L35-L44

A4-Tacks: Using strings can be easily implemented without adding new methods to FamousDefs

ShoyuVanilla: Yes, it's easier but might be incorrect, like if there are other traits with the same names.

crates/ide-assists/src/handlers/generate_asref_impl_from_borrow.rs
A4-Tacks: ok

A4-Tacks:

PR #20072, opened 2025-06-23
fix: Respect `.cargo/config.toml` `build.target-dir`

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/20070

### DIFF
--- a/crates/project-model/src/env.rs
+++ b/crates/project-model/src/env.rs
@@ -1,6 +1,6 @@
 //! Cargo-like environment variables injection.
 use base_db::Env;
-use paths::Utf8Path;
+use paths::{Utf8Path, Utf8PathBuf};
 use rustc_hash::FxHashMap;
 use toolchain::Tool;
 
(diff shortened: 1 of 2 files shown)


crates/project-model/src/env.rs
```diff
@@ -123,6 +123,26 @@ fn parse_output_cargo_config_env(manifest: &ManifestPath, stdout: &str) -> Env {
...
+    let mut cargo_config = sysroot.tool(Tool::Cargo, manifest.parent(), extra_env);
+    cargo_config
+        .args(["-Z", "unstable-options", "config", "get", "build.target-dir"])
+        .env("RUSTC_BOOTSTRAP", "1");
+    if manifest.is_rust_manifest() {
+        cargo_config.arg("-Zscript");
+    }
+    utf8_stdout(&mut cargo_config)
+        .map(|stdout| {
+            Utf8Path::new(stdout.trim_start_matches("build.target-dir =").trim_matches('"'))
```
ShoyuVanilla:
```suggestion
            Utf8Path::new(stdout.trim_start_matches("build.target-dir = ").trim_matches('"'))
```
I think we need an extra space here

crates/project-model/src/env.rs
```diff
@@ -123,6 +123,26 @@ fn parse_output_cargo_config_env(manifest: &ManifestPath, stdout: &str) -> Env {
...
+    let mut cargo_config = sysroot.tool(Tool::Cargo, manifest.parent(), extra_env);
+    cargo_config
+        .args(["-Z", "unstable-options", "config", "get", "build.target-dir"])
+        .env("RUSTC_BOOTSTRAP", "1");
+    if manifest.is_rust_manifest() {
+        cargo_config.arg("-Zscript");
+    }
+    utf8_stdout(&mut cargo_config)
+        .map(|stdout| {
+            Utf8Path::new(stdout.trim_start_matches("build.target-dir = ").trim_matches('"'))
```
lnicola, line 140: Nit: this breaks if the path contains quotes, e.g. `build.target-dir = 'a " b'` or `build.target-dir = """a " ' b"""`.

Veykril: Right. There is general cleanup potential here. For one, we invoke `cargo config` three times now for the different things, which is unnecessary, we can just invoke it plain and read out the info from that we want from that output. Secondly, it supports `json` output (and `json-value` though I have no idea what that means). We should just use the json output for more robust deserialization.

lnicola: `json` is `{"build":{"target-dir":"a \" b"}}`, `json-value` is `"a \" b"`.

Veykril: Ah okay, it made no difference when I tested it but that was because I ran it without a target config so it returned the root object in both :D

PR #20077, opened 2025-06-23
feat: Extend vscode 'run' command with optional mode argument for run…

markpots5:
…ning test(s) or bin at keyboard cursor

Fixes https://github.com/rust-lang/rust-analyzer/issues/14805

### DIFF
--- a/editors/code/src/ctx.ts
+++ b/editors/code/src/ctx.ts
@@ -1,7 +1,6 @@
 import * as vscode from "vscode";
 import type * as lc from "vscode-languageclient/node";
 import * as ra from "./lsp_ext";
-import * as commands from "./commands";
 
 import { Config, prepareVSCodeConfig } from "./config";
 import { createClient } from "./client";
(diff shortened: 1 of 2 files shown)


markpots5:
To test executing the 'run' command with 'cursor' as supplied argument one can for example add the following to keybindings.json:
`
{
  "key": "ctrl+p",
  "command": "rust-analyzer.run",
  "args": "cursor"
}
`

editors/code/src/ctx.ts
```diff
@@ -462,9 +463,17 @@ export class Ctx implements RustAnalyzerExtensionApi {
...
-                callback = factory.enabled(this);
+                if (name === "run") {
+                    // Special case: support optional argument for `run`
+                    callback = (mode?: "cursor") => {
+                        return commands.run(this, mode)();
+                    };
+                } else {
+                    // we asserted that `client` is defined
+                    callback = factory.enabled(this);
+                }
```
Veykril: I don't think this is necessary, `Cmd` (the type of callback) is already correctly typed in allowing more parameters. Instead we need to change the `enabled` field here https://github.com/rust-lang/rust-analyzer/blob/ab9e7bdc83f75e0255e7427bb2b5fbfeacc79e3c/editors/code/src/main.ts#L170 to take the special callback instead

Veykril: Ah In fact that doesn't need changes either. Just having edited the signature of `command.run` should suffice

markpots5: I changed it. I've tried to only revert the special casing in command factory (ctx.ts). However it seems adding the special callback as you suggest in your first comment is necessary to make it work.

Veykril: Ah indeed, `enabled` is not a `Cmd` but a function that returns a `Cmd` 👍

Veykril:
Thanks!

PR #20084, opened 2025-06-24
ci:  Fix cancel parallel jobs on windows

ShoyuVanilla:
Fixes rust-lang/rust-analyzer#20079

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,10 @@ env:
   RUST_BACKTRACE: short
   RUSTUP_MAX_RETRIES: 10
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   changes:
     runs-on: ubuntu-latest


lnicola:
I think we can run `curl.exe` (not that there's anything wrong with `Invoke-RestMethod`, so it doesn't matter). I wonder if it wouldn't be nicer to write that command on a single line and avoid having two versions.

ShoyuVanilla:
> I think we can run `curl.exe` (not that there's anything wrong with `Invoke-RestMethod`, so it doesn't matter). I wonder if it wouldn't be nicer to write that command on a single line and avoid having two versions.

> Can't we just change the shell for windows to bash? https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrunshell

I was going to do this in a breeze but `pwsh` was more harder than I thought and you saw my follies 😅 

Okay, I'll try with `bash` and `curl.exe`

ShoyuVanilla:
Oh, this worked well. I should have tried bash first 😂 
https://github.com/rust-lang/rust-analyzer/actions/runs/15852789504/job/44690400616
image

ShoyuVanilla:
So the following was my chain of thoughts:

- There might be no `msys/mingw` on Windows runner by default
- Even so, there might be no `curl` on it
- Even so, it might be not on `$PATH` by default

Thus, I just try to do things with `Invoke-RestMethod`, which is aliased to `curl` in `pwsh` and I thought it would work well, but it somehow fails to deliver auth header correctly and I don't know why.

What I lacked of were faith to belive Windows runner and determination to try brand new alternative on my first fail 😅

lnicola:
Welp, I knew about bash on Windows but lately I'm a little too biased against changing the defaults.

But I just realized we couldn't have kept a single step if we ran `curl.exe` 🤦.

ShoyuVanilla:
Yes, might be hard to correctly locate `curl.exe` in a single script.
BTW, I looked into the above screenshot with `bash` carefully and just found out that the initial `Invoke-RestMethod` thing actually worked with the same result and the only difference was its exit code was not `0`

https://github.com/rust-lang/rust-analyzer/actions/runs/15852059976/job/44687827150
image
(Responses are same with the former screenshot)

Besides those exit code behaviors (might be) related to HTTP status code, I have no idea why GitHub returns `403` but handles the requests as intended 🙃

PR #20105, opened 2025-06-26
Parse new const trait syntax

Veykril:
cc https://github.com/rust-lang/rust/pull/139858#issuecomment-3005310132

### DIFF
--- a/crates/syntax/rust.ungram
+++ b/crates/syntax/rust.ungram
@@ -669,7 +669,7 @@ TypeBoundList =
 
 TypeBound =
   Lifetime
-| ('~' 'const' | 'const')? 'async'? '?'? Type
+| ('~' 'const' | '[' 'const' ']' | 'const')? 'async'? '?'? Type
 | 'use' UseBoundGenericArgs
 
 UseBoundGenericArgs =
--- a/crates/parser/src/grammar/generic_params.rs
+++ b/crates/parser/src/grammar/generic_params.rs
@@ -122,7 +122,7 @@ fn lifetime_bounds(p: &mut Parser<'_>) {
 }
 
 // test type_param_bounds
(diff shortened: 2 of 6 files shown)


lnicola:
Should we ask oli-obk to drop the r-a changes from that PR? I know they shouldn't cause conflicts, but..

oli-obk:
yea sorry, overlooked that. I just searched and replaced

Veykril:
No worries, same diffs overlapping don't cause conflicts afaik and fixing this (as in the tests failing) before the next sync wouldn't have been a lot of trouble (if it was I would've raised it in your PR)

crates/syntax/rust.ungram
```diff
@@ -669,7 +669,7 @@ TypeBoundList =
 
 TypeBound =
   Lifetime
-| ('~' 'const' | 'const')? 'async'? '?'? Type
+| ('~' 'const' | '[' 'const' ']' | 'const')? 'async'? '?'? Type
```
Veykril, line 672: It is funny how we now have 3 variations of the maybe syntax in here, we should start pruning them once its stabilized

oli-obk: 2 of them are going to stay around, just `const` means you can use it in an inline const block

Veykril: Ah I was referring to `~const`, `[const]` and `const?` (note the `'?'?` rule, which I don't think is used anymore for anything)

oli-obk: the `'?'?` is in front of the "Type" (Trait) to support `?Sized`

Veykril: Oh right that's the `?` bound whoops 🤦‍♂️

PR #20109, opened 2025-06-26
feat: Generate `new` for tuple struct

Hmikihiro:
Closes: https://github.com/rust-lang/rust-analyzer/issues/20071
I will generate `fn new` for tuple structs just like struct with named fields.
For tuple structs, I use the element index of tuple instead of a field name. 

```rs
struct S<T>(String, PhantomData<T>, u32);

impl<T> S<T> {
    fn new(_0: String, _2: u32) -> Self {
        Self(_0, PhantomData, _2)
    }
}
```

```rs
struct S<T> {
    text: String,
    _marker: PhantomData<T>,
    value: u32,
}

impl<T> S<T> {
    fn new(text: String, value: u32) -> Self {
        Self { text, _marker: PhantomData, value }
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/generate_new.rs
+++ b/crates/ide-assists/src/handlers/generate_new.rs
@@ -35,10 +35,16 @@ use crate::{
 pub(crate) fn generate_new(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
     let strukt = ctx.find_node_at_offset::<ast::Struct>()?;
 
-    // We want to only apply this to non-union structs with named fields
     let field_list = match strukt.kind() {
-        StructKind::Record(named) => named,
-        _ => return None,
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
Using `_number` is a bad idea. We have [`NameGenerator`](https://github.com/rust-lang/rust-analyzer/blob/master/crates/ide-db/src/syntax_helpers/suggest_name.rs#L94) and you can also place a tabstop on each name.

Hmikihiro:
Thanks @ChayimFriedman2 for the feedback.  
I would use `NameGenerator` to generate names instead of _number, and use placeholder to change each name.
スクリーンショット 2025-06-27 18 58 06

ShoyuVanilla (approved):
Looks good to me

PR #20111, opened 2025-06-27
Bug Fix: Auto import has broken for vim-lsp

luoyancn:
1. Because of change the in [1], vim-lsp cannot complete the auto import items. May be, it`s better to use the method completion_item_edit_resolve [2] to resovle this problem. Also, maybe this change is only a workaround.

  close the bug rust-lang/rust-analyzer#19401

  [1]: https://github.com/rust-lang/rust-analyzer/blob/master/crates/rust-analyzer/src/lsp/capabilities.rs#L188
  [2]: https://github.com/rust-lang/rust-analyzer/blob/master/crates/rust-analyzer/src/lsp/capabilities.rs#L211

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -1549,7 +1549,7 @@ impl Config {
             limit: self.completion_limit(source_root).to_owned(),
             enable_term_search: self.completion_termSearch_enable(source_root).to_owned(),
             term_search_fuel: self.completion_termSearch_fuel(source_root).to_owned() as u64,
-            fields_to_resolve: if self.client_is_neovim() {
+            fields_to_resolve: if self.client_is_neovim() || self.client_is_vim_lsp() {
                 CompletionFieldsToResolve::empty()
             } else {
                 CompletionFieldsToResolve::from_client_capabilities(&client_capability_fields)
@@ -2361,6 +2361,10 @@ impl Config {
 
     pub fn client_is_neovim(&self) -> bool {
         self.client_info.as_ref().map(|it| it.name == "Neovim").unwrap_or_default()
+    }
+
+    pub fn client_is_vim_lsp(&self) -> bool {
+        self.client_info.as_ref().map(|it| it.name == "vim-lsp").unwrap_or_default()
     }
 }
 // Deserialization definitions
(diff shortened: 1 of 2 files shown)


ShoyuVanilla:
I think the current implementation is not problematic in general and "reverting" #18589 is not the right fix for this.

I might be wrong because I'm not familiar with implementation details of `vim-lsp` but plausible solution would be one of those:
- Check for the property `"additionalTextEdits"` here as well
  https://github.com/rust-lang/rust-analyzer/blob/5b852da4c3852d030ebca67746980bc0288a85ec/crates/rust-analyzer/src/lsp/capabilities.rs#L188-L193
- Make a workaround for `vim-lsp` as well, like in #18630
- Fix `vim-lsp` to provide proper default client capabilities in [here](https://github.com/prabirshrestha/vim-lsp/blob/04428c920002ac7cfacbecacb070a8af57b455d0/autoload/lsp.vim#L550)
- Edit your own vim script to provide proper client capabilities for rust-analyzer with [this](https://github.com/prabirshrestha/vim-lsp/blob/04428c920002ac7cfacbecacb070a8af57b455d0/doc/vim-lsp.txt#L945-L956)

Veykril:
That seems like a bug in `vim-lsp`, rust-analyzer does not resolve `additionalTextEdits` lazily right now, so we do not consider it when declaring whether we have a resolve provider or not which is the correct behavior. Now, we probably could resolve it lazily as well, but that will only mask the bug in `vim-lsp` (as would adding a workaround for it like we do with neovim today, which has the same bug I believe)

luoyancn:
@ShoyuVanilla Thanks for your advise.I will try to find much better solutions.

luoyancn:
@ShoyuVanilla ,Thanks for your help. May be, it`s better not to change the rust-analyzer codes. Accoding to your advise, i resolve this problem by add the follows in vimrc:

```
function! s:RustCapabilities(server_info) abort
  let caps = lsp#default_get_supported_capabilities(a:server_info)
  if get(a:server_info, 'name', '') ==# 'rust-analyzer'
    let caps.textDocument.completion.completionItem.resolveSupport = {
          \ 'properties': ['additionalTextEdits', 'tags', 'labelDetails',
          \                'detail', 'documentation', 'filterText',
          \                'textEdit', 'command']
          \ }
    return caps
  endif
endfunction
let g:lsp_get_supported_capabilities = [function('s:RustCapabilities')]
```


So, this issue is not a bug. Ignore my merge request.

PR #20114, opened 2025-06-28
Update README.md

jmikedupont2:
please mention at top level dir

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ If you want to **use** rust-analyzer's language server with your editor of
 choice, check [the manual](https://rust-analyzer.github.io/book/).
 It also contains some tips & tricks to help you be more productive when using rust-analyzer.
 
+Otherwise you can install from git with `cargo xtask install --server`
+
 ## Security and Privacy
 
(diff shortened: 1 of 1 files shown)


ShoyuVanilla:
Well, I don't think we should inform this in README. The binary that installed with `cargo xtask` is not for "retails":

https://github.com/rust-lang/rust-analyzer/blob/e9968fc5553993f26b8eab2c0c664ae92b433fe1/xtask/src/install.rs#L138-L145

That enables `force-always-assert`, which turns on panicking on many internal assertions that just leaving error logs in our "retail" build. So, this might cause irritating panics to uses, so we cannot recommend that by default.
Also, the manual page already contains various ways to install `rust-analyzer` and it's not hidden so deep inside the nested chapters, so I think current manual page link is enough

lnicola:
Agreed, let's link to https://rust-analyzer.github.io/book/rust_analyzer_binary.html instead (even though there's another link to the manual right above, maybe change that to https://rust-analyzer.github.io/book/installation.html.

And the `force-always-assert` thing isn't actually documented in the manual.

ChayimFriedman2:
As others have said this is not the place.

PR #20122, opened 2025-06-29
Remove unnecessary parens in closure

chenyukang:
From https://github.com/rust-lang/rust/pull/136906#issuecomment-3016602267


ChayimFriedman2 (approved):
Thanks!

PR #20124, opened 2025-06-30
Remove last use of `rustc_pattern_analysis::Captures`

zachs18:
It's not necessary anymore due to Rust 2024 lifetime capture rules.

cc <https://github.com/rust-lang/rust/pull/137334>


Veykril (approved):
Thanks!

PR #20132, opened 2025-07-01
Add AsMut, Borrow and BorrowMut to minicore and famous_defs

A4-Tacks:


crates/test-utils/src/minicore.rs
```diff
@@ -377,11 +380,30 @@ pub mod convert {
         fn as_ref(&self) -> &T;
     }
     // endregion:as_ref
+    // region:as_mut
+    pub trait AsMut<T: crate::marker::PointeeSized>: crate::marker::PointeeSized {
+        fn as_ref(&mut self) -> &mut T;
```
phyBrackets: this should be `as_mut`?

phyBrackets (changes requested):
and are there tests that verify these trait lookups work correctly?

crates/test-utils/src/minicore.rs
A4-Tacks: oh sorry

A4-Tacks:
> and are there tests that verify these trait lookups work correctly?

I don't know

Veykril (changes requested):
Please copy the source definitions from the actual std library (ideally from the `beta` or `nightly` channel for `minicore`

Veykril (approved):

PR #20134, opened 2025-07-01
Migrate `replace_is_method_with_if_let_method` Assist to use `SyntaxFactory`

Hmikihiro:


lnicola:
I guess this looks fine, but our MSRV is currently 1.86, so we can't use `if-let` chains. Not sure if we want to bump it right away.

Hmikihiro:
> I guess this looks fine, but our MSRV is currently 1.86, so we can't use `if-let` chains. Not sure if we want to bump it right away.

I see.
I’ll remove the if-let chains and refactor the code to be compatible with MSRV 1.86.

lnicola (approved):

PR #20140, opened 2025-07-01
feat: Implement `textDocument/documentColor`

nik-rev:
Hi, I want to add `textDocument/documentColor` method to rust-analyzer, it will allow libraries with a type like `Color { r: u8, g: u8, b: u8 }` to annotate this type with special attributes `#[rust_analyzer::color::rgb]`.

IDEs can then annotate expressions known at compile time that of type `Color`
with color swatches, similar to how it works for CSS:

image

I added the feature to Helix from the client side (https://github.com/helix-editor/helix/pull/12308) and think it'd be fun to also add it from the server side. 

---

My current plan is to have an API like this:

```
#[rust_analyzer::color::rgb]
struct Color {
    #[rust_analyzer::color::rgb::r]
    red: f32,
    #[rust_analyzer::color::rgb::g]
    green: f32,
    #[rust_analyzer::color::rgb::b]
    blue: f32,
}

// the expression `Color { ... }` will have a red square next to it!
const RED: Color = Color {
    red: 255,
    green: 0,
    blue: 0,
};
```

This is what I want for a MVP:

 1. Visit every expression that is known at compile time (is `const`)
 1. If the type `struct` of the expression contains
    an attribute `#[rust_analyzer::color::rgb]`, proceed
 1. If there are 3 `f32` fields with attribute `#[rust_analyzer::color::rgb::{r,g,b}]`,
    then we have all the necessary information to construct the type itself

Currently, the problem that I face is that I cannot obtain the `Type` of an expression. `type_of_expr` returns `None`:

```rs
for event in file.preorder() {
    let syntax::WalkEvent::Leave(syntax_node) = event else { continue };
    let Some(expr) = syntax::ast::Expr::cast(syntax_node) else { continue };
    dbg!(&expr);

    // `type_of_expr` fails... but why?
    let Some(ty) = sema.type_of_expr(&expr) else { continue };
    dbg!(ty);
}
```

I'm building rust-analyzer with `cargo run --release` then in `.helix/languages.toml` I tell the editor to use the built binary:

```toml
[language-server.rust-analyzer]
command = "/home/e/contrib/rust-analyzer/target/release/rust-analyzer"
```

This works. I get hover information, auto-complete. I get custom logs in Helix's log (`:log-open`).

The current implementation draws a single red square in the file, so I assume the problem isn't with my setup

### DIFF
--- a/crates/ide/src/document_color.rs
+++ b/crates/ide/src/document_color.rs
@@ -1,0 +1,65 @@
+#![allow(warnings)]
+
+use hir::{EditionedFileId, Semantics};
+use ide_db::RootDatabase;
+use span::FileId;
+use syntax::AstNode as _;
+
+pub struct DocumentColor {
+    pub start_line: u32,
+    pub start_char: u32,
+    pub end_char: u32,
+    pub end_line: u32,
+    pub r: f32,
+    pub g: f32,
+    pub b: f32,
+    pub a: f32,
+}
+
+/// Document colors let IDEs annotate regions of source code as having a particular color.
+///
+/// The advantage over `textDocument/documentHighlight` is that IDEs can create virtual text
+/// to show swatches, representing the color.
+///
+/// # Basic Plan
+///
+/// 1. Visit every expression that is known at compile time (is `const`)
+/// 1. If the type `struct` of the expression contains
+///    an attribute `#[rust_analyzer::color::rgb]`, proceed
+/// 1. If there are 3 `f32` fields with attribute `#[rust_analyzer::color::rgb::{r,g,b}]`,
+///    then we have all the necessary information to construct the type itself
+pub(crate) fn document_color(db: &RootDatabase, file_id: FileId) -> Vec<DocumentColor> {
+    let sema = Semantics::new(db);
+    let file_id = sema
+        .attach_first_edition(file_id)
+        .unwrap_or_else(|| EditionedFileId::current_edition(db, file_id));
+
+    let file = sema.parse(file_id);
+    let file = file.syntax();
+
+    for event in file.preorder() {
+        let syntax::WalkEvent::Leave(syntax_node) = event else { continue };
+        let Some(expr) = syntax::ast::Expr::cast(syntax_node) else { continue };
+        dbg!(&expr);
+
+        // `type_of_expr` fails... but why?
+        let Some(ty) = sema.type_of_expr(&expr) else { continue };
+        dbg!(ty);
+    }
+
+    vec![DocumentColor {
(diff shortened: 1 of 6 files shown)


ChayimFriedman2:
@nik-rev, we haven't heard back from you.

@rust-lang/rust-analyzer we also haven't heard from you: do you think we want this? I personally think if we do, we need to put some time to consider the API.

nik-rev:
> we haven't heard back from you

Hey, I am still interested in coming back to this. Let me know what kind of API you prefer. The one I had in mind with the `rust_analyzer::color` attribute is probably suitable for a POC

ChayimFriedman2:
We cannot really have a POC: we don't have a nightly system. If we publish the API, people we use it then we won't be able to break it anymore. If not, it will remain unused (or people will use it anyway). So we need to get it right from the beginning.

ChayimFriedman2:
No answer from OP, closing. @nik-rev feel free to reopen or open a new PR if you want again to work on this.

nik-rev:
Commenting in case anyone, or myself wants to pick this up in the future, again.

https://github.com/nik-contrib/rust-analyzer/commit/46e56980d9a63323e010b650bea48256f924beed

The basic MVP I tried to implement is:

- ✅ Walk all expressions
- ✅ For all expressions that are ADTs with the `#[rust_analyzer::color()]` attribute
- ❌ Evaluate the expression into a `Const<'_>`
- ❌ Extract the fields for red, green and blue colors from the `&[u8]` "memory" region
- ❌ Rest is easy

The hard part is step 3. In my PR I:

- Obtained `ExprId`, `ast::Expr` and `Type` information for all expressions
- But I wasn't able to figure out how to just evaluate an arbitrary constant expression

The PR contains this line:

```rust
dbg!(konst)
```

Which prints:

```
[crates/hir-ty/src/infer.rs:109:5] konst = UnevaluatedConst { def: ConstId(ConstId(5c01)), args: [] }
```

For the file:

```rust
const FOO: usize = 100;

#[derive(Debug)]
#[rust_analyzer::color()]
#[repr(C, packed)]
struct Color {
    r: u8,
    g: u8,
    b: u8,
    a: u8,
}

const VALUE: Color = Color {
    r: 255,
    g: 0,
    b: 0,
    a: 255,
};

fn main() {
    // let _ = VALUE;

    println!("{:?}", VALUE);
}
```

It is then unclear to me how to then turn that `ConstKind::Unevaluated` into an evaluated constant.

Veykril:
I think that part is blocked on us actually fleshing out our const eval, it is currently in a fairly broken/incomplete state

PR #20144, opened 2025-07-02
Add `fn load_workspace_into_db` variant for `ra_ap_load_cargo`'s `fn load_workspace`

regexident:
See code comment for motivation:

```rust
pub fn load_workspace(
    ws: ProjectWorkspace,
    extra_env: &FxHashMap<String, Option<String>>,
    load_config: &LoadCargoConfig,
) -> anyhow::Result<(RootDatabase, vfs::Vfs, Option<ProcMacroClient>)> {
    let lru_cap = std::env::var("RA_LRU_CAP").ok().and_then(|it| it.parse::<u16>().ok());
    let mut db = RootDatabase::new(lru_cap);

    let (vfs, proc_macro_server) = load_workspace_into_db(ws, extra_env, load_config, &mut db)?;

    Ok((db, vfs, proc_macro_server))
}

// This variant of `load_workspace` allows deferring the loading of rust-analyzer
// into an existing database, which is useful in certain third-party scenarios,
// now that `salsa` supports extending foreign databases (e.g. `RootDatabase`).
pub fn load_workspace_into_db(
    ws: ProjectWorkspace,
    extra_env: &FxHashMap<String, Option<String>>,
    load_config: &LoadCargoConfig,
    db: &mut RootDatabase,
) -> anyhow::Result<(vfs::Vfs, Option<ProcMacroClient>)> {
    // ...
}
```

Specific use case motivating this change: `ra_ap_project_model`'s flavor of metadata is not as rich as `cargo_metadata`'s `Metadata`, so a third-party may want to extend `RootDatabase` with the metadata from `cargo_metadata`, but without having to first load the entire project (i.e. its entire source code and all). Instead it would be preferable to be able to defer loading via `ra_ap_load_cargo` until _after_ `cargo_metadata` has been loaded (and most importantly: already been made use of).

Imagine a tool that first performs package/target/crate discovery via `cargo_metadata`, then shows an overview of the project at hand, allowing for selecting individual items of interest and only upon confirmation does the actual heavy lifting of loading the project workspace via `ra_ap_load_cargo`. The current API would require one to create a temporary database for loading the `cargo_metadata` stuff to do the discovery and then discard/replace that database with `RootDatabase`, which then requires the `cargo_metadata` loading to be performed a second time on that new db.

### DIFF
--- a/crates/load-cargo/src/lib.rs
+++ b/crates/load-cargo/src/lib.rs
@@ -69,6 +69,23 @@ pub fn load_workspace(
     extra_env: &FxHashMap<String, Option<String>>,
     load_config: &LoadCargoConfig,
 ) -> anyhow::Result<(RootDatabase, vfs::Vfs, Option<ProcMacroClient>)> {
+    let lru_cap = std::env::var("RA_LRU_CAP").ok().and_then(|it| it.parse::<u16>().ok());
+    let mut db = RootDatabase::new(lru_cap);
+
+    let (vfs, proc_macro_server) = load_workspace_into_db(ws, extra_env, load_config, &mut db)?;
+
+    Ok((db, vfs, proc_macro_server))
+}
+
+// This variant of `load_workspace` allows deferring the loading of rust-analyzer
+// into an existing database, which is useful in certain third-party scenarios,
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
Seems reasonable to me.

PR #20154, opened 2025-07-03
Improve settings tree title and descriptions

joshka:
- All settings are now phrased in the imperative form stating what the
  setting does rather than talking about what it controls. (E.g.:
  "Show `Debug` action." instead of "Whether to show `Debug` action"
- Categories are now displayed in title case
- Categories are now sorted lexicographically
- General category is removed (and all the settings are moved to the top
  level)
- Language for a few descriptions is made a bit less ambiguous

Before:
image
After:
image

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -61,233 +61,312 @@ pub enum MaxSubstitutionLength {
     Limit(usize),
 }
 
-// Defines the server-side configuration of the rust-analyzer. We generate
-// *parts* of VS Code's `package.json` config from this. Run `cargo test` to
-// re-generate that file.
+// Defines the server-side configuration of the rust-analyzer. We generate *parts* of VS Code's
+// `package.json` config from this. Run `cargo test` to re-generate that file.
 //
-// However, editor specific config, which the server doesn't know about, should
-// be specified directly in `package.json`.
+// However, editor specific config, which the server doesn't know about, should be specified
+// directly in `package.json`.
 //
-// To deprecate an option by replacing it with another name use `new_name` | `old_name` so that we keep
-// parsing the old name.
+// To deprecate an option by replacing it with another name use `new_name` | `old_name` so that we
+// keep parsing the old name.
 config_data! {
-    /// Configs that apply on a workspace-wide scope. There are 2 levels on which a global configuration can be configured
+    /// Configs that apply on a workspace-wide scope. There are 2 levels on which a global
+    /// configuration can be configured
     ///
-    /// 1. `rust-analyzer.toml` file under user's config directory (e.g ~/.config/rust-analyzer/rust-analyzer.toml)
+    /// 1. `rust-analyzer.toml` file under user's config directory (e.g
+    ///    ~/.config/rust-analyzer/rust-analyzer.toml)
     /// 2. Client's own configurations (e.g `settings.json` on VS Code)
     ///
-    /// A config is searched for by traversing a "config tree" in a bottom up fashion. It is chosen by the nearest first principle.
+    /// A config is searched for by traversing a "config tree" in a bottom up fashion. It is chosen
+    /// by the nearest first principle.
     global: struct GlobalDefaultConfigData <- GlobalConfigInput -> {
         /// Warm up caches on project load.
         cachePriming_enable: bool = true,
-        /// How many worker threads to handle priming caches. The default `0` means to pick automatically.
+
+        /// How many worker threads to handle priming caches. The default `0` means to pick
+        /// automatically.
         cachePriming_numThreads: NumThreads = NumThreads::Physical,
 
         /// Custom completion snippets.
-        completion_snippets_custom: FxIndexMap<String, SnippetDef> = Config::completion_snippets_default(),
+        completion_snippets_custom: FxIndexMap<String, SnippetDef> =
+            Config::completion_snippets_default(),
 
-
-        /// These paths (file/directories) will be ignored by rust-analyzer. They are
-        /// relative to the workspace root, and globs are not supported. You may
-        /// also need to add the folders to Code's `files.watcherExclude`.
+        /// List of files to ignore
+        ///
+        /// These paths (file/directories) will be ignored by rust-analyzer. They are relative to
+        /// the workspace root, and globs are not supported. You may also need to add the folders to
+        /// Code's `files.watcherExclude`.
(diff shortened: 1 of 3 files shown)


docs/book/src/configuration_generated.md
joshka, line 1: No manual changes made to this doc - purely generated by running cargo test

editors/code/package.json
```diff
@@ -354,35 +354,122 @@
...
+                "title": "Rust Analyzer"
+            },
+            {
+                "title": "Assist"
+            },
+            {
+                "title": "Cache Priming"
+            },
+            {
+                "title": "Cargo"
```
joshka, line 366: This part forces the categories to be ordered lexicographically. Otherwise they are ordered based the order which they appear in the file. It's possible that this was intentional, in which case I'd remove these lines.

editors/code/package.json
```diff
@@ -354,35 +354,122 @@
...
+                "title": "Typing"
+            },
+            {
+                "title": "Vfs"
+            },
+            {
+                "title": "Workspace"
+            },
+            {
+                "title": "rust-analyzer",
```
joshka, line 444: Change this category title from "general" to "rust-analyzer" makes this appear at the top level of the settings instead of having an unnecessary category.

editors/code/package.json
```diff
@@ -354,35 +354,122 @@
...
                 "properties": {
                     "rust-analyzer.restartServerOnConfigChange": {
-                        "markdownDescription": "Whether to restart the server automatically when certain settings that require a restart are changed.",
+                        "description": "Restart the server automatically when settings that require a restart are changed.",
```
joshka, line 447: markdownDescription is only needed when there's markdown.
I left it there for the generated descriptions to avoid having to detect whether there's markdown or not

editors/code/package.json
```diff
@@ -653,17 +740,17 @@
                 "title": "$generated-start"
             },
             {
-                "title": "assist",
+                "title": "Assist",
```
joshka, line 743: Below this point, all the changes are generated from config.rs. There are no manual changes to this file below here. The changes are generally:
- category name is title case
- description is imperative

crates/rust-analyzer/src/config.rs
```diff
@@ -61,233 +61,312 @@ pub enum MaxSubstitutionLength {
     Limit(usize),
 }
 
-// Defines the server-side configuration of the rust-analyzer. We generate
-// *parts* of VS Code's `package.json` config from this. Run `cargo test` to
-// re-generate that file.
+// Defines the server-side configuration of the rust-analyzer. We generate *parts* of VS Code's
+// `package.json` config from this. Run `cargo test` to re-generate that file.
```
joshka, line 65: In general, Ive wrapped these at 100 chars as the changes were otherwise difficult to read at times with various wrap points across each item. I've also put a single blank line between each item instead of 0-3, this makes it easier to read as the comment is more obviously grouped with the following item rather than the previous.

I also removed the vertically aligned defaults (` = true` etc.), as this formatting doesn't appear elsewhere in the source code and doesn't make the values easier to read.

crates/rust-analyzer/src/config.rs
```diff
@@ -61,233 +61,312 @@ pub enum MaxSubstitutionLength {
...
-// However, editor specific config, which the server doesn't know about, should
-// be specified directly in `package.json`.
+// However, editor specific config, which the server doesn't know about, should be specified
+// directly in `package.json`.
 //
-// To deprecate an option by replacing it with another name use `new_name` | `old_name` so that we keep
-// parsing the old name.
+// To deprecate an option by replacing it with another name use `new_name` | `old_name` so that we
+// keep parsing the old name.
 config_data! {
```
joshka, line 72: The changes in this are to:
- comments - wording etc.
- spacing - consistent wrapping etc.

The generated code doesn't care too much about single newlines, only double newlines like standard markdown.

crates/rust-analyzer/src/config.rs
```diff
@@ -3203,8 +3340,10 @@ fn schema(fields: &[SchemaField]) -> serde_json::Value {
         .iter()
         .map(|(field, ty, doc, default)| {
             let name = field.replace('_', ".");
-            let category =
-                name.find('.').map(|end| String::from(&name[..end])).unwrap_or("general".into());
+            let category = name
+                .split_once(".")
+                .map(|(category, _name)| to_title_case(category))
+                .unwrap_or("rust-analyzer".into());
```
joshka, line 3346: Defaut changed to put all non-grouped settings at the top level instead of under a "General" category. (Casing is important here, it has to match the case of the extension display name.)

crates/rust-analyzer/src/config.rs
```diff
@@ -3203,8 +3340,10 @@ fn schema(fields: &[SchemaField]) -> serde_json::Value {
         .iter()
         .map(|(field, ty, doc, default)| {
             let name = field.replace('_', ".");
-            let category =
-                name.find('.').map(|end| String::from(&name[..end])).unwrap_or("general".into());
+            let category = name
+                .split_once(".")
+                .map(|(category, _name)| to_title_case(category))
```
joshka, line 3345: Generated category names are now title case

crates/rust-analyzer/src/config.rs
```diff
@@ -3218,6 +3357,29 @@ fn schema(fields: &[SchemaField]) -> serde_json::Value {
...
 
+/// Translate a field name to a title case string suitable for use in the category names on the
+/// vscode settings page.
+///
+/// First letter of word should be uppercase, if an uppercase letter is encountered, add a space
+/// before it e.g. "fooBar" -> "Foo Bar", "fooBarBaz" -> "Foo Bar Baz", "foo" -> "Foo"
+///
+/// This likely should be in stdx (or just use heck instead), but it doesn't handle any edge cases
+/// and is intentionally simple.
+fn to_title_case(s: &str) -> String {
```
joshka, line 3368: I'm inclined to leave this method here rather than extracting it to something more general. If someone wants to make it more useful, performant etc., then that's fine, but for now it's usefullness is this one use case.

I also don't really care too much about the impl here as it's dev time code. It's likely there's edge cases a more general function would have to catch.

joshka (commented):
Pre-review comments to help navigate this change.

For any comments about the description changes, please make these in config.rs rather than the generated files to make it easier to address (except for the non-generated parts of package.json).

Veykril (approved):
Very nice thanks!

PR #20175, opened 2025-07-05
`hir_ty::match_check` cleanup: remove special handling for box patterns

dianne:
As of rust-lang/rust#143414, as part of transitioning away from box patterns and towards more general deref patterns (rust-lang/rust#87121), box patterns are now represented in terms of deref patterns in the THIR and in match checking. 

It doesn't seem like box patterns were lowered to the `match_check` IR to begin with though, so the handling of box patterns in `match_check` was dead code. Since it doesn't match with the compiler anymore and it wasn't used, I've removed it in this PR without replacement.

Once rust-lang/rust#143472 (or some successor thereof) is merged, it should be possible to properly implement `match_check` for deref patterns in rust-analyzer as well. I'd be happy to try doing so once syntax for deref patterns is sorted out.

### DIFF
--- a/crates/hir-ty/src/diagnostics/match_check.rs
+++ b/crates/hir-ty/src/diagnostics/match_check.rs
@@ -25,7 +25,6 @@ use crate::{
     db::HirDatabase,
     display::{HirDisplay, HirDisplayError, HirFormatter},
     infer::BindingMode,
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (approved):
Thanks!

PR #20178, opened 2025-07-05
chore: Cleanup cargo config queries

ShoyuVanilla:
Resolves rust-lang/rust-analyzer#20081 and fixes rust-lang/rust-analyzer#20133  also fixes the case in [this comment](https://github.com/rust-lang/rust-analyzer/issues/20070#issuecomment-3038564999)

While looking into `cargo` codes, I found that `cargo metadata --no-deps` does not touch lockfiles:

So, we are safe to use `cargo metadata --no-deps` to query `target-dir`

### DIFF
--- a/crates/project-model/src/workspace.rs
+++ b/crates/project-model/src/workspace.rs
@@ -26,7 +26,7 @@ use crate::{
     WorkspaceBuildScripts,
     build_dependencies::BuildScriptOutput,
     cargo_config_file,
-    cargo_workspace::{CargoMetadataConfig, DepKind, PackageData, RustLibSource},
+    cargo_workspace::{CargoMetadataConfig, DepKind, FetchMetadata, PackageData, RustLibSource},
     env::{cargo_config_env, inject_cargo_env, inject_cargo_package_env, inject_rustc_tool_env},
     project_json::{Crate, CrateArrayIdx},
     sysroot::RustLibSrcWorkspace,
@@ -282,10 +282,24 @@ impl ProjectWorkspace {
             .ok()
             .flatten();
 
+        let fetch_metadata = FetchMetadata::new(
+            cargo_toml,
+            workspace_dir,
+            &CargoMetadataConfig {
+                features: features.clone(),
+                targets: targets.clone(),
+                extra_args: extra_args.clone(),
+                extra_env: extra_env.clone(),
+                toolchain_version: toolchain.clone(),
+                kind: "workspace",
+            },
+            &sysroot,
+            *no_deps,
+        );
         let target_dir = config
             .target_dir
             .clone()
-            .or_else(|| cargo_target_dir(cargo_toml, extra_env, &sysroot))
+            .or_else(|| fetch_metadata.no_deps_metadata().map(|m| m.target_directory.clone()))
             .unwrap_or_else(|| workspace_dir.join("target").into());
 
         // We spawn a bunch of processes to query various information about the workspace's
@@ -319,7 +333,7 @@ impl ProjectWorkspace {
                 };
                 rustc_dir.and_then(|rustc_dir| {
                     info!(workspace = %cargo_toml, rustc_dir = %rustc_dir, "Using rustc source");
-                    match CargoWorkspace::fetch_metadata(
+                    match FetchMetadata::new(
                         &rustc_dir,
                         workspace_dir,
                         &CargoMetadataConfig {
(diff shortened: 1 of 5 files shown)


crates/project-model/src/workspace.rs
```diff
@@ -270,7 +268,9 @@ impl ProjectWorkspace {
 
         tracing::info!(workspace = %cargo_toml, src_root = ?sysroot.rust_lib_src_root(), root = ?sysroot.root(), "Using sysroot");
         progress("querying project metadata".to_owned());
-        let toolchain_config = QueryConfig::Cargo(&sysroot, cargo_toml);
+        let config_file = cargo_config_file::read(cargo_toml, extra_env, &sysroot);
+        let config_file_ = config_file.clone();
+        let toolchain_config = QueryConfig::Cargo(&sysroot, cargo_toml, &config_file_);
```
Veykril, line 273:
```suggestion
        let toolchain_config = QueryConfig::Cargo(&sysroot, cargo_toml, &config_file.clone());
```
does this not work? I'd assume this constructs a temporary

crates/project-model/src/workspace.rs
```diff
@@ -1901,3 +1899,23 @@ fn sysroot_metadata_config(
...
+    // `--no-deps` doesn't (over)write lockfiles as it doesn't do any package resolve.
+    // So we can use it to get `target_directory` before copying lockfiles
+    let mut other_options = vec!["--no-deps".to_owned()];
+    if manifest.is_rust_manifest() {
+        meta.env("RUSTC_BOOTSTRAP", "1");
+        other_options.push("-Zscript".to_owned());
+    }
+    meta.other_options(other_options);
+    meta.exec().map(|m| m.target_directory).ok()
+}
```
Veykril, line 1915: We can be even smarter here. For one this code path will be exercised in the majority of cases. And note how we fallback to `--no-deps` when the command otherwise fails to semi support offline modes and the like. So I think it would make sense to just run this command proper (as we would usually) with `--no-deps`. If that succeeds, then run it again without while using the target dir info and stuff, and if this then fails, fallback to using the `--no-deps` version (this way we save one metadata invocation).

crates/project-model/src/workspace.rs
ShoyuVanilla: Unfortunately, since this is not a function call but a tuple ctor, that doesn't work 😅 
```rust
QueryConfig::Cargo {
    0: &sysroot,
    1: cargo_toml,
    2: &config_file.clone(),
}
```
might work but I think it's not desirable

crates/project-model/src/workspace.rs
```diff
@@ -476,9 +469,7 @@ impl ProjectWorkspace {
         let target_dir = config
             .target_dir
             .clone()
-            .or_else(|| {
-                cargo_config_build_target_dir(project_json.manifest()?, &config.extra_env, &sysroot)
-            })
+            .or_else(|| cargo_target_dir(project_json.manifest()?, &config.extra_env, &sysroot))
```
ShoyuVanilla, line 472: I've done decreasing number of metadata fetches via the way you suggested but this still remains because we don't do any other metadata fetch in this function.

crates/project-model/src/workspace.rs
```diff
@@ -563,51 +556,52 @@ impl ProjectWorkspace {
         let target_dir = config
             .target_dir
             .clone()
-            .or_else(|| cargo_config_build_target_dir(detached_file, &config.extra_env, &sysroot))
+            .or_else(|| cargo_target_dir(detached_file, &config.extra_env, &sysroot))
```
ShoyuVanilla, line 559: This still remains though we have couple of metadata fetches inside this function but I think this would be more safer, because... - I'll continue on the comment below

crates/project-model/src/workspace.rs
```diff
@@ -563,51 +556,52 @@ impl ProjectWorkspace {
...
             dir,
+            &target_dir,
             &|_| (),
         );
         if let Some(loaded_sysroot) = loaded_sysroot {
             sysroot.set_workspace(loaded_sysroot);
         }
 
-        let cargo_script = CargoWorkspace::fetch_metadata(
+        let fetch_metadata = FetchMetadata::new(
```
ShoyuVanilla, line 577: continue from the above - ... This fetches with the `sysroot` that might mutated from the above `sysroot.load_workspace(..)` call with `target_dir`. So, hositing this to the above lines to pre-fetch metadata for `target_dir` and reusing it might be incorrect due to the possible mutation of `sysroot`

PR #20187, opened 2025-07-06
Add richer LSP server example (#20017)

shashforge:
## Summary

Adds a new “head-start” LSP server example inside `crates/lsp-server/examples/`. This fresh example shows how to:

* Log document edits (`textDocument/didChange`)
* Provide a stubbed completion handler (returns an empty list)
* Provide a stubbed formatting handler (returns no edits)

## Details

1. **Setup**

   * Launches JSON-RPC over stdio
   * Advertises full-text sync, completion, and formatting capabilities

2. **Message Loop**

   * Logs each `didChange` URI
   * Responds to completion with `{"isIncomplete":false,"items":[]}`
   * Responds to formatting with `[]`
   * Cleans up on shutdown/exit

This minimal template makes it easy to embed rust-analyzer as an LSP server with basic completion and formatting hooks.

---

## Testing & Validation
1. **Build**  
   ```bash
   cargo build -p rust-analyzer --example rust_analyzer_lsp_example

### DIFF
--- a/crates/hir-ty/src/display.rs
+++ b/crates/hir-ty/src/display.rs
@@ -795,6 +795,14 @@ fn render_const_scalar(
                 let Some(bytes) = memory_map.get(addr, size_one * count) else {
                     return f.write_str("<ref-data-not-available>");
                 };
+                let expected_len = count * size_one;
+                if bytes.len() < expected_len {
+                    never!(
+                        "Memory map size is too small. Expected {expected_len}, got {}",
+                        bytes.len(),
+                    );
+                    return f.write_str("<layout-error>");
+                }
                 f.write_str("&[")?;
                 let mut first = true;
                 for i in 0..count {
--- a/crates/hir-ty/src/mir/eval.rs
+++ b/crates/hir-ty/src/mir/eval.rs
@@ -31,8 +31,8 @@ use syntax::{SyntaxNodePtr, TextRange};
 use triomphe::Arc;
 
 use crate::{
-    CallableDefId, ClosureId, ComplexMemoryMap, Const, ConstData, ConstScalar, FnDefId, Interner,
-    MemoryMap, Substitution, ToChalk, TraitEnvironment, Ty, TyBuilder, TyExt, TyKind,
+    AliasTy, CallableDefId, ClosureId, ComplexMemoryMap, Const, ConstData, ConstScalar, FnDefId,
+    Interner, MemoryMap, Substitution, ToChalk, TraitEnvironment, Ty, TyBuilder, TyExt, TyKind,
     consteval::{ConstEvalError, intern_const_scalar, try_const_usize},
     db::{HirDatabase, InternedClosure},
     display::{ClosureStyle, DisplayTarget, HirDisplay},
@@ -2195,7 +2195,7 @@ impl Evaluator<'_> {
                         }
                     }
                 }
-                chalk_ir::TyKind::Array(inner, len) => {
+                TyKind::Array(inner, len) => {
                     let len = match try_const_usize(this.db, len) {
                         Some(it) => it as usize,
                         None => not_supported!("non evaluatable array len in patching addresses"),
@@ -2213,7 +2213,7 @@ impl Evaluator<'_> {
                         )?;
                     }
                 }
-                chalk_ir::TyKind::Tuple(_, subst) => {
+                TyKind::Tuple(_, subst) => {
                     let layout = this.layout(ty)?;
                     for (id, ty) in subst.iter(Interner).enumerate() {
                         let ty = ty.assert_ty_ref(Interner); // Tuple only has type argument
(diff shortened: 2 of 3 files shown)


crates/rust-analyzer/examples/rust_analyzer_lsp_example/main.rs
ShoyuVanilla, line 1: I think this example should be located under `lib/lsp-server/examples/` like the previous [one](https://github.com/rust-lang/rust-analyzer/blob/9a1fc3cdb8bff55c3094c4d81c75ad2e57aa69a1/lib/lsp-server/examples/goto_def.rs)

ShoyuVanilla:
Well, I think that the original intention of the OP in #20017 is asking opinions about adding their existing `lsp-server` example codes to rust-analyzer repository, not implementing new examples, though I don't have strong opinion on it.

> ## What the example does
> ...
> This gives anyone embedding rust-analyzer a minimal “plug-in” template for completions and formatting.

Besides that, I think this example is not that "richer" than existing `goto_def.rs` one because most of the scaffolding codes are almost same, and I can't really understand what is the intention of the message handling examples in this PR: It's just a personal thought but I think it would be better to be somewhat more exhaustive on different kinds of messages or more focused or detailed on a specific category of messages.
In short, I'd like to ask *"What would be the proper filename for this when located in the same directory with `goto_def.rs` example?"*

lnicola:
Yeah, I think a single example would be better.

crates/rust-analyzer/examples/rust_analyzer_lsp_example/main.rs
shashforge: Thanks for catching that! You’re absolutely right—I’ve now moved the example into lib/lsp-server/examples/ alongside goto_def.rs. The file is renamed to completion_and_formatting.rs, and I’ve updated the crates/lsp-server/Cargo.toml manifest accordingly. Let me know if you spot anything else!

shashforge:
> Well, I think that the original intention of the OP in #20017 is asking opinions about adding their existing `lsp-server` example codes to rust-analyzer repository, not implementing new examples, though I don't have strong opinion on it.
> 
> > ## What the example does
> > ...
> > This gives anyone embedding rust-analyzer a minimal “plug-in” template for completions and formatting.
> 
> Besides that, I think this example is not that "richer" than existing `goto_def.rs` one because most of the scaffolding codes are almost same, and I can't really understand what is the intention of the message handling examples in this PR: It's just a personal thought but I think it would be better to be somewhat more exhaustive on different kinds of messages or more focused or detailed on a specific category of messages. In short, I'd like to ask _"What would be the proper filename for this when located in the same directory with `goto_def.rs` example?"_

Thank you so much for raising these points—I really appreciate your insight! 🙏
You’re absolutely right that a richer sample could handle additional messages (hover, semantic tokens, etc.), and I think that would make a great follow-up PR once this one is settled.

As for the filename, would you be comfortable with completion_and_formatting.rs? It follows our snake_case style and clearly indicates the two features it covers, sitting neatly beside goto_def.rs. If you’d prefer a different name or any other tweaks, I’m more than happy to update—just let me know what you’d like!

shashforge:
> Yeah, I think a single example would be better.

Thank you again for all your feedback! 🙏

At this point, would you prefer that I close this PR, or is there anything else you’d like me to adjust before we merge it? Please let me know what you’d recommend, and I’ll act accordingly.

ShoyuVanilla:
@lnicola and I pondered upon this and we think in current status, this would be best to be merged into existing `goto_def.rs` as they share much scaffolding codes in common - of course, the filename `goto_def.rs` should be adjusted`.

I'll comment on some details though if you take the "merging" suggestion most of them wouldn't be needed 😅

lib/lsp-server/Cargo.toml
```diff
@@ -19,3 +19,7 @@ ctrlc = "3.4.7"
 
 [lints]
 workspace = true
+
+[[example]]
+name = "completion_and_formatting"
+path = "examples/completion_and_formatting.rs"
```
ShoyuVanilla: We don't need this because the new example code is located inside `examples/`

lib/lsp-server/examples/completion_and_formatting.rs
```diff
@@ -0,0 +1,69 @@
+//! rust-analyzer LSP example
+//!
+//! A minimal Language Server Protocol server demonstrating how to embed
+//! `rust-analyzer` as a library. To run it:
```
ShoyuVanilla: "embed `rust-analyzer`" is not a right term here. This `lsp-server` is just a crate that rust-analyzer dogfoods on to handle lsp requests and responses.
So, the examples are not for rust-analyer or embedding. They just show how to use this "library"

shashforge:
> @lnicola and I pondered upon this and we think in current status, this would be best to be merged into existing `goto_def.rs` as they share much scaffolding codes in common - of course, the filename `goto_def.rs` should be adjusted`.
> 
> I'll comment on some details though if you take the "merging" suggestion most of them wouldn't be needed 😅

Thanks for the prompt feedback—will merge the examples into the single renamed file and push the update shortly.

PR #20198, opened 2025-07-08
Migrate `pull_assignment_up` assist to use`SyntaxEditor`

Hmikihiro:

### DIFF
--- a/crates/ide-assists/src/handlers/pull_assignment_up.rs
+++ b/crates/ide-assists/src/handlers/pull_assignment_up.rs
@@ -1,7 +1,7 @@
 use syntax::{
-    AstNode,
-    ast::{self, make},
-    ted,
+    AstNode, TextRange,
+    algo::find_node_at_range,
+    ast::{self, syntax_factory::SyntaxFactory},
 };
 
 use crate::{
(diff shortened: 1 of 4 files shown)


crates/ide-assists/src/handlers/pull_assignment_up.rs
```diff
@@ -66,33 +66,37 @@ pub(crate) fn pull_assignment_up(acc: &mut Assists, ctx: &AssistContext<'_>) ->
...
+                new_target_end -= diff;
+                editor.replace(stmt, rhs.syntax());
             }
-            let assign_expr = make::expr_assignment(collector.common_lhs, tgt.clone());
-            let assign_stmt = make::expr_stmt(assign_expr);
-
-            ted::replace(tgt.syntax(), assign_stmt.syntax().clone_for_update());
+            let new_tgt_root = editor.finish().new_root().clone();
+            let new_target_range = TextRange::new(target.start(), new_target_end);
+            let new_tgt = find_node_at_range::<ast::Expr>(&new_tgt_root, new_target_range).unwrap();
```
ShoyuVanilla: I think we can simplify this with `find_node_at_offset` with the offset `target.start()` - because we trigger this assist only on `if` and `match` exprs and due to those tokens, there won't be any other child expr with the same offsets - and then we won't need `new_target_range` calculation

ShoyuVanilla: And we can avoid this `unwrap()` if we "pull up" this first half of twin-editor things outside of the closure - we can use `SyntaxEditor::new` without `edit` - though I don't think that's quite necessary 😅

Hmikihiro: > And we can avoid this unwrap() if we "pull up" this first half of twin-editor things outside of the closure - we 
> can use SyntaxEditor::new without edit - though I don't think that's quite necessary 😅

https://github.com/Hmikihiro/rust-analyzer/pull/4
I tried refactoring as suggested, though I’m not sure if the result is much cleaner overall.

ShoyuVanilla: I suggested avoiding `unwrap()` because unwrap makes people a little bit nervous, not because it would be much cleaner 😅 Anyway, I think both are good and it's up to your choice 👍

ShoyuVanilla (approved):
Looks good to me 👍

crates/ide-assists/src/handlers/pull_assignment_up.rs
Hmikihiro: Thanks. I choice to avoid `unwrap()`

ShoyuVanilla (approved):

ShoyuVanilla:
Oh, I have merge permission now

PR #20218, opened 2025-07-10
Migrate `convert_match_to_let_else` assist to use `SyntaxEditor`

Hmikihiro:

### DIFF
--- a/crates/ide-assists/src/handlers/convert_match_to_let_else.rs
+++ b/crates/ide-assists/src/handlers/convert_match_to_let_else.rs
@@ -1,8 +1,8 @@
 use ide_db::defs::{Definition, NameRefClass};
 use syntax::{
     AstNode, SyntaxNode,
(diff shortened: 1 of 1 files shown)


ShoyuVanilla (approved):
Thanks!

ShoyuVanilla:
Oh, I missed little details while reviewing this:

https://github.com/rust-lang/rust-analyzer/blob/cce300bb1731469ff85039cc67a1600964b077da/crates/ide-assists/src/handlers/convert_match_to_let_else.rs#L61-L68

We still have a `SourceChangeBuilder::replace` call remaining here and I think this should be migrated as well, and instead of `format!`, `SyntaxFactory` methods for building asts.
And Ideally, I think passing `&mut SyntaxEditor` as a parameter to `rename_variable` is more desirable if possible.

@Hmikihiro Could you handle this as a followup? Only if you have enough time and are willing to take it on 😅

PR #20223, opened 2025-07-10
Migrate `generate_impl` assist to use `SyntaxEditor`

Hmikihiro:

### DIFF
--- a/crates/ide-assists/src/handlers/generate_impl.rs
+++ b/crates/ide-assists/src/handlers/generate_impl.rs
@@ -1,14 +1,14 @@
 use syntax::{
     ast::{self, AstNode, HasName, edit_in_place::Indent, make},
-    ted,
+    syntax_editor::{Position, SyntaxEditor},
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/generate_impl.rs
```diff
@@ -51,14 +51,17 @@ pub(crate) fn generate_impl(acc: &mut Assists, ctx: &AssistContext<'_>) -> Optio
             // Generate the impl
             let impl_ = utils::generate_impl(&nominal);
```
ShoyuVanilla, line 52: I think there might be some amounts of mutable syntax tree usages in this function but let's leave this as a future work for now 😄

Hmikihiro: Got it, thanks! I checked and confirmed that generate_impl does use `ted` inside `ConstParam::remove_default`.

PR #20226, opened 2025-07-10
Migrate remaining SourceChangeBuilder::replace of `convert_match_to_let_else`

Hmikihiro:
follow up https://github.com/rust-lang/rust-analyzer/pull/20218
- Migrated the remaining calls to SourceChangeBuilder::replace to use the SyntaxEditor
- Replaced usage of format! for AST generation with SyntaxFactory methods

It seems that with the current replacement, indentation is not preserved, so it may be necessary to migrate IndentLevel::increase_indent first.


ShoyuVanilla:
> It seems that with the current replacement, indentation is not preserved, so it may be necessary to migrate IndentLevel::increase_indent first.

Oh, I'll take a look on it this weekend

PR #20233, opened 2025-07-12
examples: add `minimal_lsp.rs` and FIFO test script

shashforge:
Adds a single example, examples/minimal_lsp.rs, plus a tiny FIFO test script.
The sample now shows:
```
- completion textDocument/completion
- hover textDocument/hover
- goto-definition textDocument/definition
- full-doc formatting (rustfmt) textDocument/formatting
- dummy diagnostics textDocument/publishDiagnostics
- UTF-8 offsetEncoding in the initialize response
```

Quick test

# one-shot, all nine packets

> cargo build --example minimal_lsp
> target/debug/examples/minimal_lsp < examples/nine_packets.txt

Also includes `examples/manual_test.sh`, a tiny FIFO script that sends the canonical 9 LSP packets so the server can be verified from two terminals.

# Two-terminal test

mkfifo /tmp/lsp_pipe
 **terminal A**
cat /tmp/lsp_pipe | cargo run --example minimal_lsp
 **terminal B**
bash examples/manual_test.sh          # pumps the same nine packets

We  should see, in order: capabilities → diagnostics → completion → hover → definition → formatting → shutdown. 
(anyhow stays under [dev-dependencies]).

Fixes rust-lang/rust-analyzer#20017


shashi1687:
Hi @ShoyuVanilla ,

Thanks for the earlier feedback on #20187—it was really helpful.
Per your suggestion I’ve folded the old `goto_def.rs` logic and my completion / formatting demo into a single file, **`examples/minimal_lsp.rs`**. The new example now shows:

* completion
* hover
* goto-definition
* full-file formatting (rustfmt)
* a dummy diagnostic push
* UTF-8 `offsetEncoding` in the `initialize` response

I’ve also added a tiny FIFO script (`examples/manual_test.sh`) so the whole round-trip can be reproduced from two terminals in a few seconds.

Whenever you have a moment, could you please take another look?
Thanks again for the guidance!

shashforge:
Hi @ShoyuVanilla @lnicola ,

Thank you for the helpful guidance on my last attempt  on #20187
As you suggested, I’ve folded the logic from goto_def.rs and my earlier demo into a single file, examples/minimal_lsp.rs. 
The new example now covers

- completion
- hover
- goto-definition
- full-file formatting via rustfmt
- a dummy diagnostic push
- UTF-8 offsetEncoding in the initialize 

I’ve also added a tiny FIFO helper (examples/manual_test.sh) to replay the nine canonical packets, so the whole round-trip can be tested in two terminals.
Whenever you have a moment, I’d really appreciate another look.
Thanks again for your time and feedback!

— Shashi

ShoyuVanilla:
Thanks! I'll take a look before not too late 😅

ShoyuVanilla:
Overally looks good but some nits and a question:

lib/lsp-server/examples/minimal_lsp.rs
```diff
@@ -0,0 +1,340 @@
+// SPDX-License-Identifier: MIT
+//! Minimal Language‑Server‑Protocol example: **`minimal_lsp.rs`**
+//! =============================================================
+//!
```
ShoyuVanilla: I think this is a bit redundant. We have MIT (and Apache) license files on repository root and the filename is self-explanatory

lib/lsp-server/examples/minimal_lsp.rs
```diff
@@ -0,0 +1,340 @@
...
+//! This is a self‑contained learning sample that goes one step beyond
+//! `examples/goto_def.rs`.
```
ShoyuVanilla: As `examples/goto_def.rs` is deleted in this PR, people might be confused and I think this isn't necessary

lib/lsp-server/examples/manual_test.sh
ShoyuVanilla, line 1: I can understand the intention but does this work? This shell script uses named pipe for IPC but example lsp code uses stdin

shashforge: Thanks for checking!
The FIFO is just a convenience so I can drive the server from another terminal.  manual_test.sh writes each JSON packet to /tmp/lsp_pipe, and in the other window I run:
`cat /tmp/lsp_pipe | target/debug/examples/minimal_lsp`

cat reads the same pipe and streams the bytes straight into the server’s stdin, so the server sees exactly what it would with a direct pipe.
If you’d rather skip the FIFO altogether, the docs also show a one-liner that pipes all nine packets directly into minimal_lsp; both paths exercise the same code.

Hope that clarifies!


`manual_test.sh (writer)  ->  /tmp/lsp_pipe  ->  cat  ->  stdin of minimal_lsp`

lib/lsp-server/examples/minimal_lsp.rs
shashforge: Agree. removed

lib/lsp-server/examples/minimal_lsp.rs
shashforge: agree. removed

shashforge:
Hi @ShoyuVanilla 

Thanks so much for the careful review!

1. Header tidy-up – I removed the SPDX line and the banner underline.
2. goto_def.rs mention – dropped, since that file is gone now.
3. manual_test.sh – added a short note that the FIFO only feeds stdin, so the code path is the same as a direct pipe.

Ran cargo fmt; the example builds and clippy is clean.

If the extra script still feels like overkill, I’m happy to delete it and rely on the one-liner here-doc instead—just say the word.

Really appreciate your time and feedback. Let me know if there’s anything else you’d like tweaked!

lib/lsp-server/examples/manual_test.sh
ShoyuVanilla: Oh, I see. Could you explain that steps with the usage comments of the script?

Also, it might be helpful to include an example using background execution, like:
```bash
bash example/manual_test.sh &
cat /tmp/lsp_pipe | cargo run --example minimal_lsp # cargo forwards stdin to the example
```
This could be useful for people who don’t use a terminal multiplexer or multiple tabs.
That said, feel free to skip this background execution part if it feels too verbose or distracting.

shashforge: @ShoyuVanilla  Thanks for the suggestion! I’ve expanded the header of manual_test.sh to spell out each step and added the background-execution example you mentioned.
Let me know if you’d like any further tweaks!

lib/lsp-server/examples/minimal_lsp.rs
```diff
@@ -0,0 +1,336 @@
...
+//! | ← | `textDocument/publishDiagnostics` | pushes a dummy info diagnostic whenever the buffer changes |
+//! | ← | `textDocument/definition` | echoes an empty location array so the jump works |
+//! | ← | `textDocument/completion` | offers one hard‑coded item `HelloFromLSP` |
+//! | ← | `textDocument/hover` | shows *Hello from minimal_lsp* markdown |
+//! | ← | `textDocument/formatting` | pipes the doc through **rustfmt** and returns a full‑file edit |
+//!
+//! ### Quick start
+//! ```bash
+//! cd rust-analyzer/lib/lsp-server
+//! cargo add anyhow --dev         # once, only for examples
```
ShoyuVanilla: One last thing 😅 sorry for not discovering this earlier with other ones.
I think the above line is good to be removed, as you have added `anyhow` to our `dev-dependencies`?

shashforge: done. resolved

ShoyuVanilla (approved):
Thanks! Sorry for being too picky on this PR. As this is adding an example that serves as a documentation for newcomers, I had to be sensitive on accuracy 😅

shashforge:
> Thanks! Sorry for being too picky on this PR. As this is adding an example that serves as a documentation for newcomers, I had to be sensitive on accuracy 😅

Absolutely, and thanks again for the careful review — it only makes the example stronger.

If there are other bite-sized issues or examples that could use a fresh set of eyes, I’d be happy to tackle them next. I’m already browsing the issue tracker, but please feel free to point me toward anything you think would be especially helpful for the project. I’m keen to contribute wherever it adds the most value.

PR #20235, opened 2025-07-13
Fix assoc type where clause position

A4-Tacks:
**Code**:

```rust
trait Foo {
    type X<'a> where Self: 'a;
}
impl Foo for () {
    $0
}
```

**Current**:

```rust
trait Foo {
    type X<'a> where Self: 'a;
}
impl Foo for () {
    type X<'a> where Self: 'a = $0;
}
```

**This pr fixed**:

```rust
trait Foo {
    type X<'a> where Self: 'a;
}
impl Foo for () {
    type X<'a> = $0 where Self: 'a;
}
```


ShoyuVanilla (approved):
Thanks!

PR #20244, opened 2025-07-15
fix: Prevent lockfile overwrite loops on metadata fetch

### LINKED ISSUE #20189: VSCode stuck on fetching cargo metadata, even without dependencies
The rust-analyzer in VSCode runs into an infinite loop of fetching cargo metadata in the latest version. 
It happened after I updated to Rust 1.88 and at first blamed the macros of my own project, but managed to reproduce the error on a VSCode Profile with only the rust-analyzer extension installed, in a freshly initialized folder.
Edit: running `rustup override set 1.87`(or 1.80) does not fix it, so it's most likely not a version problem.

**rust-analyzer version**: `rust-analyzer version: 0.4.2525-standalone (2d518c73d9 2025-07-05) [/Users/USERNAME/.vscode/extensions/rust-lang.rust-analyzer-0.4.2525-darwin-arm64/server/rust-analyzer]`
(Pre-release has the same version and same problem)

**rustc version**: rustc 1.88.0 (6b00bc388 2025-06-23)

**editor or extension**: VSCode, latest version; rust-analyzer extension version: 0.3.2519

**relevant settings**: Only `$RUSTC_WRAPPER=/Path/to/sccache`.

**repository link (if public, optional)**: Default rust hello world; `cargo init`

```shell
mkdir test
cd test
cargo init
code .
(Then enter a profile with only the rust-analyzer)
```

I'm honestly not sure what I am doing wrong; I expect this to have to do something with something of my setup, else this should have happened to quite a few more people yet.
I did record the logs for this:

https://pastebin.com/tfTWELhU
(And yes, this also happens outside of the `~/Desktop` folder.)

It's making work on my Rust projects basically impossible because the CPU is caught up with fetching metadata and the results only show for a fraction of a second before the next fetch starts.

ShoyuVanilla:
Fixes #20189


ShoyuVanilla:
Oh, it's mandatory to have lockfile names as `Cargo.lock` 😅 (I tried to tread the easy way by renaming copied `Cargo.lock` into `Cargo_lock` to evade LSP documentChanges filters but unforutnately that didn't work)

ShoyuVanilla:
Some thoughts on this: [#t-compiler/rust-analyzer > Do we need copied lockfiles?](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Do.20we.20need.20copied.20lockfiles.3F/with/529092917)

PR #20247, opened 2025-07-16
Add tailexpr & -> &mut for generate_mut_trait_impl

A4-Tacks:
Implement simple `&` -> `&mut ` on tailexpr for `generate_mut_trait_impl`

https://github.com/rust-lang/rust-analyzer/issues/15581#issuecomment-1791859243

> But there should definitely be a rewrite rule for `&(...)[...]` to `&mut (...)[...]`, and perhaps other common cases. Just because the general problem is impossible doesn't mean the easy cases shouldn't be done; doing the easy cases is what assists are _for_.


ShoyuVanilla (approved):
Thanks!

PR #20255, opened 2025-07-19
Fix ide-assist: generate Deref transitive

A4-Tacks:
# Example

```rust
struct Foo<T>($0T);
```

Current output:

```rust
struct Foo<T>(T);

impl<T: std::ops::Deref> std::ops::Deref for Foo<T> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}
```

This PR fixed output:

```rust
struct Foo<T>(T);

impl<T> std::ops::Deref for Foo<T> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}
```

### DIFF
--- a/crates/ide-assists/src/utils.rs
+++ b/crates/ide-assists/src/utils.rs
@@ -567,6 +567,7 @@ pub(crate) fn generate_impl_text(adt: &ast::Adt, code: &str) -> String {
 ///
 /// This is useful for traits like `PartialEq`, since `impl<T> PartialEq for U<T>` often requires `T: PartialEq`.
 // FIXME: migrate remaining uses to `generate_trait_impl`
+#[allow(dead_code)]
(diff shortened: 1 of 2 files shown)


crates/ide-assists/src/utils.rs
```diff
@@ -567,6 +567,7 @@ pub(crate) fn generate_impl_text(adt: &ast::Adt, code: &str) -> String {
 ///
 /// This is useful for traits like `PartialEq`, since `impl<T> PartialEq for U<T>` often requires `T: PartialEq`.
 // FIXME: migrate remaining uses to `generate_trait_impl`
+#[allow(dead_code)]
```
lnicola, line 570: Do we need to keep the old function?

A4-Tacks: Will deleting this function affect other PRs?

lnicola: I don't think so, but ok.

PR #20269, opened 2025-07-21
Migrate AstNodeEdit::Indent to SyntaxEditor

Hmikihiro:
https://github.com/rust-lang/rust-analyzer/pull/20224#issuecomment-3061091923
make mapping Indented node with original node.


ShoyuVanilla (approved):
Thanks! Glad that now we have a indent/dedent functionality free of mutable syntax tree usages!

PR #20270, opened 2025-07-21
Migrate `generate new` assist to use `SyntaxEditor`

Hmikihiro:

### DIFF
--- a/crates/syntax/src/syntax_editor.rs
+++ b/crates/syntax/src/syntax_editor.rs
@@ -5,7 +5,7 @@
 //! [`SyntaxEditor`]: https://github.com/dotnet/roslyn/blob/43b0b05cc4f492fd5de00f6f6717409091df8daa/src/Workspaces/Core/Portable/Editing/SyntaxEditor.cs
 
 use std::{
-    fmt,
+    fmt, iter,
     num::NonZeroU32,
     ops::RangeInclusive,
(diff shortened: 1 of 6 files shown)


crates/syntax/src/syntax_editor.rs
```diff
@@ -41,6 +41,16 @@ impl SyntaxEditor {
         self.annotations.push((element.syntax_element(), annotation))
     }
 
+    pub fn add_annotation_all(
+        &mut self,
+        elements: Vec<impl Element>,
+        annotation: SyntaxAnnotation,
+    ) {
+        for element in elements {
+            self.annotations.push((element.syntax_element(), annotation.clone()));
```
lnicola: Nit: can you use `extend` here? It doesn't matter, just future-proofing for an eventual Clippy lint :-).

Hmikihiro: Thanks

Hmikihiro: Thanks. I fix like this.
```rs
    pub fn add_annotation_all(
        &mut self,
        elements: Vec<impl Element>,
        annotation: SyntaxAnnotation,
    ) {
        self.annotations
            .extend(elements.into_iter().map(|e| e.syntax_element()).zip(iter::repeat(annotation)));
    }
```

ShoyuVanilla (approved):
Thanks!

PR #20290, opened 2025-07-23
Use `TempDir` for copied lockfiles

ShoyuVanilla:
Fixes https://github.com/rust-lang/rust-analyzer/issues/20189 and closes https://github.com/rust-lang/rust-analyzer/pull/20244 and fixes https://github.com/rust-lang/rust-analyzer/issues/20113

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,6 +156,7 @@ smallvec = { version = "1.15.1", features = [
   "const_generics",
 ] }
 smol_str = "0.3.2"
+temp-dir = "0.1.16"
 text-size = "1.1.1"
 tracing = "0.1.41"
 tracing-tree = "0.4.0"
--- a/crates/project-model/src/cargo_workspace.rs
+++ b/crates/project-model/src/cargo_workspace.rs
@@ -552,6 +552,7 @@ impl CargoWorkspace {
 
 pub(crate) struct FetchMetadata {
     command: cargo_metadata::MetadataCommand,
+    manifest_path: ManifestPath,
     lockfile_path: Option<Utf8PathBuf>,
     kind: &'static str,
     no_deps: bool,
@@ -655,7 +656,15 @@ impl FetchMetadata {
         }
         .with_context(|| format!("Failed to run `{cargo_command:?}`"));
 
-        Self { command, lockfile_path, kind: config.kind, no_deps, no_deps_result, other_options }
+        Self {
+            manifest_path: cargo_toml.clone(),
+            command,
+            lockfile_path,
+            kind: config.kind,
+            no_deps,
+            no_deps_result,
+            other_options,
+        }
     }
 
     pub(crate) fn no_deps_metadata(&self) -> Option<&cargo_metadata::Metadata> {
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -156,6 +156,7 @@ smallvec = { version = "1.15.1", features = [
   "const_generics",
 ] }
 smol_str = "0.3.2"
+temp-dir = "0.1.16"
```
ShoyuVanilla, line 159: I chose this crate because it uses short, safe Rust codes and has no deps

crates/project-model/src/cargo_workspace.rs
```diff
@@ -672,18 +681,47 @@ impl FetchMetadata {
...
         let mut using_lockfile_copy = false;
+        let mut _temp_dir_guard = None;
         // The manifest is a rust file, so this means its a script manifest
         if let Some(lockfile) = lockfile_path {
-            let target_lockfile =
-                target_dir.join("rust-analyzer").join("metadata").join(kind).join("Cargo.lock");
+            _temp_dir_guard = temp_dir::TempDir::new().ok();
+            let target_lockfile = _temp_dir_guard
+                .and_then(|tmp| tmp.path().join("Cargo.lock").try_into().ok())
+                .unwrap_or_else(|| {
```
ShoyuVanilla, line 705: I think failures on creating `TempDir` or converting it to `Utf8PathBuf` would be quite rare, but copied disambiguation from #20244 into this fallback anyway

Cargo.toml
lnicola: TIL `temp-dir` (besides `tempdir`, which is very popular, and `tempfile`, which is unmaintained).

crates/project-model/src/cargo_workspace.rs
```diff
@@ -672,18 +681,47 @@ impl FetchMetadata {
...
             return no_deps_result.map(|m| (m, None));
         }
 
         let mut using_lockfile_copy = false;
+        let mut _temp_dir_guard = None;
         // The manifest is a rust file, so this means its a script manifest
         if let Some(lockfile) = lockfile_path {
-            let target_lockfile =
-                target_dir.join("rust-analyzer").join("metadata").join(kind).join("Cargo.lock");
+            _temp_dir_guard = temp_dir::TempDir::new().ok();
```
lnicola: Should we set a prefix here?

ShoyuVanilla: I think it's not necessary but having one would be good to some curious people who look into their `/tmp/`. Do you have some recommendations for prefix string?

lnicola: Something with `rust-analyzer` or `r-a`? :-)

Veykril: We do have one for the proc-macro server already I think on windows (where we copy the proc macro dylibs)

Veykril: Ah well, doesn't matter I guess as that is proc-macro specific https://github.com/rust-lang/rust-analyzer/blob/1bf93c38adb2dd364f7256b30113b06a9d4d1cf9/crates/proc-macro-srv/src/dylib.rs#L233

ShoyuVanilla: Oh, I just found that we also have a `RemoveFileOnDrop` for Windows there as well 😅

Veykril: Note as an aside but `RemoveFileOnDrop` shouldn't be a thing. Windows has file creation flags to mark them as to be deleted once all OS handles are dropped which we should be using instead I think.

Veykril (approved):

Cargo.toml
oxalica: > (besides `tempdir`, which is very popular, and `tempfile`, which is unmaintained).

I guess you wrote them reversed? Seems [`tempdir`](https://crates.io/crates/tempdir) is the unmaintained one, and [`tempfile`](https://crates.io/crates/tempfile) is the popular one.

PR #20293, opened 2025-07-23
Migrate `replace derive with manual impl` and `add_missing_impl_members` to use `SyntaxEditor`

Hmikihiro:
`fn add_trait_assoc_items_to_impl` use ted yet.

### DIFF
--- a/crates/ide-assists/src/utils.rs
+++ b/crates/ide-assists/src/utils.rs
@@ -23,7 +23,7 @@ use syntax::{
     ast::{
         self, HasArgList, HasAttrs, HasGenericParams, HasName, HasTypeBounds, Whitespace,
         edit::{AstNodeEdit, IndentLevel},
-        edit_in_place::{AttrsOwnerEdit, Indent, Removable},
+        edit_in_place::{AttrsOwnerEdit, Removable},
         make,
         syntax_factory::SyntaxFactory,
     },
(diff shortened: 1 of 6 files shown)


ShoyuVanilla:
Oh, the test added from https://github.com/rust-lang/rust-analyzer/pull/19938 are failing

crates/ide-assists/src/utils.rs
```diff
@@ -185,71 +185,65 @@ pub fn add_trait_assoc_items_to_impl(
     trait_: hir::Trait,
     impl_: &ast::Impl,
     target_scope: &hir::SemanticsScope<'_>,
-) -> ast::AssocItem {
+) -> Vec<ast::AssocItem> {
```
ShoyuVanilla, line 189: I think it would be good to have `#[must_use]` attributes here, as this is not mutating existing syntax tree anymore but returns a new one. (And this would help preventing tests in https://github.com/rust-lang/rust-analyzer/pull/19938 failing, though I've merged it after this PR was created 😅 )

Hmikihiro:
Sorry, I made a mistake merge.

crates/ide-assists/src/utils.rs
Hmikihiro: Thanks for finding the fix spot!

PR #20295, opened 2025-07-23
Add ide-assist: extract_to_default_generic

A4-Tacks:
Extracts selected type to default generic parameter.

```rust
struct Foo(u32, $0String$0);
```
->
```rust
struct Foo<T$0 = String>(u32, T);
```

---

```rust
fn foo(x: Option<$0i32$0>) {}
```
->
```rust
fn foo<T$0>(x: Option<T>) {}
```


Veykril:
This does not seem too useful to me, closing given we have too many assists already.

PR #20303, opened 2025-07-26
Migrate path transform

Hmikihiro:


ShoyuVanilla (approved):
Thanks! It would be ideal to replace remaining `.clone_for_update()` from the migrate codes in future, but still great improvement 👍

PR #20305, opened 2025-07-26
Migrate part of utils.rs to use SyntaxEditor

Hmikihiro:

### DIFF
--- a/crates/ide-assists/src/utils.rs
+++ b/crates/ide-assists/src/utils.rs
@@ -27,6 +27,7 @@ use syntax::{
         make,
         syntax_factory::SyntaxFactory,
     },
+    syntax_editor::SyntaxEditor,
     ted,
 };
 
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/utils.rs
```diff
@@ -329,7 +330,7 @@ fn invert_special_case(make: &SyntaxFactory, expr: &ast::Expr) -> Option<ast::Ex
 fn invert_special_case_legacy(expr: &ast::Expr) -> Option<ast::Expr> {
```
ShoyuVanilla, line 330: This one is meant to be replaced with this:

https://github.com/rust-lang/rust-analyzer/blob/b77072075c2340a5d7a3dc6c361639ee6ed0fcfd/crates/ide-assists/src/utils.rs#L266

but anyway, I think having this migrated as well is good.

PR #20311, opened 2025-07-27
Migrate `convert_tuple_struct_to_named_struct` assist to use `SyntaxEditor`

Hmikihiro:

### DIFF
--- a/crates/ide-assists/src/handlers/convert_tuple_struct_to_named_struct.rs
+++ b/crates/ide-assists/src/handlers/convert_tuple_struct_to_named_struct.rs
@@ -1,10 +1,10 @@
 use either::Either;
 use ide_db::defs::{Definition, NameRefClass};
 use syntax::{
-    SyntaxKind, SyntaxNode,
+    SyntaxKind, SyntaxNode, T,
     ast::{self, AstNode, HasAttrs, HasGenericParams, HasVisibility},
     match_ast,
-    syntax_editor::{Position, SyntaxEditor},
+    syntax_editor::{Element, Position, SyntaxEditor},
 };
 
 use crate::{AssistContext, AssistId, Assists, assist_context::SourceChangeBuilder};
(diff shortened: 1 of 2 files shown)


crates/ide-assists/src/handlers/convert_tuple_struct_to_named_struct.rs
```diff
@@ -71,66 +74,62 @@ pub(crate) fn convert_tuple_struct_to_named_struct(
...
             f.attrs().map(|attr| attr.syntax().clone_subtree().clone_for_update().into()).collect(),
         );
-        Some(field)
+        ast::RecordField::cast(field_editor.finish().new_root().clone())
     });
-    let record_fields = ast::make::record_field_list(record_fields);
-    let tuple_fields_text_range = tuple_fields.syntax().text_range();
-
-    edit.edit_file(ctx.vfs_file_id());
+    let record_fields = ast::make::record_field_list(record_fields).clone_for_update();
```
ShoyuVanilla: I think it would be good to use

https://github.com/rust-lang/rust-analyzer/blob/eb3d9095091cd364410c778d613d7180598839af/crates/syntax/src/ast/syntax_factory/constructors.rs#L963

especially when we need `.clone_for_update()` here. Though it still calls `clone_for_update()` in its body, at least we can gather them up in one place and bulk-fix later.

crates/ide-assists/src/handlers/convert_tuple_struct_to_named_struct.rs
```diff
@@ -245,7 +275,7 @@ fn generate_names(fields: impl Iterator<Item = ast::TupleField>) -> Vec<ast::Nam
         .enumerate()
         .map(|(i, _)| {
             let idx = i + 1;
-            ast::make::name(&format!("field{idx}"))
+            ast::make::name(&format!("field{idx}")).clone_for_update()
```
ShoyuVanilla: Same as the above, with
https://github.com/rust-lang/rust-analyzer/blob/eb3d9095091cd364410c778d613d7180598839af/crates/syntax/src/ast/syntax_factory/constructors.rs#L14
Everything else looks good to me!

PR #20334, opened 2025-07-29
Add ide-assist: generate_ref_impl

A4-Tacks:
Add a trait impl for reference.

```rust
trait $0Foo {
    fn foo(&mut self, n: i32) -> i32;
}
```
->
```rust
trait Foo {
    fn foo(&mut self, n: i32) -> i32;
}

impl<T: Foo + ?Sized> Foo for &mut T {
    fn foo(&mut self, n: i32) -> i32 {
        (**self).foo(n)
    }
}


Veykril:
I don't believe this pulls its weight.

A4-Tacks:
It is very troublesome to manually forward a large number of methods to references when writing libraries

PR #20363, opened 2025-08-02
Add ide-assist: extract_impl_items

A4-Tacks:
Extract selected impl items into new impl.

```rust
struct Foo;
impl Foo {
    fn foo() {}
    $0fn bar() {}
    fn baz() {}$0
}
```
->
```rust
struct Foo;
impl Foo {
    fn foo() {}
}

impl Foo {
    fn bar() {}

    fn baz() {}
}
```


A4-Tacks:

PR #20371, opened 2025-08-03
Migrate `generate_trait_from_impl` assist to use `SyntaxEditor`

Hmikihiro:


crates/syntax/src/ast/make.rs
```diff
@@ -1020,7 +1020,7 @@ pub fn trait_(
     ident: &str,
     gen_params: Option<ast::GenericParamList>,
     where_clause: Option<ast::WhereClause>,
-    assoc_items: ast::AssocItemList,
+    assoc_items: Option<ast::AssocItemList>,
```
Veykril: This being non-optional sounds correct to me as it would be an invalid trait definition otherwise (ignoring the trait alias feature for now). It should be fine to just unwrap the cast in the assist above, as should that fail that's a bug in `SyntaxEditor` (correct me if I am wrong, I am not too familiar with this API)

Hmikihiro: I see. That sounds correct.

PR #20373, opened 2025-08-03
In handlers/extract_module.rs, generate ast::Module instead of String

Hmikihiro:
When I migrate to use 'SyntaxEditor', It needs AstNode instead of String. 

bug fix:

https://github.com/user-attachments/assets/568ee0fc-8bed-4d74-828a-74241978ef5b

to:

https://github.com/user-attachments/assets/326e4f8d-2531-4927-b588-d70b02911034

### DIFF
--- a/crates/ide-assists/src/handlers/extract_module.rs
+++ b/crates/ide-assists/src/handlers/extract_module.rs
@@ -1,6 +1,5 @@
-use std::iter;
+use std::ops::RangeInclusive;
 
-use either::Either;
 use hir::{HasSource, ModuleSource};
 use ide_db::{
     FileId, FxHashMap, FxHashSet,
@@ -82,7 +81,15 @@ pub(crate) fn extract_module(acc: &mut Assists, ctx: &AssistContext<'_>) -> Opti
         curr_parent_module = ast::Module::cast(mod_syn_opt);
     }
 
-    let mut module = extract_target(&node, ctx.selection_trimmed())?;
+    let selection_range = ctx.selection_trimmed();
+    let (mut module, module_text_range) = if let Some(item) = ast::Item::cast(node.clone()) {
+        let module = extract_single_target(&item);
+        (module, node.text_range())
+    } else {
+        let (module, range) = extract_child_target(&node, selection_range)?;
+        let module_text_range = range.start().text_range().cover(range.end().text_range());
+        (module, module_text_range)
+    };
     if module.body_items.is_empty() {
         return None;
     }
@@ -92,7 +99,7 @@ pub(crate) fn extract_module(acc: &mut Assists, ctx: &AssistContext<'_>) -> Opti
     acc.add(
         AssistId::refactor_extract("extract_module"),
         "Extract Module",
-        module.text_range,
+        module_text_range,
         |builder| {
             //This takes place in three steps:
             //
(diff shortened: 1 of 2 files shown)


crates/ide-assists/src/handlers/extract_module.rs
```diff
@@ -175,31 +175,36 @@ fn generate_module_def(
...
+        let assoc_items = module
+            .body_items
+            .iter()
+            .filter_map(|item| match item {
+                ast::Item::Const(it) => Some(ast::AssocItem::Const(it.clone())),
+                ast::Item::Fn(it) => Some(ast::AssocItem::Fn(it.clone())),
+                ast::Item::MacroCall(it) => Some(ast::AssocItem::MacroCall(it.clone())),
+                ast::Item::TypeAlias(it) => Some(ast::AssocItem::TypeAlias(it.clone())),
+                _ => None,
+            })
```
DropDemBits:
```suggestion
        let body_items = std::mem::take(&mut module.body_items);
        let assoc_items = body_items
            .into_iter()
            .filter_map(ast::AssocItem::cast)
```

since it doesn't seem like we use `body_items` after this.

crates/ide-assists/src/handlers/extract_module.rs
```diff
@@ -120,7 +119,8 @@ pub(crate) fn extract_module(acc: &mut Assists, ctx: &AssistContext<'_>) -> Opti
             let import_paths_to_be_removed = module.resolve_imports(curr_parent_module, ctx);
             module.change_visibility(record_fields);
 
-            let module_def = generate_module_def(&impl_parent, &mut module, old_item_indent);
+            let module_def =
+                generate_module_def(&impl_parent, &mut module, old_item_indent).to_string();
```
DropDemBits: minor nit: Even better would be to have
```suggestion
                generate_module_def(&impl_parent, module, old_item_indent).to_string();
```

but then you'd need to disentangle the later uses `module.text_range` from the move.

DropDemBits (commented):
It's been a bit since I've done a review, so take these suggestions with a grain of salt 😅

crates/ide-assists/src/handlers/extract_module.rs
Hmikihiro: Thanks! I noticed that I can split text_range record from struct Module. for your advice.

crates/syntax/src/ast/make.rs
```diff
@@ -229,6 +229,24 @@ pub fn ty_fn_ptr<I: Iterator<Item = Param>>(
     }
 }
 
+pub fn item_list(body: Option<Vec<ast::Item>>) -> ast::ItemList {
+    let is_break_braces = body.is_some();
+    let body_newline = if is_break_braces { "\n".to_owned() } else { String::new() };
+    let body_indent = if is_break_braces { "    ".to_owned() } else { String::new() };
```
ShoyuVanilla: minor nit:
```suggestion
    let body_newline = if is_break_braces { "\n" } else { "" };
    let body_indent = if is_break_braces { "    " } else { "" };
```

crates/syntax/src/ast/make.rs
```diff
@@ -229,6 +229,24 @@ pub fn ty_fn_ptr<I: Iterator<Item = Param>>(
...
+
+    let body = match body {
+        Some(bd) => bd.iter().map(|elem| elem.to_string()).join("\n\n    "),
+        None => String::new(),
+    };
+    ast_from_text(&format!("mod C {{{body_newline}{body_indent}{body}{body_newline}}}"))
+}
+
+pub fn mod_(name: ast::Name, body: Option<ast::ItemList>) -> ast::Module {
+    let name = name.to_string();
```
ShoyuVanilla:
```suggestion
```
I guess this won't be needed as `ast:Name` implements `Display`?

Hmikihiro: Thanks you. I removed it. ast::Name already implement Display.

PR #20374, opened 2025-08-03
Add an assist handler "move_use_to_top_level" to move use statement to top level

sgasho:
closes: rust-lang/rust-analyzer#20326 

```rust
fn main() {
     use std::collections::HashMap$0;
     let map = HashMap::new();
}
```

↓

```rust
use std::collections::HashMap;

fn main() {
    let map = HashMap::new();
}
```

Attributes and comments will be moved to the top level simultaneously.

example
```rust
fn main() {
    #[allow(unused)]
    use std::collections::HashMap$0;
    let map = HashMap::new();
}
```

↓

```rust
#[allow(unused)]
use std::collections::HashMap;

fn main() {
    let map = HashMap::new();
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/move_use_to_top_level.rs
+++ b/crates/ide-assists/src/handlers/move_use_to_top_level.rs
@@ -1,4 +1,5 @@
 use crate::assist_context::{AssistContext, Assists};
+use ide_db::FileId;
 use ide_db::assists::AssistId;
 use ide_db::imports::insert_use::{ImportScope, ImportScopeKind};
 use ide_db::source_change::SourceChangeBuilder;
@@ -8,7 +9,7 @@ use syntax::{
         BlockExpr, HasModuleItem, Item, Module, SourceFile, Use, Whitespace, edit_in_place::Indent,
         make,
     },
-    ted,
+    syntax_editor::{Position, SyntaxEditor},
 };
 
 // Assist: move_use_to_top_level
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/move_use_to_top_level.rs
```diff
@@ -0,0 +1,562 @@
...
+use ide_db::assists::AssistId;
+use ide_db::imports::insert_use::{ImportScope, ImportScopeKind};
+use ide_db::source_change::SourceChangeBuilder;
+use syntax::{
+    AstNode, AstToken,
+    ast::{
+        BlockExpr, HasModuleItem, Item, Module, SourceFile, Use, Whitespace, edit_in_place::Indent,
+        make,
+    },
+    ted,
```
ShoyuVanilla: We're phasing out direct mutable syntax tree usages like `ted`. Please use APIs such as `SyntaxEditor` instead.
For context,
- https://github.com/rust-lang/rust-analyzer/issues/15710
- https://github.com/rust-lang/rust-analyzer/issues/18285
- https://github.com/rust-lang/rust-analyzer/issues/17491

sgasho: @ShoyuVanilla 
Thanks! I replaced direct mutable syntax tree usages with SyntaxEditor in the latest commit

[refactor: replace direct mutable syntax tree usages with SyntaxEditor](https://github.com/rust-lang/rust-analyzer/pull/20374/commits/7e3532889d2257c8fd0b3ac80147b6eab23e3aa6)

ChayimFriedman2:

Issue #15566, opened 2023-09-06
`std` uses a `Thin` trait alias which breaks mir eval


Veykril:
https://doc.rust-lang.org/std/ptr/fn.null_mut.html is defined as:
```rs
pub const fn null_mut<T: ?Sized + Thin>() -> *mut T;
```
where `Thin` is defined as:
```rs
pub trait Thin = Pointee<Metadata = ()>;
```

r-a currently does not handle trait aliases correctly, as such mir eval bails out on almost all code that does `Vec` allocations via `Vec::with_capacity` because we lack the knowledge that `Metadata = ()` in `null_mut` resulting in a `Expected <T as Pointee>::Metadata, found ()` error.

Veykril:
@HKalbasi this is the cause for the mir failures (or at least part of them)

HKalbasi:
Yes it seems this is the reason of failure in mir eval of debug impls. For mir lowering failures the problem was in the stats, fixed by #15567

PR #20376, opened 2025-08-04
internal: merge Trait and TraitAlias handling

fee1-dead:
Much of the current infrastructure around traits are good and cool but trait aliases don't get to enjoy type inference because they are treated as distinct entities from traits.

Doesn't seem easy to change all the existing facilities to accept something like `Either<Trait, TraitAlias>`.

We can actually just merge them and most things will work. Second commit adds a simple test.

The core of this change is in `crates/syntax/rust.ungram`.

### DIFF
--- a/crates/hir-def/src/signatures.rs
+++ b/crates/hir-def/src/signatures.rs
@@ -402,6 +402,7 @@ bitflags::bitflags! {
         const SKIP_BOXED_SLICE_DURING_METHOD_DISPATCH = 1 << 6;
         const RUSTC_PAREN_SUGAR = 1 << 7;
         const COINDUCTIVE = 1 << 8;
+        const ALIAS = 1 << 9;
     }
 }
 
@@ -425,6 +426,9 @@ impl TraitSignature {
         }
         if source.value.unsafe_token().is_some() {
             flags.insert(TraitFlags::UNSAFE);
+        }
+        if source.value.eq_token().is_some() {
+            flags.insert(TraitFlags::ALIAS);
         }
         if attrs.by_key(sym::fundamental).exists() {
             flags |= TraitFlags::FUNDAMENTAL;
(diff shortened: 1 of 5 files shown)


crates/hir-def/src/nameres/assoc.rs
```diff
@@ -51,10 +51,18 @@ impl TraitItems {
         tr: TraitId,
     ) -> (TraitItems, DefDiagnostics) {
         let ItemLoc { container: module_id, id: ast_id } = tr.lookup(db);
+        let ast_id_map = db.ast_id_map(ast_id.file_id);
+        let source = ast_id.with_value(ast_id_map.get(ast_id.value)).to_node(db);
+        if source.eq_token().is_some() {
+            // FIXME(trait-alias) probably needs special handling here
```
ShoyuVanilla, line 57: I guess this would be okay as this PR treats `TraitAlias`s somewhat like new trait with bounds and no assoc items and rust-analyzer won't demand assoc items for such traits or (currently for) trait aliases?

ShoyuVanilla (approved):
I think this is a solid improvement overall. It might end up marking trait aliases as actual traits, but it correctly resolves associated types and, more importantly, trait methods. That alone could potentially address our long-standing issues with trait aliases, such as #15566 .

My only hesitation is whether this could complicate the merging of #20329. I skimmed that PR and only noticed a few minor references to `TraitAlias`, so the impact may be limited.

That said, I'm unsure whether we should merge this before or after #20329. cc @rust-lang/rust-analyzer

crates/hir-def/src/nameres/assoc.rs
fee1-dead: Yeah, I reconsidered and thought it might not need special handling halfway into implementing this PR. I originally was thinking maybe we should return the associated items of the trait that it aliases but we already lower them with bound handling - so the aliased trait will be treated as a super trait

fee1-dead:
#20329 was merged already and GitHub does not report any merge conflicts with this PR - so I think this is good to go?

Veykril:
Surprising that there are no conflicts, but in that case yes let's merge this 👍

fee1-dead:
Rebased, only the newest commit has changed, please review :)

PR #20379, opened 2025-08-04
fix(hover): unify horizontal rule formatting to `---`

skewb1k:
Currently, hover documentation uses both `___` and `---` as horizontal rule delimiters. While both styles are valid according to the GitHub Flavored Markdown Spec, the mixed usage seems accidental to me.

This PR change all horizontal breaks to `---` because it is less ambiguous and already more common in rust-analyzer than `___`.

I also believe that such consistency makes the code and tests a bit easier to maintain.


ChayimFriedman2 (approved):
I don't know if this is better but it's also not worse.

Thanks!

lnicola:
changelog internal (first contribution) consistently use `---` for horizontal rules

Issue #20367, opened 2025-08-02
rust-project.json runnable type check does nothing.


TroyKomodo:
https://github.com/rust-lang/rust-analyzer/blob/68e7ec90bf29c9b17d0dcdb3358de5dee7f4afb5/crates/project-model/src/project_json.rs#L328

This option when constructing a rust-project.json has no effect and should either be removed or a comment should be stated why it no longer works.

Git blame doesnt really show anything when it comes to why it was added / if it ever worked.

Pavankalyan32:
You're right—I checked the codebase, and this option doesn't seem to have any effect. It's parsed from rust-project.json but never used anywhere in the logic.

I also looked at git blame and git log -S, but there’s no clear explanation of why it was added or if it ever worked.

Unless there's a reason to keep it for future plans, I'd suggest either:

Removing the field entirely, or

Adding a comment explicitly stating that it's a no-op for now.

davidbarsky:
I added it and it never worked. I wanted to migrate the `$SAVED_FILE` functionality over to this, but never got around to it.

TroyKomodo:
Removing it doesnt seem like a good option, since some people seem to use this in their code. (notably [rules_rust](https://github.com/bazelbuild/rules_rust/blob/a44ac6e1bc66edcd4e45b9277eb51ff0981de742/tools/rust_analyzer/rust_project.rs#L266-L271) in bazel).

That being said, would it be possible to use this command instead of the `rust-analyzer.check.overrideCommand` for on the fly checks?

TroyKomodo:
https://github.com/rust-lang/rust-analyzer/issues/20359#issuecomment-3166304854 I wrote a comment about this but I implemented it here, I dont particularly like my implementation but am open to suggestions on improvements, and if I should PR this.

Wilfred:
I think we should remove it: dead code doesn't buy us anything, and we can always pull stuff out of git history if we want to build a full version later.

This needn't break rules_rust in bazel, provided that we handle the now-unknown JSON value gracefully. I removed the corresponding `check` runnable in rust-project (part of buck) in https://github.com/facebook/buck2/commit/5a565c288858a249e7043a27c2e271c92daf35bd as it was confusing before.

I don't feel strongly between `overridecommand` versus a `check` runnable though. In principle a `check` runnable could be clearer if you have some files open in a non-cargo workspace and others in a normal cargo project, but VS Code allows per-workspace settings and it hasn't been an issue for us in practice.

Issue #20359, opened 2025-08-01
better testing support for rust-project.json


TroyKomodo:
The debug button 

<img width="948" height="246" alt="Image" src="https://github.com/user-attachments/assets/cffdad82-edba-4290-bd10-5f3d18473918" />

does not do anything when clicked when using rust-project.json or discoverConfig.

Likewise there doesnt seem to be a way for cargo-insta to be / snapshots to be added via rust-project. 

There also doesnt seem to be a way to run all tests in a module this way 

(using rust-project.json)
<img width="707" height="159" alt="Image" src="https://github.com/user-attachments/assets/1624f9f8-ec06-423a-a2ee-33b2df2beef2" />

(using cargo)
<img width="886" height="421" alt="Image" src="https://github.com/user-attachments/assets/b5833e50-7482-414f-a95c-4256d8dacd80" />

TroyKomodo:
https://github.com/rust-lang/rust-analyzer/blob/68e7ec90bf29c9b17d0dcdb3358de5dee7f4afb5/crates/rust-analyzer/src/target_spec.rs#L101-L103

Seems to be hard-coded to `None`

Veykril:
I believe this is a duplicate of https://github.com/rust-lang/rust-analyzer/issues/15892

Veykril:
Or maybe https://github.com/rust-lang/rust-analyzer/issues/12105

TroyKomodo:
@Veykril would you be open to prs adding support for this?

and potentially 

this one too

https://github.com/rust-lang/rust-analyzer/blob/b3b3d93fb99c9e1165b7f0acb238c21a600d1cb3/editors/code/src/debug.ts#L81-L91

TroyKomodo:
I put up some changes here for all the project-json stuff I needed to add to get my project to have a similar experience to cargo.

https://github.com/TroyKomodo/rust-analyzer/tree/troy/project-json

I still need to update the extension to allow for the debugger to work however this is a little bit trickier. 

I dont mind PRing this back too if its something thats acceptable.

PR #20396, opened 2025-08-06
fix: project json runnable kind bin substitution

TroyKomodo:
fixes a bug related to substitution in project json bin not correctly replacing the `{label}` tag with the label of the current item.


lukasoyen:
I am not sure this is a fix, as the old behavior was not a bug, but documented: https://github.com/rust-lang/rust-analyzer/blob/3d2c40d705f5663fccf4a4d89ea457876ac32df6/docs/book/src/non_cargo_based_projects.md?plain=1#L203-L207.

Though we have another comment that suggests this might be a bug: https://github.com/rust-lang/rust-analyzer/blob/3d2c40d705f5663fccf4a4d89ea457876ac32df6/crates/project-model/src/project_json.rs#L316-L319

I am all for something like this change. I would propose to introduce a new "binOne" similar to "testOne" where this substitution applies.

cc @davidbarsky @Wilfred

Wilfred:
I added the `kind: 'testOne' | string;` in #18878, as `testOne` was the only value that I could see was actually in use.

I'm not necessarily opposed to applying the substitution on all runnables, but we should update the non_cargo_based_projects.md accordingly. `test_id` makes no sense in other runnable types at least. I think we'd also need support for other types of runnables -- AFAICS this PR isn't useful on its own.

TroyKomodo:
On the comment about other runnable types, currently there is only TestOne and Run. The Check type is not used anywhere in the code see #20367

TroyKomodo:
Honestly I would prefer if we go with a solution I have on my fork. See https://github.com/rust-lang/rust-analyzer/issues/20359#issuecomment-3166304854

By adding the runnables to the crate itself instead of having a global runnables. 

The global runnables in my opinion were not well thought out since it's very likely in the case of external build systems that you have different targets for different runnables on the same crate. With the current design you can't really do this because we only have one label to substitute.

Wilfred:
> The global runnables in my opinion were not well thought out since it's very likely in the case of external build systems that you have different targets for different runnables on the same crate.

I'd actually argue that's a compelling case for global runnables! :)

By having a single command that rust-analyzer runs when you modify foo.rs, you can do nice things like merge duplicate build warnings (having a foo-proj and foo-proj-unittest is common in both buck and cargo). It also makes the JSON much smaller.

TroyKomodo:
I'm not sure I follow, I am referring to a single crate that has different labels depending on the runnable.

Right now in the bazel implementation we always prioritize the test target to allow for the `run test` button on unit tests to work, naturally this means that the `run binary` button does not.

https://github.com/bazelbuild/rules_rust/blob/a44ac6e1bc66edcd4e45b9277eb51ff0981de742/tools/rust_analyzer/aquery.rs#L239-L245

It seems buck2 does something similar here https://github.com/facebook/buck2/blob/5a565c288858a249e7043a27c2e271c92daf35bd/integrations/rust-project/src/buck.rs#L368-L373.

There are only 2 runnables that work, `TestOne` and `Run` and currently as implemented nobody allows for both to be used at the same time. In my mind the ideal solution would be to have both working. The per crate runnables was one solution to the problem another solution might be to add multiple substitution strings (other than just `{label}`). 

<details>

<summary> 
<i>It also makes the JSON much smaller. </i>
</summary>

I don't believe this is an issue at all. This json is not being added to the VCS and if you use the discover command is simply piped and does not even write to disk. The argument that "larger json slower parse times" isnt really true. This file does not get nearly large enough for that to matter. In my case its just under 1MB (pretty-printed). I am writing it out because I want to debug the generated file when using discover command. These runnables are only added to the workspace crates not all dependencies. 
```
> ls -lah .rust-project.bazel.json
-rw-r--r-- 1 troy users 935K Aug  9 13:07 .rust-project.bazel.json
```


</details>

Wilfred:
Yeah, so test crates are tricky. In rust-project we don't just take the test crate, instead we attempt to merge the dependencies between the `foo` and `foo-unittest` crates (see https://github.com/facebook/buck2/commit/f6ea0a46b8ec6e086e918b9d0c8cfe89e2356950 and follow-up commits like https://github.com/facebook/buck2/commit/4e4aaf83824245b7d063e3775172520e2bce3ea8).

Being able to say "I've modified foo.rs, what command should I run?" is easier when you have a single check command. If you have N crates that include foo.rs and they each have a check command, what should rust-analyzer do? Can it run them all, can they be run in parallel, and how do you merge the results?

FWIW [cargo deduplicates errors across crates as of Rust 1.55](https://blog.rust-lang.org/2021/09/09/Rust-1.55.0/#cargo-deduplicates-compiler-errors) and we have similar logic in a `rust-check foo.rs` command. This is nice because it's easy to write

```
// oops forgot cfg(test)
use some_test_helper;
```

and you get a squiggle complaining that your import isn't always used.

Regarding JSON size, I agree it's not a big deal. I've seen rust-project.json files up to ~20 MiB in the last week, which isn't a huge problem. Adding a lot of redundancy to the file format would make reading the files during debugging slightly more annoying, but that's all.

TroyKomodo:
> Yeah, so test crates are tricky. In rust-project we don't just take the test crate, instead we attempt to merge the dependencies between the `foo` and `foo-unittest` crates 

It is largely the same in the bazel implementation. 

> Being able to say "I've modified foo.rs, what command should I run?" is easier when you have a single check command. If you have N crates that include foo.rs and they each have a check command, what should rust-analyzer do? Can it run them all, can they be run in parallel, and how do you merge the results?

I largely agree that having the same file be apart of multiple crates is problematic when deciding which commands to run. However this PR is specifically related to the run binary type. I do want to also bring in my changes on my RA fork which include changes to complete the implementation for #20367. 

--

That being said, I am still unclear on what exactly is wrong with this PR particularly. Most of the arguments in this PR are related to changes which I intent to bring forward as a more general solution to #20359 so perhaps we should better move those discussions there?

cormacrelf:
#18043 incorporated this change and has been merged, so this is done.

PR #20408, opened 2025-08-08
Add ide-assist: generate_documentation_from_trait

A4-Tacks:
Generate documents from items defined in the trait.

Implement https://github.com/rust-lang/rust-analyzer/pull/20407#issuecomment-3167651425

Example
===

```rust
trait Foo {
    /// some docs
    fn foo(&self);
}
impl Foo for () {
    fn $0foo(&self) {}
}
```
->
```rust
trait Foo {
    /// some docs
    fn foo(&self);
}
impl Foo for () {
    /// some docs
    fn foo(&self) {}
}
```


Veykril:
This does not seem useful to me. Why would you need to duplicate the docs from the trait to the impl? And if you do you can go to def and copy

PR #20419, opened 2025-08-10
internal: Make flycheck generational

ShoyuVanilla:
Resolves https://github.com/rust-lang/rust-analyzer/issues/19854

### DIFF
--- a/crates/rust-analyzer/src/flycheck.rs
+++ b/crates/rust-analyzer/src/flycheck.rs
@@ -1,7 +1,12 @@
 //! Flycheck provides the functionality needed to run `cargo check` to provide
 //! LSP diagnostics based on the output of the command.
 
-use std::{fmt, io, process::Command, time::Duration};
+use std::{
+    fmt, io,
+    process::Command,
+    sync::atomic::{AtomicUsize, Ordering},
+    time::Duration,
+};
 
 use cargo_metadata::PackageId;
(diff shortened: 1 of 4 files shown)


crates/rust-analyzer/src/flycheck.rs
```diff
@@ -138,36 +142,62 @@ pub(crate) struct FlycheckHandle {
     sender: Sender<StateChange>,
     _thread: stdx::thread::JoinHandle,
     id: usize,
+    generation: Mutex<DiagnosticsGeneration>,
```
Veykril: Given thats just a type alias might as well use an `AtomicUsize` here (with relaxed orderings)

Veykril: Or do we need to `Sync`ness here in general? If not `Cell<usize>` ought to work

Veykril (approved):

crates/rust-analyzer/src/flycheck.rs
ShoyuVanilla: I guess there won't be concurrent problems in general but we have this line that demands this to be sync:

https://github.com/rust-lang/rust-analyzer/blob/d307bc614e04d56ef723d1f81f18f2405a8365e4/crates/rust-analyzer/src/global_state.rs#L108

I've used mutex just because some platforms might not support `AtomicUsize` but using it seems better since major platforms support it

PR #20445, opened 2025-08-13
fix: Attach db for inlay hint compute

Veykril:


lnicola:
changelog fixup #20329

PR #20446, opened 2025-08-13
next-solver fun time

lcnr:
should fix the metrics issue for hyper, unless there's also another issue here :thinking:

### DIFF
--- a/crates/hir-ty/src/layout.rs
+++ b/crates/hir-ty/src/layout.rs
@@ -182,7 +182,7 @@ pub fn layout_of_ty_ns_query<'db>(
     };
     let dl = &*target;
     let cx = LayoutCx::new(dl);
-    let infer_ctxt = interner.infer_ctxt().build(TypingMode::non_body_analysis());
+    let infer_ctxt = interner.infer_ctxt().build(TypingMode::PostAnalysis);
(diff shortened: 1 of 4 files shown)


crates/hir-ty/src/layout.rs
```diff
@@ -182,7 +182,7 @@ pub fn layout_of_ty_ns_query<'db>(
     };
     let dl = &*target;
     let cx = LayoutCx::new(dl);
-    let infer_ctxt = interner.infer_ctxt().build(TypingMode::non_body_analysis());
+    let infer_ctxt = interner.infer_ctxt().build(TypingMode::PostAnalysis);
     let cause = ObligationCause::dummy();
     let ty = deeply_normalize(infer_ctxt.at(&cause, ParamEnv::empty()), ty).unwrap_or(ty);
```
lcnr, line 187: opaque types should *just be normalized* like any other alias. For this to work, computing the layout uses `TypingMode::PostAnalysis` which then fetches the underlying type for all opaques in the program

Veykril (approved):

Veykril:
Thanks!

lnicola:
changelog fixup #20329

PR #20452, opened 2025-08-13
Add ide-assist: convert_if_let_to_let_else

A4-Tacks:
Converts let statement with if-let initializer to let-else statement.

Example
---
```rust
# //- minicore: option
fn foo(opt: Option<()>) {
    let val$0 = if let Some(it) = opt {
        it
    } else {
        return
    };
}
```
->
```rust
fn foo(opt: Option<()>) {
    let Some(val) = opt else {
        return
    };
}
```


A4-Tacks:

PR #20456, opened 2025-08-14
Add guard to let-chain for replace_match_with_if_let

A4-Tacks:
```rust
fn main() {
    match$0 Some(0) {
        Some(n) if n % 2 == 0 && n != 6 => (),
        _ => code(),
    }
}
```
->
```rust
fn main() {
    if let Some(n) = Some(0) && n % 2 == 0 && n != 6 {
        ()
    } else {
        code()
    }
}


ShoyuVanilla (approved):
Thanks!

PR #20457, opened 2025-08-14
Add ide-assist: generate_inverse_ops_impl

A4-Tacks:
Generate inverse ops implement.

Example
---
```rust
struct Foo(i32);
impl core::ops::Add<i32> $0for Foo {
    type Output = i32;

    fn add(self, rhs: i32) -> Self::Output {
        self.0 + rhs
    }
}
```
->
```rust
struct Foo(i32);
impl core::ops::Add<Foo> for i32 {
    type Output = i32;

    fn add(self, rhs: Foo) -> Self::Output {
        rhs.0 + self
    }
}

impl core::ops::Add<i32> for Foo {
    type Output = i32;

    fn add(self, rhs: i32) -> Self::Output {
        self.0 + rhs
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/generate_inverse_ops_impl.rs
+++ b/crates/ide-assists/src/handlers/generate_inverse_ops_impl.rs
@@ -1,0 +1,228 @@
+use hir::{EditionedFileId, Semantics};
+use ide_db::{
+    RootDatabase,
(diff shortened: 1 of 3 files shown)


A4-Tacks:

PR #20459, opened 2025-08-14
fix: Track diagnostic generations per package

Veykril:
This means that once flycheck is done, we will clear all diagnostics from the previous run. Why is this nice? Well, right now if you have flycheck diagnostics in crate b, then edit crate a (where crate b depends on crate a) we will get diagnostics in crate a while b's old diagnostics stick around. With this change b will get cleared properly, reducing the amount of diagnostics shown in the problems pane of editors.

r? @ShoyuVanilla

### DIFF
--- a/crates/rust-analyzer/src/diagnostics.rs
+++ b/crates/rust-analyzer/src/diagnostics.rs
@@ -26,17 +26,15 @@ pub struct DiagnosticsMapConfig {
 
 pub(crate) type DiagnosticsGeneration = usize;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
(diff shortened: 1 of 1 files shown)


ShoyuVanilla (approved):
Oh, right. In current implementation, the diagnostics from the previous runs would outlive into the next generation altogether workspacewise if there's any diagnostics reported in that workspace. I had considered this (actually, inscribing diagnostics into both workspaces and packages) for the different purpose, for granulated erasure of outdated diagnostics for scenarios like check on save is turned off and flychecks are triggered sporadically on packages but it didn't worked well and I just oversimplified things with disappointment. But as you said, this makes solid improvements in normal usage scenarios 👍

PR #20472, opened 2025-08-16
Remove conflicting advice

BenjaminBrienen:


fee1-dead:
I'm not sure if this is an improvement, because clippy lints are meant to be annoying (cause, clippy) lints that you can choose to follow or ignore.

But it's probably clearer to cite specific instances of both styles used in this repository so we can get a better picture?

BenjaminBrienen:
The "suggested" style is simply not used in the repo except in some obscure test data.
image

edit: ` !\(.* [<>+=]+ .*\)` yields 12 actual code results, of which, only a few are "true" matches

image

BenjaminBrienen:
`\b[a-z_.()0-9]* (<=|>=)+ \S*\b` yields 236 results that seem like relevant examples at a glance
image
image

PR #20508, opened 2025-08-22
Add convert unwrap_or_else for convert_bool_then

A4-Tacks:
Example
---
```rust
//- minicore:bool_impl
fn main() {
    true.t$0hen(|| 3).unwrap_or_else(|| 4)
}
```

Old output:

```rust
//- minicore:bool_impl
fn main() {
    if true {
        Some(3)
    } else {
        None
    }.unwrap_or_else(|| 4)
}
```

This PR current output:

```rust
//- minicore:bool_impl
fn main() {
    if true {
        3
    } else {
        4
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/convert_bool_then.rs
+++ b/crates/ide-assists/src/handlers/convert_bool_then.rs
@@ -177,6 +177,7 @@ pub(crate) fn convert_bool_then_to_if(acc: &mut Assists, ctx: &AssistContext<'_>
     if !assoc.implementing_ty(ctx.sema.db)?.is_bool() {
         return None;
     }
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
I dislike this. Who will write `.then().unwrap_or_else()`? It's just a convoluted way to write `if`-`else`. I don't see the value in this.

A4-Tacks:
But at least it's better than the old output

ChayimFriedman2:
In this case, yeah. The problem is that I don't believe this case is realistic.

A4-Tacks:
You're right, and there's `clippy::obfuscated_if_else`

PR #20509, opened 2025-08-22
Fix indent for move_guard_to_arm_body

A4-Tacks:
Input:

```rust
fn main() {
    match 92 {
        x $0if true
            && true
            && true =>
        {
            {
                false
            }
        },
        _ => true
    }
}
```

Old output:

```rust
fn main() {
    match 92 {
        x =>
        if true
                    && true
                    && true {
            {
                    {
                        false
                    }
                }
        },
        _ => true
    };
}
```

This PR fixed:

```rust
fn main() {
    match 92 {
        x => if true
            && true
            && true {
            {
                {
                    false
                }
            }
        },
        _ => true
    }
}
```


ShoyuVanilla (approved):
Thanks!

PR #20521, opened 2025-08-24
Add BreakExpr completion suggest

A4-Tacks:


crates/ide-completion/src/context/analysis.rs
```diff
@@ -1917,6 +1925,35 @@ fn is_in_breakable(node: &SyntaxNode) -> Option<(BreakableKind, SyntaxNode)> {
         })
 }
 
+fn find_breakvalueable(
```
Veykril: We already have https://github.com/rust-lang/rust-analyzer/blob/51d422775c8698f59958fb7f1c6eb3f9af432d77/crates/ide/src/goto_definition.rs#L514-L557 that does the job I think

So we should move that function into `ide-db` and re-use it here appropriately

A4-Tacks: Okay

PR #20533, opened 2025-08-25
Add assists: convert_match_to_let

### LINKED ISSUE #17773: Feature: Assist to convert `match` to `let`
```rust
struct Example {
    v: i32,
}

fn foo(e: Example) {
    match e {
        Example { v } => undefined(v),
    }
}
```

This `match` can be replaced with a `let`. It would be useful if rust-analyzer would offer to convert it. In this case, rustc does give an applicable suggestion, but that suggestion is only available when check has run and when the program is free of other errors, so it would be useful to have this refactor available even when the program does not compile, for use as an editing tool on the way to getting it to compile.

@rustbot label +A-assists

A4-Tacks:
Convert single arm match to let.

Close #17773

Example
---
```rust
struct Example { v: i32 }

fn foo(e: Example) {
    match e {
        $0Example { v } => undefined(v),
    }
}
```
->
```rust
struct Example { v: i32 }

fn foo(e: Example) {
    let Example { v } = e;
    undefined(v)
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/convert_match_to_let.rs
+++ b/crates/ide-assists/src/handlers/convert_match_to_let.rs
@@ -1,0 +1,374 @@
+use itertools::Itertools;
+use syntax::{
+    AstNode, SyntaxKind,
+    ast::{
+        self,
+        edit::{AstNodeEdit, IndentLevel},
+        make,
+    },
+    syntax_editor::{Position, SyntaxEditor},
+};
+
+use crate::{
+    AssistId,
+    assist_context::{AssistContext, Assists},
+};
+
+// Assist: convert_match_to_let
+//
+// Convert single arm match to let.
+//
+// ```
+// struct Example { v: i32 }
+//
+// fn foo(e: Example) {
+//     match e {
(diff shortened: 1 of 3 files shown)


Natural-selection1:
PixPin_2025-08-30_22-04-15

clippy already did this

A4-Tacks:
> clippy already did this

We need is flexible real-time conversion and quick editing

Natural-selection1:
Whether it's an erroneous update or almost all features, RA generally performs them after the file is saved. 
I believe the situation most users would encounter is having two assists with the same functionality and name(one from RA, one from Clippy), which quite confusing. 

If we reimplemente this feature that already existed in Clippy, could we all use the same justification to reimplement everything else? 
I don't think that would be appropriate.

A4-Tacks:
Firstly, most people do not set clippy to save checks

Secondly, there are many examples of this, such as:

- `merge_nested_if.rs`
- `remove_mut.rs`
- `remove_unused_imports.rs`
- `remove_parentheses.rs`

A4-Tacks:

PR #20534, opened 2025-08-25
fix: improve semicolon handling in `toggle_macro_delimiter`

A4-Tacks:


crates/ide-assists/src/handlers/toggle_macro_delimiter.rs
```diff
@@ -119,7 +127,29 @@ macro_rules! sth {
...
+    sth!$0( );
+}
+            "#,
+            r#"
+macro_rules! sth {
+    () => {};
+}
+
+fn foo() {
+    sth!{ }
```
ShoyuVanilla, line 149: How about inserting a whitespace here as well like `sth! { }`? Anyway, I think this is a decent fix by itself.

ShoyuVanilla (approved):

PR #20541, opened 2025-08-26
Add ide-assist: remove_reference

A4-Tacks:
Remove parameter reference.

Example
---

```rust
fn foo($0state: &bool) {
    if *state {}
}
fn bar() {
    let state = true;
    foo(&state);
}
```
->
```rust
fn foo(state: bool) {
    if state {}
}
fn bar() {
    let state = true;
    foo(state);
}
```


A4-Tacks:

PR #20556, opened 2025-08-28
Add ide-assist: toggle_lint_attr

A4-Tacks:
Change lint attribute level.

Example
---
```
#[$0allow(dead_code)]
fn foo() {}
```
->
```
#[expect(dead_code)]
fn foo() {}
```

---

```
#[$0forbid(dead_code)]
fn foo() {}
```
->
```
#[deny(dead_code)]
fn foo() {}
```


Veykril:
This seems too niche to me.

PR #20564, opened 2025-08-29
Add ReturnExpr support for add_return_type

A4-Tacks:
Example
---
```rust
fn foo() {
    return 45$0;
    todo!()
}
```
->
```rust
fn foo() -> i32 {
    return 45;
    todo!()
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/add_return_type.rs
+++ b/crates/ide-assists/src/handlers/add_return_type.rs
@@ -17,7 +17,8 @@ use crate::{AssistContext, AssistId, Assists};
 // fn foo() -> i32 { 42i32 }
 // ```
 pub(crate) fn add_return_type(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -> Option<()> {
-    let (fn_type, tail_expr, builder_edit_pos) = extract_tail(ctx)?;
+    let (fn_type, tail_expr, builder_edit_pos) =
+        extract_ret(ctx.find_node_at_offset()?, Either::Right(ctx.selection_trimmed()))?;
     let module = ctx.sema.scope(tail_expr.syntax())?.module();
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/add_return_type.rs
```diff
@@ -133,10 +134,16 @@ fn peel_blocks(mut expr: ast::Expr) -> ast::Expr {
     expr
 }
 
-fn extract_tail(ctx: &AssistContext<'_>) -> Option<(FnType, ast::Expr, InsertOrReplace)> {
-    let node = ctx.find_node_at_offset::<Either<ast::ClosureExpr, ast::Fn>>()?;
+fn extract_ret(
+    node: Either<ast::ReturnExpr, Either<ast::ClosureExpr, ast::Fn>>,
+    ret: Either<ast::Expr, TextRange>,
```
ChayimFriedman2, line 139: You don't need the recursion, you can go from a `TextRange` to an `AstNode` directly via `find_node_at_range()`.

ChayimFriedman2 (changes requested):

crates/ide-assists/src/handlers/add_return_type.rs
A4-Tacks: This `TextRange` is not for obtaining nodes, but for checking: `return_type_range.contains_range(range)`

ChayimFriedman2: So is it just `return_type_range.contains_range(selection_range)`?

A4-Tacks: Yes, compared to `Option<ast::Expr>`, I think `Either<ast::Expr, TextRange>` better expresses the perception of `TextRange`

PR #20565, opened 2025-08-29
feat: add "Convert range `for` to `while`" assist

A4-Tacks:
Convert for each range into while loop.

Close rust-lang/rust-analyzer#20916

```rust
fn foo() {
    $0for i in 3..7 {
        foo(i);
    }
}
```
->
```rust
fn foo() {
    let mut i = 3;
    while i < 7 {
        foo(i);
        i += 1;
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/convert_range_for_to_while.rs
+++ b/crates/ide-assists/src/handlers/convert_range_for_to_while.rs
@@ -43,9 +43,14 @@ pub(crate) fn convert_range_for_to_while(acc: &mut Assists, ctx: &AssistContext<
     let body = for_.loop_body()?;
     let last = previous_non_trivia_token(body.stmt_list()?.r_curly_token()?)?;
 
+    let description = if end.is_some() {
+        "Replace with while expression"
+    } else {
+        "Replace with loop expression"
+    };
     acc.add(
         AssistId::refactor("convert_range_for_to_while"),
-        "Replace to while or loop",
+        description,
         for_.syntax().text_range(),
         |builder| {
             let mut edit = builder.make_editor(for_.syntax());


Natural-selection1:
`while` does not seem to have better readability and may also cause unexpected variable shadowing

```rust
let i = "hi";
for i in 1..50 {
    do_something();
}
println!("{i}"); // hi
```

```rust
let i = "hi";
let mut i = 1;
while i < 50 {
    do_something();
    i += 1;
}
println!("{i}"); // 50
```

A4-Tacks:
One of the purposes is to use loop variables after the loop ends

Natural-selection1:
Perhaps some tests should be added to show that it won't take effect on the for loop of a non-range expression? Also, what about those range expressions where the upper limit is determined by a non-number, like:
```rust
fn main() {
    for i in 1..fibonacci(5) {}
}

const fn fibonacci(n: u32) -> u32 {
    match n {
        0..=2 => n,
        _ => fibonacci(n - 1) + fibonacci(n - 2),
    }
}
```
And expressions with omitted bounds:
```rust
fn main() {
    let y: u8 = 5;
    for x in y.. {}
}
```

A4-Tacks:
> Also, what about those range expressions where the upper limit is determined by a non-number, like:

This assist does not consider side effects within its range-expr, and if necessary,
`extract_variable` can be used

> And expressions with omitted bounds:

This will become a loop:

```rust
let y: u8 = 5;
let mut x = y;
loop {
    x += 1;
}
```

Natural-selection1:
> ```rust
> let y: u8 = 5;
> let mut x = y;
> loop {
>     x += 1;
> }
> ```
`panic!`, `x += 1` in `loop` is dangerous

haha, this maybe better? 
```rust
let y: u8 = 5;
let mut x = y;
while x < u8::MAX { 
    do_something();
    x += 1;
}
```
and
```rust
let y = 5; // default i32
let mut x = y;
while x < i32::MAX {
    do_something();
    x += 1;
}
```

A4-Tacks:
`..=` convert to `<=`, So `i32::MAX` can also `panic!`

A4-Tacks:
Using unbounded-range usually ends the loop prematurely, such as a `break` and `return`, which should be irrelevant

Natural-selection1:
> `..=` convert to `<=`, So `i32::MAX` can also `panic!`

That for loop never exist
PixPin_2025-08-30_14-46-54

A4-Tacks:
> That for loop never exist

But converted while exist:

```rust
while x <= i32::MAX {
    x += 1;
}
```

Natural-selection1:
Using a code action on an incorrect piece of code and getting an incorrect result is "reasonable", but now is that we are applying code action to a normal piece of code and it is causing a `panic!`.

A4-Tacks:
> That for loop never exist

Wait... this code is panic,

```rust
for i in 0i16.. {}
```

Still panic, that's no problem:

```rust
let mut i = 0i16;
loop {
    i += 1;
}
```

A4-Tacks:
|code|converted|panics|
|---|---|---|
|`a..b`|`while x<b`|no panic -> no panic|
|`a..`| `loop` | panic -> panic |
|`a..=b`|`while x<=b`|no panic -> when `b == {integer}::MAX`|

For the assists, `a..=b` minor panic changes is irrelevant

crates/ide-assists/src/handlers/convert_range_for_to_while.rs
```diff
@@ -0,0 +1,254 @@
...
+    let ast::Pat::IdentPat(pat) = for_.pat()? else { return None };
+    let iterable = for_.iterable()?;
+    let (start, end, step, inclusive) = extract_range(&iterable)?;
+    let name = pat.name()?;
+    let body = for_.loop_body()?;
+    let last = previous_non_trivia_token(body.stmt_list()?.r_curly_token()?)?;
+
+    acc.add(
+        AssistId::refactor("convert_range_for_to_while"),
+        "Replace to while or loop",
```
ShoyuVanilla: I think knowing which expression, either `loop` or `while`, will be emitted by this assist before applying it would be more preferable. Could we insert different label depending on the existence of `end` here?

crates/ide-assists/src/handlers/convert_range_for_to_while.rs
```diff
@@ -0,0 +1,254 @@
...
+            });
+            if let Some(end) = end {
+                elements.extend([
+                    make.token(T![while]).syntax_element(),
+                    make.whitespace(" ").syntax_element(),
+                    make.expr_bin(var_expr.clone(), op, end).syntax().syntax_element(),
+                ]);
+            } else {
+                elements.push(make.token(T![loop]).syntax_element());
+            }
```
ShoyuVanilla, line 85: We already have `SyntaxFactory::expr_while_loop` and you have implemented `SyntaxFactory::expr_loop` in this PR. Could we use them here?

A4-Tacks: If creating a new syntax node and replacing it, additional handling of attributes, labels, trivial, etc. is required

ShoyuVanilla: Okay, that sounds reasonable since our syntax editing infra is not so good for now, especially for trivia handlings 😄

crates/ide-assists/src/handlers/convert_range_for_to_while.rs
```diff
@@ -0,0 +1,255 @@
...
+    let for_kw = ctx.find_token_syntax_at_offset(T![for])?;
+    let for_ = ast::ForExpr::cast(for_kw.parent()?)?;
+    let ast::Pat::IdentPat(pat) = for_.pat()? else { return None };
+    let iterable = for_.iterable()?;
+    let (start, end, step, inclusive) = extract_range(&iterable)?;
+    let name = pat.name()?;
+    let body = for_.loop_body()?;
+    let last = previous_non_trivia_token(body.stmt_list()?.r_curly_token()?)?;
+
+    let description = if end.is_some() { "Replace to while" } else { "Replace to loop" };
```
ShoyuVanilla: Final nit: I think the following would be a bit clearer
```suggestion
    let description = if end.is_some() { "Replace with while expression" } else { "Replace with loop expression" };
```

ShoyuVanilla: Well, the proposition "with" would be correct here 😄

PR #20591, opened 2025-09-02
add braces indent

A4-Tacks:
- **Fix extract multiple item in impl for extract_module**
- **Fix indent for add_braces**


A4-Tacks:
Duplicated of #20845

PR #20592, opened 2025-09-02
Fix closure in match not applicable for add_braces

A4-Tacks:
Example
---

**Not applicable**:

```rust
fn foo() {
    match () {
        () => {
            t(|n|$0 n + 100);
        }
    }
}
```


ShoyuVanilla (approved):
Thanks!

PR #20604, opened 2025-09-04
Add cfg_attr predicate completion

A4-Tacks:
Partial fixes of rust-lang/rust-analyzer#14600

Example
---
```rust
#[cfg_attr($0, must_use)]
struct Foo;
```


ShoyuVanilla (approved):
Thanks!

PR #20611, opened 2025-09-04
Fix precedence parenthesis for replace_arith_op

A4-Tacks:
Example
---
```rust
fn main() {
    let x = 1*x $0+ 2;
}
```

**Before this PR**:

```rust
fn main() {
    let x = 1*x.wrapping_add(2);
}
```

**After this PR**:

```rust
fn main() {
    let x = (1*x).wrapping_add(2);
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/replace_arith_op.rs
+++ b/crates/ide-assists/src/handlers/replace_arith_op.rs
@@ -88,7 +88,11 @@ fn replace_arith(acc: &mut Assists, ctx: &AssistContext<'_>, kind: ArithKind) ->
         |builder| {
             let method_name = kind.method_name(op);
 
-            builder.replace(range, format!("{lhs}.{method_name}({rhs})"))
+            if lhs.precedence().needs_parentheses_in(ast::prec::ExprPrecedence::Postfix) {
+                builder.replace(range, format!("({lhs}).{method_name}({rhs})"))
+            } else {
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/replace_arith_op.rs
```diff
@@ -88,7 +88,11 @@ fn replace_arith(acc: &mut Assists, ctx: &AssistContext<'_>, kind: ArithKind) ->
         |builder| {
             let method_name = kind.method_name(op);
 
-            builder.replace(range, format!("{lhs}.{method_name}({rhs})"))
+            if lhs.precedence().needs_parentheses_in(ast::prec::ExprPrecedence::Postfix) {
+                builder.replace(range, format!("({lhs}).{method_name}({rhs})"))
+            } else {
+                builder.replace(range, format!("{lhs}.{method_name}({rhs})"))
+            }
```
ShoyuVanilla, line 95: Would be nice to migrate this to `SyntaxEditor` but I think this PR doesn't worsen things

ShoyuVanilla (approved):
Thanks!

crates/ide-assists/src/handlers/replace_arith_op.rs
A4-Tacks: Ok, in #20748

A4-Tacks: Ok, in #20748

PR #20617, opened 2025-09-05
Add documentation for troubleshooting flatpak permission issues

Ferynix:
Refer to https://github.com/rust-lang/rust-analyzer/issues/2873#issuecomment-1305760775 and my issue https://github.com/rust-lang/rust-analyzer/issues/20616

This change adds documentation to the troubleshooting FAQ for any future users experiencing cargo file resolving errors with rust-analyzer due to their IDE lacking permission to see the cargo binaries.

### DIFF
--- a/docs/book/src/faq.md
+++ b/docs/book/src/faq.md
@@ -5,14 +5,3 @@
 rust-analyzer fails to resolve `None`, and thinks you are binding to a variable
 named `None`. That's usually a sign of a corrupted sysroot. Try removing and re-installing
 it: `rustup component remove rust-src` then `rustup component install rust-src`.
-
-### VSCodium / another IDE cannot resolve cargo files with rust-analyzer
-
-This may be due to a misconfiguration of permissions with flatpak.
-This can be resolved by allowing flatpak to see the cargo binaries in the PATH with the following command:
-```bash
-flatpak --user override com.vscodium.codium  --env=PATH=/app/bin:/usr/bin:/home/$USER/.cargo/bin
-```
-This issue should be resolved in VS Code, if the issue persists with it or another IDE,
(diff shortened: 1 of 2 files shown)


Ferynix:
Documentation moved to the vs-code page

Feel free to edit any wording or detail improvements that should be made

lnicola:
Wait, isn't this the same as the suggestion right above?

> If you want to use Flatpak in combination with rustup, the following steps might help:
> - using Flatseal, you must add an environment variable called PATH. Set its value to the output from above, appending :~/.cargo/bin, where ~ is the path to your home directory. You must replace ~, as it won’t be expanded otherwise.

ptomato:
I think probably the suggestion given in https://github.com/VSCodium/vscodium/issues/2027#issuecomment-2367078850 is better - it doesn't require using the host's cargo installation.

I have the Codium flatpak in the system installation so I had to do it without `--user`, here's what worked for me:

```
flatpak install flathub org.freedesktop.Sdk.Extension.rust-stable//25.08
flatpak override --env=FLATPAK_ENABLE_SDK_EXT=rust-stable com.vscodium.codium
```

adding the `FLATPAK_ENABLE_SDK_EXT` variable is also something that could be done in the flatpak manifest.

lnicola:
@ptomato isn't that also right above in the docs? :smile: 

> If you use a Flatpak SDK for Rust, it must be in your `PATH`:
> * install the SDK extensions with `flatpak install org.freedesktop.Sdk.Extension.{llvm15,rust-stable}//23.08`
> * enable SDK extensions in the editor with the environment variable `FLATPAK_ENABLE_SDK_EXT=llvm15,rust-stable` (this can be done using flatseal or `flatpak override`)

But please feel free to update it to a more modern SDK etc.

ptomato:
My comment applied to the docs change in this PR, which didn't work for me. If my suggestion is already in the docs, great!

PR #20620, opened 2025-09-06
fix: add `else` keyword completion after `let` statements

A4-Tacks:
Example
---
```rust
fn foo() {
    let _ = 2 el$0
}
```
->
```rust
fn foo() {
    let _ = 2 else {
        $0
    };
}
```

---

And fix the no completion semicolon for `else` in incomplete LetStmt of rust-lang/rust-analyzer#20518

```rust
fn foo() {
    let _ = if true {} el$0
}
```

**Before this PR**:

```rust
fn foo() {
    let _ = if true {} else {
        $0
    }
}
```

**After this PR**:

```rust
fn foo() {
    let _ = if true {} else {
        $0
    };
}
```

### DIFF
--- a/crates/ide-completion/src/completions/expr.rs
+++ b/crates/ide-completion/src/completions/expr.rs
@@ -61,6 +61,7 @@ pub(crate) fn complete_expr_path(
         after_if_expr,
         in_condition,
         incomplete_let,
+        after_incomplete_let,
         in_value,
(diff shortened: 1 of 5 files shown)


A4-Tacks:
Closed as completed for #20620

A4-Tacks:
Sorry, I mistook this for an issue

PR #20627, opened 2025-09-08
Generate Deref assist for generic structs

phyBrackets:
Fixing issue rust-lang/rust-analyzer#20574

The fix ensures the assist correctly appears for generic structs without existing `Deref` implementations while maintaining accuracy for non-generic types.

### DIFF
--- a/crates/ide-assists/src/handlers/generate_deref.rs
+++ b/crates/ide-assists/src/handlers/generate_deref.rs
@@ -2,7 +2,9 @@ use hir::{ModPath, ModuleDef};
 use ide_db::{FileId, RootDatabase, famous_defs::FamousDefs};
 use syntax::{
     Edition,
-    ast::{self, AstNode, HasName, edit::AstNodeEdit, syntax_factory::SyntaxFactory},
+    ast::{
+        self, AstNode, HasGenericParams, HasName, edit::AstNodeEdit, syntax_factory::SyntaxFactory,
+    },
     syntax_editor::Position,
 };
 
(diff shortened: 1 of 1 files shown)


A4-Tacks:
It seems that `hir::Type::impls_trait` behavior has changed recently.
Previously, it always returned false for generic types, but now it always returns true

ChayimFriedman2:
Hi @phyBrackets, are you still around?

phyBrackets:
Hi @ChayimFriedman2 @A4-Tacks Sorry for the delay, been busy with other stuffs,  I've rebased the patch onto master and updated the fix,  for generic structs, instead of relying on `impls_trait` (which now returns true for generic types regardless of actual implementations), the fix routes through `has_actual_impl` which checks for concrete impl blocks in the module. 

Let me know if you have any other feedback.

crates/ide-assists/src/handlers/generate_deref.rs
```diff
@@ -211,24 +213,60 @@ fn existing_deref_impl(
...
+            } else {
+                Some(DerefType::Deref)
+            }
+        } else {
+            None
+        }
     }
 }
 
+fn has_actual_impl(
```
ChayimFriedman2, line 248: A much simpler and more correct approach is to list the impls by `Impl::all_for_type(db, strukt.ty(db))`, then filter by `Impl::trait_()`.

ChayimFriedman2: There is also no need to differentiate if the struct has generic args.

ChayimFriedman2 (commented):
I commented on the approach, but honestly I think this is entirely redundant. The behavior of `impls_trait()` is a bug, and we'll fix it (eventually), then there will be no need for this.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/20627/4fac9df3be08a7e6868ab4b8d8c77475691b52a6..d9e108674c1954694045893b1f93fa4a1736478a)*

ChayimFriedman2:
The assist is fixed now, the proper way.

PR #20629, opened 2025-09-08
Add ide-assist: generate_struct

### LINKED ISSUE #8665: Assist: Create struct from unresolved struct literal
See https://intellij-rust.github.io/2021/04/26/changelog-146.html.

A4-Tacks:
Adds a stub struct with a definition matching the expression under the cursor.

Close #8665

Example
---
```rust
fn foo() {
    let bar = Bar$0 { x: 2, y: 3, force: true };
}
```
->
```rust
fn foo() {
    let bar = Bar { x: 2, y: 3, force: true };
}

struct Bar { x: i32, y: i32, force: bool }
```


Veykril:
This should be a diagnostic quickfix, not an assist. Not sure if we have the diagnostics for this yet though.

PR #20635, opened 2025-09-08
fix: Don't trigger two flychecks when saving files that are part of targets

iorizu:
When the user pressed save in a source file that was part of a specific target, rust-analyzer would trigger two flychecks on the workspace that the file was a part of:
- A package-level flycheck (`cargo check -p <package> (--bin | --example | --test | --benchmark) <target>`)
- A workspace-level flycheck (`cargo check --workspace`)

The triggering of the second flycheck cancelled the first, so there was no correctness issue, but the behavior undesirable regardless.

This PR fixes this by making sure that the workspace level check is not triggered if a package check was already triggered.

Closes rust-lang/rust-analyzer#20436

### DIFF
--- a/crates/rust-analyzer/src/flycheck.rs
+++ b/crates/rust-analyzer/src/flycheck.rs
@@ -104,11 +104,11 @@ pub(crate) enum FlycheckConfig {
 }
 
 impl FlycheckConfig {
-    pub(crate) fn invocation_strategy_once(&self) -> bool {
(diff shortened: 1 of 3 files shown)


ShoyuVanilla:
Thanks for working on this! Unfortunately, I had no time to review this for days dealing with other issues on next-solver migration.
I'll review this tomorrow.

iorizu:
Take your time! This is a minor issue, after all.

ShoyuVanilla (approved):
Looks good to me. Thanks!

PR #20638, opened 2025-09-09
Add a FAQ entry about RA and Cargo build lock/cache conflicts

Kobzol:
This was one of the items that people complained about in the recent Compiler Performance Survey. It is "common knowledge" amongst experience RA users, but it wasn't documented in a very visible way before.

I was also thinking about actually making a separate ToC page about this, in the Configuration section, below "Non-Cargo Based Projects". What do you think about that?

### DIFF
--- a/docs/book/README.md
+++ b/docs/book/README.md
@@ -8,6 +8,7 @@ To run the documentation site locally:
 
 ```shell
 cargo install mdbook
+cargo install mdbook-toc
 cargo xtask codegen
 cd docs/book
 mdbook serve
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
Seems reasonable, thanks!

Kobzol:
The change hasn't appeared in the RA book yet. How does it get deployed? :thinking:

Veykril:
I believe we sync on Mondays, but from the looks of it @\lnicola just kicked it off manually so its up now.

Kobzol:
Oh, I didn't realize it's a cron job/manual thing. Thank you! :)

Veykril:
The website repo pulls this repo to sync, and since we tend to do releases on Monday we paired that up there so that people don't see outdated/too new configs in the docs while the changes aren't propagated to the non pre-release

lnicola:
No, it it actually gets published daily. Should be [up](https://rust-analyzer.github.io/book/faq.html) now.

PR #20641, opened 2025-09-09
Add ide-assist: merge_lifetimes

A4-Tacks:
Example
---
```rust
struct Foo<$0'a, 'b$0> {
    data: &'b [Cow<'a, str>],
}
```
->
```rust
struct Foo<'a> {
    data: &'a [Cow<'a, str>],
}
```


Veykril:
I think this is better served by renaming, the assist does not pull its weight

PR #20657, opened 2025-09-12
Fix extra semicolon before else in let-stmt

A4-Tacks:
Fixes rust-lang/rust-analyzer#20620

Example
---
```rust
fn main() {
    let x = if true {
        ()
    } $0 else {};
}
```

**Before this PR**:

```rust
fn main() {
    let x = if true {
        ()
    } else if $1 {
        $0
    }; else {};
}
```

**After this PR**:

```rust
fn main() {
    let x = if true {
        ()
    } else if $1 {
        $0
    } else {};
}
```

### DIFF
--- a/crates/ide-completion/src/completions/keyword.rs
+++ b/crates/ide-completion/src/completions/keyword.rs
@@ -270,6 +270,46 @@ fn main() {
         );
 
         check_edit(
+            "else if",
+            r#"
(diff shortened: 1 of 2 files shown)


lnicola (approved):

lnicola:
changelog fixup #20620

PR #20658, opened 2025-09-12
fix: complete `else` in more expressions

A4-Tacks:
- Support else completion in ArrayExpr, ReturnExpr and PrefixExpr etc
- Support else completion after MatchArm expression

Before this PR, the else branch could not be completed in most expressions

Example
---
```rust
fn foo() -> [i32; 1] {
    [if true {
        2
    } $0]
}
```
->
```rust
fn foo() -> [i32; 1] {
    [if true {
        2
    } else {
        $0
    }]
}
```

---

```rust
fn foo() -> i32 {
    match () {
        () => if true {
            2
        } $0
    }
}
```
->
```rust
fn foo() -> i32 {
    match () {
        () => if true {
            2
        } else {
            $0
        }
    }
}
```

### DIFF
--- a/crates/ide-completion/src/completions/pattern.rs
+++ b/crates/ide-completion/src/completions/pattern.rs
@@ -42,6 +42,11 @@ pub(crate) fn complete_pattern(
         }
     }
 
+    if pattern_ctx.after_if_expr {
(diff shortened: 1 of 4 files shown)


ShoyuVanilla (approved):
Thanks!

PR #20661, opened 2025-09-13
Fix IfExpr then branch suggest

A4-Tacks:
- And add logic operation suggest

Example
---
In the old implementation, it always suggested conditions,
this is a lot of noise, e.g `contract_checks()~(use std::intrinsics::contract_checks) const fn() -> bool`

```rust
fn foo() {
    if true {
        c$0
    }
}
```


ShoyuVanilla (approved):
Thanks!

ShoyuVanilla:
Looks like the tests have failed because of other merged PRs

PR #20673, opened 2025-09-16
Add break value completion support

### LINKED ISSUE #19817: Completion `break` value in `loop`
```rust
fn main() {
    let _: i32 = loop {
        brea$0
    };
}
```

Similar `return $0;` / `return;`

**Expect**:

```rust
fn main() {
    let _: i32 = loop {
        break $0;
    };
}
```

**Currently**:

```rust
fn main() {
    let _: i32 = loop {
        break;
    };
}
```

A4-Tacks:
Close #19817

```rust
fn foo() -> i32 {
    loop {
        $0
    }
}
```

**Before this PR**:

```rust
fn foo() -> i32 {
    loop {
        break;
    }
}
```

**After this PR**:

```rust
fn foo() -> i32 {
    loop {
        break $0;
    }
}
```


crates/ide-completion/src/context/analysis.rs
```diff
@@ -1854,7 +1860,7 @@ fn is_in_token_of_for_loop(path: &ast::Path) -> bool {
     .unwrap_or(false)
 }
 
-fn is_in_breakable(node: &SyntaxNode) -> BreakableKind {
+fn is_in_breakable(node: &SyntaxNode) -> (BreakableKind, Option<SyntaxNode>) {
```
ShoyuVanilla: `(BreakableKind, Option<SyntaxNode>)` feels a bit awkward to me, as `.1` is always `None` for `BreakableKind::None` and otherwise `Some`. How about removing the enum variant `BreakableKind::None` and returning `Option<(BreakableKind, SyntaxNode)>` instead?

A4-Tacks: Okay

PR #20681, opened 2025-09-17
Fix indent for trait_impl_redundant_assoc_item

A4-Tacks:
Example
---
```rust
trait Foo {
}
impl Foo for () {
    $0fn foo<T: Copy>(&self);
}
```

**Before this PR**:

```rust
trait Foo {
    fn foo<T>(&self)
where
    T: Copy,;
}
impl Foo for () {
    fn foo<T: Copy>(&self);
}
```

**After this PR**:

```rust
trait Foo {
    fn foo<T>(&self)
    where
        T: Copy,;
}
impl Foo for () {
    fn foo<T: Copy>(&self);
}
```


crates/stdx/src/lib.rs
```diff
@@ -214,6 +215,29 @@ pub fn trim_indent(mut text: &str) -> String {
         .collect()
 }
 
+/// Indent non empty lines, including the first line
+#[must_use]
+pub fn indent_string<'a, S>(s: S, indent_level: u8) -> Cow<'a, str>
```
ChayimFriedman2: Doesn't need to be generic, please take `&str`.

ChayimFriedman2: Also given this only returns `Cow::Borrowed` when `indent == 0` or `s.is_empty()`, I believe there is no reason to use `Cow`.

PR #20685, opened 2025-09-18
Add multiple generate for enum generate is, as, try_into

### LINKED ISSUE #11227: Assist proposals: generate .(as|is)_variant() for all variants
Would be a continuation of #7604 (and maybe other issues that I could not find).

It would add the possibility to implement the method for all variants at once. It would be made available here (at `$0`):

```
enum My$0Enum {
    FirstVariant,
    OtherVariant,
}
```

because it would be enum-wise, implementing the method for each relevant enum (no as_ if  no data inside) when not implemented yet. Actually, do as if the assist were used manually on each variant, if applicable.

Also, because authors will not know about this feature, and as a suggestion in general, it could be nice to make the assist available where the current assist (for a one-variant generation) is available.

So two assists would be added:

* Generate `is_` methods for all variants
* Generate `as_` methods for all relevant variants

Each assist would be implemented in the same file as the one-variant-wise ones, re-using most of their implementations.

A4-Tacks:
Close #11227

Examples
---
```rust
enum Variant {
    Undefined,
    $0Minor,
    M$0ajor,
}
```
->
```rust
enum Variant {
    Undefined,
    Minor,
    Major,
}

impl Variant {
    /// Returns `true` if the variant is [`Minor`].
    ///
    /// [`Minor`]: Variant::Minor
    #[must_use]
    fn is_minor(&self) -> bool {
        matches!(self, Self::Minor)
    }

    /// Returns `true` if the variant is [`Major`].
    ///
    /// [`Major`]: Variant::Major
    #[must_use]
    fn is_major(&self) -> bool {
        matches!(self, Self::Major)
    }
}
```

---

```rust
enum Value {
    Unit(()),
    $0Number(i32),
    Text(String)$0,
}
```
->
```rust
enum Value {
    Unit(()),
    Number(i32),
    Text(String),
}

impl Value {
    fn try_into_number(self) -> Result<i32, Self> {
        if let Self::Number(v) = self {
            Ok(v)
        } else {
            Err(self)
        }
    }

    fn try_into_text(self) -> Result<String, Self> {
        if let Self::Text(v) = self {
            Ok(v)
        } else {
            Err(self)
        }
    }
}
```

---

```rust
enum Value {
    Unit(()),
    $0Number(i32),
    Text(String)$0,
}
```
->
```rust
enum Value {
    Unit(()),
    Number(i32),
    Text(String),
}

impl Value {
    fn as_number(&self) -> Option<&i32> {
        if let Self::Number(v) = self {
            Some(v)
        } else {
            None
        }
    }

    fn as_text(&self) -> Option<&String> {
        if let Self::Text(v) = self {
            Some(v)
        } else {
            None
        }
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/generate_enum_is_method.rs
+++ b/crates/ide-assists/src/handlers/generate_enum_is_method.rs
@@ -1,10 +1,10 @@
-use std::slice;
-
 use ide_db::assists::GroupLabel;
+use itertools::Itertools;
 use stdx::to_lower_snake_case;
 use syntax::ast::HasVisibility;
 use syntax::ast::{self, AstNode, HasName};
 
+use crate::utils;
 use crate::{
     AssistContext, AssistId, Assists,
     utils::{add_method_to_adt, find_struct_impl},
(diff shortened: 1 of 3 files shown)


ShoyuVanilla (approved):
Looks good to me 👍

PR #20686, opened 2025-09-18
Fix selected applicable generate_default_from_enum_variant

A4-Tacks:


crates/ide-assists/src/handlers/generate_default_from_enum_variant.rs
```diff
@@ -114,6 +117,21 @@ impl Default for Variant {
...
 
+    #[test]
+    fn test_generate_default_not_applicable_with_non_empty_selection() {
+        check_assist_not_applicable(
+            generate_default_from_enum_variant,
+            r#"
+//- minicore: default
+enum Variant {
+    Undefined,
+    $0Minor$0,
```
ShoyuVanilla, line 128: I think we still should be able to generate default for this variant when we have selection over this. Is there some reason that we should refute non-empty selections?

ShoyuVanilla (approved):
Thanks!

PR #20687, opened 2025-09-18
Add ide-assist: merge_nested_else_if

A4-Tacks:
This transforms `if x {} else { if y {} }` into `if x {} else if y {}`,
This assist can only be applied with the cursor on `else`.

Example
---
```rust
fn main() {
    if x == 3 {
        true
    } $0else {
        if x == 2 {
            false
        } else {
            true
        }
    }
}
```
->
```rust
fn main() {
    if x == 3 {
        true
    } else if x == 2 {
        false
    } else {
        true
    }
}
```


Veykril:
This I think is also better served as an on-typing handler maybe, where deletion of the brace should balance out the other one.

PR #20689, opened 2025-09-18
fix: Make flycheck clearing dependency-aware

ShoyuVanilla:
Fixes https://github.com/rust-lang/rust-analyzer/issues/20651

### DIFF
--- a/crates/project-model/src/build_dependencies.rs
+++ b/crates/project-model/src/build_dependencies.rs
@@ -9,7 +9,7 @@
 use std::{cell::RefCell, io, mem, process::Command};
 
 use base_db::Env;
-use cargo_metadata::{Message, camino::Utf8Path};
+use cargo_metadata::{Message, PackageId, camino::Utf8Path};
 use cfg::CfgAtom;
 use itertools::Itertools;
 use la_arena::ArenaMap;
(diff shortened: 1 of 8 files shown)


ShoyuVanilla:
> How large are `PackageId`s? Do they need to be wrapped in an `Arc`?

TBH, I don't think it's that large to be necessarily wrapped in `Arc`s in general, but might be arbitrily large in rare cases if nested in a long filesystem path, as it looks like `"file:///absolute/path/to/my-package#0.1.0"`.
I saw some places in our codebase that wrapping `PackageId` in `Arc`, so just followed it. If you thing `Arc` is redundant, I'm not opposing to removing them.

davidbarsky:
> > How large are `PackageId`s? Do they need to be wrapped in an `Arc`?
> 
> TBH, I don't think it's that large to be necessarily wrapped in `Arc`s in general, but might be arbitrily large in rare cases if nested in a long filesystem path, as it looks like `"file:///absolute/path/to/my-package#0.1.0"`. I saw some places in our codebase that wrapping `PackageId` in `Arc`, so just followed it. If you thing `Arc` is redundant, I'm not opposing to removing them.

Ah, gotcha. Yeah, I guess they’re fine as-is, I just wasn’t sure if they were strings or u32s from the code review alone. This looks good to me!

PR #20695, opened 2025-09-18
minor: Fix one more thing in `in-rust-tree`

ShoyuVanilla:
I guess this is the last thing for the errors like `Foo and Foo have similar names, but are actually distinct types` 🙃 

We might need CI that runs cargo check with `in-rust-tree` feature and check whether we are using `ra_ap_*` things

### DIFF
--- a/crates/hir-ty/src/next_solver/inspect.rs
+++ b/crates/hir-ty/src/next_solver/inspect.rs
@@ -1,4 +1,4 @@
-pub use ra_ap_rustc_next_trait_solver::solve::inspect::*;
+pub use rustc_next_trait_solver::solve::inspect::*;
 
 use rustc_ast_ir::try_visit;
 use rustc_next_trait_solver::{


lnicola (approved):

ShoyuVanilla:
BTW we have several other errors beside this that complaining we are missing a field in a constructor of a type in `rustc_type_ir`

lnicola:
How can that be? Don't tell me `WithCachedTypeInfo` changed in the past hour or so.

lnicola:
Oh, it's gated by `#[cfg(feature = "nightly")]`. We could probably use `Fingerprint::ZERO`, but we can't refer to it.

ShoyuVanilla:
> Oh, it's gated by `#[cfg(feature = "nightly")]`. We could probably use `Fingerprint::ZERO`, but we can't refer to it.

Oh, we're in a tight spot

lnicola:
Nah, it's fine, thanks to your awesome `extern crate rustc_data_structures as ena` hack :sweat_smile:.

PR #20697, opened 2025-09-18
fix negative const generic integer literals

Oblarg:
https://github.com/rust-lang/rust-analyzer/issues/20696

openly admit this was AI-assisted, but it seemed like a sufficiently small/straightforward issue to be a good idea to attempt

### DIFF
--- a/crates/hir-ty/src/lower.rs
+++ b/crates/hir-ty/src/lower.rs
@@ -299,6 +299,29 @@ impl<'a> TyLoweringContext<'a> {
                 const_type,
                 self.resolver.krate(),
             ),
+            hir_def::hir::Expr::UnaryOp { expr: inner_expr, op: hir_def::hir::UnaryOp::Neg } => {
+                if let hir_def::hir::Expr::Literal(literal) = &self.store[*inner_expr] {
+                    // Only handle negation for signed integers and floats
+                    match literal {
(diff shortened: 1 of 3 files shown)


crates/hir-ty/src/lower.rs
```diff
@@ -299,6 +299,29 @@ impl<'a> TyLoweringContext<'a> {
...
+                                unknown_const(const_type)
+                            }
+                        }
+                        // For unsigned integers, chars, bools, etc., negation is not meaningful
+                        _ => unknown_const(const_type),
+                    }
+                } else {
+                    unknown_const(const_type)
+                }
+            }
```
ShoyuVanilla, line 324: Could you duplicate this to the following location as well?

https://github.com/rust-lang/rust-analyzer/blob/50bb3c5d820d03b41088e0d66dd3b9bbc7236c4e/crates/hir-ty/src/lower_nextsolver.rs#L266-L290

ShoyuVanilla (approved):
I believe we should eventually lower const generics with consteval for the bigger picture. That said, this PR doesn’t cause any harm—it introduces some improvements with only small changes. Approving with a nit

crates/hir-ty/src/lower.rs
Oblarg: done :)

lnicola:
changelog fix (first contribution) fix negative literals in const generics

PR #20702, opened 2025-09-19
Fix `else` completion before else keyword

A4-Tacks:
Example
---
```rust
fn foo() {
    let x = if true {
        1
    } el$0 else {
        2
    };
}
```

**Before this PR**:

```text
else~    k [LS]
else if~ k [LS]
```

**After this PR**:

```text
else if~ k [LS]
```


ShoyuVanilla (approved):
Thanks!

PR #20704, opened 2025-09-19
Add ide-assist: replace_if_matches_to_if_let

A4-Tacks:
Replace matches!() to let-chain in if condition.

Example
---
```rust
fn foo(x: Option<i32>) {
    if let Some(n) = x && $0matches!(n.checked_div(2), Some(5..8)) {
    }
}
```
->
```rust
fn foo(x: Option<i32>) {
    if let Some(n) = x && let Some(5..8) = n.checked_div(2) {
    }
}
```


A4-Tacks:

PR #20705, opened 2025-09-20
Add ide-assist: flip_range_expr

A4-Tacks:
Flips operands of a range expression.

Example
---
```rust
fn main() {
    let _ = 90..$02;
}
```
->
```rust
fn main() {
    let _ = 2..90;
}
```
---
```rust
fn main() {
    let _ = 90..$0;
}
```
->
```rust
fn main() {
    let _ = ..90;
}
```


Veykril (approved):
Thanks!

PR #20706, opened 2025-09-20
internal: avoid allocating in `stdx::replace`

A4-Tacks:


crates/stdx/src/lib.rs
```diff
@@ -187,11 +187,19 @@ pub fn is_upper_snake_case(s: &str) -> bool {
...
-    *buf = buf.replace(from, to);
+    let from_len = from.len_utf8();
+    let additional = to.len().saturating_sub(from_len);
+    buf.reserve(additional * replace_count);
+
+    let mut end = buf.len();
+    while let Some(i) = buf[..end].rfind(from) {
+        buf.replace_range(i..i + from_len, to);
+        end = i;
+    }
```
ShoyuVanilla, line 202: The implementation looks great. But seems that we don't have any test for this function so far. Could we have a simple test for this?

ShoyuVanilla (approved):
Thanks!

PR #20708, opened 2025-09-20
Fix panics on `Foo{mut x}` for destructure_struct_binding

A4-Tacks:
Example
---
```rust
struct Foo { x: () }
struct Bar { foo: Foo }
fn f(Bar { mut $0foo }: Bar) {}
```

**Before this PR**:

Panic `Option::unwrap`

**After this PR**:

```rust
struct Foo { x: () }
struct Bar { foo: Foo }
fn f(Bar { foo: Foo { mut x } }: Bar) {}
```


ShoyuVanilla (approved):
Thanks!

PR #20709, opened 2025-09-20
Fix panic `!self.data().mutable` for destructure_struct_binding

A4-Tacks:
When the reference type does not require adding a dereference or parentheses, it will panic

Example
---

```rust
struct Foo { bar: i32, baz: i32 }

fn main() {
    let $0foo = &Foo { bar: 1, baz: 2 };
    let _ = &foo.bar;
}
```

**Before this PR**:

Panic:
```
assertion failed: !self.data().mutable
```

**After this PR**:

```rust
struct Foo { bar: i32, baz: i32 }

fn main() {
    let Foo { bar, baz } = &Foo { bar: 1, baz: 2 };
    let _ = bar;
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/destructure_struct_binding.rs
+++ b/crates/ide-assists/src/handlers/destructure_struct_binding.rs
@@ -288,7 +288,7 @@ fn build_usage_edit(
         Some(field_expr) => Some({
             let field_name: SmolStr = field_expr.name_ref()?.to_string().into();
             let new_field_name = field_names.get(&field_name)?;
-            let new_expr = make.expr_path(ast::make::ext::ident_path(new_field_name));
+            let new_expr = ast::make::expr_path(ast::make::ext::ident_path(new_field_name));
 
             // If struct binding is a reference, we might need to deref field usages
             if data.is_ref {
@@ -298,7 +298,7 @@ fn build_usage_edit(
                     ref_data.wrap_expr(new_expr).syntax().clone_for_update(),
                 )
             } else {
-                (field_expr.syntax().clone(), new_expr.syntax().clone())
+                (field_expr.syntax().clone(), new_expr.syntax().clone_for_update())
             }
         }),
         None => Some((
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/destructure_struct_binding.rs
```diff
@@ -288,7 +288,7 @@ fn build_usage_edit(
         Some(field_expr) => Some({
             let field_name: SmolStr = field_expr.name_ref()?.to_string().into();
             let new_field_name = field_names.get(&field_name)?;
-            let new_expr = make.expr_path(ast::make::ext::ident_path(new_field_name));
+            let new_expr = ast::make::expr_path(ast::make::ext::ident_path(new_field_name));
```
ShoyuVanilla, line 291:
```suggestion
            let new_expr = make.expr_path(make.ident_path(new_field_name));
```
Doesn't this work?

crates/ide-assists/src/handlers/destructure_struct_binding.rs
```diff
@@ -298,7 +298,7 @@ fn build_usage_edit(
                     ref_data.wrap_expr(new_expr).syntax().clone_for_update(),
                 )
             } else {
-                (field_expr.syntax().clone(), new_expr.syntax().clone())
+                (field_expr.syntax().clone(), new_expr.syntax().clone_for_update())
```
ShoyuVanilla, line 301: Though currently we are focused on finishing migration to next-gen trait solver, we are going to get rid of mutable syntax tree usages and that's one of our major long term goals. It would be better if we could avoid calling `.clone_for_update()` here but if it's necessary then okay for now.

crates/ide-assists/src/handlers/destructure_struct_binding.rs
A4-Tacks: > Doesn't this work?

Yes, it dont work

`make.expr_path` return a mutable tree, but `ref_data.wrap_expr(new_expr)` maybe return `new_expr` or a immutable tree

if return `new_expr`, call `clone_for_update` panic for double mutable

And `wrap_expr` is being used by other assists and is extremely difficult to modify

ShoyuVanilla: Okay, as panicking is the worst thing we have here, let's merge this and then migrate later.

PR #20710, opened 2025-09-20
Fix unused_variables fixes shorthand record field

A4-Tacks:
Example
---

```rust
struct S { field : u32 }
fn main() {
    let s = S { field : 2 };
    let S { $0field } = s
}
```

**Before this PR**:

```rust
struct S { field : u32 }
fn main() {
    let s = S { field : 2 };
    let S { _field } = s
}
```

**After this PR**:

```rust
struct S { field : u32 }
fn main() {
    let s = S { field : 2 };
    let S { field: _field } = s
}
```

### DIFF
--- a/crates/ide-diagnostics/src/handlers/unused_variables.rs
+++ b/crates/ide-diagnostics/src/handlers/unused_variables.rs
@@ -6,7 +6,7 @@ use ide_db::{
     label::Label,
     source_change::SourceChange,
 };
-use syntax::{Edition, TextRange};
+use syntax::{AstNode, Edition, TextRange};
 
 use crate::{Diagnostic, DiagnosticCode, DiagnosticsContext};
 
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/unused_variables.rs
```diff
@@ -60,24 +67,23 @@ fn fixes(
...
-            var_name.display(db, edition),
-            var_name.display(db, edition)
-        )),
+        label: Label::new(format!("Rename unused {name} to {new_name}")),
         group: None,
         target: diagnostic_range.range,
         source_change: Some(SourceChange::from_text_edit(
             diagnostic_range.file_id,
-            TextEdit::replace(name_range, format!("_{}", var_name.display(db, edition))),
+            TextEdit::replace(name_range, new_name),
```
ShoyuVanilla, line 86: I think it woud be better if we use `SyntaxEditor` and `SyntaxFactor` APIs here but anyway this doesn't increase the number of mutable syntax tree API calls.

ShoyuVanilla (approved):

PR #20714, opened 2025-09-21
Fix not applicable on trailing comma for remove_dbg

A4-Tacks:
`remove_dbg` not applicable for whitespaces after trailing comma

Example
---
```rust
fn foo() {
    dbg!(
        bar(),
    );
}
```

**Before this PR**:

Assist not applicable

**After this PR**:

```rust
fn foo() {
    bar();
}
```


ShoyuVanilla (approved):
Thanks!

PR #20730, opened 2025-09-23
internal: migrate `expand_record_rest_pattern` assist to `SyntaxEditor`

A4-Tacks:
Because `add_field` uses `ted`


crates/ide-assists/src/handlers/expand_rest_pattern.rs
```diff
@@ -211,7 +210,7 @@ enum Foo {
 fn bar(foo: Foo) {
     match foo {
         Foo::A(_) => false,
-        Foo::B{ y, z  } => true,
+        Foo::B{ y, z } => true,
```
ShoyuVanilla, line 213: Oh, this PR somehow fixes extra spacings here 👍

PR #20731, opened 2025-09-23
fix: handle tuple and slice rest patterns in `expand_rest_pattern`

A4-Tacks:
Assist: expand_tuple_rest_pattern

Fills fields by replacing rest pattern in tuple patterns.

Example
---
```
fn foo(bar: (char, i32, i32)) {
    let (ch, ..$0) = bar;
}
```
->
```
fn foo(bar: (char, i32, i32)) {
    let (ch, _1, _2) = bar;
}
```

---

Assist: expand_slice_rest_pattern

Fills fields by replacing rest pattern in slice patterns.

Example
---
```
fn foo(bar: [i32; 3]) {
    let [first, ..$0] = bar;
}
```
->
```
fn foo(bar: [i32; 3]) {
    let [first, _1, _2] = bar;
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/expand_rest_pattern.rs
+++ b/crates/ide-assists/src/handlers/expand_rest_pattern.rs
@@ -113,9 +113,7 @@ fn expand_tuple_struct_rest_pattern(
     };
 
     let rest_pat = rest_pat.into();
-    let mut pats = pat.fields();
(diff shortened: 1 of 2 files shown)


ShoyuVanilla (approved):
Thanks!

Edit) Oh, we have conflicts 😅

PR #20732, opened 2025-09-23
Add ide-assist: unwrap_destructure_assign

A4-Tacks:
Unwrap the tuple destructuration assignment to different variables.

Close rust-lang/rust-analyzer#20715

Example
---
```rust
fn foo() {
    let (a, b);
    (a, b) $0= (2, 3);
}
```
->
```rust
fn foo() {
    let (a, b);
    a = 2;
    b = 3;
}
```


Veykril:
I don't think this pulls its weight as an assist to be honest.

PR #20736, opened 2025-09-24
Fix applicable on if-let-chain for invert_if

A4-Tacks:
Example
---
```rust
fn f() { i$0f x && let Some(_) = Some(1) { 1 } else { 0 } }
```

**Before this PR**:

```rust
fn f() { if !(x && let Some(_) = Some(1)) { 0 } else { 1 } }
```

**After this PR**:

Assist not applicable


ShoyuVanilla (approved):
Thanks!

PR #20741, opened 2025-09-25
Add macro segment completion

### LINKED ISSUE #11400: Add macro fragment completions
We should show completions for macro fragment types in `macro_rules!` and `macro` matchers.

A4-Tacks:
Close #11400

Example
---
```rust
macro_rules! foo {
    ($($x:$0)*) => ();
}
```

**Completion items**:

```text
ba block
ba expr
ba expr_2021
ba ident
ba item
ba lifetime
ba literal
ba meta
ba pat
ba pat_param
ba path
ba stmt
ba tt
ba ty
ba vis
```


crates/ide-db/src/syntax_helpers/node_ext.rs
```diff
@@ -542,3 +544,33 @@ pub fn macro_call_for_string_token(string: &ast::String) -> Option<MacroCall> {
     let macro_call = string.syntax().parent_ancestors().find_map(ast::MacroCall::cast)?;
     Some(macro_call)
 }
+
+pub fn is_in_macro_matcher(token: &SyntaxToken) -> bool {
+    let Some(macro_def) = token.parent_ancestors().find_map(ast::Macro::cast) else {
```
Veykril: (might have errors)
```suggestion
    let Some(Either::Right(macro_def)) = token.parent_ancestors().map_while(Either<ast::TokenTree, ast::Macro>::cast).last() else {
```
Not really too important but this will short circuit a bit faster I think

crates/ide-completion/src/context/analysis.rs
```diff
@@ -510,6 +510,21 @@ fn analyze<'db>(
...
+                    .filter(|it| !it.kind().is_trivia())
+                    .take(4)
+                    .collect_array()
+                    && dollar.kind() == T![$]
+                    && colon.kind() == T![:]
+                {
+                    CompletionAnalysis::MacroSegment
+                } else {
+                    return None;
+                }
```
Veykril, line 527: Can't this re-use the new function in node-ext or is that filtering too aggressively?

Veykril (approved):
minor question but otherwise looks good to me!

crates/ide-completion/src/context/analysis.rs
A4-Tacks: > an't this re-use the new function in node-ext

New function checks if it is in `($0) => {}` instead of `() => {$0}`

> or is that filtering too aggressively?

For `$x : ty`, filtering whitespaces makes sense

PR #20742, opened 2025-09-25
fix: don't turn unused variables into raw identifier

A4-Tacks:
Example
---
```
fn main() {
    let $0r#type = 2;
}
```

**Before this PR**:

```rust
fn main() {
    let _r#type = 2;
}
```

**After this PR**:

```rust
fn main() {
    let _type = 2;
}
```


ShoyuVanilla (approved):
Thanks!

Issue #16168, opened 2023-12-20
Re-implement `DelimiterKind::Invisible` aka `None` groups 


Veykril:
See https://github.com/rust-lang/rust-analyzer/pull/10623
cc https://github.com/rust-lang/rust-analyzer/issues/12603 https://github.com/rust-lang/rust-analyzer/issues/11521

PR #20745, opened 2025-09-25
Fix negative integer literals in const generics in declarative macro context

Oblarg:
rust-lang/rust-analyzer#20750

These currently are broken due to incorrect parentheses-wrapping behavior in the transcriber

disclaimer: again, this was AI-assisted; but it seems to be local and the AI seems to have found the correct solve


crates/mbe/src/tests.rs
ShoyuVanilla, line 1: Looks good in overall. We should wrap the expressions with `DelimiterKind::Invisible` rather than parentheses but we don't have them now #16168, so this would be the best without much hassle.
But as the current parentheses expansion behavior is a bit hacky and temporary, I think the test should be located in hover or other similar location instead of this mbe crate.

Oblarg: as you wish :)

lnicola:
changelog fixup #20697

PR #20751, opened 2025-09-26
minor: Update memchr to 2.7.6

neuschaefer:
memchr 2.7.6 contains a bugfix for aarch64_be

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1334,9 +1334,9 @@ dependencies = [
 
 [[package]]
 name = "memchr"
-version = "2.7.5"
+version = "2.7.6"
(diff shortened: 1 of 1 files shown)


lnicola:
Should we wait for https://github.com/rust-lang/stdarch/issues/1904 before merging this?

neuschaefer:
> Should we wait for [rust-lang/stdarch#1904](https://github.com/rust-lang/stdarch/issues/1904) before merging this?

I'm quite unsure an I don't have a positive or negative preference

lnicola (approved):
I don't think we have any users on that platform, so sure.

PR #20757, opened 2025-09-27
Use FileId::MAX for id assertion in PathInterner::intern

Shourya742:
Update PathInterner::intern to use FileId::MAX for id range checks, ensuring FileId MAX invariant.

### DIFF
--- a/crates/vfs/src/path_interner.rs
+++ b/crates/vfs/src/path_interner.rs
@@ -28,7 +28,7 @@ impl PathInterner {
     /// - Else, returns a newly allocated id.
     pub(crate) fn intern(&mut self, path: VfsPath) -> FileId {
         let (id, _added) = self.map.insert_full(path);
-        assert!(id < u32::MAX as usize);
+        assert!(id < FileId::MAX as usize);
         FileId(id as u32)
     }
 


bit-aloo:
I am curious why we don't use u32::MAX?

bit-aloo:
cc: @ShoyuVanilla apologies to tag, just looking for a quick review on this.

lnicola:
> I am curious why we don't use u32::MAX?

The file id bits are OR-ed with 8 edition bits and a reserved one and get passed to proc macros as a single `EditionedFileId`.

What I'm wondering is why `EditionedFileId::MAX_FILE_ID` is `0x007F_FFFF` while  `FileId::MAX` is `0x7FFF_FFFF`.

bit-aloo:
> > I am curious why we don't use u32::MAX?
> 
> The file id bits are OR-ed with 8 edition bits and a reserved one and get passed to proc macros as a single `EditionedFileId`.
> 
> What I'm wondering is why `EditionedFileId::MAX_FILE_ID` is `0x007F_FFFF` while `FileId::MAX` is `0x7FFF_FFFF`.

It’s interesting that FileId reserves one bit but doesn’t include any edition bits. Since FileId is expected to accommodate editions as well, the mask 0x7FFF_FFFF appears incomplete.

bit-aloo:
@lnicola makes sense to update the FileId upper limit to 0x007F_FFFF, right? What do you think?

Veykril (approved):
This is definitely more correct, not that we will ever observe issues from this given that r-a likely fails with that many files in memory anyways though. Thanks for finding this!

PR #20758, opened 2025-09-27
Add else-block support for convert_to_guarded_return

A4-Tacks:
Add support for else-block of never-type for `convert_to_guarded_return`

Example
---
```rust
fn main() {
    if$0 let Ok(x) = Err(92) {
        foo(x);
    } else {
        return
    }
}
```

**Before this PR**:

Assist not applicable

**After this PR**:

```rust
fn main() {
    let Ok(x) = Err(92) else {
        return
    };
    foo(x);
}
```


crates/ide-assists/src/handlers/convert_to_guarded_return.rs
```diff
@@ -35,9 +33,7 @@ use crate::{
 // ->
 // ```
 // fn main() {
-//     if !cond {
-//         return;
-//     }
+//     if !cond { return }
```
ShoyuVanilla: Looks good overall but I think we'd prefer the former `return;` stmt over `{ return }` as rustfmt does so

crates/ide-assists/src/handlers/convert_to_guarded_return.rs
```diff
@@ -421,6 +412,51 @@ fn main() {
...
+    #[test]
+    fn convert_if_let_has_never_type_else_block() {
+        check_assist(
+            convert_to_guarded_return,
+            r#"
+fn main() {
+    if$0 let Ok(x) = Err(92) {
+        foo(x);
+    } else {
+        return ;
```
ShoyuVanilla: Would be better without this single space between `return` and `;`

crates/ide-assists/src/handlers/convert_to_guarded_return.rs
A4-Tacks: This will make it more compact in let-chain

A4-Tacks: Okay

A4-Tacks:
@ShoyuVanilla Do you have time to review this PR and some older PRs?

ShoyuVanilla:
Yes, but one or two days later. I have to fix some cargo related issues as they worsen developer experiences for some people by a lot 😅

ShoyuVanilla:
But I'll review some relatively simple PRs before then

PR #20762, opened 2025-09-29
Add fixes for non_exhaustive_let diagnostic

A4-Tacks:
Example
---
```rust
fn foo() {
    let None$0 = Some(5);
}
```

->

```rust
fn foo() {
    let None = Some(5) else { return };
}
```

### DIFF
--- a/crates/ide-diagnostics/src/handlers/non_exhaustive_let.rs
+++ b/crates/ide-diagnostics/src/handlers/non_exhaustive_let.rs
@@ -24,24 +24,30 @@ pub(crate) fn non_exhaustive_let(
     .stable()
     .with_fixes(fixes(&ctx.sema, d))
 }
+
 fn fixes(sema: &Semantics<'_, RootDatabase>, d: &hir::NonExhaustiveLet) -> Option<Vec<Assist>> {
     let root = sema.parse_or_expand(d.pat.file_id);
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/non_exhaustive_let.rs
```diff
@@ -15,11 +22,68 @@ pub(crate) fn non_exhaustive_let(
...
+        return None;
+    }
+
+    let file_id = d.pat.file_id.file_id()?.file_id(sema.db);
+    let semicolon = if let_stmt.semicolon_token().is_none() { ";" } else { "" };
+    let else_block = format!(" else {{ {early_text} }}{semicolon}");
+    let insert_offset = let_stmt
+        .semicolon_token()
+        .map(|it| it.text_range().start())
+        .unwrap_or_else(|| let_stmt.syntax().text_range().end());
```
Veykril: This offset might be within a macro expansion due to the above `d.pat.value.to_node(&root)` call. So we need to upmap things out of macro expansions here to accommodate for that

A4-Tacks: Okay

A4-Tacks:
Can this PR be merged?

PR #20764, opened 2025-09-30
Fix let-expr in lhs for convert_to_guarded_return

A4-Tacks:
Example
---
```rust
fn main() {
    if$0 let Ok(x) = Err(92)
        && let Ok(y) = Ok(37)
        && x < 30
        && let Some(y) = Some(8)
    {
        foo(x, y);
    }
}
```

**Before this PR**:

```rust
fn main() {
    let Ok(x) = Err(92) else { return };
    if !(let Ok(y) = Ok(37) && x < 30) { return }
    let Some(y) = Some(8) else { return };
    foo(x, y);
}
```

**After this PR**:

```rust
fn main() {
    let Ok(x) = Err(92) else { return };
    let Ok(y) = Ok(37) else { return };
    if x >= 30 { return }
    let Some(y) = Some(8) else { return };
    foo(x, y);
}
```


Veykril (approved):
thanks!

PR #20771, opened 2025-10-01
fix: offer `invert_if` on `else`

A4-Tacks:
Example
---
```rust
fn f() {
    if cond {
        3 * 2
    } e$0lse {
        1
    }
}
```
->
```rust
fn f() {
    if !cond {
        1
    } else {
        3 * 2
    }
}
```


ChayimFriedman2 (approved):
I'm not sure it's better, but it's also not particularly worse.

PR #20781, opened 2025-10-02
Add ide-assist: flip_else_if

A4-Tacks:
Flips two if-else-if branches.

Example
---
```rust
fn foo() {
    let n = if cond1 { 2 } else$0 if cond2 { 3 } else { 0 };
}
```
->
```rust
fn foo() {
    let n = if cond2 { 3 } else if cond1 { 2 } else { 0 };
}
```


A4-Tacks:

PR #20788, opened 2025-10-03
fix: allow more string literal conversions from raw strings

A4-Tacks:
Example
---

Assist: `make_raw_string`

```rust
fn f() {
    let s = $0b"random\nstring";
}
```
->
```rust
fn f() {
    let s = br#"random
string"#;
}
```

---

Assist: `make_raw_string`

```rust
fn f() {
    let s = $0c"random\nstring";
}
```
->
```rust
fn f() {
    let s = cr#"random
string"#;
}
```

---

Assist: `add_hash`

```rust
fn f() {
    let s = $0cr"random string";
}
```
->
```rust
fn f() {
    let s = cr#"random string"#;
}
```

---

Assist: `remove_hash`

```rust
fn f() {
    let s = $0cr#"random string"#;
}
```
->
```rust
fn f() {
    let s = cr"random string";
}
```

---

Assist: `make_usual_string`

```rust
fn f() {
    let s = $0cr#"random string"#;
}
```
->
```rust
fn f() {
    let s = c"random string";
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/raw_string.rs
+++ b/crates/ide-assists/src/handlers/raw_string.rs
@@ -1,6 +1,8 @@
-use std::borrow::Cow;
-
-use syntax::{AstToken, TextRange, TextSize, ast, ast::IsString};
+use ide_db::source_change::SourceChangeBuilder;
+use syntax::{
+    AstToken,
+    ast::{self, IsString, make::tokens::literal},
+};
 
 use crate::{
     AssistContext, AssistId, Assists,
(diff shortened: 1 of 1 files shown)


A4-Tacks:
The current implementation can replace quotation marks separately instead of string tokens

Using `SyntaxEditor` requires replacing the string token

ShoyuVanilla:
Yeah, but we should eventually get rid of mutable syntax tree usages. Though this PR achieves solid improvement without much changes but the added lines won't directly work with the new API so I think it would be better to migrate it as well

PR #20803, opened 2025-10-05
Replace `--show-output` with `--nocapture`

SomeoneToIgnore:
The former does not show any output before the test is finished, which prevents long-running/stuck tests from showing any useful information.

Example: https://github.com/zed-industries/zed/blob/5b0a2f1ab66aec068a837e4d66612ba92b59a29c/crates/project/src/project_tests.rs#L8843-L8845

This is a test that repeats 10 times, each time printing the iteration number.

https://github.com/user-attachments/assets/48237692-c5ad-43c7-91a5-ad969fbc0aff


ChayimFriedman2:
There was a discussion about this in the past, I need to find it. Without inspecting the points raised in that discussion we cannot proceed with this change.

ChayimFriedman2:
Also note that you can customize this.

ChayimFriedman2:
Found it: https://github.com/rust-lang/rust-analyzer/issues/12737.

SomeoneToIgnore:
Yeah, no worries and feel free to close if not needed.

My main motivation is to have a default that produces normal output and does not require me to re-confgure every machine's rust-analyzer.

ChayimFriedman2:
I'm not rejecting this. I'm just saying that if you want this, you need to read and summarize that issue, answer the points raised there, and then we'll discuss that in the team and decide.

SomeoneToIgnore:
I did an offline discussion with @Veykril (seemed onboard as experiences the same chores) and fully agree with https://github.com/rust-lang/rust-analyzer/issues/12737#issuecomment-2559549945

The only other useful bit of information is https://github.com/rust-lang/rust-analyzer/issues/12737#issuecomment-2560145428 and that does not contradict my stance it seems.

I'll leave the rest of the discussion on you.

ChayimFriedman2:
CC @rust-lang/rust-analyzer.

My personal opinion: I slightly prefer `--nocapture` (once a while I need to debug a segfaulting/hanging/... test then I copy the test command and run it in the terminal and change `--show-output` to `--nocapture`) but not strongly, not even enough that I'll care to change the default. I also don't suffer from interleaved output because I don't run more than one test at a time via rust-analyzer.

Veykril:
Imo we should've reverted to `--nocapture` a long time ago. It is just so annoying when you have a test that hangs or hard crashes not produce output. And I don't think I've *ever* run into interleaved output issues either.

lnicola (approved):

lnicola:
Agreed, especially in the `--exact` case.

ChayimFriedman2:
This caused us to lose colors for the test harness, which is sad.

SomeoneToIgnore:
So, nothing ever works with cargo and r-a again, huh?

What if the r-a part gets more clever and runs multiple tests with the old flag, and a single one with `--nocapture`?

Or do you know the magic flag to enable colored output on top?

Veykril:
I wonder if we can just pass both? Would that be too noisy?

SomeoneToIgnore:
Seems that passing both does not help with the coloring:

image

PR #20822, opened 2025-10-09
minor: update typos checker version

Young-Flash:


.typos.toml
```diff
@@ -32,6 +32,7 @@ makro = "makro"
 trivias = "trivias"
 thir = "thir"
 jod = "jod"
+tructure = "tructure"
```
lnicola, line 35: Bah, an entire exception for an acronym used twice.

PR #20833, opened 2025-10-12
Add ide-assist: move_for_binder

A4-Tacks:
Move for-binder to where predicate.

Example
---
```rust
fn foo<F>()
where
    F: $0for<'a> Fn(&'a str)
{}
```
->
```rust
fn foo<F>()
where
    for<'a> F: Fn(&'a str)
{}
```


A4-Tacks:
> This seems fairly niche to me

This is indeed fairly niche

PR #20837, opened 2025-10-13
Add an Extension Config API

michaelgruenewald:
This adds an exported API for other extensions to add rust-analyzer-specific configurations for rust-lang/rust-analyzer#18476. It provides a simple API where an extension can just call something like

    const rustAnalyzer = await vscode.extensions.getExtension("rust-lang.rust-analyzer")?.activate();
    await rustAnalyzer?.addConfiguration(ctx.extension.id, {
        "cargo.cfgs": ["foo", "bar"],
    });

to configure aspects. The extension configurations are persisted in the workspace, but only configurations from extensions are are actually present are used, so that uninstalled extensions, etc. aren't causing any trouble.

This does pull in _lodash_ as a dependency, but IMO this is neither very heavy nor critical, and it was already in the dev closure anyway.

Also cf. the discussion in https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/VSCode.20extension.20config.20feature.20contribution

Fixes rust-lang/rust-analyzer#18476

### DIFF
--- a/editors/code/src/config.ts
+++ b/editors/code/src/config.ts
@@ -31,7 +31,7 @@ export class Config {
     workspaceState: vscode.Memento;
 
     private readonly rootSection = "rust-analyzer";
-    private readonly requiresServerReloadOpts = ["cargo", "server", "files", "showSyntaxTree"].map(
+    private readonly requiresServerReloadOpts = ["server", "files", "showSyntaxTree"].map(
         (opt) => `${this.rootSection}.${opt}`,
     );
 


editors/code/src/config.ts
```diff
@@ -4,31 +4,48 @@ import * as path from "path";
...
 export class Config {
     readonly extensionId = "rust-lang.rust-analyzer";
+
     configureLang: vscode.Disposable | undefined;
+    workspaceState: vscode.Memento;
 
-    readonly rootSection = "rust-analyzer";
-    private readonly requiresServerReloadOpts = ["server", "files", "showSyntaxTree"].map(
+    private readonly rootSection = "rust-analyzer";
+    private readonly requiresServerReloadOpts = ["cargo", "server", "files", "showSyntaxTree"].map(
```
Veykril: Why do we need this change?

michaelgruenewald: Ah, good catch, that was for an unrelated issue.

lnicola:
changelog feat (first contribution) expose an `addConfiguration` API from the Code extension

PR #20844, opened 2025-10-15
fix: offer `add_braces` on assignments

A4-Tacks:
```rust
fn foo() {
    let x =$0 n + 100;
}
```
```rust
fn foo() {
    let x = {
        n + 100
    };
}
```


A4-Tacks:
Okay, I rebased it to resolve the conflict of merging

PR #20864, opened 2025-10-19
Fix extract function invalid self param

A4-Tacks:
Example
---
```rust
trait Foo {
    fn f(&self) -> i32;

    fn foo(&self) -> i32 {
        $0self.f()+self.f()$0
    }
}
```

**Before this PR**

```rust
trait Foo {
    fn f(&self) -> i32;

    fn foo(&self) -> i32 {
        fun_name(self)
    }
}

fn $0fun_name(&self) -> i32 {
    self.f()+self.f()
}
```

**After this PR**

```rust
trait Foo {
    fn f(&self) -> i32;

    fn foo(&self) -> i32 {
        fun_name(self)
    }
}

fn $0fun_name(this: &impl Foo) -> i32 {
    this.f()+this.f()
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/extract_function.rs
+++ b/crates/ide-assists/src/handlers/extract_function.rs
@@ -1742,23 +1742,25 @@ impl<'db> Function<'db> {
         module: hir::Module,
         edition: Edition,
     ) -> ast::ParamList {
-        let this_param = self.make_this_param();
+        let this_param = self.make_this_param().map(|f| f());
         let self_param = self.self_param.clone().filter(|_| this_param.is_none());
         let params = self.params.iter().map(|param| param.to_param(ctx, module, edition));
         make::param_list(self_param, this_param.into_iter().chain(params))
     }
 
-    fn make_this_param(&self) -> Option<ast::Param> {
+    fn make_this_param(&self) -> Option<impl FnOnce() -> ast::Param> {
         if let Some(name) = self.mods.trait_name.clone()
             && let Some(self_param) = &self.self_param
         {
-            let bounds = make::type_bound_list([make::type_bound(name)]);
-            let pat = make::path_pat(make::ext::ident_path("this"));
-            let mut ty = make::impl_trait_type(bounds.unwrap()).into();
-            if self_param.amp_token().is_some() {
-                ty = make::ty_ref(ty, self_param.mut_token().is_some());
-            }
-            Some(make::param(pat, ty))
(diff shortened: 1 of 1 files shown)


flodiebold:
I'm not sure adding the method to the trait is usually going to be what the user wants?

A4-Tacks:
> I'm not sure adding the method to the trait is usually going to be what the user wants?

But it may be appropriate to do so instead of creating a generic function with all trait bounds and trait generic parameters

Veykril:
I think in this case, when the selection contains `self` locals, we might want to offer both `extract to function` and `extract to trait method`? I do agree we can't assume that the user wants a trait function in this case

A4-Tacks:
Okay, I have implemented the conversion of a simple case from `self` to `this`

crates/ide-assists/src/handlers/extract_function.rs
```diff
@@ -924,6 +928,10 @@ impl FunctionBody {
             false
         };
 
+        // FIXME: make trait arguments
+        let trait_name = ast::Trait::cast(insert_after.clone())
```
ChayimFriedman2: That seems brittle. There is no inherent correspondence between the insert_after node and the trait. Rather, pass the trait directly (or its name).

crates/ide-assists/src/handlers/extract_function.rs
```diff
@@ -1419,14 +1428,18 @@ fn fixup_call_site(builder: &mut SourceChangeBuilder, body: &FunctionBody) {
 fn make_call(ctx: &AssistContext<'_>, fun: &Function<'_>, indent: IndentLevel) -> SyntaxNode {
     let ret_ty = fun.return_type(ctx);
 
-    let args = make::arg_list(fun.params.iter().map(|param| param.to_arg(ctx, fun.mods.edition)));
     let name = fun.name.clone();
-    let mut call_expr = if fun.self_param.is_some() {
+    let args = fun.params.iter().map(|param| param.to_arg(ctx, fun.mods.edition));
+    let mut call_expr = if fun.make_this_param().is_some() {
```
ChayimFriedman2, line 1433: `make_this_param()` actually creates the param, that's redundant. Instead have `has_this_param()` or something.

A4-Tacks: A single function facilitates the maintenance of unified judgment logic, and I return a closure to solve redundant `make`

PR #20875, opened 2025-10-20
fix: Fix `signature_help` to proto conversion creating wrong offsetes

Veykril:


Veykril:
nevermind this is right, lsp strikes again where these are utf16 ...

PR #20876, opened 2025-10-20
fix: fix `signature_help` LSP conversion creating invalid UTF-16 offsets

Veykril:


ShoyuVanilla:
Was this found through the newly released Windows build of Zed? 😄

Veykril:
No we added new asserts which caught this, and then I realized r-a assumes characters are 2 bytes wide at most (which is wrong)

PR #20896, opened 2025-10-23
Remove hir-ty/src/next_solver/mapping.rs

smmalis37:
It's no longer needed, now that chalk is gone.


crates/hir-ty/src/db.rs
```diff
@@ -322,7 +322,7 @@ pub struct InternedConstParamId {
 #[salsa_macros::interned(no_lifetime, debug, revisions = usize::MAX)]
 #[derive(PartialOrd, Ord)]
 pub struct InternedOpaqueTyId {
-    pub loc: ImplTraitId,
+    pub loc: ImplTraitId<'db>,
```
ChayimFriedman2, line 325: Oh we can use `'db` here?!

I would prefer not to rely on this, as this seems to be an accidental feature. Instead, remove `no_lifetime` and introduce a lifetime to this struct.

smmalis37: I didn't expect it to work, but rustc said "use <'db> here" so i did and it worked.

smmalis37: salsa::interned doesn't seem to like me adding lifetimes:
image

ChayimFriedman2: Let's merge this then, as it's an improvement over the status quo.

smmalis37: Also adding a lifetime to InternedOpaqueTyId propagates up to SolverDefId, and farther.

lnicola:
changelog fixup #20329

PR #20897, opened 2025-10-24
Add a lot more db lifetimes

smmalis37:
As per the request on rust-lang/rust-analyzer#20896, I removed `no_lifetime` and added a 'db to InternedOpaqueTyId. This propagated way farther than I expected, but I followed them all. The macro tweaks were the annoying part, as I had to work around the differences between declaring a type in a signature and calling a method on the type. But it all seems to work.

### DIFF
--- a/crates/hir-ty/src/next_solver/def_id.rs
+++ b/crates/hir-ty/src/next_solver/def_id.rs
@@ -5,7 +5,6 @@ use hir_def::{
     GeneralConstId, GenericDefId, ImplId, StaticId, StructId, TraitId, TypeAliasId, UnionId,
 };
 use rustc_type_ir::inherent;
-use stdx::impl_from;
 
 use crate::db::{InternedClosureId, InternedCoroutineId, InternedOpaqueTyId};
 
@@ -18,7 +17,7 @@ pub enum Ctor {
 }
 
 #[derive(PartialOrd, Ord, Clone, Copy, PartialEq, Eq, Hash, salsa::Supertype)]
-pub enum SolverDefId {
+pub enum SolverDefId<'db> {
     AdtId(AdtId),
     ConstId(ConstId),
     FunctionId(FunctionId),
@@ -28,13 +27,13 @@ pub enum SolverDefId {
     TypeAliasId(TypeAliasId),
     InternedClosureId(InternedClosureId),
     InternedCoroutineId(InternedCoroutineId),
-    InternedOpaqueTyId(InternedOpaqueTyId),
+    InternedOpaqueTyId(InternedOpaqueTyId<'db>),
     EnumVariantId(EnumVariantId),
     // FIXME(next-solver): Do we need the separation of `Ctor`? It duplicates some variants.
     Ctor(Ctor),
 }
 
-impl std::fmt::Debug for SolverDefId {
+impl<'db> std::fmt::Debug for SolverDefId<'db> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let interner = DbInterner::conjure();
         let db = interner.db;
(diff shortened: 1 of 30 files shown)


crates/hir-ty/src/next_solver/def_id.rs
```diff
@@ -102,23 +101,83 @@ impl std::fmt::Debug for SolverDefId {
...
-    TraitId,
-    TypeAliasId,
-    InternedClosureId,
-    InternedCoroutineId,
-    InternedOpaqueTyId,
-    EnumVariantId,
-    Ctor
-    for SolverDefId
-);
+impl<'db> From<AdtId> for SolverDefId<'db> {
```
smmalis37, line 104: impl_from! doesn't work with generics unfortunately.

crates/query-group-macro/src/queries.rs
```diff
@@ -282,11 +282,16 @@ impl ToTokens for Intern {
...
+        self.interned_struct_path.leading_colon.to_tokens(&mut type_constructor);
+        for segment in &self.interned_struct_path.segments {
+            segment.ident.to_tokens(&mut type_constructor);
+            quote! {::}.to_tokens(&mut type_constructor);
+        }
 
         let method = quote! {
             #sig {
-                #wrapper_struct::new(self, #interned_pat)
+                #type_constructor new(self, #interned_pat)
```
smmalis37, line 294: Previously, when given the type `InternedOpaqueTyId<'db>` this would generate `InternedOpaqueTyId<'db>::new`, which is invalid syntax. Now it removes all the generics, leaving them to be inferred, producing just `InternedOpaqueTyId::new`. The same applies below.

crates/query-group-macro/src/queries.rs
```diff
@@ -314,8 +320,10 @@ impl Lookup {
         let interned_pat = ty.first().expect("at least one pat; this is a bug");
         let interned_return_ty = &interned_pat.ty;
 
+        let FnArg::Receiver(receiver) = sig.inputs.first().unwrap() else { unreachable!() };
+
         self.signature = parse_quote!(
-            fn #ident(&self, id: #interned_key) -> #interned_return_ty
+            fn #ident #generics(#receiver, id: #interned_key) -> #interned_return_ty
```
smmalis37, line 326: This &self needs the lifetime now too, hence this change.

ChayimFriedman2:
I know I've asked for this, but after seeing the diff I'm not sure I prefer that 😅

crates/hir-ty/src/next_solver/opaques.rs
```diff
@@ -13,7 +13,7 @@ interned_vec_db!(PredefinedOpaques, PredefinedOpaque);
 pub type ExternalConstraintsData<'db> =
     rustc_type_ir::solve::ExternalConstraintsData<DbInterner<'db>>;
 
-interned_vec_nolifetime_salsa!(SolverDefIds, SolverDefId);
+interned_vec_nolifetime_salsa!(SolverDefIds, SolverDefId<'db>);
```
smmalis37, line 16: Does this need more lifetime propagation too? The macro is named nolifetime but it seems to work fine despite the lifetime...

smmalis37: I guess other uses of this macro already had lifetimes

Veykril: I think its about the declaration it creates, not the things it consumes

Veykril:
Would love to get the ball rolling for the lifetime changes, could you rebase this if you get the time? Happy to merge it then

ChayimFriedman2:
We didn't get OP's response for some months and it's going to be pretty hard to rebase this now. Closing.

PR #20901, opened 2025-10-24
references.excludeTests to exclude #[cfg(test)]-gated refs

krobelus:
The references.excludeTests setting fails to exclude references from test code in various scenarios.
For example when using textDocument/references on "foo" in

	pub fn foo() {}
	#[cfg(test)]
	mod tests {
		use crate::foo; // Should be excluded
	}

same when the test module lives in a different file:

	#[cfg(test)]
	mod tests;

Try to fix that for most real-world cases.

TODO: the unit test does not actually test this because whereas
full rust-analyzer finds references in code that's disabled due to
#[cfg(test)] directives, the test logic doesn't; "ModuleData::children"
is always empty. Would appreciate help with that.

changelog fix

Closes rust-lang/rust-analyzer#18573

### DIFF
--- a/crates/ide-db/src/search.rs
+++ b/crates/ide-db/src/search.rs
@@ -1370,13 +1370,53 @@ fn is_name_ref_in_import(name_ref: &ast::NameRef) -> bool {
 }
 
 fn is_name_ref_in_test(sema: &Semantics<'_, RootDatabase>, name_ref: &ast::NameRef) -> bool {
-    name_ref.syntax().ancestors().any(|node|is_test_function(sema, &node))
+    name_ref
+        .syntax()
+        .ancestors()
+        .any(|node| is_test_function(sema, &node) || node_has_cfg_test(sema, &node))
+        || current_module_is_declared_in_cfg_test(sema, name_ref)
 }
 
 /// Returns true if the node is a function with the `#[test]` attribute.
 fn is_test_function(sema: &Semantics<'_, RootDatabase>, node: &syntax::SyntaxNode) -> bool {
     ast::Fn::cast(node.clone())
         .is_some_and(|func| sema.to_def(&func).is_some_and(|func| func.is_test(sema.db)))
+}
+
+/// Returns true if the node is only enabled in test configurations.
+fn node_has_cfg_test(sema: &Semantics<'_, RootDatabase>, node: &syntax::SyntaxNode) -> bool {
+    macro_rules! attrs {
+        ($it:expr) => {
+            sema.to_def($it).map(|node| node.attrs(sema.db))
+        };
+    }
+    use ast::*;
+    let attrs = match_ast! {
+        match node {
+            Adt(it) => attrs!(&it),
+            Const(it) => attrs!(&it),
+            ConstParam(it) => attrs!(&it),
+            Enum(it) => attrs!(&it),
+            ExternCrate(it) => attrs!(&it),
+            Fn(it) => attrs!(&it),
+            GenericParam(it) => attrs!(&it),
+            Impl(it) => attrs!(&it),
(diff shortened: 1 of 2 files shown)


crates/ide-db/src/search.rs
```diff
@@ -1370,8 +1370,77 @@ fn is_name_ref_in_import(name_ref: &ast::NameRef) -> bool {
...
+/// Returns true if the given attributes enable code only in test configurations.
+pub fn has_cfg_test(attrs: &AttrsWithOwner) -> bool {
+    fn is_cfg_test(cfg_expr: &CfgExpr) -> bool {
+        use CfgExpr::*;
+        use cfg::CfgAtom;
+        match cfg_expr {
+            Atom(CfgAtom::Flag(flag)) => *flag == sym::test,
+            All(exprs) => exprs.iter().any(is_cfg_test),
+            // N.B. possible false negatives here.
+            Invalid | Atom(CfgAtom::KeyValue { .. }) | Any(_) | Not(_) => false,
```
danielrainer, line 1431: Maybe something similar to `CfgExpr::fold` can be used for more complex cases, e.g.:

```diff
diff --git a/crates/ide-db/src/search.rs b/crates/ide-db/src/search.rs
index 41a2ea8fe0..ad3af99fb9 100644
--- a/crates/ide-db/src/search.rs
+++ b/crates/ide-db/src/search.rs
@@ -1421,17 +1421,26 @@
 
 /// Returns true if the given attributes enable code only in test configurations.
 pub fn has_cfg_test(attrs: &AttrsWithOwner) -> bool {
-    fn is_cfg_test(cfg_expr: &CfgExpr) -> bool {
+    fn is_cfg_test(cfg_expr: &CfgExpr) -> Option<bool> {
         use CfgExpr::*;
         use cfg::CfgAtom;
         match cfg_expr {
-            Atom(CfgAtom::Flag(flag)) => *flag == sym::test,
-            All(exprs) => exprs.iter().any(is_cfg_test),
-            // N.B. possible false negatives here.
-            Invalid | Atom(CfgAtom::KeyValue { .. }) | Any(_) | Not(_) => false,
+            Invalid => None,
+            Atom(CfgAtom::Flag(flag)) => Some(*flag == sym::test),
+            Atom(CfgAtom::KeyValue { .. }) => Some(false),
+            All(exprs) => {
+                exprs.iter().try_fold(false, |acc, expr| is_cfg_test(expr).map(|res| acc || res))
+            }
+            Any(exprs) => {
+                exprs.iter().try_fold(true, |acc, expr| is_cfg_test(expr).map(|res| acc && res))
+            }
+            Not(expr) => is_cfg_test(expr).map(|is_test| !is_test),
         }
     }
-    attrs.cfgs().any(|cfg_expr| is_cfg_test(&cfg_expr))
+    attrs
+        .cfgs()
+        .try_fold(false, |acc, cfg_expr| is_cfg_test(&cfg_expr).map(|is_test| acc || is_test))
+        .unwrap_or(false)
 }
 
 /// Returns true if this reference's enclosing module is declared conditional on `cfg(test)`.
```

I noticed that expressions annotated with `#[cfg(test)]` are not handled by this PR, and it seems that the approach used for other `ast` types does not work for `ast::Expr`. Does the `ast::HasAttrs` trait correspond to the types which can be annotated with `#[cfg(test)]`? If so, maybe the implementation can be based around that. It might be unnecessary to handle some of these types, since I don't know why anyone would put `#[cfg(test)]` on something like an expression.

ChayimFriedman2:
@krobelus Are you still interested in this?

ChayimFriedman2:
No answer, closing. @krobelus feel free to return to this anytime.

PR #20905, opened 2025-10-24
fix: fix array inhabitedness check

ShoyuVanilla:
Fixes https://github.com/rust-lang/rust-analyzer/issues/18334

### DIFF
--- a/crates/hir-ty/src/inhabitedness.rs
+++ b/crates/hir-ty/src/inhabitedness.rs
@@ -90,7 +90,7 @@ impl<'db> TypeVisitor<DbInterner<'db>> for UninhabitedFrom<'_, 'db> {
             TyKind::Tuple(..) => ty.super_visit_with(self),
             TyKind::Array(item_ty, len) => match try_const_usize(self.infcx.interner.db, len) {
                 Some(0) | None => CONTINUE_OPAQUELY_INHABITED,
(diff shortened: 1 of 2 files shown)


crates/hir-ty/src/inhabitedness.rs
```diff
@@ -90,7 +90,7 @@ impl<'db> TypeVisitor<DbInterner<'db>> for UninhabitedFrom<'_, 'db> {
             TyKind::Tuple(..) => ty.super_visit_with(self),
```
ShoyuVanilla, line 90: Tuple should be super-visited, as super-visit iterates over its field types

crates/hir-ty/src/inhabitedness.rs
```diff
@@ -90,7 +90,7 @@ impl<'db> TypeVisitor<DbInterner<'db>> for UninhabitedFrom<'_, 'db> {
...
             TyKind::Array(item_ty, len) => match try_const_usize(self.infcx.interner.db, len) {
                 Some(0) | None => CONTINUE_OPAQUELY_INHABITED,
-                Some(1..) => item_ty.super_visit_with(self),
+                Some(1..) => item_ty.visit_with(self),
```
ShoyuVanilla, line 93: But `item_ty` here should be directly visited

PR #20909, opened 2025-10-24
minor: Rename `downcast_[ref|mut]_unchecked` -> `downcast_unchecked_[ref|mut]`

taj-p:
An accompanying PR to https://github.com/rust-lang/rust/pull/148079 which keeps these Rust Analyzer names consistent with `core`.

---

## Intent

Renames `downcast_[ref|mut]_unchecked` to `downcast_unchecked_[ref|mut]` because we want to emphasise that it is the downcast that is unsafe, not the aliasing per:

https://github.com/rust-lang/rust/issues/90850#issuecomment-2749035832

## Tracking Issue: https://github.com/rust-lang/rust/issues/90850#issuecomment-3437027607

cc @marc0246

### DIFF
--- a/crates/stdx/src/anymap.rs
+++ b/crates/stdx/src/anymap.rs
@@ -117,7 +117,7 @@ impl<A: ?Sized + Downcast> Map<A> {
     #[inline]
     #[must_use]
     pub fn get<T: IntoBox<A>>(&self) -> Option<&T> {
-        self.raw.get(&TypeId::of::<T>()).map(|any| unsafe { any.downcast_ref_unchecked::<T>() })
+        self.raw.get(&TypeId::of::<T>()).map(|any| unsafe { any.downcast_unchecked_ref::<T>() })
     }
 
     /// Gets the entry for the given type in the collection for in-place manipulation
@@ -172,7 +172,7 @@ impl<'map, A: ?Sized + Downcast, V: IntoBox<A>> OccupiedEntry<'map, A, V> {
     #[inline]
     #[must_use]
     pub fn into_mut(self) -> &'map mut V {
-        unsafe { self.inner.into_mut().downcast_mut_unchecked() }
+        unsafe { self.inner.into_mut().downcast_unchecked_mut() }
     }
 }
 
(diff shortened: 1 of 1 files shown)


bench_data/numerous_macro_rules
```diff
@@ -315,7 +315,7 @@ macro_rules! __ra_macro_fixture313 {($name :  ident , $($s :  expr , $o :  expr
...
+macro_rules! __ra_macro_fixture317 {($base :  ident , $(+ $bounds :  ident )*)=>{ impl  fmt ::  Debug  for $base $(+ $bounds )* {# [ inline ] fn  fmt (&  self ,  f : &  mut  fmt ::  Formatter )->  fmt ::  Result { f .  pad ( stringify ! ($base $(+ $bounds )*))}} impl  UncheckedAnyExt  for $base $(+ $bounds )* {# [ inline ] unsafe  fn  downcast_unchecked_ref <  T : 'static > (&  self )-> &  T {&* ( self  as *  const  Self  as *  const  T )}# [ inline ] unsafe  fn  downcast_unchecked_mut <  T : 'static > (&  mut  self )-> &  mut  T {&  mut * ( self  as *  mut  Self  as *  mut  T )}# [ inline ] unsafe  fn  downcast_unchecked <  T : 'static > ( self :  Box <  Self >)->  Box <  T > { Box ::  from_raw ( Box ::  into_raw ( self ) as *  mut  T )}} impl <  T : $base $(+ $bounds )*>  IntoBox <$base $(+ $bounds )*>  for  T {# [ inline ] fn  into_box ( self )->  Box <$base $(+ $bounds )*> { Box ::  new ( self )}}}}
```
lnicola: This is just a test file, we don't need to actually change it.

lnicola (approved):

ShoyuVanilla:
I think this could be merged only after at least approved on rust-lang/rust. Though it's just an inherent method name change in rust-analyzer side but keeping them synced with `std` feels more comfortable as the crate name here is `stdx` 😄

bench_data/numerous_macro_rules
taj-p: Removed in https://github.com/rust-lang/rust-analyzer/pull/20909/commits/291cc5d2c7e635c06e08b8734855377223b7d842 (as a amendment to the existing commit)

PR #20926, opened 2025-10-28
Add ide-assist: convert_to_format_string

A4-Tacks:
This assist overlaps with the postfix completion `.format`,
difference is that this assist only converts one curly, and all other curls will be escaped

Example
---
```rust
fn foo() {
    let n = 2;
    let s = "n: {n$0}";
}
```
->
```rust
fn foo() {
    let n = 2;
    let s = format!("n: {n}");
}
```


Veykril:
We already have a postfix completion snippet for this which does the job

PR #20933, opened 2025-10-29
Fix range expr expected type suggest

A4-Tacks:
Example
---
```rust
enum State { Stop }
fn bar(x: core::ops::Range<State>) {}
fn foo() {
    bar(State::Stop..$0)
}
```

**Before this PR**

```text
std::ops::Range {…}~ std::ops::Range { start: Idx, end: Idx }
```

**After this PR**

```text
State::Stop~ State::Stop
```


ChayimFriedman2:
I'm closing this since we established than we need a more comprehensive approach for inferring the expected type than replicating inference.

PR #20937, opened 2025-10-30
fix: don't suggest duplicate `const` completions after `raw`

A4-Tacks:
Example
---
```rust
fn main() { let _ = &raw $0 }
```

**Before this PR**

```text
fn main() fn()
bt u32     u32
kw const
kw const
kw crate::
...
```

**After this PR**

```text
fn main() fn()
bt u32     u32
kw const
kw crate::
...
```

### DIFF
--- a/crates/ide-completion/src/completions/expr.rs
+++ b/crates/ide-completion/src/completions/expr.rs
@@ -352,7 +352,10 @@ pub(crate) fn complete_expr_path(
 
                     if !in_block_expr {
                         add_keyword("unsafe", "unsafe {\n    $0\n}");
-                        add_keyword("const", "const {\n    $0\n}");
+                        if !wants_const_token {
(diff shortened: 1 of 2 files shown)


crates/ide-completion/src/completions/expr.rs
```diff
@@ -355,7 +355,9 @@ pub(crate) fn complete_expr_path(
 
                     if !in_block_expr {
                         add_keyword("unsafe", "unsafe {\n    $0\n}");
-                        add_keyword("const", "const {\n    $0\n}");
+                        if !wants_const_token {
```
ShoyuVanilla, line 355: While this is totally correct, it looks quite counter-intuitive as it adds a const token even though it says *does not want a const token* 😅 Could you add a brief comment addressing this?

PR #20942, opened 2025-10-30
fix: False positive syntax errors on frontmatter

ShoyuVanilla:
Currently, all the frontmatters are reported as syntax errors


ShoyuVanilla:
I'm merging this because this is rather trivial

PR #20945, opened 2025-10-31
Fix typos, backtick errors/omissions

jhpratt:
The `|` character needs to be escaped due to it being in a table. The other changes are self-explanatory.


crates/ide/src/syntax_highlighting.rs
jhpratt, line 137: I'm fairly certain _something_ is supposed to be here, but I can't think of a single category that all of these would fall under.

PR #20953, opened 2025-11-01
Add ide-assist: unmerge_else_if

A4-Tacks:
Unmerge `else if` into the `else {}`.

Example
---
```rust
fn foo() {
    if cond1 {
        xxx()
    } else if$0 cond2 {
        yyy()
    } else {
        zzz()
    }
}
```
->
```rust
fn foo() {
    if cond1 {
        xxx()
    } else {
        if cond2 {
            yyy()
        } else {
            zzz()
        }
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/unmerge_else_if.rs
+++ b/crates/ide-assists/src/handlers/unmerge_else_if.rs
@@ -1,0 +1,135 @@
+use crate::{
+    AssistId,
+    assist_context::{AssistContext, Assists},
+};
+use syntax::{
+    AstNode, T,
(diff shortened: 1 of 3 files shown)


A4-Tacks:
Does this apply to non VSCode LSP front-end scenarios?

Veykril:
Yes, other IDEs should support this too, its an LSP feature after all. For context regarding the other PRs as well (as I started closing some), we decided earlier that we will not be accepting any more new IDE assists for the time being until we can catch up with the backlog of things cc https://github.com/rust-lang/rust-analyzer/pull/21316. With that I am going through most of the IDE assist addition PRs now to determine whether we have interest in them or not.

PR #20964, opened 2025-11-03
fix: Canonicalize `custom-target.json` paths when fetching sysroot metadata

ShoyuVanilla:
Fixes https://github.com/rust-lang/rust-analyzer/issues/20959

The issue might be a regression from https://github.com/rust-lang/rust-analyzer/pull/20866, as before it we used to call `cargo metadata` on sysroot with `cwd = project_root`.
But as the PR fixes some delicated issues around cargo configs that cannot be applied sysroot like vendoring ones.
I think keeping that as is would be more preferable as canonicalizing `build.target` options is a simple fix for the issue, whereas erasing options that shouldn't be applied to sysroot metadata would be quite demanding. 
But we might end up doing so or implement some alternatives as complex as it if we have two large sets of cargo config options - should be applied to sysroot / should not be applied to sysroot 🤔  

BTW, I'm gonna write more organized test suite for cargo handling things this weekend 😅

### DIFF
--- a/crates/project-model/src/workspace.rs
+++ b/crates/project-model/src/workspace.rs
@@ -374,6 +374,7 @@ impl ProjectWorkspace {
                 sysroot.load_workspace(
                     &RustSourceWorkspaceConfig::CargoMetadata(sysroot_metadata_config(
                         config,
+                        workspace_dir,
                         &targets,
                         toolchain.clone(),
                     )),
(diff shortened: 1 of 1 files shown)


ShoyuVanilla:
Well, CI reports compilation errors on a non-existing span 🤔

Veykril (approved):

PR #20967, opened 2025-11-04
Fix not applicable on `and` for replace_method_eager_lazy

A4-Tacks:
Supports:

- and <-> and_then
- then_some <-> then

Example
---
```rust
fn foo() {
    let foo = Some("foo");
    return foo.and$0(Some("bar"));
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
fn foo() {
    let foo = Some("foo");
    return foo.and_then(|| Some("bar"));
}
```


A4-Tacks:
I may need #21044 to add parameters to closure of `and_then`

ShoyuVanilla:
> I may need #21044 to add parameters to closure of `and_then`

Okay, I'll look into it some hours later

PR #20976, opened 2025-11-06
Fix not complete after inner-attr in source-file

A4-Tacks:
Fixes rust-lang/rust-analyzer#15748

Example
---
```rust
#![attr] $0
```

**Before this PR**

Empty completion list

**After this PR**

```text
ma makro!(…) macro_rules! makro
md module
kw async
kw const
kw crate::
kw enum
kw extern
kw fn
kw impl
kw impl for
kw mod
kw pub
kw pub(crate)
kw pub(super)
kw self::
kw static
kw struct
kw trait
kw type
kw union
kw unsafe
kw use
sn macro_rules
sn tfn (Test function)
sn tmod (Test module)
```

### DIFF
--- a/crates/ide-completion/src/context/analysis.rs
+++ b/crates/ide-completion/src/context/analysis.rs
@@ -1635,7 +1635,7 @@ fn classify_name_ref<'db>(
                 && let Some(t) = top.first_token()
                 && let Some(prev) =
                     t.prev_token().and_then(|t| syntax::algo::skip_trivia_token(t, Direction::Prev))
-                && ![T![;], T!['}'], T!['{']].contains(&prev.kind())
+                && ![T![;], T!['}'], T!['{'], T![']']].contains(&prev.kind())
             {
                 // This was inferred to be an item position path, but it seems
                 // to be part of some other broken node which leaked into an item
--- a/crates/ide-completion/src/tests/item_list.rs
+++ b/crates/ide-completion/src/tests/item_list.rs
@@ -106,6 +106,40 @@ fn in_item_list_after_attr() {
 }
 
 #[test]
+fn in_item_list_after_inner_attr() {
+    check_with_base_items(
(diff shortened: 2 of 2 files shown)


crates/ide-completion/src/context/analysis.rs
```diff
@@ -1635,7 +1635,7 @@ fn classify_name_ref<'db>(
                 && let Some(t) = top.first_token()
                 && let Some(prev) =
                     t.prev_token().and_then(|t| syntax::algo::skip_trivia_token(t, Direction::Prev))
-                && ![T![;], T!['}'], T!['{']].contains(&prev.kind())
+                && ![T![;], T!['}'], T!['{'], T![']']].contains(&prev.kind())
```
ShoyuVanilla, line 1638: This seems working but feels a bit ad-hoc to me. Shouldn't it be considered as top context when inserting after an inner attribute?

A4-Tacks: > Shouldn't it be considered as top context when inserting after an inner attribute?

I don't quite understand 'top context'

ShoyuVanilla: I mean, treating it as if the inner attribute weren’t there while completing it

A4-Tacks: There seems to be no problem. When `prev==T![']']`, there is no return and no side effects

ShoyuVanilla: Yeah, I'm pretty sure that this won't cause some unwanted bugs, too.
What felt too me not quite right was just this prev token filtering might be too ad-hoc, but after some checks on my own, I found that we already filtering the cases like completing after `mod foo`, `fn foo() {}` in the same manner 😅

ShoyuVanilla (approved):

PR #20978, opened 2025-11-06
Add function to fetch runnable at cursor position

ERoydev:
Expose a command in the extension that provides an interface for fetching a specific Runnable from the LSP. This lets the user select a Runnable directly instead of using the debug command, which currently prompts them to choose between multiple runnables. The fetched Runnable can then be used to call the debugSingle command directly.

Open to suggestions if there’s existing functionality that already provides this.

Here’s an example of the behavior I’m trying to address::
Screenshot 2025-11-06 at 11 31 36

### DIFF
--- a/editors/code/src/commands.ts
+++ b/editors/code/src/commands.ts
@@ -14,6 +14,7 @@ import {
     selectRunnable,
     createTaskFromRunnable,
     createCargoArgs,
+    selectRunnableAtCursor,
 } from "./run";
 import {
     isRustDocument,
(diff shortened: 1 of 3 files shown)


ERoydev:
can you suggest a solution ?

ChayimFriedman2 (changes requested):
This does not actually execute the runnable. See e.g. https://github.com/rust-lang/rust-analyzer/blob/4eac290b58a70961e78f2e0c04f61a08b995b2cb/editors/code/src/commands.ts#L1106 for how to do that.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/20978/e8ac25242fa87d5b7678580fc42c75c0dc45c4b4..35906cee5fa9b75baed8d66b3c1b1ff238e9f3b8)*

ChayimFriedman2:
@ERoydev What's with this?

ChayimFriedman2:
Closing for no response from OP.

PR #20980, opened 2025-11-06
fix: Remove some deep normalizations from infer

ShoyuVanilla:
Fixes https://github.com/rust-lang/rust-analyzer/issues/20975

The issue comes from this assertion:
https://github.com/rust-lang/rust-analyzer/blob/4bf516ee5a960c1e2eee9fedd9b1c9e976a19c86/crates/hir-ty/src/next_solver/normalize.rs#L48
The deeply normalization goals returns ambiguous errors with type unspecified integer literal because each integer types are all valid candidates.
Normaly, they should be fallbacked into `i32` but that's done manually in the type inference, not from the trait solving.

This PR removes deep normalizations on `Ty` but there are still two places calling it with `T: TypeFoldable` with non `Ty`. 
I tried to remove the deep normalization entirely from infer, this function.
https://github.com/rust-lang/rust-analyzer/blob/4bf516ee5a960c1e2eee9fedd9b1c9e976a19c86/crates/hir-ty/src/infer/unify.rs#L288-L299
But it caused few regressions especially upon const generic parameters so I left them for now. (Our const evaluation are quite wrong so might need be fixed from there)

I think this is not overlaps with https://github.com/rust-lang/rust-analyzer/pull/20974 but this can be waited and rebased upon it if so 😄

### DIFF
--- a/crates/hir-ty/src/infer/unify.rs
+++ b/crates/hir-ty/src/infer/unify.rs
@@ -284,17 +284,6 @@ impl<'db> InferenceTable<'db> {
         self.at(&ObligationCause::new()).deeply_normalize(ty.clone()).unwrap_or(ty)
     }
 
-    /// Works almost same as [`Self::normalize_associated_types_in`], but this also resolves shallow
-    /// the inference variables
-    pub(crate) fn eagerly_normalize_and_resolve_shallow_in<T>(&mut self, ty: T) -> T
-    where
-        T: TypeFoldable<DbInterner<'db>>,
-    {
-        let ty = self.resolve_vars_with_obligations(ty);
-        let ty = self.normalize_associated_types_in(ty);
-        self.resolve_vars_with_obligations(ty)
-    }
-
     pub(crate) fn normalize_alias_ty(&mut self, alias: Ty<'db>) -> Ty<'db> {
         self.infer_ctxt
             .at(&ObligationCause::new(), self.trait_env.env)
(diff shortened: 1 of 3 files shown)


crates/hir-ty/src/infer/unify.rs
```diff
@@ -298,17 +298,6 @@ impl<'db> InferenceTable<'db> {
...
-    /// the inference variables
-    pub(crate) fn eagerly_normalize_and_resolve_shallow_in<T>(&mut self, ty: T) -> T
-    where
-        T: TypeFoldable<DbInterner<'db>>,
-    {
-        let ty = self.resolve_vars_with_obligations(ty);
-        let ty = self.normalize_associated_types_in(ty);
-        self.resolve_vars_with_obligations(ty)
-    }
-
```
ShoyuVanilla, line 297: This ugly function was introduced by me (with some visitors back then) to fix some errors on non-resolved nested opaque types during mir lowering but it isn't needed anymore 😄

crates/hir-ty/src/tests/macros.rs
ShoyuVanilla, line 1: Some tests are seemingly regressed but I believe they are okay because they are errornous types

ShoyuVanilla:
Well, actually this is mostly done by #20974 😅  I'll just add a test and some minor changes - though this one is minor in overall 😅  - in this PR once it has landed

ShoyuVanilla:
I'll check whether we still need this

ShoyuVanilla:
Most of this has been resolved by #20974 and this is quite straightforward like adding a regression test and remove my ugly function. So, I'm gonna merge this

PR #20984, opened 2025-11-07
Support WhileExpr and ForExpr for add_label_to_loop

A4-Tacks:
Example
---
```rust
fn main() {
    for$0 _ in 0..5 {
        break;
        continue;
    }
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
fn main() {
    'l: for _ in 0..5 {
        break 'l;
        continue 'l;
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/add_label_to_loop.rs
+++ b/crates/ide-assists/src/handlers/add_label_to_loop.rs
@@ -35,9 +35,9 @@ use crate::{AssistContext, AssistId, Assists};
 // }
 // ```
 pub(crate) fn add_label_to_loop(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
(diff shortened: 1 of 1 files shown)


crates/syntax/src/ast/node_ext.rs
```diff
@@ -880,6 +880,53 @@ impl AstNode for TypeOrConstParam {
 
 impl HasAttrs for TypeOrConstParam {}
 
+pub enum LoopLike {
```
Veykril: We don;t need this I believe, we can use the `AnyHasLoopBody` node instead which should cover all loop types

A4-Tacks: Because `loop_expr.loop_token()` is needed, AnyHasLoopBody will be more inconvenient, or should I add methods to AnyHasLoopBody?

A4-Tacks: Okay

PR #20986, opened 2025-11-07
Integrate postcard support into proc-macro server CLI

Shourya742:
This PR introduces a postcard binary transport layer for communication between the proc-macro client and server.

### DIFF
--- a/crates/proc-macro-api/src/legacy_protocol/msg.rs
+++ b/crates/proc-macro-api/src/legacy_protocol/msg.rs
@@ -8,10 +8,7 @@ use paths::Utf8PathBuf;
 use serde::de::DeserializeOwned;
 use serde_derive::{Deserialize, Serialize};
 
-use crate::{
-    ProcMacroKind,
-    legacy_protocol::postcard::{decode_cobs, encode_cobs},
-};
+use crate::{ProcMacroKind, codec::Codec};
 
 /// Represents requests sent from the client to the proc-macro-srv.
 #[derive(Debug, Serialize, Deserialize)]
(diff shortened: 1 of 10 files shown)


crates/proc-macro-api/src/legacy_protocol/msg.rs
```diff
@@ -182,6 +205,15 @@ type ProtocolRead<R: BufRead> =
...
 
+/// Type alias for a function that reads protocol postcard messages from a buffered input stream.
+#[allow(type_alias_bounds)]
+type ProtocolReadPostcard<R: BufRead> =
+    for<'i, 'buf> fn(inp: &'i mut R, buf: &'buf mut Vec<u8>) -> io::Result<Option<&'buf mut [u8]>>;
+/// Type alias for a function that writes protocol postcard messages to an output stream.
+#[allow(type_alias_bounds)]
+type ProtocolWritePostcard<W: Write> =
+    for<'o, 'msg> fn(out: &'o mut W, msg: &'msg [u8]) -> io::Result<()>;
+
```
Veykril: We can probably reduce the duplication here between String and Vec<u8> a bit by making things a bit more generic

```suggestion
/// Type alias for a function that reads protocol postcard messages from a buffered input stream.
#[allow(type_alias_bounds)]
type ProtocolRead<R: BufRead, Buf> =
    for<'i, 'buf> fn(inp: &'i mut R, buf: &'buf mut Buf) -> io::Result<Option<&'buf mut Buf>>;
/// Type alias for a function that writes protocol postcard messages to an output stream.
#[allow(type_alias_bounds)]
type ProtocolWrite<W: Write, Buf> =
    for<'o, 'msg> fn(out: &'o mut W, msg: &'msg Buf) -> io::Result<()>;

```
and then adjust `send_task` in a similar way

crates/proc-macro-api/src/lib.rs
```diff
@@ -24,9 +24,10 @@ pub mod version {
     /// Whether literals encode their kind as an additional u32 field and idents their rawness as a u32 field.
     pub const EXTENDED_LEAF_DATA: u32 = 5;
     pub const HASHED_AST_ID: u32 = 6;
+    pub const POSTCARD_WIRE: u32 = 7;
```
Veykril: I don't think there is value in bumping the version here. To figure out if the server supports the protocol we'd need to send a version check on a json server first, then restart the server with postcard. That seems annoying. Given we can't query the binary for postcard support (mishap on my part), I think we can just try to start  the server with postcard, if it exits we know it only likes json.

crates/proc-macro-srv-cli/src/main.rs
```diff
@@ -31,7 +31,7 @@ fn main() -> std::io::Result<()> {
             clap::Arg::new("format")
                 .long("format")
                 .action(clap::ArgAction::Set)
-                .default_value("json")
```
Veykril, line 34: The default should stay as json or this will break older clients that don't support postcard

crates/proc-macro-api/src/process.rs
```diff
@@ -82,7 +83,11 @@ impl ProcMacroServerProcess {
                 if srv.version >= version::RUST_ANALYZER_SPAN_SUPPORT
                     && let Ok(mode) = srv.enable_rust_analyzer_spans()
                 {
-                    srv.protocol = Protocol::LegacyJson { mode };
```
Veykril, line 85: Regarding version, instead of this check here for the version we should be we should try spawning the process with `.args(["--format", "postcard"])`, and if that fails retry with `.args(["--format", "json"])`. Depending on whether the first one success we select the corresponding protocol

crates/proc-macro-api/src/legacy_protocol/msg.rs
bit-aloo: I removed it and added proper framing + codec abstractions. Ended up cutting most of the duplicated code. Let me know if you like the vibes around it.

crates/proc-macro-api/src/process.rs
Shourya742: Done, just reversed the order.

crates/proc-macro-api/src/lib.rs
bit-aloo: I removed it, after your comment, I realized the version field didn’t make much sense. I added fallback logic instead: try spawning the server with postcard first, and if that fails, fall back to JSON

crates/proc-macro-api/src/process.rs
bit-aloo: Done

crates/proc-macro-api/src/process.rs
```diff
@@ -28,9 +28,16 @@ pub(crate) struct ProcMacroServerProcess {
...
+pub(crate) enum Protocol {
     LegacyJson { mode: SpanMode },
+    Postcard { mode: SpanMode },
+}
+
+impl Default for Protocol {
+    fn default() -> Self {
+        Protocol::Postcard { mode: SpanMode::Id }
+    }
 }
```
Veykril: Let's not assign a default here, it feels a bit wrong to me.

crates/proc-macro-api/src/process.rs
```diff
@@ -48,57 +55,94 @@ impl ProcMacroServerProcess {
         env: impl IntoIterator<
             Item = (impl AsRef<std::ffi::OsStr>, &'a Option<impl 'a + AsRef<std::ffi::OsStr>>),
         > + Clone,
+        protocol: Protocol,
     ) -> io::Result<ProcMacroServerProcess> {
-        let create_srv = || {
-            let mut process = Process::run(process_path, env.clone())?;
+        let mut srv = {
+            let mut process = match Process::run(process_path, env.clone(), &protocol) {
```
Veykril, line 55: So I just checked something and it seems I messed up the cli parsing in the proc-macro server. It doesnt validate things so it always spawns successfully. Given you did this hcange in the version check as well it seems you noticed that though, so that should work out I tihnk 👍

crates/proc-macro-api/src/process.rs
```diff
@@ -48,57 +55,94 @@ impl ProcMacroServerProcess {
...
+        srv.version = version;
+
+        if version >= version::RUST_ANALYZER_SPAN_SUPPORT
+            && let Ok(mode) = srv.enable_rust_analyzer_spans()
+        {
+            srv.protocol = match protocol {
+                Protocol::Postcard { .. } => Protocol::Postcard { mode },
+                Protocol::LegacyJson { .. } => Protocol::LegacyJson { mode },
+            };
+        }
```
Veykril, line 124: nit
```suggestion
        if version >= version::RUST_ANALYZER_SPAN_SUPPORT
            && let Ok(new_mode) = srv.enable_rust_analyzer_spans()
        {
            match &mut srv.protocol {
                Protocol::Postcard { mode } | Protocol::LegacyJson { mode } => *mode = new_mode,
            };
        }
```

crates/proc-macro-api/src/process.rs
bit-aloo: It struck me a bit later while reviewing the code that we basically accept any garbage after --format, apart from the valid values. That was actually going to be my next PR idea. 😄

Veykril: Would love a fix for that for sure as a follow up :)

Veykril:
Also cc @jamesmunns

jamesmunns:
I'll need to catch up on the history here, but if you want something a bit more protocol'y structured, I've been working on [`ergot`](https://docs.rs/ergot), which has some better bumpers for binary compat.

PR #21000, opened 2025-11-09
fix: adjust the priority of code actions

Natural-selection1:
Closed: https://github.com/rust-lang/rust-analyzer/issues/3903

---

Example:
```rust
mod a {
    pub const POLY_SUBPIXEL_SHIFT: u32 = 8;
    pub const POLY_SUBPIXEL_SCALE: u32 = (1 << POLY_SUBPIXEL_SHIFT) - 1;
    pub fn lerp_u8() {}
    pub fn prelerp_u8() {}
}

pub mod my_module {
    pub struct HashMap;
    pub const POLY_SUBPIXEL_SCALE: i32 = 256;
}

use a::{POLY_SUBPIXEL_SHIFT, lerp_u8};

fn main() {
    println!("SCALE= {}", POLY_SUBPIXEL_SCALE);
    prelerp_u8();
    let value = POLY_SUBPIXEL_SCALE;
}

```

                 Before               |           After             
image

image

image

---

### Why choose reordering over adjusting the push order of CodeActions from the backend/frontend?

Initially, following the guidance in https://github.com/rust-lang/rust-analyzer/issues/3903, I attempted to adjust the push order in the backend but found it ineffective. 

Then I tried reordering them after the push was complete, and noticed the final presentation seemed to be sorted in segments. For example, we could make `Qualify` rank higher than `Import`, but it could never surpass `similar name`. 

I realized this was likely a frontend issue, and eventually located the relevant code here:

https://github.com/rust-lang/rust-analyzer/blob/d26e4b24d8021f856b3149b882201484ce9d3015/editors/code/src/client.ts#L176-L178

> I'm not very familiar with TS; the subsequent work relied heavily on Sonnet 4.5 : )

image

Based on the inserted logs, the issue is indeed determined by the frontend.
However, judging from the code logic, it doesn't seem possible to change the final presentation order by altering the processing sequence, so I decided to sort them after all processing was complete.

---

The order is based on matklad's idea, namely `ra diagnostic > ra assist > rustc diagnostic`
image

As for the criteria for determining priority, it might not be perfectly precise, but `it works :)`

---
The frontend testing framework seems relatively basic; I'm not sure where to write tests or how to write them.

### DIFF
--- a/editors/code/src/client.ts
+++ b/editors/code/src/client.ts
@@ -267,6 +267,12 @@ export async function createClient(
                     }
                     result.push(primary);
                 }
+
+                result.sort((a, b) => {
+                    const priorityA = getActionPriority(a);
+                    const priorityB = getActionPriority(b);
+                    return priorityA - priorityB;
+                });
                 return result;
             };
             return client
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
I disagree with this order. IMO we should prefer rustc diagnostics to our assists (not sure what the current order is).

Also the sorting should happen server-side.

ChayimFriedman2:
History repeats itself, apparently. https://github.com/rust-lang/rust-analyzer/issues/8645.

PR #21012, opened 2025-11-10
feat: Implement thin-arrow completion in fn return position

A4-Tacks:
Very cool feature that can quickly complete simple return types

Close rust-lang/rust-analyzer#20885

Example
---
```rust
fn foo() u$0
```

**Before this PR**

```text
kw where
```

**After this PR**

```text
bt u32 (adds ->) u32
kw where
...
```

```rust
fn foo() -> u32
```

### DIFF
--- a/crates/ide/src/typing.rs
+++ b/crates/ide/src/typing.rs
@@ -1239,12 +1239,6 @@ sdasdasdasdasd
     #[test]
     fn parenthesis_noop_in_item_position_with_macro() {
         type_char_noop('(', r#"$0println!();"#);
-        type_char_noop(
-            '(',
-            r#"
-fn main() $0println!("hello");
-}"#,
-        );
     }
 
     #[test]


crates/ide-completion/src/context.rs
```diff
@@ -102,6 +102,30 @@ impl PathCompletionCtx<'_> {
...
+        if ret_type.thin_arrow_token().is_some() {
+            return false;
+        }
+        match ret_type.ty() {
+            Some(ast::Type::PathType(path_ty)) => {
+                path_ty.path().is_some_and(|path| path.as_single_name_ref().is_some())
+            }
+            Some(_) => false,
+            None => true,
+        }
```
ChayimFriedman2: Why is this? Is there a problem completing an arrow on e.g. `fn foo() bar::baz| {}`?

crates/ide-completion/src/tests/type_pos.rs
```diff
@@ -93,6 +93,193 @@ fn x<'lt, T, const C: usize>() -> $0
...
 }
 
+#[test]
+fn fn_return_type_missing_thin_arrow() {
+    check_with_base_items(
+        r#"
+fn x() u$0
+"#,
+        expect![[r#"
+            en Enum                    Enum
```
ChayimFriedman2: IMO we should visibly render the `->` in the label.

A4-Tacks: This is difficult to complete, the editor needs to make the matching skip the arrow

ChayimFriedman2: Did you mean to comment on my other comment?

A4-Tacks: No

```
fo
foo
^^
```

Better than this:

```
fo
-> foo
   ^^
```

Some editors do not even support such completion

crates/ide-completion/src/context.rs
A4-Tacks: Usually, arrows are completed when input `bar`

Moreover, when completing `foo::bar`, the range is `bar` instead of `foo::bar`, so adding arrows before requires TextEdit, which is a bit troublesome

crates/ide-completion/src/tests/type_pos.rs
ChayimFriedman2: The label can be whatever we want.

crates/ide-completion/src/context.rs
ChayimFriedman2: You can add the arrow after the `)`.

crates/ide-completion/src/tests/type_pos.rs
A4-Tacks: But the editor will use the current input to match the label, and the strange prefix makes it difficult to match

crates/ide-completion/src/context.rs
A4-Tacks: So this requires TextEdit, but using TextEdit seems to be a good solution as well

crates/ide-completion/src/tests/type_pos.rs
ChayimFriedman2: `CompletionItem::lookup` can control the filtering.

A4-Tacks: Another question, if the label has an arrow, it can sometimes be strange:

```
mod foo {}
fn bar() foo::bar$0 {}
```

Outputs label `-> bar`, but complete `bar`, this is a bit counterintuitive

Perhaps you can refer to auto import, outputs label `bar(adds ->)`

ChayimFriedman2: Right, that's better.

PR #21023, opened 2025-11-11
Fix syntax_editor duplicated changed element

A4-Tacks:
Fixes rust-lang/rust-analyzer#21020

Example
---
```rust
let arg_list = make::arg_list([make::expr_literal("1").into(), make::expr_literal("2").into()]);
let mut editor = SyntaxEditor::new(arg_list.syntax().clone());
let target_expr = make::expr_literal("3").clone_for_update();

for arg in arg_list.args() {
    editor.replace(arg.syntax(), target_expr.syntax());
}

let edit = editor.finish();
let expect = expect![["(3, 3)"]];
expect.assert_eq(&edit.new_root.to_string());
```

**Before this PR**

```text
(, )3
```

**After this PR**

```text
(3, 3)
```

### DIFF
--- a/crates/syntax/src/syntax_editor/edit_algo.rs
+++ b/crates/syntax/src/syntax_editor/edit_algo.rs
@@ -152,9 +152,29 @@ pub(super) fn apply_edits(editor: SyntaxEditor) -> SyntaxEdit {
     let mut changed_elements = vec![];
     let mut changed_elements_set = rustc_hash::FxHashSet::default();
     let mut deduplicate_node = |node_or_token: &mut SyntaxElement| {
-        let SyntaxElement::Node(node) = node_or_token else { return };
+        let node;
+        let node = match node_or_token {
+            SyntaxElement::Token(token) => match token.parent() {
+                None => return,
+                Some(parent) => {
+                    node = parent;
+                    &node
+                }
+            },
+            SyntaxElement::Node(node) => node,
+        };
         if changed_elements_set.contains(node) {
-            *node = node.clone_subtree().clone_for_update();
+            let new_node = node.clone_subtree().clone_for_update();
+            match node_or_token {
+                SyntaxElement::Node(node) => *node = new_node,
+                SyntaxElement::Token(token) => {
+                    *token = new_node
+                        .children_with_tokens()
+                        .filter_map(SyntaxElement::into_token)
+                        .find(|it| it.kind() == token.kind() && it.text() == token.text())
+                        .unwrap();
+                }
+            }
         } else {
             changed_elements_set.insert(node.clone());
         }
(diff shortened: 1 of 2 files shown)


crates/syntax/src/syntax_editor/edit_algo.rs
ShoyuVanilla, line 1: The dedup logic looks correct to me, but I’m not sure whether inserting or replacing with the same syntax node multiple times is part of the intended design. Should we automatically handle this by cloning the duplicated node (as done here), or should we reject duplicates and require the caller to handle it instead?
@DropDemBits, I’d appreciate your thoughts on this 😅

DropDemBits: Oof, it's been a year but I managed to recollect enough context to form an opinion 😆

The current SyntaxEditor design admits that SyntaxNodes do have a form of identity which at the moment is tracked by SyntaxNodes sharing the same NodeData[^1]. At the same time however, the implementation detail being mutable syntax trees necessarily means that shared nodes get mutated (causing this issue), while using `clone_subtree` breaks this identity.

For now, I am *okay* with letting this through to fix this issue, but I've opened #21155 for myself to fix the broader issue.

[^1]: This doesn't necessarily lock us in to having SyntaxNodes be reference counted, just that they need to carry through some identity token when we need it, e.g. some u32 derived from a global AtomicU32 counter. In fact, lifting the concept of node identity into rowan proper should let us remove SyntaxMappings, annoying as they are!

ShoyuVanilla: Huge thanks for your clear explanation!

crates/syntax/src/syntax_editor/edit_algo.rs
```diff
@@ -150,6 +150,35 @@ pub(super) fn apply_edits(editor: SyntaxEditor) -> SyntaxEdit {
...
+                None => return,
+                Some(parent) => {
+                    node = parent;
+                    &node
+                }
+            },
+            SyntaxElement::Node(node) => node,
+        };
+        if changed_elements_set.contains(node) {
+            let new_node = node.clone_subtree().clone_for_update();
```
DropDemBits, line 167: This unfortunately does break node identity at the moment, but it doesn't seem like there's any test depending on it at the moment. There's also not really a good way to fix this at the moment.

DropDemBits: We should also be broadcasting source nodes for dependent changes, but that's trickier because we still need to preserve the mapping in that case. Again, not really a good way to fix that at the moment.

ShoyuVanilla: Oops, looks like I've merged without enough checks 😅

DropDemBits: It's fiiiine 😆

PR #21027, opened 2025-11-12
Build releases with static CRT for `-windows-msvc` targets.

ColinFinck:
This increases the binary size of `rust-analyzer.exe` from 42.4 MB to 42.6 MB.
Which should be acceptable for eliminating 7 DLL dependencies.

Fixes rust-lang/rust-analyzer#20970


ColinFinck:
CC @Veykril

CGMossa:
This solves a huge issue with installing Rust through the standalone installers on Windows. Thanks for this!

PR #21036, opened 2025-11-14
fix: make postfix completion handle all references correctly

Natural-selection1:
fix: https://github.com/rust-lang/rust-analyzer/issues/21035


ChayimFriedman2 (approved):
Thanks!

Issue #17493, opened 2024-06-25
`generate_from_impl_for_enum` should be applicable on the enum itself, generating for all variants at once


Veykril:

Veykril:
@listochkin

PR #21038, opened 2025-11-14
Support multiple variant for generate_from_impl_for_enum

A4-Tacks:
Close rust-lang/rust-analyzer#17493

Example
---
```rust
enum A { $0Foo(u32), Bar$0(i32) }
```

**Before this PR**

```rust
enum A { Foo(u32), Bar(i32) }

impl From<u32> for A {
    fn from(v: u32) -> Self {
        Self::Foo(v)
    }
}
```

**After this PR**

```rust
enum A { Foo(u32), Bar(i32) }

impl From<u32> for A {
    fn from(v: u32) -> Self {
        Self::Foo(v)
    }
}

impl From<i32> for A {
    fn from(v: i32) -> Self {
        Self::Bar(v)
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/generate_from_impl_for_enum.rs
+++ b/crates/ide-assists/src/handlers/generate_from_impl_for_enum.rs
@@ -2,7 +2,10 @@ use hir::next_solver::{DbInterner, TypingMode};
 use ide_db::{RootDatabase, famous_defs::FamousDefs};
 use syntax::ast::{self, AstNode, HasName};
 
-use crate::{AssistContext, AssistId, Assists, utils::generate_trait_impl_text_intransitive};
+use crate::{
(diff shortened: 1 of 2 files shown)


ShoyuVanilla (approved):

lnicola:
This is a little hard to discover. I think the intent of #17493 was to offer the assist on the enum itself, not on the selection.

A4-Tacks:
> This is a little hard to discover.

Perhaps when users become accustomed to applying in multiple variants, it won't be a problem anymore

We still have #20685, maybe we need some assists guidance

> I think the intent of #17493 was to offer the assist on the enum itself

Perhaps the applicability is not broad enough, for example:

```rust
enum Value {
    String(String),
    Number(u64),
    Null,
}
```

What should be done when one of the variants cannot be applied?

PR #21041, opened 2025-11-14
 Fix const eval reading wrong size for array index values

crypto-priest:


ChayimFriedman2:
Can you check if the fix is working in a full IDE, and/or if the test is passing **without the fix**?

ShoyuVanilla:
Yeah, the added tests are passing indeed without the changes in this PR

ChayimFriedman2:
I'm closing this as this doesn't address the issue and the author doesn't respond.

PR #21042, opened 2025-11-15
Improve assist qualified to top when on first segment

A4-Tacks:
Close rust-lang/rust-analyzer#18049

Example
---
```rust
mod std { pub mod fmt { pub trait Debug {} } }
fn main() {
    $0std::fmt::Debug;
    let x: std::fmt::Debug = std::fmt::Debug;
}
```

**Before this PR**

```rust
use std::fmt;

mod std { pub mod fmt { pub trait Debug {} } }
fn main() {
    fmt::Debug;
    let x: fmt::Debug = fmt::Debug;
}
```

**After this PR**

```rust
use std::fmt::Debug;

mod std { pub mod fmt { pub trait Debug {} } }
fn main() {
    Debug;
    let x: Debug = Debug;
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/replace_qualified_name_with_use.rs
+++ b/crates/ide-assists/src/handlers/replace_qualified_name_with_use.rs
@@ -1,4 +1,4 @@
-use hir::AsAssocItem;
+use hir::{AsAssocItem, ModuleDef, PathResolution};
 use ide_db::{
     helpers::mod_path_to_ast,
     imports::insert_use::{ImportScope, insert_use},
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/replace_qualified_name_with_use.rs
```diff
@@ -270,12 +279,117 @@ fn main() {
...
+        check_assist(
+            replace_qualified_name_with_use,
+            r"
+mod std { pub mod option { pub enum Option<T> { Some(T), None } } }
+fn main() {
+    let x = std::option::Option::Some$0(());
+}
+    ",
+            r"
+use std::option::Option::Some;
```
ShoyuVanilla, line 368: I think in most cases people might want importing `std::option::Option` over this 🤔 (Though triggering the assist with the cursor placed on `Option` would do the thing but many wouldn't bother it and instead feel that the assist is buggy)

A4-Tacks: The PR retained the old behavior on this test and did not change it

ShoyuVanilla: Oh, it really is 😅

ShoyuVanilla (approved):
It feels a bit exceptional on the first segment, but it's more reasonable than the current behavior and no-op(which would be the most *consistent* way) so I think this is a reasonable improvement

PR #21057, opened 2025-11-18
Implement precedence for print_hir

A4-Tacks:
Fixes rust-lang/rust-analyzer#16865

Example
---
viewHir

```rust
fn main() {
    let r = &2;
    let _ = &mut (*r as i32)
}
```

**Before this PR**

```rust
fn main() {
    let r = &2;
    let _ = &mut *r as i32;
}
```

**After this PR**

```rust
fn main() {
    let r = &2;
    let _ = &mut (*r as i32);
}
```


A4-Tacks:
Ready for merge

PR #21065, opened 2025-11-19
fix: fix field completion in irrefutable patterns

A4-Tacks:
Example
---
```rust
enum Bar { Value, Nil }
struct Foo { x: Bar }
fn foo(foo: Foo) { match foo { Foo { x: $0 } } }
```

**Before this PR**

```text
st Foo
bn Foo {…} Foo { x$1 }$0
kw mut
kw ref
```

**After this PR**

```text
en Bar
st Foo
bn Foo {…} Foo { x$1 }$0
kw mut
kw ref
```

### DIFF
--- a/crates/ide-completion/src/tests/pattern.rs
+++ b/crates/ide-completion/src/tests/pattern.rs
@@ -133,6 +133,44 @@ fn foo() {
 }
 
 #[test]
+fn refutable_in_record_pat_field() {
+    check(
+        r#"
+enum Bar { Value, Nil }
+struct Foo { x: Bar }
+fn foo(foo: Foo) { match foo { Foo { x: $0 } } }
+"#,
+        expect![[r#"
+            en Bar
+            st Foo
+            bn Foo {…} Foo { x$1 }$0
+            kw mut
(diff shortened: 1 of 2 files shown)


crates/ide-completion/src/tests/pattern.rs
```diff
@@ -132,6 +132,44 @@ fn foo() {
...
+    check(
+        r#"
+enum Bar { Value, Nil }
+struct Foo { x: Bar }
+fn foo(foo: Foo) { match foo { Foo { x: $0 } } }
+"#,
+        expect![[r#"
+            en Bar
+            st Foo
+            bn Foo {…} Foo { x$1 }$0
```
lnicola, line 146: I'm confused, are we supposed to complete `Foo` here?

Ah, nvm, we complete every type.

A4-Tacks: > I'm confused, are we supposed to complete `Foo` here?

What if it has associated constants?

```rust
impl Foo {
    const BAR: Bar = Bar::Nil;
}
```

PR #21068, opened 2025-11-19
feat: make dyn inlay hints configurable

Aditya-PS-05:
closes rust-lang/rust-analyzer#21058 

Implemented a `implied_dyn_trait_hints`  option to `InlayHintsConfig` struct (default to true, meaning the inlay hints will be shown by default and if we wants to disable them, they can do it by using this option)


ChayimFriedman2 (approved):
Thanks!

PR #21076, opened 2025-11-20
add test for Associated type of super trait not resolved in dyn syntax. #10018

asukaminato0721:
fix rust-lang/rust-analyzer#10018

Added a regression test mirroring issue rust-lang/rust-analyzer#10018 that builds a `Pin<Box<dyn FusedFuture<Output = ()> + Send>>` and confirms no spurious “no such associated item” diagnostic is emitted for the Output projection when the associated type originates from a super-trait (Future).


ShoyuVanilla (approved):
Thanks!

PR #21078, opened 2025-11-20
fix add test for Rc type inference failed #20040

asukaminato0721:
fix rust-lang/rust-analyzer#20040

Added the regression test unsized_from_keeps_type_info that builds minimal MyBox/MyRc stand-ins with CoerceUnsized + From to mirror the original Rc::from(Box<[i32]>) scenario and assert that the inferred type for rc stays MyRc<[i32]>.

The fixture uses only minicore pieces (coerce_unsized, from) so it exercises the unsized coercion path entirely within the test harness, ensuring we’ll catch future regressions without needing real Box/Rc.


ShoyuVanilla (approved):
Thanks!

PR #21079, opened 2025-11-20
fix Code action to convert between struct types #19666

asukaminato0721:
fix rust-lang/rust-analyzer#19666

Added convert_record_struct_to_unit_struct to rewrite empty record structs/variants into unit forms, updating constructors, patterns, and all referenced files via Definition::Adt usages.

Added convert_unit_struct_to_record_struct so unit structs/variants become empty record forms, including constructor/pattern rewrites even for bare identifier patterns.

Registered both assists in so they surface in the assist list alongside the existing tuple/named converters.


ChayimFriedman2:

PR #21090, opened 2025-11-21
fix: Enhance remove_parentheses assist to handle return expressions

asuto15:
close rust-lang/rust-analyzer#21063

## Summary

This PR refines the `remove_parentheses` assist’s handling of `return`-like control-flow expressions, based on the discussion in rust-lang/rust-analyzer#21092.

The assist now applies the following rules:

1. Prevent producing `return || …`
    - Parentheses around `return` are not removed when doing so would create `return ||`

2. Allow removal under prefix operators
    - Parentheses around `return`, `return <expr>`, `break`, and `continue` are removed when used under prefix operators like `!`
   (e.g., `!(return)` → `!return`, `!(return false)` → `!return false`).

## Tests

This PR adds comprehensive tests covering:

- All prohibited cases that would form `return || …` (with and without values, with or without a prefix operator).
- Valid prefix cases (`!(return)`, `!(return false)`, `!(break)`, `!(continue)`).
- Valid binary-operator cases (`true || (return)` → `true || return`).

### DIFF
--- a/crates/ide-assists/src/handlers/remove_parentheses.rs
+++ b/crates/ide-assists/src/handlers/remove_parentheses.rs
@@ -163,17 +163,18 @@ mod tests {
     }
 
     #[test]
-    fn remove_parens_prefix_with_return_no_value() {
-        // removing `()` from !(return) would make `!return` which is invalid syntax
-        check_assist_not_applicable(
+    fn remove_parens_prefix_with_ret_like_prefix() {
+        check_assist(remove_parentheses, r#"fn f() { !$0(return) }"#, r#"fn f() { !return }"#);
+        // `break`, `continue` behave the same under prefix operators
+        check_assist(remove_parentheses, r#"fn f() { !$0(break) }"#, r#"fn f() { !break }"#);
+        check_assist(remove_parentheses, r#"fn f() { !$0(continue) }"#, r#"fn f() { !continue }"#);
+        check_assist(
             remove_parentheses,
-            r#"fn main() { let _x = true && !$0(return) || true; }"#,
+            r#"fn f() { !$0(return false) }"#,
+            r#"fn f() { !return false }"#,
         );
-        check_assist_not_applicable(remove_parentheses, r#"fn f() { !$0(return) }"#);
-        check_assist_not_applicable(remove_parentheses, r#"fn f() { !$0(break) }"#);
-        check_assist_not_applicable(remove_parentheses, r#"fn f() { !$0(continue) }"#);
 
-        // Binary operators should still allow removal
+        // Binary operators should still allow removal unless a ret-like expression is immediately followed by `||` or `&&`.
         check_assist(
             remove_parentheses,
             r#"fn f() { true || $0(return) }"#,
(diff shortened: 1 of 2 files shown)


crates/ide-assists/src/handlers/remove_parentheses.rs
```diff
@@ -33,8 +34,21 @@ pub(crate) fn remove_parentheses(acc: &mut Assists, ctx: &AssistContext<'_>) ->
...
+    if let ast::Expr::ReturnExpr(_) = &expr
+        && let Some(r_paren) = parens.r_paren_token()
+        && let Some(next) = r_paren.next_token().and_then(|t| skip_trivia_token(t, Direction::Next))
+        && next.kind() == T![||]
+    {
+        cov_mark::hit!(remove_parens_return_closure);
+        return None;
+    }
+
     let parent = parens.syntax().parent()?;
```
A4-Tacks, line 36: It may be better to solve it in `ast/prec.rs` instead of adding more special cases where it's used

BTW, the `ReturnExpr` etc in `ExprPrecedence` should also always return a `Jump` precedence

crates/syntax/src/ast/prec.rs
```diff
@@ -226,9 +221,15 @@ impl Expr {
             return false;
         }
 
-        // Special-case prefix operators with return/break/etc without value
-        // e.g., `!(return)` - parentheses are necessary
-        if self.is_ret_like_with_no_value() && parent.is_prefix() {
+        // Keep parens when `(return ...)` is followed by `||`;
+        // otherwise it would become `return || <closure>` with different semantics
+        if let Expr::ReturnExpr(_) = self
```
A4-Tacks: And BreakExpr, BecomeExpr etc, but maybe it doesn't contain ContinueExpr

crates/syntax/src/ast/prec.rs
```diff
@@ -226,9 +221,15 @@ impl Expr {
...
+            && let Some(paren_expr) = self.syntax().parent().and_then(ast::ParenExpr::cast)
```
A4-Tacks: Maybe `parent` should be used instead of `self.syntax().parent()`?

crates/syntax/src/ast/prec.rs
```diff
@@ -226,9 +221,15 @@ impl Expr {
...
+            && let Some(r_paren) = paren_expr.r_paren_token()
+            && let Some(next) =
+                r_paren.next_token().and_then(|t| skip_trivia_token(t, Direction::Next))
+            && next.kind() == T![||]
```
A4-Tacks: And `&&` ? Trying use ~`BinExpr::can_cast(parent.kind())`?~

asuto15:
I tried to use `BinExpr::can_cast(parent.kind())`, but it break the case where `!(return)` must be not-applicable failed.
This suggests that relying on the parent kind is not appropriate for detecting the closure risk.

crates/syntax/src/ast/prec.rs
```diff
@@ -226,9 +221,25 @@ impl Expr {
...
+            self,
+            Expr::ReturnExpr(_)
+                | Expr::BreakExpr(_)
+                | Expr::BecomeExpr(_)
+                | Expr::YeetExpr(_)
+                | Expr::YieldExpr(_)
+        ) && let Some(paren_expr) = self.syntax().parent().and_then(ast::ParenExpr::cast)
+            && let Some(parent_expr) = paren_expr.syntax().parent().and_then(ast::Expr::cast)
+            && BinExpr::can_cast(parent_expr.syntax().kind())
+            && let Some(r_paren) = paren_expr.r_paren_token()
```
A4-Tacks: Maybe we need to sort out the relationship between `self`, `parent`, and `place_of`.
The input may not contain ParenExpr or may be on other trees

A4-Tacks:
> This suggests that relying on the parent kind is not appropriate for detecting the closure risk.

Yes, this is not very good

crates/syntax/src/ast/prec.rs
A4-Tacks: > The input may not contain ParenExpr or may be on other trees

For other calls, not just this assist

asuto15:
Based on your feedback, I will revise the approach as follows:

1. Keep `prec.rs` responsible only for expression precedence.  
  Ret-like expressions will simply return `ExprPrecedence::Jump`, without inspecting surrounding `ParenExpr` or next tokens.

2. Move all syntactic “closure‐risk” checks (e.g., `return` followed by `||`/`&&`) back into the `remove_parentheses` assist, where the actual `ParenExpr` and its neighboring tokens are reliably available.

3. Avoid relying on the syntax parent inside `needs_parens_in_expr`, since `self`, `parent`, and `place_of` may not correspond to the concrete AST structure.

A4-Tacks:
>     2. Move all syntactic “closure‐risk” checks (e.g., `return` followed by `||`/`&&`) back into the `remove_parentheses` assist, where the actual `ParenExpr` and its neighboring tokens are reliably available.

Maybe it can be resolved in the `prec.rs`

Try checking if self like ReturnExpr, ~if the `parent` is BinExpr and if place_of is a subtree of `parent.lhs()`?~

Similar:
```rust
&& let Some(node) = place_of.ancestors().find(|it| it.parent().is_none_or(|it| it == parent))
&& let Some(last_token) = node.last_token()
&& let Some(next) = last_token.next_token().and_then(|t| skip_trivia_token(t, Direction::Next))
// ...
```

A4-Tacks:
A similar case:

```rust
fn foo() -> bool {
    // Inline variable
    let x$0: bool = return false;
    !x || true;
}
```

A4-Tacks:
> Maybe it can be resolved in the `prec.rs`

This seems to have passed all the tests:

```rust
        if self.precedence() == ExprPrecedence::Jump
            && let Some(node) = place_of.ancestors().find(|it| it.parent().is_none_or(|it| &it == parent.syntax()))
            && let Some(next) = node.last_token().and_then(|t| skip_trivia_token(t.next_token()?, Direction::Next))
            && matches!(next.kind(), T![||] | T![&&])
        {
            return true;
        }
```

asuto15:
> ```rust
>        if self.precedence() == ExprPrecedence::Jump
>             && let Some(node) = place_of.ancestors().find(|it| it.parent().is_none_or(|it| &it == parent.syntax()))
>             && let Some(next) = node.last_token().and_then(|t| skip_trivia_token(t.next_token()?, Direction::Next))
>             && matches!(next.kind(), T![||] | T![&&])
>         {
>             return true;
>         }
> ```
This code works perfectly

A4-Tacks (approved):

asuto15:
Thank you for reviewing my code so actively!

PR #21092, opened 2025-11-21
fix: never remove parens from prefix ops with valueless return/break/continue

Aditya-PS-05:
fixes rust-lang/rust-analyzer#21063

The `Remove redundant parentheses` was suggesting removal of parentheses in expressions like !(return), !(break), and !(continue), which would result in invalid syntax.

So, added a case in `Expr::needs_parens_in()` which checks for two things:
1.) Prefix operator 
2.) is_ret_like_with_no_value() which handles return, yield, break, continue, become

If both are present, the parenthesis are not removed. 

Added test also.

### DIFF
--- a/crates/ide-assists/src/handlers/remove_parentheses.rs
+++ b/crates/ide-assists/src/handlers/remove_parentheses.rs
@@ -163,6 +163,30 @@ mod tests {
     }
 
     #[test]
+    fn remove_parens_prefix_with_return_no_value() {
+        // removing `()` from !(return) would make `!return` which is invalid syntax
+        check_assist_not_applicable(
+            remove_parentheses,
+            r#"fn main() { let _x = true && !$0(return) || true; }"#,
+        );
+        check_assist_not_applicable(remove_parentheses, r#"fn f() { !$0(return) }"#);
+        check_assist_not_applicable(remove_parentheses, r#"fn f() { !$0(break) }"#);
+        check_assist_not_applicable(remove_parentheses, r#"fn f() { !$0(continue) }"#);
+
+        // Binary operators should still allow removal
+        check_assist(
+            remove_parentheses,
+            r#"fn f() { true || $0(return) }"#,
+            r#"fn f() { true || return }"#,
+        );
+        check_assist(
+            remove_parentheses,
+            r#"fn f() { cond && $0(return) }"#,
+            r#"fn f() { cond && return }"#,
+        );
+    }
+
+    #[test]
     fn remove_parens_return_with_value_followed_by_block() {
         check_assist(
             remove_parentheses,
--- a/crates/syntax/src/ast/prec.rs
+++ b/crates/syntax/src/ast/prec.rs
@@ -226,6 +226,12 @@ impl Expr {
             return false;
         }
 
+        // Special-case prefix operators with return/break/etc without value
+        // e.g., `!(return)` - parentheses are necessary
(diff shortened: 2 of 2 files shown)


asuto15:
I opened PR #21090 for the same issue, but your approach looks smarter and the test coverage is broader than mine.

To the maintainers: if this solution is preferred, I can close my PR (#21090). Please let me know your preference.

Aditya-PS-05:
@asuto15 actually I started working on the issue and completed earlier than you but could not open a pr due to time constraint. So, I opened it anyway.

ShoyuVanilla (approved):
Yeah, this one is a bit cleaner than #21090 (though I think both are great), so I'm going to merge this one as @asuto15 suggested.
But in general, I think the person assigned to the issue should have priority.

A4-Tacks:
This does not solve issue, such as this test:

```rust
check_assist_not_applicable(
    remove_parentheses,
    r#"fn main() { let _x = true && $0(return) || true; }"#,
);
check_assist(
    remove_parentheses,
    r#"fn f() { cond && !$0(return) }"#,
    r#"fn f() { cond && !return }"#,
);
```

asuto15:
@A4-Tacks 
Hi, I’d like to confirm the expected output for your second example:

```rust
check_assist(
    remove_parentheses,
    r#"fn f() { cond && !$0(return) }"#,
    r#"fn f() { cond && return }"#,
);
```

Since the assist only removes parentheses (and does not remove the unary `!`),
should the expected result actually be:

```rust
fn f() { cond && !return }
```

If that’s the intended behavior, I can proceed and prepare a follow-up PR(#21090) based on that specification.

A4-Tacks:
> If that’s the intended behavior

Yes, this is intended behavior. I missed typing a character

asuto15:
Thank you!

One more question for the maintainers @ShoyuVanilla  and @A4-Tacks:

Would you prefer that we treat all valueless control-flow expressions (`return`, `break`, `continue`) uniformly?

For example, allowing:

```rust
fn f() { !($0(return)) }   // -> !return
fn f() { !($0(break)) }    // -> !break
fn f() { !($0(continue)) } // -> !continue
```

(except the `return ||` case, which must remain blocked)

This would give a simple, consistent rule.  
If this direction is acceptable, I can adjust the PR(#21090) accordingly.

A4-Tacks:
> valueless control-flow expressions

Consider this test:

```rust
check_assist_not_applicable(
    remove_parentheses,
    r#"fn main() { let _x = true && $0(return false) || true; }"#,
);
check_assist(
    remove_parentheses,
    r#"fn f() { cond && !$0(return false) }"#,
    r#"fn f() { cond && !return false }"#,
);
```

ShoyuVanilla:
> This does not solve issue, such as this test:
> 
> ```rust
> check_assist_not_applicable(
>     remove_parentheses,
>     r#"fn main() { let _x = true && $0(return) || true; }"#,
> );
> check_assist(
>     remove_parentheses,
>     r#"fn f() { cond && !$0(return) }"#,
>     r#"fn f() { cond && !return }"#,
> );
> ```

Oh, good catch

Aditya-PS-05:
Sorry I missed this discussion. Does this pr not fixes the issue or any edge case?

A4-Tacks:
> Does this pr not fixes the issue or any edge case?

The main issue is BinExpr instead of PrefixExpr

PR #21097, opened 2025-11-22
proc-macro-srv: Reimplement token trees via immutable trees

Veykril:
For macros, rust has the concept of [TokenTree](https://doc.rust-lang.org/stable/proc_macro/enum.TokenTree.html)s and [TokenStream](https://doc.rust-lang.org/stable/proc_macro/struct.TokenStream.html)s which are basically a tree-like form of the underlying tokenized input with `(){}[]` delimiters forming [Group](https://doc.rust-lang.org/stable/proc_macro/struct.Group.html)s (internal nodes) and all other tokens forming leaf nodes of the tree.

In rust-analyzer today these trees are implemented as a flattened list using indices to reduce their memory impact as we keep a lot of these memoized ([implementation](https://github.com/rust-lang/rust-analyzer/blob/master/crates/tt/src/lib.rs)). This comes at the cost of mutating operations being more expensive though and we especially feel this in proc-macro expansion.

As it turns out, one of the primitive and very commonly used operations on token streams in proc-macros is concatenation, an operation that with our model is notoriously expensive as we need to do a deepclone of the first stream, append the second stream and rewrite the internal indices. Combine this with deeply recursive concatenations as is common with the [`quote!`](https://docs.rs/quote/latest/quote/) macro and we run into issues where the proc-macro server can take 8 seconds to expand [these 8 derive invocations](https://github.com/rust-minidump/rust-minidump/blob/df483d5774cae0fa1bb847eaa3dd8f4a15c7e9d9/minidump-common/src/errors/windows.rs) of [`num_derive::FromPrimitive`](https://github.com/rust-num/num-derive/blob/7cc33515dd2ae0eb43c5795c50ce49c554e8ba02/src/lib.rs#L241-L365) on decent machines.

Fortunately, the proc-macro server has no reason to use the same implementation of TokenStreams and TokenTrees as rust-analyzer since it doesn't persist them in memory anways. So instead, to gain good concatenation performance, we can do with what rustc does and store them as immutable trees with shared internal nodes.

Before this change cache priming took 56 seconds on Zed's codebase, with the new proc-macro server it now takes 45.

### DIFF
--- a/crates/proc-macro-srv/src/tests/mod.rs
+++ b/crates/proc-macro-srv/src/tests/mod.rs
@@ -10,16 +10,44 @@ use expect_test::expect;
 fn test_derive_empty() {
     assert_expand(
         "DeriveEmpty",
-        r#"struct S;"#,
+        r#"struct S { field: &'r#lt fn(u32) -> &'a r#u32 }"#,
         expect![[r#"
             IDENT 1 struct
             IDENT 1 S
-            PUNCT 1 ; [alone]
+            GROUP {} 1 1 1
+              IDENT 1 field
+              PUNCT 1 : [alone]
+              PUNCT 1 & [joint]
+              PUNCT 1 ' [joint]
+              IDENT 1 r#lt
+              IDENT 1 fn
+              GROUP () 1 1 1
+                IDENT 1 u32
+              PUNCT 1 - [joint]
+              PUNCT 1 > [alone]
+              PUNCT 1 & [joint]
+              PUNCT 1 ' [joint]
+              IDENT 1 a
+              IDENT 1 r#u32
         "#]],
         expect![[r#"
             IDENT 42:Root[0000, 0]@0..6#ROOT2024 struct
             IDENT 42:Root[0000, 0]@7..8#ROOT2024 S
-            PUNCT 42:Root[0000, 0]@8..9#ROOT2024 ; [alone]
+            GROUP {} 42:Root[0000, 0]@9..10#ROOT2024 42:Root[0000, 0]@46..47#ROOT2024 42:Root[0000, 0]@9..47#ROOT2024
+              IDENT 42:Root[0000, 0]@11..16#ROOT2024 field
+              PUNCT 42:Root[0000, 0]@16..17#ROOT2024 : [alone]
+              PUNCT 42:Root[0000, 0]@18..19#ROOT2024 & [joint]
+              PUNCT 42:Root[0000, 0]@22..23#ROOT2024 ' [joint]
+              IDENT 42:Root[0000, 0]@22..24#ROOT2024 r#lt
+              IDENT 42:Root[0000, 0]@25..27#ROOT2024 fn
+              GROUP () 42:Root[0000, 0]@27..28#ROOT2024 42:Root[0000, 0]@31..32#ROOT2024 42:Root[0000, 0]@27..32#ROOT2024
+                IDENT 42:Root[0000, 0]@28..31#ROOT2024 u32
+              PUNCT 42:Root[0000, 0]@33..34#ROOT2024 - [joint]
+              PUNCT 42:Root[0000, 0]@34..35#ROOT2024 > [alone]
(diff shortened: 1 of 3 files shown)


Veykril:
This is basically finished, but I realized we have a severe lack of tests here so I'll add some more thorough testing later.

crates/proc-macro-api/src/legacy_protocol/msg/flat.rs
```diff
@@ -522,6 +621,105 @@ impl<'a, T: SpanTransformer> Writer<'a, '_, T> {
...
+                        suffix,
+                    });
+                    (idx << 2) | 0b01
+                }
+                proc_macro_srv::TokenTree::Punct(punct) => {
+                    let idx = self.punct.len() as u32;
+                    let id = self.token_id_of(punct.span);
+                    self.punct.push(PunctRepr {
+                        char: punct.ch as char,
+                        spacing: if punct.joint { tt::Spacing::Joint } else { tt::Spacing::Alone },
```
bit-aloo, line 683: Curious, why we don't encode `JointHidden` variant?

crates/proc-macro-api/src/legacy_protocol/msg/flat.rs
```diff
@@ -641,3 +839,122 @@ impl<T: SpanTransformer> Reader<'_, T> {
...
+                    tt::DelimiterKind::Parenthesis => proc_macro_srv::Delimiter::Parenthesis,
+                    tt::DelimiterKind::Brace => proc_macro_srv::Delimiter::Brace,
+                    tt::DelimiterKind::Bracket => proc_macro_srv::Delimiter::Bracket,
+                    tt::DelimiterKind::Invisible => proc_macro_srv::Delimiter::None,
+                },
+                stream: if stream.is_empty() { None } else { Some(TokenStream::new(stream)) },
+                span: proc_macro_srv::DelimSpan {
+                    open: read_span(repr.open),
+                    close: read_span(repr.close),
+                    // FIXME
```
bit-aloo, line 949: Shouldn’t Reader and Writer really be restricted to a SpanTransformer<Table = SpanDataIndexMap and Span = Span>? That’s the only setup that actually gives us the full span information needed to reconstruct things like entire, right?
I am also curios why is SpanTransformer even implemented for SpanId in the first place?

Shourya742 (approved):
LGTM. Just some questions.

crates/proc-macro-api/src/legacy_protocol/msg/flat.rs
Veykril: `JointHidden` becomes `Alone` after a proc-macro roundtrip, so this preserves that logic. The hidden part effectively means its hidden from proc-macros in this sense.

crates/proc-macro-api/src/legacy_protocol/msg/flat.rs
Veykril: Our macro setup doesn't really support `entire` at all, and afaik, its not really used upstream (in rustc) either anymore so it might be removed in the future.

As for `SpanId`, it's the legacy span type. r-a doesn't use it anymore but we keep it for backwards compatibility for RustRover (as well as if we need to break things, we can always fall back to that mode).

ChayimFriedman2 (approved):
Mostly skimmed, but LGTM.

crates/proc-macro-api/src/legacy_protocol/msg/flat.rs
bit-aloo: Ah!! got it thanks.

crates/proc-macro-api/src/legacy_protocol/msg/flat.rs
bit-aloo: Oh, I didn’t know rustc no longer uses it. Then it’s not really an issue once it’s gone upstream, we can remove it on our side. Thanks

Veykril:
This might've introduced a bug according to https://youtrack.jetbrains.com/issue/RUST-19283

PR #21098, opened 2025-11-22
fix: include all target types with paths outside package root

Aditya-PS-05:
closes rust-lang/rust-analyzer#21072

The issue was that only the lib targets with custom paths were added to the source root includes. So binary (and other) targets failed when their path was set to a location outside the cargo package directory.

The fix removes the filter that restricted this behavior to only lib, thus allowing all target types (Bin, Example, Test, etc.) with custom paths to have their parent directories included in source roots.

### DIFF
--- a/crates/project-model/src/workspace.rs
+++ b/crates/project-model/src/workspace.rs
@@ -844,6 +844,8 @@ impl ProjectWorkspace {
                             exclude.push(pkg_root.join("examples"));
                             exclude.push(pkg_root.join("benches"));
                         }
+                        include.sort();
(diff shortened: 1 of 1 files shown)


crates/project-model/src/workspace.rs
Veykril, line 847: lets sort -> dedup `include` here as with this change we will likely get duplicates (and I don't trust our vfs code here that it can work well with dupes)

Aditya-PS-05: done

Veykril:
Thanks!

PR #21100, opened 2025-11-22
add semantic tokens for deprecated items

xdBronch:
this lets things be rendered with a strikethrough without needing any diagnostics
image

it can be set in vscode settings like so
```
        "rules": {
            "*.deprecated": { "fontStyle": "strikethrough" }
        }
    },
```

### DIFF
--- a/crates/ide/src/syntax_highlighting/highlight.rs
+++ b/crates/ide/src/syntax_highlighting/highlight.rs
@@ -3,7 +3,7 @@
 use std::ops::ControlFlow;
 
 use either::Either;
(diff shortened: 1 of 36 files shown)


ChayimFriedman2 (changes requested):
I'm not entirely sure, but I think I prefer to handle this for each variant in a common place instead of duplicating it all over, e.g. from the `match` over the definition return `Option<AttrsWithOwner>` as well, then check it after.

xdBronch:
sure i think that makes sense, the places we handle it become a little cleaner

PR #21108, opened 2025-11-23
fix(parser): feature gate tracing spans

lmmx:
- Fixes missing feature gates in the parser crate as described in rust-lang/rust-analyzer#21107 
- Adds new default feature in the syntax crate to match, feature-gates the tracing spans there


lmmx:
Closing in favour of alternative using `default-features = false` as discussed in #21107

PR #21114, opened 2025-11-23
fix: no unused `tracing/attributes` feature

lmmx:
- Discussed in https://github.com/rust-lang/rust-analyzer/issues/21107
- Avoids activating an `attributes` feature to crates that do not use it
- Updates the 6 crates that use attributes feature to specify it in
  their Cargo.toml: {hir,hir-def,hir-ty,ide-assists,ide-db,project-model}
- Removes the `tracing` (implicitly `dep:tracing`) dependency from the parser crate `default` feature group (which now leaves it empty)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,7 +156,7 @@ smol_str = "0.3.2"
 temp-dir = "0.1.16"
 text-size = "1.1.1"
 toml = "0.9.8"
-tracing = "0.1.41"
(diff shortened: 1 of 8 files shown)


lmmx:
All checks passing and rebased to trunk :slightly_smiling_face:  Ready to merge @Veykril @ChayimFriedman2

PR #21118, opened 2025-11-24
make add_missing_impl_members can handle naming conflicts

### LINKED ISSUE #6870: Implement missing members suggestion shadows type parameters
In the following minimal example in latest nightly (using a `struct<T> DiagnosticResult { ... }`), the "implement missing members" intention shadows the type parameter `T`.
image
image
This creates rather confusing error messages if you didn't actually read the code that it inserted!
image
The two type parameters `T` and `T` are different, but I imagine that in a more complex scenario than this, it could hide the source of bugs. I think rust-analyzer should probably output code similar to
image

Natural-selection1:
fix: https://github.com/rust-lang/rust-analyzer/issues/6870

The `add_missing_impl_members.rs` file mainly involves code style adjustments and the addition of tests. The main review should focus on `utils.rs`.

### DIFF
--- a/crates/ide-assists/src/handlers/add_missing_impl_members.rs
+++ b/crates/ide-assists/src/handlers/add_missing_impl_members.rs
@@ -1,6 +1,6 @@
 use hir::HasSource;
 use syntax::{
-    Edition,
+    Edition, ToSmolStr,
     ast::{self, AstNode, make},
     syntax_editor::{Position, SyntaxEditor},
 };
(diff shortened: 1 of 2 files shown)


George-Ogden:
@Natural-selection1, what's the status of this PR? I would like to see this feature and would be happy to take over and make the rebase if you are no longer working on it.

Natural-selection1:
fell free to do anything you want : )

George-Ogden:
I have reopened this in https://github.com/rust-lang/rust-analyzer/pull/23358

PR #21133, opened 2025-11-26
fix: Do not try to connect via postcard to proc-macro-srv

Veykril:
Closes https://github.com/rust-lang/rust-analyzer/issues/21130

### DIFF
--- a/crates/proc-macro-api/src/lib.rs
+++ b/crates/proc-macro-api/src/lib.rs
@@ -22,7 +22,7 @@ use span::{ErasedFileAstId, FIXUP_ERASED_FILE_AST_ID_MARKER, Span};
 use std::{fmt, io, sync::Arc, time::SystemTime};
 
 pub use crate::codec::Codec;
-use crate::{legacy_protocol::SpanMode, process::ProcMacroServerProcess};
+use crate::process::ProcMacroServerProcess;
 
 /// The versions of the server protocol
 pub mod version {
@@ -126,11 +126,7 @@ impl ProcMacroClient {
             Item = (impl AsRef<std::ffi::OsStr>, &'a Option<impl 'a + AsRef<std::ffi::OsStr>>),
         > + Clone,
     ) -> io::Result<ProcMacroClient> {
-        let process = ProcMacroServerProcess::run(
-            process_path,
-            env,
-            process::Protocol::LegacyJson { mode: SpanMode::Id },
-        )?;
+        let process = ProcMacroServerProcess::run(process_path, env)?;
         Ok(ProcMacroClient { process: Arc::new(process), path: process_path.to_owned() })
     }
 
--- a/crates/proc-macro-api/src/process.rs
+++ b/crates/proc-macro-api/src/process.rs
@@ -33,7 +33,7 @@ pub(crate) enum Protocol {
     LegacyJson {
         mode: SpanMode,
     },
-    #[expected(dead_code)]
+    #[expect(dead_code)]
     Postcard {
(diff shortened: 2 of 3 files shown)


Veykril:
cc @Shourya742 seems like if we try to connect to a proc-macro server that does not support postcard we will just hang on it now crashing as expected, we'll need to figure out a different approach here.

bit-aloo:
> cc @Shourya742 seems like if we try to connect to a proc-macro server that does not support postcard we will just hang on it now crashing as expected, we'll need to figure out a different approach here.


I guess I know the reason for this, the json framing uses `read_line`, which blocks until it sees a newline. That means if the server doesn’t send one, we end up stuck waiting indefinitely. The problematic bit is here:

```rust
fn read<'a, R: BufRead>(
    inp: &mut R,
    buf: &'a mut String,
) -> io::Result<Option<&'a mut String>> {
    loop {
        buf.clear();

        inp.read_line(buf)?;
        buf.pop(); // Remove trailing '\n'
```

When we intialize the client with postcard, we don't write new line as the delimitor. I’ll send a patch.

Veykril:
So looking at things, the `--format` flag does seem to validate its inputs on the latest nightly, so we might just want to hold off a bit. Do a toolchain version check and then retry things here. The postcard protocol not using lineendings seems fine to me after all

bit-aloo:
> So looking at things, the `--format` flag does seem to validate its inputs on the latest nightly, so we might just want to hold off a bit. Do a toolchain version check and then retry things here. The postcard protocol not using lineendings seems fine to me after all

I was thinking that if we don’t receive a response within a specified time, we should abort the process and fall back.

Veykril:
that would work too and is definitely something we should implement.

I'll revert some parts here for now to unblock the pre-releases. And I'll also include https://github.com/rust-lang/rust-analyzer/pull/21132. It was a mistake not naming both of these flags legacy i think since we do want to replace the protocol entirely at some point.

Shourya742:
> that would work too and is definitely something we should implement.
> 
> I'll revert some parts here for now to unblock the pre-releases. And I'll also include #21132. It was a mistake not naming both of these flags legacy i think since we do want to replace the protocol entirely at some point.

I agree on the naming, but can't we just use the `json` as Protocol variant, this patch should unblock the issue:

```rust
diff --git a/crates/proc-macro-api/src/lib.rs b/crates/proc-macro-api/src/lib.rs
index a725b94f04..b438efd6dc 100644
--- a/crates/proc-macro-api/src/lib.rs
+++ b/crates/proc-macro-api/src/lib.rs
@@ -129,7 +129,7 @@ impl ProcMacroClient {
         let process = ProcMacroServerProcess::run(
             process_path,
             env,
-            process::Protocol::Postcard { mode: SpanMode::Id },
+            process::Protocol::LegacyJson { mode: SpanMode::Id },
         )?;
         Ok(ProcMacroClient { process: Arc::new(process), path: process_path.to_owned() })
     }
diff --git a/crates/proc-macro-api/src/process.rs b/crates/proc-macro-api/src/process.rs
index 1365245f98..200122aa36 100644
--- a/crates/proc-macro-api/src/process.rs
+++ b/crates/proc-macro-api/src/process.rs
@@ -31,6 +31,7 @@ pub(crate) struct ProcMacroServerProcess {
 #[derive(Debug, Clone)]
 pub(crate) enum Protocol {
+    #[allow(dead_code)]
     Postcard { mode: SpanMode },
 }
```

Shourya742 (approved):
ACK

PR #21144, opened 2025-11-27
Fix complete after `extern`, add `crate` completion

A4-Tacks:
Close rust-lang/rust-analyzer#19761

Example
---
```rust
extern "C" $0
```

**Before this PR**

Can't be completion

**After this PR**

```text
kw async
kw const
kw enum
kw fn
kw impl
kw impl for
kw mod
kw pub
kw pub(crate)
kw pub(super)
kw static
kw struct
kw trait
kw type
kw union
kw unsafe
kw use
```

---

```rust
extern $0
```

**Before this PR**

Can't be completion

**After this PR**

```rust
extern crate $0;
```


ChayimFriedman2 (approved):
Thanks!

PR #21152, opened 2025-11-28
Add render associated constant in notable trait

A4-Tacks:
Example
---
```rust
impl Notable for u32 {
    type Assoc = &str;
    type Assoc2 = char;
    const ID: u32 = 3;
}
```

**Before this PR**

```text
Implements notable traits: `Notable<Assoc = &str, Assoc2 = char>`
```

**After this PR**

```text
Implements notable traits: `Notable<Assoc = &str, Assoc2 = char, ID = 3>`
```

### DIFF
--- a/crates/ide/src/hover.rs
+++ b/crates/ide/src/hover.rs
@@ -76,6 +76,17 @@ pub enum HoverDocFormat {
     PlainText,
 }
 
+enum TyOrConst<'db> {
+    Const(hir::Const),
+    Type(hir::Type<'db>),
+}
+
+impl<'db> TyOrConst<'db> {
+    fn as_type(&self) -> Option<&hir::Type<'db>> {
+        if let Self::Type(ty) = self { Some(ty) } else { None }
+    }
+}
+
 #[derive(Debug, Clone, Hash, PartialEq, Eq, UpmapFromRaFixture)]
 pub enum HoverAction {
     Runnable(Runnable),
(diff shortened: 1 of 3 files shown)


crates/ide/src/hover.rs
```diff
@@ -526,14 +537,32 @@ fn notable_traits<'db>(
         .filter_map(move |&trait_| {
             let trait_ = trait_.into();
             ty.impls_trait(db, trait_, &[]).then(|| {
+                // FIXME: This may have a better implementation
```
ChayimFriedman2, line 563: What? No. Absolutely not. Just like you can normalize assoc types, you can normalize assoc consts, although I think we don't currently expose that in `hir`.

crates/ide/src/hover/render.rs
```diff
@@ -961,7 +911,16 @@ fn render_notable_trait(
                     f(&name.display(db, edition))?;
                     f(&" = ")?;
                     match ty {
-                        Some(ty) => f(&ty.display(db, display_target)),
+                        Some(TyOrConst::Type(ty)) => f(&ty.display(db, display_target)),
+                        Some(TyOrConst::Const(it)) => match it.eval(db) {
```
ChayimFriedman2, line 883: I guess now `assoc_types` isn't a good name anymore?

ChayimFriedman2 (changes requested):

crates/ide/src/hover/render.rs
A4-Tacks: Const is not a assoc type, maybe `match type_or_const` is better than `match assoc_types`

crates/ide/src/hover.rs
A4-Tacks: I am confused about how to implement it, `TyKind` doesn't seem to have `Const`

crates/ide/src/hover/render.rs
ChayimFriedman2: `assoc_items`?

A4-Tacks: This may imply that there are also function items included

A4-Tacks:
Strangely, the rustbot said that the conflict comes from #22464, but in reality it comes from #22583

PR #21157, opened 2025-11-28
minor: Use 'deferred task' terminology consistently

Wilfred:
Previously there was a mix of 'queued task' and 'deferred task' terminology. The deferred terminology is more helpful (it provides a hint of which tasks belong there), so standardise on that term.

### DIFF
--- a/crates/rust-analyzer/src/task_pool.rs
+++ b/crates/rust-analyzer/src/task_pool.rs
@@ -6,7 +6,7 @@ use std::panic::UnwindSafe;
 use crossbeam_channel::Sender;
 use stdx::thread::{Pool, ThreadIntent};
 
-use crate::main_loop::QueuedTask;
+use crate::main_loop::DeferredTask;
 
 pub(crate) struct TaskPool<T> {
     sender: Sender<T>,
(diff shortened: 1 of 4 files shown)


crates/rust-analyzer/src/task_pool.rs
```diff
@@ -45,11 +45,10 @@ impl<T> TaskPool<T> {
...
 }
 
-/// `TaskQueue`, like its name suggests, queues tasks.
-///
-/// This should only be used if a task must run after [`GlobalState::process_changes`]
-/// has been called.
-pub(crate) struct TaskQueue {
-    pub(crate) sender: crossbeam_channel::Sender<QueuedTask>,
-    pub(crate) receiver: crossbeam_channel::Receiver<QueuedTask>,
+/// `DeferredTaskQueue` holds deferred tasks. These are tasks that
```
davidbarsky: one small nit: i'd keep the first line of the doc comment on a separate line that summarizes what this is.

after a new, empty line, add a paragraph describing what's happening.

Wilfred: Done.

PR #21164, opened 2025-11-28
fix: Allow multiple discover operations

Wilfred:
Previously, rust-analyzer would drop discover requests that arrived before we'd finished processing the previous request.

This is particularly noticeable when starting VS Code with multiple Rust files which require separate configuration discovery, as only the first file gets IDE features.

Fix this by allowing multiple discover requests to be active. Keep track of the number of discover operations for the quiescence check, and keep the process handles until they terminate.

### DIFF
--- a/crates/rust-analyzer/src/global_state.rs
+++ b/crates/rust-analyzer/src/global_state.rs
@@ -121,9 +121,10 @@ pub(crate) struct GlobalState {
     pub(crate) test_run_remaining_jobs: usize,
 
     // Project loading
-    pub(crate) discover_handle: Option<discover::DiscoverHandle>,
+    pub(crate) discover_handles: Vec<discover::DiscoverHandle>,
     pub(crate) discover_sender: Sender<discover::DiscoverProjectMessage>,
     pub(crate) discover_receiver: Receiver<discover::DiscoverProjectMessage>,
+    pub(crate) discover_jobs_active: u32,
 
     // Debouncing channel for fetching the workspace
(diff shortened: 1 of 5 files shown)


crates/rust-analyzer/src/global_state.rs
```diff
@@ -121,9 +122,10 @@ pub(crate) struct GlobalState {
     pub(crate) test_run_remaining_jobs: usize,
 
     // Project loading
-    pub(crate) discover_handle: Option<discover::DiscoverHandle>,
+    pub(crate) discover_handles: Vec<discover::DiscoverHandle>,
     pub(crate) discover_sender: Sender<discover::DiscoverProjectMessage>,
     pub(crate) discover_receiver: Receiver<discover::DiscoverProjectMessage>,
+    pub(crate) discover_jobs_active: AtomicU32,
```
Wilfred: I think this could actually be a plain u32. I saw the mutexes elsewhere in this struct but I'm not accessing this value from other threads.

Veykril: It should be a u32, we only write to it in `&mut self` paths, so there is no point in using an atomic here

Wilfred: Done :)

Veykril:
has been a while since that test flaked

Wilfred:
Huh, it's `syntax_highlighting::tests::syntax_highlighting_not_quadratic` failing and only on macOS. I'm pretty confident that isn't affected by these changes?

Wilfred:
Rebased to trigger another CI run.

PR #21187, opened 2025-12-01
Include operator overload occurrences in SCIP index

nicolas-guichard:
Operators were explicitly filtered out, both when filtering tokens to search definitions for and when searching for actual definitions.

Fixes rust-lang/rust-analyzer#21127

### DIFF
--- a/crates/ide/src/static_index.rs
+++ b/crates/ide/src/static_index.rs
@@ -10,7 +10,7 @@ use ide_db::{
     documentation::Documentation,
     famous_defs::FamousDefs,
 };
-use syntax::{AstNode, SyntaxKind::*, SyntaxNode, SyntaxToken, T, TextRange};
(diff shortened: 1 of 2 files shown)


crates/ide/src/static_index.rs
```diff
@@ -225,12 +225,6 @@ impl StaticIndex<'_> {
             show_drop_glue: true,
             minicore: MiniCore::default(),
         };
-        let tokens = tokens.filter(|token| {
-            matches!(
-                token.kind(),
-                IDENT | INT_NUMBER | LIFETIME_IDENT | T![self] | T![super] | T![crate] | T![Self]
-            )
-        });
```
nicolas-guichard, line 233: Note: I removed the SyntaxKind-based filtering here because I didn't want to explicitly list all overloadable operators and get_definitions already filters out tokens that don't map to a definition, but I can add it back (with a proper `SyntaxKind::is_overloadable_operator`) if that's an issue.

PR #21189, opened 2025-12-01
internal: Remove mdbook-toc usage

Wilfred:
Now that there's a table of contents rendered in the left sidebar, there doesn't seem to be much value in rendering a table of contents on the page too.

The sidebar TOC was added in mdbook 0.5:
https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#05-migration-guide


lnicola:
New version is up: https://rust-analyzer.github.io/book/other_editors.html

PR #21195, opened 2025-12-03
fix: More proc-macro-srv proto fixes

Veykril:


crates/proc-macro-srv/src/token_stream.rs
```diff
@@ -40,6 +40,13 @@ impl<S> TokenStream<S> {
         TokenStreamIter::new(self)
     }
 
+    pub fn as_single_group(&self) -> Option<&Group<S>> {
```
lnicola, line 43: @Veykril nb. this chunk has Windows newlines.

Veykril: huh, mustve been the edit completion bugging out or something 🤔

PR #21198, opened 2025-12-03
Fix nested expr missing semicolon in incomplete-let

A4-Tacks:
Example
---
```rust
fn main() {
    let x = &$0
}
```

**Before this PR**

```rust
fn main() {
    let x = &loop {
        $0
    }
}
```

**After this PR**

```rust
fn main() {
    let x = &loop {
        $0
    };
}
```


ChayimFriedman2 (approved):
Thanks!

PR #21199, opened 2025-12-03
internal: migrate `generate_delegate_trait` to SyntaxEditor api

Young-Flash:
The changes for this migration are a bit more extensive than before, as some method calls indirectly use `ted::*` api

part of https://github.com/rust-lang/rust-analyzer/issues/18285

### DIFF
--- a/crates/ide-assists/src/handlers/generate_delegate_trait.rs
+++ b/crates/ide-assists/src/handlers/generate_delegate_trait.rs
@@ -296,7 +296,7 @@ fn generate_impl(
                     .collect()
             });
 
-            let delegate = finalize_delegate(&delegate, assoc_items, false)?;
+            let delegate = finalize_delegate(&make, &delegate, assoc_items, false)?;
 
             let target_scope = ctx.sema.scope(strukt.strukt.syntax())?;
             let source_scope = ctx.sema.scope(bound_def.syntax())?;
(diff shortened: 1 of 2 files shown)


crates/ide-assists/src/handlers/generate_delegate_trait.rs
```diff
@@ -254,51 +256,47 @@ fn generate_impl(
     delegee: &Delegee,
     edition: Edition,
 ) -> Option<ast::Impl> {
+    let make = SyntaxFactory::without_mappings();
```
ShoyuVanilla, line 259: Don't we need mapping here?

Young-Flash: This assist generates entirely new code (a new impl block) rather than editing existing syntax. Mappings are needed when using `SourceChangeBuilder::make_editor()` to edit existing syntax trees, where we need to track the relationship between original and new nodes for features like snippet placeholders. Since we're doing pure code generation, the mapping overhead is unnecessary.

Here is my understanding, is that right?

ShoyuVanilla: Yeah, that's correct but given the long body of this assist, I'd prefer having a comment for that 😄

PR #21203, opened 2025-12-04
Add config hide placeholders type hints

A4-Tacks:
In the inferred type hints, expand the line too long. add config to disable it.

Example
---
```json
{"rust-analyzer.inlayHints.typeHints.hideInferredTypes": true}
```

```rust
use std::collections::HashMap;
fn foo(iter: Vec<Result<HashMap<String, String>, std::io::Error>>) {
    let output = iter.into_iter().collect::<Result<Vec<_>, _>>().unwrap();
}
```

**Before this PR**

```rust
let output: Vec<HashMap<String, String>> = iter.into_iter().collect::<Result<Vec<_ = HashMap<String, String>>, _ = Error>>().unwrap();
```

**After this PR**

```rust
let output: Vec<HashMap<String, String>> = iter.into_iter().collect::<Result<Vec<_>, _>>().unwrap();
```


A4-Tacks:
> I think I prefer the phrase to be positive, all other configs are this way.

There are already hideClosureInitialization, hideClosureParameter, and hideNamedConstructor, perhaps it would be better to maintain the same style with the hide prefix?

ChayimFriedman2 (approved):
Thanks!

PR #21205, opened 2025-12-04
 fix: Skip cfg attributes in macro input attribute stripping

eihqnh:
## Description:
 This fixes a regression introduced in commit f0e372c3b6 ("Rewrite attribute handling"), where cfg attributes
     were incorrectly counted during macro input stripping. 
  ## Problem

   When using attribute macros like [maybe-async-cfg](https://github.com/nvksv/maybe-async-cfg), if cfg attributes are present on the item, rust-analyzer may
   fail to strip the macro attribute itself from the input. This results in the macro being expanded recursively
   (e.g., `TestStruct` -> `TestStructAsync`(Correct) -> `TestStructAsyncAsync`(Wrong)).

522373810-313be124-ff2d-42d1-9dd0-fae782897561

522374103-e62f4dcf-f545-4c56-9ae6-8dceb8efb09e

   ## Cause

   The `macro_input_callback` function in `crates/hir-expand/src/cfg_process.rs` uses a counter (`item_tree_attr_id`)
   to identify which attribute to strip. This counter was being incremented for cfg attributes.

   However, `collect_item_tree_attrs` (which assigns the attribute IDs) filters out cfg attributes. This caused the
   counter in `macro_input_callbac`k to desynchronize with the actual attribute IDs, leading to the macro attribute
   not being identified and stripped.

## Fix

   Explicitly skip cfg attributes when determining whether to call `should_strip_attr()` in `macro_input_callback`.
   This ensures the attribute counter remains in sync with the item tree.
image
## Repro

```rust
#[maybe_async_cfg::maybe(sync(keep_self), async(feature = "async"))]
#[derive(Debug)]
pub struct TestStruct {
    pub value: u32,
}

#[cfg(feature = "async")]
pub fn test_async_only() {
    let _x = TestStructAsync { value: 42 };
}
#[cfg(not(feature = "async"))]
pub fn test_sync_only() {
    let _y = TestStruct { value: 1 };
}
pub fn test_both_exist() {
    let _a = TestStruct { value: 1 };

    let _b = TestStructAsync { value: 2 };
}
```
```toml
[package]
name = "test-maybe-async"
version = "0.1.0"
edition = "2024"

[dependencies]
maybe-async-cfg = "=0.2.4"

[features]
async = []
```

### DIFF
--- a/crates/hir-def/src/macro_expansion_tests/proc_macros.rs
+++ b/crates/hir-def/src/macro_expansion_tests/proc_macros.rs
@@ -341,3 +341,22 @@ struct Foo;
 #[helper_should_be_ignored] struct Foo;"#]],
     );
 }
+
+#[test]
+fn attribute_macro_stripping_with_cfg() {
+    check(
+        r#"
+//- proc_macros: generate_suffixed_type
+#[cfg(all())]
+#[proc_macros::generate_suffixed_type]
+struct S;
+"#,
+        expect![[r#"
+#[cfg(all())]
+#[proc_macros::generate_suffixed_type]
+struct S;
+
+struct S;
+struct SSuffix;"#]],
+    );
+}
--- a/crates/hir-expand/src/cfg_process.rs
+++ b/crates/hir-expand/src/cfg_process.rs
@@ -162,24 +162,18 @@ fn macro_input_callback(
                                     }
                                 }
                                 Meta::TokenTree { path, tt } => {
-                                    if path.segments.len() != 1
+                                    if path.is1("cfg") {
+                                        let cfg_expr = CfgExpr::parse_from_ast(
+                                            &mut TokenTreeChildren::new(&tt).peekable(),
+                                        );
+                                        if cfg_options().check(&cfg_expr) == Some(false) {
+                                            return ControlFlow::Break(ItemIsCfgedOut);
+                                        }
+                                        strip_current_attr = true;
+                                    } else if path.segments.len() != 1
                                         || !is_item_tree_filtered_attr(path.segments[0].text())
                                     {
                                         strip_current_attr = should_strip_attr();
-                                    }
-
-                                    if path.segments.len() == 1 {
-                                        let name = path.segments[0].text();
-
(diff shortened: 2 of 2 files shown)


eihqnh:
> Thanks, that's my mistake.
> 
> LGTM, but can you construct a test? (I understand if it's too hard).

I added a simple test. Could you give me some suggestions?
`cargo test -p hir-def macro_expansion_tests::proc_macros::attribute_macro_stripping_with_cfg`
## without this change
```
cargo test -p hir-def macro_expansion_tests::proc_macros::attribute_macro_stripping_with_cfg

running 1 test
test macro_expansion_tests::proc_macros::attribute_macro_stripping_with_cfg ... FAILED

failures:

---- macro_expansion_tests::proc_macros::attribute_macro_stripping_with_cfg stdout ----


error: expect test failed
   --> crates/hir-def/src/macro_expansion_tests/proc_macros.rs:354:9

You can update all `expect!` tests by running:

    env UPDATE_EXPECT=1 cargo test

To update a single test, place the cursor on `expect` token and use `run` feature of rust-analyzer.

Expect:
----
#[cfg(all())]
#[proc_macros::generate_suffixed_type]
struct S;

struct S;
struct SSuffix;
----

Actual:
----
#[cfg(all())]
#[proc_macros::generate_suffixed_type]
struct S;

----

Diff:
----
#[cfg(all())]
#[proc_macros::generate_suffixed_type]
struct S;

struct S;
struct SSuffix;
----

failures:
    macro_expansion_tests::proc_macros::attribute_macro_stripping_with_cfg

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 466 filtered out; finished in 0.02s

error: test failed, to rerun pass `-p hir-def --lib`
```
## with this change
```
test macro_expansion_tests::proc_macros::attribute_macro_stripping_with_cfg ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 466 filtered out; fin
```

ChayimFriedman2 (approved):
Thanks!

PR #21219, opened 2025-12-07
internal: Make `MacroCallId` a tracked struct

Veykril:


Veykril:
Runs into the same issue as making blocks tracked for now. We are still re-resolving too much stuff in the `hir` crate but we can't really get around that unless we start storing a `TypeRefId` -> `hir_ty::Ty` mapping

PR #21226, opened 2025-12-07
fix: fixed Impl display to show trait generic args

benodiwal:
Closes rust-lang/rust-analyzer#21196 

Screenshot 2025-12-08 at 4 42 49 AM

Screenshot 2025-12-08 at 4 43 55 AM


benodiwal:
> I don't want to use `trait_ref()` as it will print the hir-ty representation but I prefer the hir-def one, is there a reason you don't use `target_trait` as I suggested on the issue?

 Thanks! I initially used `trait_ref()` as it was the obvious high-level API with an existing HirDisplay impl. But you're right that `target_trait` is better here since it uses the `hir-def` representation rather than going through `hir-ty` type resolution which is more appropriate for display purposes and consistent with the rest of the function. There is not any sepcific reason I used this approach, I just thought you have suggested one way to do it and its not rigid.
 I have updated it now!

PR #21236, opened 2025-12-09
feat: add "Convert `as` to `From`" assist

benodiwal:
Closes rust-lang/rust-analyzer#21229

NOTE: Narrowing casts like `i64 as i32` won't show the assist. Hence `.try_into().unwrap()` conversion when only `TryFrom/TryInto` exists can be added as a follow-up enhancement through another PR.


ChayimFriedman2:

PR #21238, opened 2025-12-10
fix: is_transmutable always panicking

Colepng:
Fixes rust-lang/rust-analyzer#21237

Since the implementation is very large and nightly only. It's better to just return `Ok(Certainty::Yes)`, as per the discussion, on the issue.


ChayimFriedman2 (approved):
Thanks!

PR #21240, opened 2025-12-10
feat: show parameter hint for missing arguments

benodiwal:
Closes rust-lang/rust-analyzer#9486

Adds inlay hint for the next expected parameter when a function call has fewer arguments than parameters. Controlled by the new config option `inlayHints.parameterHints.missingArguments.enable` (default: false).

Screenshot 2025-12-10 at 5 03 13 PM

Screenshot 2025-12-10 at 5 20 03 PM


lnicola:
I'll have to give this a try, but it feels like something we might want to enable by default.

PR #21265, opened 2025-12-14
fix: respect rustc's lint attribute application order

J3m3:
# Description

Reverse the order of returned lint attributes for a `SyntaxNode` to match rustc's behavior.

When multiple lint attributes are present, rustc overrides earlier ones with the last defined attribute. The previous iteration order was incorrect, causing earlier attributes to override the later ones.

## Before this PR

image

## After this PR

image

## Is it worthwhile?

Generally, no one wants to specify same lints with different levels in a row. However, it is technically possible for users to use `-Zcrate-attr` to inject crate-level lint attributes at the top of the crate root file. This means any overlapping lints passed by `-Zcrate-attr` should not override the corresponding ones specified in the source file.

If we're trying to mimic the behavior of `-Zcrate-attr` when [introducing a new field in the `rust-project.json` format that can inject crate-level attributes](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Primitive.20type.20inherent.20method.20lookup.20fails/with/562983853), it would need to handle lint attributes in a similar fashion.

Overall, I'm not sure this is truly worthwhile, since neither scenario seems to be a common use case (even the decision to support lint attributes hasn't been made yet). That said, I don't see a good reason to leave it as-is, especially since the fix isn't difficult to implement.

## Reference setup for reproduction

`main.rs`
```rs
#![allow(unused)]
#![deny(unused)]

fn main() {
    let should_be_denied = 0;

    #[warn(unused)]
    #[allow(unused)]
    let should_be_allowed = 0;
}
```

`rust-project.json`
```json
{
    "sysroot": "/usr/local/rustup/toolchains/1.91.1-aarch64-unknown-linux-gnu",
    "sysroot_src": "/usr/local/rustup/toolchains/1.91.1-aarch64-unknown-linux-gnu/lib/rustlib/src/rust/library",
    "crates": [
        {
            "cfg": [],
            "deps": [],
            "edition": "2024",
            "is_proc_macro": false,
            "root_module": "/root/workspace/lint-attr-order/main.rs"
        }
    ]
}
```

`.vscode/settings.json`
```json
{
    "rust-analyzer.diagnostics.experimental.enable": true
}
```

### DIFF
--- a/crates/hir/src/semantics.rs
+++ b/crates/hir/src/semantics.rs
@@ -267,7 +267,7 @@ impl<DB: HirDatabase + ?Sized> Semantics<'_, DB> {
         &self,
         krate: Crate,
         item: ast::AnyHasAttrs,
-    ) -> impl Iterator<Item = (LintAttr, SmolStr)> {
+    ) -> impl DoubleEndedIterator<Item = (LintAttr, SmolStr)> {
         let mut cfg_options = None;
         let cfg_options = || *cfg_options.get_or_insert_with(|| krate.id.cfg_options(self.db));
         let mut result = Vec::new();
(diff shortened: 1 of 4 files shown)


J3m3:
Hmm... not sure why the test failed. IIUC, `syntax_highlighting::tests::syntax_highlighting_not_quadratic` doesn't even call `hir::semantics::Semantics::lint_attrs`.

Edit: I think it occasionally fails since it's a time-based test.

ChayimFriedman2 (approved):
Thanks!

PR #21270, opened 2025-12-15
Fix "Invariant violation: file emitted multiple times" when doing `scip .`

CatherineGasnier:
Summary:

It turns out the small piece of code I'm editing can and does add a file ID multiple times, which results in "Invariant violation: file emitted multiple times" when doing rust-analyzer scip . for some rust projects.

Test Plan:
```
cargo build --bin rust-analyzer --release
rust_analyzer=$(realpath target/release/rust-analyzer)
mytmp=$(mktemp -d)
cd $mytmp
git clone https://github.com/GraphiteEditor/Graphite.git
cd Graphite
$rust_analyzer scip .
```
Before: Some error like "Invariant violation: file emitted multiple times"
After: generates an index.scip as expected

### DIFF
--- a/crates/ide/src/static_index.rs
+++ b/crates/ide/src/static_index.rs
@@ -325,12 +325,12 @@ impl StaticIndex<'_> {
             };
             let mut visited_files = FxHashSet::default();
             for module in work {
-                let file_id = module.definition_source_file_id(db).original_file(db);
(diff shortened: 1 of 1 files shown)


CatherineGasnier:
cc @ChayimFriedman2 who added that code

PR #21271, opened 2025-12-15
Fix bugs with SCIP field enclosing_range

CatherineGasnier:
Summary: Previously, the enclosing range of a function name would be the range of the whole enclosing `impl`. I'm not sure exactly why, but the old code for enclosing_range was just not doing the right thing. The new doc comments should make it clear what we're trying to achieve in this new version.

Related PR: rust-lang/rust-analyzer#21141

Test Plan:
Added unit test.
Also tested on Graphite repo

### DIFF
--- a/crates/ide/src/static_index.rs
+++ b/crates/ide/src/static_index.rs
@@ -37,6 +37,7 @@ pub struct StaticIndex<'a> {
 pub struct ReferenceData {
     pub range: FileRange,
     pub is_definition: bool,
+    pub enclosing_range: Option<FileRange>,
 }
 
 #[derive(Debug)]
@@ -45,14 +46,8 @@ pub struct TokenStaticData {
     pub documentation: Option<Documentation<'static>>,
     pub hover: Option<HoverResult>,
     /// The position of the token itself.
-    ///
     /// For example, in `fn foo() {}` this is the position of `foo`.
     pub definition: Option<FileRange>,
-    /// The position of the entire definition that this token belongs to.
-    ///
-    /// For example, in `fn foo() {}` this is the position from `fn`
-    /// to the closing brace.
-    pub definition_body: Option<FileRange>,
     pub references: Vec<ReferenceData>,
     pub moniker: Option<MonikerResult>,
     pub display_name: Option<String>,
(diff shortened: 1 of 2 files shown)


CatherineGasnier:
ok, after further examination, I think the culprit was likely in `get_definitions` which has a fixme and probably does something weird with macros (the code I was testing on had macros). I'm going for a different approach in this new version, which avoids using the result of get_definitions, and also allows to compute the enclosing_range for all occurrences, not just occurrences that are definitions. I've also added a ton of tests. The previous implementation would fail more than half of those tests.

ChayimFriedman2:
This function *should* handle macros well. I think it'll help it you will describe exactly the issue you're facing.

CatherineGasnier:
> This function _should_ handle macros well. I think it'll help it you will describe exactly the issue you're facing.

We're maintaining the Glean indexer which is based on the SCIP index produced by rust-analyzer, and we've noticed the following issues upon indexing the Graphite repo:
* in https://github.com/GraphiteEditor/Graphite/blob/f4ec76f35e2ded8b2189a328805c5bceea569d2b/frontend/wasm/src/editor_api.rs, the enclosing_range for any function name like `dispatch` on line 121 is the whole range of the `JsEditorHandle` impl at line 106, i.e. range 104-743, instead of the full range of the `dispatch` function.
* https://github.com/GraphiteEditor/Graphite/blob/f4ec76f35e2ded8b2189a328805c5bceea569d2b/website/other/bezier-rs-demos/wasm/src/bezier.rs: same issue for functions of `impl WasmBezier`
* both these impl have `#[wasm_bindgen]`. We haven't examined other files for now.

I've tested this PR on that Graphite repo and it fixed those issues

ChayimFriedman2:
The original code is correct. A definition range that equals the entire macro call range can happen if the upmapped range of the function name is not contained in the upmapped range of the full function. So it's probably somehow because `wasm_bindgen` messes with the spans, although I can't tell how exactly.

ChayimFriedman2:
As I said this code likely isn't the culprit, and I also provided some hints to what it might be.

PR #21276, opened 2025-12-15
fix: use `cmark_with_options` to write shortcuts links to the output

dfireBird:
fixes: rust-lang/rust-analyzer#21214


ChayimFriedman2 (approved):
Thanks!

PR #21279, opened 2025-12-16
internal: add special `ErasedFileAstId` used for bypassing downmapping

J3m3:
Introduce `NO_DOWNMAP_ERASED_FILE_AST_ID_MARKER`, which prevents `Span`s from being mapped down into macro expansions.

This is a preparatory step for adding a new field to the `rust-project.json` format that can inject crate-level attributes. `Span`s for those attributes will be marked with
`NO_DOWNMAP_ERASED_FILE_AST_ID_MARKER`, indicating that they should not be mapped down into macro expansions.

### DIFF
--- a/crates/hir-def/src/macro_expansion_tests/mod.rs
+++ b/crates/hir-def/src/macro_expansion_tests/mod.rs
@@ -19,7 +19,7 @@ use std::{any::TypeId, iter, ops::Range, sync};
 use base_db::RootQueryDb;
 use expect_test::Expect;
 use hir_expand::{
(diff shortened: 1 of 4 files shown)


J3m3:
I've separated this into a dedicated PR since the next one would already be quite large. Hopefully this is the last preparatory step for the `-Zcrate-attr` feature. I'll open the next PR as soon as this one lands.

J3m3:
Thanks, that was quick! Let me spend some time to write a better commit message for the next PR...

PR #21298, opened 2025-12-18
feat: pretty print attributes up to `cfg(false)`

J3m3:
Extend ItemTree pretty-printing to include attributes even when a cfg expression evaluates to false, emitting explicit `cfg(cfg_expr)` marker after the attributes.

This improves testability of ItemTree attribute handling by exposing more information in pretty-printed output.


ChayimFriedman2 (approved):
Thanks!

PR #21306, opened 2025-12-20
fix: Auto completion should use fully qualified syntax in case of ambiguity #18895

asukaminato0721:
fix rust-lang/rust-analyzer#18895

Updated the completion renderer to detect when a trait method is being completed via a type path and rewrite the edit to the UFCS form (`<Ty as Trait>::method`) before emitting the completion item, including the new UfcsData/CallSnippetRewrite plumbing and trait-container detection in render_fn so both plain inserts and call-paren snippets respect the wider replacement range.

Extended add_call_parens to accept the optional rewrite data, build snippets using the qualified call head, and emit a TextEdit that spans the entire qualifier when necessary.

Added helper utilities to compute the qualifier text/range, render the trait path, and decide when UFCS rewriting is appropriate so the logic stays isolated and testable.

Added a regression test that exercises completing Dog:: when only a trait implementation exists to ensure the completion now expands to `<Dog as Animal>::baby_name()`.

### DIFF
--- a/crates/ide-completion/src/render/function.rs
+++ b/crates/ide-completion/src/render/function.rs
@@ -180,7 +180,9 @@ fn render(
         .detail(detail)
         .lookup_by(name.as_str().to_smolstr());
 
-    if let Some(data) = &ufcs_data {
+    if complete_call_parens.is_none()
+        && let Some(data) = &ufcs_data
+    {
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
As @Veykril said this PR is not correct.

PR #21319, opened 2025-12-22
internal: add merge imports indent tests

A4-Tacks:


crates/syntax/src/syntax_editor/edits.rs
```diff
@@ -214,18 +214,18 @@ impl Removable for ast::Use {
     fn remove(&self, editor: &mut SyntaxEditor) {
         let make = SyntaxFactory::without_mappings();
 
-        let next_ws = self
+        let prev_ws = self
```
ChayimFriedman2, line 476: Isn't this going to leave extra spaces in cases like this?
```rust
use foo; // to be removed

// other code
```

crates/ide-assists/src/handlers/merge_imports.rs
```diff
@@ -106,12 +106,21 @@ trait Merge: AstNode + Clone {
...
         let mut merged = self.clone();
+        let mut replace = self;
+        let mut update_replace = |item: Self| {
+            if item.syntax().text_range().start() < replace.syntax().text_range().start() {
+                std::mem::replace(&mut replace, item)
+            } else {
+                item
+            }
+        };
+
```
ChayimFriedman2: What does this code does? Please leave a comment.

ChayimFriedman2 (dismissed):

crates/syntax/src/syntax_editor/edits.rs
A4-Tacks: Keeping whitespace in this case is also a reasonable solution. Do we need a fallback?

Usually, removing the whitespace before rather than after is more conducive to indentation

ChayimFriedman2: But `use`s specifically are commonly the first child and followed by a blank line. I don't see a reason to degrade one scenario for the other. If you handle both then fine.

A4-Tacks: > But `use`s specifically are commonly the first child and followed by a blank line.

Uh, no, there are curly and whitespaces before it in ItemList, and there are often whitespaces before it in SourceFile as well

And the test utils will also add a whitespaces line at the beginning

> If you handle both then fine.

If the siblings before and after are deleted, it may easily cause overlapping deletion when the caller batch deletes

ChayimFriedman2: A simple heuristic is: remove the previous, and if the `use` is the first child and has blank lines after it remove them as well.

A4-Tacks: > I don't see a reason to degrade one scenario for the other.

In SourceFile, it is rare to truly remove all uses, so it is difficult to leave useless blank lines at the beginning of file

> A simple heuristic is: remove the previous, and if the `use` is the first child and has blank lines after it remove them as well.

It's possible to make overlapping edits, isn't it?

ChayimFriedman2: > In SourceFile, it is rare to truly remove all uses, so it is difficult to leave useless blank lines at the beginning of file

rustfmt will fix the indentation anyway for both ways. If you want to not need that, I won't stop you, but I don't want to regress other cases.

> It's possible to make overlapping edits, isn't it?

Nope. It's a bug, I think it will even panic.

A4-Tacks: > Nope. It's a bug, I think it will even panic.

If it is possible to delete both the before and after uses when deleting a use, then this API will be difficult to use because many uses involve batch deletion of adjacent uses

> I won't stop you, but I don't want to regress other cases.

Other cases are not common, and overall, this PR has improved trivias

ChayimFriedman2: > If it is possible to delete both the before and after uses when deleting a use, then this API will be difficult to use because many uses involve batch deletion of adjacent uses

What API? Only this assist need to delete things correctly.

> Other cases are not common, and overall, this PR has improved trivias

I've stated my opinion. @Veykril what is your opinion?

Veykril: Imo we shouldn't spend time on trying to fix whitespace at all as we cannot possibly get it right with the current infra without going through massive hoops. This all will be fixed by us having a proper formatter at some point.

People that do not use rustfmt simply have bad a time with r-a right now.

That is, we shouldn't try to fix indents, but newlines is a different matter though, as rustfmt does not necessarily normalize these. So we should check what happens with that here and if we regress in that regard or not, if yes we shouldn't do this.

A4-Tacks: At least for this assist, it replacing the first uses and deleting follows uses, there will be no blank line left on begin of file

I searched all the references, and only this one assist used this call, so there was no **any** regress

ChayimFriedman2:
Given @Veykril agrees with me this is not needed, I'll close this PR.

A4-Tacks:
But I noticed that there was no regress:

> if we regress in that regard or not, if yes we shouldn't do this.

ChayimFriedman2:
@Veykril As I and @A4-Tacks seem to disagree what to do with this PR here, I'll leave it to your judgment as the team lead.

ChayimFriedman2:
@Veykril this is awaiting your decision.

A4-Tacks:
It seems to have been fixed in #22351, I only retain tests

Veykril (approved):

Issue #4318, opened 2020-05-05
Display regular inlay hints at the end of the line


Alexendoo:
The addition of [type hints for chained method calls](https://github.com/rust-analyzer/rust-analyzer/pull/3710) is a fantastic feature, and it got me thinking that it would be nice to have the option of displaying regular hints like that, e.g.

```rust
let x = something; /* type hint */
```

rather than

```rust
let x/*: type hint */ = something;
```

A nice benefit of this would be your code would never be moved around by the type hints, either during edits or when RA has finished loading

It would also help with long type names and/or narrow windows, as types wouldn't push code off screen

bjorn3:
How should this work with the following?

```rust
let (x/*: type hint */, y/*: type hint */) = something;

**Comment by lnicola** [comment]
```rust
let (x, y) = something; /* (type hint, type hint)
```

But this might be a little awkward for irrefutable patterns:

```rust
let Foo(Bar(x /* i32 */)) = f();
let Foo(Bar(x)) = f(); /* Foo(Bar(i32)) */
```

iago-lito:
+1 here, because I think virtual text can only be shown at the end of lines in Vim, so we can enjoy the hint in:
```rust
let x = something; /* type hint */
```
but not in
```rust
let x/*: type hint */ = something;
```

For chained hints, we could end up with something like:
```rust
let n = map // Hashmap<String, String>
    .get("test".to_string()) // Option<&String>
    .map(|s| s.len())  // Option<usize>
    .unwrap(); // usize
```
But I'm not sure where the hint for inner `s` argument would be supposed to go. The more hints, the less room :\

Zooce:
This feature would be a game changer. Even if it couldn't handle nested hints like @iago-lito 's example above.

Zooce:
I don't know Typescript, but it looks like this _might_ be possible if the `createHintStyle` function in inlay_hints.ts had a way to change the start of the `vscode.DecorationOptions.range` parameter to be the length of the line containing the hint. Not sure if the corresponding inlay_hints.rs functions have a way to get this information.

faldor20:
This would be great. I hate how the type hints make things jump about while i'm typing so having it be on the end, like Ionide for f# would be great. Also makes it easier to read IMHO

kirillbobyrev:
[...]

PR #21322, opened 2025-12-22
impl Display type hint inlay hints at the end of the line #4318

asukaminato0721:
fix rust-lang/rust-analyzer#4318

Added a new rust-analyzer.inlayHints.typeHints.location setting (values inline/end_of_line, default inline) and plumbed it through the server config so clients can request end-of-line placement as described in issue.

Updated the binding-type inlay producer so that when end_of_line is selected it extends the hint range to the end of the surrounding let statement and forces a left pad, which makes the client render the type after the statement terminator.

### DIFF
--- a/crates/ide/src/inlay_hints/bind_pat.rs
+++ b/crates/ide/src/inlay_hints/bind_pat.rs
@@ -8,10 +8,12 @@ use ide_db::{RootDatabase, famous_defs::FamousDefs};
 
 use itertools::Itertools;
 use syntax::{
+    TextRange,
     ast::{self, AstNode, HasGenericArgs, HasName},
     match_ast,
 };
 
+use super::TypeHintsPlacement;
 use crate::{
     InlayHint, InlayHintPosition, InlayHintsConfig, InlayKind,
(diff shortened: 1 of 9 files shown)


Ahbar1999:
why is this still open ?

ChayimFriedman2:
@Ahbar1999 Because no-one from the team has reviewed this. We're all volunteers here.

Ahbar1999:
Dang it, it's a much needed feature I hope somebody reviews it soon

crates/ide/src/inlay_hints/bind_pat.rs
```diff
@@ -101,16 +105,26 @@ pub(super) fn hints(
...
+    }
     acc.push(InlayHint {
-        range: match type_ascriptable {
-            Some(Some(t)) => text_range.cover(t.text_range()),
-            _ => text_range,
-        },
+        range,
         kind: InlayKind::Type,
         label,
         text_edit,
```
Decodetalkers, line 125: Maybe it will be kind of weird if the inlay hint is put at the end of the line and still started with ":", I think the text should also be modified with the change of placing way

Issue #13370, opened 2022-10-09
Snippets don't dedent the reciever text


DropDemBits:
Applying the `dbg` snippet to the following code

```rs
fn main() {
            (1 +
                1).$0
}
```

results in the following text in VSCode:

```rs
fn main() {
            dbg!((1 +
                            1))
}
```

instead of the expected

```rs
fn main() {
            dbg!((1 +
                1))
}
```

Also, currently this `check_edit` passes even though the expected output doesn't match the real output (i.e. there's a mismatch somewhere in how edits are applied).

```rs
check_edit(
    "dbg",
    r#"
//- minicore: option
fn main() {
            (1 +
                1).$0
}
"#,
    r#"
fn main() {
            dbg!((1 +
                1))
}
"#,
);
```

---

**rust-analyzer version**: rust-analyzer version: 0.3.1229-standalone (5c28ad193 2022-10-01)

**rustc version**: rustc 1.64.0 (a55dd71d5 2022-09-19)

PR #21324, opened 2025-12-23
Fix postfix completion indentation compensation

A4-Tacks:
Editor adds the current indentation to the content of the code snippet

This PR dedentation is used to offset the editor snippet indentation

Fixes rust-lang/rust-analyzer#13370

Example
---
```rust
fn foo(x: Option<i32>, y: Option<i32>) {
    let _f = || {
        x
            .and(y)
            .map(|it| it+2)
            .$0
    };
}
```

**Before this PR**

```rust
fn foo(x: Option<i32>, y: Option<i32>) {
    let _f = || {
        let $0 = x
                        .and(y)
                        .map(|it| it+2);
    };
}
```

**After this PR**

```rust
fn foo(x: Option<i32>, y: Option<i32>) {
    let _f = || {
        let $0 = x
            .and(y)
            .map(|it| it+2);
    };
}
```

### DIFF
--- a/crates/ide-completion/src/completions/postfix.rs
+++ b/crates/ide-completion/src/completions/postfix.rs
@@ -355,12 +355,20 @@ fn get_receiver_text(
         range.range = TextRange::at(range.range.start(), range.range.len() - TextSize::of('.'))
     }
     let file_text = sema.db.file_text(range.file_id.file_id(sema.db));
-    let mut text = file_text.text(sema.db)[range.range].to_owned();
+    let text = file_text.text(sema.db);
+    let indent_spaces = indent_of_tail_line(&text[TextRange::up_to(range.range.start())]);
+    let mut text = stdx::dedent_by(indent_spaces, &text[range.range]);
 
     // The receiver texts should be interpreted as-is, as they are expected to be
     // normal Rust expressions.
     escape_snippet_bits(&mut text);
-    text
+    return text;
+
+    fn indent_of_tail_line(text: &str) -> usize {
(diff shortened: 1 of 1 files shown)


A4-Tacks:
> Are all editors indenting like that?

The language server protocol can adjust this behavior

- `InsertTextMode::AS_IS`
- `InsertTextMode::ADJUST_INDENTATION`

> It feels a bit wrong to change the indentation of existing code

A dedentation is the least disruptive implementation

> it's not a severe issue since formatting will fix that anyway...

Frequent formatting greatly affects the user experience,
And the range formatting does not work well when partial filling in the syntax tree

DropDemBits:
Small note that this would fixup #13370, thanks!

> The language server protocol can adjust this behavior
> 
> - `InsertTextMode::AS_IS`
> - `InsertTextMode::ADJUST_INDENTATION`


I'd rather go for the route for using `InsertTextMode::AS_IS` where possible instead of the default `InsertTextMode::ADJUST_INDENTATION`, that way we don't have to rely on the client editor's implementation of how it inserts indentation and we can likely just inspect the receiver text to the indentation needed to fixup multi-line snippets instead of looking at the entire file.

A4-Tacks:
> we can likely just inspect the receiver text to the indentation needed to fixup multi-line snippets instead of looking at the entire file.

I didn't look at the entire file, I only extracted the indentation of the current line

DropDemBits:
I meant needing to pull the file text, which I realized is something that we already do 😅

A4-Tacks:
Can this PR be merge now?

PR #21326, opened 2025-12-23
fix: keep flycheck generations in sync across multiple workspaces

Veykril:
The diagnostics collection globally tracks the generation for all loaded workspaces as its shared between them, yet the flycheck actors track their own separate generations per workspace. This mismatch could cause flycheck to not work correctly when multiple workspaces were loaded.


ShoyuVanilla (approved):
Oops, I made so many flounders in this feature 😅

PR #21334, opened 2025-12-25
internal: Introduce cargo-machete ci step

Veykril:


lnicola:
Only 5? We were in a pretty good place.

PR #21338, opened 2025-12-26
Add select stmts support for replace_let_with_if_let

A4-Tacks:
- ~~And migrate `edit_in_place::Indent` into `edit::AstNodeEdit`~~

Example
---
```rust
enum E<T> { X(T), Y(T) }

fn main() {
    $0let x = E::X(92);
    let y = x;$0
    let _ = ();
}
```

**Before this PR**

```
enum E<T> { X(T), Y(T) }

fn main() {
    if let x = E::X(92) {
    }
    let y = x;
    let _ = ();
}
```

**After this PR**

```rust
enum E<T> { X(T), Y(T) }

fn main() {
    if let x = E::X(92) {
        let y = x;
    }
    let _ = ();
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/replace_let_with_if_let.rs
+++ b/crates/ide-assists/src/handlers/replace_let_with_if_let.rs
@@ -39,6 +39,11 @@ pub(crate) fn replace_let_with_if_let(acc: &mut Assists, ctx: &AssistContext<'_>
     let let_stmt = let_kw.parent().and_then(ast::LetStmt::cast)?;
     let init = let_stmt.initializer()?;
     let original_pat = let_stmt.pat()?;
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
To me the more intuitive behavior will be to replace it with an if-let-chain, but I'd prefer neither, just not supporting this case.

A4-Tacks:
This supports any statement, not just `let`, such as replacing `let y = x;` with `fun(x)`

If necessary, refutable lets can be collected in if-let-chain in the future

ChayimFriedman2:
I understand that, but my claim is that because it is ambiguous we should not do it.

A4-Tacks:
You are right.

PR #21340, opened 2025-12-27
Refactor: Bidirectional messages

Shourya742:
This PR builds on the discussion in the Zulip thread: [#t-compiler/rust-analyzer > proc-macro-srv: bidirectional messaging](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/proc-macro-srv.3A.20bidirectional.20messaging/near/565116882)

 At a high level, it removes the rigid `FnMut` callback introduced in the initial bidirectional messaging implementation on the `srv` side, and instead introduces a `ProcMacroClientInterface` trait owned by the `srv`. This allows the `cli` to implement the interface for any transport, without the `srv` needing to know about transport details. The `srv` simply calls trait methods to obtain the data it needs.

 The PR also removes the duplicated `SubRequest` and `SubResponse` types added in the earlier bidirectional PR,

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1877,7 +1877,6 @@ name = "proc-macro-srv-cli"
 version = "0.0.0"
 dependencies = [
  "clap",
- "crossbeam-channel",
  "postcard",
  "proc-macro-api",
  "proc-macro-srv",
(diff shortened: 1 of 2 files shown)


Veykril (approved):
Nice looking better already

PR #21372, opened 2025-12-30
[WIP] Allow overriding paths to rust component binaries

Pandapip1:
My setup means that `cargo`, `rustup`, etc... are not usually in the path. This means I can't use the `rust-analyzer` extension since it assumes they are. This adds a few options to allow setting the absolute path to the binaries, like how the R language extension does it.

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -1602,6 +1602,66 @@
                 }
             },
             {
+                "title": "Environment",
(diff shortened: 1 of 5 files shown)


ChayimFriedman2:
You can set the `CARGO` environment variable, I don't think we need another way.

PR #21373, opened 2025-12-31
fix: Properly handle directories that are nested within excluded directories

ChayimFriedman2:
That is `/excluded` is excluded, and `/excluded/project` is the project.

### DIFF
--- a/crates/project-model/src/workspace.rs
+++ b/crates/project-model/src/workspace.rs
@@ -753,7 +753,6 @@ impl ProjectWorkspace {
 
                         let exclude = vec![
                             pkg_root.join(".git"),
-                            pkg_root.join("target"),
                             pkg_root.join("tests"),
                             pkg_root.join("examples"),
                             pkg_root.join("benches"),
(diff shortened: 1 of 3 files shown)


lnicola (approved):

ChayimFriedman2:
Fixed the bug; see the second commit.

ShoyuVanilla (approved):
I agree that not excluding `./target/` by default is the correct choice, as stated in the commit message. Excluding it would be even quite incorrect, since it is hard-coded rather than respecting the cargo config.

That said, I'm a bit concerned that this could overwhelm our vfs or project discovery in cases where large projects are cloned inside the target directory 😅

Veykril:
I am actually surprised we had `target` excluded to be honest 🤔 
> That said, I'm a bit concerned that this could overwhelm our vfs or project discovery in cases where large projects are cloned inside the target directory 😅

Yea that might cause issues to be fair. Can we exclude target but re-include the relevant subdirectories? (that is `target/<profile>/build`) To slight reduce this risk

ChayimFriedman2:
To be clear; `target` was *already* not excluded; the bug in the code prevented it from being.

PR #21374, opened 2025-12-31
fix: don't fire `non_camel_case_types` lint for structs/enums marked with `repr(C)`

Jefffrey:
If a struct/enum is marked with `repr(C)` and it doesn't follow camel case convention, rustc actually ignores this.

See corresponding rustc code:

https://github.com/rust-lang/rust/blob/94f8f4083ef82743d44ccb842c2bb65978efec28/compiler/rustc_lint/src/nonstandard_style.rs#L146-L153

And the original commit: https://github.com/rust-lang/rust/commit/c26cd9f05dc20eb67cf90d2ebb18728807b53022


ChayimFriedman2 (approved):
Thanks!

PR #21376, opened 2025-12-31
Allow finding references from doc comments

SomeoneToIgnore:
https://github.com/user-attachments/assets/7f27574c-991f-4a12-9ad0-0bc2c273d8b9

### DIFF
--- a/crates/ide/src/references.rs
+++ b/crates/ide/src/references.rs
@@ -38,7 +38,8 @@ use syntax::{
 };
 
 use crate::{
-    Analysis, FilePosition, HighlightedRange, NavigationTarget, TryToNav, highlight_related,
+    Analysis, FilePosition, HighlightedRange, NavigationTarget, TryToNav,
+    doc_links::token_as_doc_comment, highlight_related,
 };
 
 /// Result of a reference search operation.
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
I have a concern: this does not find the most obvious reference - where the cursor is. And all other references in doc comments.

SomeoneToIgnore:
My main use case for the current impl is to look up code whilst keeping the view of the documentation in the current editor, and did not expect the documentation links to appear there.

Also, I suspect that if we start showing all docs' references, that will make the reference lookup even slower and bloated with results?

So, personally, I'd love to keep things as they are, as I am rarely interested in docs' usages, but very interested in the real code related.

I do not know internals good enough, but suspect that if we want to track those references in docs as proposed, we need to go over almost all the "stack" starting with lowering?
I'm not sure I will be able to do that, and neither I am interested right now: I want to make completions to work inside doc comments and this was a practice PR before that.

Veykril (approved):

Veykril:
> I have a concern: this does not find the most obvious reference - where the cursor is. And all other references in doc comments.

This would be ideal behavior, but shouldn't block this PR. Its an improvement over the current state after all.

PR #21377, opened 2025-12-31
proc-macro-srv: support file and local_file via bidirectional callbacks

Shourya742:
This PR extends the proc-macro bidirectional protocol to support Span::file() and Span::local_file().

### DIFF
--- a/crates/load-cargo/src/lib.rs
+++ b/crates/load-cargo/src/lib.rs
@@ -540,7 +540,7 @@ impl ProcMacroExpander for Expander {
         current_dir: String,
     ) -> Result<tt::TopSubtree, ProcMacroExpansionError> {
         let mut cb = |req| match req {
-            SubRequest::LocalFileName { file_id } => {
+            SubRequest::LocalFilePath { file_id } => {
                 let file = FileId::from_raw(file_id);
                 let source_root_id = db.file_source_root(file).source_root_id(db);
                 let source_root = db.source_root(source_root_id).source_root(db);
@@ -548,9 +548,9 @@ impl ProcMacroExpander for Expander {
                 let name = source_root
                     .path_for_file(&file)
                     .and_then(|path| path.as_path())
-                    .and_then(|path| path.file_name().map(|filename| filename.to_owned()));
+                    .map(|path| path.to_string());
 
-                Ok(SubResponse::LocalFileNameResult { name })
+                Ok(SubResponse::LocalFilePathResult { name })
             }
             SubRequest::SourceText { file_id, start, end } => {
                 let file = FileId::from_raw(file_id);
@@ -558,7 +558,7 @@ impl ProcMacroExpander for Expander {
                 let slice = text.get(start as usize..end as usize).map(ToOwned::to_owned);
                 Ok(SubResponse::SourceTextResult { text: slice })
             }
-            SubRequest::FileName { file_id } => {
+            SubRequest::FilePath { file_id } => {
                 let file = FileId::from_raw(file_id);
                 let source_root_id = db.file_source_root(file).source_root_id(db);
                 let source_root = db.source_root(source_root_id).source_root(db);
@@ -566,10 +566,10 @@ impl ProcMacroExpander for Expander {
                 let name = source_root
                     .path_for_file(&file)
                     .and_then(|path| path.as_path())
-                    .and_then(|path| path.file_name().map(|filename| filename.to_owned()))
+                    .map(|path| path.to_string())
                     .unwrap_or_default();
 
-                Ok(SubResponse::FileNameResult { name })
+                Ok(SubResponse::FilePathResult { name })
             }
         };
         match self.0.expand(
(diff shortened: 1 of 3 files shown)


crates/load-cargo/src/lib.rs
```diff
@@ -540,12 +540,40 @@ impl ProcMacroExpander for Expander {
...
                 let slice = text.get(start as usize..end as usize).map(ToOwned::to_owned);
                 Ok(SubResponse::SourceTextResult { text: slice })
             }
+            SubRequest::FileName { file_id } => {
+                let file = FileId::from_raw(file_id);
+                let source_root_id = db.file_source_root(file).source_root_id(db);
+                let source_root = db.source_root(source_root_id).source_root(db);
+
+                let name = source_root.path_for_file(&file).and_then(|path| {
+                    path.name_and_extension().map(|(name, ext)| match ext {
```
lnicola: Nit: doesn't `path.file_name()` work?

crates/load-cargo/src/lib.rs
```diff
@@ -540,12 +540,40 @@ impl ProcMacroExpander for Expander {
...
     ) -> Result<tt::TopSubtree, ProcMacroExpansionError> {
         let mut cb = |req| match req {
+            SubRequest::LocalFileName { file_id } => {
+                let file = FileId::from_raw(file_id);
+                let source_root_id = db.file_source_root(file).source_root_id(db);
+                let source_root = db.source_root(source_root_id).source_root(db);
+
+                let name = source_root
+                    .path_for_file(&file)
+                    .and_then(|path| path.clone().into_abs_path())
```
lnicola: Nit:

```rust
                    .and_then(|path| path.as_path())
                    .and_then(|path| path.file_name().map(|filename| filename.to_owned()));
```

crates/load-cargo/src/lib.rs
bit-aloo: `filename` can be an `AbsPath` or a `VirtualPath` and doesn’t need to exist locally. local_filename must be an AbsPath that exists on disk.

With `path.file_name()` we only get `AbsPath` which only appears if file exist locally.

crates/load-cargo/src/lib.rs
bit-aloo: Thanks updated.

crates/load-cargo/src/lib.rs
lnicola: Ah, because there's no `VirtualPath::file_name`.

Veykril (commented):
Why do local file name and file name differ for us here? I'd expect both of the functions to return the same results for us, reading the corresponding function docs https://doc.rust-lang.org/stable/proc_macro/struct.Span.html#method.file and https://doc.rust-lang.org/stable/proc_macro/struct.Span.html#method.local_file

bit-aloo:
> Why do local file name and file name differ for us here? I'd expect both of the functions to return the same results for us, reading the corresponding function docs https://doc.rust-lang.org/stable/proc_macro/struct.Span.html#method.file and https://doc.rust-lang.org/stable/proc_macro/struct.Span.html#method.local_file

Ahh, right, fetch the local filename first, and if that doesn’t exist, fall back to manually constructing it. I was too focused on the display-side logic for the filename and totally missed the obvious. Thanks for pointing it out!

Veykril:
No my point was more that for rust-analyzer, both functions will effectively return the same info. rustc has some extra stuff like path remapping and what not that are only handled in one of the two functions, we don't though.

Also ntoably, both functions are supposed to return full paths, not just file names.

bit-aloo:
> No my point was more that for rust-analyzer, both functions will effectively return the same info. rustc has some extra stuff like path remapping and what not that are only handled in one of the two functions, we don't though.
> 
> Also ntoably, both functions are supposed to return full paths, not just file names.

~~Are you suggesting we don't return a file name, when absolute path doesn't exists? because currently, we first try to get the path, and if it doesn't exist, just return the filename. And considering `filename` method is just for display purpose, so that should be kinda ok? Otherwise, we gonna return empty string, which won't be much informative.~~

bit-aloo:
> No my point was more that for rust-analyzer, both functions will effectively return the same info. rustc has some extra stuff like path remapping and what not that are only handled in one of the two functions, we don't though.
> 
> Also ntoably, both functions are supposed to return full paths, not just file names.

After some more thought, the suggestion clicked. I’ve aligned filename and localfilename so they now emit the same info. Thanks for the suggestion, it helped me understand this part better.

Veykril:
There is still some slight confusion here I think, your implementation returns the file name only, but it should really return the entire file path.

That is, these are also misnamed as they are about the full path, not just the file name.

bit-aloo:
> There is still some slight confusion here I think, your implementation returns the file name only, but it should really return the entire file path.
> 
> That is, these are also misnamed as they are about the full path, not just the file name.

Shoot, thanks for mentioning that. Ah!!! my bad... so sorry, I completely mixed it with filename for some reason. Updated.

PR #21380, opened 2026-01-01
feat: Implement fine grained client side request cancellation support

Veykril:
cc https://github.com/salsa-rs/salsa/pull/1007

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3754,3 +3754,15 @@ dependencies = [
  "memchr",
  "time",
 ]
+
+[[patch.unused]]
+name = "salsa"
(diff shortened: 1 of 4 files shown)


crates/hir-ty/src/tests/regression.rs
```diff
@@ -2363,6 +2363,7 @@ fn test() {
 }
 "#,
         expect![[r#"
+            46..49 'Foo': Foo<N>
```
Veykril, line 2366: For some reason the salsa upgrade made this pop up? Its not wrong but I am also not sure why that now happens, bugfixes I guess :)

PR #21385, opened 2026-01-02
internal: Parallelize proc macro expansion

Shourya742:
This PR replaces the single proc-macro server per workspace with a small pool of server processes.

### DIFF
--- a/crates/proc-macro-api/src/lib.rs
+++ b/crates/proc-macro-api/src/lib.rs
@@ -155,8 +155,15 @@ impl ProcMacroClient {
         version: Option<&Version>,
         num_process: usize,
     ) -> io::Result<ProcMacroClient> {
-        let process = ProcMacroServerProcess::spawn(process_path, env, version)?;
-        Ok(ProcMacroClient { process: Arc::new(process), path: process_path.to_owned() })
+        let pool_size = num_process;
+        let mut workers = Vec::with_capacity(pool_size);
+        for _ in 0..pool_size {
+            let worker = ProcMacroServerProcess::spawn(process_path, env.clone(), version)?;
+            workers.push(worker);
+        }
+
+        let pool = ProcMacroServerPool::new(workers);
+        Ok(ProcMacroClient { pool: Arc::new(pool), path: process_path.to_owned() })
     }
 
     /// Invokes `spawn` and returns a client connected to the resulting read and write handles.
@@ -170,11 +177,20 @@ impl ProcMacroClient {
             Box<dyn process::ProcessExit>,
             Box<dyn io::Write + Send + Sync>,
             Box<dyn io::BufRead + Send + Sync>,
-        )>,
+        )> + Clone,
         version: Option<&Version>,
+        num_process: usize,
     ) -> io::Result<ProcMacroClient> {
-        let process = ProcMacroServerProcess::run(spawn, version, || "<unknown>".to_owned())?;
-        Ok(ProcMacroClient { worker: Arc::new(process), path: process_path.to_owned() })
+        let pool_size = num_process;
+        let mut workers = Vec::with_capacity(pool_size);
+        for _ in 0..pool_size {
+            let worker =
+                ProcMacroServerProcess::run(spawn.clone(), version, || "<unknown>".to_owned())?;
+            workers.push(worker);
+        }
+
+        let pool = ProcMacroServerPool::new(workers);
+        Ok(ProcMacroClient { pool: Arc::new(pool), path: process_path.to_owned() })
     }
 
     /// Returns the absolute path to the proc-macro server.
--- a/crates/proc-macro-api/src/process.rs
+++ b/crates/proc-macro-api/src/process.rs
@@ -210,9 +210,8 @@ impl ProcMacroServerProcess {
         callback: Option<SubCallback<'_>>,
     ) -> Result<Result<Vec<(String, ProcMacroKind)>, String>, ServerError> {
         match self.protocol {
-            Protocol::LegacyJson { .. } | Protocol::LegacyPostcard { .. } => {
-                legacy_protocol::find_proc_macros(self, dylib_path)
-            }
+            Protocol::LegacyJson { .. } => legacy_protocol::find_proc_macros(self, dylib_path),
+
             Protocol::BidirectionalPostcardPrototype { .. } => {
                 let cb = callback.expect("callback required for bidirectional protocol");
                 bidirectional_protocol::find_proc_macros(self, dylib_path, cb)
@@ -279,7 +278,7 @@ impl ProcMacroServerProcess {
         let result = match self.protocol {
             Protocol::LegacyJson { .. } => legacy_protocol::expand(
                 proc_macro,
-                    self,
+                self,
                 subtree,
                 attr,
                 env,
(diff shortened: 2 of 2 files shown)


crates/proc-macro-api/src/process.rs
```diff
@@ -29,6 +33,7 @@ pub(crate) struct ProcMacroServerProcess {
     protocol: Protocol,
     /// Populated when the server exits.
     exited: OnceLock<AssertUnwindSafe<ServerError>>,
+    active: AtomicU32,
```
Veykril: Why  do we have this, given a single server is still sequential, this will always be either 0 or 1.

crates/proc-macro-api/src/pool.rs
```diff
@@ -0,0 +1,120 @@
...
+        Self { workers: workers.into() }
+    }
+}
+
+impl ProcMacroServerPool {
+    pub(crate) fn exited(&self) -> Option<&ServerError> {
+        for worker in &*self.workers {
+            worker.exited()?;
+        }
+        self.workers[0].exited()
```
Veykril, line 27:
```suggestion
        None
```

crates/proc-macro-api/src/pool.rs
```diff
@@ -0,0 +1,120 @@
+//! This module represents Process Pool
```
Veykril:
```suggestion
//! A pool of proc-macro server processes
```

crates/proc-macro-api/src/pool.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    fn pick_process(&self) -> Result<&ProcMacroServerProcess, ServerError> {
+        self.workers
+            .iter()
+            .filter(|w| w.exited().is_none())
+            .min_by_key(|w| w.number_of_active_req())
+            .ok_or_else(|| ServerError {
+                message: "all proc-macro server workers have exited".into(),
+                io: None,
+            })
```
Veykril: If all servers have a request running right now, any new caller that wants to expand will end up picking and blocking on the same worker (first or last, whatever min by key returns in this case).

crates/proc-macro-api/src/lib.rs
```diff
@@ -217,21 +209,7 @@ impl ProcMacro {
...
-            subtree = subtree_changed.view();
-
-            if let Some(attr) = &mut attr {
-                attr_changed = tt::TopSubtree::from_subtree(*attr);
-                self.change_fixup_to_match_old_server(&mut attr_changed);
-                *attr = attr_changed.view();
-            }
-        }
-
-        self.process.expand(
```
Veykril, line 272: We can keep this transform here isntead of moving that to the pool I think?

crates/proc-macro-api/src/lib.rs
```diff
@@ -186,8 +178,8 @@ impl ProcMacro {
         self.kind
     }
 
-    fn needs_fixup_change(&self) -> bool {
-        let version = self.process.version();
+    fn needs_fixup_change(&self, process: &ProcMacroServerProcess) -> bool {
+        let version = process.version();
```
Veykril: It would make sense to cache the version in the pool I think

crates/proc-macro-api/src/process.rs
```diff
@@ -40,7 +45,7 @@ pub(crate) enum Protocol {
 
 /// Maintains the state of the proc-macro server process.
 #[derive(Debug)]
-struct ProcessSrvState {
+pub(crate) struct ProcessSrvState {
```
Veykril, line 82: is this change needed?

crates/proc-macro-api/src/pool.rs
bit-aloo: We would like to return exit status when all the process has exited? 
If any worker is still running -> return None. If all workers have exited -> return the exit error of worker 0

Veykril: Oh, wait this is an option 🤦 Yea I misread this

crates/proc-macro-api/src/pool.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    pub(crate) fn pick_process(&self) -> Result<&ProcMacroServerProcess, ServerError> {
+        self.workers
+            .iter()
+            .filter(|w| w.exited().is_none())
+            .min_by_key(|w| w.number_of_active_req())
+            .ok_or_else(|| ServerError {
+                message: "all proc-macro server workers have exited".into(),
+                io: None,
+            })
```
Veykril:
```suggestion

    let mut min = None;
    self.workers
        .iter()
        .filter(|w| w.exited().is_none())
        .find(|w| {
            let number_of_active_req= w.number_of_active_req();
            if number_of_active_req== 0 {
                return true;
            }
            min = Some(min.map_or(number_of_active_req, |min| min.min(number_of_active_req)));
            false
        })
        .or(min)
        .ok_or_else(|| ServerError {
            message: "all proc-macro server workers have exited".into(),
            io: None,
        });
```
something like this as a slight perf improvement (really miniscule), we don't need to look at every process, if one of them has no requests its a minimum already.

crates/proc-macro-api/src/pool.rs
```diff
@@ -0,0 +1,83 @@
...
+            .collect())
+    }
+
+    pub(crate) fn version(&self) -> u32 {
+        self.version
+    }
+}
+
+pub(crate) fn default_pool_size() -> usize {
```
Veykril: Let's make this configurable, similar to how `cachePriming_numThreads` is configurable (just for processes instead of threads), so
```diff
#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
#[serde(rename_all = "snake_case")]
- pub enum NumThreads {
+ pub enum NumProcesses {
    Physical,
-    Logical,
    #[serde(untagged)]
    Concrete(usize),
}
```

Veykril: and let's make the default `Concrete(1)` for now so we can have people test this as an opt-in for a bit

crates/proc-macro-api/src/pool.rs
bit-aloo: Done, makes sense for early return.

crates/proc-macro-api/src/pool.rs
bit-aloo: I implemented this, with one caveat: in a few CLI commands we don’t instantiate Config. In those cases I didn’t instantiate config through just to fetch the process count and instead defaulted to 1.

ChayimFriedman2:
This PR is based on requesting expansion from the least overloaded worker. I wonder, now that the protocol is bidirectional, can we instead make the proc macro server request "more work" from r-a? This will more evenly parallelize in case some expansion takes a long time.

bit-aloo:
> This PR is based on requesting expansion from the least overloaded worker. I wonder, now that the protocol is bidirectional, can we instead make the proc macro server request "more work" from r-a? This will more evenly parallelize in case some expansion takes a long time.

While working on the current pool design, I also considered a client-side approach closer to work stealing or availability-based scheduling. In this model, the client would wait until any worker becomes free and then dispatch the next request. This keeps scheduling centralized and fits naturally with the existing pool setup. I can add this in a follow-up.

If we instead move scheduling to the server side, things get more involved. The client would need a dedicated listener thread to receive subrequest/request signals from the servers and dispatch work accordingly, along with a way to track request completion. One nice property of this approach is that it pushes the architecture toward a more asyncy type model,  the client could fire-and-forget requests and later query the listener thread when it actually needs the result.

Veykril:
Do you mean the client should batch up multiple expansion requests? That makes things  fairly complicated again, just like multiplexing the connection to allow multiple inflight expansions (due to proc-macros observing the process environment). I think parallelization should for now be done via multiple processes as that can live separately from the protocol and is easy enough to implement.

Veykril:
🤦 completely forgot i wanted to merge this on monday, guess tomorrow it is then

lnicola (approved):

PR #21389, opened 2026-01-02
Fix some TryEnum reference assists

A4-Tacks:
- Fix `convert_to_guarded_return`
- Fix `replace_let_with_if_let`
- Fix `replace_if_let_with_match`

Example
---
```rust
fn foo() -> &'static Option<i32> {
    &None
}

fn main() {
    let x$0 = foo();
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
fn foo() -> &'static Option<i32> {
    &None
}

fn main() {
    let Some(x) = foo() else { return };
}
```

---

```rust
fn main(action: Action) {
    $0let x = compute();
}

fn compute() -> &'static Option<i32> { &None }
```

**Before this PR**

```rust
fn main(action: Action) {
    if let x = compute() {
    }
}

fn compute() -> &'static Option<i32> { &None }
```

**After this PR**

```rust
fn main(action: Action) {
    if let Some(x) = compute() {
    }
}

fn compute() -> &'static Option<i32> { &None }
```

---

```rust
fn foo(x: &Option<i32>) {
    $0if let Some(x) = x {
        println!("{}", x)
    } else {
        println!("none")
    }
}
```

**Before this PR**

```rust
fn foo(x: &Option<i32>) {
    match x {
        Some(x) => println!("{}", x),
        _ => println!("none"),
    }
}
```

**After this PR**

```rust
fn foo(x: &Option<i32>) {
    match x {
        Some(x) => println!("{}", x),
        None => println!("none"),
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/desugar_try_expr.rs
+++ b/crates/ide-assists/src/handlers/desugar_try_expr.rs
@@ -61,7 +61,7 @@ pub(crate) fn desugar_try_expr(acc: &mut Assists, ctx: &AssistContext<'_>) -> Op
     let expr = try_expr.expr()?;
     let expr_type_info = ctx.sema.type_of_expr(&expr)?;
 
-    let try_enum = TryEnum::from_ty_without_strip(&ctx.sema, &expr_type_info.original)?;
+    let try_enum = TryEnum::from_ty(&ctx.sema, &expr_type_info.original)?;
 
     let target = try_expr.syntax().text_range();
     acc.add(


A4-Tacks:
> Actually, I think I prefer to put the `strip_references()` in `TryEnum::from_ty()`.

I'm not sure if it will damage `crates/ide-assists/src/handlers/desugar_try_expr.rs`

Can I extract a method like `from_ty_without_strip()`?

ChayimFriedman2:
I actually think it's good even there. True, `?` is not valid with references, but it can make life easier for people and can't hurt.

ChayimFriedman2 (approved):

PR #21392, opened 2026-01-02
internal: Enable `jemalloc` for Linux glibc releases

lnicola:
- `x86_64-pc-windows-msvc` is currently excluded by https://github.com/rust-lang/rust-analyzer/blob/master/crates/rust-analyzer/Cargo.toml#L86
 - `i686-pc-windows-msvc` seems to be excluded in the same way, but actually fails to build:

```
  configure: error: in `/d/a/rust-analyzer/rust-analyzer/target/i686-pc-windows-msvc/release/build/tikv-jemalloc-sys-feae94799be1c2c3/out/build':
  configure: error: C compiler cannot create executables
  See `config.log' for more details

  thread 'main' (7844) panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\tikv-jemalloc-sys-0.5.4+5.3.0-patched\build.rs:351:9:
  command did not execute successfully: "sh" "D:/a/rust-analyzer/rust-analyzer/target/i686-pc-windows-msvc/release/build/tikv-jemalloc-sys-feae94799be1c2c3/out/build/configure" "--disable-cxx" "--enable-doc=no" "--enable-shared=no" "--with-jemalloc-prefix=_rjem_" "--with-private-namespace=_rjem_" "--host=i686-pc-win32" "--build=x86_64-pc-win32" "--prefix=D:\\a\\rust-analyzer\\rust-analyzer\\target\\i686-pc-windows-msvc\\release\\build\\tikv-jemalloc-sys-feae94799be1c2c3\\out"
  expected success, got: exit code: 77
```

 - `x86_64-unknown-linux-musl` fails for some reason with:

```
  running: cd "/__w/rust-analyzer/rust-analyzer/target/x86_64-unknown-linux-musl/release/build/tikv-jemalloc-sys-3abac4942835b5d1/out/build" && "make" "-j" "4"

  --- stderr

  thread 'main' (2014) panicked at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tikv-jemalloc-sys-0.5.4+5.3.0-patched/build.rs:347:19:
  failed to execute command: No such file or directory (os error 2)
```
- `aarch64-apple-darwin` fails with SIGSEGV:
```
Error: cannot generate PGO profiles

Caused by:
    command was terminated by a signal `target/aarch64-apple-darwin/release/rust-analyzer analysis-stats -q --run-all-ide-things /Users/runner/work/rust-analyzer/rust-analyzer/rust-analyzer-pgo/clap-rs-clap`: 11
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -30,12 +30,14 @@ jobs:
             target: x86_64-pc-windows-msvc
             code-target: win32-x64
             pgo: clap-rs/clap@v4.5.36
+            allocator: jemalloc
           - os: windows-latest
             target: i686-pc-windows-msvc
             pgo: clap-rs/clap@v4.5.36
           - os: windows-latest
             target: aarch64-pc-windows-msvc
             code-target: win32-arm64
+            allocator: jemalloc
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
             # Use a container with glibc 2.28
@@ -43,15 +45,18 @@ jobs:
             container: quay.io/pypa/manylinux_2_28_x86_64
             code-target: linux-x64
             pgo: clap-rs/clap@v4.5.36
+            allocator: jemalloc
           - os: ubuntu-24.04-arm
             target: aarch64-unknown-linux-gnu
             container: quay.io/pypa/manylinux_2_28_aarch64
             code-target: linux-arm64
             pgo: clap-rs/clap@v4.5.36
+            allocator: jemalloc
           - os: ubuntu-latest
             target: arm-unknown-linux-gnueabihf
             zig_target: arm-unknown-linux-gnueabihf.2.28
             code-target: linux-armhf
+            allocator: jemalloc
           - os: macos-14
             target: x86_64-apple-darwin
             code-target: darwin-x64
@@ -67,6 +72,8 @@ jobs:
 
     env:
       RA_TARGET: ${{ matrix.target }}
+      ALLOCATOR: ${{ matrix.allocator && format('--allocator {0}', matrix.allocator) || '' }}
+      PGO: ${{ matrix.pgo && format('--pgo {0}', matrix.pgo) || '' }}
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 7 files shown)


ChayimFriedman2:
@lnicola Do you want to look into the failures?

lnicola:
@ChayimFriedman2 no, I'm happy enabling it just for x64 Linux by default. I just have to clean this up a bit and maybe figure out why the feature flag doesn't get passed on `x86_64-pc-windows-msvc`. It won't make a difference because `jemallocator` is gated by `[target.'cfg(not(target_env = "msvc"))'.dependencies]`, but I didn't see anything in `xtask dist` that would drop it.

kkocdko:
Yeah it improved a lot, but I have a silly question: if jemalloc was integrated into rust-analyzer, will that prevent users from using their own allocator?

Currently I use jemalloc by LD_PRELOAD, with some tweaks:

```sh
# > allocator: jemalloc
export MALLOC_CONF="background_thread:true,dirty_decay_ms:2000,muzzy_decay_ms:2000,percpu_arena:phycpu,metadata_thp:auto" # stats_print:true
export LD_PRELOAD=/media/kkocdko/KK_TMP_1/usr/lib/libjemalloc.so.2 # 5.3.1
```

ChayimFriedman2:
@kkocdko If you mean from using LD_PRELOAD like you, I believe that yes. However there will also be less reason to do that.

kkocdko:
IMO, dynamic link to `libjemalloc.so.2` will be better, this allows user to use LD_PRELOAD, I tried `rustflags = ["-C", "link-arg=-ljemalloc"]` (may needs rpath), it works fine, but then we needs to ship two files? Seems not a good idea. I see that `zed-editor` already switch back from mimalloc to system default allocator due to the memory overhead.

Or, only enable built-in jemalloc on static musl target?

ChayimFriedman2:
It's possible to compile r-a with a different allocator. I don't think we need more than that.

lnicola:
Test run: https://github.com/lnicola/rust-analyzer/actions/runs/32018125453, https://github.com/lnicola/rust-analyzer/releases/tag/nightly

ChayimFriedman2:
This is fine from my side. @rust-lang/rust-analyzer does someone object?

lnicola:
Maybe we should also enable it for the dashboard? I'm not sure.

PR #21393, opened 2026-01-02
feat: Add location links for generic parameter type hints

benodiwal:
Closes rust-lang/rust-analyzer#21010

### DIFF
--- a/crates/ide/src/inlay_hints/bind_pat.rs
+++ b/crates/ide/src/inlay_hints/bind_pat.rs
@@ -1258,7 +1258,7 @@ where
     }
 
     #[test]
-    fn generic_param_inlay_hint_has_location_link() {
+    fn type_param_inlay_hint_has_location_link() {
         check_expect(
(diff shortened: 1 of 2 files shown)


crates/ide/src/inlay_hints/bind_pat.rs
```diff
@@ -1255,4 +1256,40 @@ where
 "#,
         );
     }
+
+    #[test]
+    fn generic_param_inlay_hint_has_location_link() {
+        check_expect(
+            InlayHintsConfig { type_hints: true, ..DISABLED_CONFIG },
+            r#"
+fn identity<T>(t: T) -> T {
```
Veykril, line 1265: lets put a const parameter and lifetime parameter here as well to test them all

benodiwal:
@Veykril added support for `const parameter` and `lifetime parameter`

crates/ide/src/inlay_hints/bind_pat.rs
benodiwal: added the tests

PR #21394, opened 2026-01-02
Implement Span::line() and Span::column() for proc-macro server

tilladam:
Add proper line/column resolution for proc-macro spans via a callback mechanism. Previously these methods returned hardcoded 1 values.

The implementation adds:

SubRequest::LineColumn and SubResponse::LineColumnResult to the bidirectional protocol
ProcMacroClientInterface::line_column() method
Callback handling in load-cargo using LineIndex
Server implementation in RaSpanServer that uses the callback

This is in service of resolving an issue with Slint: https://github.com/slint-ui/slint/issues/685

The Slint side now implement this also as per:
https://github.com/slint-ui/slint/commit/13ce62d9b26306c8dddabaeb7f635daa4b4f315e

Owing to my limited experience with Rust (and rust-analyzer) this was created with help from Claude. I have attempted to steer and review to the best of my ability and attempted sufficient test coverage.


Veykril:
Note that the implementations here are still unused by rust-analyzer as we are currently hashing out the concrete protocol for the bidirectional stuff. So this landing on stable will still take a while. Additionally, `line` and `column` are very iffy for us, as they completely nuke incrementality if a macro makes use of them.

Will also need a rebase.

crates/load-cargo/src/lib.rs
```diff
@@ -546,6 +546,18 @@ impl ProcMacroExpander for Expander {
                 let slice = text.get(start as usize..end as usize).map(ToOwned::to_owned);
                 Ok(SubResponse::SourceTextResult { text: slice })
             }
+            SubRequest::LineColumn { file_id, offset } => {
+                let file = FileId::from_raw(file_id);
+                let text = db.file_text(file).text(db);
+                let line_index = ide_db::line_index::LineIndex::new(text);
+                let line_col =
+                    line_index.try_line_col(ide_db::line_index::TextSize::new(offset)).unwrap();
```
Veykril: This unwrap can technically be hit if the proc-macro server is buggy so we should handle this failure instead of panicking within rust-analyzer

crates/proc-macro-api/src/bidirectional_protocol/msg.rs
```diff
@@ -11,11 +11,19 @@ use crate::{
...
 #[derive(Debug, Serialize, Deserialize)]
 pub enum SubResponse {
-    SourceTextResult { text: Option<String> },
+    SourceTextResult {
+        text: Option<String>,
+    },
+    /// Line and column are 1-based.
+    LineColumnResult {
+        line: usize,
+        column: usize,
```
Veykril:
```suggestion
        line: u32,
        column: u32,
```
usize is unnecessary here I believe, neither can be bigger than u32 given files are limited to 4gb in Rust (afaik)

crates/proc-macro-srv-cli/src/main_loop.rs
```diff
@@ -192,6 +192,31 @@ impl<C: Codec> proc_macro_srv::ProcMacroClientInterface for ProcMacroClientHandl
...
+            _ => return None,
+        };
+
+        match msg {
+            bidirectional::BidirectionalMessage::SubResponse(
+                bidirectional::SubResponse::LineColumnResult { line, column },
+            ) => Some((line, column)),
+            _ => None,
+        }
+    }
```
Veykril: This can be simplified now (you'll notice when rebasing atop latest master)

crates/proc-macro-srv/src/server_impl/rust_analyzer_span.rs
```diff
@@ -263,14 +263,16 @@ impl server::Span for RaSpanServer<'_> {
...
+        self.callback.as_mut().and_then(|cb| cb.line_column(file_id, offset)).map_or(1, |(l, _)| l)
     }
 
-    fn column(&mut self, _span: Self::Span) -> usize {
-        // FIXME requires db to resolve line index, THIS IS NOT INCREMENTAL
-        1
+    fn column(&mut self, span: Self::Span) -> usize {
+        let file_id = span.anchor.file_id.file_id().index();
+        let offset: u32 = span.range.start().into();
+        self.callback.as_mut().and_then(|cb| cb.line_column(file_id, offset)).map_or(1, |(_, c)| c)
```
Veykril: This is incorrect, the range in a span is relative to the anchored item (the AstId).
Which I also just realized means we have done the same mistake in `source_text` cc @Shourya742 
It would probably make more sense to just sent the entire span over the protocol here instead of taking it apart.

bit-aloo: If we want to send spans over the protocol, they need to be serde-compatible. This came up while I was implementing the other file-path related methods, and it’s essentially the only requirement for putting spans on the wire. Do we want to make Span serde-enabled?

Veykril: Alternatively we can deconstruct them into their parts and send all of that over and reconstruct. Point is we need to send the ast-id over too as the range is otherwise not useful.

bit-aloo: I will update the source root text.

PR #21397, opened 2026-01-03
Fix source text

Shourya742:
This PR does a bunch of stuff:

* Switch from `SourceDB` to `ExpandDB`, which provides access to AST/node information needed for anchored spans.
* ~~Make `Span` serde-compatible so it can be sent over the wire. (API's become nicer)~~
* ~~Simplify proc-macro-srv callbacks and `SubRequest` variants to carry only `Span`.~~
* Fix source-text resolution by interpreting span ranges relative to their anchored AST items.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2648,7 +2648,6 @@ dependencies = [
  "la-arena 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc-hash 2.1.1",
  "salsa",
- "serde",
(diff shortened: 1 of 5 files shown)


bit-aloo:
Shoot, span brings in salsa with itself. Reverting serde from span.

Veykril (approved):
Got some smaller thoughts about some things here but I'll try to address them myself as a follow up as I am not entirely clear on something here either

PR #21399, opened 2026-01-04
fix: Properly lower `SelfOnly` predicates

ShoyuVanilla:
Fixes https://github.com/rust-lang/rust-analyzer/issues/19339

In the following code:
```rust
trait Foo {
    type Assoc1;
    type Assoc2;
}

struct Bar {}

trait Baz {}
```

And we have some trait bounds like `Self: Foo<Assoc1 = Bar, Assoc2: Baz>`, rustc considers

- the projection predicate `<Self as Foo>::Assoc1 = Bar` as a self bound
- the assoc bound predicate `<Self as Foo>::Assoc2: Baz` as a non-self bound

https://github.com/rust-lang/rust/blob/00f49d22042dce59561b36eb075bbf5e343c23c4/compiler/rustc_hir_analysis/src/hir_ty_lowering/bounds.rs#L547-L668

As I have previously stated in some FIXME comment 😅, we and rustc has a bit different predicates lowering implementation: rustc lowers them multiple times with granular filtering options, while we lower them as little as possible.
I think our current implementation makes sense for memory usages and we don't need [much granular filtering options like rustc](https://doc.rust-lang.org/1.92.0/nightly-rustc/rustc_hir_analysis/hir_ty_lowering/enum.PredicateFilter.html) (yet) so I just reordered those assoc item bounds at the end of the predicates and discern them by range.

### DIFF
--- a/crates/hir-ty/src/lower.rs
+++ b/crates/hir-ty/src/lower.rs
@@ -77,6 +77,7 @@ pub struct ImplTraits {
 #[derive(PartialEq, Eq, Debug, Hash)]
 pub struct ImplTrait {
     pub(crate) predicates: StoredClauses,
+    pub(crate) assoc_ty_bounds_start: u32,
 }
 
 pub type ImplTraitIdx = Idx<ImplTrait>;
@@ -164,6 +165,12 @@ impl<'db> LifetimeElisionKind<'db> {
         // FIXME: We should use the elided lifetime here, or `ElisionFailure`.
         LifetimeElisionKind::Elided(Region::error(interner))
     }
+}
+
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub(crate) enum GenericPredicateSource {
+    SelfOnly,
+    AssocTyBound,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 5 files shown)


crates/hir-ty/src/lower.rs
```diff
@@ -1699,18 +1760,33 @@ pub(crate) fn type_alias_bounds<'db>(
...
+) -> EarlyBinder<'db, &'db [Clause<'db>]> {
+    let (bounds, assoc_ty_bounds_start, _) = type_alias_bounds_with_diagnostics(db, type_alias);
+    bounds.map_bound(|it| &it.as_slice()[..assoc_ty_bounds_start as usize])
+}
+
+fn type_alias_bounds_with_diagnostics<'db>(
     db: &'db dyn HirDatabase,
     type_alias: TypeAliasId,
-) -> (EarlyBinder<'db, Clauses<'db>>, Diagnostics) {
+) -> (EarlyBinder<'db, Clauses<'db>>, u32, Diagnostics) {
```
ChayimFriedman2: Return a struct? The role of the `u32` isn't clear from the outside.

crates/hir-ty/src/lower.rs
```diff
@@ -1759,6 +1849,7 @@ pub struct GenericPredicates {
     // then the implicit trait predicate for the child, if `is_trait` is `true`.
```
ChayimFriedman2, line 1859: Update this comment?

crates/hir-ty/src/lower.rs
```diff
@@ -1827,6 +1933,14 @@ impl GenericPredicates {
     /// Returns the predicates, minus the implicit `Self: Trait` predicate for a trait.
```
ChayimFriedman2: Update this comment?

crates/hir-ty/src/lower.rs
```diff
@@ -1999,8 +2120,12 @@ where
     add_implicit_trait_predicate(interner, def, predicate_filter, &mut predicates, &mut is_trait);
 
     let diagnostics = create_diagnostics(ctx.diagnostics);
+    let assoc_ty_bounds_start = predicates.len() as u32;
+    predicates.extend(assoc_ty_bounds);
```
ChayimFriedman2: This will mean `query_own()` will get the assoc type bounds from the parent as well, which isn't what we want. I think you'll need to have two `assoc_ty_bounds_start`s.

ShoyuVanilla: Oh, right. I'm messing up bounds from parents here 😅

PR #21400, opened 2026-01-04
internal: Clean up proc-macro-srv callback trait

Veykril:


Veykril:
cc @Shourya742 slight follow up to your PR, pushing some things around

crates/base-db/src/editioned_file_id.rs
```diff
@@ -26,6 +26,9 @@ const _: () = {
         krate: Crate,
     }
 
+    // FIXME: This poses an invalidation problem, if one constructs an `EditionedFileId` with a
+    // different crate then whatever the input of a memo used, it will invalidate the memo causing
+    // it to recompute even if the crate is not really used.
```
Veykril, line 31: cc @ChayimFriedman2 unsure if i mentioned that problem in the PR review that introduced this API, but something to be aware of. If we always use `guess_origin` it won't really prose a problem, but otherwise we end up needlessly invalidating entries and recompute which is unforutnate

crates/proc-macro-srv/src/lib.rs
```diff
@@ -94,9 +95,9 @@ impl<'env> ProcMacroSrv<'env> {
 pub type ProcMacroClientHandle<'a> = &'a mut (dyn ProcMacroClientInterface + Sync + Send);
 
 pub trait ProcMacroClientInterface {
-    fn file(&mut self, file_id: u32) -> String;
-    fn source_text(&mut self, file_id: u32, ast_id: u32, start: u32, end: u32) -> Option<String>;
-    fn local_file(&mut self, file_id: u32) -> Option<String>;
+    fn file(&mut self, file_id: span::FileId) -> String;
+    fn source_text(&mut self, span: Span) -> Option<String>;
+    fn local_file(&mut self, file_id: span::FileId) -> Option<String>;
```
Veykril, line 100: This is the main interesting change. This API should cleanly use the corresponding types so that the caller (the server impl) does not have to care about how the RPC ends up looking, that's up to the RPC impl how these should taken apart

crates/base-db/src/editioned_file_id.rs
ChayimFriedman2: That's not really a problem because when a file already exists in the db we will pick the existing crate.

Veykril: It is a problem if you turn a span's `EditionedFileId` into the salsa variation without going through `relevant_crates` (what the guess origin function does), since then its on the caller to supply a crate

ChayimFriedman2: The caller supply a new crate only the def map. All other places derive an already-existing file id from the def map.

crates/proc-macro-srv-cli/Cargo.toml
```diff
@@ -18,6 +18,7 @@ clap = {version = "4.5.42", default-features = false, features = ["std"]}
 
 [features]
 default = []
+# default = ["sysroot-abi"]
```
bit-aloo, line 21: Can be removed?

PR #21403, opened 2026-01-04
fix: Suppress false positive missing assoc item diag on specialization

ShoyuVanilla:
Fixes https://github.com/rust-lang/rust-analyzer/issues/21378

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -920,6 +920,48 @@ impl Module {
                     }
                 }
 
+                // HACK: When specialization is enabled in the current crate, and there exists
+                // *any* blanket impl that provides a default implementation for the missing item,
+                // suppress the missing associated item diagnostic.
+                // This can lead to false negatives when the impl in question does not actually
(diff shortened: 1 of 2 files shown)


crates/hir/src/lib.rs
```diff
@@ -920,6 +920,48 @@ impl Module {
...
+                // specialize that blanket impl, but determining the exact specialization
+                // relationship here would be significantly more expensive.
+                if !missing.is_empty() {
+                    let krate = self.krate(db).id;
+                    let def_map = crate_def_map(db, krate);
+                    if def_map.is_unstable_feature_enabled(&sym::specialization)
+                        || def_map.is_unstable_feature_enabled(&sym::min_specialization)
+                    {
+                        missing.retain(|(assoc_name, assoc_item)| {
+                            let AssocItem::Function(_) = assoc_item else {
```
ChayimFriedman2, line 936: Technically the full `specialization` supports specialization of types as well (and also `default impl`), but I don't think we need to support that.

PR #21404, opened 2026-01-04
proc-macro-api: Add more tests

Veykril:
All of these were AI generated and then hand-picked/cleaned up by me

### DIFF
--- a/crates/tt/src/storage.rs
+++ b/crates/tt/src/storage.rs
@@ -381,18 +381,12 @@ pub struct TopSubtree {
 
 impl TopSubtree {
     pub fn empty(span: DelimSpan) -> Self {
-        Self {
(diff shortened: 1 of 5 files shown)


crates/tt/src/storage.rs
```diff
@@ -381,18 +381,12 @@ pub struct TopSubtree {
...
-                CompressedSpanPart::from_span(&span.open),
-                CompressedSpanPart::from_span(&span.close),
-            ]),
-        }
+        TopSubtreeBuilder::new(crate::Delimiter {
+            kind: DelimiterKind::Invisible,
+            open: span.open,
+            close: span.close,
+        })
+        .build()
```
Veykril, line 389: Notable change, given that the eq impl of `TopSubtree` relies on the backing span store, we should make sure that building an empty tree results in the same object no matter what constructor is being used, thus this change

PR #21405, opened 2026-01-04
Implement Span::line() and Span::column() for proc-macro server

tilladam:
Add proper line/column resolution for proc-macro spans via a callback mechanism. Previously these methods returned hardcoded 1 values.

The implementation adds:

SubRequest::LineColumn and SubResponse::LineColumnResult to the bidirectional protocol
ProcMacroClientInterface::line_column() method
Callback handling in load-cargo using LineIndex
Server implementation in RaSpanServer that uses the callback
Test infrastructure and a test for LineColumn

This is in service of resolving an issue with Slint: https://github.com/slint-ui/slint/issues/685

The Slint side now implement this also as per:
https://github.com/slint-ui/slint/commit/13ce62d9b26306c8dddabaeb7f635daa4b4f315e

Owing to my limited experience with Rust (and rust-analyzer) this was created with help from Claude. I have attempted to steer and review to the best of my ability and attempted sufficient test coverage.

### DIFF
--- a/crates/load-cargo/src/lib.rs
+++ b/crates/load-cargo/src/lib.rs
@@ -554,20 +554,30 @@ impl ProcMacroExpander for Expander {
                 Ok(SubResponse::LocalFilePathResult { name })
             }
             SubRequest::SourceText { file_id, ast_id, start, end } => {
-                let ast_id = span::ErasedFileAstId::from_raw(ast_id);
-                let editioned_file_id = span::EditionedFileId::from_raw(file_id);
-                let span = Span {
-                    range: TextRange::new(TextSize::from(start), TextSize::from(end)),
-                    anchor: SpanAnchor { file_id: editioned_file_id, ast_id },
-                    ctx: SyntaxContext::root(editioned_file_id.edition()),
-                };
-                let range = db.resolve_span(span);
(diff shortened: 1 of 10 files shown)


crates/load-cargo/src/lib.rs
```diff
@@ -554,21 +554,35 @@ impl ProcMacroExpander for Expander {
...
+                let line_col =
+                    line_index.try_line_col(range.range.start()).ok_or_else(|| ServerError {
+                        message: format!("invalid offset {offset} for file {file_id}"),
+                        io: None,
+                    })?;
+                // proc_macro::Span line/column are 1-based
+                Ok(SubResponse::LineColumnResult {
+                    line: line_col.line as u32 + 1,
+                    column: line_col.col as u32 + 1,
+                })
```
bit-aloo: We can return a default value instead of propagating an error, given how things are structured right now. If an error occurs in the callback, we end up interrupting the client request while the server continues to block waiting for a response. This is something we’ll improve, but since we’re still prototyping, returning a default value is the simplest and safest option for now.

tilladam: {1,1} then as in the previous implementation?

Veykril (approved):

PR #21408, opened 2026-01-05
default_field_values

kouhe3:
fix rust-lang/rust-analyzer#19780 

Add enum and field to track `=` and `..`

Create a new function to determine match and missing msg

behavior
```rust
#![feature(default_field_values)]
fn main() {
    let u1 = UserInfo { id: 10, $0 ,.. }; // trigger suggestion: age,email
               ^^^^^^ 💡 missing field 
                    | -email
}

struct UserInfo {
    id: i32,
    age: f32 = 1.0,
    email: String,
}
```

### DIFF
--- a/crates/hir-def/src/expr_store.rs
+++ b/crates/hir-def/src/expr_store.rs
@@ -32,7 +32,7 @@ use crate::{
     expr_store::path::Path,
     hir::{
         Array, AsmOperand, Binding, BindingId, Expr, ExprId, ExprOrPatId, Label, LabelId, Pat,
-        PatId, RecordFieldPat, Statement,
+        PatId, RecordFieldPat, RecordSpread, Statement,
     },
     nameres::{DefMap, block_def_map},
     type_ref::{LifetimeRef, LifetimeRefId, PathId, TypeRef, TypeRefId},
@@ -575,8 +575,8 @@ impl ExpressionStore {
                 for field in fields.iter() {
                     f(field.expr);
                 }
-                if let &Some(expr) = spread {
-                    f(expr);
+                if let RecordSpread::Expr(expr) = spread {
+                    f(*expr);
                 }
             }
             Expr::Closure { body, .. } => {
@@ -706,8 +706,8 @@ impl ExpressionStore {
                 for field in fields.iter() {
                     f(field.expr);
                 }
-                if let &Some(expr) = spread {
-                    f(expr);
+                if let RecordSpread::Expr(expr) = spread {
+                    f(*expr);
                 }
             }
             Expr::Closure { body, .. } => {
--- a/crates/hir-def/src/expr_store/pretty.rs
+++ b/crates/hir-def/src/expr_store/pretty.rs
@@ -16,7 +16,8 @@ use crate::{
     attrs::AttrFlags,
     expr_store::path::{GenericArg, GenericArgs},
     hir::{
-        Array, BindingAnnotation, CaptureBy, ClosureKind, Literal, Movability, Statement,
+        Array, BindingAnnotation, CaptureBy, ClosureKind, Literal, Movability, RecordSpread,
+        Statement,
         generics::{GenericParams, WherePredicate},
     },
     lang_item::LangItemTarget,
@@ -139,7 +140,7 @@ pub fn print_variant_body_hir(db: &dyn DefDatabase, owner: VariantId, edition: E
     }
 
     for (_, data) in fields.fields().iter() {
-        let FieldData { name, type_ref, visibility, is_unsafe } = data;
+        let FieldData { name, type_ref, visibility, is_unsafe, default_value: _ } = data;
         match visibility {
             crate::item_tree::RawVisibility::Module(interned, _visibility_explicitness) => {
                 w!(p, "pub(in {})", interned.display(db, p.edition))
@@ -679,10 +680,17 @@ impl Printer<'_> {
                         p.print_expr(field.expr);
                         wln!(p, ",");
                     }
-                    if let Some(spread) = spread {
-                        w!(p, "..");
-                        p.print_expr(*spread);
-                        wln!(p);
+                    match spread {
+                        RecordSpread::None => {}
+                        RecordSpread::FieldDefaults => {
+                            w!(p, "..");
+                            wln!(p);
+                        }
+                        RecordSpread::Expr(spread_expr) => {
+                            w!(p, "..");
+                            p.print_expr(*spread_expr);
+                            wln!(p);
+                        }
                     }
                 });
                 w!(self, "}}");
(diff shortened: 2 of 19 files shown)


crates/hir-def/src/hir.rs
```diff
@@ -260,6 +260,7 @@ pub enum Expr {
         path: Option<Box<Path>>,
         fields: Box<[RecordLitField]>,
         spread: Option<ExprId>,
+        has_spread_syntax: bool,
```
ChayimFriedman2: I think it's better to make the existing `spread` an `enum Spread { None, Default, Spread(ExprId) }` (the names are up to you).

ChayimFriedman2 (changes requested):

crates/hir-def/src/hir.rs
kouhe3: i will try

crates/ide-completion/src/completions/record.rs
kouhe3: Due to miss != match commit. I suggest here use new record_literal_matched_fields  function

crates/hir-def/src/hir.rs
```diff
@@ -187,6 +187,13 @@ impl From<ast::LiteralKind> for Literal {
     }
 }
 
+#[derive(Debug, Clone, Eq, PartialEq)]
```
ChayimFriedman2: Derive `Copy`.

crates/hir-def/src/hir.rs
```diff
@@ -187,6 +187,13 @@ impl From<ast::LiteralKind> for Literal {
...
+pub enum RecordSpread {
+    None,
+    Empty,
```
ChayimFriedman2: I dislike calling this `Empty` as it does not explain what it does, especially to users unfamiliar with the unstable feature. However I don't really have a good name... rustc calls this `Rest` (in the context of them calling the enum `StructRest`) which is also not a good name. Perhaps `pub enum RecordSpread { None, FieldDefaults, FromBase(ExprId) }`?

crates/hir-def/src/signatures.rs
```diff
@@ -754,6 +754,7 @@ pub struct FieldData {
     pub type_ref: TypeRefId,
     pub visibility: RawVisibility,
     pub is_unsafe: bool,
+    pub has_default_value: bool,
```
ChayimFriedman2: We will likely want the default for some things, so it makes sense to store an `Option<ExprId>` here and lower the expression.

crates/hir-def/src/expr_store.rs
```diff
@@ -575,8 +575,8 @@ impl ExpressionStore {
                 for field in fields.iter() {
                     f(field.expr);
                 }
-                if let &Some(expr) = spread {
-                    f(expr);
+                if let RecordSpread::Expr(ref expr) = *spread {
```
ChayimFriedman2: Redundant `ref` then `*`, ditto below.

crates/hir-def/src/expr_store/pretty.rs
```diff
@@ -139,7 +140,7 @@ pub fn print_variant_body_hir(db: &dyn DefDatabase, owner: VariantId, edition: E
     }
 
     for (_, data) in fields.fields().iter() {
-        let FieldData { name, type_ref, visibility, is_unsafe } = data;
+        let FieldData { name, type_ref, visibility, is_unsafe, .. } = data;
```
ChayimFriedman2: Please ignore only specific fields (no `..` patterns, only `field: _`), we do not want to miss new fields here.

Also we probably want to print the default value, but we can't since we don't store it - relevant to what I said in the other comment.

crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -571,25 +571,70 @@ pub fn record_literal_missing_fields(
...
+        })
         .collect();
     if missed_fields.is_empty() {
         return None;
     }
-    Some((variant_def, missed_fields, exhaustive))
+    Some((variant_def, missed_fields))
+}
+
+pub fn record_literal_matched_fields(
```
ChayimFriedman2: This and its friend `record_pattern_matched_fields()` don't belong in this file, `record_literal_missing_fields()` and `record_pattern_missing_fields()` are here because they're used by diagnostics.

Also "missing" vs. "matched" doesn't really capture the distinction, both are both missing and **non**-matched (which I suppose was the intent). Something like `missing_considering_spread` and `missing_without_spread` is better.

crates/hir-ty/src/infer/mutability.rs
```diff
@@ -132,8 +133,11 @@ impl<'db> InferenceContext<'_, 'db> {
             Expr::Become { expr } => {
                 self.infer_mut_expr(*expr, Mutability::Not);
             }
-            Expr::RecordLit { path: _, fields, spread } => {
-                self.infer_mut_not_expr_iter(fields.iter().map(|it| it.expr).chain(*spread))
+            Expr::RecordLit { path: _, fields, spread, .. } => {
+                self.infer_mut_not_expr_iter(fields.iter().map(|it| it.expr));
+                if let RecordSpread::Expr(expr) = *spread {
+                    self.infer_mut_not_expr_iter(std::iter::once(expr));
```
ChayimFriedman2:
```suggestion
                    self.infer_mut_expr(expr, Mutability::Not);
```

crates/hir-ty/src/mir/lower.rs
```diff
@@ -867,16 +867,16 @@ impl<'a, 'db> MirLowerCtx<'a, 'db> {
...
+                let spread_place = match *spread {
+                    RecordSpread::Expr(it) => {
                         let Some((p, c)) = self.lower_expr_as_place(current, it, true)? else {
                             return Ok(None);
                         };
                         current = c;
                         Some(p)
                     }
-                    None => None,
+                    RecordSpread::None | RecordSpread::Empty => None,
```
ChayimFriedman2: I think we should separate them, and for `RecordSpread::Empty` cause `not_supported!()`, because it needs a different implementation.

crates/ide-completion/src/completions/record.rs
```diff
@@ -90,7 +90,13 @@ pub(crate) fn complete_record_expr_fields(
                 item.add_to(acc, ctx.db);
                 return;
             }
-            missing_fields
+            if update_exists {
```
ChayimFriedman2: This `if` should not exist here, instead call `record_literal_matched_fields()` to begin with.

ChayimFriedman2 (changes requested):

crates/hir-ty/src/diagnostics/expr.rs
kouhe3: I will try to move to source_analyzer

PR #21410, opened 2026-01-06
Make proc-macro bidirectional calls cancellation safe

Shourya742:
Catch panics in subrequest callbacks, reply with Cancel, and propagate cancellation to the proc-macro server

### DIFF
--- a/crates/proc-macro-srv-cli/src/main_loop.rs
+++ b/crates/proc-macro-srv-cli/src/main_loop.rs
@@ -214,7 +214,7 @@ fn handle_failure(failure: Result<bidirectional::SubResponse, ProcMacroClientErr
     }
 }
 
-impl<C: Codec> proc_macro_srv::ProcMacroClientInterface for ProcMacroClientHandle<'_, C> {
+impl proc_macro_srv::ProcMacroClientInterface for ProcMacroClientHandle<'_> {
     fn file(&mut self, file_id: proc_macro_srv::span::FileId) -> String {
         match self.roundtrip(bidirectional::SubRequest::FilePath { file_id: file_id.index() }) {
             Ok(bidirectional::SubResponse::FilePathResult { name }) => name,


crates/proc-macro-api/src/bidirectional_protocol.rs
```diff
@@ -56,9 +57,17 @@ pub fn run_conversation<C: Codec>(
                 return Ok(BidirectionalMessage::Response(response));
             }
             BidirectionalMessage::SubRequest(sr) => {
-                let resp = callback(sr)?;
-                let reply = BidirectionalMessage::SubResponse(resp);
-                let encoded = C::encode(&reply).map_err(wrap_encode)?;
+                let resp = match catch_unwind(AssertUnwindSafe(|| callback(sr))) {
```
ChayimFriedman2, line 61: I would slightly prefer to have the callback type `+ UnwindSafe` than this `AssertUnwindSafe`.

crates/proc-macro-srv-cli/src/main_loop.rs
```diff
@@ -204,20 +221,32 @@ impl<C: Codec> proc_macro_srv::ProcMacroClientInterface for ProcMacroClientHandl
             start: range.start().into(),
             end: range.end().into(),
         }) {
-            Some(bidirectional::BidirectionalMessage::SubResponse(
-                bidirectional::SubResponse::SourceTextResult { text },
-            )) => text,
-            _ => None,
+            Ok(bidirectional::SubResponse::SourceTextResult { text }) => text,
```
ChayimFriedman2, line 235: Looks like this has a potential for code reuse, something like:
```rust
            Ok(bidirectional::SubResponse::SourceTextResult { text }) => text, // Or other expected responses
            other => handle_failure(other),
```

crates/proc-macro-srv/src/lib.rs
```diff
@@ -139,7 +152,13 @@ impl ProcMacroSrv<'_> {
                 });
             match thread.unwrap().join() {
                 Ok(res) => res,
-                Err(e) => std::panic::resume_unwind(e),
+
+                Err(payload) => {
+                    if let Some(cancel) = payload.downcast_ref::<ProcMacroCancelMarker>() {
+                        return Err(PanicMessage { message: Some(cancel.reason.clone()) });
```
ChayimFriedman2: That doesn't seem right. This will lead us to show an error message for the user, which definitely isn't what we should be doing on cancellation. We probably need another error variant.

Maybe it's after all fine because if it's being cancelled the error won't be shown, but I still prefer another variant.

ChayimFriedman2 (changes requested):
Few comments.

crates/proc-macro-srv/src/lib.rs
Shourya742: Added another variant for this.

crates/proc-macro-api/src/bidirectional_protocol.rs
bit-aloo: This is tricky because the callback captures &dyn ExpandDatabase (later at implementation), which is not `UnwindSafe`. Requiring `UnwindSafe` on the callback would therefore force unwind-safety guarantees on `ExpandDatabase`.

Veykril: Yea the issue is that our database contains a `parking_lot::RwLock` (due the `DashMap` usage) which is not `RefUnwindSafe` as unlike std's verison it does not implement poisoning.

Let's put a comment note here though regarding that, we might be able to get rid of that issue in the future as this is all due to the VFS stuff only

crates/proc-macro-srv-cli/src/main_loop.rs
```diff
@@ -170,27 +171,49 @@ impl<'a, C: Codec> ProcMacroClientHandle<'a, C> {
...
+    }
+}
 
-        match bidirectional::BidirectionalMessage::read::<_, C>(&mut self.stdin.lock(), self.buf) {
-            Ok(Some(msg)) => Some(msg),
-            _ => None,
+fn handle_failure(failure: Result<bidirectional::SubResponse, ProcMacroClientError>) -> ! {
+    match failure {
+        Err(ProcMacroClientError::Cancelled { reason }) => {
+            panic_any(ProcMacroCancelMarker { reason });
```
Veykril:
```suggestion
            resume_unwind(ProcMacroCancelMarker { reason });
```
There is no need to incur the panic handler in this case

crates/proc-macro-srv-cli/src/main_loop.rs
```diff
@@ -170,27 +171,49 @@ impl<'a, C: Codec> ProcMacroClientHandle<'a, C> {
...
+        }
+        Err(err) => {
+            panic!("proc-macro IPC failed: {err:?}");
+        }
+        Ok(other) => {
+            panic!("unexpected SubResponse {other:?}");
         }
```
Veykril, line 213: Let's `panicy_any` these with a custom type as well like `InternalError(String)` or so, that way we can easily extend the final new protocol version such that we can differentiate actual proc-macro panics and internal failures like thes

crates/proc-macro-srv/src/lib.rs
```diff
@@ -117,10 +144,12 @@ impl ProcMacroSrv<'_> {
...
         callback: Option<ProcMacroClientHandle<'_>>,
-    ) -> Result<token_stream::TokenStream<S>, PanicMessage> {
+    ) -> Result<token_stream::TokenStream<S>, ExpandError> {
         let snapped_env = self.env;
-        let expander = self.expander(lib.as_ref()).map_err(|err| PanicMessage {
-            message: Some(format!("failed to load macro: {err}")),
+        let expander = self.expander(lib.as_ref()).map_err(|err| {
+            ExpandError::Panic(PanicMessage {
+                message: Some(format!("failed to load macro: {err}")),
+            })
```
Veykril: This then can also make use of that new internal error type since this is technically not a proc-macro panic

crates/proc-macro-srv/src/lib.rs
```diff
@@ -138,8 +167,14 @@ impl ProcMacroSrv<'_> {
...
-                Err(e) => std::panic::resume_unwind(e),
+                Ok(res) => res.map_err(ExpandError::Panic),
+
+                Err(payload) => {
+                    if let Some(cancel) = payload.downcast_ref::<ProcMacroCancelMarker>() {
+                        return Err(ExpandError::Cancelled { reason: Some(cancel.reason.clone()) });
+                    }
+                    std::panic::resume_unwind(payload)
+                }
             }
```
Veykril, line 193: and here we can then also downcast to that internal error variant as well

crates/proc-macro-srv-cli/src/main_loop.rs
bit-aloo: Here we are not continuing an existing panic. If something goes wrong on the client side, we catch the panic there and convert it into a cancellation message. Up to this point, there is no panic on the server side. We then intentionally originate a new panic here with a marker, which is later caught at the thread join boundary.

Veykril: We do want to `resume_unwind` this as otherwise on every cancellation in this flow we will print the backtrace. `resume_unwind` is really just panicking but without invoking the panic handler. This is really more us trying to bubble up an error in this case without being noisy about it. note that I am only talking about this branch, the other two remaining panics is fine.

bit-aloo: oh!!! I understand it better now, thanks for clarification.

Veykril (approved):

PR #21411, opened 2026-01-06
Fix not applicable on ambiguous ident pat for merge_match_arms

A4-Tacks:
Example
---
```rust
enum X {
    A,
    B,
    C,
}
use X::*;
fn main() {
    match A {
        $0A => todo!(),
        B => todo!(),
        C => todo!(),
    }
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
enum X {
    A,
    B,
    C,
}
use X::*;
fn main() {
    match A {
        A | B => todo!(),
        C => todo!(),
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/merge_match_arms.rs
+++ b/crates/ide-assists/src/handlers/merge_match_arms.rs
@@ -160,8 +160,11 @@ fn get_arm_types<'db>(
                     }
                 }
                 ast::Pat::IdentPat(ident_pat) => {
-                    if let Some(name) = ident_pat.name() {
+                    if let Some(name) = ident_pat.name()
+                        && ctx.sema.to_def(ident_pat).is_some()
+                    {
                         let pat_type = ctx.sema.type_of_binding_in_pat(ident_pat);
+
                         map.insert(name.text().to_string(), pat_type);
                     }
                 }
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/merge_match_arms.rs
```diff
@@ -160,9 +160,14 @@ fn get_arm_types<'db>(
                     }
                 }
                 ast::Pat::IdentPat(ident_pat) => {
+                    let has_type = ctx.sema.type_of_pat(local_pat).is_some();
```
ChayimFriedman2: Is there a pattern without a type? This is weird. Every pattern has a type, even binding patterns. Are you sure this code is treating the other way around, that is binding patterns, correctly?

A4-Tacks: > Are you sure this code is treating the other way around, that is binding patterns, correctly?

There are behavioral differences in patterns and bindings:

| pattern | `type_of_pat` | `type_of_binding_in_pat` |
| ---     | ---           | ---                      |
| None    | Some(Ty)      | None                     |
| x       | Some(Ty)      | Some(Ty)                 |

A4-Tacks: Distinguish local variables using the differences between these two methods

At least testing shows that it works

ChayimFriedman2: This table confirms that every pattern has a type. I also checked and this indeed regresses the following, i.e. suggests the assist where it shouldn't:
```rust
fn main() {
    match A {
        a => todo!(),
        B => todo!(),
        C => todo!(),
    }
}
```

ChayimFriedman2: You can use `to_def()` from `ast::IdentPat` to `hir::Local` to check if it refers to a local.

A4-Tacks: Good idea, okay

PR #21415, opened 2026-01-07
feat: Allow rust paths in symbol search

Veykril:
Big diff is mostly due to the first two commits adding some more docs, and the test outputs changing vastly from a new option wrapping

AI disclosure: Changes were made with help of opus 4.5

### DIFF
--- a/crates/hir/src/symbols.rs
+++ b/crates/hir/src/symbols.rs
@@ -21,12 +21,9 @@ use hir_ty::{
 };
 use intern::Symbol;
 use rustc_hash::FxHashMap;
-use syntax::{
-    AstNode, AstPtr, SyntaxNode, SyntaxNodePtr, ToSmolStr,
-    ast::{HasModuleItem, HasName},
-};
+use syntax::{AstNode, AstPtr, SyntaxNode, SyntaxNodePtr, ToSmolStr, ast::HasName};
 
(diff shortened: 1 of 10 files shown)


ChayimFriedman2:
In a perfect timing, [the last VSCode release removed their horrible filtering](https://code.visualstudio.com/updates/v1_108#_go-to-symbol-in-workspace-supports-special-characters-in-query), which hopefully should allow this to work as well in VSCode!

Veykril:
That only changes it to accept # and * I think though, they didn't actually address the root issue they just vibe coded the fix to my issue on the vscode issue tracker calling out those two symbols I think 😅

ChayimFriedman2:
I guess we can get `:` special-cased as well then?

Veykril:
Maybe

unlimitedsola:
@Veykril Hi, I'm interested in trying out this feature but I can't seem to get it to work. When I type `std::sync::RwLock`, it returns an empty result. Based on my observation, it seems that it always returns empty result if the query contains `::`. I've tried both Neovim and VSCode, none of them are working for me. Could you suggest what I might be missing? Thanks!

Veykril:
To make this work properly clients will need some glue code as usually the clients do some additional filtering themselves which causes all paths to be filtered out. The only editor this currently works in (that I know of) is Zed

PR #21416, opened 2026-01-07
Implement Span::ByteRange for proc-macro-srv

Shourya742:


crates/load-cargo/src/lib.rs
```diff
@@ -591,6 +591,16 @@ impl ProcMacroExpander for Expander {
 
                 Ok(SubResponse::FilePathResult { name })
             }
+            SubRequest::ByteRange { file_id, ast_id, start, end } => {
```
Veykril, line 597: Let's add a comment to this, `LineColumn` and `SourceText` about them not being incremental. We might want to consider allowing to enable/disable these for incrementality reasons in the future

bit-aloo: Added here: 65869af905f39e931aba627cdb74c58d47dbf39c

PR #21420, opened 2026-01-08
Fix not disable string escape highlights

A4-Tacks:
Example
---
with config `strings: false`

```rust
fn main() {
    format_args!("foo\nbar\invalid");
}
```

**Before this PR**

```rust
fn main() {
    format_args!("foo\nbar\invalid");
                  // ^^ EscapeSequence
                       // ^^ InvalidEscapeSequence
}
```

**After this PR**

```rust
fn main() {
    format_args!("foo\nbar\invalid");
}
```

### DIFF
--- a/crates/ide/src/syntax_highlighting.rs
+++ b/crates/ide/src/syntax_highlighting.rs
@@ -513,21 +513,21 @@ fn string_injections(
             );
 
             if !string.is_raw() {
-                highlight_escape_string(hl, &string);
+                highlight_escape_string(hl, config, &string);
             }
         }
     } else if let Some(byte_string) = ast::ByteString::cast(token.clone()) {
         if !byte_string.is_raw() {
-            highlight_escape_string(hl, &byte_string);
+            highlight_escape_string(hl, config, &byte_string);
         }
     } else if let Some(c_string) = ast::CString::cast(token.clone()) {
         if !c_string.is_raw() {
-            highlight_escape_string(hl, &c_string);
+            highlight_escape_string(hl, config, &c_string);
         }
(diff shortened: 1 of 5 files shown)


crates/ide/src/syntax_highlighting.rs
```diff
@@ -218,6 +218,7 @@ pub(crate) fn highlight(
     let mut hl = highlights::Highlights::new(root.text_range());
     let krate = sema.scope(&root).map(|it| it.krate());
     traverse(&mut hl, &sema, config, InRealFile::new(file_id, &root), krate, range_to_highlight);
+    hl.retain_mut(|hl_range| filter_by_config(&mut hl_range.highlight, config));
```
A4-Tacks: Is it better to filter the entire tree here, or is it better to pass `config` to each `add`?

crates/ide/src/syntax_highlighting/highlights.rs
```diff
@@ -26,6 +26,10 @@ impl Highlights {
         self.root.add(hl_range);
     }
 
+    pub(crate) fn retain_mut(&mut self, mut f: impl FnMut(&mut HlRange) -> bool) {
```
ChayimFriedman2: Can call this just `retain()`, the reason `Vec::retain_mut()` is named this way is due to a past mistake and it's IMO not a good name. Ditto for below.

crates/ide/src/syntax_highlighting.rs
ChayimFriedman2: I believe it's better to filter on the spot, not the entire tree. This is faster, and the tree can be large.

PR #21421, opened 2026-01-08
fix: Fixes for builtin derive expansions

ChayimFriedman2:
- Do not store the `MacroCallId` of the "real" expansion anywhere, so that the IDE layer could not expand it by mistake
 - Fix a stupid bug where we used the directive of the `derive` itself instead of of the macro, leading us to re-expand it again and again.

Fixes rust-lang/rust-analyzer#21382.

### DIFF
--- a/crates/hir/src/semantics.rs
+++ b/crates/hir/src/semantics.rs
@@ -13,7 +13,7 @@ use std::{
 use base_db::FxIndexSet;
 use either::Either;
 use hir_def::{
-    DefWithBodyId, MacroId, StructId, TraitId, VariantId,
+    BuiltinDeriveImplId, DefWithBodyId, HasModule, MacroId, StructId, TraitId, VariantId,
     attrs::parse_extra_crate_attrs,
     expr_store::{Body, ExprOrPatSource, HygieneId, path::Path},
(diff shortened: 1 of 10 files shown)


crates/hir/src/semantics.rs
```diff
@@ -690,8 +706,9 @@ impl<'db> SemanticsImpl<'db> {
             .derive_helpers_in_scope(InFile::new(sa.file_id, id))?
             .iter()
             .filter(|&(name, _, _)| *name == attr_name)
-            .map(|&(_, macro_, call)| (macro_.into(), call))
+            .filter_map(|&(_, macro_, call)| Some((macro_.into(), call.left()?)))
             .collect();
+        // FIXME: We filter our builtin derive "fake" expansions, is this correct? Should we still expose them somehow?
```
Veykril, line 711: That will mean 
```rust
#[derive(Default)]
enum Foo {
	#[default]
	Bar,
}
```
hover and stuff on `#[default]` will no longer work i think, thats a bit annoying

ChayimFriedman2: I checked and hover works. This also makes sense, as hover goes through resolve_path and not descend, which will ignore the call id (and only use the `MacroId`).

PR #21423, opened 2026-01-09
Project json compatibility improvements

cormacrelf:
Adds better backwards- and forwards-compatibility to the project JSON deserializer. For example, today, if you use `flycheck` runnables from rust-lang/rust-analyzer#18043 with an old rust-analyzer, it bails on deserializing the whole structure and you can't do anything. That's fixed (at least for new rust-analyzer). Otherwise adds and recommends more #[serde(default)].

Updated:
- added JSON deserialization testing for these newer pieces of the project model. These have been in fairly widespread use for about 18 months or so now. At least if someone is breaking these users, they'll know it when the tests fail.

### DIFF
--- a/crates/rust-analyzer/src/main_loop.rs
+++ b/crates/rust-analyzer/src/main_loop.rs
@@ -1180,6 +1180,8 @@ impl GlobalState {
             } => self.diagnostics.clear_check_older_than_for_package(id, package_id, generation),
             FlycheckMessage::Progress { id, progress } => {
                 let format_with_id = |user_facing_command: String| {
+                    // When we're running multiple flychecks, we have to include a disambiguator in
+                    // the title, or the editor complains. Note that this is a user-facing string.
                     if self.flycheck.len() == 1 {
                         user_facing_command
                     } else {


crates/project-model/src/project_json.rs
```diff
@@ -368,18 +368,26 @@ pub enum RunnableKind {
...
     /// May include {label} which will get the label from the `build` section of a crate.
     Flycheck,
+
+    /// For forwards-compatibility, i.e. old rust-analyzer binary with newer workspace discovery tools
+    Unknown,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
 pub struct ProjectJsonData {
+    #[serde(default)]
```
ChayimFriedman2: `Option` doesn't need `#[serde(default)]`.

crates/project-model/src/project_json.rs
```diff
@@ -414,6 +422,8 @@ struct CrateData {
     build: Option<BuildData>,
     #[serde(default)]
     proc_macro_cwd: Option<Utf8PathBuf>,
+    //
+    // New fields should be #[serde(default)].
```
ChayimFriedman2: I don't think we need this comment everywhere. One comment at the file's top should suffice.

cormacrelf:
I think it's warranted generally for the project JSON as opposed to config, not just because people commit their JSON to git, but also because the JSON is generated by other tools that have an independent release cycle. Especially if you're using workspace.discoverConfig. In contrast, RA config is usually done by humans and checked at the time of writing.

I recall, from the design discussions around the runnable kinds, Matklad describing some principles for JSON formats and pushing for a vec of runnables rather than a struct, so we would be more relaxed as to what's acceptable and more flexible to accept multiple of each kind, etc. Originally, it was a struct like `{ run: Option<Runnable>, test: Option<Runnable>, ... }`, which actually *is* forwards compatible for new runnable kinds since Serde ignores unknown fields. The alternative that we landed on is not, and I think it should be brought back to parity.

As a practical example, https://github.com/facebook/buck2/pull/767 will make the rust-project JSON generator incompatible with most users' copies of rust-analyzer unless we wait for RA's half of this functionality to make it to a stable rustup release so that we can point users somewhere. I wish I had made this compatibility PR a year ago so this would not be the case.

crates/project-model/src/project_json.rs
cormacrelf: I seem to have a brain worm from my early usage of serde in like 2015 and I've been doing this ever since. Seems you are right.

Also there are tests for old JSON files without new fields that should catch any new required fields and fail. I think I'll just add a test that parses a BuildData and a RunnableData so that we get coverage for these as well.

Veykril:
I think it's fine for us to say we don't necessarily support all version of this format here but keep a supported window (like we do with rust versions). Either way sounds like we should merge this quickly (and maybe beta back port as well?)

cormacrelf:
I'm not sure given the Meta people would be able to merge that PR quick enough for a backport to matter, but maybe it would be helpful to the rest of the community. You can do it if you have the time.

crates/project-model/src/project_json.rs
```diff
@@ -368,6 +368,9 @@ pub enum RunnableKind {
     /// Template for checking a target, emitting rustc JSON diagnostics.
     /// May include {label} which will get the label from the `build` section of a crate.
     Flycheck,
+
+    /// For forwards-compatibility, i.e. old rust-analyzer binary with newer workspace discovery tools
```
Wilfred, line 372: Nit: it's not that they're newer tools, it's that they're tool that can output additional types of runnable.

crates/project-model/src/project_json.rs
```diff
@@ -380,6 +383,8 @@ pub struct ProjectJsonData {
     crates: Vec<CrateData>,
     #[serde(default)]
     runnables: Vec<RunnableData>,
+    //
+    // New fields should be Option or #[serde(default)]. This applies to most of this datastructure.
```
Wilfred, line 387: "most of this datastructure" is ambiguous: which parts of the datastructure? How about "everything should be optional except crates, so we can add support for new fields that old rust-project.json tools aren't generating yet".

Wilfred:
Wearing my rust-project maintainer hat, we package r-a the same way as OSS does: including the binary in the VS Code extension. So if a user hasn't updated their extension, adding new types of runnables to rust-project.json is going to break their setup.

Given that, making r-a tolerant to unknown runnable types seems worthwhile.

davidbarsky:
> since `rust-project.json` is often committed and used by different r-a versions.

With my former-Meta engineer/current rust-analyzer team member hat on, I recommend _not_ committing and instead generating a `rust-project.json` on-the-fly through the `discoverProjectConfig`, but I agree with Wilfred: not breaking people is good, but I think relying can dramatically shorten the migration periods where rust-analyzer needs to be tolerant of old/new versions of `rust-project.json`.

> Wearing my rust-project maintainer hat, we package r-a the same way as OSS does: including the binary in the VS Code extension. So if a user hasn't updated their extension, adding new types of runnables to rust-project.json is going to break their setup.

The details are already fading from my memory: rust-project was packaged with dotslash, so the repo/dotslash version might move forward, but the extension might be not updated, right?

crates/project-model/src/project_json.rs
cormacrelf: That is implied by this comment being attached to a serde(other) variant called "Unknown"

PR #21425, opened 2026-01-09
Edition::EditionFuture for forwards compatibility

cormacrelf:
Like rust-lang/rust-analyzer#21423 but for editions.

I would like it so that RA is not necessarily bricked as soon as a new edition comes out. Usually a new edition means RA will have to update (new syntax, new keywords, new trait stuff, etc) but we should allow it to lurch on a little longer if that is not bothering a given user. I have been forced to upgrade because of this and it would have been nice at the time to do that the following month instead of in a mega-upgrade of rustc, 100 crates, rust-analyzer, and buck2 all at once.

I'm not sure if this is the exact right approach. Maybe we just assume all editions will be years and parse the string to a number? Then we can at least inject the prelude using a non-interned symbol to construct a `use ::std::prelude::rust_xxxx::*` import, rather than falling back to the latest edition and probably missing some symbols.

### DIFF
--- a/crates/edition/src/lib.rs
+++ b/crates/edition/src/lib.rs
@@ -10,6 +10,8 @@ pub enum Edition {
     Edition2018,
     Edition2021,
     Edition2024,
+    /// For forwards-compatibility, i.e. old rust-analyzer with newer rustc
+    EditionFuture,
 }
 
 impl Edition {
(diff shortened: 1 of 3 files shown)


ChayimFriedman2:
I definitely do not want us to introduce a new "fake" variant to the main `Edition` enum. I do support, though, deserializing unknown editions into the oldest or latest edition - whatever people prefer (and assuming the consensus on https://github.com/rust-lang/rust-analyzer/pull/21423 will be positive).

ChayimFriedman2:
@cormacrelf, we haven't heard from you so I'm closing this PR. Thank you for your interest in contributing!

PR #21426, opened 2026-01-09
Add config to disable keywords highlight

A4-Tacks:
The editor has already highlighted the keywords, and users may not necessarily need r-a highlighted keywords


ChayimFriedman2:
I... don't know how I feel about *yet another* config.

Sure, the editor will usually highlight them, but is it this bad to highlight them us too? Also FWIW the editor handling is incorrect at least in some editors, e.g. VSCode does not highlight contextual keywords like `union` or `raw`.

A4-Tacks:
> but is it this bad to highlight them us too?

The keyword highlighting of r-a is difficult to customize, so I usually disable the highlighting of r-a.
But disable r-a highlighting means that symbols and formatting cannot be highlighted either

BTW, I also plan to add configuration in other PR to disable symbol highlighting

> Also FWIW the editor handling is incorrect at least in some editors

This config is enabled by default, it's not a big problem

ChayimFriedman2:
> The keyword highlighting of r-a is difficult to customize

How so? It provides even more granularity than the usual editor, and of course the colors are completely your choice.

> This config is enabled by default, it's not a big problem

It **is** a big problem for maintainability and searchability of the configs. Not this specific one, of course, but the dozens of configs we're accumulating (we already have almost 200 options!) In the past the approach was to say "yes" to any user wanting a config (and finding someone to implement it, of course). Now I believe the default should be "no" unless the config will matter for many users, or will be **very important** to few. Of course that's only my personal opinion.

CC @rust-lang/rust-analyzer what are your thoughts on this, and on the subject in general?

A4-Tacks:
> It provides even more granularity than the usual editor

But disabling r-a highlighting, editor configuration can refine colors to specific keywords instead of all keywords or category

ChayimFriedman2:
r-a does provide some tags to keywords groups. If you want more granular control than that then I'll repeat my saying that I don't believe r-a should support that.

Veykril:
This is technically solved by the lsp augmentsSyntax or whatever that s called capability I think which will make r-a drop most regular highlighting that are unnecessary, unfortunately the text mate grammar VSCode ships for rust is really bad so we've disabled that capability in the extension code.

davidbarsky:
I think @Wilfred got commit bit on the underlying Textmate grammar recently and might be able to get improvements into it. I'm pretty sure that VS Code will pick any changes on their next submodule sync.

Wilfred:
Yep, happy to review any PRs on https://github.com/dustypomerleau/rust-syntax. I haven't noticed it being terribly bad, would be interested to hear what issues you've seen.

Veykril:
I don't recall specifics, just that it vastly differs from what our semantic tokens map to meaning unanalyzed things look completely different than analyzed ones

Wilfred:
My mental model of semantic highlighting is that it should be additional highlighting for things that the textmate grammar can't handle. For example, highlighting whether a variable is mutable:

cannot_be_changed_r_a

Textmate grammars are dumb, fast and available even when you have no LSP at all. For example, highlighting on the web via monaco benefits from the textmate grammar too.

Feel free to nerd snipe me, it's a [fun rabbit hole](https://www.wilfred.me.uk/blog/2014/09/27/the-definitive-guide-to-syntax-highlighting/) (shameless blog post link).

ChayimFriedman2:
What's the status of this @Veykril?

Veykril:
Still of the stance that this is unnecessary given what the LSP has here. I also do not use VSCode anymore so I don't know about the state of the textmate grammar there. Happy to undo that hack on the extension side if anyone wants to look into that.

PR #21433, opened 2026-01-09
internal: Include private definitions in generated rustdoc

Wilfred:
rust-analyzer has handy prebuilt `cargo doc` output at https://rust-lang.github.io/rust-analyzer/ide/

However, it doesn't include private definitions, which makes it less useful when trying to learn unfamiliar parts of the codebase.

Instead, pass `--document-private-items` so the HTML includes information on private types and modules too. rustdoc renders these with a padlock icon, so it's still clear that they're private.

This change also exposes some more rustdoc warnings, which I've fixed.

### DIFF
--- a/.github/workflows/rustdoc.yaml
+++ b/.github/workflows/rustdoc.yaml
@@ -24,7 +24,7 @@ jobs:
       run: rustup update --no-self-update stable
 
     - name: Build Documentation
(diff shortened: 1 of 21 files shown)


ChayimFriedman2 (approved):
Thanks!

PR #21434, opened 2026-01-09
fix: Remove code made redundant by method resolution rewrite

ChayimFriedman2:
Its job is now done elsewhere, and it's also wrong (not accounting for autoderef)

Fixes rust-lang/rust-analyzer#21429.

### DIFF
--- a/crates/hir-ty/src/tests/regression.rs
+++ b/crates/hir-ty/src/tests/regression.rs
@@ -891,13 +891,14 @@ use core::ops::Deref;
 
 struct BufWriter {}
 
-struct Mutex<T> {}
-struct MutexGuard<'a, T> {}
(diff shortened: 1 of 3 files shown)


crates/hir-ty/src/tests/regression.rs
```diff
@@ -891,28 +891,33 @@ use core::ops::Deref;
 
 struct BufWriter {}
 
-struct Mutex<T> {}
```
ChayimFriedman2, line 894: The changed tests are due to variance. The new solver is very sensitive to variance. Making type params bivariant, commonly done in tests by not including a field with them, makes the solver unable to reason about equivalences between them, often ending with unknown types.

PR #21439, opened 2026-01-10
internal: Land integration test infra for proc-macro-srv-cli

Veykril:
Very bare bones and only tests parts of the legacy-json protocol. Will be filling this out as a follow up

Follow up to, making use of the changes there https://github.com/rust-lang/rust-analyzer/pull/21438


Veykril:
Didn't manage to flesh this out more over the weekend so I'll land this as is for now.

PR #21450, opened 2026-01-12
minor: Fix overlapping cfg attributes for wasm32-unknown-emscripten target

robertoaloi:
The `wasm32-unknown-emscripten` target sets both `cfg(unix)` and `cfg(target_arch = "wasm32")`, causing the `mod imp` definitions in `process.rs` to overlap and fail to compile:

```
$ rustup target add wasm32-unknown-emscripten
$ cargo build -p stdx --target wasm32-unknown-emscripten
error[E0428]: the name `imp` is defined multiple times
   --> crates/stdx/src/process.rs:260:1
    |
 80 | mod imp {
    | ------- previous definition of the module `imp` here
...
260 | mod imp {
    | ^^^^^^^ `imp` redefined here
    |
    = note: `imp` must be defined only once in the type namespace of this module

For more information about this error, try `rustc --explain E0428`.
```

This PR ensures the three platform-specific implementations are mutually exclusive.
Before this fix, the above command fails with duplicate `mod imp` definitions. After this fix, it compiles successfully.

### DIFF
--- a/crates/stdx/src/process.rs
+++ b/crates/stdx/src/process.rs
@@ -76,7 +76,7 @@ pub fn spawn_with_streaming_output(
     Ok(Output { status, stdout, stderr })
 }
 
-#[cfg(unix)]
+#[cfg(all(unix, not(target_arch = "wasm32")))]
 mod imp {
     use std::{
         io::{self, prelude::*},


lnicola:
Thanks, are you actually building it for WASM?

robertoaloi:
> Thanks, are you actually building it for WASM?

No, but we use the unpublished `stdx` crate as a dependency for [ELP](https://github.com/whatsapp/erlang-language-platform) and this was breaking one of our internal checks.

ChayimFriedman2:
I really really do not recommend depending on r-a's `stdx`. Make your own copy. If there is one unsupported crate in rust-analyzer, it's `stdx`.

In fact, would I have known this before the merge I'd probably say do not merge this. Now that it's merged though I'd leave it as-is.

robertoaloi:
@ChayimFriedman2 Thanks for the insight! I fully understand the reasoning. At the same time, ELP is basically a hard-fork of Rust Analyzer (but for Erlang), so it still makes sense to somehow follow upstream. In case of issues, we can always fork/duplicate the crate.

PR #21456, opened 2026-01-13
fix: Make `naked_asm!()` always return `!`

ChayimFriedman2:
As it should.

Fixes rust-lang/rust-analyzer#21454.

### DIFF
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -8,7 +8,7 @@ use hir_def::{
     expr_store::path::{GenericArgs as HirGenericArgs, Path},
     hir::{
         Array, AsmOperand, AsmOptions, BinaryOp, BindingAnnotation, Expr, ExprId, ExprOrPatId,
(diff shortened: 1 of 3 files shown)


luodeb:
Sorry, naked_asm doesn't always seem to return `Never`.  
Such as this code
```rust
#[unsafe(naked)]
fn naked_func() -> i32 {
    unsafe {
        naked_asm!(
             "
             mov x0, #42
             ret
            "
       );
    }
}
```

I don't know if this change will have any impact.

ShoyuVanilla:
> Sorry, naked_asm doesn't always seem to return `Never`. Such as this code
> 
> ```rust
> #[unsafe(naked)]
> fn naked_func() -> i32 {
>     unsafe {
>         naked_asm!(
>              "
>              mov x0, #42
>              ret
>             "
>        );
>     }
> }
> ```
> 
> I don't know if this change will have any impact.

Never type is coercible to any type so it won't make type mismatch error for such cases

ChayimFriedman2:
It does always return never, but as @ShoyuVanilla said, never is coercible to any type.

Issue #11409, opened 2022-02-04
Match fill arms postfix


JosephTLyons:
If you have

```rust
let a = Some(1);
```

when you type `a.match` and complete it, you get

```rust
match a {
    Some(_) => {},
    None => {},
}
```

If you have a custom `enum`

```rust
enum State {
    Blocked,
    Running,
    Stopped,
}
```

```
let state = State::Blocked;
```

typing `state.match` and completing it gives

```rust
match state {
    _ => {},
}
```
As a note, RA has an assist to automatically fill arms:

![SCR-20220204-39u](https://user-images.githubusercontent.com/19867440/152488269-1e679cec-af67-403d-8c69-5a50d9991e26.png)

Veykril:
The reason why we aren't autofilling arms on completion in the general is that a user may not want to always match all variants of an enum(option is special because there are 2 variants and you have to match with 2 arms for the match to be useful)

Nadrieril:
Could there be a threshold, maybe fill all of them if there are only 2-3 cases since it's easy to remove 1-2 cases?

Veykril:
Yes, I think generalizing this for matches of 1 and 2 arms makes sense, 3 I would say no to as in that case you might want to go back to 2 arms, the interesting match and a `_` fallback. Whereas with 2 arms you'd merely replace the uninteresting one with `_`.

jonathancyu:
Is it viable to create a .matchall postfix snippet? I understand the reasoning, however I find myself expanding .match then using the fill match arms code action that I think having a snippet that fills all arms is warranted.

Nadrieril:
This could alternatively be helped by a ["split this pattern into subpatterns"](https://github.com/rust-lang/rust-analyzer/issues/8365) code action: step one generates a match with `_`, step two expands it into all the variants.

PR #21457, opened 2026-01-13
feat: Add destructure_enum assist

Wilfred:
r-a already has the ability to fill in match arms as part of the add_missing_match_arms assist. However, there's currently no way to go from `flavor` to `match flavor { Apple => {} ... }`.

If a user does want the IDE to fill in match arms, they have to write the `match` and `{}` themselves. Other language LSPs supporting pattern matching (e.g. merlin for OCaml) support writing the whole match statement.

Whilst I'm aware there's a freeze on new assists, this assist is just a generalisation of add_missing_match_arms and reuses the existing logic there (which has great test coverage). I think it's a really compelling feature and hopefully a small ask.

AI disclosure: Written with a little bit of help from Claude in a few places.

### DIFF
--- a/crates/ide-assists/src/handlers/add_missing_match_arms.rs
+++ b/crates/ide-assists/src/handlers/add_missing_match_arms.rs
@@ -388,20 +388,20 @@ fn does_pat_match_variant(pat: &Pat, var: &Pat) -> bool {
 }
 
 #[derive(Eq, PartialEq, Clone)]
-enum ExtendedEnum {
+pub(crate) enum ExtendedEnum {
     Bool,
     Enum { enum_: hir::Enum, use_self: bool },
 }
 
 #[derive(Eq, PartialEq, Clone, Copy, Debug)]
-enum ExtendedVariant {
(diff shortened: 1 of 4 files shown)


A4-Tacks:
Currently, for `Option` and `Result`, postfix completion overlaps with this assistant

Perhaps we should strengthen postfix completion instead of adding assistants?

Related #11409

crates/ide-completion/src/completions/postfix.rs:165

Wilfred:
Huh, I had totally forgotten about `.match`. Thanks for the issues too: in addition to #11409, I think the linked #8365 is very relevant. This PR is essentially adding case splitting (maybe that's a better name for this assist?).

The downsides of `.match` is that it's much harder to discover (not valid Rust syntax and I don't think there's precedent in other languages?) and it's awkward to use when you have larger expressions. For example, if I have `foo.bar.baz` I can highlight the whole expression,  highlight `foo.bar` or just highlight `foo` and destructure depending on the type of those subexpressions.

ChayimFriedman2:

PR #21466, opened 2026-01-14
Remove postcard from legacy

Shourya742:
This PR removes the postcard codec from the legacy protocol, as postcard is not self-describing and breaks legacy message encoding. In addition, the bidirectional message types have been updated to remove flatten, making them compatible with postcard in the bidirectional path.


bit-aloo:
@Veykril I didn’t remove the codec abstraction for now, but it does look redundant given that each protocol now has its own codec scheme. There’s probably not much benefit in keeping it, so I can remove it in a follow-up if we agree. That would require cleaning up a couple of related bits as well.

Veykril:
yea I think that makes sense to remove as well then

PR #21475, opened 2026-01-15
fix: lookup flycheck by ID instead of vector index

lukasoyen:
After rust-lang/rust-analyzer#18043 introduced per-package flycheck for JSON projects, the code assumed that `world.flycheck` indices matched `world.workspaces` indices. However, not all workspaces have flycheck enabled (e.g., JSON projects without a flycheck template configured), so the flycheck vector can be shorter than the workspaces vector.

This caused an index-out-of-bounds panic when saving a file in a JSON project without flycheck configured:

  thread 'Worker' panicked at notification.rs:
  index out of bounds: the len is 0 but the index is 0

Fix by looking up the flycheck handle by its ID (which is the workspace index set during spawn) rather than using the workspace index directly as a vector index.

### DIFF
--- a/crates/rust-analyzer/src/handlers/notification.rs
+++ b/crates/rust-analyzer/src/handlers/notification.rs
@@ -387,14 +387,21 @@ fn run_flycheck(state: &mut GlobalState, vfs_path: VfsPath) -> bool {
                                         } => false,
                                     });
                                 if let Some(idx) = package_workspace_idx {
-                                    let workspace_deps =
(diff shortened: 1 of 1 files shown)


lukasoyen:
cc @cormacrelf

cormacrelf:
18043 only added line 396 so this bug pre-dates it. In any case I would also check the use of indices in giving display strings to flychecks but I think that should be fine as it fills in gaps with empty strings.

(Edit: By that I mean, I think you can have cargo projects without flychecks too. So those would have caused errors already.)

PR #21487, opened 2026-01-18
feat: Support extract variable in macro call

A4-Tacks:
Implement based first and last token mappings

Example
---
```rust
macro_rules! foo {
    (= $($t:tt)*) => { $($t)* };
}

fn main() {
    let x = foo!(= $02 + 3$0 + 4);
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
macro_rules! foo {
    (= $($t:tt)*) => { $($t)* };
}

fn main() {
    let $0var_name = 2+3;
    let x = foo!(= var_name + 4);
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -2830,6 +2830,7 @@ fn main() {
 
     #[test]
     fn extract_variable_in_token_tree() {
+        // FIXME: Keep the original trivia instead of extracting macro expanded?
         check_assist_by_label(
             extract_variable,
             r#"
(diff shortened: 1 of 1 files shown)


A4-Tacks:
At present, it only supports extraction into outside of macros, and there may be improvements in the future

ChayimFriedman2:
Then test that it's not applicable in a macro?

A4-Tacks:
At least for small macro calls, extracting to outside the macro is very practical

Supporting extraction to inside the macros may disrupt mappings
If this is check and not applicable, it is easy to false negative

ChayimFriedman2:
I'm just saying that whatever the current behavior is, it's good to test it, so we won't have surprises now or in the future.

A4-Tacks:
> it's good to test it

Oh, I have already pushed the test

PR #21491, opened 2026-01-19
fix: Insert type vars and normalize for the type of a used `static`

ChayimFriedman2:
They have their own special path, so they slipped through.

Fixes rust-lang/rust-analyzer#21478.


ChayimFriedman2:
I'm a bit worried from the fact that without type var insertion this fails, as it seems to me that it should typeck well even with an error const (in actuality some coercion takes place and transforms the entire type into an error type). But this also replicates in rustc (run the following with `-Zunpretty=hir,typed`):
```rust
use std::sync::LazyLock;

const L: usize = loop {};
static LAZY: LazyLock<[u32; L]> = LazyLock::new(|| []);

fn foo() {
    let t = LazyLock::force(&LAZY);
}
```
So it's probably a t-types matter.

PR #21495, opened 2026-01-19
Fix not complete 'else' before tuple

A4-Tacks:
Example
---
```rust
fn foo() -> (i32, i32) {
    if foo {} el$0
    (2, 3)
}
```

**Before this PR**

```rust
...
kw crate::
kw false
kw for
...
```

**After this PR**

```rust
...
kw crate::
kw else
kw else if
kw false
kw for
...
```

### DIFF
--- a/crates/ide-completion/src/context/analysis.rs
+++ b/crates/ide-completion/src/context/analysis.rs
@@ -2030,9 +2030,10 @@ fn is_after_if_expr(node: SyntaxNode) -> bool {
         Some(stmt) => stmt.syntax().clone(),
         None => node,
     };
-    let prev_sibling =
-        non_trivia_sibling(node.into(), Direction::Prev).and_then(NodeOrToken::into_node);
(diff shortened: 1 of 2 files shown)


crates/ide-completion/src/context/analysis.rs
```diff
@@ -2030,9 +2030,9 @@ fn is_after_if_expr(node: SyntaxNode) -> bool {
         Some(stmt) => stmt.syntax().clone(),
         None => node,
     };
-    let prev_sibling =
-        non_trivia_sibling(node.into(), Direction::Prev).and_then(NodeOrToken::into_node);
-    iter::successors(prev_sibling, |it| it.last_child_or_token()?.into_node())
+    let Some(prev_token) = previous_non_trivia_token(node) else { return false };
+    iter::successors(prev_token.parent(), |it| it.parent())
```
```suggestion
    prev_token.parent_ancestors()
```

PR #21504, opened 2026-01-21
fix: Add recursion limit when hovering on const values

Wilfred:
This fixes rust-lang/rust-analyzer#21503

AI disclosure: I minimised the crash manually and wrote the unit test, then asked Claude to write a fix. I've reviewed all the changes and they seem reasonable.

I'm in two minds about this solution: the sample code shouldn't produce an infinite recursion as far as I can see, so maybe there's a better fix. However, I also suspect it'd be possible to write a genuine infinite recursion const function, in which case you wouldn't want r-a to stack overflow even though the code wouldn't compile.

Let me know what you think.

### DIFF
--- a/crates/hir-ty/src/display.rs
+++ b/crates/hir-ty/src/display.rs
@@ -702,6 +702,7 @@ impl<'db> HirDisplay<'db> for Const<'db> {
                 &const_bytes.value.inner().memory,
                 &const_bytes.value.inner().memory_map,
                 const_bytes.ty,
+                CONST_SCALAR_RENDER_DEPTH_LIMIT,
             ),
             ConstKind::Unevaluated(unev) => {
                 let c = unev.def.0;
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
I don't think it can be possible to create an infinite recursion. When evaluating yes, but not when rendering.

Wilfred:
@ChayimFriedman2 It shouldn't be possible, but it definitely is today unfortunately -- see the test added.

Perhaps the type parameters aren't being substituted correctly? It looks like rendering is infinitely recursing on `Option<()>::SIGNATURE` instead of recursing into `()::SIGNATURE` as it should.

Veykril:
Didn't we have a similar issue with `impl Trait` where `trait Trait<T = Self>` overflowing. How did we solve that? Feels like we could take a similar approach here as well.

ShoyuVanilla:
https://github.com/rust-lang/rust-analyzer/pull/16877
I'm not sure is this what you are mentioning but you've introduced `generic_args_sans_defaults` 😄

Veykril:
Ah, right, we didn't solve it I worked around it 😆

Wilfred:
I see there isn't any substitution in the environment here, that feels like it might be the problem?

https://github.com/rust-lang/rust-analyzer/blob/2e3535826072546f88b746633db65f453f783028/crates/hir-ty/src/display.rs#L940-L944

ChayimFriedman2:
From what I can tell we should not call `trait_environment()` at all there but rather pass the `ParamEnv` from the caller. In general a `ParamEnv` represents the assumptions in the generic body you're in, and we're instantiating the fields.

(Also FWIW you almost never want to instantiate a `ParamEnv`, the only exception AFAIK is canonicalization).

ChayimFriedman2:
As I said I don't believe this is the correct route. You can check the `ParamEnv` thing, it could be related.

PR #21508, opened 2026-01-22
Support else-branch for move_guard

A4-Tacks:
Example
---
```rust
fn main() {
    match 92 {
        x @ 0..30 $0if x % 3 == 0 => false,
        x @ 0..30 if x % 2 == 0 => true,
        _ => false
    }
}
```

**Before this PR**

```rust
fn main() {
    match 92 {
        x @ 0..30 => if x % 3 == 0 {
            false
        },
        x @ 0..30 if x % 2 == 0 => true,
        _ => false
    }
}
```

**After this PR**

```rust
fn main() {
    match 92 {
        x @ 0..30 => if x % 3 == 0 {
            false
        } else if x % 2 == 0 {
            true
        },
        _ => false
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/move_guard.rs
+++ b/crates/ide-assists/src/handlers/move_guard.rs
@@ -440,6 +440,32 @@ fn main() {
             r#"
 fn main() {
     match 92 {
+        x @ 0..30 if x % 3 == 0 => false,
+        x @ 0..30 $0if x % 2 == 0$0 => true,
+        x @ 0..30 => false,
+        _ => true
+    }
+}
+"#,
+            r#"
+fn main() {
+    match 92 {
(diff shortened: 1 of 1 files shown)


A4-Tacks:
> This will transform the code wrongly if the user is selecting the non-first arm.

From testing, it is expected

ChayimFriedman2:
I meant where there is an empty selection, so all matching arms are eligible, but we'll reorder them to put the arm where the cursor is first.

A4-Tacks:
> I meant where there is an empty selection, so all matching arms are eligible

Why are we doing this? The current logic is consistent with `Merge match arms`

ChayimFriedman2:
Okay, I see now. I think this isn't the best behavior but since it's consistent, LGTM.

Thanks!

PR #21513, opened 2026-01-24
Improve extract_function name

A4-Tacks:
If the name contains `_`, it is likely to be descriptive

Example
---
```rust
fn foo(kind: i32) {
    let is_complex = $0kind != 0$0;
}
```

**Before this PR**

```rust
fn foo(kind: i32) {
    let is_complex = fun_name(kind);
}

fn fun_name(kind: i32) -> bool {
    kind != 0
}
```

**After this PR**

```rust
fn foo(kind: i32) {
    let is_complex = is_complex(kind);
}

fn is_complex(kind: i32) -> bool {
    kind != 0
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/extract_function.rs
+++ b/crates/ide-assists/src/handlers/extract_function.rs
@@ -257,7 +257,7 @@ fn make_function_name(
 
     let mut name = body
         .suggest_name()
-        .filter(|name| name.contains('_'))
+        .filter(|name| name.len() > 2)
         .unwrap_or_else(|| default_name.to_owned());
     let mut counter = 0;
     while names_in_scope.contains(&name) {


A4-Tacks:
> Why presume that names without `_` are non-descriptive?

Without a `_`, it may contain some short variables, like `i`, `item`, `ele`, `guard`, etc

Not only does it need to be descriptive, but it also needs to be applicable to function names

ChayimFriedman2:
This doesn't seem to me like a good reason to exclude *all* one-word names. We can have an exclusion list, but there are also many useful one-word names.

A4-Tacks:
> We can have an exclusion list

This may be difficult to cover most meaningless short names?

> but there are also many useful one-word names.

Do you have any examples?

ChayimFriedman2:
> This may be difficult to cover most meaningless short names?

It's not the end of the world if we miss some name, the only impact will be that the function name will be different.

> Do you have any examples?

Just from a single file in r-a: `snapshot`, `unsize`, `obligations`, `adjustments`, `mutbl`, `traits`, `reborrow`, and I stopped here.

A4-Tacks:
It does look that way

PR #21516, opened 2026-01-24
Publish no-server to Code Marketplace and OpenVSX

Master-Hash:
fix https://github.com/rust-lang/rust-analyzer/issues/18578

I believe it won't break anything.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -264,8 +264,6 @@ jobs:
           name: ${{ env.TAG }}
           token: ${{ secrets.GITHUB_TOKEN }}
 
-      - run: rm dist/rust-analyzer-no-server.vsix
-
       - run: npm ci
         working-directory: ./editors/code
 


ChayimFriedman2:
@rust-lang/rust-analyzer do we want this?

lnicola:
I feel it might be quite confusing, and it's a hack for our lack of support of some architectures that we might (tm) add in the future. But once we publish it to the marketplace, it will stay there.

Are there other popular extensions that publish multiple variants?

ChayimFriedman2:
Well, you're kinda our expert on these subjects @lnicola (and I also think it's hacky). So, close this?

Master-Hash:
I want to say, a vsix without binary is beneficial to whose rust-analyzer is managed by apt, dnf, rustup, etc

Official vscode registry is rather slow

Master-Hash:
> Are there other popular extensions that publish multiple variants?

https://marketplace.visualstudio.com/items?itemName=myriad-dreamin.tinymist

lnicola:
Maybe we should go back to downloading the server binary at run-time :upside_down_face:.

ChayimFriedman2:
> Maybe we should go back to downloading the server binary at run-time 🙃.

Oh no, it was a relief to remove that code.

lnicola:
> https://marketplace.visualstudio.com/items?itemName=myriad-dreamin.tinymist

I'm not sure what I'm looking at. They have Tinymist Typst and Tinymist Typst HTML on the Marketplace, but the latter seems to be a different extension. I couldn't find it in the source repo, but their release assets have:

image

So I guess they have a browser-only version, but it's not published to the Marketplace?

Anyway, I see they've added the `web` platform:

> The currently available platforms are: win32-x64, win32-arm64, linux-x64, linux-arm64, linux-armhf, alpine-x64, alpine-arm64, darwin-x64, darwin-arm64 and web.

> If you want a platform-specific extension to also support running in the browser as a [web extension](https://code.visualstudio.com/api/extension-guides/web-extensions), it must target the web platform when publishing. The web platform respects the browser entry point in the package.json. To disable the extension capabilities that are not supported in the web, we recommend using when clauses in the package.json instead of shipping separate package.json for the web platform or removing parts of the VSIX that do not work in the web.

We could misappropriate `web` to mean "runs on desktop, but you'll need to provide the server binary", but it still feels a bit wrong.

Anyway, I don't think we want to publish a second version to the Marketplace.

> Oh no, it was a relief to remove that code.

But HTTP clients are [really nice](https://www.stepsecurity.io/blog/axios-compromised-on-npm-malicious-versions-drop-remote-access-trojan) these days!

lnicola:
Anyway, I don't know if this PR works. Do target-specific extensions override universal ones? I think we've ran into a Marketplace bug related to this when they were originally introduced.

Master-Hash:
This doesn't mean a second extension.

图片

It means "All Platforms" here.

And "universal" in your quoted screenshot.

Master-Hash:
> Do target-specific extensions override universal ones?

No, they coexist.

> I think we've ran into a Marketplace bug related to this when they were originally introduced.

Yes, I inspected the discussions. I believe they are fixed now.

lnicola:
> No, they coexist.

But you still get the one for your target if it's available, right?

Maybe we should also upgrade the vsce / ovsx helpers before, but I'm a bit afraid these days..

Master-Hash:
> > No, they coexist.
> 
> But you still get the one for your target if it's available, right?
> 
> Maybe we should also upgrade the vsce / ovsx helpers before, but I'm a bit afraid these days..

Users can choose the universal target somewhere manually (sorry I forgot), after that, vscode will download the correct update target.

lnicola:
> Users can choose the universal target somewhere manually (sorry I forgot), after that, vscode will download the correct update target.

I mean, if I just install Tinymist, will I get the x64 version or the universal one? I know I could check, but I don't have a sandboxed VS Code around and

> I'm a bit afraid these days

Master-Hash:
> Maybe we should also upgrade the vsce / ovsx helpers before, but I'm a bit afraid these days..

Afraid of Supply Chain Attack? https://news.ycombinator.com/item?id=47513932 may help you

Master-Hash:
> I mean, if I just install Tinymist, will I get the x64 version or the universal one? I know I could check, but I don't have a sandboxed VS Code around and

the x64 version

Master-Hash:
> Maybe we should also upgrade the vsce / ovsx helpers before

I think so.

I only know "not overwriting" happens in my environment & tinymist's, but I'm not confident it works here

lnicola:
I fired off a new nightly release in https://github.com/rust-lang/rust-analyzer/actions/runs/23790412943.

lnicola:
@Master-Hash I wonder if it matters whether we publish the universal or platform specific VSIXes first...

Master-Hash:
> @Master-Hash I wonder if it matters whether we publish the universal or platform specific VSIXes first...

I'm not sure (as I fail to reproduce the following issues)

I suggest testing in nightly first

lnicola:
I'm not suggesting we change the order, I'm just worried it might make a difference. Shell glob expansion sorts them lexicographically, but `ovsx` and `vsce` seem to behave differently, even though `ovsx` is basically a `vsce` wrapper:

image

PR #21518, opened 2026-01-25
internal: Use parser expect where possible

pksunkara:
I had this change lying around from more than a year ago when I sent rust-lang/rust-analyzer#16786. Was cleaning up my local and found it.

We should use `p.expect` instead of manually doing `p.error` after `p.eat`.

### DIFF
--- a/crates/parser/test_data/parser/err/0002_duplicate_shebang.rast
+++ b/crates/parser/test_data/parser/err/0002_duplicate_shebang.rast
@@ -28,7 +28,7 @@ SOURCE_FILE
         NAME_REF
           IDENT "rusti"
   WHITESPACE "\n"
-error 23: expected `[`
+error 23: expected L_BRACK
 error 23: expected an item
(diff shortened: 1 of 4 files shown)


crates/parser/test_data/parser/err/0002_duplicate_shebang.rast
```diff
@@ -28,7 +28,7 @@ SOURCE_FILE
         NAME_REF
           IDENT "rusti"
   WHITESPACE "\n"
-error 23: expected `[`
+error 23: expected L_BRACK
```
lnicola, line 31: Should we pretty-print these, I wonder?

pksunkara: If this is visible to the user, yes.

ChayimFriedman2: It is.

PR #21522, opened 2026-01-25
fix: Add semicolon for toggle_macro_delimiter

A4-Tacks:
Example
---
```rust
macro_rules! sth {
    () => {};
}

sth!$0{ }
```

(old test `sth!{};` is a syntax error in item place)

**Before this PR**

```rust
macro_rules! sth {
    () => {};
}

sth![ ]
```

**After this PR**

```rust
macro_rules! sth {
    () => {};
}

sth![ ];
```

### DIFF
--- a/crates/ide-assists/src/handlers/toggle_macro_delimiter.rs
+++ b/crates/ide-assists/src/handlers/toggle_macro_delimiter.rs
@@ -86,7 +86,14 @@ pub(crate) fn toggle_macro_delimiter(acc: &mut Assists, ctx: &AssistContext<'_>)
                 }
                 MacroDelims::LCur | MacroDelims::RCur => {
                     editor.replace(ltoken, make.token(T!['[']));
-                    editor.replace(rtoken, make.token(T![']']));
+                    if semicolon.is_some() || !needs_semicolon(token_tree) {
+                        editor.replace(rtoken, make.token(T![']']));
+                    } else {
+                        editor.replace_with_many(
+                            rtoken,
+                            vec![make.token(T![']']).into(), make.token(T![;]).into()],
+                        );
+                    }
                 }
             }
             editor.add_mappings(make.finish_with_mappings());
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/toggle_macro_delimiter.rs
```diff
@@ -95,6 +102,15 @@ pub(crate) fn toggle_macro_delimiter(acc: &mut Assists, ctx: &AssistContext<'_>)
     )
 }
 
+fn needs_macro_value(makro: ast::MacroCall) -> bool {
+    makro
+        .syntax()
+        .parent()
+        .and_then(ast::MacroExpr::cast)
+        .and_then(|it| ast::StmtList::cast(it.syntax().parent()?))
+        .is_some()
```
ChayimFriedman2: I do not understand the logic here. If this is a MacroExpr and the child of a StmtList, it *does* need semicolon, no?

A4-Tacks: For `{m!{} n!{}}`, it ast is `StmtList { ExprStmt { MacroExpr }, MacroExpr }`

- `m!` ExprStmt does not require a value, it requires a semicolon
- `n!` MacroExpr requires a value, meaning no semicolons are needed

ChayimFriedman2: But what about things like `foo!(m! {})`? Even if this works, please add a test for this.

A4-Tacks: It has been tested in `test_nested_macros` that currently does not support nested toggle delimiters

ChayimFriedman2: These are not nested macro calls?

ChayimFriedman2: Sorry, I meant `foo(m! {})` (`foo` is not a macro).

A4-Tacks:
> There is also the problem of removing the semicolon in item lists when changing other brackets to `{}`

Is there a problem with this? If the delimiter is `{}`, there is no need for a semicolon, and adding a semicolon in item list is a syntax error

ChayimFriedman2:
If the delimiter was `()` or `[]` it had a semicolon, then you changed it to `{}` and suddenly the semicolon is a syntax error.

A4-Tacks:
~~But this has already been fixed by #20534~~

But hasn't this already been implemented? `m!();` -> `m!{}`

ChayimFriedman2:
Oops I thought it wasn't :sweat_smile:

PR #21526, opened 2026-01-26
fix: Fix not complete `.not` in condition

A4-Tacks:
Example
---
```rust
fn foo(cond: bool) {
    if cond.$0
}
```

**Before this PR**

```text
...
sn deref         *expr
sn ref           &expr
...
```

**After this PR**

```text
...
sn deref         *expr
sn not           !expr
sn ref           &expr
...
```


A4-Tacks:
Okay, rebased and fixed test

PR #21536, opened 2026-01-27
fix: offer `toggle_macro_delimiter` in nested macro

A4-Tacks:
Example
---
```rust
prt!{abc!$0(3 + 5)};
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
prt!{abc!{3 + 5}};
```

### DIFF
--- a/crates/ide-assists/src/handlers/toggle_macro_delimiter.rs
+++ b/crates/ide-assists/src/handlers/toggle_macro_delimiter.rs
@@ -415,7 +415,6 @@ prt!{(3 + 5)}
         )
     }
 
-    // FIXME @alibektas : Inner macro_call is not seen as such. So this doesn't work.
     #[test]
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/toggle_macro_delimiter.rs
```diff
@@ -405,7 +418,7 @@ prt!{(3 + 5)}
     // FIXME @alibektas : Inner macro_call is not seen as such. So this doesn't work.
```
ChayimFriedman2: Remove this?

ChayimFriedman2 (approved):
A bit hacky solution over using semantic info but works, and has the advantage of working inside macro definitions as well.

ChayimFriedman2 (approved):

PR #21537, opened 2026-01-27
minor: Sync from downstream

lnicola:


lnicola:
Um, where are we supposed to get `rustc_proc_macro` from?

bjorn3:
rustc_proc_macro is part of the rustc-dev component, but using `proc_macro` in this repo is fine. Both crates are effectively identical in stage 2 rustc builds such as we ship with rustup. Their difference only matters during bootstrapping.

PR #21538, opened 2026-01-27
Fix linking of postcard test

Shourya742:


lnicola:
Is this the same as https://github.com/rust-lang/rust/pull/151751/changes/d400dca751671b531c0826e30d9f7eba24cff8f2? We'll get it during the next sync, probably not worth having two identical commits.

bit-aloo:
> Is this the same as [rust-lang/rust@d400dca](https://github.com/rust-lang/rust/commit/d400dca751671b531c0826e30d9f7eba24cff8f2)? We'll get it during the next sync, probably not worth having two identical commits.

Just saw it, I was not aware. Thanks for mentioning. Ya they are the same, I will close this PR

PR #21545, opened 2026-01-29
fix: auto-import macro when exclamation mark is already typed

ddmoney420:
## Summary
Fixes auto-import for macros when the user has already typed the `!` character.

Fixes rust-lang/rust-analyzer#20898

## Problem
When using auto-completion with auto-import for macros:
- Typing `warn` and completing → correctly adds both `warn!()` and the import
- Typing `warn!` and completing → inserts `warn!()` but **fails to add the import**

## Root Cause
In `flyimport.rs`, when determining what name to search for:
- For `warn`, `ctx.token` is an identifier, so `import_name()` returns `"warn"`
- For `warn!`, `ctx.token` is `!` (not an identifier), so `import_name()` returns `""`

## Solution
Added `import_name_from_macro_call()` helper that extracts the macro name when the current token is `!` inside a `MACRO_CALL` node.

## Test plan
- [x] Added test with both `check()` and `check_edit()`
- [x] All flyimport tests pass

### DIFF
--- a/crates/ide-completion/src/completions/flyimport.rs
+++ b/crates/ide-completion/src/completions/flyimport.rs
@@ -130,7 +130,15 @@ pub(crate) fn import_on_the_fly_path(
         } => qualified,
         _ => return None,
     };
-    let potential_import_name = import_name(ctx);
+    let mut potential_import_name = import_name(ctx);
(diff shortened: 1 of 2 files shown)


ddmoney420:
Closing this PR — I opened too many contributions across projects. Apologies for the noise.

PR #21546, opened 2026-01-29
fix: generate function uses closure type directly for closure references

ddmoney420:
## Summary
Fixes the "Generate function" assist to use the closure type directly when a closure is passed by reference.

Fixes rust-lang/rust-analyzer#21288

## Problem
When using "Generate function" on `bar(&|x| true)`, the `&` prefix was kept unnecessarily.

## Solution
Modified `fn_arg_type` to detect closure references and strip the `&`, generating `impl Fn(_) -> bool` instead of `&impl Fn(_) -> bool`.

## Test plan
- [x] Added tests for `&closure` and `&mut closure` cases
- [x] All generate_function tests pass

### DIFF
--- a/crates/ide-assists/src/handlers/generate_function.rs
+++ b/crates/ide-assists/src/handlers/generate_function.rs
@@ -1145,8 +1145,16 @@ fn fn_arg_type(
         generic_params.extend(ty.generic_params(ctx.db()));
 
         if ty.is_reference() || ty.is_mutable_reference() {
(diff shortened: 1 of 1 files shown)


A4-Tacks:
I was unable to reproduce the unexpected semicolons in test-utils. Have you found the reason?

ChayimFriedman2:
This is not a real fix. Please do not submit AI-generated PRs without properly reviewing them.

PR #21554, opened 2026-01-30
Improve pgo builds

Ddystopia:
* feat: allow local absolute paths
* feat: allow measuring against multiple crates

---

I'm not as familiar with PGO, do I understand correctly that by running more commands, more measurements do accumulate and then get merged?

### DIFF
--- a/xtask/src/dist.rs
+++ b/xtask/src/dist.rs
@@ -11,7 +11,7 @@ use zip::{DateTime, ZipWriter, write::SimpleFileOptions};
 
 use crate::{
     date_iso,
-    flags::{self, Malloc, PgoTrainingCrate},
+    flags::{self, Malloc, PgoTrainingCrates},
     project_root,
     util::detect_target,
 };
@@ -102,7 +102,7 @@ fn dist_server(
     target: &Target,
     allocator: Malloc,
     zig: bool,
-    pgo: Option<PgoTrainingCrate>,
+    pgo: Option<PgoTrainingCrates>,
     dev_rel: bool,
 ) -> anyhow::Result<()> {
     let _e = sh.push_env("CFG_RELEASE", release);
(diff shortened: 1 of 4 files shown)


Ddystopia:
One possible improvement would be to report error in case the string passed to `--pgo` is an absolute path but doesn't exists.

ChayimFriedman2:
Do you actually need this, or just thought this is nice to have?

Ddystopia:
Well I used it to compile RA to better suit my projects. Does this count as "I actually need this"?

ChayimFriedman2:
Was there a measurable benefit?

Ddystopia:
I didn't benchmark it, I'll try to do this in a span of two weeks

ChayimFriedman2:
The reason I ask is that I suspect there won't be any perf benefit from customizing to a project (all rust projects look the same, when looking at the level of branch and cache predictions), therefore I don't want to accept this just because.

Ddystopia:
Yeah, I get the point, but my objective is that my projects are quite different as it is either wasm front-ends or embedded software with lots of automatically generated FFI bindings and ra often just dies. I would try to benchmark it against stock ra if there would be some significant benefit.

lnicola:
See https://github.com/rust-lang/rust-analyzer/pull/19585#issuecomment-2806856737, or better, run your own numbers.

ChayimFriedman2:
@Ddystopia Did you check that?

Ddystopia:
@ChayimFriedman2 

Hi, I did not, I am quite busy lately :sweat_smile: . I'll probably get around to this during this weekend. Note my previous comment from a week ago:
> I didn't benchmark it, I'll try to do this in a span of two weeks

Ddystopia:
I ran tests, there seem to be small gains, not too huge though. It also has some `uncaught type error`. I'll be using it locally, together with `target-cpu=native` (I didn't look into how to enable it during compilation with `xtastk` and will do it later, local-only). Those are some seconds so I'll take them, if you think it's not worth it I'm fine if you close the PR.

[simple-fw.txt](https://github.com/user-attachments/files/25328792/simple-fw.txt)
[complex-fw.txt](https://github.com/user-attachments/files/25328793/complex-fw.txt)
[simple-backend.txt](https://github.com/user-attachments/files/25328791/simple-backend.txt)
[complex-backend.txt](https://github.com/user-attachments/files/25328794/complex-backend.txt)

lnicola:
Is `simple` our release binary (PGOed on `clap`) and `complex` PGOed on specific project, plus `target-cpu=native`? I see a difference on `fw`, but not on `backend`.

Ddystopia:
Both are without target-cpu=native, simple is master with pgo on rust-analyzer (I thought you used it). And yeah I also only see the difference on fw.

PR #21561, opened 2026-01-31
fix: Fix more glob issues

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/19224

Definitely not the proper way to fix it, but as a stopgap until we rewrite all of this it should suffice as this bug makes using the windows crates a pain.


Veykril:
Not gonna merge this until monday in case this breaks name res

PR #21572, opened 2026-02-02
Implement the new homogeneous & heterogeneous try blocks

goffrie:
Match https://github.com/rust-lang/rust/pull/148725 / https://github.com/rust-lang/rust/pull/149489 from rustc nightly.
This adds parser support for the bikeshed syntax & desugaring.

### DIFF
--- a/crates/hir-def/src/expr_store/lower.rs
+++ b/crates/hir-def/src/expr_store/lower.rs
@@ -426,7 +426,7 @@ pub struct ExprCollector<'db> {
     /// and we need to find the current definition. So we track the number of definitions we saw.
     current_block_legacy_macro_defs_count: FxHashMap<Name, usize>,
 
-    current_try_block_label: Option<LabelId>,
+    current_try_block: Option<TryBlock>,
 
     label_ribs: Vec<LabelRib>,
     unowned_bindings: Vec<BindingId>,
@@ -470,6 +470,13 @@ impl RibKind {
 enum Awaitable {
     Yes,
     No(&'static str),
+}
+
+enum TryBlock {
+    // `try { ... }`
+    Homogeneous { label: LabelId },
+    // `try bikeshed Ty { ... }`
+    Heterogeneous { label: LabelId },
 }
 
 #[derive(Debug, Default)]
@@ -532,7 +539,7 @@ impl<'db> ExprCollector<'db> {
             lang_items: OnceCell::new(),
             store: ExpressionStoreBuilder::default(),
             expander,
-            current_try_block_label: None,
+            current_try_block: None,
             is_lowering_coroutine: false,
             label_ribs: Vec::new(),
             unowned_bindings: Vec::new(),
@@ -1069,7 +1076,9 @@ impl<'db> ExprCollector<'db> {
                 self.alloc_expr(Expr::Let { pat, expr }, syntax_ptr)
             }
             ast::Expr::BlockExpr(e) => match e.modifier() {
-                Some(ast::BlockModifier::Try(_)) => self.desugar_try_block(e),
+                Some(ast::BlockModifier::Try { try_token: _, bikeshed_token: _, result_type }) => {
+                    self.desugar_try_block(e, result_type)
+                }
                 Some(ast::BlockModifier::Unsafe(_)) => {
                     self.collect_block_(e, |id, statements, tail| Expr::Unsafe {
                         id,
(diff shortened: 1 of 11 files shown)


goffrie:
(i previously stated that type inference was broken, but I was just missing a minicore import)

crates/syntax/rust.ungram
```diff
@@ -473,7 +473,7 @@ TryExpr =
   Attr* Expr '?'
 
 BlockExpr =
-  Attr* Label? ('try' | 'unsafe' | ('async' 'move'?) | ('gen' 'move'?) | 'const') StmtList
+  Attr* Label? ('try' ('bikeshed' Type)? | 'unsafe' | ('async' 'move'?) | ('gen' 'move'?) | 'const') StmtList
```
ChayimFriedman2: This is a bit too complicated for my taste. I would prefer a separate `TryBlockModifier = 'try' ('bikeshed' Type)?`, or alternatively `HeterogeneousTryBlockModifier = 'bikeshed' Type`, I'll let you choose between the two.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -472,6 +472,11 @@ enum Awaitable {
     No(&'static str),
 }
 
+enum TryBlock {
+    Homogeneous { label: LabelId },
```
ChayimFriedman2, line 477: Probably good to add a comment demonstrating the syntax for each variant.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -1720,8 +1731,36 @@ impl<'db> ExprCollector<'db> {
...
+                // `{ let <name>: <ty> = <expr>; <name> }`
+                let name = self.generate_new_name();
+                let type_ref = self.lower_type_ref_disallow_impl_trait(ty);
+                let binding = self.alloc_binding(
+                    name.clone(),
+                    BindingAnnotation::Unannotated,
+                    HygieneId::ROOT,
+                );
+                let pat = self.alloc_pat_desugared(Pat::Bind { id: binding, subpat: None });
+                let tail_expr = self.alloc_expr(Expr::Path(Path::from(name)), ptr);
```
```suggestion
                let tail_expr = self.alloc_expr_desugared_with_ptr(Expr::Path(Path::from(name)), ptr);
```

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -1720,8 +1731,36 @@ impl<'db> ExprCollector<'db> {
...
+                self.alloc_expr(
```
```suggestion
                self.alloc_expr_desugared_with_ptr(
```

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -1720,8 +1731,36 @@ impl<'db> ExprCollector<'db> {
             unreachable!("block was lowered to non-block");
         };
         *tail = Some(next_tail);
-        self.current_try_block_label = old_label;
-        expr_id
+        self.current_try_block = old_try_block;
+        match result_type {
+            Some(ty) => {
+                // `{ let <name>: <ty> = <expr>; <name> }`
```
ChayimFriedman2, line 1739: Technically in the MIR forcing a type is not the same as introducing a new local variable, but our MIR is horribly broken anyway and I believe in typeck the effect should be the same.

crates/syntax/rust.ungram
goffrie: Fair enough, TryBlockModifier sounds good.

crates/hir-def/src/expr_store/lower.rs
goffrie: You are correct of course, but I don't think we have a construct for type ascription in an expression (which is what rustc uses) :')

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -1720,8 +1733,38 @@ impl<'db> ExprCollector<'db> {
...
+                let name = self.generate_new_name();
+                let type_ref = self.lower_type_ref_disallow_impl_trait(ty);
+                let binding = self.alloc_binding(
+                    name.clone(),
+                    BindingAnnotation::Unannotated,
+                    HygieneId::ROOT,
+                );
+                let pat = self.alloc_pat_desugared(Pat::Bind { id: binding, subpat: None });
+                self.add_definition_to_binding(binding, pat);
```
goffrie, line 1748: i think i was missing this `add_definition_to_binding` before but I confess that I don't know what it actually does.

PR #21581, opened 2026-02-03
internal: Run clippy as a separate CI step

Wilfred:
Currently clippy is run in CI as part of the macOS build. This is a little confusing, because clippy failures just show as "Rust (macos-latest)" which make it look like a macOS build failure.

Instead, treat clippy as a separate build step, like miri and rustfmt.

This should also make CI a little faster, because it reduces macOS runner usage (which tend to be slower than Linux on GitHub actions), and it reduces the number of steps where we need to install clippy.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -96,7 +96,7 @@ jobs:
         run: |
           rustup update --no-self-update stable
           rustup default stable
-          rustup component add --toolchain stable rust-src clippy rustfmt
+          rustup component add --toolchain stable rust-src rustfmt
           # We also install a nightly rustfmt, because we use `--file-lines` in
           # a test.
           rustup toolchain install nightly --profile minimal --component rustfmt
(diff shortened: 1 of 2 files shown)


Wilfred:
Huh, this has found a new clippy issue in r-a:

```
error: casting to the same type is unnecessary (`f64` -> `f64`)
    --> crates/hir-ty/src/mir/eval.rs:1628:62
     |
1628 |                         rustc_type_ir::FloatTy::F64 => Owned((value as f64).to_le_bytes().to_vec()),
     |                                                              ^^^^^^^^^^^^^^ help: try: `{ value }`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#unnecessary_cast
     = note: `-D clippy::unnecessary-cast` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(clippy::unnecessary_cast)]`
```

Wilfred:
I've resolved the above issue, it doesn't occur when rust-src is installed. I'm happy to send a PR to fix the clippy issue too, but I don't want to inundate y'all with PRs. Let me know if you have a preference :)

Veykril:
yea lets just fix that with this PR, though the suggested fix is weird, the braces will also be unnecessary there

Wilfred:
OK, fixed the clippy warning too.

lnicola:
It doesn't seem to be the case any more, but at one point macOS runners were actually faster than the x64 Linux ones.

PR #21583, opened 2026-02-04
fix: Fix loses associated bounds for replace_derive_with_manual_impl

A4-Tacks:
Fixes rust-lang/rust-analyzer#19878

Example
---
```rust
#[derive(Clo$0ne)]
struct Foo<T: core::ops::Deref>(T::Target);
```

**Before this PR**

```rust
struct Foo<T: core::ops::Deref>(T::Target);

impl<T: core::ops::Deref + Clone> Clone for Foo<T> {
    $0fn clone(&self) -> Self {
        Self(self.0.clone())
    }
}
```

**After this PR**

```rust
struct Foo<T: core::ops::Deref>(T::Target);

impl<T: core::ops::Deref + Clone> Clone for Foo<T>
where T::Target: Clone
{
    $0fn clone(&self) -> Self {
        Self(self.0.clone())
    }
}
```

### DIFF
--- a/crates/ide-assists/src/utils.rs
+++ b/crates/ide-assists/src/utils.rs
@@ -767,7 +767,7 @@ fn generate_impl_inner(
     });
     let generic_args =
         generic_params.as_ref().map(|params| params.to_generic_args().clone_for_update());
-    let trait_where_clause = trait_
+    let adt_assoc_bounds = trait_
         .as_ref()
         .zip(generic_params.as_ref())
         .and_then(|(trait_, params)| generic_param_associated_bounds(adt, trait_, params));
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/utils.rs
```diff
@@ -781,7 +787,7 @@ fn generate_impl_inner(
             false,
             trait_,
             ty,
-            None,
+            trait_where_clause,
```
ShoyuVanilla: This behaves correctly, but the variable name and passing it as the param with the same name might be misleading. We’re collecting clauses from the ADT here, not from the trait.

ShoyuVanilla (approved):
https://github.com/rust-lang/rust-analyzer/blob/9efc1f709f3c8134c3acac5d3592a8e4c184a0c6/crates/syntax/src/ast/make.rs#L324-L337

We might need to improve the above API as well to accept iterator or sth for where clauses but I think this PR is good as is for now

Issue #21582, opened 2026-02-04
Invisible delimiters after asm `sym` operand causes parsing errors


oxalica:
**rust-analyzer version**: 0.3.2711-standalone [/home/oxa/.config/VSCodium/User/globalStorage/rust-lang.rust-analyzer/rust-analyzer]

**rustc version**: rustc 1.93.0 (254b59607 2026-01-19)

**editor or extension**: VSCode

**repository link (if public, optional)**: https://github.com/oxalica/sjlj2/blob/8c9bfbb566f7e2b72250e0d007756715056d4da2/src/lib.rs#L360

**code snippet to reproduce**:

The following minimized code compiles fine with rustc, but rust-analyzer emits an error.

```rust
macro_rules! m {
    ($e:expr) => {
        core::arch::asm!("/*{f}*/", f = sym $e, out("ax") _)
    };
}

fn generic<T>() {}

fn main() {
    unsafe {
        m!(generic::<i32>); 
     // ^^ invalid `_` expression, expected type `()` [rust-analyzer(typed-hole)]
    }
}
```

The issue occurs when the next TT after `sym` is a invisible-delimited group captured by macro `expr` specifier. I have to admit that it is technically more correct to use `path` specifier in this case (and then rust-analyzer will parse everything fine), but `expr` should work since it is also accepted by rustc.

railgun-0402:
@oxalica 
Hi! I’d like to work on this. 

I’ll try to reproduce it locally and investigate how the sym operand is parsed when the next token is an invisible-delimited group produced by macro expansion. I’ll aim to add a regression test first, then propose a fix.

railgun-0402:
Hi! I opened a PR to address this issue: https://github.com/rust-lang/rust-analyzer/pull/21588.

When capturing a path via $e:expr, macro expansion treats it as Fragment::Expr and wraps it in parentheses to preserve operator precedence.

Feedback welcome 🙏

PR #21588, opened 2026-02-04
Fix asm sym operand parsing for parenthesized expr fragments

railgun-0402:
Fixes rust-lang/rust-analyzer#21582

## Summary
When capturing a path via `$e:expr`, macro expansion treats it as `Fragment::Expr` and wraps it in parentheses to preserve operator precedence.
This results in `sym (generic::<i32>)`, but the existing parser expected a path immediately after `sym` and failed when seeing `(`, leading to a parse failure and a bogus typed-hole diagnostic.

## Changes
- Parse the operand after `sym` as an expression
- During lowering, extract a path from the expression (so parenthesized forms are handled correctly)

## Why this approach
This keeps parsing more permissive while keeping the semantic requirement (“`sym` expects a path”) enforced during lowering.

## Testing
- Added/updated regression test for rust-lang/rust-analyzer#21582
- `cargo test` (and any focused test targets you ran)

### AI disclosure: 
I used ChatGPT to help draft the PR description and clarify the root-cause explanation. 
The implementation and tests were written and validated by me.

### DIFF
--- a/crates/ide-diagnostics/src/handlers/typed_hole.rs
+++ b/crates/ide-diagnostics/src/handlers/typed_hole.rs
@@ -442,4 +442,30 @@ fn rdtscp() -> u64 {
 }"#,
         );
     }
+
+    #[test]
+    fn asm_sym_with_macro_expr_fragment() {
+        // Regression test for issue #21582
+        // When `$e:expr` captures a path and is used in `sym $e`, the path gets
+        // wrapped in parentheses during macro expansion due to invisible delimiters.
+        // This should not cause false positive typed-hole errors.
+        check_diagnostics(
+            r#"
+//- minicore: asm
+macro_rules! m {
+    ($e:expr) => {
+        core::arch::asm!("/*{f}*/", f = sym $e, out("ax") _)
+    };
+}
+
+fn generic<T>() {}
+
+fn main() {
+    unsafe {
+        m!(generic::<i32>);
+    }
+}
+"#,
+        );
+    }
 }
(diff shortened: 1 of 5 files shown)


ChayimFriedman2:
Please note that we require disclosure of AI usage, if you used them.

oxalica:
I'm not sure if it is good to overly relax the syntax to accept any `sym <expr>`, which is not accepted by rustc currently. Note that rustc will reject it if you write `sym (func)` directly (see also [syntax reference](https://doc.rust-lang.org/reference/inline-assembly.html)).

railgun-0402:
Good point — I agree we shouldn’t accept a broader sym <expr> syntax than rustc.

The problematic case here is $e:expr in macro expansion, where the operand can be wrapped in invisible delimiters for precedence, even though users didn’t write parentheses.

I’ll adjust the change to keep the syntax strict (path-only), and only handle the invisible-delimiter wrapping so that sym (func) written directly is still rejected, matching rustc. I’ll also add a regression test for sym (func) to ensure we stay compatible.

railgun-0402:
Hi @oxalica — thanks for the earlier feedback!

I revised the approach to avoid relaxing `sym` syntax beyond rustc.
Instead of accepting `sym <expr>`, this PR keeps the operand path-only and fixes the macro-expansion behavior that introduced the problematic parentheses.

- In the macro transcriber, we avoid wrapping *path-like* expr fragments (e.g. `generic::<i32>`) in parentheses, since that breaks `asm!` `sym` operands coming from `$e:expr`.
  (Conservative heuristic: requires `::` and only allows path-like tokens.)
- Added a regression test for the `$e:expr` + `sym $e` case (issue #21582) to ensure no bogus typed-hole diagnostic.

Could you please take another look at the transcriber heuristic and the test coverage?
Thanks!

ChayimFriedman2:
No no no. If we're accepting that we should patch something until we fix macros properly (which I'm not sure is a good idea but maybe it is), it is definitely not a good idea to do that in macro_rules general handling. Change the parser to accept parentheses inside `sym` instead:

https://github.com/rust-lang/rust-analyzer/blob/4eac290b58a70961e78f2e0c04f61a08b995b2cb/crates/parser/src/grammar/expressions/atom.rs#L409-L413

railgun-0402:
Squashed!

[squashed my commits](https://github.com/rust-lang/rust-analyzer/pull/21588/commits)

PR #21594, opened 2026-02-06
feat: offer block let fallback postfix complete

A4-Tacks:
Related https://github.com/rust-lang/rust-analyzer/issues/5290#issuecomment-749487258

@flodiebold Can this close the issue?

Um... it's a bit strange if it's inside the chain call,
e.g `foo.bar().let.baz()` -> `{let $1 = foo.bar();$0}.baz()`

Or only offer when the parent is a MatchArm?

Example
---
```rust
fn main() {
    match 2 {
        bar => bar.$0
    }
}
```
->
```rust
fn main() {
    match 2 {
        bar => {
            let $1 = bar;
            $0
        }
    }
}
```


ChayimFriedman2:
I don't think we want this in general expressions, but inside a match arm it'll be nice to have both postfix and normal `let` (that complete the block as well), I've wanted this several times.

PR #21619, opened 2026-02-11
migrate covert_tuple_return_type to struct  assist to syntax editor

Shourya742:

### DIFF
--- a/crates/ide-assists/src/handlers/convert_tuple_return_type_to_struct.rs
+++ b/crates/ide-assists/src/handlers/convert_tuple_return_type_to_struct.rs
@@ -12,7 +12,11 @@ use ide_db::{
 };
 use syntax::{
     AstNode, SyntaxNode,
-    ast::{self, HasName, edit::IndentLevel, edit_in_place::Indent, syntax_factory::SyntaxFactory},
+    ast::{
+        self, HasName,
+        edit::{AstNodeEdit, IndentLevel},
+        syntax_factory::SyntaxFactory,
+    },
     match_ast,
     syntax_editor::SyntaxEditor,
 };
@@ -286,7 +290,7 @@ fn add_tuple_struct_def(
     let struct_def = syntax_factory.struct_(visibility, struct_name, None, field_list);
 
     let indent = IndentLevel::from_node(parent);
-    struct_def.reindent_to(indent);
+    let struct_def = struct_def.indent(indent);
 
     edit.insert(parent.text_range().start(), format!("{struct_def}\n\n{indent}"));
 }


crates/ide-assists/src/handlers/convert_tuple_return_type_to_struct.rs
```diff
@@ -5,15 +5,16 @@ use ide_db::{
...
+    imports::insert_use::{ImportScope, insert_use_with_editor},
     search::{FileReference, UsageSearchResult},
     source_change::SourceChangeBuilder,
     syntax_helpers::node_ext::{for_each_tail_expr, walk_expr},
 };
 use syntax::{
     AstNode, SyntaxNode,
-    ast::{self, HasName, edit::IndentLevel, edit_in_place::Indent, make},
-    match_ast, ted,
+    ast::{self, HasName, edit::IndentLevel, edit_in_place::Indent, syntax_factory::SyntaxFactory},
```
ShoyuVanilla: We should migrate this mutable `ast::edit_in_place::Indent::indent` usages to `ast:;edit::AstNodeEdit::indent` as well

bit-aloo: Thanks. I see that we use this in several other places as well, including in the already migrated assists. I’ll open a follow-up to move those over too.

bit-aloo:
Seems like miri is flagging crossbeam-utils: https://github.com/rust-lang/rust-analyzer/actions/runs/21934438687/job/63345527614?pr=21619

ShoyuVanilla:
Looks like it just fails while running build script for it, before miri do its own work 🤔 Maybe the GH cache is corrupted somehow?

~~Could you rebase this onto the latest master branch? I'm not sure if that would fix the problem but worth trying~~
edit: Nevermind. https://github.com/rust-lang/rust-analyzer/pull/21627 fails even though its parent commit is the latest master.

crates/ide-assists/src/handlers/convert_tuple_return_type_to_struct.rs
ShoyuVanilla: Yeah, IIRC, the immutable API for handling indentation has been implemented after some migrations already done and some migration works just missed it even though they got merged after it 😅

bit-aloo:
> Looks like it just fails while running build script for it, before miri do its own work 🤔 Maybe the GH cache is corrupted somehow?
> 
> ~Could you rebase this onto the latest master branch? I'm not sure if that would fix the problem but worth trying~ edit: Nevermind. #21627 fails even though its parent commit is the latest master.

Pinned the toolchain here: https://github.com/rust-lang/rust-analyzer/pull/21628, for miri job.

PR #21622, opened 2026-02-11
refactor: encapsulate TyLoweringContext usage in hir-ty

raushan728:
Waiting

### DIFF
--- a/crates/hir-ty/src/lib.rs
+++ b/crates/hir-ty/src/lib.rs
@@ -94,7 +94,7 @@ pub use infer::{
 };
 pub use lower::{
     GenericPredicates, ImplTraits, LifetimeElisionKind, TyDefId, TyLoweringContext, ValueTyDefId,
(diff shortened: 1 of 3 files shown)


ShoyuVanilla:
That test is just sensitive to query graph changes, not flaky. You could just update the expectation to the actual output

raushan728:
thanks for the correction @ShoyuVanilla . I've updated the test expectations. I realized my previous local run was mistaken—the test is indeed sensitive to the query changes from this fix.

ChayimFriedman2:
I'm really not pleased with the solution here, and given that we need to get rid of the distinction between `TyLoweringContext` and `InferenceContext` anyway (as I said on Zulip), I would much prefer to solve this by that. That is larger work, of course.

raushan728:
Thanks @ChayimFriedman2. I just saw your messages on Zulip, let's discuss it further there."

PR #21623, opened 2026-02-11
feat: offer on is_some_and for replace_is_method_with_if_let_method

A4-Tacks:
Example
---
```rust
fn main() {
    let x = Some(1);
    if x.is_som$0e_and(|it| it != 3) {}
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
fn main() {
    let x = Some(1);
    if let Some(it) = x && it != 3 {}
}
```


nikolaevx:
Looks like the predicate gets dropped when the argument is not a closure

A4-Tacks:
Okay, fixes in #21787

PR #21624, opened 2026-02-11
feat: custom auto-import order

eduardorittner:
This PR adds the ability to configure the order of auto-imports suggested by rust-analyzer. The changes are pretty simple, however there are some points, currently:

- There is no checking to make sure that all `ImportGroupKind`s are present in the config
- All `ImportGroupKind`s which are not explictly set in the config are given the same, lowest priority. This means that for a config like
```json
"group_order": {
    "std",
    "this-crate",
    "extern-crate",
    "one",
}
```
both "this-module" and "super-module" would be given the same priority, so the order in which they appear is probably random? An alternative to this behavior is to have all missing items have the same as their default priority except shifted down to be lower than all the explictly configured items. In this particular case, this would mean that "this-module" and "super-module" would have the second to lowest and lowest priorities, respectively.
- It seems a bit clumsy to have a fixed-size array, since adding new import groups requires changing all instances of this type, but at the same time the places where it's used are relatively few, I don't thing import groups are going to change that much over the years, and on top of that the dynamic nature of `Vec` is not needed. Thoughts?

Disclaimer: I used AI agents in writing the initial version of the code, but I changed most of it (it decided to use `Vec<String>` for the config which seemed like a bad decision).

Closes rust-lang/rust-analyzer#13960


eduardorittner:
For some reason clippy and cross compilation are failing to compile at all while all other jobs are fine. From what I can tell these failing jobs have stale code which does not reflect the actual contents of the branch. Not what the fix is

PR #21635, opened 2026-02-13
fix: Better import placement + merging

tascord:
fix stuff relating to imports.

1. `#[cfg]`'s act as hard barriers for import globbing
    ```rs
    use { /* ...*/ };

    #[cfg(target_arch = "x86_64")]
    pub mod api;
    // 🡴 The cfg'd mod here causes RA to no longer see the use above. Quick-action imports will be added to a NEW use block here, rather than at the top of the file.
    ```

2. `#[cfg]` attribute ordering matters (was a FIXME)
    ```rs
    #[cfg(a)] #[cfg(b)] use foo;
    #[cfg(b)] #[cfg(a)] use bar; // RA wouldn't merge these before
    ```

3. If an existing `#[cfg]` `use` block exists, RA will still prefer to create a new import block instead of merging, when in `#[cfg]`'d functions. (rust-lang/rust-analyzer#9391)
    ```rs
    #[cfg(target_os = "linux")]
    use existing::Item;

    #[cfg(target_os = "linux")]
    fn main() {
        // Auto-importing `NewItem` here would previously create a local import
    }
    ```

### DIFF
--- a/crates/ide-db/src/imports/merge_imports.rs
+++ b/crates/ide-db/src/imports/merge_imports.rs
@@ -4,7 +4,7 @@ use std::cmp::Ordering;
 use itertools::{EitherOrBoth, Itertools};
 use parser::T;
 use syntax::{
-    Direction, SyntaxElement, algo,
+    Direction, SyntaxElement, ToSmolStr, algo,
     ast::{
         self, AstNode, HasAttrs, HasName, HasVisibility, PathSegmentKind, edit_in_place::Removable,
         make,
@@ -691,10 +691,10 @@ pub fn eq_attrs(
     attrs0: impl Iterator<Item = ast::Attr>,
     attrs1: impl Iterator<Item = ast::Attr>,
 ) -> bool {
-    let mut attrs0: Vec<_> = attrs0.map(|attr| attr.syntax().text().to_string()).collect();
-    let mut attrs1: Vec<_> = attrs1.map(|attr| attr.syntax().text().to_string()).collect();
-    attrs0.sort();
-    attrs1.sort();
+    let mut attrs0: Vec<_> = attrs0.map(|attr| attr.syntax().text().to_smolstr()).collect();
+    let mut attrs1: Vec<_> = attrs1.map(|attr| attr.syntax().text().to_smolstr()).collect();
+    attrs0.sort_unstable();
+    attrs1.sort_unstable();
 
     attrs0 == attrs1
 }
--- a/crates/ide-db/src/imports/insert_use/tests.rs
+++ b/crates/ide-db/src/imports/insert_use/tests.rs
@@ -1529,6 +1529,16 @@ fn merge_gated_imports_with_different_values() {
 }
 
 #[test]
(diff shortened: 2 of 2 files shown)


tascord:
@Billy-Sheppard am i missing anything in this one?

Billy-Sheppard:
Does it solve this issue here?

```rust
    // `x`
    use foo::Foo; 

    #[cfg(target_arch = "x86_64")] // `y`
    use bar::Baz;
```

When I auto-import within a `#[cfg(target_arch = "x86_64")]` scope, I want it to go into the `y` import block when using `imports_granularity = "One"`.
Not into:
```rust
    #[cfg(target_arch = "x86_64")]
    fn buzz() {
        use buzz::Buzz;
    }
```

To make something like
```rust
    // `x`
    use foo::Foo; 

    #[cfg(target_arch = "x86_64")] // `y`
    use {
        bar::Baz,
        buzz::Buzz,
    };
```

JLesDev (approved):
LGTM 🚀

crates/ide-db/src/imports/merge_imports.rs
```diff
@@ -692,13 +692,12 @@ pub fn eq_attrs(
     attrs1: impl Iterator<Item = ast::Attr>,
 ) -> bool {
     // FIXME order of attributes should not matter
```
tascord: remove

ShoyuVanilla (commented):
Looks great overall 👍 

> 2\. `#[cfg]` attribute ordering matters (was a FIXME)
>
>         
>         #[cfg(a)] #[cfg(b)] use foo;
>         #[cfg(b)] #[cfg(a)] use bar; // RA wouldn't merge these before
Could you add a test for this case as well?

crates/ide-db/src/imports/merge_imports.rs
```diff
@@ -691,14 +691,12 @@ pub fn eq_attrs(
...
-        .flat_map(|attr| attr.syntax().descendants_with_tokens())
-        .flat_map(|it| it.into_token());
-    let attrs1 = attrs1
-        .flat_map(|attr| attr.syntax().descendants_with_tokens())
-        .flat_map(|it| it.into_token());
-    stdx::iter_eq_by(attrs0, attrs1, |tok, tok2| tok.text() == tok2.text())
+    let mut attrs0: Vec<_> = attrs0.map(|attr| attr.syntax().text().to_string()).collect();
+    let mut attrs1: Vec<_> = attrs1.map(|attr| attr.syntax().text().to_string()).collect();
+    attrs0.sort();
+    attrs1.sort();
```
ShoyuVanilla: nit: Might be slightly more performant?
```suggestion
    let mut attrs0: Vec<_> = attrs0.map(|attr| attr.syntax().text().to_smolstr()).collect();
    let mut attrs1: Vec<_> = attrs1.map(|attr| attr.syntax().text().to_smolstr()).collect();
    attrs0.sort_unstable();
    attrs1.sort_unstable();
```

PR #21647, opened 2026-02-14
fix: handle `ref mut` bindings in `contains_explicit_ref_binding`

Albab-Hasan:
the standalone `contains_explicit_ref_binding` function only checked for `BindingAnnotation::Ref`, missing `BindingAnnotation::RefMut`. this caused `let ref mut x = expr` to incorrectly take the coercion path instead of preserving the exact type of the rhs expression. the method version used for match arms already handles both `Ref` and `RefMut` correctly.

### DIFF
--- a/crates/hir-ty/src/infer/pat.rs
+++ b/crates/hir-ty/src/infer/pat.rs
@@ -673,7 +673,7 @@ impl<'db> InferenceContext<'_, 'db> {
 pub(super) fn contains_explicit_ref_binding(body: &Body, pat_id: PatId) -> bool {
     let mut res = false;
     body.walk_pats(pat_id, &mut |pat| {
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
Can you add a test?

Albab-Hasan:
@ChayimFriedman2 ok

Albab-Hasan:
@ChayimFriedman2 hey, is this good? pls let me know.

Albab-Hasan:
@ChayimFriedman2 hey is there anyway I can ask you some questions if you dont mind?

ChayimFriedman2:
On Zulip is a good idea.

Albab-Hasan:
@ChayimFriedman2 ok, do you have a org or smth?

ShoyuVanilla:
https://rust-lang.zulipchat.com/#narrow/stream/185405-t-compiler.2Frust-analyzer

PR #21649, opened 2026-02-15
fix: use `ExprIsRead::Yes` for rhs of ordinary assignments

Albab-Hasan:
the rhs of an ordinary assignment `x = *never_ptr` was inferred with `ExprIsRead::No`, which prevented `NeverToAny` coercion for place expressions of type `!`. this caused false type mismatches and missing divergence detection. the destructuring assignment path and let binding path both correctly use `ExprIsRead::Yes` for the rhs value, since the value is always consumed (read). this makes the ordinary assignment path consistent with both.


ChayimFriedman2 (approved):
True, the RHS of an assignment does constitute read (the LHS does not).

Thanks!

PR #21652, opened 2026-02-15
fix: Fix predicates of builtin derive traits with two parameters defaulting to `Self`

ChayimFriedman2:
I.e. `PartialEq` and `PartialOrd`.

I believe this should fix rust-lang/rust-analyzer#21641; @nicola-gigante can you check?


nicola-gigante:
Thanks for the quick fix! I'll check as soon as possible and let you know!

nicola-gigante:
@ChayimFriedman2 yes, it works now!

PR #21654, opened 2026-02-16
fix: use `ExprIsRead::Yes` for rhs of binary operators

Albab-Hasan:
the rhs of binary operators (including compound assignments like `+=`) was inferred with `ExprIsRead::No`, which prevented divergence detection for place expressions of type `!`. this caused false type mismatches when a function's return type depended on the block diverging after something like `x += *never_ptr`. the rhs of any binary operator is always consumed (read), so this uses `ExprIsRead::Yes` to be consistent with the ordinary assignment and let binding paths.

### DIFF
--- a/crates/hir-ty/src/tests/never_type.rs
+++ b/crates/hir-ty/src/tests/never_type.rs
@@ -826,8 +826,7 @@ fn binop_lhs_never_place_diverges() {
 fn foo() {
     unsafe {
         let p: *mut ! = 0 as _;
-        *p + 1;
-//      ^^ adjustments: NeverToAny
+        *p += 1;
     }
 }
 "#,


crates/hir-ty/src/infer/op.rs
```diff
@@ -178,9 +178,9 @@ impl<'a, 'db> InferenceContext<'a, 'db> {
                 // trait matching creating lifetime constraints that are too strict.
                 // e.g., adding `&'a T` and `&'b T`, given `&'x T: Add<&'x T>`, will result
                 // in `&'a T <: &'x T` and `&'b T <: &'x T`, instead of `'a = 'b = 'x`.
-                let lhs_ty = self.infer_expr_no_expect(lhs_expr, ExprIsRead::No);
+                let lhs_ty = self.infer_expr_no_expect(lhs_expr, ExprIsRead::Yes);
```
ChayimFriedman2, line 181: This is not correct (same below); the LHS of a (non-binop) assignment should not constitute read.

crates/hir-ty/src/tests/never_type.rs
```diff
@@ -802,6 +802,37 @@ fn foo() {
...
+        let mut x: i32 = 0;
+        x += *p;
+    }
+}
+"#,
+    );
+}
+
+#[test]
+fn binop_lhs_never_place_diverges() {
```
ChayimFriedman2, line 822: This... does not check what it declares?

Albab-Hasan: my bad. one sec, ill fix it

Albab-Hasan:
@ChayimFriedman2 hey, so im really confused rn.

> This... does not check what it declares?

I get this. i pushed the fix. but im not sure what ur trying to say in the first comment.

the changes on lines 181 and 183 are in the _ => branch thats regular binops like + not assignments. both operands of + are consumed by `Add::add` so they're read. i tried reverting those lines back to `ExprIsRead::No` and the `binop_lhs_never_place_diverges` test breaks *p + 1 where p: *const ! stops triggering NeverToAny coercion and get expected i32 got (). the ordinary assignment LHS (x in x = expr) is already `ExprIsRead::No` in expr.rs and wasnt touched here.

crates/hir-ty/src/tests/never_type.rs
ChayimFriedman2: Did you fix it?

Albab-Hasan: > Did you fix it?

yes

ChayimFriedman2: How? It still does not test coercion in the LHS.

The other comment was my mistake.

Albab-Hasan: sorry i just switched to `check_infer_with_mismatches` thinking that showingthe inferred types was enough.

Albab-Hasan: its fixed now

ChayimFriedman2: It seems there is a misunderstanding. The LHS in the assignment `a = b` is `a`. Asserting that a never coercion happens in the LHS means something like `*p += 1` where `*p` is `*const !`.

Albab-Hasan:
@ChayimFriedman2 can i get some instructions or should i close this pr and work on one liners again?

Albab-Hasan:
@ChayimFriedman2 sorry for wasting your time

ChayimFriedman2:
The change is good, only the test is problematic. A good test will be something like (not verified):
```rust
let a: *const ! = 0 as _;
*a += 1;
```

Albab-Hasan:
@ChayimFriedman2 ok ill try to fix the test

Albab-Hasan:
@ChayimFriedman2 can you review the changes? the test is fixed

crates/hir-ty/src/tests/never_type.rs
ChayimFriedman2: This test is still not correct as there is no assignment now.

Albab-Hasan:
@ShoyuVanilla hey can you check this one out too, i want to know if im missing anything?

crates/hir-ty/src/tests/never_type.rs
ShoyuVanilla, line 1: Well, this still isn't an assignment as Chayim said and more importantly, both tests aren't compiled with rustc.
I think the test code should be the valid Rust code which is compiled with rustc, unless the behavior we'd like to test is diagnostics or error resilience that not panicking or completing/inferencing the partially broken code.
I'd suggest trying out other code that compiled with rustc to test this PR.

Albab-Hasan:
@ChayimFriedman2 its done. let me know if i missed anything

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21654/744b554ce7dddfce22f77a1f07143cf3eaed0cd6..ab44d8263ef17a6e5d81e77db5f32c9bd4aa5029)*

PR #21660, opened 2026-02-17
internal: Configure codecov

Wilfred:
It would be wonderful to have test coverage tracking for rust-analyzer. I've added a basic configuration.

This will need adding a token to the secrets on this repository, but this is a good first step.

You can see a coverage report in my fork: https://app.codecov.io/github/Wilfred/rust-analyzer/commit/33f0a8efda0c2418343df02f590b42d08783ea89/tree

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,10 @@
+coverage:
+  range: 40...60
+  status:
+    patch: off
+    project:
+      default:
+        informational: true
+
+# Don't leave comments on PRs
+comment: false
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -1,0 +1,44 @@
+name: Coverage
+
+on: [pull_request, push]
+
+env:
+  CARGO_INCREMENTAL: 0
+  CARGO_NET_RETRY: 10
+  CI: 1
+  RUST_BACKTRACE: short
+  RUSTUP_MAX_RETRIES: 10
+
+jobs:
+  coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        run: |
+          rustup update --no-self-update nightly
+          rustup default nightly
+          rustup component add --toolchain nightly rust-src rustc-dev rustfmt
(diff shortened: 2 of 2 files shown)


Veykril:
cc @Kobzol re secret. Also does any other rust repo do code coverage yet? Curious if we already have in project solutions here to use

Kobzol:
Looks like crates.io have some codecov config (https://github.com/rust-lang/crates.io/blob/d1cd75749886a8c133af561e7a6f348d28679fcc/codecov.yml#L4), but I'm not sure if it's exercised in CI. @turbo87 any experiences with that?

Regarding the secret, opened a [Zulip thread](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Installing.20a.20secret.20in.20the.20rust-analyzer.20repo/with/574496916).

Turbo87:
> Looks like crates.io have some codecov config ([rust-lang/crates.io@`d1cd757`/codecov.yml#L4](https://github.com/rust-lang/crates.io/blob/d1cd75749886a8c133af561e7a6f348d28679fcc/codecov.yml#L4)), but I'm not sure if it's exercised in CI.

not anymore, no

> any experiences with that?

we tried it out for a while and it worked okayish, but it significantly impacted our compile and test times in CI since we were running it on PRs too. we ultimately removed it again from CI since we figured the tradeoff wasn't worth it at the moment but apparently forgot to remove the config file again.

Wilfred:
FWIW the coverage build here took 5 minutes, comparable to the normal test builds (Linux: 4 minutes, macOS: 7 minutes, Windows: 7 minutes).

I guess if PR performance was a concern we could configure coverage to only run on pushes on the master branch. That'd probably still give us most of the value.

Wilfred:
Zulip discussion for adding a secret: https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Installing.20a.20secret.20in.20the.20rust-analyzer.20repo/with/574501256

Wilfred:
CODECOV_TOKEN is now set as a repo secret, so I think this PR is good to go :)

Veykril:
Yea I also don't really trust the coverage % here

PR #21661, opened 2026-02-17
Add integration tests for flycheck

Wilfred:
rust-lang/rust-analyzer#21571 ended up introducing problems with stale diagnostics in cargo projects with custom check commands.

Flycheck isn't terribly well covered by tests today, so add some integration (aka heavy/slow) tests to ensure that basic functionality always works. They're not crazy slow, around a second runtime for each.

I want to follow up with additional tests for custom check diagnostics, but this gets things started.

AI disclosure: I wrote the first version of these tests with Claude and Opus, but I've reviewed all the changes and iterated on the details. Happy to make any other changes if you like.

### DIFF
--- a/crates/rust-analyzer/src/flycheck.rs
+++ b/crates/rust-analyzer/src/flycheck.rs
@@ -382,6 +382,7 @@ enum FlycheckCommandOrigin {
     ProjectJsonRunnable,
 }
 
+#[derive(Debug)]
 enum StateChange {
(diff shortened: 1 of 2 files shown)


Veykril:
Clippy isn't happy.

Thanks! we are severely lacking integration tests so this is a nice addition :)

Wilfred:
Fixed clippy :)

PR #21665, opened 2026-02-18
fix: prevent qualifying parameter names in add_missing_impl_members

raushan728:
Fixes rust-lang/rust-analyzer#21663

**What changed:**
The path transformation logic was incorrectly qualifying parameter names that matched definitions in scope (e.g., `ptr` becoming `std::ptr`).

I updated `transform_ident_pat` in `path_transform.rs` to skip the transformation for non-value definitions (like modules, functions, traits, non-unit structs, and non-unit variants) since they act as bindings in patterns. Valid pattern values (unit structs/variants, constants, and statics) are still transformed correctly.

Added a comprehensive test to verify all these edge cases.

### DIFF
--- a/crates/ide-db/src/path_transform.rs
+++ b/crates/ide-db/src/path_transform.rs
@@ -580,10 +580,10 @@ impl Ctx<'_> {
                     }
                 }
 
-                if let hir::ModuleDef::Variant(v) = def {
-                    if v.kind(self.source_scope.db) != hir::StructKind::Unit {
-                        return None;
-                    }
+                if let hir::ModuleDef::Variant(v) = def
+                    && v.kind(self.source_scope.db) != hir::StructKind::Unit
+                {
+                    return None;
                 }
 
                 let cfg = FindPathConfig {


raushan728:
hi @ChayimFriedman2 it's done. plz take a look when u  free

crates/ide-db/src/path_transform.rs
```diff
@@ -553,6 +553,11 @@ impl Ctx<'_> {
                     return None;
                 }
 
+                // Similarly, modules cannot be used in pattern position.
+                if matches!(def, hir::ModuleDef::Module(_)) {
```
A4-Tacks, line 557: Can this filter other definitions? For example, structures or functions?

A4-Tacks: BTW, I am filtering identifier patterns in #21411

raushan728: @A4-Tacks got it adding filters for structs, fns, etc. updating the PR shortly.

A4-Tacks: Note that in `fn foo(Foo: i32) {}`,
- `struct Foo {}`: `Foo` is the identify pattern
- `struct Foo;`: `Foo` required the transform (like `const Foo = ...`)

raushan728: that makes perfect sense! unit structs/constants can act as values in patterns, while braced structs act as bindings. I'll make sure the filter handles this distinction. thanks for catching that edge case

raushan728:
@A4-Tacks updated with new filters and tests

PR #21671, opened 2026-02-18
fix: offer block `.let` in ref-expr in match arm

A4-Tacks:
Related rust-lang/rust-analyzer#21600, rust-lang/rust-analyzer#21594

Example
---
```rust
fn main() {
    match 2 {
        bar => &bar.l$0
    }
}
```

**Before this PR**

```text
sn deref         *expr
sn match match expr {}
```

**After this PR**

```text
sn deref         *expr
sn let             let
sn letm        let mut
sn match match expr {}
```

### DIFF
--- a/crates/ide-completion/src/completions/postfix.rs
+++ b/crates/ide-completion/src/completions/postfix.rs
@@ -16,7 +16,7 @@ use itertools::Itertools;
 use stdx::never;
 use syntax::{
     SmolStr,
-    SyntaxKind::{BLOCK_EXPR, EXPR_STMT, STMT_LIST},
+    SyntaxKind::{EXPR_STMT, STMT_LIST},
     T, TextRange, TextSize, ToSmolStr,
     ast::{self, AstNode, AstToken},
     format_smolstr, match_ast,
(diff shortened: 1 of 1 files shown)


crates/ide-completion/src/completions/postfix.rs
```diff
@@ -111,13 +112,8 @@ pub(crate) fn complete_postfix(
...
 
     let try_enum = TryEnum::from_ty(&ctx.sema, receiver_ty);
-    let mut is_in_cond = false;
-    if let Some(parent) = dot_receiver_including_refs.syntax().parent()
-        && let Some(second_ancestor) = parent.parent()
-    {
-        if let Some(parent_expr) = ast::Expr::cast(parent) {
-            is_in_cond = is_in_condition(&parent_expr);
-        }
+    let is_in_cond = is_in_condition(&dot_receiver_including_refs);
```
ChayimFriedman2, line 115: I don't understand the change here. Why change `is_in_condition()` from the parent to the child?

A4-Tacks: `dot_receiver` = `a`

Old behavior is wrong, old:

- On `a.let`, `dot_receiver_including_refs` = `a`
- On `&a.let`, `dot_receiver_including_refs` = `&a.let`

So in #21600, it is necessary to consider that `second_ancestor` skipped `STMT_LIST` and arrived at `BLOCK_EXPR`

New:

- On `a.let`, `dot_receiver_including_refs` = `a.let`, `receiver_accessor` = `a.let`
- On `&a.let`, `dot_receiver_including_refs` = `&a.let`, `receiver_accessor` = `a.let`

PR #21675, opened 2026-02-18
fix: exclude macro refs in tests when excludeTests is enabled

akashchakrabortymsc-cmd:
Fixes rust-lang/rust-analyzer#21667

## Problem
`rust-analyzer.references.excludeTests` was not excluding 
references inside macros like `dbg!()`, `assert_eq!()`, `vec![]` 
even when they were inside a `#[test]` function.

## Root Cause
The old `is_name_ref_in_test` used `.ancestors()` which cannot 
cross macro boundaries.

## Fix
Use `sema.ancestors_with_macros()` instead, which correctly 
walks up the syntax tree across macro boundaries.


ChayimFriedman2 (changes requested):
The change is correct, but the test is not: the macros are not defined (our tests do not have access to std), which is why the references don't appear.

akashchakrabortymsc-cmd:
Hi, I've fixed the test to use a custom macro instead of std macros. Please review again!

PR #21681, opened 2026-02-19
internal: Skip rustfmt test if nightly toolchain is missing

Wilfred:
Currently the rustfmt slow test fails if you don't have a nightly toolchain installed. We install a nightly toolchain on CI, but it's a little confusing that tests can fail on a fresh checkout on a new laptop.

Instead, skip this test if there's no nightly toolchain.

AI disclosure: Written with a little help from Claude.

### DIFF
--- a/crates/rust-analyzer/tests/slow-tests/main.rs
+++ b/crates/rust-analyzer/tests/slow-tests/main.rs
@@ -22,6 +22,7 @@ mod testdir;
 
 use std::{collections::HashMap, path::PathBuf, time::Instant};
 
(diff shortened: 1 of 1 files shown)


Wilfred:
Argh, `cargo clippy` doesn't catch issues in the workspace. Hopefully CI should be green now.

lnicola:
This makes sense, but I don't understand why the test is green for me without an installed nightly.

EDIT: because `RUN_SLOW_TESTS`, not `SKIP_SLOW_TESTS` :sweat_smile:

PR #21692, opened 2026-02-22
fix: Fix another case where we forgot to put the type param for `PartialOrd` and `PartialEq` in builtin derives

ChayimFriedman2:
Should fix https://github.com/rust-lang/rust-analyzer/issues/21641#issuecomment-3940990752; @nicola-gigante can you check?


nicola-gigante:
Yes, it works, thanks! 👍

PR #21695, opened 2026-02-23
fix: Improve inserted order for trait_impl_redundant_assoc_item

A4-Tacks:
Example
---
```rust
trait Marker {
    fn foo();
    fn baz();
}
impl Marker for Foo {
    fn foo() {}
    fn missing() {}$0
    fn baz() {}
}
```

**Before this PR**

```rust
trait Marker {
    fn missing();
    fn foo();
    fn baz();
}
impl Marker for Foo {
    fn foo() {}
    fn missing() {}
    fn baz() {}
}
```

**After this PR**

```rust
trait Marker {
    fn foo();
    fn missing();
    fn baz();
}
impl Marker for Foo {
    fn foo() {}
    fn missing() {}
    fn baz() {}
}
```

### DIFF
--- a/crates/ide-diagnostics/src/handlers/trait_impl_redundant_assoc_item.rs
+++ b/crates/ide-diagnostics/src/handlers/trait_impl_redundant_assoc_item.rs
@@ -5,8 +5,10 @@ use ide_db::{
     label::Label,
     source_change::SourceChangeBuilder,
 };
-use syntax::ToSmolStr;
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
I'm not sure we want this complexity. This is a small thing and if the items aren't in the same order as the trait it's not helpful anyway.

flodiebold:
IMO it's worth it, and the complexity is pretty well-contained.

PR #21699, opened 2026-02-24
fix: Detect E0804 when casting raw ptr-to-dyn adds auto traits

Albab-Hasan:
the `check_ptr_ptr_cast` function had an early return when the source and destination dyn trait objects shared the same principal trait def id:

    if src_principal == dst_principal {
        return Ok(());
    }

this bypassed all three remaining checks that the code comment explicitly listed as required: generic argument equality, projection equality and the auto-trait superset check. as a resultcasts like `*mut dyn Trait as *mut (dyn Trait + Send)` were silently accepted instead of emitting e0804.

the fix removes the early return. `CastError::PtrPtrAddingAutoTraits` and its diagnostic were already implemented they just couldnt be reached. the supertrait case (`trait Trait: Send`) continues to work correctly because the auto trait check already accounts for implied auto traits via supertrait elaboration.

### DIFF
--- a/crates/hir-ty/src/infer/cast.rs
+++ b/crates/hir-ty/src/infer/cast.rs
@@ -328,11 +328,7 @@ impl<'db> CastCheck<'db> {
                     //
                     // Note that trait upcasting goes through a different mechanism (`coerce_unsized`)
                     // and is unaffected by this check.
-                    (Some(src_principal), Some(dst_principal)) => {
-                        if src_principal == dst_principal {
-                            return Ok(());
-                        }
-
+                    (Some(src_principal), Some(_)) => {
                         // We need to reconstruct trait object types.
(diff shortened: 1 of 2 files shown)


Albab-Hasan:
@ChayimFriedman2 hey, can you check this out?

ShoyuVanilla (approved):
IIRC, this was added deliberately to suppress the false positives, but yeah, I guess this would be good to be removed as our type inferences is mature enough by now.
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21699/05da4cf3c4dfb8aa3430343f41e73c4cfad46bd5..da7b8bbec2d7ebde73da260ef90a22882050f873)*

PR #21705, opened 2026-02-25
fix: cfg_select supports non token-tree tokens

A4-Tacks:
Fixes rust-lang/rust-analyzer#21702

- Fix single non-curly groups is ignored `true => ((),)`, old: `(),`, new: `((),)`

Example
---
```rust
const _: i32 = cfg_select! { true => 2 + 3, _ => 3 + 4 };
```

**Before this PR**

```rust
const _: i32 = cfg_select! { true => 2 /* expected a token tree after `=>` */+ 3, _ => 3 + 4 };
```

**After this PR**

```rust
const _: i32 = 2 + 3;
```

### DIFF
--- a/crates/hir-expand/src/builtin/fn_macro.rs
+++ b/crates/hir-expand/src/builtin/fn_macro.rs
@@ -381,16 +381,40 @@ fn cfg_select_expand(
                 );
             }
         }
-        let expand_to_if_active = match iter.next() {
-            Some(tt::TtElement::Subtree(_, tt)) => tt.remaining(),
-            _ => {
+        let expand_to_if_active = match iter.peek() {
+            Some(tt::TtElement::Subtree(sub, tt)) if sub.delimiter.kind == DelimiterKind::Brace => {
+                iter.next();
+                tt.remaining()
+            }
+            None | Some(TtElement::Leaf(tt::Leaf::Punct(tt::Punct { char: ',', .. }))) => {
                 let err_span = iter.peek().map(|it| it.first_span()).unwrap_or(span);
+                iter.next();
                 return ExpandResult::new(
                     tt::TopSubtree::empty(tt::DelimSpan::from_single(span)),
                     ExpandError::other(err_span, "expected a token tree after `=>`"),
                 );
             }
+            Some(_) => {
+                let expr = expect_fragment(
+                    db,
+                    &mut iter,
+                    parser::PrefixEntryPoint::Expr,
+                    tt.top_subtree().delimiter.delim_span(),
+                );
+                if let Some(err) = expr.err {
+                    return ExpandResult::new(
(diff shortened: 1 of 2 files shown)


crates/hir-expand/src/builtin/fn_macro.rs
```diff
@@ -384,16 +384,34 @@ fn cfg_select_expand(
...
+            None | Some(TtElement::Leaf(tt::Leaf::Punct(tt::Punct { char: ',', .. }))) => {
                 let err_span = iter.peek().map(|it| it.first_span()).unwrap_or(span);
+                iter.next();
                 return ExpandResult::new(
                     tt::TopSubtree::empty(tt::DelimSpan::from_single(span)),
                     ExpandError::other(err_span, "expected a token tree after `=>`"),
                 );
             }
+            Some(_) => {
+                let savepoint = iter.savepoint();
```
ChayimFriedman2: That's not how rustc does this and this is not the right way. This will break on e.g. paths with commas like `foo::<T, U>`. You need to parse an expression with the parser.

A4-Tacks: But it still needs to support pat, so I roughly delimited it with commas

ChayimFriedman2: What do you mean by "it still need to support pat"? You need to use the parser to parse an expr like with macro_rules matcher.

A4-Tacks: Like `let cfg_select! { _ => () } = ();`,
but now I realize that it's actually parsing according to the expression, it just happens to be a valid pat

crates/hir-expand/src/builtin/fn_macro.rs
```diff
@@ -413,6 +423,39 @@ fn cfg_select_expand(
                 "none of the predicates in this `cfg_select` evaluated to true",
             ),
         ),
+    };
+
+    /// Parse one expression, skip comma in turbofish
+    fn parse_expr<'a>(iter: &mut TtIter<'a>) -> tt::TokenTreesView<'a> {
```
ChayimFriedman2: That is definitely not enough. The only way to parse an expression reliably is to invoke the parser.

A4-Tacks: > That is definitely not enough.

Is there anything else besides the closure returning the location?

> The only way to parse an expression reliably is to invoke the parser.

Do we have an expression parser for TtIter?

ChayimFriedman2: > Is there anything else besides the closure returning the location?

What do you mean?

> Do we have an expression parser for TtIter?

No, but you can look at the macro_rules matcher. Look at the function named `expect_fragment()`.

ChayimFriedman2 (approved):

PR #21716, opened 2026-02-27
fix: Use -Zlockfile-path on every 1.95 nightly

lnicola:
Closes rust-lang/rust-analyzer#21715


lnicola:
image

Oh, we'll probably have to backport this.

ShoyuVanilla:
Aaargh sorry 😢

demosdemon:
Will this break once 1.95 is beta?

ShoyuVanilla:
> Will this break once 1.95 is beta?

No, the purpose of this PR is exactly to make it never happen

PR #21719, opened 2026-02-27
fix: Improve newline for make `match`, `if`, `while` etc

A4-Tacks:
- Remove unused `ast::make::match_arm_with_guard`

Example
---
```rust
fn main() {
    if true {
        $0if true
            && false
        {
            foo()
        }
    }
}
```

**Before this PR**

```rust
fn main() {
    if true {
        match true
            && false {
            true => foo(),
            false => (),
        }
    }
}
```

**After this PR**

```rust
fn main() {
    if true {
        match true
            && false
        {
            true => foo(),
            false => (),
        }
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/move_guard.rs
+++ b/crates/ide-assists/src/handlers/move_guard.rs
@@ -567,7 +567,8 @@ fn main() {
     match 92 {
         x => if true
             && true
-            && true {
+            && true
+        {
             {
                 false
             }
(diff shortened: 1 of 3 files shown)


ChayimFriedman2 (approved):
Generally I'm not a fan of whitespace fixes and I prefer to let rustfmt do the job, but this one is simple enough.

*[View changes since this review](                                                                                                                                                                       )*

PR #21726, opened 2026-03-01
Fix: align `is_rust()` with rustc by correcting constructor ABI in next solver

Albab-Hasan:
resolves a TODO in `crates/hir-ty/src/next_solver/abi.rs` .

TODO: rustc does not consider `RustCall` to be true here but chalk does the `rustc_type_ir::inherent::Abi` trait documents `is_rust()` as * "if this ABI is `extern "Rust"`"*. `RustCall` is `extern "rust-call"` a specific ABI used internally for the closure calling convention so returning `true` for it violates the trait contract and diverges from rustcs behavior.

cause:  `fn_sig_for_struct_constructor` and `fn_sig_for_enum_variant_constructor` in `lower.rs` were assigning `FnAbi::RustCall` to tuple struct and enum variant constructors a holdover from the chalk era. in rustc these constructors have `extern "Rust"` ABI. because of this `is_rust()` was forced to include `RustCall` so that constructors would still pass the `is_fn_trait_compatible()` check in the next solver (used at 5 call sites in `rustc_next_trait_solver::solve::assembly::structural_traits` to gate `Fn`/`FnMut`/`FnOnce` impl resolution for `FnDef` and `FnPtr` types).

fix:
- change struct and enum variant constructor signatures to use `FnAbi::Rust`
- change `is_rust()` to `matches!(self, FnAbi::Rust)`, matching rustc's semantics

added two tests comfirming tuple struct and enum variant constructors correctly implement `Fn` traits through the next solver after this change.

### DIFF
--- a/crates/hir-ty/src/lower.rs
+++ b/crates/hir-ty/src/lower.rs
@@ -2468,7 +2468,7 @@ fn fn_sig_for_struct_constructor(
     let inputs_and_output =
         Tys::new_from_iter(DbInterner::new_no_crate(db), params.chain(Some(ret.as_ref())));
     StoredEarlyBinder::bind(StoredPolyFnSig::new(Binder::dummy(FnSig {
-        abi: FnAbi::RustCall,
+        abi: FnAbi::Rust,
         c_variadic: false,
         safety: Safety::Safe,
         inputs_and_output,
@@ -2487,7 +2487,7 @@ fn fn_sig_for_enum_variant_constructor(
     let inputs_and_output =
         Tys::new_from_iter(DbInterner::new_no_crate(db), params.chain(Some(ret.as_ref())));
     StoredEarlyBinder::bind(StoredPolyFnSig::new(Binder::dummy(FnSig {
-        abi: FnAbi::RustCall,
+        abi: FnAbi::Rust,
         c_variadic: false,
         safety: Safety::Safe,
         inputs_and_output,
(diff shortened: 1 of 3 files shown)


Albab-Hasan:
rustc's actual implementation in compiler/rustc_middle/src/ty/context.rs:

```
impl<'tcx> rustc_type_ir::inherent::Abi<TyCtxt<'tcx>> for ExternAbi {
    fn rust() -> Self {
        ExternAbi::Rust
    }

    fn is_rust(self) -> bool {
        matches!(self, ExternAbi::Rust)
    }
}
```

my fix is pretty much the same in structure:

```
fn is_rust(self) -> bool {
    matches!(self, FnAbi::Rust)
}
```

Albab-Hasan:
@ChayimFriedman2 hey, can you check this out? if you have some free time.

ChayimFriedman2 (approved):

ChayimFriedman2:
Thanks!

Albab-Hasan:
@ChayimFriedman2 thanks for checking it out. also is there i anyway i can talk to you in zulip? i have some questions to ask if u dont mind. i cant seem to find u there.

ChayimFriedman2:
I'm available on Zulip (Chayim Refael Friedman), but if you have questions, it's better to ask them in the public, so that more people can answer and learn.

PR #21738, opened 2026-03-03
fix: Stale diagnostics when a custom check command is configured

Wilfred:
We can't use the flycheck scope, because that value varies depending on how the flycheck was triggered. See also rust-lang/rust-analyzer#21571, which was reverted due to issues with scope.

Instead, treat empty diagnostics as a flycheck for the entire workspace, add comments explaining the JSON diagnostic format, and add an integration test.


Wilfred:
Fixed the issue link to keep rustbot happy.

ChayimFriedman2:
@Veykril Did you mean to merge this?

Veykril:
wanted to wait until the next stable release

PR #21740, opened 2026-03-04
feat: Allow crate authors to declare that their trait prefers to be imported `as _`

ChayimFriedman2:
For example for extension traits.

Provide an attribute for that. It'll affect flyimport and the autoimport quickfix, as explained in the code.

The origin of this is me being angry at the next solver traits that have names identical to the types and make autoimport a nightmare, especially in completion. Then in general we don't do a good job for traits that would be better of imported `as _`, I at least leave them imported normally because it's more convenient for flyimport. So this fixes that.

### DIFF
--- a/docs/book/src/contributing/lsp-extensions.md
+++ b/docs/book/src/contributing/lsp-extensions.md
@@ -1,5 +1,5 @@
 <!---
-lsp/ext.rs hash: 235f56089da3dbb5
+lsp/ext.rs hash: dc4ba5f417c74aa6
 
 If you need to change the above hash to make the test pass, please check if you
 need to adjust this doc as well and ping this issue:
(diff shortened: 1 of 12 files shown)


docs/book/src/contributing/lsp-extensions.md
```diff
@@ -1,5 +1,5 @@
 <!---
-lsp/ext.rs hash: 235f56089da3dbb5
```
ChayimFriedman2, line 2: This file didn't contain info about `CompletionImport` at all, so I just updated the hash.

ChayimFriedman2:
Can this be reviewed please? I'd like to use this for the solver's trait.

ShoyuVanilla (approved):
I think this feature is good to have

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21740/05808ac8271197d2d33f65918a16dd0643af6fba..f42e02aaff84d7b613ab9057036e93010c23b93f)*

PR #21741, opened 2026-03-04
Add opt-in eager diagnostic invalidation for agentic workflows

alliprice:
## Summary

Adds an opt-in `diagnostics.eagerInvalidation` setting that immediately clears native diagnostics for a file when its content changes, before reanalysis completes.

## Motivation

The current debounce behavior is correct for GUI editors - it prevents diagnostic flicker while a human is typing. This default should remain unchanged.

However, LSP is increasingly used by headless agentic tools (AI coding assistants, automated refactoring systems) that consume diagnostics programmatically. For these clients, stale diagnostics between an edit and reanalysis are incorrect input to decision loops. This can cause:

- Re-fixing already fixed errors
- Missing newly introduced errors
- Looping on diagnostics that no longer apply

This is not about servers being buggy - it is about different client requirements. Interactive editors need stable UI. Programmatic clients need immediate staleness signals.

## Changes

- **New setting**: `diagnostics.eagerInvalidation` (default: `false`)
  - When enabled, `textDocument/didChange` triggers an immediate clear of native diagnostics for the changed file
  - Server continues normal analysis pipeline
  - Fresh diagnostics arrive once reanalysis completes

- **Implementation**:
  - Added `diagnostics_eagerInvalidation` to `config.rs` with accessor `eager_diagnostic_invalidation()`
  - Hooked into `handle_did_change_text_document` in `handlers/notification.rs`
  - Reuses existing `diagnostics.clear_native_for()` path to publish empty diagnostics
  - Only clears if file ID can be resolved from VFS

- **Tests**: New integration test verifies diagnostics are cleared synchronously when setting is enabled

- **Documentation**: Updated generated configuration docs

## Configuration

```json
{
  "rust-analyzer.diagnostics.eagerInvalidation": false
}
```

Setting this to `true` is intended for non-UI rapid-edit workflows such as agentic coding tools.

## Test Plan

1. Run existing test suite to verify no regressions in default behavior: `cargo test`
2. Run new integration test to verify eager clearing: `cargo test --test slow-tests test_eager_diagnostic_invalidation_on_did_change`
3. Manual verification in VS Code or other LSP client:
   - With setting disabled: observe normal debounced diagnostic updates
   - With setting enabled: observe immediate diagnostic clearing on edit

## AI Disclosure

AI tools (Claude, Anthropic) were used for code authoring, test writing, and documentation. All changes were reviewed, understood, and validated by a human.

## Related Work

This is part of a coordinated effort across multiple LSP server ecosystems to support agentic workflows. See: https://github.com/lsp-eager-invalidation

Similar PRs:
- typescript-language-server (TypeScript)
- pyright (Python)
- gopls (Go)

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -701,6 +701,11 @@ config_data! {
         /// List of rust-analyzer diagnostics to disable.
         diagnostics_disabled: FxHashSet<String> = FxHashSet::default(),
 
+        /// Clear native diagnostics for a file as soon as its content changes, before
+        /// reanalysis completes. This prevents stale diagnostics from lingering between
+        /// an edit and the next analysis pass.
+        diagnostics_eagerInvalidation: bool = false,
+
         /// Show native rust-analyzer diagnostics.
         diagnostics_enable: bool = true,
 
@@ -711,11 +716,6 @@ config_data! {
         /// Map of prefixes to be substituted when parsing diagnostic file paths. This should be the
         /// reverse mapping of what is passed to `rustc` as `--remap-path-prefix`.
         diagnostics_remapPrefix: FxHashMap<String, String> = FxHashMap::default(),
-
-        /// Clear native diagnostics for a file as soon as its content changes, before
-        /// reanalysis completes. This prevents stale diagnostics from lingering between
-        /// an edit and the next analysis pass.
-        diagnostics_eagerInvalidation: bool = false,
 
         /// Run additional style lints.
         diagnostics_styleLints_enable: bool = false,
--- a/docs/book/src/configuration_generated.md
+++ b/docs/book/src/configuration_generated.md
@@ -591,16 +591,6 @@ Map of prefixes to be substituted when parsing diagnostic file paths. This shoul
 reverse mapping of what is passed to `rustc` as `--remap-path-prefix`.
 
 
-## rust-analyzer.diagnostics.eagerInvalidation {#diagnostics.eagerInvalidation}
-
-Default: `false`
-
-Clear native diagnostics for a file as soon as its content changes, before
-reanalysis completes. This prevents stale diagnostics from lingering between
-an edit and the next analysis pass. Intended for non-UI rapid-edit workflows
-such as agentic coding tools.
-
-
 ## rust-analyzer.diagnostics.styleLints.enable {#diagnostics.styleLints.enable}
 
 Default: `false`


ChayimFriedman2:
These tools can turn off flycheck then trigger it manually when needed (we have a command for that).

ChayimFriedman2:
As an unrelated note, while AI usage is fine (with proper control and disclosure), I'd prefer that text written for communication, e.g. the PR description, will be written without AI help (errors are fine). Although we don't have a consensus for that yet; CC @rust-lang/rust-analyzer.

lnicola:
> These tools can turn off flycheck then trigger it manually when needed

I'm not sure they can. They could (or they could pull the diagnostics), which leads to another question: who will teach all of these agents to enable the server-specific settings?

ShoyuVanilla:
Native diagnostics are from rust-analyzer. While we can turn off the flycheck and feed the AI agents directly with outputs from cargo, we can't do the simliar thing to the native ones.
So, this one might be useful though I have no idea how often the stale diagnostics mess up with the AI agents, as our native diagnostics are quite fast.

IMHO, using AI for code or PR descriptions is fine in itself. However, the author should carefully review the generated output to ensure it is correct and not unnecessarily verbose.

But submitting a PR without even running tests or clippy makes it hard to believe that the changes were properly reviewed.
Of course 100% human written PRs are sometimes misseing formatting, clippy or tests but while they are mere mistakes, for AI composed ones they are the lack of the evidence for self-reviewing.

lnicola:
> While we can turn off the flycheck and feed the AI agents directly with outputs from cargo, we can't do the simliar thing to the native ones.

The native diagnostics are incomplete and have a bunch of false positives too, I think they're actually counter-productive for this use case.

ShoyuVanilla:
> The native diagnostics are incomplete and have a bunch of false positives too, I think they're actually counter-productive for this use case.

Yeah, I guess so, too. But I'm not so confident with that as I've never tried that 😅

Veykril:
To be honest I think if you have an agent running rust-analyzer you should either disable rust-analyzer's diagnostics or disable the diagnostic tool from the agent and have it invoke plain cargo. rust-analyzer's native diagnostics are a) too slow for agents, b) sometimes incorrect, c) incomplete, and the flycheck diagnostics can hide useful information from the actual cargo/rustc output. This change doesn't even help that much imo, this clears diagnostics in the buffer that was changed but does not do so for other buffers whose diagnostics might change from the edit, those are still delayed and may still confuse the agent. So overall I do not believe this to be as useful as the author may expect.

lnicola:
> have it invoke plain cargo

Yeah, assuming `cargo check` is enough. Maybe the agent should ask the AI for a diagnostics command and run it automatically after edits.

alliprice:
Thanks for the detailed feedback.

On CI, the clippy lint and codegen regeneration are fixed on the branch now. I had my own tests passing, but that was my miss on process.

On the transitive effects, I think that's good criticism and potentially an argument not to do this at all. But my hunch is that this would solve the vast majority of cases and allow agents to receive majority value from the native diagnostics.

On native vs flycheck, I get that native diagnostics are the unreliable tier. Clearing stale unreliable diagnostics still seems better than leaving stale unreliable diagnostics, but I hear the argument that agents maybe shouldn't rely on native diagnostics at all. Is there room for this as a narrow improvement, or is the consensus position that agentic clients should bypass rust-analyzer diagnostics entirely?  

On "who teaches this", this is config ideally set by tool vendors (Claude Code, Cursor, etc.) in their LSP initialization, not something individual agents or users would toggle

PR #21744, opened 2026-03-04
fix: keep comments for 'Fill match arms'

A4-Tacks:
Fixes rust-lang/rust-analyzer#21718

Example
---
```rust
enum E { A, B, C }

fn foo(t: E) -> i32 {
    match $0t {
        // variant a
        E::A => 2
        // comment on end
    }
}
```

**Before this PR**

```rust
enum E { A, B, C }

fn foo(t: E) -> i32 {
    match t {
        E::A => 2,
        E::B => ${1:todo!()},
        E::C => ${2:todo!()},$0
    }
}
```

**After this PR**

```rust
enum E { A, B, C }

fn foo(t: E) -> i32 {
    match t {
        // variant a
        E::A => 2,
        // comment on end
        E::B => ${1:todo!()},
        E::C => ${2:todo!()},$0
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/add_missing_match_arms.rs
+++ b/crates/ide-assists/src/handlers/add_missing_match_arms.rs
@@ -263,8 +263,8 @@ pub(crate) fn add_missing_match_arms(acc: &mut Assists, ctx: &AssistContext<'_>)
             let mut arms_edit = ArmsEdit { match_arm_list, place: old_place, last_arm: None };
 
             arms_edit.remove_wildcard_arms(ctx, &mut editor);
-            arms_edit.add_comma_after_last_arm(ctx, &mut editor);
-            arms_edit.append_arms(&missing_arms, &mut editor);
+            arms_edit.add_comma_after_last_arm(ctx, &make, &mut editor);
+            arms_edit.append_arms(&missing_arms, &make, &mut editor);
 
             if let Some(cap) = ctx.config.snippet_cap {
                 if let Some(it) = missing_arms
@@ -375,7 +375,7 @@ impl ArmsEdit {
         }
     }
 
-    fn append_arms(&self, arms: &[ast::MatchArm], editor: &mut SyntaxEditor) {
+    fn append_arms(&self, arms: &[ast::MatchArm], make: &SyntaxFactory, editor: &mut SyntaxEditor) {
         let Some(mut before) = self.place.last_token() else {
             stdx::never!("match arm list not contain any token");
             return;
@@ -389,8 +389,8 @@ impl ArmsEdit {
             !self.place.text().contains_char('\n') || before.kind() == SyntaxKind::WHITESPACE;
         let indent = IndentLevel::from_node(&self.place);
         let arm_indent = indent + 1;
-        let indent = make::tokens::whitespace(&format!("\n{indent}"));
-        let arm_indent = make::tokens::whitespace(&format!("\n{arm_indent}"));
+        let indent = make.whitespace(&format!("\n{indent}"));
+        let arm_indent = make.whitespace(&format!("\n{arm_indent}"));
         let elements = arms
             .iter()
             .flat_map(|arm| [arm_indent.clone().into(), arm.syntax().clone().into()])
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/add_missing_match_arms.rs
```diff
@@ -372,6 +354,96 @@ fn cursor_at_trivial_match_arm_list(
...
+        ctx: &AssistContext<'_>,
+        node: &impl AstNode,
+    ) -> Option<std::ops::RangeInclusive<syntax::SyntaxElement>> {
+        let range = ctx.sema.original_range_opt(node.syntax())?;
+
+        if !self.place.text_range().contains_range(range.range) {
+            return None;
+        }
+
+        Some(utils::cover_edit_range(&ast::AnyAstNode(self.place.clone()), range.range))
```
ChayimFriedman2: A better solution to the dubious `AnyAstNode` is to just make this function take a `&SyntaxNode`, since it doesn't really need any specific node.

crates/ide-assists/src/handlers/add_missing_match_arms.rs
```diff
@@ -372,6 +354,96 @@ fn cursor_at_trivial_match_arm_list(
...
+}
+
+impl ArmsEdit {
+    fn remove_wildcard_arms(&mut self, ctx: &AssistContext<'_>, editor: &mut SyntaxEditor) {
+        for arm in self.match_arm_list.arms() {
+            if !matches!(arm.pat(), Some(Pat::WildcardPat(_))) {
+                self.last_arm = Some(arm);
+                continue;
+            }
+            if !arm.expr().is_none_or(is_empty_expr) {
```
ChayimFriedman2, line 355: That seems wrong? You need one `if !matches!(arm.pat(), Some(Pat::WildcardPat(_))) && !arm.expr().is_none_or(is_empty_expr)`, otherwise `self.last_arm` will be set for empty expr.

crates/ide-assists/src/handlers/add_missing_match_arms.rs
```diff
@@ -372,6 +354,96 @@ fn cursor_at_trivial_match_arm_list(
...
+        if let Some(prev) = before.prev_token()
+            && prev.kind() == SyntaxKind::WHITESPACE
+        {
+            before = prev;
+        }
+        let open_curly =
+            !self.place.text().contains_char('\n') || before.kind() == SyntaxKind::WHITESPACE;
+        let indent = IndentLevel::from_node(&self.place);
+        let arm_indent = indent + 1;
+        let indent = make::tokens::whitespace(&format!("\n{indent}"));
```
ChayimFriedman2: Use a `SyntaxFactory`, here and next line?

crates/ide-assists/src/handlers/add_missing_match_arms.rs
A4-Tacks:
```
filter(|_| if is_wildcard {
    if is_empty {
        false
    } else {
        true
    }
} else {
    true
})
```
->
```
filter(|_| if is_wildcard {
    !is_empty
} else {
    true
})
```
->
```
filter(|_| if !is_wildcard {
    true
} else {
    !is_empty
})
```
->
```
filter(|_| !is_wildcard || !is_empty)
```
->
```
for_each(|_| {
    if !is_wildcard || !is_empty { continue }
    delete()
})
```
->
```
for_each(|_| {
    if !is_wildcard  { continue }
    if !is_empty { continue }
    delete()
})
```

ChayimFriedman2: That's not the same because you also do the `last_arm` assignment inside the `if`, and if I understand correctly it should not execute for empty exprs or wildcards.

A4-Tacks: If the arm is wildcard **and** body is empty, it will be deleted, otherwise it will update `last_arm`

ChayimFriedman2: So what `last_arm` shouldn't include? Should it include empty exprs? Wildcards? Wildcards with empty body?

A4-Tacks: `last_arm` is the last arm that has not been deleted, as cannot be inserted on the deleted node

PR #21745, opened 2026-03-04
fix: Fix a bug in associated type lowering

ChayimFriedman2:
When rewriting assoc type shorthand lowering (`TypeParam::AssocType`), I took into account that if two traits have the same assoc type, even if one is a supertrait of the other, it's an error. However it turns out that assoc type predicates (`Trait<AssocType = Foo>`) use the same infrastructure, and there it's allowed to specify an ambiguous assoc type when it refers to a sub-trait. So make a distinction between ambiguous assoc type that comes from a supertrait and one that does not, and if it comes from a supertrait, keep the resolved assoc type for cases that need it.

Fixes rust-lang/rust-analyzer#21742.

### DIFF
--- a/crates/hir-ty/src/lower.rs
+++ b/crates/hir-ty/src/lower.rs
@@ -1680,10 +1680,16 @@ impl SupertraitsInfo {
     }
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-enum TypeParamAssocTypeShorthandError {
-    AssocTypeNotFound,
-    AmbiguousAssocType,
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+enum AssocTypeShorthandResolution {
+    Resolved(StoredEarlyBinder<(TypeAliasId, StoredGenericArgs)>),
+    Ambiguous {
+        /// If one resolution belongs to a sub-trait and one to a supertrait, this contains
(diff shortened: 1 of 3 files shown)


crates/hir-ty/src/lower.rs
```diff
@@ -1782,34 +1789,53 @@ fn resolve_type_param_assoc_type_shorthand(
...
+                    AssocTypeShorthandResolution::Resolved(_) => {
+                        AssocTypeShorthandResolution::Resolved(current_result)
+                    }
+                    AssocTypeShorthandResolution::Ambiguous { .. } => {
+                        AssocTypeShorthandResolution::Ambiguous {
+                            sub_trait_resolution: Some(current_result),
+                        }
+                    }
+                    AssocTypeShorthandResolution::NotFound
+                    | AssocTypeShorthandResolution::Cycle => unreachable!(),
```
lnicola, line 1830: Should this be a `never!()`?

lnicola:
Could we include the unresolved associated types in the CI metrics?

crates/hir-ty/src/lower.rs
ChayimFriedman2: Nah, it's handled just above.

ChayimFriedman2:
> Could we include the unresolved associated types in the CI metrics?

Not easily (if you mean unresolved in ty lowering), because they just end up as error types.

PR #21750, opened 2026-03-04
fix: Do not use PostAnalysis TypingMode for IDE method resolution

ChayimFriedman2:
As explained in the comments, PostAnalysis is good for most IDE things but not method resolution.

This fixes a bug which should not be impacted by this at all - return position impl trait in trait. It is currently lowered to an opaque, while it should be lowered to an anonymous associated type. But today when it is lowered as an opaque, this opaque of course has no definition and will normalize to an error, preventing method resolution on it from succeeding in some cases.

Fixes rust-lang/rust-analyzer#21749.

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -5952,7 +5952,16 @@ impl<'db> Type<'db> {
     ) -> R {
         let module = resolver.module();
         let interner = DbInterner::new_with(db, module.krate(db));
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
Sorry for the Copilot review, I was curious to see how much it will understand from the jargon. Looks like it understand nothing and just repeats what I said, as expected 😂

PR #21764, opened 2026-03-06
smol_str: Improve API surface

Veykril:
in anticipation of a 1.0 cut

Some of this was done with Claude

### DIFF
--- a/lib/smol_str/src/lib.rs
+++ b/lib/smol_str/src/lib.rs
@@ -34,13 +34,17 @@ use core::{
 pub struct SmolStr(Repr);
 
 impl SmolStr {
+    /// The maximum byte length of a string that can be stored inline
+    /// without heap allocation.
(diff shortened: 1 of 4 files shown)


lib/smol_str/src/lib.rs
```diff
@@ -850,10 +948,18 @@ macro_rules! format_smolstr {
 /// A builder that can be used to efficiently build a [`SmolStr`].
 ///
 /// This won't allocate if the final string fits into the inline buffer.
-#[derive(Clone, Default, Debug, PartialEq, Eq)]
+#[derive(Clone, Default, Debug)]
```
A4-Tacks, line 951: It looks like this is breaking change.
If we do this, the next bump version will need to be 0.4.0 instead of 0.3.6

Veykril: Yea that's intentional, we are thinking of cutting a 1.0 hence we can break things now

PR #21767, opened 2026-03-06
fix: Remove crate from salsa `EditionedFileId`

Veykril:
The way we've done things here breaks invariants of salsa

Fixes https://github.com/rust-lang/rust-analyzer/issues/21721


ChayimFriedman2:
@Veykril did you measure the memory impact?

PR #21768, opened 2026-03-06
fix: add ident_pat qualifier to fully fn param

A4-Tacks:
- Do not show commas on label

Example
---
```rust
fn f(foo: (), bar: u32) {}
fn g(foo: (), mut ba$0)
```

**Before this PR**

```rust
fn f(foo: (), bar: u32) {}
fn g(foo: (), bar: u32)
```

**After this PR**

```rust
fn f(foo: (), bar: u32) {}
fn g(foo: (), mut bar: u32)
```

### DIFF
--- a/crates/ide-completion/src/completions/fn_param.rs
+++ b/crates/ide-completion/src/completions/fn_param.rs
@@ -30,14 +30,27 @@ pub(crate) fn complete_fn_param(
         _ => return None,
     };
 
+    let qualifier = param_qualifier(param);
     let comma_wrapper = comma_wrapper(ctx);
     let mut add_new_item_to_acc = |label: &str| {
-        let mk_item = |label: &str, range: TextRange| {
-            CompletionItem::new(CompletionItemKind::Binding, range, label, ctx.edition)
+        let label = label.strip_prefix(qualifier.as_str()).unwrap_or(label);
+        let insert = if label.starts_with('#') {
+            // FIXME: `#[attr] it: i32` -> `#[attr] mut it: i32`
+            label.to_smolstr()
+        } else {
+            format_smolstr!("{qualifier}{label}")
+        };
+        let mk_item = |insert_text: &str, range: TextRange| {
+            let mut item =
+                CompletionItem::new(CompletionItemKind::Binding, range, label, ctx.edition);
+            if insert_text != label {
+                item.insert_text(insert_text);
+            }
+            item
         };
         let item = match &comma_wrapper {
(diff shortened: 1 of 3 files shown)


crates/ide-completion/src/completions/fn_param.rs
```diff
@@ -30,14 +30,27 @@ pub(crate) fn complete_fn_param(
...
     let comma_wrapper = comma_wrapper(ctx);
     let mut add_new_item_to_acc = |label: &str| {
-        let mk_item = |label: &str, range: TextRange| {
-            CompletionItem::new(CompletionItemKind::Binding, range, label, ctx.edition)
+        let label = label.strip_prefix(qualifier.as_str()).unwrap_or(label);
+        let insert = if label.starts_with('#') {
+            // FIXME: `#[attr] it: i32` -> `#[attr] mut it: i32`
+            label.to_smolstr()
+        } else {
+            format_smolstr!("{qualifier}{label}")
```
ChayimFriedman2, line 41: Why strip then re-add?

A4-Tacks: | original     | current  | actual       | expect           |
| ---          | ---      | ---          | ---              |
| `x: i32`     | `mut $0` | `x: i32`     | `mut x: i32`     |
| `mut x: i32` | `ref $0` | `mut x: i32` | `ref mut x: i32` |

ChayimFriedman2: What does this table mean? What is `original`, `current`, `actual` and `expect`?

A4-Tacks: - original: other fully fn define in ancestors, like `fn foo(x: i32)`
- current: current complete parameter
- actual: current behavior (before this PR)
- expect: expect behavior (after this PR)

ChayimFriedman2: I'm not saying the new behavior isn't good. It is. But you `strip_prefix(qualifier)` then immediately add it at the beginning, this is redundant.

A4-Tacks: - if `strip_prefix(qualifier)` returns Some, is redundant
- if `strip_prefix(qualifier)` returns None, is not redundant

If using other solutions seems more lengthy

ChayimFriedman2: Now I see that there is a difference in case of `#`. Nevermind.

ChayimFriedman2 (approved):

PR #21774, opened 2026-03-07
Migrate unqualify_method_call assist to SyntaxEditor

akashchakrabortymsc-cmd:
part of https://github.com/rust-lang/rust-analyzer/issues/18285

Migrates the `unqualify_method_call` assist handler from the old `ted::`/`SourceChangeBuilder` text-edit API to the new `SyntaxEditor`/`SyntaxFactory` API.

### Changes

- Replace text-range based edits (`edit.delete`, `edit.insert`, `edit.replace`) with node-based edits using `SyntaxEditor`
- Replace `builder.make_mut()` with `builder.make_editor()`
- Replace `insert_use` with `insert_use_with_editor`
- Handle receiver parenthesization using `make.expr_paren()` and `ExprPrecedence::Postfix` instead of manual text-range insertion

### Testing

All 9 existing tests pass.

### DIFF
--- a/crates/ide-assists/src/handlers/unqualify_method_call.rs
+++ b/crates/ide-assists/src/handlers/unqualify_method_call.rs
@@ -1,8 +1,5 @@
 use hir::AsAssocItem;
-use syntax::{
-    TextRange,
-    ast::{self, AstNode, HasArgList, prec::ExprPrecedence},
-};
+use syntax::ast::{self, AstNode, HasArgList, prec::ExprPrecedence, syntax_factory::SyntaxFactory};
 
 use crate::{AssistContext, AssistId, Assists};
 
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/unqualify_method_call.rs
```diff
@@ -51,51 +45,42 @@ pub(crate) fn unqualify_method_call(acc: &mut Assists, ctx: &AssistContext<'_>)
...
-            if let Some((open, close)) = parens {
-                edit.insert(open, "(");
-                edit.insert(close, ")");
-            }
-            edit.replace(replace_comma, format!(".{method_name}("));
+        |builder| {
+            let make = SyntaxFactory::with_mappings();
+            let mut editor = builder.make_editor(call.syntax());
+
+            let new_arg_list = make.arg_list(args.args().skip(1));
```
A4-Tacks, line 56: This will lose the trivia between args, such as a multi line ArgList, but it is also imperfect before migration, it is okay

PR #21779, opened 2026-03-07
fix: replace TODO placeholders in next-solver IrPrint with proper formatting

Albab-Hasan:
seven `IrPrint::print_debug` implementations in the next-solver were placeholder stubs that returned "TODO: <typename>" instead of meaningful output. these are called via the Display impl (and Debug for PatternKind) of these types so any solver trace log containing them was unreadable.

implemented proper formatting for each:
- TraitPredicate      -> "T: Trait" / "!T: Trait"
- HostEffectPredicate -> "const T: Trait" / "[const] T: Trait"
- NormalizesTo        -> "AliasTerm(...) -> Term"
- SubtypePredicate    -> "A <: B"
- CoercePredicate     -> "A -> B"
- FnSig               -> "fn([inputs]) -> output"
- PatternKind         -> "start..=end" / "or([...])" / "!null"

also removed the unused `type_name_of_val` import.

### DIFF
--- a/crates/hir-ty/src/next_solver/ir_print.rs
+++ b/crates/hir-ty/src/next_solver/ir_print.rs
@@ -188,7 +188,7 @@ impl<'db> IrPrint<ty::NormalizesTo<Self>> for DbInterner<'db> {
         t: &ty::NormalizesTo<Self>,
         fmt: &mut std::fmt::Formatter<'_>,
     ) -> std::fmt::Result {
-        write!(fmt, "{} -> {:?}", t.alias, t.term)
+        write!(fmt, "NormalizesTo({} -> {:?})", t.alias, t.term)
     }
 }
 impl<'db> IrPrint<ty::SubtypePredicate<Self>> for DbInterner<'db> {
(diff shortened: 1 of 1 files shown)


crates/hir-ty/src/next_solver/ir_print.rs
```diff
@@ -210,7 +215,7 @@ impl<'db> IrPrint<ty::CoercePredicate<Self>> for DbInterner<'db> {
         t: &ty::CoercePredicate<Self>,
         fmt: &mut std::fmt::Formatter<'_>,
     ) -> std::fmt::Result {
-        fmt.write_str(&format!("TODO: {:?}", type_name_of_val(t)))
```
ChayimFriedman2, line 213: This prints the same as `NormalizesTo`, so I think it's better to qualify both, e.g. something like `CoercePredicate({:?} -> {:?})`.

Albab-Hasan:
@ChayimFriedman2 done

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21779/a1b86d600f88be98643e5dd61d6ed26eda17c09e..b47ef98446b029f38f5342f30c264117718923af)*

PR #21781, opened 2026-03-07
Migrate two assists to clone_subtree to ease SyntaxEditor transition

false200:
## Summary
- `convert_bool_then` now clones its condition with `clone_subtree()` before building the `bool::then` call, since that node isn’t edited again in this assist.
- `convert_bool_to_enum` now finishes the generated `Bool` enum with `clone_subtree()` because it’s only inserted and never mutated afterward.

These are small, safe steps toward the “Migrating rust-analyzer assists to SyntaxEditor” project  
(tracking issue: https://github.com/rust-lang/rust-analyzer/issues/18285;  
rationale: https://github.com/rust-lang/rust-analyzer/issues/15710;  
prototype: https://github.com/ChayimFriedman2/rowan/tree/next-rowan).

## Testing
- `cargo test --package ide-assists -- handlers::convert_bool_then::tests`
- `cargo test --package ide-assists -- handlers::convert_bool_to_enum::tests`
- `cargo test -p ide-assists --lib`

### DIFF
--- a/crates/ide-assists/src/handlers/convert_bool_then.rs
+++ b/crates/ide-assists/src/handlers/convert_bool_then.rs
@@ -102,11 +102,7 @@ pub(crate) fn convert_if_to_bool_then(acc: &mut Assists, ctx: &AssistContext<'_>
                 ast::Expr::BlockExpr(block) => unwrap_trivial_block(block),
                 e => e,
             };
-            let cond = if invert_cond {
-                invert_boolean_expression(&make, cond)
-            } else {
-                cond.clone_for_update()
-            };
+            let cond = if invert_cond { invert_boolean_expression(&make, cond) } else { cond };
 
             let parenthesize = matches!(
                 cond,


ChayimFriedman2:
Are the `clone_subtree()` needed?

false200:
> Are the `clone_subtree()` needed?

Yes, in both places the node is only read or inserted and never mutated afterward, so clone_subtree() is sufficient here. clone_for_update() would only be needed if the node were later modified (e.g., passed through SyntaxEditor for edits), which doesn’t happen in these assists.
The change also passes all local tests (cargo test -p ide-assists --lib).

image

ChayimFriedman2:
I'm not asking if it's sufficient, I'm asking if we need it or can remove it as well.

false200:
> I'm not asking if it's sufficient, I'm asking if we need it or can remove it as well.



> I'm not asking if it's sufficient, I'm asking if we need it or can remove it as well.

Do you mean whether the clone is needed at all, or if we could remove the cloning entirely in those places?

ChayimFriedman2:
Yes.

false200:
> Yes.

You're right, the clone_subtree() calls aren’t actually required here.

false200:
> Yes.

I’ve removed them and updated the PR.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21781/a1b86d600f88be98643e5dd61d6ed26eda17c09e..5c47b6b28ac62b26ccf3b7ccdc460c7b7d1a69bb)*

PR #21784, opened 2026-03-08
fix: remove angle brackets if all lifetime args removed in inline type alias code assist.

80avin:
fixes https://github.com/rust-lang/rust-analyzer/issues/21769

In `Inline type alias into all uses` code assist,
* Remove comma and whitespace along with lifetime arg
* cleanup empty angle brackets

### DIFF
--- a/crates/ide-assists/src/handlers/inline_type_alias.rs
+++ b/crates/ide-assists/src/handlers/inline_type_alias.rs
@@ -12,7 +12,7 @@ use itertools::Itertools;
 use syntax::ast::syntax_factory::SyntaxFactory;
 use syntax::syntax_editor::SyntaxEditor;
 use syntax::{
-    AstNode, NodeOrToken, SyntaxNode,
+    AstNode, NodeOrToken, SyntaxKind, SyntaxNode, T,
     ast::{self, HasGenericParams, HasName},
 };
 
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/inline_type_alias.rs
```diff
@@ -322,12 +322,27 @@ fn create_replacement(
...
+                    }
                     removals.push(NodeOrToken::Node(syntax.clone()));
-
-                    if let Some(ws) = syntax.next_sibling_or_token() {
-                        removals.push(ws.clone());
+                    if let Some(ws) = syntax.next_sibling_or_token()
+                        && ws.kind() == SyntaxKind::WHITESPACE
+                    {
+                        removals.push(ws);
                     }
```
A4-Tacks, line 360: When deleting multiple parameters but the list is not cleared, the comma has not been deleted yet
Similar code (needs `removals.unique()`):
```rust
removals.extend(next_prev().find_map(|dir| {
    let tokens = syntax.siblings_with_tokens(dir)
        .take_while(|it| it.is_token())
        .take_while_inclusive(|it| it.kind() != T![,])
        .collect::<Vec<_>>();
    tokens.iter().any(|it| it.kind() == T![,]).then_some(tokens)
}).flatten())
```

crates/ide-assists/src/handlers/inline_type_alias.rs
```diff
@@ -322,12 +322,27 @@ fn create_replacement(
...
+                    if let Some(lifetime_arg) =
+                        old_lifetime.syntax().parent().and_then(ast::LifetimeArg::cast)
+                    {
+                        // If the LifetimeArg is a child of GenericArgList, remove the LifetimeArg
+                        // wrapper
+                        if lifetime_arg
+                            .syntax()
+                            .parent()
+                            .and_then(ast::GenericArgList::cast)
+                            .is_some()
```
A4-Tacks: Is this condition always true?

80avin: Didn't know that LifetimeArg is always a child of GenericArgList.
Fixed. Please resolve if it appears right.

crates/ide-assists/src/handlers/inline_type_alias.rs
80avin: Fixed. Also added one more test case.
Please resolve if it seems right.

A4-Tacks:
Strange, why did CI fail

@ChayimFriedman2 Do you know why there is a null pointer here?

https://github.com/rust-lang/rust-analyzer/actions/runs/23118102241/job/67147080801

ChayimFriedman2:
See [#t-compiler/rust-analyzer > UB @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/UB/near/579649352).

PR #21788, opened 2026-03-09
Do not re-query source roots per crate in analysis-stats

Veykril:
Not ideal as we may pick a crate that analysis is not actually using but we do not encode targets in the crate graph so we can't really tell what main library crate will be


ChayimFriedman2:
This could still report more memory than actually used because we populate item trees we don't need.

The problem is that accurate crate discovery requires def map, and we compute it after...

PR #21792, opened 2026-03-10
make matching brace work when cursor not at bracket

bend-n:
by jumping to the parents enclosing bracket.

see [#t-compiler/rust-analyzer > experimental/matching braces when not on a brace @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/experimental.2Fmatching.20braces.20when.20not.20on.20a.20brace/near/566138066)

this would make the behavior match vscodes go to bracket command.

behavior pictured:

https://github.com/user-attachments/assets/ca6bd28f-48a5-414d-a750-45e0f128b361

### DIFF
--- a/crates/ide/src/matching_brace.rs
+++ b/crates/ide/src/matching_brace.rs
@@ -17,10 +17,9 @@ use syntax::{
 pub(crate) fn matching_brace(file: &SourceFile, offset: TextSize) -> Option<TextSize> {
     const BRACES: &[SyntaxKind] =
         &[T!['{'], T!['}'], T!['['], T![']'], T!['('], T![')'], T![<], T![>], T![|], T![|]];
-
-    if let Some((brace_token, brace_idx)) = file
-        .syntax()
-        .token_at_offset(offset)
+    let current = file.syntax().token_at_offset(offset);
+    if let Some((brace_token, brace_idx)) = current
+        .clone()
         .filter_map(|node| {
             let idx = BRACES.iter().position(|&brace| brace == node.kind())?;
             Some((node, idx))
@@ -39,10 +38,8 @@ pub(crate) fn matching_brace(file: &SourceFile, offset: TextSize) -> Option<Text
             .find(|node| node.kind() == matching_kind && node != &brace_token)?;
         Some(matching_node.text_range().start())
     } else {
-        // when the offset is not at a brace
-        let thingy = file.syntax().token_at_offset(offset).last()?;
-        // find first parent
-        thingy.parent_ancestors().find_map(|x| {
+        // when the offset is not at a brace, find first parent
+        current.last()?.parent_ancestors().find_map(|x| {
             x.children_with_tokens()
                 .filter_map(|it| it.into_token())
                 // with ending brace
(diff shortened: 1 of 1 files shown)


crates/ide/src/matching_brace.rs
```diff
@@ -17,25 +17,40 @@ use syntax::{
...
+        }
+        let matching_kind = BRACES[brace_idx ^ 1];
+        let matching_node = parent
+            .children_with_tokens()
+            .filter_map(|it| it.into_token())
+            .find(|node| node.kind() == matching_kind && node != &brace_token)?;
+        Some(matching_node.text_range().start())
+    } else {
+        // when the offset is not at a brace
+        let thingy = file.syntax().token_at_offset(offset).last()?;
```
ChayimFriedman2: We already have `file.syntax().token_at_offset(offset)`. There is a potential for code reuse here.

ChayimFriedman2: Also `thingy` is *not exactly* a descriptive name.

crates/ide/src/matching_brace.rs
```diff
@@ -17,25 +17,40 @@ use syntax::{
...
+        // find first parent
+        thingy.parent_ancestors().find_map(|x| {
+            x.children_with_tokens()
+                .filter_map(|it| it.into_token())
+                // with ending brace
+                .filter(|node| BRACES.contains(&node.kind()))
```
ChayimFriedman2, line 46: This goes to either the opening brace or the closing brace, whichever is closer. I don't think this is the intended behavior. Instead, always go the the open brace, e.g. `BRACES.iter().copied().position(|brace| brace == node.kind()).is_some_and(|idx| idx % 2 == 0)`

bend-n: it goes to the opening brace if there is no closing brace, you mean? but it should always go to the closing brace

ChayimFriedman2: I don't care if it goes to the closing brace or the opening brace (IMO the opening one is more natural), but it should be consistent. Currently it will jump to the closest brace, whether opening or closing.

bend-n: i dont know what you mean by closest. closest to cursor? it always goes to the closing brace in my experience?

ChayimFriedman2: Closest to cursor, yes. Counting by depth.

ChayimFriedman2: Oh forgive me, I forgot braces always come in pairs, so there isn't such thing as the closest brace. You can ignore this comment.

crates/ide/src/matching_brace.rs
```diff
@@ -17,25 +17,37 @@ use syntax::{
...
-    let parent = brace_token.parent()?;
-    if brace_token.kind() == T![|] && !ast::ParamList::can_cast(parent.kind()) {
-        cov_mark::hit!(pipes_not_braces);
-        return None;
+        .last()
+    {
+        let parent = brace_token.parent()?;
+        if brace_token.kind() == T![|] && !ast::ParamList::can_cast(parent.kind()) {
+            cov_mark::hit!(pipes_not_braces);
+            return None;
```
ChayimFriedman2, line 32: Why is that? I see no reason to exclude pipes.

bend-n: thats not my code, its from https://github.com/rust-lang/rust-analyzer/pull/5047

PR #21798, opened 2026-03-10
Allow defining inline structs in `diagnostics!`

ada4a:
Just a random TODO comment I thought I'd implement. I have some doubts about this though:
- is this still desired?
- is removing `pub` and `struct` really worth it? I feel like one'd write them almost instinctively and be surprised when it doesn't work. Maybe I could adjust the macro to look for `$($($vis:vis)? strukt:ident)?` instead?
- the macro definition ended up being _somewhat_ cursed
- formatting is completely broken

Also, the diff is pretty hard to review, sorry about that. I found that commenting out `IncorrectCase` helped a _little_ bit, hence the second commit that uncomments it.

### DIFF
--- a/crates/hir/src/diagnostics.rs
+++ b/crates/hir/src/diagnostics.rs
@@ -445,7 +445,7 @@ diagnostics![AnyDiagnostic<'db> ->
 
     ;
 
-    // IncorrectCase,
+    IncorrectCase,
 ];
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]


ChayimFriedman2:
I personally think the old way is fine, and don't want this change.

lnicola:
Not sure about the complexity, but that enum is a bit weird, do we still want it to be an enum in the long run?

ada4a:
> Not sure about the complexity, but that enum is a bit weird, do we still want it to be an enum in the long run?

That I don't feel qualified enough to answer, but the complexity bothers me as well. I'll close this and open an alternative PR removing the TODO comment itself -- if someone ends up wanting this new and enhanced macro after all, they could just copy this PR

PR #21801, opened 2026-03-11
feat: supports minus to underscore rename

A4-Tacks:
Typing minus is usually much more convenient than typing underscore

Example
---
Rename to `"a-b"`
```rust
fn main() { let i$0 = 1; }
```

**Before this PR**

Invalid name `a-b`: not an identifier

**After this PR**

```rust
fn main() { let a_b = 1; }
```

### DIFF
--- a/crates/ide-db/src/rename.rs
+++ b/crates/ide-db/src/rename.rs
@@ -701,6 +701,7 @@ pub enum IdentifierKind {
 
 impl IdentifierKind {
     pub fn classify(edition: Edition, new_name: &str) -> Result<(Name, IdentifierKind)> {
+        let new_name = &new_name.replace('-', "_");
         match parser::LexedStr::single_token(edition, new_name) {
             Some(res) => match res {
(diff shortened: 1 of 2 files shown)


lnicola (commented):
The change looks fine to me, but that's not a typo I've ever made and I'm not aware of other LSP servers that do this. The Python one in VS Code simply renames to `a-b`.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21801/8ed3cca4d30610fd0d3c1179c85418de2dc0a7c1..0b217b2296d56a8383b18542bd1045eec45db4c4)*

A4-Tacks:
> I'm not aware of other LSP servers that do this. The Python one in VS Code simply renames to `a-b`.

This feature is very useful, but there seems to be no other LSP or IDE implementation

PyCharm, jedils, tsserver, luals, ocamllsp and clangd have not implemented this

ChayimFriedman2:
I'm also not a fan of this feature.

PR #21803, opened 2026-03-12
ide-assists: replace unnecessary clone_for_update with clone_subtree…

Amit5601:
The condition expression in `convert_bool_then` is not mutated after cloning. Replacing `clone_for_update()` with `clone_subtree()` avoids creating an unnecessary mutable syntax node. All ide-assists tests pass locally.


A4-Tacks:
Duplicated #21781

PR #21809, opened 2026-03-13
fix: offer on const like path-expr for 'extract_variable'

A4-Tacks:
Fixes rust-lang/rust-analyzer#15820

Example
---
```rust
struct Foo;
fn foo() -> Foo {
    $0Foo$0
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
struct Foo;
fn foo() -> Foo {
    let foo = Foo;
    foo
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -75,7 +75,7 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_>) -> Op
             .next()
             .and_then(ast::Expr::cast)
         {
-            expr.syntax().ancestors().find_map(valid_target_expr)?.syntax().clone()
+            expr.syntax().ancestors().find_map(valid_target_expr(ctx))?.syntax().clone()
         } else {
             return None;
         }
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (changes requested):
The test passes without your changes, so there is something wrong in your checks. Also this is expected - we return `true` for any expr.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21809/6254616e97f358e67b70dfc0463687f5f7911c1a..b7c4100a27734fe7140df39012bd315ad9e911ee)*

A4-Tacks:
> we return `true` for any expr.

The old behavior is to return `false` for any `PATH_EXPR`

ChayimFriedman2:
Oh, right. So why not just remove that?

A4-Tacks:
What is 'remove that'?

ChayimFriedman2:
Just make `PATH_EXPR` always return true, like the rest of exprs.

A4-Tacks:
But it's strange to extract a local variable as a local variable

Especially when extracting on empty selection, it is usually noisy

PR #21817, opened 2026-03-14
fix: Fix overlap edit on record to tuple assist uses self

A4-Tacks:
Fixes rust-lang/rust-analyzer#21595

Due to macro processing, syntax mapping was not possible, so the PR changed it to incremental edit instead of full node replacement

- **Migrate 'convert_named_struct_to_tuple_struct' assist**
- **Fix field ref in macro for 'convert_tuple_struct_to_named_struct'**
- **Fix overlap edit on tuple to record assist expr uses self**

Example
---
```rust
struct $0Foo { field1: u32 }
impl Foo {
    fn clone(&self) -> Self {
        Self { field1: self.field1 }
    }
}
```

**Before this PR**

Panic

**After this PR**

```rust
struct Foo(u32);
impl Foo {
    fn clone(&self) -> Self {
        Self(self.0)
    }
}
```

---

```rust
struct T$0(u8);
fn test(t: T) {
    T(t.0);
}
```

**Before this PR**

Panic

**After this PR**

```rust
struct T { field1: u8 }
fn test(t: T) {
    T { field1: t.field1 };
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/convert_named_struct_to_tuple_struct.rs
+++ b/crates/ide-assists/src/handlers/convert_named_struct_to_tuple_struct.rs
@@ -1,9 +1,7 @@
-use std::ops::RangeInclusive;
-
 use either::Either;
 use ide_db::{defs::Definition, search::FileReference};
 use syntax::{
-    NodeOrToken, SyntaxElement, SyntaxKind, SyntaxNode, T, TextRange,
+    NodeOrToken, SyntaxKind, SyntaxNode, T,
     algo::next_non_trivia_token,
     ast::{
         self, AstNode, HasAttrs, HasGenericParams, HasVisibility, syntax_factory::SyntaxFactory,
@@ -12,7 +10,9 @@ use syntax::{
     syntax_editor::{Element, Position, SyntaxEditor},
 };
 
-use crate::{AssistContext, AssistId, Assists, assist_context::SourceChangeBuilder};
+use crate::{
+    AssistContext, AssistId, Assists, assist_context::SourceChangeBuilder, utils::cover_edit_range,
+};
 
 // Assist: convert_named_struct_to_tuple_struct
 //
@@ -242,7 +242,7 @@ where
 {
     let make = SyntaxFactory::without_mappings();
     let orig = ctx.sema.original_range_opt(field_list.syntax())?;
-    let list_range = cover_range(source, orig.range);
+    let list_range = cover_edit_range(source, orig.range);
 
     let l_curly = match list_range.start() {
         NodeOrToken::Node(node) => node.first_token()?,
@@ -265,7 +265,7 @@ where
 
     for name_ref in fields(&field_list) {
         let Some(orig) = ctx.sema.original_range_opt(name_ref.syntax()) else { continue };
-        let name_range = cover_range(source, orig.range);
+        let name_range = cover_edit_range(source, orig.range);
 
         if let Some(colon) = next_non_trivia_token(name_range.end().clone())
             && colon.kind() == T![:]
@@ -306,7 +306,7 @@ fn edit_field_references(
                     // Only edit the field reference if it's part of a `.field` access
                     if name_ref.syntax().parent().and_then(ast::FieldExpr::cast).is_some() {
                         edit.replace_all(
-                            cover_range(&source, r.range),
+                            cover_edit_range(&source, r.range),
                             vec![make.name_ref(&index.to_string()).syntax().clone().into()],
                         );
                     }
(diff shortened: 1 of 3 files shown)


crates/ide-assists/src/handlers/convert_tuple_struct_to_named_struct.rs
```diff
@@ -147,93 +151,121 @@ fn edit_struct_references(
...
+                process_delimiter(ctx, source, editor, &arg_list, first_insert);
+            },
+            _ => {}
         }
-        edit.add_file_edits(file_id.file_id(ctx.db()), editor);
+    }
+    Some(())
+}
+
+fn process_delimiter(
```
bit-aloo, line 231: We can pass syntax_factory to this method, instead of instantiating new one.

crates/ide-assists/src/handlers/convert_tuple_struct_to_named_struct.rs
```diff
@@ -147,93 +151,121 @@ fn edit_struct_references(
...
-                            if let (Some(start), Some(end)) = (start, end) {
-                                let replace_range = RangeInclusive::new(start, end);
-                                editor.replace_all(replace_range, vec![new_node.into()]);
-                            }
-                        }
+        edit.add_file_edits(file_id.file_id(ctx.db()), editor);
+    }
+}
+
+fn process_struct_name_reference(
```
bit-aloo, line 166: We can pass syntax_factory to this method, instead of instantiating  a new one.

crates/ide-assists/src/handlers/convert_named_struct_to_tuple_struct.rs
```diff
@@ -243,19 +298,46 @@ fn edit_field_references(
...
+    let Some(whitespace) = whitespace else { return };
+    let NodeOrToken::Token(token) = whitespace.syntax_element() else { return };
+
+    if token.kind() == SyntaxKind::WHITESPACE && !token.text().contains('\n') {
+        edit.delete(token);
+    }
+}
+
+fn remove_trailing_comma(w: ast::WhereClause) -> SyntaxNode {
+    let w = w.syntax().clone_subtree();
```
bit-aloo, line 331: I am not entirely sure about using clone_subtree() here. Since the goal of the migration is to move the deep-copy logic into SyntaxEditor, or at least avoid having it directly inside the handlers, I was wondering if we could instead reuse the editor instance created earlier through the builder and perform the operations through it. That might allow us to batch the mutations and keep the logic more centralized.

crates/ide-assists/src/handlers/convert_named_struct_to_tuple_struct.rs
```diff
@@ -230,11 +229,67 @@ fn process_struct_name_reference(
     Some(())
 }
 
+fn record_to_tuple_struct_like<T, I>(
+    ctx: &AssistContext<'_>,
```
bit-aloo, line 233: syntax_factory can be passed as a parameter

crates/ide-assists/src/handlers/convert_tuple_struct_to_named_struct.rs
A4-Tacks: Passing `SyntaxFactory::without_mappings()` doesn't make sense, does it

crates/ide-assists/src/handlers/convert_named_struct_to_tuple_struct.rs
A4-Tacks: Due to handling macros, I am unable to utilize syntax mappings, which is very painful

crates/ide-assists/src/handlers/convert_tuple_struct_to_named_struct.rs
bit-aloo: That’s a fair point, particularly since, in the without_mapping case, it’s effectively just a thin wrapper around make. My only concern is that it might lead to some code fragmentation, as we would end up redefining the same wrapper in multiple methods where it’s needed.

crates/ide-assists/src/handlers/convert_named_struct_to_tuple_struct.rs
bit-aloo: Can you please explain a bit more about the pain points?

A4-Tacks: Due to macro processing, `replace_all` is actually used, but this does not support mapping,
and `SyntaxNode -> FileRange -> RangeInclusive<SyntaxElement>` also does not support mapping

bit-aloo: Ah!!! ya make sense...  Thanks for being explicit.

Shourya742 (approved):
wont block, this on other comment. LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21817/6254616e97f358e67b70dfc0463687f5f7911c1a..34839f600f550855177cee66b758fbc306bd13b3)*

PR #21822, opened 2026-03-15
fix: fill match arms on last comma and empty expr

A4-Tacks:
For Vim, triggering code action at the end of line usually occurs before the last character rather than after it

Example
---
```rust
fn foo() {
    match Some(2) {
        Some(_) => ()$0,
    }
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
fn foo() {
    match Some(2) {
        Some(_) => (),
        None => ${1:todo!()},$0
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/add_missing_match_arms.rs
+++ b/crates/ide-assists/src/handlers/add_missing_match_arms.rs
@@ -245,14 +245,7 @@ pub(crate) fn add_missing_match_arms(acc: &mut Assists, ctx: &AssistContext<'_>)
                 .arms()
                 .filter(|arm| {
                     if matches!(arm.pat(), Some(ast::Pat::WildcardPat(_))) {
-                        let is_empty_expr = arm.expr().is_none_or(|e| match e {
-                            ast::Expr::BlockExpr(b) => {
-                                b.statements().next().is_none() && b.tail_expr().is_none()
-                            }
-                            ast::Expr::TupleExpr(t) => t.fields().next().is_none(),
-                            _ => false,
-                        });
-                        if is_empty_expr {
+                        if arm.expr().is_none_or(is_empty_expr) {
                             false
                         } else {
                             cov_mark::hit!(add_missing_match_arms_empty_expr);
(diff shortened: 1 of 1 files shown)


A4-Tacks:
> Do I understand correctly that this PR degrades experience for non-Vim users,

No, no, this makes it more convenient for both Vim and non-Vim users

> by making it so cursor after the end of the line won't trigger the assist?

In fact, this PR can be triggered even without end of line

Use `$0...$0` to indicate trigger range:

| old                | new                |
| ---                | ---                |
| `{ a => 2,$0 }$0`  | `{ a => 2$0, }$0`  |
| `{ a => (),$0 }$0` | `{ a => $0(), }$0` |
| `{ a => {}$0 }$0`  | `{ a => $0{} }$0`  |

ChayimFriedman2:
My question was whether this:
```rust
fn foo() {
    match Some(2) {
        Some(_) => (),$0
    }
}
```
Work now.

A4-Tacks:
But in Vim class editors, this is not very common. Vim class editors are usually:

```rust
fn foo() {
    match Some(2) {
        Some(_) => ()$0,
    }
}
```

ChayimFriedman2:
But in non-Vim editors it is.

A4-Tacks:
At least it hasn't gotten worse

ChayimFriedman2:
That's the problem, it did. This works currently.

A4-Tacks:
Isn't triggering on the trailing comma and empty arm a reasonable improvement?

ChayimFriedman2:
Sure it is, but not when it comes at the expense of that. Why shouldn't we have both?

A4-Tacks:
> but not when it comes at the expense of that

I don't understand, what has been expense

ChayimFriedman2:
Without this change a cursor after the comma is supported, but not before. With it only cursors before the comma is supported. I want to support both.

A4-Tacks:
> With it only cursors before the comma is supported.

There's no such thing, this doesn't disrupt old behavior

> https://github.com/rust-lang/rust-analyzer/pull/21822#issuecomment-4096907046

The tail trigger is a range, and this PR only expands the range rather than interrupting it

ChayimFriedman2 (approved):
Ah. Ok. Then we didn't understood each other.

Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21822/90c8906e6443e7cee18cece9c2621a8b1c10794c..aef7b79af90fd490635e728916b4576ec223ac42)*

PR #21823, opened 2026-03-15
Implement signature type inference

Veykril:


Veykril:
Have to hook this up in analysis_stats still and add some IDE level tests (assuming none cover this yet) but otherwise this should be good to go

Veykril:
This also does not yet hook up inference in variant fields, thats a separate follow up

Veykril:
Oh, and type expectations are missing for generic const argument position

crates/hir-ty/src/next_solver/def_id.rs
```diff
@@ -26,13 +26,15 @@ pub enum SolverDefId {
...
+    AnonConstId(AnonConstId),
     TraitId(TraitId),
     TypeAliasId(TypeAliasId),
     InternedClosureId(InternedClosureId),
     InternedCoroutineId(InternedCoroutineId),
     InternedOpaqueTyId(InternedOpaqueTyId),
     EnumVariantId(EnumVariantId),
     // FIXME(next-solver): Do we need the separation of `Ctor`? It duplicates some variants.
+    // FIXME: Thuis is wrong, salsa::Supertype folds these into the previous variants.
     Ctor(Ctor),
```
Veykril, line 36: Just noticed this, this is a bug, if we move these IDs through salsa they will get folded into AdtId::StructId/EnumVariantId. This might be causing issues as Ctor has a different rust type associated with it (value-ty) compared to AdtId::StructId

ChayimFriedman2: True, although we can remove the `salsa::Supertype` on `SolverDefId`, we aren't using it at all.

Veykril: ah, well then there is no bug either nice

crates/hir-def/src/resolver.rs
```diff
@@ -854,14 +857,15 @@ impl<'db> Resolver<'db> {
     pub fn update_to_inner_scope(
         &mut self,
         db: &'db dyn DefDatabase,
-        owner: DefWithBodyId,
+        owner: impl Into<ExpressionStoreOwner>,
```
ChayimFriedman2, line 860: A bit big method and sad to see it becoming generic. I understand it has a lot of callers so you don't want to take `ExpressionStoreOwner` directly, but maybe polymorphize it?

crates/hir-def/src/expr_store.rs
```diff
@@ -364,6 +387,7 @@ impl ExpressionStoreBuilder {
                     binding_owners,
                     block_scopes: block_scopes.into_boxed_slice(),
                     ident_hygiene,
+                    const_expr_origins: const_expr_origins.unwrap_or_default(),
```
ChayimFriedman2, line 393: `shrink_to_fit()`?

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -1045,17 +1070,30 @@ impl<'db> ExprCollector<'db> {
     }
 
     fn lower_const_arg_opt(&mut self, arg: Option<ast::ConstArg>) -> ConstRef {
-        ConstRef { expr: self.collect_expr_opt(arg.and_then(|it| it.expr())) }
+        let const_expr_origins = self.store.const_expr_origins.take();
```
ChayimFriedman2, line 1073: I think a comment to explain why we take it will be helpful. If I understand correctly, that's because a nested const isn't a root, right? And it will be inferred with its parent.

crates/hir-ty/src/infer.rs
```diff
@@ -179,6 +178,57 @@ fn infer_cycle_result(db: &dyn HirDatabase, _: salsa::Id, _: DefWithBodyId) -> I
...
+        let expected = match origin {
+            // Array lengths are always `usize`.
+            ConstExprOrigin::ArrayLength => Expectation::has_type(ctx.types.types.usize),
+            // Const parameter default: look up the param's declared type.
+            ConstExprOrigin::ConstParam(local_id) => Expectation::has_type(db.const_param_ty_ns(
+                ConstParamId::from_unchecked(TypeOrConstParamId { parent: def, local_id }),
+            )),
+            // Path const generic args: determining the expected type requires
+            // path resolution.
+            // FIXME
```
ChayimFriedman2, line 214: This is problematic indeed.

The way I see it, we will need to get rid of separate `TyLoweringContext` and `InferenceContext`, and instead of having a single place to infer all anon consts, we will need to infer them on demand from the lowered type. I don't know yet whether this will mean a single query for all ty lowering or many queries like now.

How do you expect to solve that?

crates/hir-ty/src/infer.rs
```diff
@@ -179,6 +178,57 @@ fn infer_cycle_result(db: &dyn HirDatabase, _: salsa::Id, _: DefWithBodyId) -> I
...
+            ConstExprOrigin::GenericArgsPath => Expectation::None,
+        };
+        ctx.infer_expr(root_expr, &expected, ExprIsRead::Yes);
+    }
+
+    ctx.type_inference_fallback();
```
ChayimFriedman2: We still need the full set of calls as in `infer_query_with_inspect()`, e.g. `infer_closures()` because there can be (nested) closures. So it makes sense to extract it into a function

crates/hir-ty/src/infer.rs
```diff
@@ -555,9 +605,37 @@ impl InferenceResult {
...
+    /// Look up inference results for a specific anonymous const in a signature.
+    ///
+    /// This delegates to [`Self::for_signature`] on the anon const's owner.
+    /// The returned `InferenceResult` contains types for *all* expressions in
+    /// the owner's signature store, not just this anon const's sub-tree.
+    /// Callers should index into it with `loc.expr` to get the root expression's
+    /// type.
+    // FIXME: This function doesn't make sense in that we can't return a full inference result here
+    // as the anon const is just part of an inference result.
+    pub fn for_anon_const(db: &dyn HirDatabase, id: AnonConstId) -> &InferenceResult {
```
ChayimFriedman2: Why does it exist? It isn't called.

crates/hir-ty/src/next_solver/interner.rs
```diff
@@ -2006,7 +2002,10 @@ impl<'db> Interner for DbInterner<'db> {
         // FIXME: Make this a query? I don't believe this can be accessed from bodies other than
         // the current infer query, except with revealed opaques - is it rare enough to not matter?
         let InternedCoroutine(owner, expr_id) = def_id.0.loc(self.db);
-        let body = self.db.body(owner);
+        let Some(body_owner) = owner.as_def_with_body() else {
```
ChayimFriedman2, line 2005: This looks incorrect, here and below? Signatures may contain coroutines as well.

crates/hir-ty/src/drop.rs
```diff
@@ -132,9 +132,12 @@ fn has_drop_glue_impl<'db>(
         TyKind::Slice(ty) => has_drop_glue_impl(infcx, ty, env, visited),
         TyKind::Closure(closure_id, subst) => {
             let owner = db.lookup_intern_closure(closure_id.0).0;
-            let infer = InferenceResult::for_body(db, owner);
+            let Some(body_owner) = owner.as_def_with_body() else {
```
ChayimFriedman2, line 135: Here too we should handle signatures.

crates/hir-ty/src/layout.rs
```diff
@@ -333,7 +333,10 @@ pub fn layout_of_ty_query(
         }
         TyKind::Closure(id, args) => {
             let def = db.lookup_intern_closure(id.0);
-            let infer = InferenceResult::for_body(db, def.0);
+            let Some(body_owner) = def.0.as_def_with_body() else {
```
ChayimFriedman2, line 336: As well as here.

ChayimFriedman2 (approved):
Excited for this!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21823/4eac290b58a70961e78f2e0c04f61a08b995b2cb..c061236562d4a83023456dac646ee6b2f4dd6d88)*

crates/hir-def/src/expr_store/lower.rs
Veykril: Yea, that was my thinking but thats actually wrong. I am planning on changing that to consider nested const args to also be roots but doing so currently breaks some things so that needs to wait for follow up proper const lowering

crates/hir-ty/src/infer.rs
Veykril: My only thought here so far was to infer all anon const with the rest of the body or signature they are in, then fetching the type of the anon const itself will just be a lookup in the main inference context. How to thread the expectations through I have not really thought about yet.

crates/hir-ty/src/next_solver/interner.rs
Veykril: Ah, yea I forgot to look at the remaining `as_def_with_body` calls

Veykril: Will do this in a follow up as this is a really invasive change actually

PR #21827, opened 2026-03-16
Don't trigger GC on slow tests

ChayimFriedman2:
It can lead to UB.

### DIFF
--- a/crates/rust-analyzer/src/main_loop.rs
+++ b/crates/rust-analyzer/src/main_loop.rs
@@ -303,6 +303,15 @@ impl GlobalState {
         .map(Some)
     }
 
+    fn trigger_garbage_collection(&mut self) {
+        if cfg!(test) {
+            // Slow tests run the main loop in multiple threads, but GC isn't thread safe.
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


Wilfred:
Aha, that explains a lot, thanks :)

A4-Tacks:
It seems that there are still empty pointers after this PR:

https://github.com/rust-lang/rust-analyzer/actions/runs/23485977075

It seems that SIGSEGV frequently appears after this PR, could it be OOM?

https://github.com/rust-lang/rust-analyzer/actions/runs/23439227044

```
process didn't exit successfully: `/home/runner/work/rust-analyzer/rust-analyzer/target/llvm-cov-target/debug/deps/slow_tests-b8257e4d576b48c3` (signal: 11, SIGSEGV: invalid memory reference)
```

ChayimFriedman2:
That is very weird because the panic is in garbage collection in slow tests, but this PR explicitly disables GC in slow tests. Perhaps `cfg(test)`  is not set for some reason?

ChayimFriedman2:
Oh right, I forgot slow tests are integration tests, and for those `cfg(test)` is indeed not set. We'll need another way.

PR #21828, opened 2026-03-16
fix: SCIP generation should prime caches in parallel

Wilfred:
Previously load_workspace_into_db() called parallel_prime_caches() with a worker count of 1. This made SCIP generation slower than it needed to be.

On my M4 Macbook Pro this reduces the time taken to generate SCIP for rust-analyzer from ~39 seconds to ~35 seconds.

AI disclosure: Partially written with Claude Opus 4.6.

### DIFF
--- a/crates/load-cargo/src/lib.rs
+++ b/crates/load-cargo/src/lib.rs
@@ -42,6 +42,7 @@ pub struct LoadCargoConfig {
     pub load_out_dirs_from_check: bool,
     pub with_proc_macro_server: ProcMacroServerChoice,
     pub prefill_caches: bool,
+    pub num_worker_threads: usize,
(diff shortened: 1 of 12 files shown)


ChayimFriedman2:
FWIW I don't like this change. I believe the default behavior should be conservative, and if SCIP wants to control cache priming it should set `config.prefill_caches = false` and prime them itself, likely with a config for the number of threads.

Actually maybe we should get rid of `config.prefill_caches` completely... Though it's a breaking change.

Wilfred:
OK, I've changed it to a CLI argument for the SCIP subcommand. Does this look sensible, or do you want me to push cache priming into the SCIP entry point?

PR #21829, opened 2026-03-16
minor: Small clean-ups

ada4a:
Some things I've noticed while working on https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Type.20inference.20on.20non-speculative.20file.20fails.20--.20due.20to.20.60.2E.2E.60.3F/with/579457494, which I wasn't able to finish, hence this separate small PR.

- extract unrelated `if` out of `matches!`
- in `include_references`:
  - add docs
  - use a while-let chain
  - add a more complex test

### DIFF
--- a/crates/ide-completion/src/completions/postfix.rs
+++ b/crates/ide-completion/src/completions/postfix.rs
@@ -402,6 +402,10 @@ fn receiver_accessor(receiver: &ast::Expr) -> ast::Expr {
         .unwrap_or_else(|| receiver.clone())
 }
 
+/// Given an `initial_element`, tries to expand it to include deref(s), and then references.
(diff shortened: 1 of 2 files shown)


crates/ide-completion/src/completions/postfix.rs
```diff
@@ -1071,6 +1073,7 @@ fn main() {
         postfix_completion_for_block("const");
     }
 
+    #[track_caller]
```
lnicola: Isn't it useful to know which `check_edit` failed?

ada4a: Ah, you're right, `#[track_caller]` removes that information. Not sure what I was thinking here. Removed the change

PR #21831, opened 2026-03-17
fix: don't offer unwrap result/option return type assist in trait impls

Albab-Hasan:
fixes rust-lang/rust-analyzer#13560

the `unwrap_result_return_type` and `unwrap_option_return_type` assists  were incorrectly offered inside trait impl methods. applying the assist removes the return type breaking the trait contract and producing a compile error.

this fix disables both assists when the function is a trait impl method, checked via `implemented_trait`. this approach was chosen over showing the assist with a warning, since silently producing uncompilable code is wrong behavior.

added a test to cover this case.

### DIFF
--- a/crates/ide-assists/src/handlers/unwrap_return_type.rs
+++ b/crates/ide-assists/src/handlers/unwrap_return_type.rs
@@ -1,4 +1,5 @@
 use either::Either;
+use hir::AsAssocItem;
 use ide_db::{
     famous_defs::FamousDefs,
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
Assists aren't required to produce compilable code, and this can be useful even in trait methods (for example, if you want to change the trait signature as well).

Albab-Hasan:
@ChayimFriedman2 youre right. my mistake. shouldve done more research before making the pr. ill close it

PR #21834, opened 2026-03-17
ide-assists: add tests for const, static in add_braces

Amit5601:
The `add_braces` assist already supports const, static and let-else initializers, but there were no tests covering these cases. This PR adds tests for 'const initializers', 'static initializers', 'let-else expressions'. These tests document the current behavior of the assist without changing it.

### DIFF
--- a/crates/ide-assists/src/handlers/add_braces.rs
+++ b/crates/ide-assists/src/handlers/add_braces.rs
@@ -9,7 +9,7 @@ use crate::{AssistContext, AssistId, Assists};
 
 // Assist: add_braces
 //
-// Adds braces to closure bodies, match arm expressions and assignment bodies.
+// Adds braces to closure bodies, match arm expressions and Initializer expression.
 //
 // ```
 // fn foo(n: i32) -> i32 {
@@ -52,7 +52,7 @@ pub(crate) fn add_braces(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<(
         match expr_type {
             ParentType::ClosureExpr => "Add braces to this closure body",
             ParentType::MatchArmExpr => "Add braces to this match arm expression",
-            ParentType::Assignment => "Add braces to this assignment expression",
+            ParentType::Initializer => "Add braces to this Initializer expression",
         },
         expr.syntax().text_range(),
         |builder| {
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/add_braces.rs
```diff
@@ -264,4 +264,54 @@ fn foo() {
...
+    );
+}
+
+    #[test]
+    fn suggest_add_braces_for_let_else_initializer() {
+        check_assist(
+            add_braces,
+            r#"
+fn foo(opt: Option<i32>) {
+    let Some(x) $0= opt else return;
```
A4-Tacks: `else return` is syntax error, and let-else is normal let-stmt, it has been tested, but let-expr has not been tested

Amit5601: @A4-Tacks Thanks! I'll update the test and fix the CI failures.

Amit5601: @A4-Tacks i tried adding a test for let-expr  but it seems the current assist doesn't handle it correctly and produces broken output. as of now, I have removed the test for let-expr. The PR focuses on valid cases (const, static).

A4-Tacks: `if let Some(x: i32) =$0 Some(2) {}` generates valid output

crates/ide-assists/src/handlers/add_braces.rs
```diff
@@ -52,7 +52,7 @@ pub(crate) fn add_braces(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<(
         match expr_type {
             ParentType::ClosureExpr => "Add braces to this closure body",
             ParentType::MatchArmExpr => "Add braces to this match arm expression",
-            ParentType::Assignment => "Add braces to this assignment expression",
+            ParentType::Initializer => "Add braces to this Initializer expression",
```
A4-Tacks, line 55: Why rename? In the future, it can offer on assignment (bin-expr)

crates/ide-assists/src/handlers/add_braces.rs
Amit5601: @A4-Tacks  I’ll take another look at let-expr cases and see how the assist behaves there. For now I focused on const/static since those were clearly handled, but I’ll revisit let-expr more carefully.

Amit5601: @A4-Tacks seems like the coverage job is failing with a segfault in slow_tests, unrelated to the changes in this PR. Let me know if I should investigate further or if it's a flaky CI issue.

ChayimFriedman2: You need to rebase.

PR #21838, opened 2026-03-18
fix: skip usages inside macro expansions in destructure struct/tuple binding

Albab-Hasan:
when a local binding is used inside a macro call (e.g. write!(s, "{}", x.field) or write!(s, "{}", t.0)), the usages name node lives in the macro expansion tree while the syntax editor is rooted at the source file tree. passing the expansion node to `editor.replace()` causes `apply_edits` to panic when it resolves the `SyntaxNodePtr` against the wrong tree.

fix by comparing `usage.range` (original file range, mapped back through `original_range()`) against `usage.name.text_range()` (expansion buffer range). they diverge for all macro expanded usages matching the guard already used in `inline_local_variable`.

this also replaces the `MacroStmts` ancestor check in `destructure_tuple_binding`, which only covered macros expanding to multiple statements and missed single expression macros (MacroExpr).

macro expanded field accesses are left unchanged. the user must update them manually.

fixes rust-lang/rust-analyzer#20716

### DIFF
--- a/crates/ide-assists/src/handlers/destructure_struct_binding.rs
+++ b/crates/ide-assists/src/handlers/destructure_struct_binding.rs
@@ -17,7 +17,7 @@ use syntax::{
 
 use crate::{
     assist_context::{AssistContext, Assists, SourceChangeBuilder},
-    utils::ref_field_expr::determine_ref_and_parens,
+    utils::{cover_edit_range, ref_field_expr::determine_ref_and_parens},
 };
 
 // Assist: destructure_struct_binding
@@ -358,6 +358,7 @@ fn update_usages(
     data: &StructEditData,
     field_names: &FxHashMap<SmolStr, SmolStr>,
 ) {
+    let source = ctx.source_file();
     let make = SyntaxFactory::with_mappings();
     let edits = data
         .usages
@@ -366,7 +367,9 @@ fn update_usages(
         .collect_vec();
     editor.add_mappings(make.finish_with_mappings());
     for (old, new) in edits {
-        editor.replace(old, new);
+        if let Some(range) = ctx.sema.original_range_opt(&old) {
+            editor.replace_all(cover_edit_range(source, range.range), vec![new.into()]);
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/ide-assists/src/handlers/destructure_struct_binding.rs
```diff
@@ -1006,4 +1012,33 @@ mod tests {
...
+}
+"#,
+            r#"
+use core::fmt::Write;
+struct Foo { y: i8 }
+
+fn main() {
+    let mut s = String::new();
+    let Foo { y } = Foo { y: 8 };
+    write!(s, "{}", x.y).unwrap();
```
A4-Tacks: Ideally, this should be fixed rather than ignored,
Referring to the `cover_edit_range` in #21817 ?

ChayimFriedman2:
Generally, outside of few exceptions, when we have some code in macro file and we need it in source code, we upmap it, not ignore it.

Albab-Hasan:
@A4-Tacks done. let me know if there anything else that needs to be changed.

thanks

crates/ide-assists/src/handlers/destructure_struct_binding.rs
```diff
@@ -355,17 +355,29 @@ fn new_field_name(base_name: SmolStr, names_in_scope: &FxHashSet<SmolStr>) -> Sm
...
+    for usage in &data.usages {
+        if usage.range != usage.name.text_range() {
+            // Usage is inside a macro expansion. Try to map the field
+            // expression back to the original source and replace it there.
+            if let Some(field_expr) = usage.name.syntax().ancestors().find_map(ast::FieldExpr::cast)
+                && let Some(name_ref) = field_expr.name_ref()
+                && let Some(new_name) = field_names.get(&SmolStr::from(name_ref.to_string()))
+                && let Some(file_range) = ctx.sema.original_range_opt(field_expr.syntax())
+            {
+                builder.replace(file_range.range, new_name.to_string());
```
A4-Tacks: You should use SyntaxEditor, `cover_edit_range` is already available in #21817
Perhaps you should let `build_usage_edit` handle macros naturally

Albab-Hasan:
@A4-Tacks sorry didnt realize that the pr was merged.

let me know if this is good.

thanks

crates/ide-assists/src/handlers/destructure_struct_binding.rs
```diff
@@ -365,33 +365,54 @@ fn update_usages(
...
+
 fn build_usage_edit(
     ctx: &AssistContext<'_>,
     make: &SyntaxFactory,
     data: &StructEditData,
     usage: &FileReference,
     field_names: &FxHashMap<SmolStr, SmolStr>,
-) -> Option<(SyntaxNode, SyntaxNode)> {
+) -> Option<UsageEdit> {
+    let is_macro = usage.range != usage.name.text_range();
```
A4-Tacks: Regardless of whether it is within a macro or not, you can always use `cover_edit_range` and `replace_all`
Sorry, I came up with a better solution:
Directly `cover_edit_range` and `replace_all` at the call site without changing `build_usage_edit`

Albab-Hasan:
@A4-Tacks let me know if this is good.

thanks

Albab-Hasan:
@A4-Tacks hey, i wanted to ask a question if you dont mind. I want to start reviewing prs.

Can you tell me the process of how you review prs?

A4-Tacks:
Non member review may be confusing?

Albab-Hasan:
np

PR #21839, opened 2026-03-18
fix: don't add a second semicolon after postfix completions

A4-Tacks:
Make `.let` `.letm` `.return` `.break` not duplicate semicolons

Example
---
```rust
fn main() {
    baz.l$0;
}
```

**Before this PR**

```rust
fn main() {
    let $0 = baz;;
}
```

**After this PR**

```rust
fn main() {
    let $0 = baz;
}
```

### DIFF
--- a/crates/ide-completion/src/completions/postfix.rs
+++ b/crates/ide-completion/src/completions/postfix.rs
@@ -66,6 +66,12 @@ pub(crate) fn complete_postfix(
         Some(it) => it,
         None => return,
     };
+    let semi =
+        if expr_ctx.in_block_expr && ctx.token.next_token().is_none_or(|it| it.kind() != T![;]) {
+            ";"
+        } else {
+            ""
+        };
 
     let cfg = ctx.config.find_path_config(ctx.is_nightly);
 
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
Of all the nice changes you're making currently to the `let` postfix completion, any chance you do the same for the non-postfix `let`?

Thanks anyway, of course!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21839/4eac290b58a70961e78f2e0c04f61a08b995b2cb..88a2710bcb5d8e76113c74174d223feb445d8186)*

A4-Tacks:
> for the non-postfix `let`?

Is it used for the skill of entering semicolons first and then `let` in some macros?

ChayimFriedman2:
AFAIK it inserts `let $0 = ;`.

A4-Tacks:
Are you referring to situations like `le$0expr();`?

This is quite difficult to handle.  The expression snippets are relatively simple, so it's better to use `expr().le$0;`

ChayimFriedman2:
No, I'm referring to e.g. typing `let` inside a match arm, it will be nice if it would open a block.

A4-Tacks:
Isn't this #21594? And the expression snippets are all hard coded short snippets, without perceiving the surrounding syntax tree, right

ChayimFriedman2:
#21594 was about the postfix snippet, which I'm unused to use, therefore I'm requesting you to do it for non-postfix snippets as well :)

A4-Tacks:
However, technically, non-postfix snippets (expressions snippets) are difficult to implement this

PR #21844, opened 2026-03-19
fix: Support filesystems that don't send Create events

Wilfred:
On some filesystems, particularly FUSE on Linux, we don't get Create(...) events. We do get Access(Open(Any)) events, so handle those consistently with create/modify/remove events.

This fixes missed file notifications when using Sapling SCM with EdenFS, although I believe the problem can occur on other FUSE environments.

See the individual commit messages for more details.

### DIFF
--- a/crates/vfs-notify/src/lib.rs
+++ b/crates/vfs-notify/src/lib.rs
@@ -14,7 +14,7 @@ use std::{
 };
 
 use crossbeam_channel::{Receiver, Sender, select, unbounded};
-use notify::{Config, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
+use notify::{Config, EventKind, RecommendedWatcher, RecursiveMode, Watcher, event::AccessKind};
 use paths::{AbsPath, AbsPathBuf, Utf8PathBuf};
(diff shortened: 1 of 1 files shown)


Wilfred:
I've tested on FUSE a bunch, but not much on non-FUSE. I can get you some more thorough benchmarks next week, I don't have much time for r-a this week I'm afraid.

Wilfred:
Yep, testing this with non-FUSE (i.e. a normal filesystem) and server-side watching seems fine. I've tested a rust-analyzer build with rust-analyzer open as a repository in VS Code, and performance seems entirely normal in my testing.

I've also cherry-picked this PR for the users I support at work and not seen any issues there either, so I'm pretty confident these changes are good.

Happy to do any other tests you can think of, if I've missed anything :)

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21844/63b3eff38ef1c216480147dd53b0e4365d55f269..19e0346ed891bf343e87659be21234834d323260)*

PR #21846, opened 2026-03-19
fix: Add `-Zjson-target-spec` on Cargo calls where needed

### LINKED ISSUE #21821: Custom targets triplets loaded from json now require "-Z json-target-spec"
**rust-analyzer version**: rust-analyzer version: 0.3.2819-standalone

**rustc version**: rustc 1.96.0-nightly (03749d625 2026-03-14)

**editor or extension**: VSCode 1.111.0, rust-lang.rust-analyzer 0.3.2819

**relevant settings**: 
Extension  settings:
```json
{
    "rust-analyzer.cargo.target": "<path to project>/.cargo/armv5te-nintendo-ds-newlibeabi.json"
}
```

**repository link**: 
- Library: [https://github.com/BlueTheDuck/nds-rs](https://github.com/BlueTheDuck/nds-rs)

**code snippet to reproduce**: N/A

Last January the following changes were merged into Cargo: rust-lang/cargo#16557

This pull request made it so using custom targets now required passing `-Z json-target-spec` when loading the definition from a JSON file.

In particular, this change affected the command `cargo metadata` that RA uses to gather information about the projects. 

Example output right after loading a project:
> Failed to read Cargo metadata with dependencies for <path to project>/Cargo.toml: cargo metadata exited with an error: error: .json target specs require -Zjson-target-spec

ChayimFriedman2:
Fixes rust-lang/rust-analyzer#21821.

CC @ShoyuVanilla assuming you know - do we need to gate this only for affected Cargo versions? If so, how?

### DIFF
--- a/crates/toolchain/src/lib.rs
+++ b/crates/toolchain/src/lib.rs
@@ -143,3 +143,33 @@ pub fn probe_for_binary(path: Utf8PathBuf) -> Option<Utf8PathBuf> {
     };
     iter::once(path).chain(with_extension).find(|it| it.is_file())
 }
+
+pub const MINIMUM_TOOLCHAIN_VERSION_REQUIRING_JSON_TARGET_SPEC_FLAG: semver::Version =
+    semver::Version {
+        major: 1,
+        minor: 95,
+        patch: 0,
+        pre: semver::Prerelease::EMPTY,
+        build: semver::BuildMetadata::EMPTY,
+    };
+
+/// Uses targets in a Cargo process.
+pub fn cargo_use_targets(
+    toolchain_version: Option<&semver::Version>,
+    cmd: &mut Command,
+    targets: impl IntoIterator<Item = impl AsRef<str>>,
+) {
+    let mut has_json_target = false;
+    for target in targets {
+        let target = target.as_ref();
+        cmd.args(["--target", target]);
+        has_json_target |= target.ends_with(".json");
+    }
+    if has_json_target
+        && toolchain_version.is_some_and(|version| {
+            *version >= MINIMUM_TOOLCHAIN_VERSION_REQUIRING_JSON_TARGET_SPEC_FLAG
+        })
+    {
+        cmd.arg("-Zjson-target-spec");
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3000,6 +3000,7 @@ name = "toolchain"
 version = "0.0.0"
 dependencies = [
  "camino",
+ "semver",
 ]
 
 [[package]]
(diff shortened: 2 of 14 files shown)


ShoyuVanilla:
Yeah, the older versions without that flag will error with it. For the later, I guess we could just do it based on the toolchain version. Would that be hard because of some quirks in our codebase? (I can't check whether it's hard or not since I'm outside rn. Sry 😅 )

Veykril:
We already have the sysroot here so we should have the toolchain version queried already somewhere

ShoyuVanilla:
As Lukas said we should the toolchain version available there, I guess you could compare the toolchain version like the following lines:

https://github.com/rust-lang/rust-analyzer/blob/803fb517b6eb0fb976c2eea7c5330bd1211e7287/crates/project-model/src/cargo_config_file.rs#L171-L173

MavenRain:
Hi @ChayimFriedman2, I've dropped #22435 in favour of this PR.  Two notes that may help unstick it:

  1. The cargo metadata invocation in `crates/project-model/src/cargo_workspace.rs:698-702` also passes `.json` target specs (via `--filter-platform`) and needs the same fix.  It's not covered in this PR.  The user-reported error in #21821 surfaces from that path.

  2. For the version gate, `-Zjson-target-spec` was added in rust-lang/cargo#16557, milestoned 1.95.0 (merged 2026-01-25).  The existing pattern at build_dependencies.rs:521-530 (COMP_TIME_DEPS_MIN_TOOLCHAIN_VERSION) works as precedent: `toolchain.is_some_and(|v| *v >= MIN_VERSION_FOR_JSON_TARGET_SPEC)`.
  
I'd be happy to push a commit to this branch that adds the metadata site and version gating if you'd like, or leave it for you.  Either is fine.

ChayimFriedman2:
r? @ShoyuVanilla 

But feel free to re-assign.

crates/toolchain/src/lib.rs
```diff
@@ -143,3 +143,33 @@ pub fn probe_for_binary(path: Utf8PathBuf) -> Option<Utf8PathBuf> {
...
+    targets: impl IntoIterator<Item = impl AsRef<str>>,
+) {
+    let mut has_json_target = false;
+    for target in targets {
+        let target = target.as_ref();
+        cmd.args(["--target", target]);
+        has_json_target |= target.ends_with(".json");
+    }
+    if has_json_target
+        && toolchain_version.is_some_and(|version| {
```
ShoyuVanilla, line 169: I'm a bit uncertain on whether we should make this default when the `toolchain_version.is_none()`, as it will fail when the toolchain is empty but actually `1.95.0 =<`, like in flychecks with `InvocationStrategy::Once`.
But as you have written https://github.com/rust-lang/rust-analyzer/pull/23012 this would be okay

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21846/235acb4bc6e4d27ad1c6f41b0a5bea613f44704d..dc50fc20322779c94584e0adbd836ccf377ca5f0)*

ShoyuVanilla (approved):

PR #21847, opened 2026-03-20
Replace direct usage of make with syntax factory and migrate assist to syntaxEditor

Shourya742:
This PR migrates direct use of make in assist with methods from syntax Factory, it also make migrates a couple of assists to `SyntaxEditor` and `Syntax Factory`, which are `qualify_method_call`, `qualify_path` and `desugar_try_expr`

### DIFF
--- a/crates/ide-assists/src/handlers/replace_let_with_if_let.rs
+++ b/crates/ide-assists/src/handlers/replace_let_with_if_let.rs
@@ -64,11 +64,8 @@ pub(crate) fn replace_let_with_if_let(acc: &mut Assists, ctx: &AssistContext<'_>
                     }
                 }
             };
-            let init_expr = if let_expr_needs_paren(&init, &make) {
-                make.expr_paren(init).into()
-            } else {
-                init
-            };
+            let init_expr =
+                if let_expr_needs_paren(&init) { make.expr_paren(init).into() } else { init };
 
             let block = make.block_expr([], None);
             let block = block.indent(IndentLevel::from_node(let_stmt.syntax()));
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/replace_let_with_if_let.rs
```diff
@@ -85,9 +88,8 @@ pub(crate) fn replace_let_with_if_let(acc: &mut Assists, ctx: &AssistContext<'_>
     )
 }
 
-fn let_expr_needs_paren(expr: &ast::Expr) -> bool {
-    let fake_expr_let =
-        ast::make::expr_let(ast::make::tuple_pat(None).into(), ast::make::ext::expr_unit());
+fn let_expr_needs_paren(expr: &ast::Expr, make: &SyntaxFactory) -> bool {
+    let fake_expr_let = make.expr_let(make.tuple_pat(None).into(), make.expr_unit());
```
A4-Tacks, line 90: This does not require mapping, or even a mutable tree

bit-aloo: That thought occurred to me as well while migrating, but I wasn’t sure whether it makes sense to introduce a new constructor method for this. Curious to hear your suggestion, ideally without reusing make directly. For now, I used without_mapping variant here, at least to avoid the mapping.

A4-Tacks (approved):

crates/ide-assists/src/handlers/replace_let_with_if_let.rs
A4-Tacks: > but I wasn’t sure whether it makes sense to introduce a new constructor method for this.

It seems that the goal is to completely remove the mutable tree?

bit-aloo: Yup

PR #21856, opened 2026-03-23
Replace direct use of make in replace derive with manual impl with SyntaxFactory

Shourya742:
Replace the direct use of `make` in the `replace_derive_with_manual_impl` assist with `SyntaxFactory`.

This PR also introduces variants of utility methods based on `SyntaxFactory`, which will replace the non-`SyntaxFactory` versions once all dependents are migrated.

### DIFF
--- a/crates/ide-assists/src/handlers/replace_derive_with_manual_impl.rs
+++ b/crates/ide-assists/src/handlers/replace_derive_with_manual_impl.rs
@@ -4,7 +4,7 @@ use itertools::Itertools;
 use syntax::{
     SyntaxKind::WHITESPACE,
     T,
-    ast::{self, AstNode, HasName, make},
+    ast::{self, AstNode, HasName, syntax_factory::SyntaxFactory},
     syntax_editor::{Position, SyntaxEditor},
 };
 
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/utils.rs
```diff
@@ -279,6 +282,79 @@ pub fn add_trait_assoc_items_to_impl(
...
+#[must_use]
+pub fn add_trait_assoc_items_to_impl_with_factory(
+    make: &SyntaxFactory,
+    sema: &Semantics<'_, RootDatabase>,
+    config: &AssistConfig,
+    original_items: &[InFile<ast::AssocItem>],
+    trait_: hir::Trait,
+    impl_: &ast::Impl,
+    target_scope: &hir::SemanticsScope<'_>,
+) -> Vec<ast::AssocItem> {
```
ChayimFriedman2, line 295: These methods are large enough that I prefer to share them, not duplicate.

bit-aloo: We would be removing them pretty soon, as these are only used in assist. Having these in a common flow would require additional complexity for something we would remove again later.

PR #21857, opened 2026-03-23
internal: Remove clone_for_update from flip_range_expr

Amit5601:
This PR removes the use of `.clone_for_update()` in the `flip_range_expr` assist. The `SyntaxEditor` can handle the insertion of the immutable nodes directly, so the mutable clones are no longer necessary.


ChayimFriedman2:
Thanks!

A4-Tacks:
The original tree can be auto mutable and only supports replaces, not inserts

crates/syntax/src/syntax_editor/edit_algo.rs:197

A4-Tacks:
This PR is wrong

Unless you add features to `edit_algo`

ChayimFriedman2:
As @A4-Tacks said you can't do that, apparently.

Issue #21837, opened 2026-03-18
Reports false type errors on macro, but actually compiles without errors


lzyor:
**rust-analyzer version**: 0.3.2828-standalone

**rustc version**: rustc 1.90.0 (1159e78c4 2025-09-14) ~ rustc 1.94.0 (4a4ef493e 2026-03-02)

**editor or extension**: VSCode 1.109.5

**code snippet to reproduce**:
```rust
use ambassador::{Delegate, delegatable_trait};

#[delegatable_trait]
pub trait Shout {
    fn shout(&self, input: &str) -> String;
}

pub struct Cat;

impl Shout for Cat {
    fn shout(&self, input: &str) -> String {
        format!("{} - meow!", input)
    }
}

pub struct Dog;

impl Shout for Dog {
    fn shout(&self, input: &str) -> String {
        format!("{} - wuff!", input)
    }
}

#[derive(Delegate)]
#[delegate(Shout)]
pub enum Animal {
    Cat(Cat),
    Dog(Dog),
}

pub fn main() {
    let foo_animal = Animal::Cat(Cat);
    println!("{}", foo_animal.shout("BAR"));
}
```

Simply run the example code from the `ambassador` crate documentation: https://crates.io/crates/ambassador

rust-analyzer reports errors:

```text
---> #[derive(Delegate)]
expected &{unknown}, found Dog rust-analyzer(E0308)
expected &{unknown}, found Cat rust-analyzer(E0308)
```

However, there are actually no errors, and the macro-expanded code compiles successfully.

ChayimFriedman2:
Likely related to order of name resolution.

AayushMainali-Github:
@ChayimFriedman2  I opened a PR for this with a fix and regression test. Would appreciate a look when
  you get a chance.

zirconium-n:
I hit this problem as well. Here's some more info.

`#[delegatable_trait]` expands to:

```rust
#[macro_export]
#[doc(hidden)]
macro_rules! _ambassador_impl_Shout {
    (body_struct(< > , $ty:ty, $field_ident:tt)) => {
        ambassador_impl_Shout!{
            body_struct(< > , $ty,($field_ident),($field_ident),($field_ident))
        }
    };
    (body_struct(< > , $ty:ty,($($ident_owned:tt)*),($($ident_ref:tt)*),($($ident_ref_mut:tt)*))) => {
[...]

PR #21858, opened 2026-03-23
hir-ty: collect impls from named const bodies

AayushMainali-Github:
This fixes impl collection for impls defined inside named const bodies.

  `hir-ty` already descended into unnamed const items (`const _: () = { ... }`) when collecting impls, but it skipped named consts. That means impls generated inside named const bodies could be ignored during method resolution.

  This change makes impl collection recurse into all const bodies instead of only unnamed consts, and adds regression tests covering both trait and inherent impls inside named consts.

  I have not independently confirmed the exact expansion shape emitted by `ambassador` for the issue example, so I cannot claim with certainty that this is the only root cause of rust-lang/rust-analyzer#21837. However, this does fix a concrete gap in custom-derive-style impl collection that matches the review feedback and is plausibly related to the reported behavior.

  Testing:
  - Added `trait_impl_in_named_const`
  - Added `inherent_impl_in_named_const`

### DIFF
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -36,7 +36,7 @@ use hir_def::{
     AdtId, AssocItemId, ConstId, ConstParamId, DefWithBodyId, ExpressionStoreOwner, FieldId,
     FunctionId, GenericDefId, GenericParamId, ItemContainerId, LocalFieldId, Lookup, TraitId,
     TupleFieldId, TupleId, TypeAliasId, TypeOrConstParamId, VariantId,
-    expr_store::{ConstExprOrigin, ExpressionStore, path::Path},
+    expr_store::{ConstExprOrigin, ExpressionStore, HygieneId, path::Path},
     hir::{BindingAnnotation, BindingId, ExprId, ExprOrPatId, LabelId, PatId},
     lang_item::LangItems,
     layout::Integer,
@@ -1644,11 +1644,7 @@ impl<'body, 'db> InferenceContext<'body, 'db> {
         let mut path_ctx = ctx.at_path(path, node);
         let interner = DbInterner::conjure();
         let (resolution, unresolved) = if value_ns {
-            let hygiene = match node {
-                ExprOrPatId::ExprId(expr) => self.store.expr_path_hygiene(expr),
-                ExprOrPatId::PatId(pat) => self.store.pat_path_hygiene(pat),
-            };
-            let Some(res) = path_ctx.resolve_path_in_value_ns(hygiene) else {
+            let Some(res) = path_ctx.resolve_path_in_value_ns(HygieneId::ROOT) else {
                 return (self.err_ty(), None);
             };
             match res {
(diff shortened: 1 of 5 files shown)


crates/hir-ty/src/infer.rs
```diff
@@ -1644,7 +1644,11 @@ impl<'body, 'db> InferenceContext<'body, 'db> {
         let mut path_ctx = ctx.at_path(path, node);
         let interner = DbInterner::conjure();
         let (resolution, unresolved) = if value_ns {
-            let Some(res) = path_ctx.resolve_path_in_value_ns(HygieneId::ROOT) else {
+            let hygiene = match node {
+                ExprOrPatId::ExprId(expr) => self.store.expr_path_hygiene(expr),
+                ExprOrPatId::PatId(pat) => self.store.pat_path_hygiene(pat),
+            };
+            let Some(res) = path_ctx.resolve_path_in_value_ns(hygiene) else {
```
ChayimFriedman2:
```suggestion
            let Some(res) = path_ctx.resolve_path_in_value_ns(self.store.expr_ot_pat_path_hygiene(node)) else {
```

crates/test-fixture/src/lib.rs
```diff
@@ -900,6 +915,39 @@ impl ProcMacroExpander for Issue18089ProcMacroExpander {
...
+        _: &TopSubtree,
+        _: Option<&TopSubtree>,
+        _: &Env,
+        _: Span,
+        call_site: Span,
+        _: Span,
+        _: String,
+    ) -> Result<TopSubtree, ProcMacroExpansionError> {
+        Ok(quote! { call_site =>
+            const __DELEGATE_SHOUT_IMPL: () = {
```
ChayimFriedman2: Impls inside named consts for types not defined inside the const are not considered. If this is what `ambassador` does, chances are that is in fact the cause of the bug.

crates/test-fixture/src/lib.rs
```diff
@@ -577,6 +577,21 @@ pub fn derive_identity(item: TokenStream) -> TokenStream {
         ),
         (
             r#"
+#[proc_macro_derive(DelegateShout)]
```
ChayimFriedman2: I don't see why you need a proc macro for that. A `macro_rules` should be sufficient.

AayushMainali-Github:
> I'm not convinced this is indeed the cause, in fact I'm pretty convinced you're wrong. See my comments.
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21858/da2fe7fed8960a4e351818d52bf444d7c9569da7..02f5985fea46f7a5cac5bf9baab1786fa3defefe)_

Sure I will look into it

AayushMainali-Github:
Hey @ChayimFriedman2, So I reworked this based on your comments. I dropped the `infer.rs` change and moved the fix to impl collection: we were already descending into `const _: ...` bodies, but not named consts. This patch makes it recurse into named const bodies too and adds direct regression tests for that.

I haven’t checked the exact `ambassador` expansion yet, so I’m not claiming full certainty for #21837. Does this look closer to what you had in mind?

ChayimFriedman2:
We do not collect named consts deliberately. This is a won't fix.

AayushMainali-Github:
Makes sense, thanks for clarifying. I’ll close the PR then.

PR #21861, opened 2026-03-24
migrate convert_closure_to_fn assist to SyntaxEditor

ogunbor:
Migrates `convert_closure_to_fn` assist from mutable tree 
editing (`ted::`) to the `SyntaxEditor` API.

Removes the `ted` import and replaces the single `ted::replace` 
call with an equivalent `SyntaxEditor` pattern.

Part of rust-lang/rust-analyzer#18285


ChayimFriedman2:
Already has an open PR, https://github.com/rust-lang/rust-analyzer/pull/21790.

PR #21876, opened 2026-03-25
fix: wrap ty-anchor in non-path type constuctor

A4-Tacks:
Fixes rust-lang/rust-analyzer#21873

Example
---
```rust
struct TestStruct { r: &'static [i32] }
fn test_fn() {
    let s = TestStruct{ $0 };
}
```

**Before this PR**

```rust
struct TestStruct { r: &'static [i32] }
fn test_fn() {
    let s = TestStruct{ r: &'static  };
}
```

**After this PR**

```rust
struct TestStruct { r: &'static [i32] }
fn test_fn() {
    let s = TestStruct{ r: <&'static [i32]>::default()  };
}
```

### DIFF
--- a/crates/syntax/src/ast/make.rs
+++ b/crates/syntax/src/ast/make.rs
@@ -74,10 +74,18 @@ pub mod ext {
         expr_from_text("_")
     }
     pub fn expr_ty_default(ty: &ast::Type) -> ast::Expr {
-        expr_from_text(&format!("{ty}::default()"))
+        if !ty.needs_angles_in_path() {
+            expr_from_text(&format!("{ty}::default()"))
+        } else {
+            expr_from_text(&format!("<{ty}>::default()"))
+        }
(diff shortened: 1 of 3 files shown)


crates/syntax/src/ast/make.rs
```diff
@@ -74,10 +74,18 @@ pub mod ext {
         expr_from_text("_")
     }
     pub fn expr_ty_default(ty: &ast::Type) -> ast::Expr {
-        expr_from_text(&format!("{ty}::default()"))
+        if matches!(ty, ast::Type::PathType(_)) && ty.generic_arg_list().is_none() {
```
ChayimFriedman2: Seems worthy to extract the logic somewhere in `syntax`, like we have `needs_parens_in()`.

ShoyuVanilla (approved):

PR #21877, opened 2026-03-26
fix: Correct Display label for Event::FetchWorkspaces

cuiweixie:
## Summary

- Fix `fmt::Display` for `Event::FetchWorkspaces`: it incorrectly printed `Event::SwitchWorkspaces`, which was misleading in tracing output.
- Add a unit test that asserts the display string matches the variant name.

## Notes

## Verification

On the parent commit, `Event::FetchWorkspaces` mapped to `"Event::SwitchWorkspaces"` in `Display`; the new test fails in that state and passes with the fix:

```bash
# With only the test on the buggy `Display` (parent `main_loop.rs` + test block): test fails.
# With the corrected `Display` line: test passes.
```

### DIFF
--- a/crates/rust-analyzer/src/main_loop.rs
+++ b/crates/rust-analyzer/src/main_loop.rs
@@ -92,7 +92,7 @@ impl fmt::Display for Event {
             Event::DeferredTask(_) => write!(f, "Event::DeferredTask"),
             Event::TestResult(_) => write!(f, "Event::TestResult"),
             Event::DiscoverProject(_) => write!(f, "Event::DiscoverProject"),
-            Event::FetchWorkspaces(_) => write!(f, "Event::SwitchWorkspaces"),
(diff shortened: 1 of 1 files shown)


cuiweixie:
> Please remove the test. We don't need a test for something so simple.
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21877/9253d39eab8b9c9da3c1412fc94764e01d55a02b..945628ba51dc55cb87124a07d5f1ed78d2915e7b)_

Done

ShoyuVanilla:
Looks like we can't use bor commands here 😅  Could you squash two commits into one?

cuiweixie:
> Looks like we can't use bor commands here 😅 Could you squash two commits into one?

Done

PR #21878, opened 2026-03-26
fix: Report correct crates_total in prime caches Done progress

cuiweixie:
## Summary

When the parallel prime-caches progress channel closed, the final `Done` callback set `crates_total` to `crate_def_maps_done` instead of `crate_def_maps_total`. The total should match the def-map indexing phase (same as other `ParallelPrimeCachesProgress` updates).

## Notes


ChayimFriedman2:
It doesn't matter, done == equal at this stage.

PR #21881, opened 2026-03-26
fix: Correct missing-args messages for sched_getaffinity and getenv shims

cuiweixie:
## Summary

When arguments were missing, `sched_getaffinity` and `getenv` reported `libc::write args are not provided`. Use shim-specific messages instead.

## Notes


lnicola:
@cuiweixie can you run `cargo fmt` and squash?

PR #21886, opened 2026-03-26
Replace make constructor with syntaxFactory in utils/gen trait fn body

Shourya742:
This PR replaces the `make` constructor with `SyntaxFactory` in `gen_trait_fn_body`, which directly affects two assists that are also updated in this PR:

1. `add_missing_impl_members`
2. `replace_derive_with_manual_impl`

### DIFF
--- a/crates/syntax/src/ast/syntax_factory/constructors.rs
+++ b/crates/syntax/src/ast/syntax_factory/constructors.rs
@@ -2053,13 +2053,7 @@ impl SyntaxFactory {
         &self,
         parts: impl IntoIterator<Item = &'a str>,
     ) -> Option<ast::Path> {
-        let mut iter = parts.into_iter();
-        let base = self.ident_path(iter.next()?);
-        let path = iter.fold(base, |base, s| {
-            let segment = self.ident_path(s);
-            self.path_concat(base, segment)
-        });
-        Some(path)
+        make::ext::path_from_idents(parts).map(|path| path.clone_for_update())
     }
 
     pub fn token_tree_from_node(&self, node: &SyntaxNode) -> ast::TokenTree {


crates/syntax/src/ast/syntax_factory/constructors.rs
```diff
@@ -2049,6 +2049,23 @@ impl SyntaxFactory {
         self.path_unqualified(self.path_segment(self.name_ref(ident)))
     }
 
+    pub fn path_from_idents<'a>(
+        &self,
+        parts: impl IntoIterator<Item = &'a str>,
+    ) -> Option<ast::Path> {
+        let mut iter = parts.into_iter();
+        let base = self.ident_path(iter.next()?);
+        let path = iter.fold(base, |base, s| {
```
A4-Tacks: Why not call `make::ext::path_from_idents`

crates/syntax/src/ast/syntax_factory/constructors.rs
```diff
@@ -2049,6 +2049,23 @@ impl SyntaxFactory {
...
+            let segment = self.ident_path(s);
+            self.path_concat(base, segment)
+        });
+        Some(path)
+    }
+
+    pub fn token_tree_from_node(&self, node: &SyntaxNode) -> ast::TokenTree {
+        make::ext::token_tree_from_node(node).clone_for_update()
```
A4-Tacks: Doesn't this require mapping?

crates/syntax/src/ast/syntax_factory/constructors.rs
bit-aloo: I reverted this change. I had originally implemented the manual construction because all the constituent methods had already been ported to SyntaxFactory.

crates/syntax/src/ast/syntax_factory/constructors.rs
bit-aloo: Added. Thanks

crates/syntax/src/ast/syntax_factory/constructors.rs
```diff
@@ -2049,6 +2049,31 @@ impl SyntaxFactory {
...
+        make::ext::path_from_idents(parts).map(|path| path.clone_for_update())
+    }
+
+    pub fn token_tree_from_node(&self, node: &SyntaxNode) -> ast::TokenTree {
+        let ast = make::ext::token_tree_from_node(node).clone_for_update();
+
+        if let Some(mut mapping) = self.mappings() {
+            let mut builder = SyntaxMappingBuilder::new(ast.syntax().clone());
+            let target = ast
+                .l_curly_token()
```
A4-Tacks: This could also be `l_paren` and `l_brack`

bit-aloo: Mapping here is itself wrong. My bad for not thinking deeper for this. Considering there is no child of same kind in output. Even the above implementation should panic while casting to a node, considering all the children are token. Removed.

PR #21887, opened 2026-03-27
fix a tests that assunes the wrong alignment on s390x

fneddy:
on s390x vectors are aligned at 8 bytes not 16 or vector length like on x86.


lnicola:
Do all the tests (including `RUN_SLOW_TESTS=1`) pass on s390x with this change?

fneddy:
i ran `RUN_SLOW_TESTS=1 cargo test` on s390x and all is green :+1:

PR #21892, opened 2026-03-28
feat: add expected name on simple enum variant

A4-Tacks:
Example
---
```rust
struct Other;
struct String;
enum Foo {
    String($0)
}
```

**Before this PR**

```text
en Foo Foo []
st Other Other []
sp Self Foo []
st String String []
```

**After this PR**

```text
st String String [name]
en Foo Foo []
st Other Other []
sp Self Foo []
```

### DIFF
--- a/crates/ide-completion/src/context/analysis.rs
+++ b/crates/ide-completion/src/context/analysis.rs
@@ -827,6 +827,14 @@ fn expected_type_and_name<'db>(
                         .map(|c| (Some(c.return_type()), None))
                         .unwrap_or((None, None))
                 },
+                ast::Variant(it) => {
+                    let is_simple_variant = matches!(
+                        it.field_list(),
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
I'm not sure what this does. Why did the scoring change?

A4-Tacks:
Suggest variant name within simple tuple variants

This is beneficial for writing `String(String)`, `Error(Error)`, etc

> Why did the scoring change?

Because the name matches, as described in the test

ChayimFriedman2 (approved):
Got it, it doesn't hurt I guess.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21892/4407174f9fb6f163ca403f5dbbc3aff5d3836fd6..e3ff47611781f9b1bbf56741cb55c2ccf1e9e5ec)*

PR #21901, opened 2026-03-28
Only allocate item blocks if they actually contain items or statement macros

Veykril:
Should help with https://github.com/rust-lang/rust-analyzer/pull/21897's slight memory regression


lnicola:
@Veykril this broke `analysis-stats` on `clap`, https://github.com/rust-lang/rust-analyzer/actions/runs/23697924204/job/69036307405

Veykril:
oh well that cfg is wrong 🤦

PR #21905, opened 2026-03-29
Migrate assists to SyntaxFactory with_mappings

ogunbor:
Migrates the following assists from `SyntaxFactory::without_mappings()` 
to `SyntaxFactory::with_mappings()`, adding the required 
`editor.add_mappings(make.finish_with_mappings())` call before 
each `add_file_edits`.

- replace_named_generic_with_impl
- extract_type_alias
- generate_default_from_new
- generate_fn_type_alias

Part of rust-lang/rust-analyzer#18285


ChayimFriedman2:
There is no point in using `with_mappings()` when `without_mappings()` works.

PR #21906, opened 2026-03-29
feat: exclude refs(find all refs) from deps and stdlib

### LINKED ISSUE #17184: feat: improve references search
- _[exclude_tests](https://github.com/rust-lang/rust-analyzer/blob/1a5bb27c018c947dab01ab70ffe1d267b0481a17/crates/rust-analyzer/src/handlers/request.rs#L1153-L1158) should be pushed down as a parameter to find_all_refs._
- _Something can be a test and a read at the same time._

_originally post by matklad at https://github.com/rust-lang/rust-analyzer/pull/16441#discussion_r1564288229_

Young-Flash:
follow up https://github.com/rust-lang/rust-analyzer/pull/16441, close https://github.com/rust-lang/rust-analyzer/issues/17184

### DIFF
--- a/crates/ide-db/src/search.rs
+++ b/crates/ide-db/src/search.rs
@@ -615,7 +615,7 @@ impl<'a> FindUsages<'a> {
         search_scope: &SearchScope,
         name: &str,
     ) -> bool {
-        if self.scope.is_some() || self.exclude_library_files {
+        if self.scope.is_some() {
             return false;
         }
 
@@ -897,12 +897,16 @@ impl<'a> FindUsages<'a> {
         let finder = Finder::new(name.as_bytes());
         // The search for `Self` may return duplicate results with `ContainerName`, so deduplicate them.
         let mut self_positions = FxHashSet::default();
+        let is_possibly_self = is_possibly_self.into_iter().filter(|position| {
+            !self.exclude_library_files
+                || !is_library_file(self.sema.db, position.file_id.file_id(self.sema.db))
+        });
         tracing::info_span!("Self_search").in_scope(|| {
             search(
                 self,
                 &finder,
                 name,
-                is_possibly_self.into_iter().map(|position| {
+                is_possibly_self.map(|position| {
                     (position.file_text(self.sema.db).clone(), position.file_id, position.range)
                 }),
                 |path, name_position| {
--- a/crates/ide-db/src/search.rs
+++ b/crates/ide-db/src/search.rs
@@ -449,6 +449,8 @@ impl Definition {
             scope: None,
             include_self_kw_refs: None,
             search_self_mod: false,
+            excluded_categories: ReferenceCategory::empty(),
+            exclude_library_files: false,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


ChayimFriedman2:
I'm not keen on another config to be honest. I prefer an idea @Veykril has suggested in the past (@Veykril do you agree?): search in dependencies if the search is triggered from dependency source, exclude them if the search is from user code. This seems intuitive.

crates/ide-db/src/search.rs
```diff
@@ -465,6 +467,10 @@ pub struct FindUsages<'a> {
     include_self_kw_refs: Option<hir::Type<'a>>,
     /// whether to search for the `self` module
     search_self_mod: bool,
+    /// categories to exclude while collecting usages
+    excluded_categories: ReferenceCategory,
```
ChayimFriedman2, line 471: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21906/9eb97ea96d8400e8957ddd56702e962614296583..7490a86f806da7fd936e146517696a6fc01b8b11)*

I prefer this to be `included_categories`.

crates/ide-db/src/search.rs
```diff
@@ -931,6 +951,14 @@ impl<'a> FindUsages<'a> {
                 Some(scope) => base.intersection(scope),
             }
         };
+        if self.exclude_library_files {
```
ChayimFriedman2, line 954: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21906/9eb97ea96d8400e8957ddd56702e962614296583..7490a86f806da7fd936e146517696a6fc01b8b11)*

It's better to do that in `scope_files()`. This will also handle the case you forgot in `short_associated_function_fast_search()`.

crates/ide/src/references.rs
```diff
@@ -166,7 +182,8 @@ pub(crate) fn find_all_refs(
                     is_mut: matches!(def, Definition::Local(l) if l.is_mut(sema.db)),
                     nav,
                 }
-            });
+            })
+            .filter(|decl| !(exclude_library_refs && is_library_file(sema.db, decl.nav.file_id)));
```
ChayimFriedman2, line 186: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21906/9eb97ea96d8400e8957ddd56702e962614296583..7490a86f806da7fd936e146517696a6fc01b8b11)*

No need to do that here given it's already filtered by the search.

crates/ide/src/references.rs
```diff
@@ -503,13 +527,138 @@ fn test() {
...
+    foo();
+}
+"#,
+            false,
+            false,
+            expect![[r#"
+                FileId(0) 9..12 import
+                FileId(0) 31..34
+            "#]],
+        );
```
ChayimFriedman2, line 588: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21906/9eb97ea96d8400e8957ddd56702e962614296583..7490a86f806da7fd936e146517696a6fc01b8b11)*

I think we always want the definition, even when inside a library. Only exclude other references.

ChayimFriedman2 (changes requested):

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21906/9eb97ea96d8400e8957ddd56702e962614296583..8b707e3846ab5d979f6d70597d86e47985f4e1c2)*

PR #21908, opened 2026-03-29
fix: transform `return` into `break` when inlining functions

### LINKED ISSUE #19377: Inlining a method with an early return does not handle early return correctly.
**rust-analyzer version**: rust-analyzer version: 0.3.2319-standalone [/home/lennart/.vscode/extensions/rust-lang.rust-analyzer-0.3.2319-linux-x64/server/rust-analyzer]

**editor or extension**: VSCode v0.3.2319

**relevant settings**: Base settings

**repository link (if public, optional)**: https://github.com/pc2/sus-compiler

Using the code action "Inline `early_return`" also inlines the `return None` statement unmodified, which does not match the original function's behavior 
```rust
fn early_return(a: Option<usize>) -> Option<()> {
    let Some(x) = a else { return None };
    return Some(());
}
fn use_early_return() -> Option<()> {
    early_return(Some(6))?;

    Some(())
}
```
produces
```rust
fn use_early_return() -> Option<()> {
    {
        let a = Some(6);
        let Some(x) = a else { return None };
        return Some(());
    }?;

    Some(())
}
```

Albab-Hasan:
when inlining a function containing `return` expressions the `return` would incorrectly exit the caller rather than the inlined block. wrap the inlined body in a labeled block (`'inline: { ... }`) and rewrite `return <expr>` to `break 'inline <expr>`.

returns inside closures, async, const, and gen blocks are not changed because they have their own return scope.

Closes rust-lang/rust-analyzer#19377

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -2904,23 +2904,21 @@ impl Function {
         }
     }
 
-    /// Returns the text ranges of `return` keywords in this function's body,
+    /// Returns the `return` expressions in this function's body,
     /// excluding those inside closures or async blocks.
-    pub fn return_points(self, db: &dyn HirDatabase) -> Vec<TextRange> {
+    pub fn return_points(self, db: &dyn HirDatabase) -> Vec<ast::ReturnExpr> {
         let func_id: FunctionId = match self.id {
             AnyFunctionId::FunctionId(id) => id,
             _ => return vec![],
         };
         let (body, source_map) = Body::with_source_map(db, func_id.into());
-        let fn_file_id = func_id.loc(db).id.file_id;
 
         fn collect_returns(
             body: &Body,
             source_map: &hir_def::expr_store::ExpressionStoreSourceMap,
             db: &dyn HirDatabase,
-            fn_file_id: HirFileId,
             expr_id: ExprId,
-            acc: &mut Vec<TextRange>,
+            acc: &mut Vec<ast::ReturnExpr>,
         ) {
             match &body[expr_id] {
                 Expr::Closure { .. } | Expr::Const(_) => return,
@@ -2929,30 +2927,18 @@ impl Function {
                         && let Some(ret_expr) = source.value.cast::<ast::ReturnExpr>()
                     {
                         let root = db.parse_or_expand(source.file_id);
-                        let node = ret_expr.to_node(&root);
-                        if let Some(return_token) = node.return_token() {
-                            let token_range = return_token.text_range();
-                            if source.file_id == fn_file_id {
-                                acc.push(token_range);
-                            } else if let Some((file_range, _)) =
-                                InFile::new(source.file_id, token_range)
-                                    .original_node_file_range_opt(db)
-                                && file_range.file_id == fn_file_id
-                            {
-                                acc.push(file_range.range);
-                            }
-                        }
+                        acc.push(ret_expr.to_node(&root));
                     }
                 }
                 _ => {}
             }
             body.walk_child_exprs(expr_id, |child| {
-                collect_returns(body, source_map, db, fn_file_id, child, acc);
+                collect_returns(body, source_map, db, child, acc);
             });
         }
 
         let mut returns = vec![];
-        collect_returns(body, source_map, db, fn_file_id, body.root_expr(), &mut returns);
+        collect_returns(body, source_map, db, body.root_expr(), &mut returns);
         returns
     }
 
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -2906,7 +2906,7 @@ impl Function {
 
     /// Returns the `return` expressions in this function's body,
     /// excluding those inside closures or async blocks.
-    pub fn return_points(self, db: &dyn HirDatabase) -> Vec<ast::ReturnExpr> {
+    pub fn return_points(self, db: &dyn HirDatabase) -> Vec<InFile<ast::ReturnExpr>> {
         let func_id: FunctionId = match self.id {
             AnyFunctionId::FunctionId(id) => id,
             _ => return vec![],
@@ -2918,7 +2918,7 @@ impl Function {
             source_map: &hir_def::expr_store::ExpressionStoreSourceMap,
             db: &dyn HirDatabase,
             expr_id: ExprId,
-            acc: &mut Vec<ast::ReturnExpr>,
+            acc: &mut Vec<InFile<ast::ReturnExpr>>,
         ) {
             match &body[expr_id] {
                 Expr::Closure { .. } | Expr::Const(_) => return,
@@ -2927,7 +2927,7 @@ impl Function {
                         && let Some(ret_expr) = source.value.cast::<ast::ReturnExpr>()
                     {
                         let root = db.parse_or_expand(source.file_id);
-                        acc.push(ret_expr.to_node(&root));
+                        acc.push(InFile::new(source.file_id, ret_expr.to_node(&root)));
                     }
                 }
                 _ => {}
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -2904,34 +2904,55 @@ impl Function {
         }
     }
 
-    /// Returns the return expressions in this function's body that belong to the
-    /// function itself (not inside closures or async blocks which create their own
-    /// return scope). Uses HIR analysis, which handles returns inside macro expansions.
-    pub fn return_points(self, db: &dyn HirDatabase) -> Vec<InFile<AstPtr<ast::ReturnExpr>>> {
+    /// Returns the text ranges of `return` keywords in this function's body,
+    /// excluding those inside closures or async blocks.
+    pub fn return_points(self, db: &dyn HirDatabase) -> Vec<TextRange> {
         let func_id: FunctionId = match self.id {
             AnyFunctionId::FunctionId(id) => id,
             _ => return vec![],
         };
         let (body, source_map) = Body::with_source_map(db, func_id.into());
-        let mut returns = vec![];
-        let mut to_visit = vec![body.root_expr()];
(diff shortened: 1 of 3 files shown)


ChayimFriedman2:
I have a few concerns with this PR:

 - It extends an assist using mutable tree editing, which I feel isn't the correct call at the moment.
 - To handle macros and other cursed cases, we'll need at least to descend into the tokens, and that complicates things because we can do it only for the original function. Furthermore I believe the correct way to handle that is by exposing a function from `hir`, say `fn return_points(func: Function) -> Vec<SyntaxNode>`, not using textual analysis which is inherently incomplete.

Albab-Hasan:
@ChayimFriedman2 Thanks for the review. 

agree the syntactic approach is likely incomplete and not the best way to go. walking `.descendants()` on the cloned tree can't see returns inside macro expansions.

im thinking of adding a method on `hir::Function` (something like `return_points()`) that uses Body + BodySourceMap to find returns at the HIR level. Since HIR lowering already expands macros `Expr::Return` naturally covers macro generated returns too. the walk would skip expressions that create their own return scope (closures, async blocks, const blocks). then `expr_syntax()` maps the results back to syntax locations for the assist to consume.

for macro-generated returns specifically: we can detect them (the `InFile` from `expr_syntax()` will have a macro `file_id`) but i dont think we can rewrite them since the macro will still expand to return after inlining. i think the safest option is to not offer the assist when the function contains macro-generated returns. does that sound right?

ChayimFriedman2:
Using the HIR is exactly what I meant.

As for macros, we can upmap the `return` keyword, and if it maps to something sensible (i.e. in the source code we're editing), replace it with `break 'block`.

crates/hir/src/lib.rs
```diff
@@ -2824,6 +2824,35 @@ impl Function {
         }
     }
 
+    /// Returns the return expressions in this function's body that belong to the
+    /// function itself (not inside closures or async blocks which create their own
+    /// return scope). Uses HIR analysis, which handles returns inside macro expansions.
```
ChayimFriedman2:
```suggestion
    /// return scope).
```
No need to say it's using HIR analysis, it's obvious for any function in this file.

crates/hir/src/lib.rs
```diff
@@ -2824,6 +2824,35 @@ impl Function {
...
+    pub fn return_points(self, db: &dyn HirDatabase) -> Vec<InFile<AstPtr<ast::ReturnExpr>>> {
+        let func_id: FunctionId = match self.id {
+            AnyFunctionId::FunctionId(id) => id,
+            _ => return vec![],
+        };
+        let (body, source_map) = Body::with_source_map(db, func_id.into());
+        let mut returns = vec![];
+        let mut to_visit = vec![body.root_expr()];
```
ChayimFriedman2: A recursion is simpler IMO than an out-of-line traversal.

Albab-Hasan:
@ChayimFriedman2 done

crates/ide-assists/src/handlers/inline_call.rs
```diff
@@ -333,6 +333,66 @@ fn inline(
...
+
+    // Collect return expressions before modifying the body tree, as modifications shift
+    // text ranges. Returns inside async functions don't need transformation because the
+    // body gets wrapped in `async move { ... }` which preserves return semantics.
+    let is_async_fn = function.is_async(sema.db);
+    let returns_to_replace: Vec<ast::ReturnExpr> = if is_async_fn {
+        vec![]
+    } else if file_id.is_macro() {
+        // Function defined in a macro — the body was prettified so HIR source map ranges
+        // don't match. Walk the prettified body syntactically instead.
```
ChayimFriedman2: What? No. You can use `return_points()` on a function defined in a macro just fine.

crates/ide-assists/src/handlers/inline_call.rs
```diff
@@ -333,6 +333,66 @@ fn inline(
...
+                    let return_expr = ret.value.to_node(&root);
+                    let hir::FileRange { file_id, range } =
+                        sema.original_range_opt(return_expr.syntax())?;
+                    if file_id != function_def_file_id {
+                        return None;
+                    }
+                    body.syntax()
+                        .descendants()
+                        .filter_map(ast::ReturnExpr::cast)
+                        .find(|r| r.syntax().text_range() == range)
```
ChayimFriedman2: Inside a (function-like) macro we don't have `ReturnExpr`, we have a `TokenTree`. That's why I deliberately suggested mapping the `return` **keyword**, and collecting `Vec<TextRange>`.

crates/ide-assists/src/handlers/inline_call.rs
```diff
@@ -568,18 +627,38 @@ fn inline(
         });
     }
 
+    // Transform return expressions into break expressions with a labeled block.
+    // This is needed because `return` in the inlined body would otherwise exit
+    // the *caller* function, not just the inlined block.
+    if !returns_to_replace.is_empty() {
```
ChayimFriedman2: And then here you can just do a textual edit of the upmapped `return` with `break 'label`.

ChayimFriedman2 (changes requested):

Albab-Hasan:
@ChayimFriedman2 done. sorry for the delay, let me know if anything needs to be changed

crates/hir/src/lib.rs
```diff
@@ -2824,6 +2824,58 @@ impl Function {
         }
     }
 
+    /// Returns the text ranges of `return` keywords in this function's body,
+    /// excluding those inside closures or async blocks.
+    pub fn return_points(self, db: &dyn HirDatabase) -> Vec<TextRange> {
```
ChayimFriedman2: This function should still IMO return `Vec<ast::ReturnExpr>`, without upmapping them, as a general-purpose API (that maybe will be useful elsewhere as well). Do the upmapping in this assist.

ChayimFriedman2 (changes requested):

crates/ide-assists/src/handlers/inline_call.rs
```diff
@@ -605,6 +626,33 @@ fn is_in_type_path(self_tok: &syntax::SyntaxToken) -> bool {
         .is_some()
 }
 
+/// Replaces `return` / `return expr` with `break 'inline` / `break 'inline expr`
+/// for all `ReturnExpr` nodes that belong directly to the function body
+/// (i.e., not nested inside closures or async blocks).
+fn replace_returns_with_breaks(body: &ast::BlockExpr) {
```
ChayimFriedman2: This doesn't use `return_points()`.

crates/hir/src/lib.rs
```diff
@@ -2824,6 +2824,44 @@ impl Function {
         }
     }
 
+    /// Returns the `return` expressions in this function's body,
+    /// excluding those inside closures or async blocks.
+    pub fn return_points(self, db: &dyn HirDatabase) -> Vec<ast::ReturnExpr> {
```
ChayimFriedman2: Probably better to return `Vec<InFile<ast::ReturnExpr>>`, even if this is not needed here.

ChayimFriedman2 (changes requested):

crates/ide-assists/src/handlers/inline_call.rs
Albab-Hasan: `replace_returns_with_breaks` operates on the cloned body (a different syntax tree) so the `ReturnExpr` nodes from `return_points()` dont correspond to it. how do you want to use `return_points()`?

ChayimFriedman2: If you map the returns on the original body, the ranges will stay the same. Also as I said use text replacement.

Albab-Hasan: thanks

Albab-Hasan:
@ChayimFriedman2 done. can u pls check it out

ChayimFriedman2:
The current revision is not what I had in mind still. But I realized that what I had in mind is very difficult because the various edits we do with the cloned body make it difficult to keep track of the ranges.

This is actually a perfect place to use `SyntaxEditor`, as it is able of keeping this track itself. This assist was not migrated yet however.

I also stole `return_points()` for a branch of mine and this made me realize that it should be a method on `Semantics` (I called it `fn_return_points()`), as it works with ASTs, so it needs to cache them in the `Semantics`.

Albab-Hasan:
@ChayimFriedman2 thanks, ive moved `return_points()` to semantics as `fn_return_points()` as you suggested.

for the `SyntaxEditor` migration i looked into it but the entire assist currently uses `clone_for_update()` + `ted`, and all the range based `covering_element` lookups (for params, self, Self type, etc.) depend on that. migrating just the return transformation to `SyntaxEditor` doesnt work cleanly since `SyntaxEditor::finish()` produces a new tree that breaks the existing range tracking. this needs a full assist migration which is gonna take a while for me to do. (btw should i push into this pr or should it be in a followup pr?)

also really cool that you picked up `fn_return_points()` for your rename PR. glad the API turned out to be useful beyond this assist.

Albab-Hasan:
@ChayimFriedman2 can i get a approval on weather i should work with this approach or not?

ChayimFriedman2:
Indeed, inline_call is one of the more complicated assists to migrate. And it should definitely be done in a separate PR.

If you want to work on migrating it, that's of course fine. If you do not, that is also fine, but I think I'll close this PR as unactionable for now. We can return to it after the assist is migrated.

Albab-Hasan:
@ChayimFriedman2 that sounds great. ill start working on the migration.

also i would appreciate it if u could check out this pr of mine. its been out for a while and might face merged conflicts soon.

https://github.com/rust-lang/rust-analyzer/pull/21864

Albab-Hasan:
@ChayimFriedman2 hi, can i get a review on this?

PR #21912, opened 2026-03-30
feat: offer on tail-expr with else-branch for if_let_to_guarded assist

A4-Tacks:
For assist 'convert_to_guarded_return'

Example
---
```rust
fn main() -> i32 {
    if$0 true {
        foo();
    } else {
        bar()
    }
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
fn main() -> i32 {
    if false {
        return bar();
    }
    foo();
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/convert_to_guarded_return.rs
+++ b/crates/ide-assists/src/handlers/convert_to_guarded_return.rs
@@ -1068,6 +1068,37 @@ fn main() {
     }
 
     #[test]
+    fn convert_let_inside_for_with_else() {
+        check_assist(
+            convert_to_guarded_return,
+            r#"
+fn main() {
+    for n in ns {
+        if$0 let Some(n) = n {
+            foo(n);
+            bar();
+        } else {
+            baz()
+        }
+    }
+}
+"#,
+            r#"
+fn main() {
+    for n in ns {
+        let Some(n) = n else {
+            baz();
+            continue
+        };
+        foo(n);
+        bar();
+    }
+}
+"#,
+        );
+    }
+
+    #[test]
     fn convert_let_stmt_inside_fn() {
         check_assist(
             convert_to_guarded_return,


crates/ide-assists/src/handlers/convert_to_guarded_return.rs
```diff
@@ -239,33 +230,119 @@ fn let_stmt_to_guarded_return(
...
+        };
+        let whitespace = last_element.prev_sibling_or_token().filter(|it| it.kind() == WHITESPACE);
+
+        let make = SyntaxFactory::without_mappings();
+        let mut edit = SyntaxEditor::new(block_expr.syntax().clone());
+
+        if let Some(tail_expr) = block_expr.tail_expr()
+            && !self.kind.is_unit()
+        {
+            let early_expr = self.kind.make_early_expr(sema, &make, Some(tail_expr.clone()));
```
ChayimFriedman2, line 278: Looks like this can lose the tail expr if inside a loop, where it's `continue`.

A4-Tacks: Due to `!self.kind.is_unit()`, this will not allow `continue`

crates/ide-assists/src/handlers/convert_to_guarded_return.rs
```diff
@@ -239,33 +230,119 @@ fn let_stmt_to_guarded_return(
...
+
+        if self.is_never_block {
+            return block_expr.reset_indent();
+        }
+
+        let block_expr = block_expr.reset_indent().clone_subtree();
+        let last_stmt = block_expr.statements().last().map(|it| it.syntax().clone());
+        let tail_expr = block_expr.tail_expr().map(|it| it.syntax().clone());
+        let Some(last_element) = tail_expr.clone().or(last_stmt.clone()) else {
+            return make.tail_only_block_expr(self.kind.make_early_expr(sema, make, None));
```
ChayimFriedman2, line 268: Sorry I was pointing at the wrong place, I meant here 😅

A4-Tacks: Is there any problem here? For empty blocks, use a new early block instead

ChayimFriedman2: Oh right didn't notice the block is empty here 👍

PR #21913, opened 2026-03-30
Fix intra-doc link

Shourya742:
CI on master is failing because of doc linking, this PR fixes that. More info here: https://github.com/rust-lang/rust-analyzer/actions/runs/23736993343/job/69144103755


ChayimFriedman2:
Thanks!

PR #21919, opened 2026-03-31
feat: support labeled block for convert_to_guarded_return

A4-Tacks:
Example
---
```rust
fn main() {
    'l: {
        if$0 let Some(n) = n {
            foo(n);
            bar();
        }
    }
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
fn main() {
    'l: {
        let Some(n) = n else { break 'l };
        foo(n);
        bar();
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/convert_to_guarded_return.rs
+++ b/crates/ide-assists/src/handlers/convert_to_guarded_return.rs
@@ -101,8 +101,8 @@ fn if_expr_to_guarded_return(
         return None;
     }
 
-    let parent_container = parent_block.syntax().parent()?;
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
I'm not entirely sure this is better, but I'm neutral about this, so, so be it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21919/63b3eff38ef1c216480147dd53b0e4365d55f269..8dd9f6ce3c009463348dc4c852e8225af812f7a4)*

PR #21920, opened 2026-03-31
fix: Improve label on add_missing_match_arms assist

Wilfred:
"Fill match arms" is an extremely helpful assist, but the name is pretty opaque to new users (at least it was to me). We actually renamed the file in rust-lang/rust-analyzer#10299 from `fill_match_arms.rs` to `add_missing_match_arms.rs` but the label hasn't changed from its original text in rust-lang/rust-analyzer#733.

Instead, show "Add missing match arms" if there are multiple missing match arms, and show "Add missing match arm `Foo::Bar`" when there's only a single missing match arm.

Partially generated by Claude Opus.

### DIFF
--- a/crates/ide-assists/src/handlers/add_missing_match_arms.rs
+++ b/crates/ide-assists/src/handlers/add_missing_match_arms.rs
@@ -1,4 +1,4 @@
-use std::iter::{self, Peekable};
+use std::iter;
 
 use either::Either;
 use hir::{Adt, AsAssocItem, Crate, FindPathConfig, HasAttrs, ModuleDef, Semantics};
@@ -93,8 +93,8 @@ pub(crate) fn add_missing_match_arms(acc: &mut Assists, ctx: &AssistContext<'_>)
     } else {
         None
     };
-    let (mut missing_pats, is_non_exhaustive, has_hidden_variants): (
-        Peekable<Box<dyn Iterator<Item = (ast::Pat, bool)>>>,
+    let (missing_pats, is_non_exhaustive, has_hidden_variants): (
+        Vec<(ast::Pat, bool)>,
         bool,
         bool,
     ) = if let Some(enum_def) = resolve_enum_def(&ctx.sema, &expr, self_ty.as_ref()) {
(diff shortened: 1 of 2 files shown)


crates/ide-assists/src/handlers/add_missing_match_arms.rs
```diff
@@ -223,9 +223,17 @@ pub(crate) fn add_missing_match_arms(acc: &mut Assists, ctx: &AssistContext<'_>)
         return None;
     }
 
+    let first = missing_pats.next();
```
ChayimFriedman2: I prefer to have `missing_pats: Vec<_>` by collecting beforehand.

crates/ide-assists/src/handlers/add_missing_match_arms.rs
```diff
@@ -223,9 +223,17 @@ pub(crate) fn add_missing_match_arms(acc: &mut Assists, ctx: &AssistContext<'_>)
...
+    let label = match (&first, missing_pats.peek()) {
+        (Some((pat, false)), None) => format!("Add missing match arm `{pat}`"),
+        (Some((_, true)), None) => "Add missing match arm `_`".to_owned(),
```
ChayimFriedman2: I think this arm will never hit. Also we may filter hidden patterns (that we do after) before this.

ChayimFriedman2 (changes requested):

crates/ide-assists/src/handlers/add_missing_match_arms.rs
Wilfred: Funny, that's what my first attempt did! The other nice thing about collecting is that I can show the exact number of missing arms in the label. 

My concern was that most assists are deliberately lazy, hence me trying to just consume a bit of the iterator. What do you think?

ChayimFriedman2: This allocates anyway, so it doesn't really matter IMO.

Wilfred:
Rebased, changed to use .collect() and show the exact number of missing arms, added test for the catch-all case.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21920/12e079ec6dd0b375f8adad6e996c676b12a0153d..fd92d9d39338fbd33af4865c65fb3dfb87f1b530)*

PR #21926, opened 2026-04-01
fix: Fix param inlayHints on empty expr and comma

A4-Tacks:
Example
---
```rust
pub fn test(a: i32, b: i32, c: i32) {}
fn main() {
    test(, 2,);
    test(, , 3);
}
```

**Before this PR**

```rust
    test(, 2,);
         //^ a
    test(, , 3);
           //^ a
```

**After this PR**

```rust
    test(, 2,);
         //^ b
    test(, , 3);
           //^ c
```

### DIFF
--- a/crates/ide/src/inlay_hints/param_name.rs
+++ b/crates/ide/src/inlay_hints/param_name.rs
@@ -37,8 +37,9 @@ pub(super) fn hints(
     let hints = callable
         .params()
         .into_iter()
-        .zip(arg_list.args())
+        .zip(arg_list.args_maybe_empty())
         .filter_map(|(p, arg)| {
+            let arg = arg?;
             // Only annotate hints for expressions that exist in the original file
             let range = sema.original_range_opt(arg.syntax())?;
             if range.file_id != file_id {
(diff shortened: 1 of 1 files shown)


crates/ide-db/src/syntax_helpers/node_ext.rs
```diff
@@ -579,3 +580,25 @@ pub fn is_in_macro_matcher(token: &SyntaxToken) -> bool {
             && tt.syntax().text_range().contains_range(range)
     })
 }
+
+/// Comma separated args, argument may be empty
+pub fn split_args(args: &ast::ArgList) -> impl Iterator<Item = Option<ast::Expr>> {
```
ChayimFriedman2: I prefer this to be a method in `syntax`.

ChayimFriedman2: Also a more descriptive name, like `args_maybe_empty()`.

A4-Tacks: ~~I don't think we should accept `&SyntaxNode` because he only handles `ArgList`~~

Oh, it's `syntax` crate

ChayimFriedman2 (approved):

PR #21927, opened 2026-04-01
fix: Not suggest name in nested type in variant

A4-Tacks:
Partial of rust-lang/rust-analyzer#21892

Example
---
```rust
struct Other;
struct Vec<T>(T);
enum Foo {
    Vec(Vec<$0>)
}
```

**Before this PR**

```text
st Vec<…> Vec<{unknown}> [name]
en Foo Foo []
st Other Other []
sp Self Foo []
```

**After this PR**

```text
en Foo Foo []
st Other Other []
sp Self Foo []
st Vec<…> Vec<{unknown}> []
```

### DIFF
--- a/crates/ide-completion/src/context/analysis.rs
+++ b/crates/ide-completion/src/context/analysis.rs
@@ -828,10 +828,15 @@ fn expected_type_and_name<'db>(
                         .unwrap_or((None, None))
                 },
                 ast::Variant(it) => {
+                    let is_simple_field = |field: ast::TupleField| {
+                        let Some(ty) = field.ty() else { return true };
+                        matches!(ty, ast::Type::PathType(_)) && ty.generic_arg_list().is_none()
+                    };
                     let is_simple_variant = matches!(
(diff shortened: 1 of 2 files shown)


crates/ide-completion/src/context/analysis.rs
```diff
@@ -828,10 +828,15 @@ fn expected_type_and_name<'db>(
                         .unwrap_or((None, None))
                 },
                 ast::Variant(it) => {
+                    let is_simple_field = |field: ast::TupleField| {
+                        let Some(ty) = field.ty() else { return true };
+                        matches!(ty, ast::Type::PathType(_)) && ty.generic_arg_list().is_none()
```
ChayimFriedman2, line 833: This will mean we'll also not suggest on `Foo($0<>)`; do you think we should suggest on that?

A4-Tacks: I don't think it's necessary, this kind of completion is rare

PR #21928, opened 2026-04-01
feat: support macro expansion in `#[doc = ...]` attributes

so1ve:
fixes rust-lang/rust-analyzer#7688, fixes rust-lang/rust-analyzer#8092, fixes rust-lang/rust-analyzer#11137, fixes rust-lang/rust-analyzer#20494, closes rust-lang/rust-analyzer#21303

Macro expansion part is hand-written; AI (specifically, GPT 5.4) was used to generate tests.

### DIFF
--- a/crates/hir-def/src/attrs/docs.rs
+++ b/crates/hir-def/src/attrs/docs.rs
@@ -455,9 +455,7 @@ fn extend_with_attrs<'a, 'db>(
                         // expansion (see FIXME above).
                         let is_from_cfg_attr =
                             top_attr.as_simple_call().is_some_and(|(name, _)| name == "cfg_attr");
-                        if !is_from_cfg_attr
-                            && let Some(expr) = top_attr.expr()
-                        {
+                        if !is_from_cfg_attr && let Some(expr) = top_attr.expr() {
                             let (exp, ctx) = expander.get_or_insert_with(|| {
                                 let resolver = make_resolver();
                                 let def_map = resolver.top_level_def_map();


ChayimFriedman2:
I cannot vouch for which is better but note that #21303 also implements that however by text-matching on `include_str!()` and supporting it alone. Does this PR does the same, or does it fully expand all macros?

so1ve:
Yes, this PR uses rust-analyzer's internal macro expansion mechanism, enabling all macro expansion within `#[doc]` attributes, instead of special-casing include_str. See [this test case](https://github.com/so1ve/rust-analyzer/blob/015ee833890256dfd54d7bd5ff21b14f7d60a9fb/crates/ide/src/hover/tests.rs#L11524-L11559) for details.

ChayimFriedman2:
Then this PR is preferred.

crates/hir-def/src/attrs.rs
```diff
@@ -652,6 +682,21 @@ impl Docs {
         }
     }
 
+    fn extend_with_unmapped_doc_str(&mut self, doc: &str, indent: &mut usize) {
+        for line in doc.split('\n') {
```
ChayimFriedman2: This has a potential for code sharing with above.

crates/hir-def/src/attrs.rs
```diff
@@ -757,6 +804,20 @@ pub struct DeriveInfo {
     pub helpers: Box<[Symbol]>,
 }
 
+struct DocMacroExpander<'db> {
```
ChayimFriedman2: The docs stuff is becoming big enough to extract it to a submodule.

crates/hir-def/src/attrs.rs
```diff
@@ -398,6 +403,28 @@ fn attrs_source(
     (owner, None, None, krate)
 }
 
+fn resolver_for_attr_def_id(db: &dyn DefDatabase, owner: AttrDefId) -> Resolver<'_> {
+    match owner {
+        AttrDefId::ModuleId(id) => id.resolver(db),
```
ChayimFriedman2, line 400: A module can have two resolvers, depending on whether the docs are inner or outer (where outer, this is the resolver of the parent module). So you need to account for that. Also please add a test for that.

crates/hir-def/src/attrs.rs
```diff
@@ -785,7 +846,128 @@ fn extract_cfgs(result: &mut Vec<CfgExpr>, attr: Meta) -> ControlFlow<Infallible
...
+                        if name.text() == "doc" =>
+                    {
+                        result.extend_with_doc_attr(value, indent);
+                    }
+                    Meta::NamedKeyValue { name: Some(name), value: None, .. }
+                        if name.text() == "doc" =>
+                    {
+                        if let (Some(expander), Some(source_ctx)) =
+                            (expander.as_deref_mut(), source_ctx)
+                            && let Some(expr) = top_attr.expr()
```
ChayimFriedman2: This is technically not correct in the presence of `cfg_attr`. If you do not want to handle that, fine, but leave a FIXME. Otherwise, you will need to change `expand_cfg_attr()`.

crates/hir-def/src/attrs.rs
```diff
@@ -1292,10 +1494,25 @@ impl AttrFlags {
     pub fn docs(db: &dyn DefDatabase, owner: AttrDefId) -> Option<Box<Docs>> {
         let (source, outer_mod_decl, _extra_crate_attrs, krate) = attrs_source(db, owner);
         let inner_attrs_node = source.value.inner_attributes_node();
+        let resolver = resolver_for_attr_def_id(db, owner);
```
ChayimFriedman2: Constructing a resolver is pretty expensive, so it's better to do it lazily.

crates/hir-def/src/attrs.rs
```diff
@@ -1292,10 +1494,25 @@ impl AttrFlags {
...
+        let def_map = crate_def_map(db, krate);
```
ChayimFriedman2:
```suggestion
        let def_map = resolver.top_level_def_map();
```

crates/hir-def/src/attrs.rs
```diff
@@ -1292,10 +1494,25 @@ impl AttrFlags {
...
+        let recursion_limit = if cfg!(test) {
+            std::cmp::min(32, def_map.recursion_limit() as usize)
+        } else {
+            def_map.recursion_limit() as usize
+        };
```
ChayimFriedman2: No need for that.

crates/hir-def/src/attrs.rs
so1ve: Oh I was planning to implement cfg_attr expansion, but noticed ongoing discussions about the default behavior (rust-lang/rust-analyzer#18444)

ChayimFriedman2: That issue is completely unrelated.

crates/ide/src/hover/tests.rs
```diff
@@ -11404,3 +11404,333 @@ pub trait MyTrait {
...
+
+            plain outer doc
+            expanded from parent
+            inner module docs
+        "#]],
+    );
+}
+
+#[test]
+fn test_hover_doc_attr_macro_on_outlined_mod_combined_with_inner_docs() {
```
ChayimFriedman2: This test is the same as above. Also you don't check inner docs.

so1ve:
Just applied the suggestions!

crates/hir-def/src/attrs/docs.rs
```diff
@@ -0,0 +1,774 @@
...
+                            result.extend_with_unmapped_doc_str(&expanded, indent);
+                        }
+                    }
+                    _ => {}
+                },
+            }
+            ControlFlow::Continue(())
+        },
+    );
+}
```
ChayimFriedman2, line 491:
```suggestion
    outer_resolver: &dyn Fn() -> Option<Resolver<'db>>,
```

crates/hir-def/src/attrs/docs.rs
```diff
@@ -0,0 +1,774 @@
...
+
+pub(crate) fn extract_docs<'a, 'db>(
+    db: &'db dyn DefDatabase,
+    krate: Crate,
+    // For outer docs on an outlined module, use the parent module's resolver.
+    // For inline docs (and non-module items), use the item's own resolver.
+    outer_resolver: Option<impl FnOnce() -> Resolver<'db>>,
+    inline_resolver: impl FnOnce() -> Resolver<'db>,
```
ChayimFriedman2:
```suggestion
    inline_resolver: &dyn Fn() -> Resolver<'db>,
```
Monomorphization is not helpful here.

crates/hir-def/src/attrs.rs
```diff
@@ -1292,10 +965,30 @@ impl AttrFlags {
...
+            if let AttrDefId::ModuleId(module_id) = owner {
+                module_id
+                    .containing_module(db)
+                    .map(|parent| move || -> Resolver<'_> { parent.resolver(db) })
+            } else {
+                None
+            }
+        } else {
+            None
+        };
```
ChayimFriedman2:
```suggestion
        let parent = if outer_mod_decl.is_some() && AttrDefId::ModuleId(module_id) = owner {
            module_id.containing_module(db)
        } else {
            None
        };
        let outer_resolver = || parent.map(|it| it.resolver(db)); 
```

crates/hir-def/src/attrs/docs.rs
ChayimFriedman2: Or even:
```rust
resolvers: &dyn Fn() -> (Option<Resolver<'db>>, Resolver<'db>)
```

crates/hir-def/src/attrs/docs.rs
```diff
@@ -0,0 +1,774 @@
...
+        result.outline_mod = Some((outer_mod_decl.file_id, result.docs_source_map.len()));
+    }
+
+    let inline_source_map_start = result.docs_source_map.len();
+    let mut indent = usize::MAX;
+    // For inline docs, use the item's own resolver.
+    let needs_expansion = has_doc_macro_attr(source.value.syntax())
+        || inner_attrs_node.as_ref().is_some_and(has_doc_macro_attr);
+    let (mut inline_expander, inline_source_ctx) = if needs_expansion {
+        let resolver = inline_resolver();
```
ChayimFriedman2: Potential for code reuse with above here.

crates/hir-def/src/attrs/docs.rs
```diff
@@ -0,0 +1,774 @@
...
+        &mut result,
+        source.value.syntax(),
+        false,
+        &mut indent,
+        get_cfg_options,
+        &mut cfg_options,
+        inline_expander.as_mut(),
+        inline_source_ctx.as_ref(),
+    );
+    if let Some(inner_attrs_node) = &inner_attrs_node {
```
ChayimFriedman2, line 548: You can `.chain()` it like we do for other kind of attributes, it's shorter.

crates/hir-def/src/attrs/docs.rs
```diff
@@ -0,0 +1,774 @@
...
+                        // `top_attr` points to the `cfg_attr(...)` node, not the
+                        // inner `doc = macro!()`.  In that case `top_attr.expr()`
+                        // would not yield the macro expression we need, so skip
+                        // expansion (see FIXME above).
+                        let is_from_cfg_attr =
+                            top_attr.as_simple_call().is_some_and(|(name, _)| name == "cfg_attr");
+                        if !is_from_cfg_attr
+                            && let (Some(expander), Some(source_ctx)) =
+                                (expander.as_deref_mut(), source_ctx)
+                            && let Some(expr) = top_attr.expr()
```
ChayimFriedman2: This still isn't good for `cfg_attr`. The problem is this line. For `cfg_attr`, `top_attr` is the `cfg_attr`, therefore it won't have `expr()`. Fixing this is difficult as we need an `Expr` that doesn't exist here for the ast id and for sanely parsing the macro call.

crates/hir-def/src/attrs/docs.rs
```diff
@@ -0,0 +1,774 @@
...
+        file_id: expansion_file_id,
+        ast_id_map: expander.db.ast_id_map(expansion_file_id),
+        span_map: expander.db.span_map(expansion_file_id),
+    };
+    Some((expr, new_source_ctx))
+}
+
+/// Quick check: does this syntax node have any `#[doc = expr]` attributes where the
+/// value is not a simple string literal (i.e., it needs macro expansion)?
+fn has_doc_macro_attr(node: &SyntaxNode) -> bool {
```
ChayimFriedman2: This is not needed, you can just make the resolver in `extend_with_attrs()`.

ChayimFriedman2 (changes requested):

so1ve:
Updated!

asukaminato0721:
amazing!

lnicola:
Unlike #21303, this didn't add `.md` to the included extensions, so many of the linked issues and even the test won't work.

Should we bite the bullet and load those into the VFS?

Veykril:
Yea I think we should

ChayimFriedman2:
There aren't many markdown files usually so I believe it shouldn't cause perf problems.

lnicola:
@asukaminato0721 want to do the honors? :)

so1ve:
Ah, I never considered that the expansion of `include_str` requires the use of the VFS 😅 I was misled by the test, it seems that `include_str` in the test can run normally

ChayimFriedman2:
Tests always includes the specified files in the VFS.

asukaminato0721:
> @asukaminato0721 want to do the honors? :)

sure :)

joshka:
This does not appear to actually work (for the axum source code at least).

It does not appear to have fixed the problem:

Installed version:
Image

https://github.com/tokio-rs/axum/blob/441216428893d13544f12722b54dcaaadd47135a/axum/src/error_handling/mod.rs#L1
Module level doc:
Image

Function level doc:
https://github.com/tokio-rs/axum/blob/441216428893d13544f12722b54dcaaadd47135a/axum/src/routing/method_routing.rs#L1002
Image

I'd recommend re-opening at least one of the mentioned issues to track this (and perhaps consolidating on one of them?)

ChayimFriedman2:
@joshka As explained literally just above, the expansion does work but rust-analyzer in its current version does not load md files. If you'll make the file extension `.rs` it'll work. This will be fixed by the next release.

joshka:
Ah, I see now. I missed that the VFS piece meant markdown wouldn’t work yet. Thanks for clarifying.

PR #21931, opened 2026-04-02
Move mutability responsibility from caller to edit_algo

bit-aloo:
This PR moves responsibility from the caller to the edit pipeline. The goal is to eliminate all `clone_subtree` and `clone_for_update` calls at the caller site and handle them within the pipeline. I may still be overlooking some edge cases, but this approach feels like a more appropriate way to manage mutability.

### DIFF
--- a/crates/syntax/src/syntax_editor/edit_algo.rs
+++ b/crates/syntax/src/syntax_editor/edit_algo.rs
@@ -192,11 +192,8 @@ pub(super) fn apply_edits(editor: SyntaxEditor) -> SyntaxEdit {
                     }
                 };
             }
-            Change::Replace(SyntaxElement::Node(target), Some(SyntaxElement::Node(new_target))) => {
+            Change::Replace(SyntaxElement::Node(target), Some(SyntaxElement::Node(_))) => {
                 *target = tree_mutator.make_syntax_mut(target);
-                if new_target.ancestors().any(|node| node == tree_mutator.immutable) {
-                    *new_target = new_target.clone_for_update();
-                }
             }
             Change::Replace(target, _) | Change::ReplaceWithMany(target, _) => {
                 *target = tree_mutator.make_element_mut(target);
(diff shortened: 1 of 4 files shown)


bit-aloo:
@A4-Tacks would appreciate your thoughts on this idea?

A4-Tacks:
I am not familiar with `SyntaxEditor` internal impl, you can have someone else review this

crates/syntax/src/syntax_editor/edit_algo.rs
```diff
@@ -209,6 +206,35 @@ pub(super) fn apply_edits(editor: SyntaxEditor) -> SyntaxEdit {
             }
         }
 
+        match &mut changes[index as usize] {
+            Change::Insert(_, SyntaxElement::Node(node))
+            | Change::Replace(_, Some(SyntaxElement::Node(node))) => {
+                if node.parent().is_some() {
+                    *node = node.clone_subtree().clone_for_update();
+                } else {
+                    if node.ancestors().any(|im_node| im_node == tree_mutator.immutable) {
```
ChayimFriedman2: If `!node.parent.is_some()` then `node.ancestors()` contains this node alone.

crates/syntax/src/syntax_editor/edit_algo.rs
```diff
@@ -209,6 +206,35 @@ pub(super) fn apply_edits(editor: SyntaxEditor) -> SyntaxEdit {
...
+                        *node = node.clone_for_update();
+                    }
+                }
+            }
+            Change::InsertAll(_, elements)
+            | Change::ReplaceWithMany(_, elements)
+            | Change::ReplaceAll(_, elements) => {
+                for element in elements.iter_mut() {
```
ChayimFriedman2:
```suggestion
                for element in elements {
```

ChayimFriedman2 (commented):
I agree, this is a better approach.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21931/72fdd72d6cf1a5532f82310ec1bc39d80d1ed56d..5b5aa91b9c252f3f1925d1d43d82410941caf8ef)*

crates/syntax/src/syntax_editor/edit_algo.rs
bit-aloo: Ah, my bad. Thanks

This check is quite limiting. It only considers nodes that is the syntax tree the editor is based on. Ideally, we should also make newly created nodes, those not part of the original syntax tree mutable at this stage. If we can expose `mutable` from `NodeData` in Rowan, we could eliminate most `clone_for_update` calls from the constructor layer (and elsewhere), and instead confine that responsibility to the edit flow.

bit-aloo: Can we do a patch release to rowan to just expose this `mutable` field from `NodeData`?.

ChayimFriedman2: > Can we do a patch release to rowan to just expose this mutable field from NodeData?

We can, but how will this help?

A4-Tacks: Rowan's 0.16.1 has a `SyntaxNode::is_mutable` now, but r-a still uses 0.15.17

bit-aloo: The only reason we can’t blindly call `clone_for_update` on nodes without a parent is that it is not idempotent. Attempting to make an already mutable node mutable again triggers a panic due to the runtime assertion `assert!(!node.mutable())`. If we had a way to determine whether a node is already mutable, we could safely avoid these panics.

ChayimFriedman2: Okay, published rowan v0.15.18 with `is_mutable()`.

bit-aloo: Thanks @ChayimFriedman2, this should be good now. We’re no longer performing double mutation. This will also lead to a fair amount of cleanup; I’ll be opening additional PRs soon.

ChayimFriedman2 (approved):

PR #21936, opened 2026-04-03
fix: Fix extract variable on arg with comma

A4-Tacks:
Example
---
```rust
fn main() {
    let x = 2;
    foo(
        x + x,
        $0x - x,$0
    )
}
```

**Before this PR**

```rust
fn main() {
    let x = 2;
    let $0var_name = x + x;
    foo(
        var_name,
        x - x,
    )
}
```

**After this PR**

```rust
fn main() {
    let x = 2;
    let $0var_name = x - x;
    foo(
        x + x,
        var_name,
    )
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -283,13 +283,17 @@ fn peel_parens(mut expr: ast::Expr) -> ast::Expr {
 /// Check whether the node is a valid expression which can be extracted to a variable.
 /// In general that's true for any expression, but in some cases that would produce invalid code.
 fn valid_target_expr(ctx: &AssistContext<'_>) -> impl Fn(SyntaxNode) -> Option<ast::Expr> {
-    |node| match node.kind() {
+    let selection = ctx.selection_trimmed();
+    move |node| match node.kind() {
         SyntaxKind::LOOP_EXPR | SyntaxKind::LET_EXPR => None,
         SyntaxKind::BREAK_EXPR => ast::BreakExpr::cast(node).and_then(|e| e.expr()),
         SyntaxKind::RETURN_EXPR => ast::ReturnExpr::cast(node).and_then(|e| e.expr()),
         SyntaxKind::BLOCK_EXPR => {
             ast::BlockExpr::cast(node).filter(|it| it.is_standalone()).map(ast::Expr::from)
         }
+        SyntaxKind::ARG_LIST => ast::ArgList::cast(node)?
+            .args()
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -283,13 +283,17 @@ fn peel_parens(mut expr: ast::Expr) -> ast::Expr {
...
-    |node| match node.kind() {
+    let selection = ctx.selection_trimmed();
+    move |node| match node.kind() {
         SyntaxKind::LOOP_EXPR | SyntaxKind::LET_EXPR => None,
         SyntaxKind::BREAK_EXPR => ast::BreakExpr::cast(node).and_then(|e| e.expr()),
         SyntaxKind::RETURN_EXPR => ast::ReturnExpr::cast(node).and_then(|e| e.expr()),
         SyntaxKind::BLOCK_EXPR => {
             ast::BlockExpr::cast(node).filter(|it| it.is_standalone()).map(ast::Expr::from)
         }
+        SyntaxKind::ARG_LIST => ast::ArgList::cast(node)?
```
ChayimFriedman2, line 294: This will also pass when multiple args are selected. Also `ARG_LIST` isn't the only place for things like that (e.g. `ARRAY_EXPR` is too). IMO we should disable the assist when the comma is selected, but I also agree to enable it if it's confined to where only a single item is selected.

A4-Tacks: > IMO we should disable the assist when the comma is selected

When multiple parameters are selected, the old behavior will also pass, and I don't think this will break some habits

A4-Tacks: > e.g. `ARRAY_EXPR` is too

No, `ARRAY_EXPR` is quite normal. Its behavior is to extract the entire array

Need to determine if the selected single expression should be an array, not an arglist?

ChayimFriedman2: > When multiple parameters are selected, the old behavior will also pass, and I don't think this will break some habits

True, but it's still incorrect.

A4-Tacks: It seems difficult, if multiple args are selected in arglist and `valid_target_expr` returns `None`, it will degenerate into its original behavior because `valid_target_expr` is applied to `node.descendants()`

ChayimFriedman2: Fine then.

PR #21938, opened 2026-04-03
fix: Fix indent for convert_let_else_to_match

A4-Tacks:
Example
---
```rust
mod indent {
    fn foo() {
        let Ok(x) = f() else$0 {
            log();
            unreachable!(
                "..."
            );
        };
    }
}
```

**Before this PR**

```rust
mod indent {
    fn foo() {
        let x = match f() {
            Ok(x) => x,
            _ => {
                    log();
                    unreachable!(
                        "..."
                    );
                }
        };
    }
}
```

**After this PR**

```rust
mod indent {
    fn foo() {
        let x = match f() {
            Ok(x) => x,
            _ => {
                log();
                unreachable!(
                    "..."
                );
            }
        };
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/convert_let_else_to_match.rs
+++ b/crates/ide-assists/src/handlers/convert_let_else_to_match.rs
@@ -33,9 +33,9 @@ pub(crate) fn convert_let_else_to_match(acc: &mut Assists, ctx: &AssistContext<'
     let let_stmt = LetStmt::cast(let_stmt)?;
     let else_block = let_stmt.let_else()?.block_expr()?;
     let else_expr = if else_block.statements().next().is_none() {
-        else_block.tail_expr()?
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
I have in the past expressed sentiments that dislike PRs that fix indentation, and AFAIK @Veykril agrees with me. However this PR also adds mutable tree editing for an assist that has been already migrated, therefore I'm closing it on that basis.

A4-Tacks:
> However this PR also adds mutable tree editing

Didn't I use `ast::edit` instead of `ast::edit_in_place`?

ChayimFriedman2:
Actually yes, my mistake. I'll approve this, but please don't make more such PRs at least until we finish the migration. Their benefit is doubtful and these APIs may also need a change.

PR #21940, opened 2026-04-03
fix: support multiple snippet placeholders in VS Code extension

Amit5601:
Fixes rust-lang/rust-analyzer#12312.
Added a missing escaped dollar sign (`\$`) to the regex in `hasSnippet` within `editors/code/src/snippets.ts`. The previous regex `\{\d+:[^}]*\}` failed to match VS Code's official placeholder syntax, which requires a leading dollar sign (e.g., `${1:foo}`). Because of this, assists returning multiple placeholders (like "Generate Documentation") were failing the `hasSnippet` check and being inserted as raw text.

### DIFF
--- a/editors/code/src/snippets.ts
+++ b/editors/code/src/snippets.ts
@@ -53,7 +53,7 @@ export async function applySnippetTextEdits(editor: vscode.TextEditor, edits: vs
 }
 
 function hasSnippet(snip: string): boolean {
-    const m = snip.match(/\$\d+|\{\d+:[^}]*\}/);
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21940/f4b77dc99d9925667246e2887783b79bdc46a50d..f182e0f0fc64e29aa293d88af2071ba860bc68c1)*

PR #21941, opened 2026-04-03
internal: extract should_count_for_attr_id predicate for AttrId counter sync

Vastargazing:
Closes rust-lang/rust-analyzer#21939.

`collect_item_tree_attrs()` and `macro_input_callback()` must make identical decisions about which attributes increment `item_tree_attr_id`, but that logic was duplicated in two places. `cfg` is handled via a separate special-case outside `is_item_tree_filtered_attr()`, so the invariant was easy to desynchronize — and already did (PR rust-lang/rust-analyzer#21205).

This PR extracts a single predicate `should_count_for_attr_id()` that both functions use, making the invariant explicit in one place.

## What changed

- Added `should_count_for_attr_id()` in `attrs.rs` with a doc-comment stating the invariant
- Refactored `collect_item_tree_attrs` to use it (replaces the `let filter = match` block)
- Refactored `macro_input_callback` to use it (36 lines → 18)
- No behaviour change — pure refactor

## AI disclosure

I used AI-assisted tooling during investigation to help navigate and cross-reference paths and prior changes across a large surface area. The bug analysis, invariant identification, proposed refactor, and final patch are my own.

### DIFF
--- a/crates/hir-expand/src/attrs.rs
+++ b/crates/hir-expand/src/attrs.rs
@@ -390,15 +390,14 @@ pub fn collect_item_tree_attrs<'a, BreakValue>(
             if let Meta::TokenTree { path, tt } = &attr
                 && path.is1("cfg")
             {
-                let cfg =
-                    CfgExpr::parse_from_ast(&mut TokenTreeChildren::new(tt).peekable());
+                let cfg = CfgExpr::parse_from_ast(&mut TokenTreeChildren::new(tt).peekable());
                 if cfg_options().check(&cfg) == Some(false) {
                     return ControlFlow::Break(Either::Right(cfg));
                 }
-            } else if should_count_for_attr_id(&attr) {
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
This does not help synchronization in my eyes. Also I suspect that despite what you said, the patch was generated by a LLM. The PR text certainly was. Please respect the maintainers, if you repeat this we'll reach out to moderation.

Vastargazing:
> This does not help synchronization in my eyes. Also I suspect that despite what you said, the patch was generated by a LLM. The PR text certainly was. Please respect the maintainers, if you repeat this we'll reach out to moderation.

I'm a Russian speaker — I used AI assistance for translating and structuring 
the English text, not for generating the patch. I should have been clearer 
about that in the disclosure. Understood, closing. Thank you for the feedback.

PR #21942, opened 2026-04-03
fix: silence type mismatch diagnostic when type is unknown

Amit5601:
Fixes rust-lang/rust-analyzer#21512.
Silences the unhelpful `expected {unknown}, found...` secondary diagnostic that triggers when type inference has already failed (e.g., from an unsatisfied trait bound). `type_mismatch` now returns an `Option<Diagnostic>` and returns `None` if either the expected or actual type `is_unknown()`.


ChayimFriedman2 (approved):
It is perhaps better to filter them at creation time, but I can't decide, so this is also fine.

Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21942/5f147ccdf270e8c9cc4f1e8b7cc662fb3c47b99e..dc20744592e40392b1d9e33e54babe92588f55fd)*

PR #21947, opened 2026-04-03
minor: update generated lints

dybucc:
I noticed rust-lang/rust-analyzer#11928 referred to these issues, so I got a CI job going to be run a
bit like the nightly channel on a daily cron job, which should at least get out
the need for somebody to update it. If this is viable, I can open a PR with the
workflow file.


ChayimFriedman2:
I think once a day will be too burdensome, but we can do that once a week.

dybucc:
Just opened a PR with the workflow set for weekly runs on Sundays. I've
personally seen daily updates actually refreshing the list, so maybe the load's
worth it?

PR #21951, opened 2026-04-04
fix: report `expected type, found {` in parser

Amit5601:
Fixes rust-lang/rust-analyzer#19858.

Improves the parse error message when a developer forgets the `->` before a function body block. Added a specific match arm for the `{` token in `type_with_bounds_cond` to report `"expected type, found \`{\`"` instead of the generic `"expected type"` fallback.


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/21951/f4b77dc99d9925667246e2887783b79bdc46a50d..77f0ee37481e202e858809343dfcf758257fcb17)*

PR #21952, opened 2026-04-04
fix: offer 'type_mismatch' some fixes inside macro

A4-Tacks:
- Supports macro for `add_missing_ok_or_some` and `str_ref_to_owned`

Example
---
```rust
macro_rules! identity { ($($t:tt)*) => ($($t)*) }
identity! {
    fn test() -> String {
        "a"$0
    }
}
```

**Before this PR**

Invalid trigger range and edit range

**After this PR**

```rust
macro_rules! identity { ($($t:tt)*) => ($($t)*) }
identity! {
    fn test() -> String {
        "a".to_owned()
    }
}
```

---

```rust
macro_rules! identity { ($($t:tt)*) => ($($t)*) }
identity! {
    fn div(x: i32, y: i32) -> Result<i32, ()> {
        if y == 0 { return Err(()); }
        x / y$0
    }
}
```

**Before this PR**

Invalid trigger range and edit range

**After this PR**

```rust
macro_rules! identity { ($($t:tt)*) => ($($t)*) }
identity! {
    fn div(x: i32, y: i32) -> Result<i32, ()> {
        if y == 0 { return Err(()); }
        Ok(x / y)
    }
}
```

### DIFF
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -101,7 +101,7 @@ fn add_missing_ok_or_some(
 ) -> Option<()> {
     let root = ctx.sema.db.parse_or_expand(expr_ptr.file_id);
     let expr = expr_ptr.value.to_node(&root);
-    let expr_range = expr.syntax().text_range();
+    let expr_range = ctx.sema.original_range_opt(expr.syntax())?.range;
     let scope = ctx.sema.scope(expr.syntax())?;
 
     let expected_adt = d.expected.as_adt()?;
@@ -135,13 +135,13 @@ fn add_missing_ok_or_some(
                     // Empty block
                     let indent = block_indent + 1;
                     builder.insert(
-                        block.syntax().text_range().start() + TextSize::from(1),
+                        expr_range.start() + TextSize::from(1),
                         format!("\n{indent}{variant_name}(())\n{block_indent}"),
                     );
                 } else {
                     let indent = IndentLevel::from(1);
                     builder.insert(
-                        block.syntax().text_range().end() - TextSize::from(1),
+                        expr_range.end() - TextSize::from(1),
                         format!("{indent}{variant_name}(())\n{block_indent}"),
                     );
                 }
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/type_mismatch.rs
ChayimFriedman2, line 104: Need to check the file id, don't offer the fix if it's different from the current file.

crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -320,16 +312,13 @@ fn str_ref_to_owned(
...
     let to_owned = ".to_owned()".to_owned();
 
-    let edit = TextEdit::insert(expr.syntax().text_range().end(), to_owned);
-    let source_change = SourceChange::from_text_edit(
-        expr_ptr.file_id.original_file(ctx.sema.db).file_id(ctx.sema.db),
-        edit,
-    );
-    acc.push(fix("str_ref_to_owned", "Add .to_owned() here", source_change, expr_range));
+    let edit = TextEdit::insert(range.end(), to_owned);
+    let source_change = SourceChange::from_text_edit(file_id.file_id(ctx.sema.db), edit);
```
ChayimFriedman2, line 320: Here too don't offer the fix if the file is not the current file, otherwise this can change unrelated macro code.

ChayimFriedman2 (changes requested):

crates/ide-diagnostics/src/handlers/type_mismatch.rs
A4-Tacks: > otherwise this can change unrelated macro code.

Checking `file_id` doesn't seem to cure it, if the macro is defined in the same file

> Here too don't offer the fix if the file is not the current file

If it's not the current file, the fix seems difficult to apply

Fix is a code-action that is separate from diagnostics. If they are not in the same location, they will not be detected

ChayimFriedman2: > Checking file_id doesn't seem to cure it, if the macro is defined in the same file

Right, true. It is easier to notice though. You can check if it's from the macro by checking if the `SyntaxContext` is root.

A4-Tacks: In diagnostics, this seems to have no precedent

ChayimFriedman2: We can set the precedent :) But if you want, you can merge this as-is.

ChayimFriedman2 (approved):

PR #21953, opened 2026-04-04
fix: Improve add some on block like expression

A4-Tacks:
After rust-lang/rust-analyzer#21952, make the PR also support macros

Example
---
```rust
fn foo() -> Result<(), ()> {
    for _ in 0..5 {}$0
}
```

**Before this PR**

```rust
fn foo() -> Result<(), ()> {
    Ok(for _ in 0..5 {})
}
```

**After this PR**

```rust
fn foo() -> Result<(), ()> {
    for _ in 0..5 {}
    Ok(())
}
```

### DIFF
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -110,7 +110,8 @@ fn add_missing_ok_or_some(
 ) -> Option<()> {
     let root = ctx.sema.db.parse_or_expand(expr_ptr.file_id);
     let expr = expr_ptr.value.to_node(&root);
-    let expr_range = ctx.sema.original_range_opt(expr.syntax())?.range;
+    let hir::FileRange { file_id, range: expr_range } =
+        ctx.sema.original_range_opt(expr.syntax())?;
     let scope = ctx.sema.scope(expr.syntax())?;
 
     let expected_adt = d.expected.as_adt()?;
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -168,6 +168,21 @@ fn add_missing_ok_or_some(
...
+            && expr.syntax().parent().and_then(ast::StmtList::cast).is_some()
+        {
+            // Fix for forms like `fn foo() -> Result<(), String> { for _ in 0..8 {} }`
+            let mut builder = TextEdit::builder();
+            let indent = expr.indent_level();
+            builder
+                .insert(expr.syntax().text_range().end(), format!("\n{indent}{variant_name}(())"));
+
+            let source_change = SourceChange::from_text_edit(
+                expr_ptr.file_id.original_file(ctx.sema.db).file_id(ctx.sema.db),
```
ChayimFriedman2: You can't expect the expanded node to be in the original file.

crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -168,24 +168,31 @@ fn add_missing_ok_or_some(
...
             }
             return Some(());
+        } else if expr.is_block_like()
+            && expr.syntax().parent().and_then(ast::StmtList::cast).is_some()
+        {
+            // Fix for forms like `fn foo() -> Result<(), String> { for _ in 0..8 {} }`
+            let mut builder = TextEdit::builder();
+            let indent = expr.indent_level();
+            builder
+                .insert(expr.syntax().text_range().end(), format!("\n{indent}{variant_name}(())"));
```
ChayimFriedman2: This should be the upmapped range.

A4-Tacks: Oh, sorry, I forgot

PR #21955, opened 2026-04-04
fix: add semicolon for postfix format unit like snippets

A4-Tacks:
Example
---
```rust
fn main() {
    "{ 2+2 } { SomeStruct { val: 1, other: 32 } :?}".println
}
```

**Before this PR**

```rust
fn main() {
    println!("{} {:?}", 2 + 2, SomeStruct { val: 1, other: 32 })
}
```

**After this PR**

```rust
fn main() {
    println!("{} {:?}", 2 + 2, SomeStruct { val: 1, other: 32 });
}
```

### DIFF
--- a/crates/ide-completion/src/completions/postfix/format_like.rs
+++ b/crates/ide-completion/src/completions/postfix/format_like.rs
@@ -40,7 +40,7 @@ static KINDS: &[(&str, &str)] = &[
     ("logw", "log::warn!"),
     ("loge", "log::error!"),
 ];
-static HAS_VALUE: &[&str] = &["format"];
+static SNIPPET_RETURNS_NON_UNIT: &[&str] = &["format"];
 
 pub(crate) fn add_format_like_completions(
     acc: &mut Completions,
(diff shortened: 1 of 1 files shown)


crates/ide-completion/src/completions/postfix/format_like.rs
```diff
@@ -40,13 +40,15 @@ static KINDS: &[(&str, &str)] = &[
     ("logw", "log::warn!"),
     ("loge", "log::error!"),
 ];
+static HAS_VALUE: &[&str] = &["format"];
```
ChayimFriedman2: Name is a bit unclear. Perhaps `FORMAT_SNIPPET_RETURNS_UNIT`?

crates/ide-completion/src/completions/postfix.rs
```diff
@@ -310,7 +310,7 @@ pub(crate) fn complete_postfix(
     if let ast::Expr::Literal(literal) = dot_receiver.clone()
         && let Some(literal_text) = ast::String::cast(literal.token())
     {
-        add_format_like_completions(acc, ctx, &dot_receiver_including_refs, cap, &literal_text);
+        add_format_like_completions(acc, ctx, dot_receiver, cap, &literal_text, semi);
```
ChayimFriedman2, line 313: We could also handle match arms here and insert a comma.

crates/ide-completion/src/completions/postfix/format_like.rs
A4-Tacks: Since it's not public, I think `SNIPPET_RETURNS_NON_UNIT` is enough

crates/ide-completion/src/completions/postfix.rs
A4-Tacks: I will test this in the new PR

PR #21957, opened 2026-04-05
fix: enhance part of onEnter

asukaminato0721:
improve some onEnter handling, now it supports some more cases.

### DIFF
--- a/crates/ide/src/typing/on_enter.rs
+++ b/crates/ide/src/typing/on_enter.rs
@@ -1,12 +1,11 @@
-//! Handles the `Enter` key press. At the momently, this only continues
-//! comments, but should handle indent some time in the future as well.
+//! Handles the `Enter` key press, including comment continuation and
+//! indentation in brace-delimited constructs.
 
 use ide_db::{FilePosition, RootDatabase};
 use syntax::{
     AstNode, SmolStr, SourceFile,
     SyntaxKind::*,
-    SyntaxNode, SyntaxToken, TextRange, TextSize, TokenAtOffset,
-    algo::find_node_at_offset,
+    SyntaxToken, TextRange, TextSize, TokenAtOffset,
     ast::{self, AstToken, edit::IndentLevel},
 };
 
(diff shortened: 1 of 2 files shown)


crates/ide/src/typing/on_enter.rs
```diff
@@ -19,8 +19,10 @@ use ide_db::text_edit::TextEdit;
 // - <kbd>Enter</kbd> inside triple-slash comments automatically inserts `///`
 // - <kbd>Enter</kbd> in the middle or after a trailing space in `//` inserts `//`
 // - <kbd>Enter</kbd> inside `//!` doc comments automatically inserts `//!`
-// - <kbd>Enter</kbd> after `{` indents contents and closing `}` of single-line
-//   blocks, `use` trees, and item lists
+// - <kbd>Enter</kbd> inside an empty brace pair like `{$0}` expands it into a
+//   multi-line indented form
+// - <kbd>Enter</kbd> after `{` also indents simple single-line blocks, `use`
```
ChayimFriedman2: Redundant now?

crates/ide/src/typing/on_enter.rs
```diff
@@ -61,15 +63,18 @@ pub(crate) fn on_enter(db: &RootDatabase, position: FilePosition) -> Option<Text
     }
 
     if token.kind() == L_CURLY {
-        // Typing enter after the `{` of a block expression, where the `}` is on the same line
+        if let Some(edit) = on_enter_in_empty_braces(token, position) {
+            cov_mark::hit!(indent_block_contents);
+            return Some(edit);
+        }
+
         if let Some(edit) = find_node_at_offset(file.syntax(), position.offset - TextSize::of('{'))
```
ChayimFriedman2: And all of these are also redundant.

asukaminato0721: I think this is for indent in nonempty block?

ChayimFriedman2: Right. IMO this is redundant, but we can also handle this via tokens: if the closing brace is one the same line, extract anything between the braces with spaces after/before the braces skipped to a new indented line.

PR #21958, opened 2026-04-05
fix: modify name to avoid duplicate warnings lints

edragain2nd:
fix rust-lang/rust-analyzer#21943
I think we should rename one of these two 'warnings' lints to avoid confusion.


edragain2nd:
It is the first time for me to create a pull request. Any suggestions on what I might have missed would be greatly appreciated. ：）

ChayimFriedman2:
First, this file is auto-generated, don't change it manually. Second, we should not change one them, rather remove one of them, and make sure it won't break anything else.

edragain2nd:
tks for pointing out the mistake i make !

PR #21979, opened 2026-04-07
feat: When renaming a field, rename variables in constructors as well

### LINKED ISSUE #10489: Renaming a struct field should rename constructor arguments of the same name
When renaming a struct field, one often has to go clean up any constructors manually by also renaming the variable there. It would be nice if this was automated.

Example:

```rust
struct Struct {
    // renamed from struct_field
    field: String,
}

impl Struct {
    // this parameter should have been renamed too
    fn new(struct_field: String) -> Self {
        // so that struct init shorthand is used here
        Self { field: struct_field }
    }
}
```

Related: #6547

ChayimFriedman2:
`fn_return_points()` was adapted from rust-lang/rust-analyzer#21908 by @Albab-Hasan.

Fixes rust-lang/rust-analyzer#10489.

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -1984,6 +1984,14 @@ impl Variant {
             Variant::EnumVariant(e) => (*e).name(db),
         }
     }
+
+    pub fn adt(&self, db: &dyn HirDatabase) -> Adt {
+        match *self {
+            Variant::Struct(it) => it.into(),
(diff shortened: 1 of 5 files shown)


ShoyuVanilla (approved):

usamoi:
It is a good feature, but the behavior is a bit unexpected. I only became sure it was a feature, not a bug, after coming here😅

PR #21985, opened 2026-04-07
fix: Use the official npm mirror for all VS Code dependencies

Wilfred:
c420e48547e444e1f978be9dbf045d459936b0d5 changed a few packages to be installed from npmmirror.com rather than npmjs.org. This is harmless (there's a SHA512 hash to confirm integrity) but it prevents building the VS Code extension in locked down environments that only permit official package repositories.

It looks like npmmirror.com is an unofficial mirror run by Taobao (it replaced npm.taobao.org) and it can lag behind the official npm repository.

### DIFF
--- a/editors/code/package-lock.json
+++ b/editors/code/package-lock.json
@@ -257,7 +257,7 @@
         },
         "node_modules/@emnapi/core": {
             "version": "1.9.1",
-            "resolved": "https://registry.npmmirror.com/@emnapi/core/-/core-1.9.1.tgz",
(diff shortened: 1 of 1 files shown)


lnicola:
This is what `npm update @vscode/vsce ovsx` produced, I was also confused about the domain, but IIRC it was already used in the lockfile.

Did you just replace the URLs?

Wilfred:
I did just replace the URLs and confirm that I could still build the extension, yep.

Wilfred:
@lnicola do you have anything in `~/.npmrc` by any chance?

lnicola:
I see, looks like I do.

lnicola (approved):

PR #22003, opened 2026-04-09
fix: complete variants of hidden enums through public aliases

### LINKED ISSUE #21891: Cannot complete variant in hidden enum alias
**rust-analyzer version**: rust-analyzer 0.4.2833-standalone

**editor or extension**: Vim

**src/lib.rs**:
```rust
#[doc(hidden)]
pub enum Foo {
    Variant,
}

pub type Bar = Foo;
```

**src/main.rs**:
```rust
fn main() {
    let x = mycrate::Bar::Var$0; // cannot complete Variant
}
```

Amit5601:
Fixes rust-lang/rust-analyzer#21891.

Previously, the completion engine would aggressively bail out if the parent enum was marked . This pushes the visibility check down to the individual variants so they can be properly completed when accessed via a public type alias.

### DIFF
--- a/crates/ide-completion/src/completions.rs
+++ b/crates/ide-completion/src/completions.rs
@@ -276,9 +276,6 @@ impl Completions {
         path_ctx: &PathCompletionCtx<'_>,
         e: hir::Enum,
     ) {
-        if !ctx.check_stability_and_hidden(e) {
-            return;
-        }
         e.variants(ctx.db)
             .into_iter()
             .for_each(|variant| self.add_enum_variant(ctx, path_ctx, variant, None));
--- a/crates/ide-completion/src/tests/item.rs
+++ b/crates/ide-completion/src/tests/item.rs
@@ -380,3 +380,23 @@ foo!(f$0);
         "#]],
     );
 }
+
+#[test]
+fn completes_variant_through_hidden_enum_alias() {
+    check(
+        r#"
+//- /lib.rs crate:dep
+#[doc(hidden)]
+pub enum Foo { Variant }
+pub type Bar = Foo;
+
+//- /main.rs crate:main deps:dep
+fn main() {
+    let x = dep::Bar::V$0;
+}
+"#,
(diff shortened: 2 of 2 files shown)


crates/ide-completion/src/tests/item.rs
```diff
@@ -380,3 +380,22 @@ foo!(f$0);
...
+fn completes_variant_through_hidden_enum_alias() {
+    check(
+        r#"
+#[doc(hidden)]
+pub enum Foo { Variant }
+
+pub type Bar = Foo;
+
+fn main() {
+    let x = Bar::V$0;
```
A4-Tacks: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22003/64ddb549bc9a70d011328746fa46a8883f937b6b..6685b157a35161280864ee695a90bd94bd7ea912)*

This test is invalid, you need to reference it in another crate

A4-Tacks (changes requested):

crates/ide-completion/src/completions.rs
```diff
@@ -276,11 +276,9 @@ impl Completions {
         path_ctx: &PathCompletionCtx<'_>,
         e: hir::Enum,
     ) {
-        if !ctx.check_stability_and_hidden(e) {
-            return;
-        }
         e.variants(ctx.db)
             .into_iter()
+            .filter(|&variant| ctx.check_stability_and_hidden(variant))
```
A4-Tacks: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22003/64ddb549bc9a70d011328746fa46a8883f937b6b..6685b157a35161280864ee695a90bd94bd7ea912)*

This check is redundant, it already exists in add_enum_variant

A4-Tacks (changes requested):

A4-Tacks (approved):

PR #22010, opened 2026-04-10
Fix rustfmt relative custom command

erfanio:
When `rustfmt.overrideCommand` is a relative path, it's joined with the workspace root to make it an absolute path. Without this the command path can't be resolved, especially because rustfmt changes the cwd to make sure rustfmt.toml works correctly.

Currently `run_rustfmt` only does this when it finds a `target_spec` using `target_spec_for_file` which only works for the root file of the target.

This commit changes it to use `TargetSpec::for_file` which works for any file in the crate.

### DIFF
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -2442,8 +2442,7 @@ fn run_rustfmt(
         }
         RustfmtConfig::CustomCommand { command, args } => {
             let cmd = Utf8PathBuf::from(&command);
-            let target_spec =
-                crates.first().and_then(|&crate_id| snap.target_spec_for_file(file_id, crate_id));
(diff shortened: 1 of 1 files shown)


erfanio:
Looking at this code path a bit closer, I actually think maybe we don't need to find the target_spec at all. The code path tries to find a sensible root even if the file is not part of any target, so limiting formatting to files that are part of a target feel unnecessary here. See https://github.com/erfanio/rust-analyzer/commit/abd932c0bb

Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22010/64ddb549bc9a70d011328746fa46a8883f937b6b..342d9ed0dc298abceac9617622bcad02855d7c07)*

PR #22014, opened 2026-04-11
 Fix verify_only_whitespaces panicking on vertical tab (\x0B)

TG199:
- `verify_only_whitespaces` used `is_ascii_whitespace()` to validate whitespace tokens, which follows the WHATWG Infra
  standard and excludes vertical tab (\x0B).
  - The Rust lexer treats \x0B as valid whitespace (https://doc.rust-lang.org/reference/whitespace.html), so the
   function incorrectly panicked when encountering it.
  - Fix replaces the check with one that matches Rust's lexer definition exactly: is_ascii_whitespace() || byte == 0x0B.

  Test plan

  - Added a failing test (test_verify_only_whitespaces_accepts_rust_lexer_whitespace) that calls verify_only_whitespaces
   with all 6 Rust lexer whitespace characters, including \x0B
  - Test panicked before the fix, passes after
  - Full test suite passes: cargo test

### DIFF
--- a/crates/syntax/src/ast/make/quote.rs
+++ b/crates/syntax/src/ast/make/quote.rs
@@ -183,7 +183,7 @@ pub(crate) const fn verify_only_whitespaces(text: &str) {
     let text = text.as_bytes();
     let mut i = 0;
     while i < text.len() {
-        if !text[i].is_ascii_whitespace() {
+        if !text[i].is_ascii_whitespace() && text[i] != 0x0B {
             panic!("non-whitespace found in whitespace token");
         }
         i += 1;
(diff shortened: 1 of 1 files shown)


A4-Tacks (commented):
The `verify_only_whitespaces` function is only used by `quote!` for compile time literal validation
And no one would create a hard whitespace containing `'\x0B'` in `make`


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22014/9eb97ea96d8400e8957ddd56702e962614296583..302c2258763b3ef2e41cb71b58f984321f7179ca)*

PR #22015, opened 2026-04-11
fix: trigger completion inside format strings in VS Code

### LINKED ISSUE #22001: VSCode cannot auto trigger format_string completion
**rust-analyzer version**: rust-analyzer 0.4.2848-standalone

```rust
fn main() {
    let num = 2;
    const nun: i32 = 2;
    format!("{}")
}
```

Cursor after `"{` input `n`

In VSCode, it is necessary to manually press `Ctrl+Space` to trigger completion
In vim, no this bug

Image

Amit5601:
Fixes rust-lang/rust-analyzer#22001
This PR adds `"{".to_owned()` to the `trigger_characters` array in the server capabilities (`crates/rust-analyzer/src/caps.rs`). 
This instructs the VS Code client to automatically query the language server for completions the moment a user opens a curly brace.

### DIFF
--- a/crates/rust-analyzer/src/lsp/capabilities.rs
+++ b/crates/rust-analyzer/src/lsp/capabilities.rs
@@ -55,6 +55,7 @@ pub fn server_capabilities(config: &Config) -> ServerCapabilities {
                 ".".to_owned(),
                 "'".to_owned(),
                 "(".to_owned(),
+                "{".to_owned(),
             ]),
             all_commit_characters: None,
             completion_item: config.caps().completion_item(),


crates/rust-analyzer/src/lsp/capabilities.rs
```diff
@@ -55,6 +55,7 @@ pub fn server_capabilities(config: &Config) -> ServerCapabilities {
                 ".".to_owned(),
                 "'".to_owned(),
                 "(".to_owned(),
+                "{".to_owned(),
```
A4-Tacks, line 58: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22015/9eb97ea96d8400e8957ddd56702e962614296583..48a6fb00a754bf76dcd3056f59878457bbdf7d67)*

Is this really useful? The issue mentioned that even after entering `{nu$0}`, it did not auto complete
And setting `{` as the trigger character can cause a lot of trouble for those who want to open braces

Amit5601: @A4-Tacks Thank you for pointing that out!

The root cause isn't actually the LSP server's trigger_characters. The issue is that VS Code aggressively suppresses standard word completions inside strings by default to prevent noise, whereas Vim does not.

For anyone running into this issue in VS Code, you don't need a server update. You just need to enable string suggestions for Rust in your settings.json:

JSON
"[rust]": {
    "editor.quickSuggestions": {
        "other": true,
        "comments": false,
        "strings": true
    }
}
Since this is a client-side editor preference rather than a missing server capability, I'm going to go ahead and close this PR. Thanks again for the review!

A4-Tacks: Can't this be implemented in the extension? After all, this is a feature of r-a

A4-Tacks: Ok, pr ed in #22018

Issue #21602, opened 2026-02-07
build-std: ERROR Received compiler message for unknown package


Paladynee:
# Short description
enabling `build-std` in `.cargo/config.toml` will result in ERRORs in the Output panel of VSCode pointing to compiler builtins.

Minimal reproduction:

https://github.com/tristanpoland/test_analyzer_bug
default hello world program, with `.cargo/config.toml` containing this:
```toml
[unstable]
build-std = ["core"]
```

It'll give these lines in the Output > rust-analyzer Language Server section:
```
[Error - 12:37:36 AM] Server process exited with code 0.
2026-02-08T00:37:36.907400434+01:00 ERROR Received compiler message for unknown package: path+file:///home/eurydice/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/compiler-builtins/compiler-builtins#compiler_builtins@0.1.160
 path+file:///home/eurydice/coreign/rustaxiomtest#0.1.0
2026-02-08T00:37:36.907447367+01:00 ERROR Received compiler message for unknown package: path+file:///home/eurydice/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/compiler-builtins/compiler-builtins#compiler_builtins@0.1.160
 path+file:///home/eurydice/coreign/rustaxiomtest#0.1.0
```

I don't know if these lines mean anything or if something failed at all, but I was having other problems in one of my local workspaces when I came across these lines in the output, and it looks like it's reproducible on nightly and stable, on both linux and windows.

```
rust-analyzer version: 1.95.0-nightly (f889772 2026-02-05) [/home/eurydice/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/rust-analyzer]
rustc 1.95.0-nightly (f889772d6 2026-02-05)
```
extension version: rust-analyzer  v0.3.2777

pratham-srivastava-07:
I looked into this. When build-std is enabled, cargo emits compiler messages for stdlib crates (e.g. core) that aren't in rust-analyzer's package map (built from cargo metadata). The code logs these as ERROR/never!() but since the server exits with code 0, this is clearly non-fatal. I'd like to fix this by downgrading the log level to debug for the unknown-package case, and possibly adding a check to silently skip packages whose IDs resolve to sysroot/rustup toolchain paths. 

Happy to submit a PR if this approach looks good.

[...]

PR #22016, opened 2026-04-11
fix: recognize sysroot packages in build script output processing

pratham-srivastava-07:
This PR fixes issue rust-lang/rust-analyzer#21602 

## Fix

  Insert sysroot package IDs into `by_id` (mapped to `None`) alongside
  user workspace packages (mapped to `Some`). When the build script
  output callback encounters a `None` entry, it silently discards the
  output — sysroot build outputs are already handled separately via
  `rustc_crates()`.

### DIFF
--- a/crates/project-model/src/build_dependencies.rs
+++ b/crates/project-model/src/build_dependencies.rs
@@ -24,6 +24,7 @@ use crate::{
     CargoConfig, CargoFeatures, CargoWorkspace, InvocationStrategy, ManifestPath, Package, Sysroot,
     TargetKind,
     cargo_config_file::{LockfileCopy, LockfileUsage, make_lockfile_copy},
+    sysroot::RustLibSrcWorkspace,
     utf8_stdout,
 };
 
@@ -92,7 +93,7 @@ impl WorkspaceBuildScripts {
             sysroot,
             toolchain,
         )?;
-        Self::run_per_ws(cmd, workspace, progress)
+        Self::run_per_ws(cmd, workspace, sysroot, progress)
     }
 
     /// Runs the build scripts by invoking the configured command *once*.
@@ -100,6 +101,7 @@ impl WorkspaceBuildScripts {
     pub(crate) fn run_once(
         config: &CargoConfig,
         workspaces: &[&CargoWorkspace],
+        sysroots: &[&Sysroot],
         progress: &dyn Fn(String),
         working_directory: &AbsPathBuf,
     ) -> io::Result<Vec<WorkspaceBuildScripts>> {
@@ -118,7 +120,8 @@ impl WorkspaceBuildScripts {
         // NB: Cargo.toml could have been modified between `cargo metadata` and
         // `cargo check`. We shouldn't assume that package ids we see here are
         // exactly those from `config`.
-        let mut by_id = FxHashMap::default();
+        // `None` entries mark sysroot packages: recognized but not stored in any workspace output.
+        let mut by_id: FxHashMap<Arc<PackageId>, Option<(Package, usize)>> = FxHashMap::default();
         // some workspaces might depend on the same crates, so we need to duplicate the outputs
         // to those collisions
         let mut collisions = Vec::new();
@@ -132,28 +135,51 @@ impl WorkspaceBuildScripts {
                     if by_id.contains_key(&workspace[package].id) {
                         collisions.push((&workspace[package].id, idx, package));
                     } else {
-                        by_id.insert(workspace[package].id.clone(), (package, idx));
+                        by_id.insert(workspace[package].id.clone(), Some((package, idx)));
                     }
                 }
                 res
             })
             .collect();
 
+        // Insert sysroot package IDs so that messages for stdlib crates emitted when
+        // build-std is enabled are recognized rather than reported as unknown packages.
+        for sysroot in sysroots {
+            if let RustLibSrcWorkspace::Workspace { ws, .. } = sysroot.workspace() {
+                for package in ws.packages() {
+                    by_id.entry(ws[package].id.clone()).or_insert(None);
+                }
+            }
+        }
+
         let errors = Self::run_command(
             cmd,
             |package, cb| {
-                if let Some(&(package, workspace)) = by_id.get(package) {
-                    cb(&workspaces[workspace][package].name, &mut res[workspace].outputs[package]);
-                } else {
-                    tracing::error!("Received compiler message for unknown package: {}", package);
+                match by_id.get(package) {
+                    Some(Some((package, workspace))) => {
+                        cb(
+                            &workspaces[*workspace][*package].name,
+                            &mut res[*workspace].outputs[*package],
+                        );
+                    }
+                    Some(None) => {
+                        // Sysroot package (e.g. core when build-std is enabled);
+                        // its outputs are handled separately.
+                    }
+                    None => {
+                        tracing::error!(
+                            "Received compiler message for unknown package: {}",
+                            package
+                        );
+                    }
                 }
             },
             progress,
         )?;
         res.iter_mut().for_each(|it| it.error.clone_from(&errors));
         collisions.into_iter().for_each(|(id, workspace, package)| {
-            if let Some(&(p, w)) = by_id.get(id) {
-                res[workspace].outputs[package] = res[w].outputs[p].clone();
+            if let Some(Some((p, w))) = by_id.get(id) {
+                res[workspace].outputs[package] = res[*w].outputs[*p].clone();
             }
         });
 
@@ -281,6 +307,7 @@ impl WorkspaceBuildScripts {
     fn run_per_ws(
         cmd: Command,
         workspace: &CargoWorkspace,
+        sysroot: &Sysroot,
         progress: &dyn Fn(String),
     ) -> io::Result<WorkspaceBuildScripts> {
         let mut res = WorkspaceBuildScripts::default();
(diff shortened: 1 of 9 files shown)


crates/project-model/src/build_dependencies.rs
```diff
@@ -132,28 +135,51 @@ impl WorkspaceBuildScripts {
...
+                match by_id.get(package) {
+                    Some(Some((package, workspace))) => {
+                        cb(
+                            &workspaces[*workspace][*package].name,
+                            &mut res[*workspace].outputs[*package],
+                        );
+                    }
+                    Some(None) => {
+                        // Sysroot package (e.g. core when build-std is enabled);
+                        // its outputs are handled separately.
```
Veykril: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22016/9eb97ea96d8400e8957ddd56702e962614296583..80ab66806c19115d95edbd15db28f3940a943984)*

I am pretty sure if we are running into build-std, we need to actually handle the packages here and then disable the special case handling in those cases. Otherwise we will end up trying to load in correct proc-macros for std I think. (in theory at least)

pratham-srivastava-07: > _[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22016/9eb97ea96d8400e8957ddd56702e962614296583..80ab66806c19115d95edbd15db28f3940a943984)_
> 
> I am pretty sure if we are running into build-std, we need to actually handle the packages here and then disable the special case handling in those cases. Otherwise we will end up trying to load in correct proc-macros for std I think. (in theory at least)

Got it, makes sense. Quick question before I rework this...should I detect that build-std is active by checking if we receive compiler messages for sysroot packages during the run, or should I check the cargo config/command args upfront?

Veykril: I think we should pro-actively recognize build-std, by checking whether build-std is requested, https://github.com/rust-lang/rust-analyzer/blob/49094dbbeae543853ff4b89da27e0eab59101cd8/crates/project-model/src/cargo_config_file.rs

hishamhm:
@pratham-srivastava-07 Thank you for this PR! Any pending blockers? It seems that the CI failure was just for a simple clippy refactor. I hit #21602 recently, so happy to see that progress into fixing it has already been made!

pratham-srivastava-07:
> @pratham-srivastava-07 Thank you for this PR! Any pending blockers? It seems that the CI failure was just for a simple clippy refactor. I hit #21602 recently, so happy to see that progress into fixing it has already been made!

Hi @hishamhm, thanks for the heads up...have been away for a while, onto this right now and then proceed further

crates/project-model/src/build_dependencies.rs
pratham-srivastava-07: hey @Veykril, 

Following up on the red clippy check. I dug into it and it's not coming from the build-std changes...it's clippy::collapsible_match firing on crates/parser/src/grammar.rs (the visibility() fn), which is untouched by this PR and identical to master. It looks like a clippy 1.95 lint-bump that hits existing code:

error: this `if` can be collapsed into the outer `match`   --> crates/parser/src/grammar.rs:244
note: `-D clippy::collapsible-match` implied by `-D warnings`

The build-std code itself is clippy-clean. The fix is a one-arm refactor (folding the inner if into the match guard), which I've confirmed is behavior-preserving.

 How would you prefer I handle it.... fold the grammar.rs fix into this PR as a separate commit to get CI check correct, or should it go as its own PR against master? Want to keep the scope where you'd like it.

ChayimFriedman2: You need to rebase.

crates/project-model/src/build_dependencies.rs
```diff
@@ -92,17 +96,30 @@ impl WorkspaceBuildScripts {
...
         )?;
-        Self::run_per_ws(cmd, workspace, progress)
+        let build_std = CargoConfigFile::load(
+            workspace.manifest_path(),
+            &config.extra_env,
+            sysroot,
+            config.config_path.as_deref(),
+        )
+        .as_ref()
+        .is_some_and(CargoConfigFile::build_std_requested);
```
Veykril, line 106: We shouldn't need to load the cargo config here. At this point we should already have had it loaded before somewhere so we should thread that through to here. If we don't store the cargo-config in the workspace yet we should

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22016/6a4a48911d843d70369d9af8453a6dfa05bc14cc..76dec3249ad9c9d4c73255acfe448d2ee2be7990)*

crates/project-model/src/build_dependencies.rs
```diff
@@ -132,28 +150,113 @@ impl WorkspaceBuildScripts {
...
+            .zip(sysroots.iter())
+            .map(|(ws, sysroot)| {
+                CargoConfigFile::load(
+                    ws.manifest_path(),
+                    &config.extra_env,
+                    sysroot,
+                    config.config_path.as_deref(),
+                )
+                .as_ref()
+                .is_some_and(CargoConfigFile::build_std_requested)
```
Veykril, line 173: likewise

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22016/6a4a48911d843d70369d9af8453a6dfa05bc14cc..76dec3249ad9c9d4c73255acfe448d2ee2be7990)*

hishamhm:
@Veykril Friendly ping, since @pratham-srivastava-07 has rebased and now the PR passed CI. Anything else still pending here? Thanks!

crates/project-model/src/workspace.rs
```diff
@@ -1463,6 +1487,7 @@ fn detached_file_to_crate_graph(
         rustc_cfg.clone(),
         load,
         // FIXME: This looks incorrect but I don't think this causes problems.
+        &WorkspaceBuildScripts::default(),
         crate_ws_data.clone(),
```
Veykril, line 1491:
```suggestion
        &WorkspaceBuildScripts::default(),
        // FIXME: This looks incorrect but I don't think this causes problems.
        crate_ws_data.clone(),
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22016/6a4a48911d843d70369d9af8453a6dfa05bc14cc..76dec3249ad9c9d4c73255acfe448d2ee2be7990)*

crates/project-model/src/workspace.rs
```diff
@@ -1095,6 +1116,7 @@ fn project_json_to_crate_graph(
         rustc_cfg.clone(),
         load,
         // FIXME: This looks incorrect but I don't think this matters.
+        &WorkspaceBuildScripts::default(),
         crate_ws_data.clone(),
```
Veykril, line 1120:
```suggestion
        &WorkspaceBuildScripts::default(),
        // FIXME: This looks incorrect but I don't think this matters.
        crate_ws_data.clone(),
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22016/6a4a48911d843d70369d9af8453a6dfa05bc14cc..76dec3249ad9c9d4c73255acfe448d2ee2be7990)*

crates/project-model/src/workspace.rs
```diff
@@ -1040,13 +1066,8 @@ impl ProjectWorkspace {
...
             (
-                ProjectWorkspaceKind::Cargo { cargo, rustc, build_scripts: _, error: _ },
-                ProjectWorkspaceKind::Cargo {
-                    cargo: o_cargo,
-                    rustc: o_rustc,
-                    build_scripts: _,
-                    error: _,
-                },
+                ProjectWorkspaceKind::Cargo { cargo, rustc, .. },
+                ProjectWorkspaceKind::Cargo { cargo: o_cargo, rustc: o_rustc, .. },
```
Veykril, line 1070: please keep the exhaustive match, otherwise we might add a field and forget to eq that 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22016/6a4a48911d843d70369d9af8453a6dfa05bc14cc..76dec3249ad9c9d4c73255acfe448d2ee2be7990)*

crates/project-model/src/workspace.rs
```diff
@@ -1005,6 +1030,7 @@ impl ProjectWorkspace {
                         rustc_cfg.clone(),
                         cfg_overrides,
                         build_scripts,
+                        &WorkspaceBuildScripts::default(),
```
Veykril, line 1033:
```suggestion
                        // FIXME: cargo script might support build-std in the future?
                        &WorkspaceBuildScripts::default(),
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22016/6a4a48911d843d70369d9af8453a6dfa05bc14cc..76dec3249ad9c9d4c73255acfe448d2ee2be7990)*

crates/project-model/src/workspace.rs
```diff
@@ -816,7 +836,7 @@ impl ProjectWorkspace {
                 .into_iter()
                 .chain(mk_sysroot())
                 .collect::<Vec<_>>(),
-            ProjectWorkspaceKind::Cargo { cargo, rustc, build_scripts, error: _ } => {
+            ProjectWorkspaceKind::Cargo { cargo, rustc, build_scripts, error: _, .. } => {
```
Veykril, line 839: also exhaustive matching please

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22016/6a4a48911d843d70369d9af8453a6dfa05bc14cc..76dec3249ad9c9d4c73255acfe448d2ee2be7990)*

PR #22017, opened 2026-04-11
fix: prevent premature UnresolvedField diagnostic on type variables

### LINKED ISSUE #18872: Confusing "no field found" diagnostic
```rust
struct S {
    field: (),
}
fn f(_: impl FnOnce(S)) {}
fn main() {
    let cb = |s| _ = s.field;
    f(cb)
}
```
Here we report a "no field `field` on type `S`" diagnostic on the field access. Now this diagnostic is correct (rustc will fail here as well), the issue is that by the time we try to resolve the field we don't actually yet know what type `s` has, we only learn about this fact later on when the closure is used and we resolve all outstanding type variables once at the end of inference. This is where we then resolve the unknown type to `S` within the diagnostic which causes this confusing output. We should prevent the last part from happening.

Amit5601:
Fixes rust-lang/rust-analyzer#18872.
Added a `!receiver_ty.is_ty_var()` guard in `infer_field_access` to prevent premature `UnresolvedField` diagnostics when the closure parameter type isn't known yet. Added a regression test.

### DIFF
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -1702,12 +1702,14 @@ impl<'db> InferenceContext<'_, 'db> {
                     receiver,
                     tgt_expr,
                 );
-                self.push_diagnostic(InferenceDiagnostic::UnresolvedField {
-                    expr: tgt_expr,
-                    receiver: receiver_ty.store(),
-                    name: name.clone(),
-                    method_with_same_name_exists: resolved.is_ok(),
-                });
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
We should still issue a diagnostic, but the text of it should be different.

Amit5601:
@ChayimFriedman2 After digging into this, I realized my approach triggered the hir anti-spam filter in the test environment. I don't have the cycles to wire up the rest of the architectural bridge right now, so I'm going to close this PR to keep the queue clean.

PR #22022, opened 2026-04-12
feat: Support `impl` and `mut` restrictions

### LINKED ISSUE #22021: Support `impl` restriction syntax
The latest nightly supports `impl` restrictions. I am looking into how well it works before removing the "incomplete" flag on the feature, and immediately noticed that the syntax isn't supported by rust-analyzer. That's not particularly surprising given that it wasn't implemented until recently (and only semantically enforced for a day).

See the RFC (https://github.com/rust-lang/rfcs/pull/3323), the tracking issue (https://github.com/rust-lang/rust/issues/105077), and the implementation of the parsing (https://github.com/rust-lang/rust/pull/152943), which should be more than sufficient for showing the syntax that needs to be supported.

As far as r-a is concerned, I _think_ simply parsing it would be sufficient to avoid any errors. Right now it's a syntax error as you'd expect.

ChayimFriedman2:
I chose the order of mut restrictions to be `unsafe mut(...)` for unsafe fields, but it's undecided yet.

Closes https://github.com/rust-lang/rust-analyzer/issues/22021.

### DIFF
--- a/crates/ide-completion/src/context/analysis.rs
+++ b/crates/ide-completion/src/context/analysis.rs
@@ -1592,7 +1592,7 @@ fn classify_name_ref<'db>(
                     kind_macro_call(it)?
                 },
                 ast::Meta(meta) => make_path_kind_attr(meta)?,
-                ast::Visibility(it) => PathKind::Vis { has_in_token: it.in_token().is_some() },
+                ast::VisibilityInner(it) => PathKind::Vis { has_in_token: it.in_token().is_some() },
                 ast::UseTree(_) => PathKind::Use,
                 // completing inside a qualifier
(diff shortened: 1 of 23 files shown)


Veykril (commented):
We could try to accept both orderings for now (and maybe even retain that for recovery reasons, just emitting a soft error then)?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22022/7b6e1249b7320e16792e31ce67bb2e5f4acd6a8b..198bf2ac6bfdf2a5418c4b785b4e798dc6428c51)*

Veykril (approved):

PR #22031, opened 2026-04-13
Demoting completion relevance when an inherent impl already exists

### LINKED ISSUE #13197: Prefer completing `impl T[rait] for` over `impl T[ype]` when an inherent impl already exists
Given the following incomplete program:

```rust
trait Foob {}
struct Fooa {}
impl Fooa {}

impl Foo
```

If completion is invoked at the end, the current behavior appears to be to offer "Fooa" and "Foob" in alphabetical order; thus the default completion is "Fooa".

I propose that rust-analyzer could improve on this by completing the example to "Foob" (or even better, "`Foob for `"), demoting any type name for which there is _already_ an inherent `impl` block, since it is much less common to want to create a second inherent `impl` block than to create a trait `impl`.

WeiTheShinobi:
Fixed https://github.com/rust-lang/rust-analyzer/issues/13197

Added a new field `has_local_inherent_impl` to `CompletionRelevance` for demoting completion relevance when an inherent impl already exists.

My first contribution to this project. Thanks for the review!

### Case

```
trait Foob {}
struct Fooa {}
impl Fooa {}

impl Foo$0
```

### Before
```
st Fooa Fooa []
tt Foob  []
```

### After
```
tt Foob  []
st Fooa Fooa [has_local_inherent_impl]
```

---

About `CompletionRelevance`: 
```
if has_local_inherent_impl {
        score -= 5;
}
```

Not sure if `score -= 5` is good or not?

### DIFF
--- a/crates/ide-completion/src/item.rs
+++ b/crates/ide-completion/src/item.rs
@@ -184,6 +184,8 @@ pub struct CompletionRelevance {
     pub function: Option<CompletionRelevanceFn>,
     /// true when there is an `await.method()` or `iter().method()` completion.
     pub is_skipping_completion: bool,
+    /// if inherent impl already exists in current module, user may not want to implement it again.
+    pub has_local_inherent_impl: bool,
 }
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub struct CompletionRelevanceTraitInfo {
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22031/88005041f85726ff6067db68334a74e3a72c4812..d13247a878bcaea1e64e4d4d1fa922cc2176051b)*

PR #22035, opened 2026-04-14
minor: Ignore a hangy test

ChayimFriedman2:
CC @Wilfred.


Wilfred:
Thanks for the ping. I'm guessing it's another race condition, might be a legitimate bug. I'll play with running several copies of the test to see if I can get a few local reproductions.

PR #22048, opened 2026-04-15

ChayimFriedman2:
A bigger update than usual.

Best reviewed commit by commit, the commit messages are also important.


crates/hir-ty/src/next_solver/consts/valtree.rs
```diff
@@ -0,0 +1,712 @@
...
+        self.interned.fmt(f)
+    }
+}
+
+/// The raw bytes of a simple value.
+///
+/// This is a packed struct in order to allow this type to be optimally embedded in enums
+/// (like Scalar).
+#[derive(Clone, Copy, Eq, PartialEq, Hash)]
+#[repr(packed)]
```
Veykril: `#[repr(Rust, packed)]`, no need to disallow the clippy lint when we can just be explicit about the ABI here imo?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22048/b5122a9cb5934988d5dece2c9e07101b58ddcd1e..d2a22b14b6cab33405553ef0b2e57e1f1f3f3d8f)*

Veykril (approved):

PR #22050, opened 2026-04-15
internal: Represent lowered coroutines with closures

ChayimFriedman2:
This is how rustc does it, because they're similar in many ways, and most importantly: they share (or will share, not yet) capture analysis.

Also:
 - Copy closure inference from rustc again, this time really precisely.
 - Have separate ID types for coroutine closures and normal closures, for better type safety.
 - Validate, when constructing interned closures/coroutines/coroutine closures and `cfg(debug_assertions)`, that the `ExprId` is of an expr of the correct kind.

### DIFF
--- a/crates/hir-def/src/expr_store/lower.rs
+++ b/crates/hir-def/src/expr_store/lower.rs
@@ -46,8 +46,9 @@ use crate::{
     },
     hir::{
         Array, Binding, BindingAnnotation, BindingId, BindingProblems, CaptureBy, ClosureKind,
-        Expr, ExprId, Item, Label, LabelId, Literal, MatchArm, Movability, OffsetOf, Pat, PatId,
-        RecordFieldPat, RecordLitField, RecordSpread, Statement, generics::GenericParams,
+        CoroutineSource, Expr, ExprId, Item, Label, LabelId, Literal, MatchArm, Movability,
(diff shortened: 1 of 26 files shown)


crates/hir-def/src/expr_store/lower.rs
```diff
@@ -1417,7 +1446,22 @@ impl<'db> ExprCollector<'db> {
                         };
                         ClosureKind::Coroutine(movability)
                     } else if e.async_token().is_some() {
-                        ClosureKind::Async
+                        // It's important that this expr is allocated immediately before the closure.
+                        // We rely on it for `async_block_for_async_closure()`.
```
```suggestion
                        // We rely on it for `coroutine_for_closure()`.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22050/17d028e9d7161cf6ebf21e21d3c4a1835099e039..a2e7b2614afb689ca57707c86dc7756fbe15da01)*

Veykril: unless I missed `async_block_for_async_closure`

Veykril (approved):

crates/hir-def/src/expr_store/lower.rs
ChayimFriedman2: Yeah right I renamed this method.

PR #22054, opened 2026-04-15
fix(21943): remove duplicate lints

### LINKED ISSUE #21943: Duplicate `warnings` lint complete
**rust-analyzer version**: rust-analyzer 0.4.2848-standalone

**editor or extension**: Vim

```rust
#[allow(w$0)]
```

```text
#[allow(w)]
        warnings                                    f [LS]   mass-change the level for lints which produce
        warnings                                    f [LS]   warnings
        wasm_c_abi                                  f [LS]
```

jdrouet:
In this PR, I filter the lints that have already been added by name.
There are 2 main groups, the default lints and the rustdoc lints.
This should avoid having duplicates.

Fixes rust-lang/rust-analyzer#21943

PS: I've tried regenerating the lints, it introduced many changes which are not really related, so I manually removed them instead.

### DIFF
--- a/crates/ide-db/src/generated/lints.rs
+++ b/crates/ide-db/src/generated/lints.rs
@@ -1790,13 +1790,6 @@ pub const DEFAULT_LINTS: &[Lint] = &[
         warn_since: None,
         deny_since: None,
     },
-    Lint {
-        label: "warnings",
-        description: r##"lint group for: all lints that are set to issue warnings"##,
-        default_severity: Severity::Allow,
-        warn_since: None,
-        deny_since: None,
-    },
 ];
 
 pub const DEFAULT_LINT_GROUPS: &[LintGroup] = &[
--- a/xtask/src/codegen/lints.rs
+++ b/xtask/src/codegen/lints.rs
@@ -3,7 +3,7 @@
 #![allow(clippy::disallowed_types)]
 
 use std::{
-    collections::{HashMap, hash_map},
+    collections::{HashMap, HashSet, hash_map},
     fs,
     path::Path,
     str::FromStr,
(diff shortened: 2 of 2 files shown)


crates/ide-completion/src/tests/attribute.rs
```diff
@@ -1372,6 +1372,13 @@ mod lint {
             r#"#[allow(rustdoc::bare_urls struct Test;"#,
         );
     }
+
+    #[test]
```
ChayimFriedman2: We don't need a test for this IMO.

ChayimFriedman2 (changes requested):
Using a hashmap for deduplication, especially with the default hasher but not only, risks ordering changes across runs. Instead, sort and `dedup()`.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22054/17d028e9d7161cf6ebf21e21d3c4a1835099e039..96cfb4999c325b946fffa8d59491ac6a1d132c48)*

jdrouet:
> Using a hashmap for deduplication, especially with the default hasher but not only, risks ordering changes across runs. Instead, sort and `dedup()`.
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22054/17d028e9d7161cf6ebf21e21d3c4a1835099e039..96cfb4999c325b946fffa8d59491ac6a1d132c48)_

Thanks for your answer, but I'm not sure I understand properly.
As you said, the `HashSet` is only used for deduplication, so, as long as we don't change the order of the values in the `Vec`, the result will be the same considering we only call `.contains` 🤔 The hasher is not involved in this, is it?

Thanks in advance 😉

ChayimFriedman2 (approved):
Right, my mistake (I thought the deduplication is also inside lints and groups, not just between lints and groups). But please remove the test.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22054/17d028e9d7161cf6ebf21e21d3c4a1835099e039..96cfb4999c325b946fffa8d59491ac6a1d132c48)*

crates/ide-completion/src/tests/attribute.rs
jdrouet: addressed in https://github.com/rust-lang/rust-analyzer/pull/22054/commits/2c933e46ebd346950f3f1878b0cd51aa602f79bb 👍

ChayimFriedman2 (approved):
Please squash, then LGTM.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22054/17d028e9d7161cf6ebf21e21d3c4a1835099e039..2c933e46ebd346950f3f1878b0cd51aa602f79bb)*

jdrouet:
One commit is probably enough for this 😉 done!

ShoyuVanilla (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22054/c41c134c0f767b149c0c20a4adea1b05da947844..07cf5e4ebeadb977adc1248374614e18b9c8a0fa)*

PR #22055, opened 2026-04-15
fix: Some fixes for `upvars_mentioned()`

ChayimFriedman2:
- Support non-body expression stores.
 - Fix support for `self`.

### DIFF
--- a/crates/hir-ty/src/upvars.rs
+++ b/crates/hir-ty/src/upvars.rs
@@ -250,8 +250,9 @@ fn resolve_maybe_upvar<'db>(
     path: &Path,
 ) {
     if let Path::BarePath(mod_path) = path
-        && matches!(mod_path.kind, PathKind::Plain)
-        && mod_path.segments().len() == 1
+        && matches!(mod_path.kind, PathKind::Plain | PathKind::SELF)
+        // `self` is length zero.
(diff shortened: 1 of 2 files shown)


crates/hir-ty/src/upvars.rs
```diff
@@ -36,18 +36,85 @@ impl Upvars {
...
+        upvars_mentioned_impl(db, owner.into())
+    }
+}
+
+pub fn upvars_mentioned_impl(
+    db: &dyn HirDatabase,
+    owner: ExpressionStoreOwnerId,
 ) -> Option<Box<FxHashMap<ExprId, Upvars>>> {
-    let body = Body::of(db, owner);
+    let store = ExpressionStore::of(db, owner);
```
Veykril, line 112:
```suggestion
    let store = ExpressionStore::of(db, owner);
    if store.const_expr_origins().is_empty() {
        return None;
    }
```
We can skip the resolver construction here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22055/1e7decd129e29a6fb4a6629cad2e9d2a6ee53439..ddfe86ca394c4a625c9d91067456c32dce9b1fb0)*

Veykril (approved):

PR #22058, opened 2026-04-16
fix: Improve prettify_macro_expansion()

A4-Tacks:
- Add basic tests for prettify_macro_expansion
- Migrate prettify_macro_expansion to use SyntaxEditor

Example
---
```rust
const _: () = {
    loop {break}
    foo()
};
```

**Before this PR**

```rust
const _:() = {
    loop{
        break
    }foo()
};
```

**After this PR**

```rust
const _: () = {
    loop {
        break
    }
    foo()
};
```

---

```rust
const _: () = {
    match 2 {
        tmp => foo!(),
    };
};
```

**Before this PR**

```rust
const _:() = {
    match 2 {
        tmp => foo!(),

        };
};
```

**After this PR**

```rust
const _: () = {
    match 2 {
        tmp => foo!(),
    };
};
```

### DIFF
--- a/crates/ide-assists/src/handlers/inline_macro.rs
+++ b/crates/ide-assists/src/handlers/inline_macro.rs
@@ -258,7 +258,7 @@ macro_rules! whitespace {
         if true {}
     };
 }
-fn f() { if true{}; }
+fn f() { if true {}; }
 "#,
         )
     }
--- a/crates/ide-assists/src/handlers/inline_macro.rs
+++ b/crates/ide-assists/src/handlers/inline_macro.rs
@@ -1,6 +1,6 @@
 use hir::db::ExpandDatabase;
 use ide_db::syntax_helpers::prettify_macro_expansion;
-use syntax::ast::{self, AstNode};
+use syntax::ast::{self, AstNode, edit::AstNodeEdit};
 
 use crate::{AssistContext, AssistId, Assists};
 
@@ -52,6 +52,7 @@ pub(crate) fn inline_macro(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option
             // Don't call `prettify_macro_expansion()` outside the actual assist action; it does some heavy rowan tree manipulation,
             // which can be very costly for big macros when it is done *even without the assist being invoked*.
             let expanded = prettify_macro_expansion(ctx.db(), expanded, &span_map, target_crate_id);
+            let expanded = ast::edit::indent(&expanded, unexpanded.indent_level());
             editor.replace(unexpanded.syntax(), expanded);
             builder.add_file_edits(ctx.vfs_file_id(), editor);
         },
@@ -299,12 +300,12 @@ macro_rules! foo {
 }
 fn main() {
     cfg_if!{
-    if #[cfg(test)]{
-        1;
-    }else {
-        1;
-    }
-};
+        if #[cfg(test)]{
+            1;
+        }else {
+            1;
+        }
+    };
 }
 "#,
         );
--- a/crates/ide-assists/src/handlers/inline_macro.rs
+++ b/crates/ide-assists/src/handlers/inline_macro.rs
@@ -46,12 +46,14 @@ pub(crate) fn inline_macro(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option
         "Inline macro".to_owned(),
         text_range,
         |builder| {
+            let editor = builder.make_editor(unexpanded.syntax());
             let expanded = ctx.sema.parse_or_expand(macro_call.into());
             let span_map = ctx.sema.db.expansion_span_map(macro_call);
             // Don't call `prettify_macro_expansion()` outside the actual assist action; it does some heavy rowan tree manipulation,
             // which can be very costly for big macros when it is done *even without the assist being invoked*.
             let expanded = prettify_macro_expansion(ctx.db(), expanded, &span_map, target_crate_id);
-            builder.replace(text_range, expanded.to_string())
+            editor.replace(unexpanded.syntax(), expanded);
+            builder.add_file_edits(ctx.vfs_file_id(), editor);
         },
     )
 }
@@ -207,15 +209,15 @@ macro_rules! num {
             inline_macro,
             r#"
 macro_rules! foo {
-  () => {foo!()}
+  ($t:tt) => {foo!(1)}
 }
-fn f() { let result = foo$0!(); }
+fn f() { let result = foo$0!(0); }
 "#,
             r#"
 macro_rules! foo {
-  () => {foo!()}
+  ($t:tt) => {foo!(1)}
 }
-fn f() { let result = foo!(); }
+fn f() { let result = foo!(1); }
 "#,
         );
     }
--- a/crates/ide/src/expand_macro.rs
+++ b/crates/ide/src/expand_macro.rs
@@ -433,7 +433,9 @@ fn main() {
             expect![[r#"
                 match_ast!
                 {
-                    if let Some(it) = ast::TraitDef::cast(container.clone()){}else if let Some(it) = ast::ImplDef::cast(container.clone()){}else {
+                    if let Some(it) = ast::TraitDef::cast(container.clone()){
+                    }else if let Some(it) = ast::ImplDef::cast(container.clone()){
+                    }else {
                         {
                             continue
                         }
--- a/crates/ide/src/expand_macro.rs
+++ b/crates/ide/src/expand_macro.rs
@@ -357,7 +357,7 @@ fn main() {
 "#,
             expect![[r#"
                 bar!
-                for _ in 0..42{}"#]],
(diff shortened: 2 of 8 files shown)


ChayimFriedman2:
Is there a real reason to improve this function? Where important, we pass its output to rustfmt. We only need it so that tokens won't be combined.

A4-Tacks:
> Is there a real reason to improve this function?

Yes, when I use `inline_macro`, the old output often looks terrible

ChayimFriedman2:
Then we just should pass it to rustfmt there too.

A4-Tacks:
No, most of the time I use `inline_macro` to debug macros,
and rustfmt has poor support for formatting nested macros
And when developing macros, the expanded result may not be syntax correct, while rustfmt can only handle correct syntax

crates/ide-assists/src/handlers/inline_macro.rs
```diff
@@ -46,12 +46,15 @@ pub(crate) fn inline_macro(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option
         "Inline macro".to_owned(),
         text_range,
         |builder| {
+            let edit = builder.make_editor(unexpanded.syntax());
```
bit-aloo:
```suggestion
            let editor = builder.make_editor(unexpanded.syntax());
```

crates/syntax/src/ast/edit.rs
```diff
@@ -252,6 +252,10 @@ impl ast::IdentPat {
     }
 }
 
+pub fn indent(node: &SyntaxNode, level: IndentLevel) -> SyntaxNode {
+    level.clone_increase_indent(node)
+}
+
```
bit-aloo, line 258: better if add an associated method in IndentLevel

```rust
impl IndentLevel {
     #[must_use]
    pub fn indent(self, node: &SyntaxNode) -> SyntaxNode {
        self.clone_increase_indent(node)
    }
}
```

bit-aloo (commented):
Some suggestions and rest looks good. 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22058/4a244d4c6bf18bae57626dcaf81bf6442ad59380..4fa7e1c5e09d9353a75578d2c562a4b41e416e06)*

crates/syntax/src/ast/edit.rs
A4-Tacks: I did not use it as a method for `IndentLevel` because `IndentLevel::clone_increase_indent` is private

crates/syntax-bridge/src/prettify_macro_expansion.rs
```diff
@@ -161,3 +163,283 @@ fn is_text(k: SyntaxKind) -> bool {
...
+
+        fn remove_whitespaces(node: &SyntaxNode) {
+            let mut to_remove = vec![];
+            node.preorder_with_tokens().for_each(|it| match it {
+                WalkEvent::Enter(NodeOrToken::Token(tok)) if tok.kind().is_trivia() => {
+                    to_remove.push(tok);
+                }
+                _ => (),
+            });
+            to_remove.iter().for_each(ted::remove)
```
bit-aloo: Lets use SyntaxEditor for this, instead of ted.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22058/41716020ddd2c50ec36a1b356fbe06ef3d57cc39..dc8bf1f2657b9d40b5a9d3121fcd007a3fccc801)*

A4-Tacks: Done, Can I close #22158?

bit-aloo: This seems pretty clean. We can close that.

bit-aloo: Done

crates/ide/src/expand_macro.rs
```diff
@@ -433,9 +433,7 @@ fn main() {
             expect![[r#"
                 match_ast!
                 {
-                    if let Some(it) = ast::TraitDef::cast(container.clone()){}
-                    else if let Some(it) = ast::ImplDef::cast(container.clone()){}
-                    else {
+                    if let Some(it) = ast::TraitDef::cast(container.clone()){}else if let Some(it) = ast::ImplDef::cast(container.clone()){}else {
```
ChayimFriedman2: That seems like a regression?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22058/41716020ddd2c50ec36a1b356fbe06ef3d57cc39..c8ff15e048811c58c510bf9350fe767bf2d97f3d)*

crates/syntax-bridge/src/prettify_macro_expansion.rs
```diff
@@ -45,17 +44,23 @@ pub fn prettify_macro_expansion(
...
+                let is_last_child =
+                    node.parent().is_some_and(|parent| parent.last_child().as_ref() == Some(&node));
+                let is_always_newline = matches!(node.kind(), ATTR);
+                let is_non_last_newline = match node.kind() {
+                    MATCH_ARM | STRUCT | ENUM | UNION | FN | IMPL | MACRO_RULES | EXTERN_BLOCK
+                    | EXTERN_CRATE | MODULE => true,
+                    EXPR_STMT if Some(R_CURLY) == node.last_token().map(|it| it.kind()) => true,
+                    _ => false,
+                };
+                if !is_last_child && is_non_last_newline || is_always_newline {
```
ChayimFriedman2: Parentheses help here, even if they aren't required.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22058/41716020ddd2c50ec36a1b356fbe06ef3d57cc39..c8ff15e048811c58c510bf9350fe767bf2d97f3d)*

crates/syntax-bridge/src/prettify_macro_expansion.rs
```diff
@@ -77,16 +82,12 @@ pub fn prettify_macro_expansion(
 
         match tok.kind() {
             k if is_text(k)
-                && is_next(|it| !it.is_punct() || matches!(it, T![_] | T![#]), false) =>
+                && is_next(|it| !it.is_punct() || matches!(it, T![_] | T![#] | L_CURLY), false) =>
```
ChayimFriedman2, line 83: Why did you put `L_CURLY` here and removed the previous location? It seems appropriate to push a newline before `}`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22058/41716020ddd2c50ec36a1b356fbe06ef3d57cc39..c8ff15e048811c58c510bf9350fe767bf2d97f3d)*

crates/syntax-bridge/src/prettify_macro_expansion.rs
```diff
@@ -126,6 +123,12 @@ pub fn prettify_macro_expansion(
                 mods.push(do_ws(before, tok));
                 mods.push(do_ws(after, tok));
             }
+            T![:] if is_next(|it| it != T![:], false) && is_last(|it| it != T![:], false) => {
+                // XXX: Why input included WHITESPACE?
+                if is_next(|it| it != SyntaxKind::WHITESPACE, false) {
+                    mods.push(do_ws(after, tok));
+                }
+            }
```
ChayimFriedman2, line 129: Won't this put a space after the `::` in a path, e.g. `foo::bar`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22058/41716020ddd2c50ec36a1b356fbe06ef3d57cc39..c8ff15e048811c58c510bf9350fe767bf2d97f3d)*

A4-Tacks: See `is_next(|it| it != T![:], false) && is_last(|it| it != T![:], false)`, this is tested

crates/ide/src/expand_macro.rs
A4-Tacks: I don't think so. As long as there are elements in braces, they will insert newline

The old logic insert newline after any `}`. which does not conform to rustfmt

ChayimFriedman2: rustfmt puts a newline between the parentheses. Doing that will be better, but putting a newline before the `else` is preferable to no newline at all.

crates/syntax-bridge/src/prettify_macro_expansion.rs
A4-Tacks: This is adding a space instead of a new line, like `if true{}` -> `if true {}`

Placing it here is more catch some cases than placing it later

ChayimFriedman2 (approved):

PR #22064, opened 2026-04-16
ide: show Run lens for crate-root main under cfg(test)

### LINKED ISSUE #21948: Benchmarks don't have a "▶Run" button
**rust-analyzer version**: 0.3.2845-standalone

**rustc version**: rustc 1.94.0 (4a4ef493e 2026-03-02)

**editor or extension**: vs code

**relevant settings**: none

```rust
// benches/bench.rs
fn main() {
    println!("bench");
}
```

This does not display a "Run" lens, even when the benchmark is configured with `harness = false` in Cargo.toml (which isn't necessary for integration tests, they always show a "Run" lens).

0xmuon:
Fix: rust-lang/rust-analyzer#21948 
fixes missing “▶ Run” code lenses for benchmark crates that define a crate-root `fn main()` but are analyzed under `cfg(test)`, which previously caused the main runnable to be suppressed. Now, crate-root `fn main()` still produces a Bin runnable even when the crate is `cfg(test)`.It also preserves existing behavior for test-only modules by continuing to suppress “run binary” lenses for functions that only look like main via `#[export_name = "main"]` inside `#[cfg(test)]` contexts, avoiding extra runnables.
Tests cover both scenarios, including a new expect-test ensuring `fn main()` remains runnable under `cfg(test)`, and the existing `exported-main-in-cfg(test)` behavior.

### DIFF
--- a/crates/ide/src/runnables.rs
+++ b/crates/ide/src/runnables.rs
@@ -320,7 +320,7 @@ pub(crate) fn runnable_fn(
 ) -> Option<Runnable> {
     let edition = def.krate(sema.db).edition(sema.db);
     let under_cfg_test = has_cfg_test(def.module(sema.db).attrs(sema.db).cfgs(sema.db));
-    let kind = if !under_cfg_test && def.is_main(sema.db) {
+    let kind = if def.is_main(sema.db) && !(under_cfg_test && def.exported_main(sema.db)) {
         RunnableKind::Bin
     } else {
         let test_id = || {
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (commented):
The original code is correct. `under_cfg_test` does not look at the active cfg, rather whether the function has `#[cfg(test)]`. Did the added test not pass before?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22064/b8458013c217be4fccefc4e4f194026fa04ab4ca..e72288a1922058ea701e6a14be3887046b84a2d4)*

0xmuon:
`under_cfg_test` here is computed from `def.module(..).attrs(..).cfgs(..)` (i.e the module iss cfg options in the crate graph), not from a `#[cfg(test)]` attribute attached to the function itself. In the new fixture `//- /lib.rs crate:foo cfg:test`, the crate is analyzed with `cfg(test)` enabled, so `under_cfg_test` becomes true for the crate-root module even though fn main has no `#[cfg(test)]` attr. With the previous condition (`if !under_cfg_test && def.is_main(..)`), that suppressed the Bin runnable, so `main_runnable_even_with_cfg_test` would fail before this change.
The now it keeps a real `fn main()` runnable even under `cfg(test)`, while still suppressing the “fake main” case via `#[export_name="main"]` in `cfg(test)` contexts (`under_cfg_test && def.exported_main(..)`), which remains covered by `exported_main_is_test_in_cfg_test_mod`.

ChayimFriedman2:
`under_cfg_test` checks whether the module (right, not the function) has `#[cfg(test)]`. It is false (or should be false) for the fixture.

0xmuon:
oh right, `under_cfg_test` comes from `def.module(..).attrs(..).cfgs(..)` (cfg attributes on the module/crate root), not the crate’s active cfg set. With the current fixture `cfg:test`, `under_cfg_test` stays false, and I confirmed the new test would also pass on the parent commit. So this test as written doesn’t demonstrate the change, we likely need a different reproduction (or adjust how we detect “bench analyzed under cfg(test)” if that’s the root cause). what's your take?

ChayimFriedman2:
That this is not the cause of the bug.

PR #22067, opened 2026-04-17
fix: add parentheses on record expr for replace_let_with_if_let

A4-Tacks:
Example
---
```rust
fn main() {
    $0let x = Foo { x };
}
```

**Before this PR**

```rust
fn main() {
    if let x = Foo {
    }
}
```

**After this PR**

```rust
fn main() {
    if let x = (Foo { x }) {
    }
}
```


ChayimFriedman2 (approved):
Right, `needs_parens_in_place_of()` checks ancestry.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22067/064046069f65d07277e5b38b1bef4e558d709086..05ada7afe4b49442fb4a872e72a35dca2f3a5963)*

PR #22070, opened 2026-04-17
fix: adjust name of extract_type_alias

### LINKED ISSUE #13748: Extract type alias `Type` name clash
The "Extract type alias" helper that creates a new type alias. The generated alias is always named `Type`. Which is a problem if you are working in a file with a type named `Type` in scope (e.g. https://docs.rs/syn/latest/syn/enum.Type.html), as the follow up rename action changes the existing reference as well.

Here is the place where it uses the `Type` name:
https://github.com/rust-lang/rust-analyzer/blob/afe8f6b92210505ddf69d6056a9bd197c0bf7fa1/crates/ide-assists/src/handlers/extract_type_alias.rs#L72-L88

An ideal fix would be to identify clashes and pick a non-clashing name. However a easier fix would just be something extremely unlikely to clash e.g. `_extracted_alias`

A4-Tacks:
Fixes rust-lang/rust-analyzer#13748

Example
---
**Before this PR**

```rust
struct Type;
type $0Type = (u8, u8, u8);

struct S { field: Type }
```

**After this PR**

```rust
struct Type;
type $0Type1 = (u8, u8, u8);

struct S { field: Type1 }
```

### DIFF
--- a/crates/ide-assists/src/handlers/extract_type_alias.rs
+++ b/crates/ide-assists/src/handlers/extract_type_alias.rs
@@ -1,6 +1,6 @@
 use either::Either;
 use hir::HirDisplay;
-use ide_db::syntax_helpers::node_ext::walk_ty;
+use ide_db::syntax_helpers::{node_ext::walk_ty, suggest_name::NameGenerator};
 use syntax::{
     ast::{self, AstNode, HasGenericArgs, HasGenericParams, HasName, edit::IndentLevel},
     syntax_editor,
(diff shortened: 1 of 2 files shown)


bit-aloo (approved):

ChayimFriedman2 (commented):
I have no problem with this PR, but it does not fix the issue, just makes it somewhat less likely. A real fix will be to use a different name if it's already used, or ask the user for a name in the assist.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22070/26d7ba3f3d17987ce7a8e20fba1ceddbc9d85b56..252ec7f1d5b6df99bc37e5679d684d9232d35ea7)*

ChayimFriedman2 (approved):

PR #22073, opened 2026-04-17
fix: handle empty `/**/` in block_to_line assist

### LINKED ISSUE #22071: `block_to_line` (convert block comment to line comments) panics on empty `/**/`
```rust
fn foo() {
    /**/
    foo();
}
```

reproduce:
1. cursor on the `/**/` line
2. Ctrl+. → "Replace block comment with line comments"

panic:
```
request handler panicked: byte index 4 is out of bounds / begin <= end (4 <= 2) when slicing `/**/`
```

```
panicked at crates/ide-assists/src/handlers/convert_comment_block.rs:49:29:
begin <= end (4 <= 2) when slicing `/**/`
```

cause:
In `convert_comment_block.rs:49`:
```rust
let text = comment.text();
let text = &text[comment.prefix().len()..(text.len() - "*/".len())].trim();
```

this one is pretty interesting. no one would be converting empty comments but its still valid rust syntax. would this be considered a bug?

- rustc: 1.94.0
- rust-analyzer: 0.3.2862
- OS: Ubuntu
- Editor: VS Code

Albab-Hasan:
block_to_line sliced &text[prefix.len()..text.len()-2], which panicked on `/**/` since its prefix matches the entire comment (len 4) and text.len() - 2 = 2, producing the invalid range [4..2].

use strip_prefix/strip_suffix and emit a plain `//` when the body is empty after trimming.

fixes https://github.com/rust-lang/rust-analyzer/issues/22071

### DIFF
--- a/crates/ide-assists/src/handlers/convert_comment_block.rs
+++ b/crates/ide-assists/src/handlers/convert_comment_block.rs
@@ -382,6 +382,21 @@ fn main() {
     }
 
     #[test]
+    fn empty_block_to_line() {
+        check_assist(
+            convert_comment_block,
+            r#"
+/**/$0
+fn main() {}
+"#,
+            r#"
+
+fn main() {}
+"#,
+        );
+    }
+
+    #[test]
     fn end_of_line_block_to_line() {
         check_assist_not_applicable(
             convert_comment_block,
--- a/crates/syntax/src/ast/token_ext.rs
+++ b/crates/syntax/src/ast/token_ext.rs
@@ -32,11 +32,7 @@ impl ast::Comment {
     }
 
     pub fn prefix(&self) -> &'static str {
-        let &(prefix, _kind) = CommentKind::BY_PREFIX
-            .iter()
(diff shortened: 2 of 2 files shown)


ChayimFriedman2 (changes requested):
I prefer to fix `CommentKind::prefix()` so it will only include the prefix, aka. `/*`, and not the full `/**/`. It seems to work for all places using it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22073/26d7ba3f3d17987ce7a8e20fba1ceddbc9d85b56..c1f0e60cce51c4028921178400a9a1932c4a05b2)*

Albab-Hasan:
@ChayimFriedman2 done

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22073/26d7ba3f3d17987ce7a8e20fba1ceddbc9d85b56..beacd8a12328f7932df7debd73acfa6f945a851f)*

PR #22074, opened 2026-04-17
internal: support RUSTFMT_TOOLCHAIN for xtask codegen

A4-Tacks:


ChayimFriedman2 (approved):
Seems fine to me.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22074/26d7ba3f3d17987ce7a8e20fba1ceddbc9d85b56..70e4af2bf5f7dd005edb15c6d8ecca4a24eb048a)*

xtask/src/codegen.rs
```diff
@@ -130,16 +130,17 @@ impl fmt::Display for Location {
 fn reformat(text: String) -> String {
     let sh = Shell::new().unwrap();
     let rustfmt_toml = project_root().join("rustfmt.toml");
-    let version = cmd!(sh, "rustup run stable rustfmt --version").read().unwrap_or_default();
+    let toolchain = &std::env::var("RUSTFMT_TOOLCHAIN").unwrap_or("stable".to_owned());
```
Veykril, line 133: What is this env var? I never heard of that, if anything we should allow `RUSTFMT` I think, as that is also how `RUSTC` and `CARGO` env vars tend to be

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22074/26d7ba3f3d17987ce7a8e20fba1ceddbc9d85b56..70e4af2bf5f7dd005edb15c6d8ecca4a24eb048a)*

A4-Tacks: I haven't investigated the source of this variable,
I just hope a temporary variable like `UPDATE_EXPECT` can be used to adjust the toolchain

Because I only use nightly, I want to run `RUSTFMT_TOOLCHAIN=nightly cargo codegen`

PR #22076, opened 2026-04-17
internal: Allow only security updates for Dependabot

ChayimFriedman2:


joshka:
I'm assuming this is to reduce noise - have you considered tuning this to be less frequent instead?
(i.e. weekly / monthly with a [cooldown](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-) setting).

For rust updates it can be useful to use the new `increase-if-necessary` [versioning-strategy](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#versioning-strategy--), which allows you to set a proper minimal manifest version and only update it when there's an actual requirement that gets updated.

ChayimFriedman2:
My main reason is that I don't think it's a good idea to bump minor versions unless you have a specific reason.

Veykril (approved):

lnicola:
I quite like the dependabot PRs because they have a handy link to the diff view, especially with `increase-if-necessary`. Unfortunately, what's on GitHub doesn't have to match what we pull from crates.io.

lnicola:
I mean, it's fine if y'all feel like it's not worth upgrading all the time, I'm not _completely against_ merging this.

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    # Disable version updates for npm dependencies
```
Veykril, line 4: Actually, will this prevent security updates from happening? The docs don't really talk about this I think. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22076/8edb1119ab626029ec973fc95517c21c24339e6c..b372b8e2b52158ebd3357a5d52e9fbe5c17063d0)*

Veykril:
Imo I'd be fine with just setting the general cadence up to 2 weeks or something like that

.github/dependabot.yml
ChayimFriedman2: It will not. GitHub explicitly says that [in the example](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/customizing-dependabot-security-prs#example-1-configuration-for-security-updates-only).

PR #22078, opened 2026-04-18
perf: optimize allocation strategies of output/parser/event

landaire:
This change does a couple of different, but semi-related things. Happy to split this change up into multiple PRs if desired to keep these in isolation.

**Note:** This is my first contribution to rust-analyzer and also was mostly done by Claude. All tests pass and from my _human_ analysis of the change things look good. This description written by a human -- tables assembled by Claude.

Memory profiling done on an M5 MacBook Pro.

### `Event` Size Shrinkage

Through memory profiling, it was realized that the `parser::Parser`, a large consumer of memory, retains many `parser::Event`s that are forced to be a 24-byte structure, mostly because of the `Error` variant which held a `String`. Claude reworked this pattern to instead use an internal `Vec<String>` and the `Error` variant holds a 32-bit index into this vector.

Claude also noted that `forward_parent` field in `Event::Start` is always non-zero and can be changed to a `NonZeroU32` for further memory savings.

|                    | baseline | after   |
|--------------------|----------|---------|
| Parser::new bytes  | 2.40 GiB | 818 MiB |
| calls              | 76,546   | 76,546  |

### `Output.event` optimization

The `Output.event` vec uses a naive `push` strategy for allocation, but can be pre-allocated to avoid unnecessary memory growth from the vector's reallocation strategy.

|                                    | baseline        | after          |
|------------------------------------|-----------------|----------------|
| Output::enter_node (grow path)     | 584 MiB / 214k  | 0              |
| Output::with_event_capacity (new)  | —               | 209 MiB / 48k  |
| net                                | 584 MiB / 214k  | 209 MiB / 48k  |

-----------

Overall results:

| metric                | baseline       | after both wins | Δ        |
|-----------------------|----------------|-----------------|----------|
| total bytes allocated | **8.33 GiB**   | **6.36 GiB**    | **−23.6%** |
| total blocks          | 55,653,282     | 55,488,471      | −164,811 |
| at-gmax live bytes    | 1.06 GiB       | 1.06 GiB        | —        |
| at-gmax live blocks   | 10,567,626     | 10,567,623      | —        |
| at-end live bytes     | 633 MiB        | 631 MiB         | −2 MiB   |

### DIFF
--- a/crates/parser/src/event.rs
+++ b/crates/parser/src/event.rs
@@ -2,7 +2,7 @@
 //! It is intended to be completely decoupled from the
 //! parser, so as to allow to evolve the tree representation
 //! and the parser algorithm independently.
-use std::mem;
+use std::{mem, num::NonZeroU32};
 
 use crate::{
     SyntaxKind::{self, *},
@@ -12,6 +12,12 @@ use crate::{
 /// `Parser` produces a flat list of `Event`s.
 /// They are converted to a tree-structure in
 /// a separate pass, via `TreeBuilder`.
+///
+/// Kept to 8 bytes: error messages live in a side table on the `Parser`
+/// (the `errors` vec) and `Event::Error` only stores an index into it.
+/// `forward_parent` uses `NonZeroU32` so `Option` is niche-optimised away
+/// (the offset is always ≥ 1 because the forward parent sits later in the
+/// event stream).
 #[derive(Debug, PartialEq)]
 pub(crate) enum Event {
     /// This event signifies the start of the node.
@@ -53,10 +59,7 @@ pub(crate) enum Event {
     /// ```
     ///
     /// See also `CompletedMarker::precede`.
-    Start {
-        kind: SyntaxKind,
-        forward_parent: Option<u32>,
-    },
+    Start { kind: SyntaxKind, forward_parent: Option<NonZeroU32> },
 
     /// Complete the previous `Start` event
     Finish,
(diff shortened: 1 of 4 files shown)


ChayimFriedman2:
First, do you have any benchmark proving that parsing indeed became faster with these changes?

Second, about shrinking `Event`: `Event` is very short-living, therefore I don't see the benefit in shrinking it. I also don't expect to have an effect from cache.

I have no problems merging the `with_capacity()` change alone, though, it can never hurt.

landaire:
@ChayimFriedman2 sorry, not sure I follow on this:

>First, do you have any benchmark proving that parsing indeed became faster with these changes?

Maybe the `perf` tag is misleading but this is entirely about memory usage. I didn't do any runtime speed benchmarking. If you'd like me to do some, I'm happy to see what the implications of this change is.

>Second, about shrinking Event: Event is very short-living, therefore I don't see the benefit in shrinking it

While it might be short-living, it has a pretty major impact on peak memory usage by RA which I imagine does indeed have cascading effects. Especially for individuals on more resource-constrained systems (Apple is once again shipping MacBooks with 8GiB of RAM after all).

ChayimFriedman2:
It shouldn't have an impact even on peak memory usage because we aren't parsing many files concurrently, at most the number of cores. Unless profiling will show otherwise, of course.

landaire:
Ah I see what you're saying. Yeah I think the `Event` might come down to misinterpreting the `dhat` output and doing a bit too much at once. Let me run things again and see if my original interpretation still holds true.

Veykril:
The changes to `Event` do have the benefit of the type becoming a POD with no drop glue, which I think can have a beneficial effect when the event vector gets resized. I am fine with these changes personally. Though profiling would of course be nice either way.

ChayimFriedman2:
Not when it is resized but when it's dropped, still this can be beneficial.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22078/251df518d73abb5c5d573c4d5d266a3edae9ca5a..ec88d12aa3f8074de0258a31a8448550d3195c58)*

PR #22082, opened 2026-04-18
Implement `pattern_type` macro

Veykril:
Does not implement actually type level things though


crates/syntax/rust.ungram
```diff
@@ -706,6 +707,9 @@ FnPtrType =
 ForType =
   ForBinder Type
 
+PatternType =
+  'in' Type 'is' Pat
```
Veykril, line 711: Given this is all hidden behind a macro right now I took the liberty of going with a prefix approach for parsing simplicity for now

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22082/251df518d73abb5c5d573c4d5d266a3edae9ca5a..b27b697c06cca9d7e77d3efb431417f5688660fd)*

ChayimFriedman2:
This does not support the `!null` pattern. I'm not sure if extending normal patterns will cause problems, and we need a special parser.

Veykril:
Yeas it doesn't, ~~though I think !null is not yet used in the standard library~~ 🤦 I even copied a counter-example into minicore. In that case I'd still like to merge the PR as is I think, can be fixed up as a follow up. Once again I wish the project wouldn't dogfood features in the standard libraries..

That does raise the question of what to do about the new type only pattern here. Ideally we would re-use the pattern grammar here as adding a new type-pattern grammar is a lot to deal with.

ChayimFriedman2:
The problem of using it as normal pattern is that all other places dealing with patterns will need to deal it. I don't personally have an opinion what's better.

ChayimFriedman2:
I think we need to prioritize this and also support for `!null`, this is causing errors for people on nightly. See https://github.com/rust-lang/rust-analyzer/issues/22297.

Veykril:
Okay this also implements !null patterns as part of normal patterns for now. Its simpler to ignore this in the relevant places. The actual type system support isn't implemented yet (aside from lowering range pats and notnull pats to type IR, no or pat support yet), so also no layout support. 

Would like to get this merged without that for now as I don't know if I find the time to look into the layout part this week, though happy if someone else wants to take a stab at that

ChayimFriedman2:
I can look into the layout, although I'm not sure if there is actually anything we need to do.

Veykril:
> I can look into the layout, although I'm not sure if there is actually anything we need to do.

Well we do if we want to see sizes reported on hover again for almost any type since NonNull is using this now

PR #22083, opened 2026-04-18
fix: respect `#[deprecated]` attr when deciding if a `ModuleDef` completion is `deprecated`

### LINKED ISSUE #22062: No deprecated-ness info for item/const suggestions
Completions for functions correctly display the item name with a strikeout, but the ones for items (e.g. structs), constants, and possibly more, do not. This happens independently of whether the suggestion involves an auto-import.

**rust-analyzer version**: `rust-analyzer 1.94.1 (e408947 2026-03-25)`

**rustc version**: `rustc 1.94.1 (e408947bf 2026-03-25)`

**editor or extension**: VSCode, extension version 0.3.2862

**relevant settings**: none

```rust
#[deprecated]
fn bar() {}

#[deprecated]
const BAR: i32 = 0;

#[deprecated]
struct Bar;

mod foo {
    #[deprecated]
    pub fn baz() {}

    #[deprecated]
    pub const BAZ: i32 = 0;

    #[deprecated]
    pub struct Baz;
}

fn main() {
    let x = bar$0 // suggests:
                  // - using `bar`, shows as deprecated
                  // - using `BAR`, does _not_ show as deprecated
                  // - using `Bar`, does _not_ show as deprecated

    let x = baz$0 // suggests:
                  // - importing and using `foo::baz`, shows as deprecated
                  // - importing and using `foo::BAZ`, does _not_ show as deprecated
                  // - importing and using `foo::Baz`, does _not_ show as deprecated
}
```

ada4a:
A `ModuleDef` can be deprecated for two reasons:
- the def itself is marked as `#[deprecated]`
- the def is an assoc item whose trait is marked as deprecated

Previously, only the second case was checked. This PR adds a check for the first case as well.

The second commit adds "broken" test cases to demonstrate the bug, the third commit adds the fix and updates the test cases accordingly. Together, they fix rust-lang/rust-analyzer#22062.

The fourth commit is a refactoring of `.is_deprecated()` and `.is_deprecated_assoc_item()`: instead of needing to remember to call both functions separately, one now only needs to call `.is_deprecated()`, while passing a new argument to it in case the def can be an assoc item.

### DIFF
--- a/crates/hir/src/attrs.rs
+++ b/crates/hir/src/attrs.rs
@@ -38,7 +38,11 @@ pub enum AttrsOwner {
     Field(FieldId),
     LifetimeParam(LifetimeParamId),
     TypeOrConstParam(TypeOrConstParamId),
-    /// Things that do not have attributes. Used for builtin derives.
+    /// Things that do not have attributes.
+    ///
+    /// Used for:
+    /// - builtin derives
+    /// - builtin types (as those do not have attributes)
     Dummy,
 }
 
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -556,6 +556,23 @@ impl HasCrate for ModuleDef {
     }
 }
 
+impl HasAttrs for ModuleDef {
+    fn attr_id(self, db: &dyn HirDatabase) -> attrs::AttrsOwner {
+        match self {
+            ModuleDef::Module(it) => it.attr_id(db),
+            ModuleDef::Function(it) => it.attr_id(db),
+            ModuleDef::Adt(it) => it.attr_id(db),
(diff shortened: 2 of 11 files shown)


ChayimFriedman2 (approved):
This does not handle enum variants whose enum is deprecated, but still an improvement.

Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22083/251df518d73abb5c5d573c4d5d266a3edae9ca5a..400c929bf84f499bed2e01f87f7c7d442568f01a)*

ada4a:
> This does not handle enum variants whose enum is deprecated

Opened an issue to not forget this: #22090

PR #22084, opened 2026-04-18
implement custom `Debug` for `MiniCore`

ada4a:
When the contents correspond to the default, just output `MiniCore("<default>")` instead of the whole thing.

Helps reduce the length of the debug output in the common case.

### DIFF
--- a/crates/ide-db/src/lib.rs
+++ b/crates/ide-db/src/lib.rs
@@ -385,7 +385,7 @@ pub enum Severity {
     Allow,
 }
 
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22084/251df518d73abb5c5d573c4d5d266a3edae9ca5a..09fd5c7817bf6e2cff86f6e2fd4e21deff4de42a)*

PR #22085, opened 2026-04-18
feat(completion): reduce relevance for deprecated items

### LINKED ISSUE #18554: Should prefer associated items over a possible import with the same name
**rust-analyzer version**: 0.3.2188-standalone (ba56d9b9b 2024-11-17)
**editor or extension**: VSCode; version of the RA extension is v0.3.2188

(I also tested with nightly / pre-release versions; the behavior was the same.)

```rust
fn main() {
    let x = f32::INF
}
```

Put the cursor at the end of the line and press ^Space for completions. The first two suggestions will both complete to `f32::INFINITY`, but the *first* one on the list is one which will insert `use core::f32;` too.

Image

Since those constants are sort-of-deprecated, I don’t think RA should be preferring them, and in the general case, I would think that given any two items with the *same name*, it makes sense to rank completing the associated item over importing something new.

ada4a:
Fixes rust-lang/rust-analyzer#18554, albeit not in a way suggested in the issue: instead of prioritising assoc items, we deprioritise suggestions that are deprecated.


ada4a:
Depends on https://github.com/rust-lang/rust-analyzer/pull/22083, so

@rustbot blocked

crates/ide-completion/src/item.rs
```diff
@@ -708,9 +733,60 @@ mod tests {
...
+            CompletionItemKind::Expression,
+            Default::default(),
+            "",
+            syntax::Edition::DEFAULT,
+        );
+        builder.set_relevance(CompletionRelevance { is_deprecated: true, ..Default::default() });
+        let item = builder.build(&Default::default());
+        assert!(item.deprecated);
+        assert!(item.relevance.is_deprecated);
+    }
```
ada4a: Hm, CI doesn't like this test having `#[should_panic]`. The reason I added this is: the only way to avoid the panic is to run with debug assertions off (which is the next test), but running tests without debug assertions is pretty unusual.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22085/251df518d73abb5c5d573c4d5d266a3edae9ca5a..454a668815fb05b220b6c1a3c9cc054b21b6830b)*

ChayimFriedman2: You can instead make `Builder::set_deprecated()` decide, meaning it will also set `is_deprecated` for the relevance, and `set_relevance()` will do `relevance.is_deprecated = self.is_deprecated`. Then you can leave the debug assert in place. This is also a better design IMO.

ada4a: But then, if I call only `set_relevance()`, with `relevance.is_deprecated=true`, then the code would reset that to `false`, meaning the final `item.deprecated` and `item.relevance.is_deprecated` will both be `false`, no?

ada4a: Ah, you say "leave the debug assert in place", so you mean calling `set_relevance()` with `relevance.is_deprecated=true` would be disallowed. But that would theoretically only work when debug assertions are on? So _theoretically_, it would be possible to sneak it through in release mode.

Obviously one can just assume that the contributors will be reasonable and won't ignore the debug assert... but idk having fallback logic in `Builder::build` just in case doesn't hurt I think.

Idk, let me know what you think

ChayimFriedman2: It will panic, like it does currently.

ChayimFriedman2: IMO it's fine to assert in release mode as well, but even if not, I don't think we need a fallback.

ada4a: > it's fine to assert in release mode as well

That sounds a bit too harsh, but it does simplify the logic, so sure. One problem though: this test will still need to be `#[should_panic]` -- should I try to silence `tidy` maybe?

ChayimFriedman2: > this test will still need to be #[should_panic]

No it will not, just call `set_deprecated` on the item instead.

ada4a: I do have a test that calls `set_deprecated` -- it's the one before this one. What I intend to test in this one is precisely the behaviour when calling `set_relevance` with `is_deprecated=true`: ideally, it would be just "allow, reconcile later in `Builder::build`; but then I added a debug assert, so it became "hit a debug assert" (and now it'd become "hit a regular assert"). The initial behaviour could be tested without `#[should_panic]`, but the current one can't be.

You could of course argue that this test is just me being paranoid and that I might as well remove it.

ChayimFriedman2: We don't need a test for such things.

Urelatedly, the API I proposed seems better to me eyes.

ada4a: Okay, removed

ada4a: Hm, it looks like this change somehow allowed the creation of `CompletionItem`s with `deprecated = true` and `relevance.is_deprecated = false`. This seems to hint that the deprecatedness is sometimes set not through `set_deprecated`.

It could be interesting to track down where exactly that's happening currently, but maybe it'd be the most resilient to do reconciliation in `Builder::build` after all -- at that point we _know_ that no one will modify the deprecatedness info anymore

ChayimFriedman2: I don't have a preferrence.

ada4a: Ohh, it's because I forgot to set `self.relevance.is_deprecated = self.deprecated;` in `set_relevance()`. But I mean...

Let me articulate why I don't like this approach. It requires spreading the deprecatedness logic across two places: `set_deprecated` and `set_relevance`. This is different from all the other fields, because for those:
1. the setters don't have any extra logic in them -- they just set the field
2. if any post-processing is required, it's done inside `Builder::build`

ChayimFriedman2: Feel free to do what you propose if you think it's better.

crates/ide-completion/src/item.rs
```diff
@@ -590,21 +596,15 @@ impl Builder {
             None => TextEdit::replace(self.source_range, insert_text),
         };
 
+        // `deprecated` can be set through both `Self::set_deprecated` and `Self::set_relevance` --
+        // reconcile the possibly different values.
+        let deprecated = self.deprecated || self.relevance.is_deprecated;
```
ChayimFriedman2: IMO we should let `self.deprecated` be the deciding factor. That is, `self.relevance.is_deprecated = self.deprecated`.

ChayimFriedman2 (changes requested):
Please add a test that shows the score effect, like in the linked issue.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22085/0b696c09d6d75841e4cfd1dd27f476ba07c45a96..c684aef30a6ce703a7bceeee754147732b6a246a)*

crates/ide-completion/src/item.rs
ada4a: Right, because we disallow setting through `self.relevance` anyway. Good catch, thank you:)

ada4a:
> Please add a test that shows the score effect, like in the linked issue.

That should be `render::tests::float_consts_relevance`

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22085/4429a9c91ae01b2f260c242bb9f383bf722a1b4a..24dca9fd90eec75a051cdb7784446b484fe65386)*

PR #22088, opened 2026-04-19
snippet edit support no longer experimental

BenjaminBrienen:
blocked on https://github.com/gluon-lang/lsp-types/pull/311


ChayimFriedman2:
The switch from `SnippetCap` is wrong, IMO. It makes things less type-safe.

A4-Tacks:
I don't think replacing `SnippetCap` with `bool` is a good idea, typed programming helps us avoid making mistakes

BenjaminBrienen:
I thought that making it a boolean would bring it in-line with all the other capability switches.

ChayimFriedman2:
Given that we already have it typed, I'd prefer that we don't do that. We can make the other options typed, if you want.

BenjaminBrienen:
I split SnippetCap into Workspace and Completion variants because they are actually based on two different things and avoiding mixing them using the type system seems like it would be the "whole point", so to speak.

BenjaminBrienen:
I'm sorry that the Uri migration is mixed in

BenjaminBrienen:
Maybe this should be split into:
- update to lsp-types 0.97
- split SnippetCap into Completion* and Workspace*
- and after the blocker is released: update to lsp-types 0.98

does that sound good, or do you want to do all three at once and wait for upstream?

crates/rust-analyzer/src/lsp/to_proto.rs
DropDemBits, line 211: The LSP proposal version of how snippets are represented (new `lsp_types::{SnippetTextEdit, StringValue}` structs) is unfortunately different from how the lsp_ext version of the snippet is defined (a different insert_text_format).

This should really be emitting a `lsp_types::Edit to align with the proposal spec, though it'd be better to split up the snippet capability change from the snippet format change.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22088/4c57da3b27bf5dd3fdcbddd4ae82e1a27e5dc776..c0f12bb50cf3aa80998aad61c4f7c0b5bc20676d)*

crates/ide-completion/src/render/pattern.rs
```diff
@@ -174,68 +177,62 @@ fn render_pat(
...
-        }
-        None => {
-            format!(
-                "{name} {{ {}{} }}",
-                fields.map(|field| field.name(db).display_no_db(edition).to_smolstr()).format(", "),
-                if fields_omitted { ", .." } else { "" },
-                name = name
-            )
-        }
+    if completion_snippet_cap.is_some() {
```
DropDemBits, line 195: Not particularly driven either way, but I don't mind preserving the original match structure of this code (along with the other block below)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22088/4c57da3b27bf5dd3fdcbddd4ae82e1a27e5dc776..c0f12bb50cf3aa80998aad61c4f7c0b5bc20676d)*

DropDemBits (changes requested):

PR #22092, opened 2026-04-19
Sync function call args check fudging with rustc

ShoyuVanilla:
Apply changes in https://github.com/rust-lang/rust/pull/149320 and https://github.com/rust-lang/rust/pull/150316, which fix some inference failures on rustc

### DIFF
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -35,7 +35,8 @@ use crate::{
     lower::{GenericPredicates, lower_mutability},
     method_resolution::{self, CandidateId, MethodCallee, MethodError},
     next_solver::{
-        ErrorGuaranteed, FnSig, GenericArg, GenericArgs, TraitRef, Ty, TyKind, TypeError,
+        ClauseKind, ErrorGuaranteed, FnSig, GenericArg, GenericArgs, TraitRef, Ty, TyKind,
+        TypeError,
         infer::{
             BoundRegionConversionTime, InferOk,
             traits::{Obligation, ObligationCause},
@@ -1898,6 +1899,15 @@ impl<'db> InferenceContext<'_, 'db> {
         // Whether the function is variadic, for example when imported from C
         c_variadic: bool,
     ) {
+        let formal_input_tys: Vec<_> = formal_input_tys
+            .iter()
+            .map(|&ty| {
+                let generalized_ty = self.table.next_ty_var();
+                let _ = self.demand_eqtype(call_expr.into(), ty, generalized_ty);
+                generalized_ty
+            })
+            .collect();
+
         // First, let's unify the formal method signature with the expectation eagerly.
         // We use this to guide coercion inference; it's output is "fudged" which means
         // any remaining type variables are assigned to new, unrelated variables. This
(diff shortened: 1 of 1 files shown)


crates/hir-ty/src/infer/expr.rs
```diff
@@ -1917,18 +1927,23 @@ impl<'db> InferenceContext<'_, 'db> {
...
 
                         // Record all the argument types, with the args
                         // produced from the above subtyping unification.
-                        Ok(Some(
-                            formal_input_tys
-                                .iter()
-                                .map(|&ty| self.table.infer_ctxt.resolve_vars_if_possible(ty))
-                                .collect(),
-                        ))
+                        Ok(Some(formal_input_tys.clone()))
```
ShoyuVanilla, line 1946: This isn't (yet) applied to rustc, but as we call `resolve_vars_if_possible` in `fudge_inference_if_ok`, no need to do it again here
https://github.com/rust-lang/rust-analyzer/blob/fba71bb991356828a1b0a435a3fb34a806060e98/crates/hir-ty/src/next_solver/infer/snapshot/fudge.rs#L105

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22092/fba71bb991356828a1b0a435a3fb34a806060e98..4e67fc025a4bd1b9553725cce1c183fd71319c3d)*

crates/hir-ty/src/infer/expr.rs
```diff
@@ -1898,6 +1899,15 @@ impl<'db> InferenceContext<'_, 'db> {
         // Whether the function is variadic, for example when imported from C
         c_variadic: bool,
     ) {
+        let formal_input_tys: Vec<_> = formal_input_tys
+            .iter()
+            .map(|&ty| {
+                let generalized_ty = self.table.next_ty_var();
+                let _ = self.demand_eqtype(call_expr.into(), ty, generalized_ty);
```
ChayimFriedman2, line 1906: This is interesting, if you'd demand supertype/subtype I would understand, but isn't demand_eqtype equivalent to setting the type var?

crates/hir-ty/src/infer/expr.rs
```diff
@@ -1917,18 +1927,23 @@ impl<'db> InferenceContext<'_, 'db> {
                         // No argument expectations are produced if unification fails.
                         let origin = ObligationCause::new();
                         ocx.sup(&origin, self.table.param_env, expected_output, formal_output)?;
+
+                        for &ty in &formal_input_tys {
+                            ocx.register_obligation(Obligation::new(
```
ChayimFriedman2, line 1932: We have `self.table.register_wf_obligation()`, for short.

ChayimFriedman2 (approved):
Having authored these PRs, I assume you understand in this more than I do 😅 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22092/fba71bb991356828a1b0a435a3fb34a806060e98..4e67fc025a4bd1b9553725cce1c183fd71319c3d)*

crates/hir-ty/src/infer/expr.rs
ShoyuVanilla: Ah, I forgot about it 😅 Thanks!

crates/hir-ty/src/infer/expr.rs
ShoyuVanilla: Yeah, it's a bit subtle thing. Relating a type with a ty var generalizes it, which entails normalization:

https://github.com/rust-lang/rust-analyzer/blob/848e6aa69bcb6a8204186f58fc9ce6e6cd16c100/crates/hir-ty/src/next_solver/infer/relate/generalize.rs#L42

crates/hir-ty/src/infer/expr.rs
ShoyuVanilla: Oh wait, I think we shouldn't use it, since we are probing in this context so we shouldn't leave anything to the `table.fulfillment_cx` when leaving out of the probe. That's why we are creating a fresh new `let ocx = ObligationCtxt::new(..)` in this scope

PR #22098, opened 2026-04-19
fix: avoid prelude paths when `imports.preferPrelude` is false

### LINKED ISSUE #21914: Option for import path when re-exports exist — preferences/excludes
# Use case

A crate (lib) I'm working on provides a global prelude module, from which programmers can quickstart the usage of the crate by doing `crate_name::prelude::*;`.

Internally though, I want to avoid relying on the prelude re-exports, and choose their original location instead and makes those imports independent from changes to the prelude's re-exports.

Unfortunately, completion suggestions from `rust-analyzer` will always suggest the shortest path to an item that is not yet imported.

# Potential solutions

1. Option to ignore re-exports
2. Option to ignore a list of modules' re-exports
3. Option to ignore a list of specific re-exports
4. Option to enable prompting for disambiguation when multiple paths to an item in the same crate exist
5. Option to "canonicalize" an import (i.e. "rephrase" an import path that uses a re-export to an import path to the original location of the item)

Thank you for considering this feature :)

PS: May be related to #20468 , but I'm not certain that it's about the same thing, as they talk more about external modules rather than internal re-exports.

Souradip121:
## What changes

- Rename `Choice::prefer_due_to_prelude` → `has_prelude_segment`; track it
  unconditionally instead of gating on the config value
- Make `Choice::try_select` bidirectional: deprioritize prelude paths when
  `prefer_prelude: false`, prefer them when `true`
- Add `container_max_len` helper that relaxes the path-search limit by 1 when
  the current best is a prelude path and `prefer_prelude: false`, so a canonical
  path one segment longer is not silently cut off before it can be considered

## Before / After

```rust
// Crate exposes two paths to the same item:
//   krate::prelude::Foo          (3 segments — re-export)
//   krate::module::sub::Foo      (4 segments — canonical)
```

```rust
// Before  (prefer_prelude: false, the default)
use krate::prelude::Foo;        // shorter path always won
```

```rust
// After   (prefer_prelude: false, the default)
use krate::module::sub::Foo;   // canonical path is preferred

// After   (prefer_prelude: true)
use krate::prelude::Foo;        // prelude path preferred, unchanged
```

## Notes

- No new config key — `rust-analyzer.imports.preferPrelude: false` (the default)
  now has correct semantics in both directions
- `find_in_prelude` early-return (Rust language prelude items already in scope)
  is intentionally untouched; this fix only affects ranked path selection
- Closes rust-lang/rust-analyzer#21914

Used AI for faster development, but properly reviewed the code myself after writing it.

### DIFF
--- a/crates/hir-def/src/find_path.rs
+++ b/crates/hir-def/src/find_path.rs
@@ -414,19 +414,6 @@ fn find_in_sysroot(
         });
 }
 
-/// Computes the maximum path length allowed for a container module when searching for an item.
-///
-/// When `prefer_prelude` is false and the current best path goes through a `prelude` module,
-/// we relax the limit by 1 so that non-prelude paths one segment longer can still be found
-/// and then preferred by `Choice::try_select`.
-fn container_max_len(prefer_prelude: bool, max_len: usize, best_choice: &Option<Choice>) -> usize {
-    match best_choice {
-        Some(best) if !prefer_prelude && best.has_prelude_segment => best.path.len(),
-        Some(best) => best.path.len() - 1,
-        None => max_len,
-    }
-}
-
 fn find_in_dep(
     ctx: &FindPathCtx<'_>,
     visited_modules: &mut FxHashSet<(ItemInNs, ModuleId)>,
@@ -447,13 +434,7 @@ fn find_in_dep(
 
         // Determine best path for containing module and append last segment from `info`.
         // FIXME: we should guide this to look up the path locally, or from the same crate again?
-        let choice = find_path_for_module(
-            ctx,
-            visited_modules,
-            info.container,
-            true,
-            container_max_len(ctx.cfg.prefer_prelude, max_len, best_choice),
-        );
+        let choice = find_path_for_module(ctx, visited_modules, info.container, true, max_len - 1);
         let Some(mut choice) = choice else {
             continue;
         };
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
Note that we require disclosure of AI tools, if you used them. See [Use of AI tools in CONTRIBUTING.md](https://github.com/rust-lang/rust-analyzer/blob/master/CONTRIBUTING.md#use-of-ai-tools).

Souradip121:
@ChayimFriedman2 Yep, used AI, already disclosed in the commit message.

ChayimFriedman2:
The current code passing the best path so far's length as `max_len` is wrong, see [discussion on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/find_path.20max_len/with/590732987). Instead, we should pass the current `max_len` minus 1, and this will also solve the need for `container_max_len`. You can do that in this PR or in a separate PR.

Souradip121:
@ChayimFriedman2 doing in this PR only.

Souradip121:
@ChayimFriedman2 check once!

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22098/9006feed5f7160a2f87710e93a4a3c18faf21df7..ef72f16d312d810edc30c7a9420d859c65c28aca)*

PR #22100, opened 2026-04-19
hir-ty: report solver overflow as NextSolverError::Overflow

cammeresi:
When the next_solver returned `Maybe { cause: Overflow }` and the evaluation made no inference progress, `try_evaluate_obligations` re-registered the obligation instead of alerting the caller.  Callers like `structurally_normalize_term` could not distinguish overflow from successful ambiguous resolution and would return the inference variable they had allocated, letting it leak past the infcx (e.g. across salsa query boundaries into stored MIR bodies where it caused a panic during const evaluation).

Emit `NextSolverError::Overflow` in that case so callers can fail normalization instead of propagating a dangling variable (which in my code led thence to a panic).

[The above is mostly written by AI but reviewed by a human, but now the human is talking.]

I got rust-analyzer into a panic and unable to work with a bunch of my program via some very complex types that AI could only minimize down to a mere 111 lines that were standalone. It was caused by an array sized by const code that computed the size of a complex type that itself was composed of a chain of associated types that recursed several levels.

We found two ways to avoid the panic:  (1) increasing the recursion limit from a hardcoded limit of 50, which differs from rustc's configurable value that defaults to 128, and allowed reaching a solution; and (2) actually propagating the overflow out to a point where things can take notice and stop rather than proceeding in an incorrect state.  (1) may be desired independently, but it caused issues with two regression tests.

### DIFF
--- a/crates/hir-ty/src/next_solver/fulfill.rs
+++ b/crates/hir-ty/src/next_solver/fulfill.rs
@@ -10,7 +10,7 @@ use rustc_next_trait_solver::{
 use rustc_type_ir::{
     Interner, TypeSuperVisitable, TypeVisitable, TypeVisitableExt, TypeVisitor,
     inherent::{IntoKind, Span as _},
-    solve::{Certainty, NoSolution},
+    solve::{Certainty, MaybeCause, NoSolution},
 };
 
 use crate::next_solver::{
@@ -211,6 +211,11 @@ impl<'db> FulfillmentCtxt<'db> {
 
                 match certainty {
                     Certainty::Yes => {}
+                    Certainty::Maybe { cause: MaybeCause::Overflow { .. }, .. }
+                        if has_changed == HasChanged::No =>
+                    {
+                        errors.push(NextSolverError::Overflow(obligation));
+                    }
                     Certainty::Maybe { .. } => self.obligations.register(obligation, stalled_on),
                 }
             }


ChayimFriedman2:
This code is [copied from rustc](https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_trait_selection/solve/fulfill.rs.html#181-270), so it's very unlikely to be wrong. Further done we return if nothing has changed.

cammeresi:
I think it's more about signaling that inference failed via emitting an error than about returning sooner rather than later.

But if I'm off base here, that's fine, I can close.

ShoyuVanilla (changes requested):
I agree with Chayim.
The progresses in other goals may make the previously overflowing goal to proceed. This is why rustc and rust-analyzer do this in a fixpoint iteration.
So, doing this would make some inferences should be succeed to fail, though it might fix the things for your code.

> When the next_solver returned `Maybe { cause: Overflow }` and the evaluation made no inference progress, `try_evaluate_obligations` re-registered the obligation instead of alerting the caller. Callers like `structurally_normalize_term` could not distinguish overflow from successful ambiguous resolution and would return the inference variable they had allocated, letting it leak past the infcx (e.g. across salsa query boundaries into stored MIR bodies where it caused a panic during const evaluation).
> 
> Emit `NextSolverError::Overflow` in that case so callers can fail normalization instead of propagating a dangling variable (which in my code led thence to a panic).

So, it looks like some new inference vars are allocated per each iteration over and over and I have seen such cases in rustc as well and fixed them. It would be nice if you could minimize and share the failing code

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22100/adef948679e1f550805eeb2a78d10e25c0279f54..119b12a00d1b1132eb39e2ba5500ed0b49eef468)*

ShoyuVanilla:
> I think it's more about signaling that inference failed via emitting an error than about returning sooner rather than later.

This is not a mere matter of the order, as overflowed goals with `HasChanged::No` could be proved by progresses in other goals, so erroring them out as soon as we encounter them might cause the inference errors in valid codes.

PR #22101, opened 2026-04-20
fix: Port call expr type checking and closure upvar inference from rustc

ChayimFriedman2:
The first commits are small.

The biggest differences from rustc are due to the fact that we need to precisely store all source spans for captures, for the convert_closure_to_fn assist.


crates/hir-ty/src/layout/tests/closure.rs
```diff
@@ -125,13 +125,17 @@ fn capture_specific_fields2() {
 #[test]
 fn capture_specific_fields() {
     size_and_align_expr! {
+        minicore: fn;
```
ChayimFriedman2, line 128: The new call implementation is sensitive to `FnX` in core; anything that call closures need them, or it'll panic.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/adef948679e1f550805eeb2a78d10e25c0279f54..3ccf7e6f274387472810c058c2cee19b8093f1aa)*

crates/hir-ty/src/tests/closure_captures.rs
```diff
@@ -159,7 +189,7 @@ fn main() {
     let closure = || { let &mut ref b = a; };
 }
 "#,
-        expect!["53..79;20..21;67..72 ByRef(Shared) *a &'? bool"],
+        expect!["53..79;20..21;62..72 ByRef(Immutable) *a &'<erased> bool"],
```
ChayimFriedman2, line 192: I have to admit I didn't check all the changed ranges here to make sense 😅 If the reviewer wants to do that, well..., that's not easy. But `convert_closure_to_fn` relies on the ranges and its test passes, and I did check the capture kinds.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/adef948679e1f550805eeb2a78d10e25c0279f54..3ccf7e6f274387472810c058c2cee19b8093f1aa)*

crates/ide-assists/src/handlers/convert_closure_to_fn.rs
```diff
@@ -756,7 +780,7 @@ impl C {
 }
 "#,
             r#"
-struct A { b: &'static B, c: i32 }
+struct A { b: &'static mut B, c: i32 }
```
ChayimFriedman2, line 783: Here I changed it to `mut` because leaving it as is (there is a test down with the same structs I left as-is), the capture will be `&B` instead of `&i32`. This is due to `truncate_capture_for_optimization()`, which we didn't have earlier.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/adef948679e1f550805eeb2a78d10e25c0279f54..3ccf7e6f274387472810c058c2cee19b8093f1aa)*

crates/ide/src/hover/tests.rs
```diff
@@ -423,7 +423,7 @@ fn main() {
 
             ## Captures
             * `x.f1` by move
-            * `(*x.f2.0.0).f` by mutable borrow
+            * `x.f2.0.0.f` by mutable borrow
```
ChayimFriedman2, line 426: This is due to a small change (in my opinion, and in the code): I decided that we don't need a separate `display()` for the place, and we can use `display_for_source_code()`. The difference between the two used to be that the non-source-code version always printed derefs, with parentheses if needed. I decided now that it's nicer to show without them, given that the user expects autoderef anyway. Note that in the closure_captures.rs tests we do print all derefs, to ensure better coverage.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/adef948679e1f550805eeb2a78d10e25c0279f54..3ccf7e6f274387472810c058c2cee19b8093f1aa)*

crates/ide/src/inlay_hints.rs
```diff
@@ -1085,19 +1085,20 @@ fn foo() {
     fn closure_dependency_cycle_no_panic() {
         check(
             r#"
+//- minicore: fn
 fn foo() {
     let closure;
-     // ^^^^^^^ impl Fn()
+     // ^^^^^^^ impl FnOnce()
```
ChayimFriedman2, line 1091: The functions in this test error in rustc, so whatever we infer is good enough.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/adef948679e1f550805eeb2a78d10e25c0279f54..3ccf7e6f274387472810c058c2cee19b8093f1aa)*

crates/rust-analyzer/tests/slow-tests/main.rs
```diff
@@ -1577,6 +1577,6 @@ fn test() {
 
     let res: serde_json::Value = serde_json::from_str(res.as_str().unwrap()).unwrap();
     let arr = res.as_array().unwrap();
-    assert_eq!(arr.len(), 2);
+    assert_eq!(arr.len(), 1);
```
ChayimFriedman2, line 1580: There previously was a duplicate goal.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/adef948679e1f550805eeb2a78d10e25c0279f54..b8d0e9de761d697aca896ad9a73705b3ceac4a15)*

ChayimFriedman2:
This analysis_stats failure - turns out we're just missing match arm guards in `walk_child_exprs()` 🫢

crates/hir-def/src/expr_store.rs
```diff
@@ -675,6 +675,9 @@ impl ExpressionStore {
                 f(*expr);
                 arms.iter().for_each(|arm| {
                     f(arm.expr);
+                    if let Some(guard) = arm.guard {
```
Veykril, line 678: In these kinds of functions we should really enforce restructuring patterns to catch things 😅 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/c41c134c0f767b149c0c20a4adea1b05da947844..37febba555479b99d72463a6dbeaa826acf22f56)*

bench_data/glorious_old_parser
```diff
@@ -1,3 +1,4 @@
+//- minicore: fn
```
Veykril, line 1: Curious where we panic without language items, ideally that wouldn't happen (even if in practice no one will ever write a no_core rust program without the fn traits ...). Just looks a bit odd to add a mini core directive here haha

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/c41c134c0f767b149c0c20a4adea1b05da947844..37febba555479b99d72463a6dbeaa826acf22f56)*

crates/hir-ty/src/utils.rs
```diff
@@ -163,3 +165,54 @@ pub(crate) fn detect_variant_from_bytes<'a>(
     };
     Some((var_id, var_layout))
 }
+
+pub(crate) struct EnumerateAndAdjust<I> {
```
Veykril, line 169: Could you put down a one liner what this does / is meant for? Not really apparent from just the name without context

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/c41c134c0f767b149c0c20a4adea1b05da947844..37febba555479b99d72463a6dbeaa826acf22f56)*

crates/hir-ty/src/infer/callee.rs
```diff
@@ -0,0 +1,543 @@
...
+        let expr_ty = self.table.try_structurally_resolve_type(original_callee_ty);
+
+        let mut autoderef = GeneralAutoderef::new_from_inference_context(self, expr_ty);
+        let mut result = None;
+        while result.is_none() && autoderef.next().is_some() {
+            result =
+                Self::try_overloaded_call_step(call_expr, callee_expr, arg_exprs, &mut autoderef);
+        }
+
+        // FIXME: rustc does some ABI checks here, but the ABI mapping is in rustc_target and we don't have access to that crate.
```
Veykril, line 55: Curious what mapping you are speaking of here 🤔 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/c41c134c0f767b149c0c20a4adea1b05da947844..37febba555479b99d72463a6dbeaa826acf22f56)*

crates/hir-ty/src/infer.rs
```diff
@@ -637,11 +636,226 @@ pub struct InferenceResult {
     /// the first `rest` has implicit `ref` binding mode, but the second `rest` binding mode is `move`.
     pub(crate) binding_modes: ArenaMap<PatId, BindingMode>,
 
-    pub(crate) closure_info: FxHashMap<InternedClosureId, (Vec<CapturedItem>, FnTrait)>,
-    // FIXME: remove this field
-    pub mutated_bindings_in_closure: FxHashSet<BindingId>,
```
Veykril, line 642: 🥳 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/c41c134c0f767b149c0c20a4adea1b05da947844..37febba555479b99d72463a6dbeaa826acf22f56)*

crates/hir-ty/src/infer/closure/analysis/expr_use_visitor.rs
```diff
@@ -0,0 +1,1706 @@
+//! A different sort of visitor for walking fn bodies. Unlike the
+//! normal visitor, which just walks the entire body in one shot, the
+//! `ExprUseVisitor` determines how expressions are being used.
+//!
+//! In the compiler, this is only used for upvar inference, but there
+//! are many uses within clippy.
```
Veykril:
```suggestion
//! This is only used for upvar inference
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22101/c41c134c0f767b149c0c20a4adea1b05da947844..37febba555479b99d72463a6dbeaa826acf22f56)*

Veykril (approved):

bench_data/glorious_old_parser
Veykril: I wonder if we should add this directive as enabled by default for tests now

ChayimFriedman2: It panicked in the past as ``Expected to find a suitable `Fn`/`FnMut`/`FnOnce` implementation for ...`` in callee.rs, this was copied from rustc and I preferred not to change it. But I was forced to, because it panicked in the sysroot run in CI (it doesn't have a proper sysroot configured), and there I could not inject a minicore 😅

crates/hir-def/src/expr_store.rs
ChayimFriedman2: Agree, but it's better to leave that to another PR.

crates/hir-ty/src/infer/callee.rs
ChayimFriedman2: https://github.com/rust-lang/rust/blob/e22c616e4e87914135c1db261a03e0437255335e/compiler/rustc_hir_typeck/src/callee.rs#L166

PR #22103, opened 2026-04-20
Add VS Code create new project command

### LINKED ISSUE #19338: Create a new rust project from VSCode (cargo new / cargo init)
Something that I regularly do is create new rust projects (e.g. for testing out some functionality, bug testing, writing a demo etc.) Right now, the process for this is to have a terminal pointing at my root projects folder, run `cargo new project-name`, then open code pointing at the project (`code project-name` or `Cmd+O`, navigate and open).

This should be simpler and provided within VSCode.

For comparison / implementation ideas. The Java plugin for VSCode has a flow that seems reasonable. Select the project type, select the containing folder, select the project name:

Action: `Java: Create Java Project` also appears in the Explorer `No Folder Opened` view:

Image

Select Project Type (this would be just as simple as bin/lib for rust projects, but it could be worth considering making this extensible enough to integrate cargo-generate eventually here.)

Image

Select folder (this remembers the last location which was used to create a new project in)

Image

Enter project name:

Image

I searched for a similar issue and didn't find one already - not sure if I've missed it though as this seems like the sort of request which might have come up previously.

joshka:
Fixes rust-lang/rust-analyzer#19338.

## Summary

Add a VS Code command for creating a new Rust project from the editor.

The command prompts for a project kind, parent folder, and package name, then runs `cargo new` with the same effective environment used by the extension. After creation, it can open the project, open it in a new window, or add it to the current VS Code workspace.

## Details

- Add the `rust-analyzer.newProject` command.
- Add command palette, Explorer empty-state, and walkthrough entries.
- Add `rust-analyzer.projectCreation.openAfterCreate` to configure the post-create action.
- Validate project names before invoking Cargo for common input issues, while leaving Cargo as the source of truth for final package creation.
- Clarify VS Code workspace wording in the post-create prompt.
- Keep path display simple and avoid home-directory `~` shortening for now. That can be re-added later if wanted.

## Testing

- `npm -C editors/code run typecheck`
- `npm -C editors/code test`
- `npm exec -- prettier src/new_project.ts tests/unit/commands.test.ts --check`
- `npm exec -- eslint src/new_project.ts tests/unit/commands.test.ts`

## AI disclosure

I used Codex/AI tooling to research similar editor implementations, refine the implementation plan, review parts of the code structure, and draft/refine portions of this change. I reviewed the resulting code and notes before opening this PR.

### DIFF
--- a/editors/code/src/commands.ts
+++ b/editors/code/src/commands.ts
@@ -33,6 +33,7 @@ import { log } from "./util";
 import type { SyntaxElement } from "./syntax_tree_provider";
 
 export * from "./run";
+export { newProject } from "./new_project";
 
 export function analyzerStatus(ctx: CtxInit): Cmd {
     const tdcp = new (class implements vscode.TextDocumentContentProvider {
--- a/editors/code/src/new_project.ts
+++ b/editors/code/src/new_project.ts
@@ -1,0 +1,311 @@
+import * as vscode from "vscode";
+
+import type { Ctx, Cmd } from "./ctx";
+import * as ra from "./lsp_ext";
+import { cargoPath } from "./toolchain";
+import { log, spawnAsync } from "./util";
+
+type NewProjectKind = "bin" | "lib";
+type NewProjectOpenAction = "open" | "openNewWindow" | "addToWorkspace";
+
+type NewProjectTemplate = {
+    detail: string;
+    id: NewProjectKind;
+    label: string;
+};
+
+type NewProjectCargo = {
+    cargo: string;
+    cargoEnv: NodeJS.ProcessEnv;
+};
+
+const NEW_PROJECT_TEMPLATES: readonly NewProjectTemplate[] = [
+    {
+        id: "bin",
+        label: "Binary Application",
+        detail: "Create a Cargo binary package (`cargo new --bin`)",
+    },
+    {
+        id: "lib",
+        label: "Library",
+        detail: "Create a Cargo library package (`cargo new --lib`)",
+    },
+] as const;
+
+export function newProject(ctx: Ctx): Cmd {
+    return async () => {
+        const cargo = await resolveNewProjectCargo(ctx);
+
+        const selectedKind = await promptForNewProjectTemplate();
+        if (!selectedKind) {
+            return;
+        }
+
+        const parentFolder = await promptForNewProjectParentFolder();
+        if (!parentFolder) {
+            return;
+        }
+
+        const projectName = await promptForNewProjectName(parentFolder);
+        if (!projectName) {
+            return;
+        }
+
+        if (!(await createNewProject(cargo, parentFolder, selectedKind, projectName))) {
+            return;
+        }
+
+        const projectUri = vscode.Uri.joinPath(parentFolder, projectName);
+        const defaultAction = determineNewProjectOpenAction(
+            ctx.config.projectCreationOpenAfterCreate,
+            Boolean(vscode.workspace.workspaceFolders?.length),
+        );
+        const action =
+            defaultAction === "ask"
+                ? await promptForNewProjectOpenAction(
+                      projectName,
+                      Boolean(vscode.workspace.workspaceFolders?.length),
+                  )
+                : defaultAction;
+
+        if (action) {
+            await executeNewProjectOpenAction(ctx, action, projectUri);
+        }
+    };
+}
+
+async function resolveNewProjectCargo(ctx: Ctx): Promise<NewProjectCargo> {
+    // Use the same effective environment rust-analyzer uses elsewhere so project creation sees
+    // toolchain wrappers, PATH overrides, and CARGO_HOME changes from configuration.
+    const cargoEnv = { ...process.env, ...ctx.config.serverExtraEnv };
+    return { cargo: await cargoPath(cargoEnv), cargoEnv };
+}
+
+async function promptForNewProjectTemplate(): Promise<NewProjectKind | undefined> {
+    const selected = await vscode.window.showQuickPick(NEW_PROJECT_TEMPLATES, {
+        placeHolder: "Select a Rust project kind",
+    });
+    return selected?.id;
+}
+
+async function promptForNewProjectParentFolder(): Promise<vscode.Uri | undefined> {
+    const selectedFolder = await vscode.window.showOpenDialog({
+        title: "Select the parent folder for the new Rust project",
+        openLabel: "Select parent folder",
+        canSelectFiles: false,
+        canSelectFolders: true,
+        canSelectMany: false,
+    });
+    if (!selectedFolder?.length) {
+        return undefined;
+    }
+    return selectedFolder[0];
+}
+
+const CARGO_MANIFEST_NAME_PATTERN = /^[\p{Alphabetic}\p{Number}_-]+$/u;
+
+// Keep local validation focused on stable checks that can be reported in the input box, then let
+// `cargo new` remain the source of truth for package-name-specific identifier and keyword rules.
+export function validateNewProjectName(
+    value: string,
+    existingNames: readonly string[],
+): string | undefined {
+    const trimmedValue = value.trim();
+    if (trimmedValue.length === 0) {
+        return "Project name cannot be empty.";
+    }
+    if (trimmedValue.includes("/") || trimmedValue.includes("\\")) {
+        return "Project name cannot contain '/' or '\\' characters.";
+    }
+    if (trimmedValue === "." || trimmedValue === "..") {
+        return "Project name cannot be '.' or '..'.";
+    }
+    if (!CARGO_MANIFEST_NAME_PATTERN.test(trimmedValue)) {
+        return "Project name can contain only alphanumeric characters, '-' or '_'.";
+    }
+    if (existingNames.includes(trimmedValue)) {
+        return "A file or folder with this name already exists.";
+    }
+    return undefined;
+}
+
+async function promptForNewProjectName(parentFolder: vscode.Uri): Promise<string | undefined> {
+    let existingNames: string[] = [];
+    try {
+        const entries = await vscode.workspace.fs.readDirectory(parentFolder);
+        existingNames = entries.map(([name]) => name);
+    } catch (error) {
+        log.error("Failed to read project parent folder", error);
+        void vscode.window.showErrorMessage("Failed to read the selected parent folder.");
+        return undefined;
+    }
+
+    const projectName = await vscode.window.showInputBox({
+        prompt: `Enter the new project name to create inside ${parentFolder.fsPath}`,
+        validateInput: async (value) => validateNewProjectName(value, existingNames),
+    });
+    return projectName?.trim();
+}
+
+export function cargoNewArgs(kind: NewProjectKind, name: string): string[] {
+    return ["new", kind === "bin" ? "--bin" : "--lib", name];
+}
+
+async function createNewProject(
+    cargo: NewProjectCargo,
+    parentFolder: vscode.Uri,
+    kind: NewProjectKind,
+    projectName: string,
+): Promise<boolean> {
+    const args = cargoNewArgs(kind, projectName);
+    const createResult = await vscode.window.withProgress(
+        {
+            location: vscode.ProgressLocation.Notification,
+            title: `Creating Rust project ${projectName}`,
+        },
+        async () =>
+            spawnAsync(cargo.cargo, args, {
+                cwd: parentFolder.fsPath,
+                env: cargo.cargoEnv,
+            }),
+    );
+
+    if (createResult.error || createResult.status !== 0) {
+        const details = formatProcessDetails(createResult);
+        await showNewProjectError("Failed to create Rust project.", details || undefined, {
+            cargo: cargo.cargo,
+            args,
+            cwd: parentFolder.fsPath,
+            error: createResult.error?.message,
+            status: createResult.status,
+            stderr: createResult.stderr || undefined,
+            stdout: createResult.stdout || undefined,
+        });
+        return false;
+    }
+
+    return true;
+}
+
+function formatProcessDetails(result: { error?: Error; stderr: string; stdout: string }): string {
+    return [result.stderr, result.stdout, result.error?.message]
+        .filter((value): value is string => Boolean(value && value.trim().length > 0))
+        .join("\n")
+        .trim();
+}
+
+async function showNewProjectError(
+    message: string,
+    details: string | undefined,
+    logContext: {
+        cargo: string;
+        args: string[];
+        cwd?: string;
+        error?: string;
+        status: number | null;
+        stderr?: string;
+        stdout?: string;
+    },
+): Promise<void> {
+    // Keep command-failure logging focused on the invocation and process output. Environment
+    // variables may contain secrets such as API keys, tokens, and credentials, so failure logs
+    // must not dump the merged env here.
+    const commandLine = [logContext.cargo, ...logContext.args].join(" ");
+    log.error(message);
+    log.error(`command: ${commandLine}`);
+    if (logContext.cwd) {
+        log.error(`cwd: ${logContext.cwd}`);
+    }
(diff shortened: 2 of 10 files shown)


editors/code/src/commands.ts
```diff
@@ -886,6 +889,402 @@ export function reloadWorkspace(ctx: CtxInit): Cmd {
...
+        detail: "Create a Cargo library package (`cargo new --lib`)",
+    },
+] as const;
+
+/**
+ * Entry point for the `rust-analyzer: Create New Project...` command.
+ *
+ * This function keeps the user-visible workflow in one place: resolve Cargo, gather the project
+ * inputs, run `cargo new`, and then apply the requested post-create workspace action.
+ */
```
joshka, line 924: I intentionally steered codex to produce docs as the intent helps make the code readable IMO. Intention revealing names are good, but they don't convey tradeoffs / rationale / etc. This might be a bit noisier than most places in RA, but this is an intentional choice to improve such things.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22103/adef948679e1f550805eeb2a78d10e25c0279f54..d7ff960be22a850cbb22e8ff4dea09e185e98585)*

editors/code/src/commands.ts
```diff
@@ -886,6 +889,402 @@ export function reloadWorkspace(ctx: CtxInit): Cmd {
...
+export function newProject(ctx: Ctx): Cmd {
+    return async () => {
```
joshka, line 926: I split this method up from a longer one that had each of the pieces in a single method. It was difficult to follow / understand. This feels a bit table of contents and similarly problematic - there's probably a good balance, but I feel this is closer to the right size than the opposite (all the helper methods inlined here).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22103/adef948679e1f550805eeb2a78d10e25c0279f54..d7ff960be22a850cbb22e8ff4dea09e185e98585)*

editors/code/src/commands.ts
```diff
@@ -886,6 +889,402 @@ export function reloadWorkspace(ctx: CtxInit): Cmd {
...
+/**
+ * Resolves Cargo using rust-analyzer's effective environment and verifies that it is runnable.
+ *
+ * Returns `undefined` after surfacing an actionable error when Cargo cannot be launched, so the
+ * top-level command can treat toolchain failures the same way as user cancellation.
+ */
+async function resolveNewProjectCargo(ctx: Ctx): Promise<NewProjectCargo | undefined> {
+    // Use the same effective environment rust-analyzer uses elsewhere so project creation sees
+    // toolchain wrappers, PATH overrides, and CARGO_HOME changes from configuration.
+    const cargoEnv = { ...process.env, ...ctx.config.serverExtraEnv };
```
joshka, line 985: This is the part I'm least familiar with being correct choices. I think it's reasonably correct based on my understanding of what is going on here, but would value a second opinion.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22103/adef948679e1f550805eeb2a78d10e25c0279f54..d7ff960be22a850cbb22e8ff4dea09e185e98585)*

editors/code/src/commands.ts
```diff
@@ -886,6 +889,402 @@ export function reloadWorkspace(ctx: CtxInit): Cmd {
     return async () => ctx.client.sendRequest(ra.reloadWorkspace);
 }
 
+type NewProjectKind = "bin" | "lib";
```
joshka, line 892: This is a bunch of code to dump in an existing module. I'd tend to split additive code like this into a new separate module as it cuts down the cognitive load a bit, but went with keeping with the existing convention here rather than doing that. Ideally this code probably belongs in newProject.ts

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22103/adef948679e1f550805eeb2a78d10e25c0279f54..d7ff960be22a850cbb22e8ff4dea09e185e98585)*

editors/code/src/commands.ts
```diff
@@ -886,6 +889,402 @@ export function reloadWorkspace(ctx: CtxInit): Cmd {
...
+        status: number | null;
+        stderr?: string;
+        stdout?: string;
+    },
+): Promise<void> {
+    // Keep command-failure logging focused on the invocation and process output. Environment
+    // variables may contain secrets such as API keys, tokens, and credentials, so failure logs
+    // must not dump the merged env here.
+    const commandLine = [logContext.cargo, ...logContext.args].join(" ");
+    log.error(message);
```
joshka, line 1177: Not sure if there's an existing better logging approach here. The initial implementation just showed a notification, but I figure that there's a bunch of failure cases where the details of what's going on in cargo invocation is likely a bit more detailed for whatever reason. Maybe it's unlikely to hit this point. This should probably be extracted to a common function somewhere. Codex didn't find an obvious existing pattern / code to reuse for this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22103/adef948679e1f550805eeb2a78d10e25c0279f54..d7ff960be22a850cbb22e8ff4dea09e185e98585)*

editors/code/src/toolchain.ts
```diff
@@ -160,9 +160,44 @@ export function cargoPath(env?: Env): Promise<string> {
     if (env?.["RUSTC_TOOLCHAIN"]) {
         return Promise.resolve("cargo");
     }
+    if (env) {
```
joshka, line 163: I'm not sure I 100% understand this part of the change.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22103/adef948679e1f550805eeb2a78d10e25c0279f54..d7ff960be22a850cbb22e8ff4dea09e185e98585)*

editors/code/src/toolchain.ts
```diff
@@ -160,9 +160,44 @@ export function cargoPath(env?: Env): Promise<string> {
...
+        return getPathForExecutableWithEnv("cargo", env);
+    }
```
joshka, line 165: The new-project command builds an effective Cargo environment from rust-analyzer configuration (`server.extraEnv`) before running `cargo new`. Resolve `cargo` from that same env here so the command does not discover one Cargo binary and then execute another.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22103/adef948679e1f550805eeb2a78d10e25c0279f54..d7ff960be22a850cbb22e8ff4dea09e185e98585)*

joshka:
image

image

image

image

image

image

image

image

ChayimFriedman2:
This is a large wall of code, and I'm not even sure this feature is in scope for r-a.

joshka:
> This is a large wall of code, and I'm not even sure this feature is in scope for r-a.

It's working, self-reviewed, tested manually and with unit tests, with a firm spec in the linked issue that I spent time iterating on prior to throwing the tooling at it, plus refactoring and improving the generated code.

It exists because other extensions do this already, and it allows me to live in vscode rather than a terminal in order to create rust applications / libraries. This seems like a solidly good way to validate whether it's scope.

r-a is the place where rust projects live in vscode. From that perspective creating a new rust project in vscode should live in r-a.

editors/code/src/commands.ts
```diff
@@ -886,6 +889,402 @@ export function reloadWorkspace(ctx: CtxInit): Cmd {
...
+}
+
+/**
+ * Formats the selected parent folder for user-facing prompts, shortening home-relative paths to
+ * `~` to keep the input box text readable.
+ */
+function displayProjectBasePath(parentFolder: vscode.Uri): string {
+    const fsPath = parentFolder.fsPath;
+    const home = os.homedir();
+    if (fsPath === home) {
```
stefnotch: This won't ever be true on Windows, because the drive letter won't match. In general, I tend to be extremely suspicious of any path manipulation, because it's so easy to get it wrong.

> Will handle UNC paths and normalize windows drive letters to lower-case
> https://code.visualstudio.com/api/references/vscode-api#Uri

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22103/adef948679e1f550805eeb2a78d10e25c0279f54..28f842efd63d56331ec0e984ac1439a08231f080)*

editors/code/src/commands.ts
```diff
@@ -886,6 +889,402 @@ export function reloadWorkspace(ctx: CtxInit): Cmd {
...
+    });
+    if (!selectedFolder?.length) {
+        return undefined;
+    }
+    return selectedFolder[0];
+}
+
+// Keep local validation intentionally narrow: reject obviously invalid folder names up front,
+// then let `cargo new` remain the source of truth for package-name-specific rules.
+export function validateNewProjectName(
```
stefnotch: Is there any upside in doing this validation?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22103/adef948679e1f550805eeb2a78d10e25c0279f54..28f842efd63d56331ec0e984ac1439a08231f080)*

editors/code/src/commands.ts
```diff
@@ -886,6 +889,402 @@ export function reloadWorkspace(ctx: CtxInit): Cmd {
...
+async function promptForNewProjectOpenAction(
+    projectName: string,
+    hasWorkspaceFolders: boolean,
+): Promise<NewProjectOpenAction | undefined> {
+    let message = `Would you like to open ${projectName}?`;
+    const open = "Open";
+    const openNewWindow = "Open in New Window";
+    const choices = [open, openNewWindow];
+
+    const addToWorkspace = "Add to Workspace";
```
stefnotch: What does "workspace" mean in this context? Is it about adding it to the cargo workspace or about the vscode workspace?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22103/adef948679e1f550805eeb2a78d10e25c0279f54..28f842efd63d56331ec0e984ac1439a08231f080)*

stefnotch:
As a general question, which other extensions do this? I don't have any extensions installed with a similar feature, except for [the Typst one](https://marketplace.visualstudio.com/items?itemName=myriad-dreamin.tinymist). Though I don't fully understand the Typst one. I think Typst does it, because it doesn't rely on a CLI?

editors/code/src/commands.ts
joshka: yes - you get immediate results in the ui where it prompts for a name rather than after you hit return.

image

editors/code/src/commands.ts
joshka: vscode workspace. This is about running cargo new. That cannot add a crate to a workspace.

joshka:
> As a general question, which other extensions do this? I don't have any extensions installed with a similar feature, except for [the Typst one](https://marketplace.visualstudio.com/items?itemName=myriad-dreamin.tinymist). Though I don't fully understand the Typst one. I think Typst does it, because it doesn't rely on a CLI?

(From the issue plan details) https://github.com/rust-lang/rust-analyzer/issues/19338#issuecomment-4277204885:
I also checked several other extensions to ground the expected behavior and tradeoffs here,
including Swift, Java, Python, Go, and .NET.

editors/code/src/commands.ts
joshka: Removed this home-relative `~` shortening for now to simplify the PR. We can re-add it later if it proves useful or necessary.

editors/code/src/commands.ts
stefnotch: https://doc.rust-lang.org/cargo/reference/manifest.html#the-name-field

Then why not harden the check to be like cargo itself? It seems like a pretty straightforward regex.

editors/code/src/commands.ts
stefnotch: I see. Would it be possible to change the wording?
If I used this cargo new command in a sub-folder of a cargo workspace, then the project would be automatically added to the cargo workspace.

At which point I'd be confused about why the extension is asking about a workspace. Until I figure out that it is asking about a vscode workspace.

joshka: Updated the prompt and action label to say "VS Code workspace" explicitly, so it does not read as Cargo workspace membership.

joshka:
I pushed a cleanup pass aimed at making this easier to review: the new-project workflow now lives in `editors/code/src/new_project.ts`, with `commands.ts` only re-exporting the command. I also trimmed the helper docs down to the comments that explain behavior or tradeoffs.

Scope-wise, this still delegates project creation to `cargo new`; it does not edit Cargo manifests. The post-create "Add to VS Code Workspace" action only controls the VS Code workspace/window behavior.

joshka:
https://github.com/user-attachments/assets/92a3f6a0-7d25-4dea-9d46-a58430c63f86

joshka:
ping

joshka:
ping

editors/code/src/new_project.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+    if (trimmedValue.includes("/") || trimmedValue.includes("\\")) {
+        return "Project name cannot contain '/' or '\\' characters.";
+    }
+    if (trimmedValue === "." || trimmedValue === "..") {
+        return "Project name cannot be '.' or '..'.";
+    }
+    if (!CARGO_MANIFEST_NAME_PATTERN.test(trimmedValue)) {
+        return "Project name can contain only alphanumeric characters, '-' or '_'.";
+    }
```
workingjubilee: These are all the same error, aren't they? Why have three separate checks for it?

joshka:
@rustbot ready
r? @Veykril

(unsure if you're the best person to review this [#t-compiler/rust-analyzer > New project command in VSCode @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/New.20project.20command.20in.20VSCode/near/597448807))

editors/code/src/new_project.ts
```diff
@@ -0,0 +1,311 @@
...
+        title: "Select the parent folder for the new Rust project",
+        openLabel: "Select parent folder",
+        canSelectFiles: false,
+        canSelectFolders: true,
+        canSelectMany: false,
+    });
+    if (!selectedFolder?.length) {
+        return undefined;
+    }
+    return selectedFolder[0];
```
workingjubilee: This is just `return selectedFolder[0]`, isn't it? Does tsc reject this if you don't first check the length?

editors/code/src/new_project.ts
```diff
@@ -0,0 +1,311 @@
...
+}
```
workingjubilee: I feel like these kinds of "does one thing, only used once" functions mostly make code harder to read.

Veykril (approved):

Veykril:
Formatting failures

joshka:
Iterating on the windows CI failure (I'm not actively testing on windows generally. Leaving that to CI here.

joshka:
The Windows CI failures on this PR turned up three separate issues in the VS Code-side cargo/toolchain resolution changes:

1. We were only checking `$CARGO_HOME/bin/cargo`, not `$CARGO_HOME/bin/cargo.exe`, so the Windows `CARGO_HOME` lookup fell through and returned plain `"cargo"` instead of the resolved executable path.

2. The new project flow passes a copied env object (`{ ...process.env, ...serverExtraEnv }`) into `cargoPath(...)`. On Windows that commonly preserves the path variable as `Path`, not `PATH`, so the original supplied-env lookup could miss Cargo even when it was on the user’s PATH.

3. After fixing the resolution logic, the remaining Windows failure was in the test itself: the assertion compared raw path strings, but the resolved `fsPath` came back with different drive-letter casing (`c:\...` vs `C:\...`). That was not a production bug, just a Windows path-normalization mismatch in the test.

joshka:
@Veykril can this be merged?

joshka:
Thanks!

PR #22104, opened 2026-04-20
perf: Do not check solver's cache validity on every access

ChayimFriedman2:
This should help speed.

Instead, we do it only when an `Interner` is constructed, since an interner cannot be held across revisions.

There is a problem, though: an interner *can* be held across different databases. To solve that, we also reinit the cache when attaching databases, and take the assumption that everything significant (i.e. that can access the cache) will need to attach the db. This is somewhat risky, but we'll take it for the speed.

Theoretically we could *only* reinit on db attach, but it's less risky this way, and with-crate interner construction is rare.


ChayimFriedman2:
This would really benefit from perf run like rustc, but we only have after-merge perf stats... So I'll just bite the bullet and merge this.

PR #22108, opened 2026-04-20
update lsp-types to 0.97

BenjaminBrienen:
Alternative to: https://github.com/rust-lang/rust-analyzer/pull/22115


BenjaminBrienen:
I don't know how to fix the slow tests. The issue seems to be the testdir not being matched and elided.

PR #22112, opened 2026-04-20
Simplify `drop_maybe_dangle` function.

WeiTheShinobi:
Follow `FIXME`, simplify the `drop_maybe_dangle` function.

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -881,7 +881,6 @@ impl Module {
             }
 
             let drop_maybe_dangle = (|| {
-                // FIXME: This can be simplified a lot by exposing hir-ty's utils.rs::Generics helper
                 let trait_ = trait_?;
                 let drop_trait = interner.lang_items().Drop?;
                 if drop_trait != trait_.into() {


ChayimFriedman2:
Two things:

 - This FIXME is outdated. This used to be bad, but currently it's not terrible at all.
 - Even if it was not, the solution is not to blindly move this function to `hir-ty`. We should keep the crate separate.

You can remove the FIXME.

WeiTheShinobi:
> Two things:
> 
> * This FIXME is outdated. This used to be bad, but currently it's not terrible at all.
> * Even if it was not, the solution is not to blindly move this function to `hir-ty`. We should keep the crate separate.
> 
> You can remove the FIXME.

I understand now, thanks!

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22112/4429a9c91ae01b2f260c242bb9f383bf722a1b4a..c2f5dab6fd1c408e589f0d22b25ef853674afe27)*

PR #22115, opened 2026-04-20
switch out lsp-types for gen-lsp-types

BenjaminBrienen:
Blocked on:
- https://github.com/ribru17/gen-lsp-types/issues/8
- https://github.com/ribru17/gen-lsp-types/issues/9
- https://github.com/ribru17/gen-lsp-types/issues/10
- https://github.com/ribru17/gen-lsp-types/pull/11
- https://github.com/ribru17/gen-lsp-types/pull/14

Alternative to: https://github.com/rust-lang/rust-analyzer/pull/22108


crates/rust-analyzer/src/bin/main.rs
ribru17: I think this special logic can be removed, since the typo does not exist in gen-lsp-types

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -2669,35 +2639,6 @@ fn crate_path(root_file_path: &VfsPath) -> Option<VfsPath> {
...
-
-    Ok(())
-}
-
-fn resolve_resource_op(op: &lsp_types::DocumentChange) -> ResourceOperationKind {
-    match op {
-        lsp_types::DocumentChange::CreateFile(_) => ResourceOperationKind::Create,
-        lsp_types::DocumentChange::RenameFile(_) => ResourceOperationKind::Rename,
-        lsp_types::DocumentChange::DeleteFile(_) => ResourceOperationKind::Delete,
-        lsp_types::DocumentChange::TextDocumentEdit(_) => todo!(),
```
DropDemBits: @BenjaminBrienen  `DocumentChange::TextDocumentEdit` client support is checked via `workspace.workspaceEdit.documentChanges`, but I think that's out of scope for this since according to the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#workspaceEdit), `workspace.workspaceEdit.resourceOperations` should imply support for the latter? Or at the very least, we always assume `TextDocumentEdit`s are [always supported](https://github.com/BenjaminBrienen/rust-analyzer/blob/039ac844f1c6ef78472f9790f2ff05376d3dcdbc/crates/rust-analyzer/src/lsp/to_proto.rs#L1368).

---

The LSP metamodel seems to also not define a ResourceOperationKind enum, which is also why the enum ends up not existing in the first place...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/8954b66d43225e62c92e8bbcc8500191b5cceb1e..b996006e50c828eaeaeb0dfebc08237a0959d68f)*

BenjaminBrienen: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#resourceOperationKind

ribru17:
Heads up @BenjaminBrienen I just pushed v0.4.0, will probably affect this PR

BenjaminBrienen:
I already updated to it :)

BenjaminBrienen:
I'm assuming you would like a side of squash with this meal

ribru17:
Fyi, the `lsp-types` issues fixed in `gen-lsp-types` are:

- https://github.com/gluon-lang/lsp-types/issues/310
- https://github.com/gluon-lang/lsp-types/issues/308
- https://github.com/gluon-lang/lsp-types/issues/284
- https://github.com/gluon-lang/lsp-types/issues/278
- https://github.com/gluon-lang/lsp-types/issues/277
- https://github.com/gluon-lang/lsp-types/issues/260
- https://github.com/gluon-lang/lsp-types/issues/245
- https://github.com/gluon-lang/lsp-types/issues/93

Along with: https://github.com/rust-lang/rust-analyzer/blob/4c57da3b27bf5dd3fdcbddd4ae82e1a27e5dc776/crates/rust-analyzer/src/bin/main.rs#L218

crates/rust-analyzer/src/lsp/utils.rs
```diff
@@ -16,10 +16,8 @@ pub(crate) fn invalid_params_error(message: String) -> LspError {
     LspError { code: lsp_server::ErrorCode::InvalidParams as i32, message }
 }
 
-pub(crate) fn notification_is<N: lsp_types::notification::Notification>(
-    notification: &Notification,
-) -> bool {
-    notification.method == N::METHOD
+pub(crate) fn notification_is<N: lsp_types::Notification>(notification: &Notification) -> bool {
+    notification.method == N::METHOD.to_string()
```
DropDemBits: nit:
```suggestion
    notification.method.as_str() == N::METHOD.as_str()
```

crates/rust-analyzer/src/lsp/utils.rs
```diff
@@ -39,37 +37,34 @@ impl Progress {
...
-                    else { return };
-                    if let Ok(Some(_item)) = crate::from_json::<
-                        <lsp_types::request::ShowMessageRequest as lsp_types::request::Request>::Result,
-                    >(
-                        lsp_types::request::ShowMessageRequest::METHOD, &result
+                    else {
+                        return;
+                    };
+                    if let Ok(Some(_item)) = crate::from_json::<Option<MessageActionItem>>(
+                        lsp_types::ShowMessageRequest::METHOD.to_string(),
```
DropDemBits:
```suggestion
                        lsp_types::ShowMessageRequest::METHOD.as_str(),
```

crates/rust-analyzer/src/config.rs
```diff
@@ -2657,8 +2657,6 @@ impl Config {
     }
 
     pub fn snippet_cap(&self) -> Option<SnippetCap> {
-        // FIXME: Also detect the proposed lsp version at caps.workspace.workspaceEdit.snippetEditSupport
-        // once lsp-types has it.
```
BenjaminBrienen, line 2679: I already have a draft PR based on top of this one that implements the split and uses the capability properly, so I just removed the comment for now. Is that fine?

BenjaminBrienen: https://github.com/rust-lang/rust-analyzer/pull/22088

crates/rust-analyzer/src/lsp/semantic_tokens.rs
```diff
@@ -1,170 +1,245 @@
...
+impl fmt::Display for SupportedModifiers {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let string: String = self.into();
+        write!(f, "{string}")
     }
-];
+}
+
+const LAST_STANDARD_MOD: usize =
+    count_tts!(ASYNC DOCUMENTATION Declaration Static DefaultLibrary Deprecated);
```
DropDemBits: This could probably be
```suggestion
const STANDARD_MOD: [SupportedModifiers; _] = [
	SupportedModifiers::Async,
	SupportedModifiers::Documentation,
	SupportedModifiers::Declaration,
	SupportedModifiers::Static,
	SupportedModifiers::DefaultLibrary,
	SupportedModifiers::Deprecated,
];
const LAST_STANDARD_MOD: usize = STANDARD_MOD.len() - 1;
```
to ensure that the names match the actual enum definitions?

crates/rust-analyzer/src/lsp/ext.rs
```diff
@@ -32,11 +31,14 @@ pub enum InternalTestingFetchConfigResponse {
     CheckWorkspace(bool),
 }
 
-impl Request for InternalTestingFetchConfig {
+impl Request for InternalTestingFetchConfigRequest {
     type Params = InternalTestingFetchConfigParams;
     // Option is solely to circumvent Default bound.
     type Result = Option<InternalTestingFetchConfigResponse>;
```
DropDemBits, line 37: The option-wrapping seems to not be required anymore, thankfully!

crates/rust-analyzer/src/diagnostics/test_data/rustc_range_map_lsp_position.txt
```diff
@@ -1,16 +1,8 @@
...
-            username: "",
-            password: None,
-            host: None,
-            port: None,
-            path: "/test/crates/test_diagnostics/src/main.rs",
-            query: None,
-            fragment: None,
-        },
+        url: Uri(
+            "/test/crates/test_diagnostics/src/main.rs",
```
DropDemBits: Q: Is the `file` scheme still needed in e.g. VSCode?

At some point in the past it might've been required, but I'm not sure if this is still needed today or if it'll happily accept an unschema'd path.

crates/rust-analyzer/src/handlers/dispatch.rs
```diff
@@ -107,14 +107,14 @@ impl RequestDispatcher<'_> {
...
-        R: lsp_types::request::Request<
+        R: lsp_types::Request<
                 Params: DeserializeOwned + panic::UnwindSafe + Send + fmt::Debug,
                 Result: Serialize + Default,
             > + 'static,
     {
         if !self.global_state.vfs_done {
             if let Some(lsp_server::Request { id, .. }) =
-                self.req.take_if(|it| it.method == R::METHOD)
+                self.req.take_if(|it| it.method == R::METHOD.to_string())
```
DropDemBits:
```suggestion
                self.req.take_if(|it| it.method.as_str() == R::METHOD.as_str())
```

crates/rust-analyzer/src/handlers/dispatch.rs
```diff
@@ -136,14 +136,14 @@ impl RequestDispatcher<'_> {
...
-        R: lsp_types::request::Request<
+        R: lsp_types::Request<
                 Params: DeserializeOwned + panic::UnwindSafe + Send + fmt::Debug,
                 Result: Serialize,
             > + 'static,
     {
         if !self.global_state.vfs_done || self.global_state.incomplete_crate_graph {
             if let Some(lsp_server::Request { id, .. }) =
-                self.req.take_if(|it| it.method == R::METHOD)
+                self.req.take_if(|it| it.method == R::METHOD.to_string())
```
DropDemBits:
```suggestion
                self.req.take_if(|it| it.method.as_str() == R::METHOD.as_str())
```

crates/rust-analyzer/src/handlers/dispatch.rs
```diff
@@ -182,14 +182,14 @@ impl RequestDispatcher<'_> {
...
-        R: lsp_types::request::Request<
+        R: lsp_types::Request<
                 Params: DeserializeOwned + panic::UnwindSafe + Send + fmt::Debug,
                 Result: Serialize + Default,
             > + 'static,
     {
         if !self.global_state.vfs_done {
             if let Some(lsp_server::Request { id, .. }) =
-                self.req.take_if(|it| it.method == R::METHOD)
+                self.req.take_if(|it| it.method == R::METHOD.to_string())
```
DropDemBits:
```suggestion
                self.req.take_if(|it| it.method.as_str() == R::METHOD.as_str())
```

crates/rust-analyzer/src/handlers/dispatch.rs
```diff
@@ -278,11 +278,11 @@ impl RequestDispatcher<'_> {
 
     fn parse<R>(&mut self) -> Option<(lsp_server::Request, R::Params, String)>
     where
-        R: lsp_types::request::Request,
+        R: lsp_types::Request,
         R::Params: DeserializeOwned + fmt::Debug,
     {
-        let req = self.req.take_if(|it| it.method == R::METHOD)?;
-        let res = crate::from_json(R::METHOD, &req.params);
+        let req = self.req.take_if(|it| it.method == R::METHOD.to_string())?;
```
DropDemBits:
```suggestion
        let req = self.req.take_if(|it| it.method.as_str() == R::METHOD.as_str())?;
        let res = crate::from_json(R::METHOD.as_str(), &req.params);
```

crates/rust-analyzer/src/config.rs
DropDemBits: I hadn't known about the draft PR, sorry!

DropDemBits (commented):
Going through all of the `it.method == R::METHOD` changes made me realize that there's no request-equivalent helper function of `notification_is` 😅

My main concern though is if the `file://` scheme is still required though, as I think the tests ensure that we're emitting self-consistent values rather than if those values actually make sense for the LS client.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/4c57da3b27bf5dd3fdcbddd4ae82e1a27e5dc776..9ea32c3f2d7fa1993f571798efa803ff7b0ace79)*

crates/rust-analyzer/src/lsp/semantic_tokens.rs
BenjaminBrienen: I think this is just an artifact of when I expanded a macro that was no longer working for me. I can see if this can be removed.

crates/rust-analyzer/src/diagnostics/test_data/rustc_range_map_lsp_position.txt
BenjaminBrienen: When I run r-a with the `Run Extension (Debug Build)` task, the editor that pops up does not seem to have r-a installed, so I am confused how to test it locally. [Here is the spec:](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#uri)
image

<details>

#### <a href="#uri" name="uri" class="anchor">URI</a>

URI's are transferred as strings. The URI's format is defined in [https://tools.ietf.org/html/rfc3986](https://tools.ietf.org/html/rfc3986)

```
  foo://example.com:8042/over/there?name=ferret#nose
  \_/   \______________/\_________/ \_________/ \__/
   |           |            |            |        |
scheme     authority       path        query   fragment
   |   _____________________|__
  / \ /                        \
  urn:example:animal:ferret:nose
```

We also maintain a node module to parse a string into `scheme`, `authority`, `path`, `query`, and `fragment` URI components. The GitHub repository is [https://github.com/Microsoft/vscode-uri](https://github.com/Microsoft/vscode-uri), and the npm module is [https://www.npmjs.com/package/vscode-uri](https://www.npmjs.com/package/vscode-uri).

Many of the interfaces contain fields that correspond to the URI of a document. For clarity, the type of such a field is declared as a `DocumentUri`. Over the wire, it will still be transferred as a string, but this guarantees that the contents of that string can be parsed as a valid URI.

Care should be taken to handle encoding in URIs. For example, some clients (such as VS Code) may encode colons in drive letters while others do not. The URIs below are both valid, but clients and servers should be consistent with the form they use themselves to ensure the other party doesn't interpret them as distinct URIs. Clients and servers should not assume that each other are encoding the same way (for example a client encoding colons in drive letters cannot assume server responses will have encoded colons). The same applies to casing of drive letters - one party should not assume the other party will return paths with drive letters cased the same as itself.

```
file:///c:/project/readme.md
file:///C%3A/project/readme.md
```

<div class="anchorHolder"><a href="#documentUri" name="documentUri" class="linkableAnchor"></a></div>

```typescript
type DocumentUri = string;
```

There is also a tagging interface for normal non document URIs. It maps to a `string` as well.

```typescript
type URI = string;
```

</details>

crates/rust-analyzer/src/lsp/semantic_tokens.rs
BenjaminBrienen: Oh, I misunderstood the suggestions. Yeah, that looks like a good idea.

crates/rust-analyzer/src/diagnostics/test_data/rustc_range_map_lsp_position.txt
BenjaminBrienen: If we want to go down the path of having `file://` in the values, then I propose that we activate the fluent_uri feature of gen-lsp-types and bite the bullet. Please let me know before I waste time on that and it isn't wanted.

BenjaminBrienen: I can grab most of those changes from here, but I am sure that it will take a few more hours to integrate cleanly and make sure all the tests are right
https://github.com/rust-lang/rust-analyzer/pull/22108

BenjaminBrienen: I got this in my fork project, which makes me lean harder towards using fluent_uri.

```

thread 'LspServer' (924940) panicked at crates/wgsl-analyzer/src/lsp/from_proto.rs:19:65:
called `Result::unwrap()` on an `Err` value: "file:///home/techn0/source/bevy/crates/bevy_ui_render/src/gradient.wgsl"
```

ribru17: Maybe starting with `url` would be best, since it has the most compatibility with the previous `lsp-types` version? Then a follow up to change to `fluent_uri` would be easier to review. I could be wrong though, maybe there is some context I lack which makes this harder

BenjaminBrienen: It is now updated to use url::Url internally

BenjaminBrienen:
~~It is failing because of something to do with the difference between resource operation and document change, I think. It is just one slow test. I couldn't figure out the reason.~~

crates/rust-analyzer/src/lsp/capabilities.rs
```diff
@@ -43,7 +45,9 @@ pub fn server_capabilities(config: &Config) -> ServerCapabilities {
                 Some(SaveOptions::default().into())
             },
         })),
-        hover_provider: Some(HoverProviderCapability::Simple(true)),
+        notebook_document_sync: None,
+        selection_range_provider: Some(SelectionRangeProvider::Bool(true)),
```
ribru17, line 48: nit: can use `true.into()` here if it makes it more readable; same with all the other enum variants here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..c7b031d0dc07e9f77d96fa412ee77fa3eb76d920)*

crates/rust-analyzer/src/lsp/capabilities.rs
```diff
@@ -244,33 +258,34 @@ impl ClientCapabilities {
...
-                        CodeActionKind::QUICKFIX,
-                        CodeActionKind::REFACTOR,
-                        CodeActionKind::REFACTOR_EXTRACT,
-                        CodeActionKind::REFACTOR_INLINE,
-                        CodeActionKind::REFACTOR_REWRITE,
+                        CodeActionKind::Empty,
+                        CodeActionKind::QuickFix,
+                        CodeActionKind::Refactor,
+                        CodeActionKind::Refactor,
+                        CodeActionKind::Refactor,
```
ribru17:
```suggestion
                        CodeActionKind::RefactorExtract,
                        CodeActionKind::RefactorInline,
```
typo?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..c7b031d0dc07e9f77d96fa412ee77fa3eb76d920)*

crates/rust-analyzer/src/lsp/capabilities.rs
BenjaminBrienen: I don't know whether that is more clear what value is being created.

crates/rust-analyzer/src/lsp/capabilities.rs
BenjaminBrienen: I guess there is test coverage missing here

crates/rust-analyzer/src/handlers/dispatch.rs
```diff
@@ -278,11 +278,11 @@ impl RequestDispatcher<'_> {
 
     fn parse<R>(&mut self) -> Option<(lsp_server::Request, R::Params, String)>
     where
-        R: lsp_types::request::Request,
+        R: lsp_types::Request,
         R::Params: DeserializeOwned + fmt::Debug,
     {
-        let req = self.req.take_if(|it| it.method == R::METHOD)?;
-        let res = crate::from_json(R::METHOD, &req.params);
+        let req = self.req.take_if(|it| it.method.as_str() == R::METHOD.as_str())?;
```
Veykril, line 284:
```suggestion
        let req = self.req.take_if(|it| it.method == R::METHOD)?;
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..c7c5cae2dece164b3bca3f3a0ba71fb288dd54ca)*

crates/rust-analyzer/src/lsp/capabilities.rs
Veykril: Yea I prefer the explicit version here as well

crates/rust-analyzer/src/lsp/from_proto.rs
```diff
@@ -12,18 +13,18 @@ use crate::{
...
 
-pub(crate) fn vfs_path(url: &lsp_types::Url) -> anyhow::Result<vfs::VfsPath> {
+pub(crate) fn vfs_path(url: &lsp_types::Uri) -> anyhow::Result<vfs::VfsPath> {
     abs_path(url).map(vfs::VfsPath::from)
 }
 
 pub(crate) fn offset(
     line_index: &LineIndex,
-    position: lsp_types::Position,
+    position: &lsp_types::Position,
```
Veykril: why this change?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..c7c5cae2dece164b3bca3f3a0ba71fb288dd54ca)*

crates/rust-analyzer/src/lsp/utils.rs
```diff
@@ -16,10 +16,8 @@ pub(crate) fn invalid_params_error(message: String) -> LspError {
     LspError { code: lsp_server::ErrorCode::InvalidParams as i32, message }
 }
 
-pub(crate) fn notification_is<N: lsp_types::notification::Notification>(
-    notification: &Notification,
-) -> bool {
-    notification.method == N::METHOD
+pub(crate) fn notification_is<N: lsp_types::Notification>(notification: &Notification) -> bool {
+    notification.method.as_str() == N::METHOD.as_str()
```
Veykril, line 20:
```suggestion
    notification.method == N::METHOD
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..c7c5cae2dece164b3bca3f3a0ba71fb288dd54ca)*

crates/rust-analyzer/src/lsp/utils.rs
```diff
@@ -172,15 +181,26 @@ pub(crate) fn apply_document_changes(
...
+        Err((whole_document, reversed_partial_changes)) => {
+            text = mem::take(&mut whole_document.text);
+            r_partial_changes = reversed_partial_changes;
+        }
+        Ok(partials) => {
+            text = file_contents.to_owned();
+            r_partial_changes = partials;
+        }
+    }
+    if r_partial_changes.is_empty() {
```
Veykril, line 203: Is this restructuring necessary?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..c7c5cae2dece164b3bca3f3a0ba71fb288dd54ca)*

Cargo.lock
Veykril, line 1: Did you bump the lockfile while updating this? There is a lot of noise in here, could you try undoing that if that is the case?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..c7c5cae2dece164b3bca3f3a0ba71fb288dd54ca)*

crates/rust-analyzer/src/lsp/from_proto.rs
```diff
@@ -65,19 +66,19 @@ pub(crate) fn text_range(
...
+    url: &lsp_types::Uri,
 ) -> anyhow::Result<Option<FileId>> {
     snap.url_to_file_id(url)
 }
 
 /// Returns `None` if the file was excluded.
 pub(crate) fn file_position(
     snap: &GlobalStateSnapshot,
-    tdpp: lsp_types::TextDocumentPositionParams,
+    tdpp: &lsp_types::TextDocumentPositionParams,
```
Veykril, line 77: Likewise, why the extra ref here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..c7c5cae2dece164b3bca3f3a0ba71fb288dd54ca)*

crates/rust-analyzer/src/lsp/ext.rs
```diff
@@ -45,12 +47,14 @@ pub struct InternalTestingFetchConfigParams {
...
-pub enum AnalyzerStatus {}
+pub enum AnalyzerStatusRequest {}
 
-impl Request for AnalyzerStatus {
+impl Request for AnalyzerStatusRequest {
     type Params = AnalyzerStatusParams;
     type Result = String;
-    const METHOD: &'static str = "rust-analyzer/analyzerStatus";
+    const METHOD: LspRequestMethod =
+        LspRequestMethod::Custom(Cow::Borrowed("rust-analyzer/analyzerStatus"));
```
Veykril:
```suggestion
        LspRequestMethod::new("rust-analyzer/analyzerStatus");
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..c7c5cae2dece164b3bca3f3a0ba71fb288dd54ca)*

Veykril: and same treatment for the other places here

crates/rust-analyzer/src/lsp/semantic_tokens.rs
```diff
@@ -1,170 +1,247 @@
...
-        NUMBER,
-        OPERATOR,
-        PARAMETER,
-        PROPERTY,
-        STRING,
-        STRUCT,
-        TYPE_PARAMETER,
-        VARIABLE,
-        TYPE,
+impl From<&SupportedType> for String {
```
Veykril: Do we need to `From` impls here? implementing Display should suffice as that gives us `to_string()` for free

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..c7c5cae2dece164b3bca3f3a0ba71fb288dd54ca)*

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -207,7 +207,7 @@ impl GlobalState {
             if matches!(
                 &event,
                 Event::Lsp(lsp_server::Message::Notification(Notification { method, .. }))
-                if method == lsp_types::notification::Exit::METHOD
+                if method == lsp_types::ExitNotification::METHOD.as_str()
```
Veykril, line 210:
```suggestion
                if *method == lsp_types::ExitNotification::METHOD
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..c7c5cae2dece164b3bca3f3a0ba71fb288dd54ca)*

crates/rust-analyzer/src/lsp/utils.rs
BenjaminBrienen: I wanted to eliminate unreachable code and avoid an unwrap

crates/rust-analyzer/src/handlers/dispatch.rs
BenjaminBrienen: can't compare `std::string::String` with `gen_lsp_types::LspRequestMethod`

crates/rust-analyzer/src/lsp/utils.rs
BenjaminBrienen: can't compare `std::string::String` with `gen_lsp_types::LspRequestMethod`

crates/rust-analyzer/src/lsp/ext.rs
BenjaminBrienen: done

crates/rust-analyzer/src/main_loop.rs
BenjaminBrienen: can't compare `std::string::String` with `gen_lsp_types::LspNotificationMethod`

crates/rust-analyzer/src/lsp/semantic_tokens.rs
BenjaminBrienen: done

crates/rust-analyzer/src/lsp/from_proto.rs
BenjaminBrienen: It's something I thought I needed, but then when I didn't, I figured borrowed is better anyway and kept it. Should I undo it?

Cargo.lock
BenjaminBrienen: I have discarded the change and committed the new one after rebuilding. Hopefully that produces the desired change.

crates/rust-analyzer/src/lsp/from_proto.rs
BenjaminBrienen: After experimenting with removing it, I see that it does avoid a clone in `handle_goto_declaration` and in `handle_completion`. Those clones are removed in my change.

crates/rust-analyzer/src/lsp/from_proto.rs
BenjaminBrienen: It seems like removing this does not negatively affect the existing callsites. Especially since this type is `Copy`, I will undo this and update those callsites.

crates/rust-analyzer/src/bin/main.rs
```diff
@@ -207,25 +207,15 @@ fn run_server() -> anyhow::Result<()> {
 
     tracing::info!("InitializeParams: {}", initialize_params);
     let lsp_types::InitializeParams {
+        #[expect(deprecated, reason = "migration TODO")]
```
lnicola: Nit: is this really a TODO? And should we also handle `root_path`? Maybe not, since nobody has complained.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..e55875bea7323e4a9562a2ee1078b2d1b5c0f188)*

BenjaminBrienen: The field is deprecated, so I had to add an expect and I didn't know what else to put. Do you have a suggestion?

lnicola: "compatibility with old clients"?

BenjaminBrienen: done

crates/rust-analyzer/src/lsp/semantic_tokens.rs
```diff
@@ -1,170 +1,241 @@
...
+            SupportedType::SelfKeyword => "selfKeyword".to_owned(),
+            SupportedType::SelfTypeKeyword => "selfTypeKeyword".to_owned(),
+            SupportedType::Semicolon => "semicolon".to_owned(),
+            SupportedType::Static => "static".to_owned(),
+            SupportedType::ToolModule => "toolModule".to_owned(),
+            SupportedType::TypeAlias => "typeAlias".to_owned(),
+            SupportedType::Union => "union".to_owned(),
+            SupportedType::UnresolvedReference => "unresolvedReference".to_owned(),
+        };
+        write!(f, "{string}")
```
lnicola:
```suggestion
        f.write_str(string)
```

crates/rust-analyzer/src/lsp/semantic_tokens.rs
```diff
@@ -1,170 +1,241 @@
...
-        (SELF_TYPE_KEYWORD, "selfTypeKeyword") => KEYWORD,
-        (SEMICOLON, "semicolon"),
-        (STATIC, "static") => VARIABLE,
-        (TOOL_MODULE, "toolModule") => DECORATOR,
-        (TYPE_ALIAS, "typeAlias") => TYPE,
-        (UNION, "union") => TYPE,
-        (UNRESOLVED_REFERENCE, "unresolvedReference"),
+impl fmt::Display for SupportedType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let string: String = match *self {
```
lnicola:
```suggestion
        let string: &'static str = match *self {
            SupportedType::Comment => ::lsp_types::SemanticTokenTypes::Comment.as_str(),
            SupportedType::Decorator => ::lsp_types::SemanticTokenTypes::Decorator.as_str(),
            SupportedType::EnumMember => ::lsp_types::SemanticTokenTypes::EnumMember.as_str(),
            SupportedType::Enum => ::lsp_types::SemanticTokenTypes::Enum.as_str(),
            SupportedType::Function => ::lsp_types::SemanticTokenTypes::Function.as_str(),
            SupportedType::Interface => ::lsp_types::SemanticTokenTypes::Interface.as_str(),
            SupportedType::Keyword => ::lsp_types::SemanticTokenTypes::Keyword.as_str(),
            SupportedType::Macro => ::lsp_types::SemanticTokenTypes::Macro.as_str(),
            SupportedType::Method => ::lsp_types::SemanticTokenTypes::Method.as_str(),
            SupportedType::Namespace => ::lsp_types::SemanticTokenTypes::Namespace.as_str(),
            SupportedType::Number => ::lsp_types::SemanticTokenTypes::Number.as_str(),
            SupportedType::Operator => ::lsp_types::SemanticTokenTypes::Operator.as_str(),
            SupportedType::Parameter => ::lsp_types::SemanticTokenTypes::Parameter.as_str(),
            SupportedType::Property => ::lsp_types::SemanticTokenTypes::Property.as_str(),
            SupportedType::String => ::lsp_types::SemanticTokenTypes::String.as_str(),
            SupportedType::Struct => ::lsp_types::SemanticTokenTypes::Struct.as_str(),
            SupportedType::TypeParameter => ::lsp_types::SemanticTokenTypes::TypeParameter.as_str(),
            SupportedType::Variable => ::lsp_types::SemanticTokenTypes::Variable.as_str(),
            SupportedType::Type => ::lsp_types::SemanticTokenTypes::Type.as_str(),
            SupportedType::Label => ::lsp_types::SemanticTokenTypes::Label.as_str(),
            SupportedType::Angle => "angle",
            SupportedType::Arithmetic => "arithmetic",
            SupportedType::AttributeBracket => "attributeBracket",
            SupportedType::Attribute => "attribute",
            SupportedType::Bitwise => "bitwise",
            SupportedType::Boolean => "boolean",
            SupportedType::Brace => "brace",
            SupportedType::Bracket => "bracket",
            SupportedType::BuiltinAttribute => "builtinAttribute",
            SupportedType::BuiltinType => "builtinType",
            SupportedType::Char => "character",
            SupportedType::Colon => "colon",
            SupportedType::Comma => "comma",
            SupportedType::Comparison => "comparison",
            SupportedType::ConstParameter => "constParameter",
            SupportedType::Const => "const",
            SupportedType::DeriveHelper => "deriveHelper",
            SupportedType::Derive => "derive",
            SupportedType::Dot => "dot",
            SupportedType::EscapeSequence => "escapeSequence",
            SupportedType::FormatSpecifier => "formatSpecifier",
            SupportedType::Generic => "generic",
            SupportedType::InvalidEscapeSequence => "invalidEscapeSequence",
            SupportedType::Lifetime => "lifetime",
            SupportedType::Logical => "logical",
            SupportedType::MacroBang => "macroBang",
            SupportedType::Negation => "negation",
            SupportedType::Parenthesis => "parenthesis",
            SupportedType::ProcMacro => "procMacro",
            SupportedType::Punctuation => "punctuation",
            SupportedType::SelfKeyword => "selfKeyword",
            SupportedType::SelfTypeKeyword => "selfTypeKeyword",
            SupportedType::Semicolon => "semicolon",
            SupportedType::Static => "static",
            SupportedType::ToolModule => "toolModule",
            SupportedType::TypeAlias => "typeAlias",
            SupportedType::Union => "union",
            SupportedType::UnresolvedReference => "unresolvedReference",
        };
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..6c2aba87b245f3aa15ad65d02f513d4cfc6c37e5)*

crates/rust-analyzer/src/lsp/semantic_tokens.rs
```diff
@@ -1,170 +1,241 @@
...
-        (INTRA_DOC_LINK, "intraDocLink"),
-        (LIBRARY, "library"),
-        (MACRO_MODIFIER, "macro"),
-        (MUTABLE, "mutable"),
-        (PROC_MACRO_MODIFIER, "procMacro"),
-        (PUBLIC, "public"),
-        (REFERENCE, "reference"),
-        (TRAIT_MODIFIER, "trait"),
-        (UNSAFE, "unsafe"),
+impl fmt::Display for SupportedModifiers {
```
lnicola, line 219:
```rust
impl fmt::Display for SupportedModifiers {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        let string = match self {
            SupportedModifiers::Async => SemanticTokenModifiers::Async.as_str(),
            SupportedModifiers::Documentation => SemanticTokenModifiers::Documentation.as_str(),
            SupportedModifiers::Declaration => SemanticTokenModifiers::Declaration.as_str(),
            SupportedModifiers::Static => SemanticTokenModifiers::Static.as_str(),
            SupportedModifiers::DefaultLibrary => SemanticTokenModifiers::DefaultLibrary.as_str(),
            SupportedModifiers::Deprecated => SemanticTokenModifiers::Deprecated.as_str(),
            SupportedModifiers::Associated => "associated",
            SupportedModifiers::AttributeModifier => "attribute",
            SupportedModifiers::Callable => "callable",
            SupportedModifiers::Constant => "constant",
            SupportedModifiers::Consuming => "consuming",
            SupportedModifiers::ControlFlow => "controlFlow",
            SupportedModifiers::CrateRoot => "crateRoot",
            SupportedModifiers::Injected => "injected",
            SupportedModifiers::IntraDocLink => "intraDocLink",
            SupportedModifiers::Library => "library",
            SupportedModifiers::MacroModifier => "macro",
            SupportedModifiers::Mutable => "mutable",
            SupportedModifiers::ProcMacroModifier => "procMacro",
            SupportedModifiers::Public => "public",
            SupportedModifiers::Reference => "reference",
            SupportedModifiers::TraitModifier => "trait",
            SupportedModifiers::Unsafe => "unsafe",
        };
        f.write_str(string)
    }
}
```

And drop the `impl From<&SupportedModifiers> for String` above.

crates/rust-analyzer/Cargo.toml
```diff
@@ -53,6 +53,8 @@ semver.workspace = true
 memchr = "2.7.5"
 cargo_metadata.workspace = true
 process-wrap.workspace = true
+strum = "0.28.0"
```
lnicola: Eh, not a fan of this, I'd rather have a quick declarative macro.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..6c2aba87b245f3aa15ad65d02f513d4cfc6c37e5)*

crates/rust-analyzer/src/lib.rs
```diff
@@ -73,7 +75,7 @@ pub use crate::{
 };
 
 pub fn from_json<T: DeserializeOwned>(
-    what: &'static str,
+    what: impl fmt::Display,
```
lnicola:
```suggestion
    what: &str,
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22115/33f097b09ba82406dec061c78fe542c009fc1e7b..6c2aba87b245f3aa15ad65d02f513d4cfc6c37e5)*

crates/rust-analyzer/src/lsp/semantic_tokens.rs
BenjaminBrienen: added

crates/rust-analyzer/src/lsp/semantic_tokens.rs
BenjaminBrienen: added

crates/rust-analyzer/src/lib.rs
BenjaminBrienen: added

crates/rust-analyzer/Cargo.toml
BenjaminBrienen: I am not very good at macros, so I added it hardcoded.

crates/rust-analyzer/src/lsp/semantic_tokens.rs
BenjaminBrienen: done

crates/rust-analyzer/Cargo.toml
lnicola: `strum_macros` too.

lnicola:
```rust
macro_rules! declare_enum {
    ($name:ident {
        $($variant:ident),* $(,)?
    }) => {
        #[repr(u32)]
        #[derive(Debug, PartialEq, Clone, Copy)]
        pub(crate) enum $name {
            $($variant,)*
        }

        impl $name {
            pub(crate) fn iter() -> Iter<'static, Self> {
                static ITEMS: &[$name] = &[
                    $(
                        $name::$variant,
                    )*
                ];
                ITEMS.iter()
            }
        }
    };
}

declare_enum!(SupportedType {
    Comment,
    Decorator,
    EnumMember,
    Enum,
    Function,
    Interface,
    Keyword,
    Macro,
    Method,
    Namespace,
    Number,
    Operator,
    Parameter,
    Property,
    String,
    Struct,
    TypeParameter,
    Variable,
    Type,
    Label,
    Angle,
    Arithmetic,
    AttributeBracket,
    Attribute,
    Bitwise,
    Boolean,
    Brace,
    Bracket,
    BuiltinAttribute,
    BuiltinType,
    Char,
    Colon,
    Comma,
    Comparison,
    ConstParameter,
    Const,
    DeriveHelper,
    Derive,
    Dot,
    EscapeSequence,
    FormatSpecifier,
    Generic,
    InvalidEscapeSequence,
    Lifetime,
    Logical,
    MacroBang,
    Negation,
    Parenthesis,
    ProcMacro,
    Punctuation,
    SelfKeyword,
    SelfTypeKeyword,
    Semicolon,
    Static,
    ToolModule,
    TypeAlias,
    Union,
    UnresolvedReference,
});

declare_enum!(SupportedModifiers {
    Async,
    Documentation,
    Declaration,
    Static,
    DefaultLibrary,
    Deprecated,
    Associated,
    AttributeModifier,
    Callable,
    Constant,
    Consuming,
    ControlFlow,
    CrateRoot,
    Injected,
    IntraDocLink,
    Library,
    MacroModifier,
    Mutable,
    ProcMacroModifier,
    Public,
    Reference,
    TraitModifier,
    Unsafe,
});
```

Assisted by my friendly neighborhood LLM, just to see if it could. It managed to write two separate macros, but couldn't merge them.

BenjaminBrienen: added with some changes

BenjaminBrienen: > `strum_macros` too.

caught by `cargo-machete`, too :)

Veykril (approved):

Veykril:
If you could rebase once more, will get this merged then 🙏 
Thanks!

stothej:
@Veykril If I can ask a question from the peanut gallery... Isn't there some inherent risk with swapping to a library that's less than 2 months old, from an (admittedly unmaintained) stable one? It's also not a drop-in replacement, unfortunately (hence the 2k line diff). As an example, the version in Cargo.toml is already 2 versions out of date (0.4.0 instead of 0.6.0).

I've seen @ribru17 send PRs of this library to a lot of other projects who are depending on gluon's lsp_types - and then referencing this PR as social proof for it going in. That feels gross to me - but OSS libs are a chicken/egg problem, so there's that.

Anyways, just raising a flag as someone who maintains a bunch of OSS projects and LSP servers (using lsp_server) - take this comment with whatever grains of salt are required.

Personally I was hoping that Microsoft's lsprotocol would have gone somewhere, and solved this problem. I've vendored Helix's lsp-types in my projects for now.

Edit: Changed spam to send - as the former sounded too much like a personal attack, while this is a comment about technical stability and long-term viability.

lnicola:
@stothej we're switching from a hand-written crate to an automatically generated one, which is somewhat we've wanted anyway. Even if the author goes away, we can take over or fork it, and it's going to be easier. Most of the 2000-line diff is to test outputs (because fields were missing in the old crate) and different module and struct names.

> and then referencing this PR as social proof for it going in

Which seems totally fair, given that ribru17 listened to our feedback incorporated the changes we've requested.

PR #22122, opened 2026-04-21
fix: Pass proc_macro_cwd to Analysis::from_single_file()

Wilfred:
If the directory that rust-analyzer was started in no longer exists, all semantic highlighting requests crash and the user sees a lot of errors in their editor.

    [Error - 2:31:20 PM] Request textDocument/semanticTokens/full failed.
      Message: request handler panicked: called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
      Code: -32603

This is because Analysis::from_single_file() used std::env::current_dir().unwrap() for proc_maco_cwd, and current_dir() returns Err if the working directory no longer exists.

I've encountered this when the user's project is a filesystem that is unmounted and remounted. I believe it can also happen when the user has multiple workspaces open in their editor and they delete the directory that rust-analyzer happened to be started in.

Instead, pass proc_macro_cwd explicitly to from_single_file. This is more correct, and removes a FIXME.

I've also checked the other usages of std::env::current_dir().unwrap() but all the other cases look like unit tests.

AI: Code was partially written with Claude.

### DIFF
--- a/crates/ide/src/lib.rs
+++ b/crates/ide/src/lib.rs
@@ -70,7 +70,7 @@ use ide_db::ra_fixture::RaFixtureAnalysis;
 use ide_db::{
     FxHashMap, FxIndexSet,
     base_db::{
-        CrateOrigin, CrateWorkspaceData, Env, FileSet, SourceDatabase, VfsPath,
+        AbsPathBuf, CrateOrigin, CrateWorkspaceData, Env, FileSet, SourceDatabase, VfsPath,
         salsa::{Cancelled, Database},
     },
     prime_caches, symbol_index,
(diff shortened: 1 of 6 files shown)


crates/ide/src/lib.rs
```diff
@@ -253,7 +253,7 @@ impl Analysis {
     // Creates an analysis instance for a single file, without any external
     // dependencies, stdlib support or ability to apply changes. See
     // `AnalysisHost` for creating a fully-featured analysis.
-    pub fn from_single_file(text: String) -> (Analysis, FileId) {
+    pub fn from_single_file(text: String, proc_macro_cwd: AbsPathBuf) -> (Analysis, FileId) {
```
ChayimFriedman2: Can make this take `Arc<AbsPathBuf>`, because that's what we have in highlight injection.

ChayimFriedman2 (approved):
One minor request.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22122/8eb5cb9fcc8a016ecdfe900b7ff2671c4b94cb9f..554e361683652dbcebe12bc5c200c26049087b8a)*

crates/ide/src/lib.rs
Wilfred: Done :)

ChayimFriedman2 (approved):

PR #22124, opened 2026-04-21
fix: generate-method skips trait impl blocks when picking insertion site

### LINKED ISSUE #22123: Assist "Generate method" selects current impl block even if it's a trait impl block
"Generate method" assist adds the code to the current impl even if this is a trait implementation.

I believe this is related to #21684 fixing #21550

**rust-analyzer version**: `rust-analyzer 1.95.0 (5980761 2026-04-14)`
**rustc version**: `rustc 1.95.0 (59807616e 2026-04-14)`

```rust
trait Foo {
	fn foo();
}

struct Bar;

impl Bar {
	fn func1() {}

    // (3)
}

impl Foo for Bar {
	fn foo(&self) {
		self.func2(); // (1)
	}

    // (2)
}
```

Using the "Generate `func2` method" assist at `(1)` adds the `func2` method skeleton at `(2)` in the trait implementation block whereas I would expect it to default to `(3)` in the `Bar` struct block implementation.

cc @akashchakrabortymsc-cmd

Brumbelow:
Follow up to rust-lang/rust-analyzer#21684

- The enclosing-impl shortcut accepted any impl whose self_ty matched the adt, so a call to a missing inherent method from inside a trait impl body appended the stub to the trait impl
- This produced code that does not compile (the trait does not declare the generated fn).

The fix:
- Tighten the filter with trait_(db).is_none(); fall back to the existing find_struct_impl path (or new-impl synthesis) for the trait-impl cursor case. 
- Closes rust-lang/rust-analyzer#22123.

### DIFF
--- a/crates/ide-assists/src/handlers/generate_function.rs
+++ b/crates/ide-assists/src/handlers/generate_function.rs
@@ -148,7 +148,9 @@ fn gen_method(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
 
     let enclosing_impl = ctx.find_node_at_offset::<ast::Impl>();
     let cursor_impl = enclosing_impl.filter(|impl_| {
-        ctx.sema.to_def(impl_).map_or(false, |def| def.self_ty(ctx.sema.db).as_adt() == Some(adt))
+        ctx.sema.to_def(impl_).is_some_and(|def| {
+            def.self_ty(ctx.sema.db).as_adt() == Some(adt) && def.trait_(ctx.sema.db).is_none()
+        })
     });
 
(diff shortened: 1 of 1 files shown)


Brumbelow:
I can remove the comments here too (if needed), just force of habit. LMK either way

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22124/0705955879fe720c915e8c9bc571eb2113a7166e..849c42f60e146aaa7628357101856852c8e7bcdc)*

PR #22126, opened 2026-04-21
fix: guard replace_let_with_if_let against malformed input

### LINKED ISSUE #22106: `replace_let_with_if_let` panics on unterminated string literal
when invoking `Replace let with if let` on a `let` binding whose initializer is an unterminated string literal r-a panics inside the assist handler. the syntax tree is parseable but incomplete and `SyntaxFactory::expr_if` calls `ast.then_branch().unwrap()` on the constructed node without guarding against the case where the branch is absent returning `None` for malformed input and causing an unconditional `unwrap` failure.

reproduce:
```rust
let s = "foo
```

place cursor on the `l` of `let` and apply `Replace let with if let`.

expected:
the assist should either transform the binding correctly or report that it cannot handle the incomplete expression, without crashing the server.

actual:
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', crates/syntax/src/ast/syntax_factory/constructors.rs:1141

the language server crashes and the editor loses all diagnostics and completions until it restarts.

env:
- rustc: 1.94.0
- rust-analyzer: 0.3.2854
- OS: Ubuntu
- Editor: VS Code

Brumbelow:
## Changes
- added `let_stmt.semicolon_token()?` guard at the top of the assist handler. 
- A malformed `let` swallows the trailing `;`, which is the reliable signal. Previously that state flowed through to `make::expr_if`, whose re-parse produced an IfExpr without a then-branch and panicked at `constructors.rs:1155`. 
- also added regression test `replace_let_not_applicable_on_unterminated_string`

Closes rust-lang/rust-analyzer#22106

### DIFF
--- a/crates/ide-assists/src/handlers/replace_let_with_if_let.rs
+++ b/crates/ide-assists/src/handlers/replace_let_with_if_let.rs
@@ -37,6 +37,10 @@ use crate::{AssistContext, AssistId, Assists};
 pub(crate) fn replace_let_with_if_let(acc: &mut Assists, ctx: &AssistContext<'_>) -> Option<()> {
     let let_kw = ctx.find_token_syntax_at_offset(T![let])?;
     let let_stmt = let_kw.parent().and_then(ast::LetStmt::cast)?;
+    // Bail on malformed input (e.g. unterminated string literal in the initializer
+    // swallows the trailing `;` and subsequent tokens). `make::expr_if` would
+    // otherwise re-parse a broken format string and panic.
+    let_stmt.semicolon_token()?;
     let init = let_stmt.initializer()?;
     let original_pat = let_stmt.pat()?;
 
@@ -101,7 +105,7 @@ fn let_expr_needs_paren(expr: &ast::Expr) -> bool {
 
 #[cfg(test)]
 mod tests {
-    use crate::tests::check_assist;
+    use crate::tests::{check_assist, check_assist_not_applicable};
 
     use super::*;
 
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
I don't think we should change our assists to deal with broken input that causes `make` to panic.

We have a solution for `make` - we can create the nodes directly not via parsing, i.e. the `quote!` macros, see `make` constructors that use that for examples. However it's really inconvenient and hard to keep with changes to the AST. Ideally we should generate it from our ungram file, but this is a large undertaking. As such, I prefer not to fix this issue for now.

Brumbelow:
> I don't think we should change our assists to deal with broken input that causes `make` to panic.
> 
> We have a solution for `make` - we can create the nodes directly not via parsing, i.e. the `quote!` macros, see `make` constructors that use that for examples. However it's really inconvenient and hard to keep with changes to the AST. Ideally we should generate it from our ungram file, but this is a large undertaking. As such, I prefer not to fix this issue for now.
> 
> Thank you for your interest in contribution anyway!



> I don't think we should change our assists to deal with broken input that causes `make` to panic.
> 
> We have a solution for `make` - we can create the nodes directly not via parsing, i.e. the `quote!` macros, see `make` constructors that use that for examples. However it's really inconvenient and hard to keep with changes to the AST. Ideally we should generate it from our ungram file, but this is a large undertaking. As such, I prefer not to fix this issue for now.
> 
> Thank you for your interest in contribution anyway!

Yes I experimented with direct creation and it required a lot more overhead than I anticipated. 

I will continue testing for alternate routes or solutions to see what I can find. Thanks for review!

PR #22129, opened 2026-04-21
internal: Do not lower normal assignments into `Pat::Path`

ChayimFriedman2:
Lower them as `Pat::Expr` instead.

This is a bit difficult to implement due to the way rustc checks whether a pat is a destructuring assignment, but it gains us the very desirable property that `Pat::Expr` is the only way to have an normal assignment, which simplifies the code and prevents bugs.

### DIFF
--- a/crates/hir-def/src/expr_store/lower.rs
+++ b/crates/hir-def/src/expr_store/lower.rs
@@ -13,6 +13,7 @@ use cfg::CfgOptions;
 use either::Either;
 use hir_expand::{
     HirFileId, InFile, MacroDefId,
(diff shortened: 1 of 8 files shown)


ShoyuVanilla (approved):
Yeah, this has been very irritating 👍 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22129/8954b66d43225e62c92e8bbcc8500191b5cceb1e..9a95464ab19e3eb3467a8b9f84ba420b405e250e)*

PR #22131, opened 2026-04-21
internal: Remove some `Arc`s

ChayimFriedman2:
Unfortunately, this causes `Semantics` to become invariant over its lifetime, which necessitates large changes, but the changes from `hir` downwards are mechanical and could use a skim review.


bit-aloo (approved):
The changes look fine. I have very limited understanding or context prior to this PR. We are adding a separate, specific lifetime to the database types and removing the `Arc`-based variant of SpanMap. Could you please explain?


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22131/322fc570fac21c9ed2de3b0121c24cb87d7088e9..dd46336d2a78a96bfc96fee0834dbe546fd15e98)*

ChayimFriedman2:
We prefer queries to return references because that's cheaper and `Copy`. Old salsa couldn't do that, but new salsa can.

crates/hir-expand/src/db.rs
```diff
@@ -358,10 +359,11 @@ fn parse_or_expand(db: &dyn ExpandDatabase, file_id: HirFileId) -> SyntaxNode {
 
 // FIXME: We should verify that the parsed node is one of the many macro node variants we expect
 // instead of having it be untyped
+#[salsa_macros::tracked(returns(ref))]
```
Veykril: This now has no LRU anymore

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22131/4c57da3b27bf5dd3fdcbddd4ae82e1a27e5dc776..c54e1d40adb82ca389d2c82be5d8eb48704c6816)*

crates/hir/src/semantics.rs
```diff
@@ -167,7 +167,7 @@ pub struct Semantics<'db, DB: ?Sized> {
 
 pub struct SemanticsImpl<'db> {
     pub db: &'db dyn HirDatabase,
-    s2d_cache: RefCell<SourceToDefCache>,
+    s2d_cache: RefCell<SourceToDefCache<'db>>,
```
Veykril, line 169: I take it this is the cause of the invariance? Not much we can do there, and it shouldn't matter either way. We should not be needing covariance for anything I'd assume

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22131/4c57da3b27bf5dd3fdcbddd4ae82e1a27e5dc776..c54e1d40adb82ca389d2c82be5d8eb48704c6816)*

crates/hir-def/src/item_tree.rs
```diff
@@ -174,21 +171,17 @@ pub(crate) fn file_item_tree_query(
...
+            attrs: FxHashMap::default(),
+            small_data: FxHashMap::default(),
+            big_data: FxHashMap::default(),
+            top_attrs: AttrsOrCfg::empty(),
+            vis: ItemVisibilities { arena: ThinVec::new() },
+        }
     } else {
         item_tree.shrink_to_fit();
-        Arc::new(item_tree)
+        item_tree
```
Veykril: Note this might have a small memory regression, empty item trees can occur frequently with macros (though we still wanna do this change either way I think)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22131/4c57da3b27bf5dd3fdcbddd4ae82e1a27e5dc776..c54e1d40adb82ca389d2c82be5d8eb48704c6816)*

crates/hir/src/semantics.rs
ChayimFriedman2: Yep.

crates/hir-expand/src/db.rs
ChayimFriedman2: Right, my mistake.

crates/hir-def/src/item_tree.rs
ChayimFriedman2: I've thought about that, but you know what, I'll check what the memory impact is.

ChayimFriedman2: Okay, this takes a considerable amount of memory (44mb on self), I'll revert this.

lnicola (approved):

PR #22132, opened 2026-04-22
feat: add .new postfix completion based on expected type (rust-lang/r…

### LINKED ISSUE #20851: Generic postfix `new` completion
Imagine the following

```
let other_thing: OtherThing = ...;
let thing: RefCell<OtherThing> = other_thing$0;
```
imo it would be really nice if we could offer a postfix `.new` here that just does the expected thing of producing a `RefCell::new(other_thing$0)` edit that works for any expected type with such a `new` constructor.

Amit5601:
Fixes rust-lang/rust-analyzer#20851 
Added a .new postfix completion to easily construct types when the expected type is known.
Added the postfix_new unit test to verify it triggers correctly.

### DIFF
--- a/crates/ide-completion/src/completions/postfix.rs
+++ b/crates/ide-completion/src/completions/postfix.rs
@@ -3,7 +3,7 @@
 mod format_like;
 
 use base_db::SourceDatabase;
-use hir::{ItemInNs, Semantics};
+use hir::{HirDisplay, ItemInNs, Semantics};
 use ide_db::{
     RootDatabase, SnippetCap,
     documentation::{Documentation, HasDocs},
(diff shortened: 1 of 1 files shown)


crates/ide-completion/src/completions/postfix.rs
```diff
@@ -116,6 +116,16 @@ pub(crate) fn complete_postfix(
     postfix_snippet("dbgr", "dbg!(&expr)", &format!("dbg!(&{receiver_text})")).add_to(acc, ctx.db);
     postfix_snippet("call", "function(expr)", &format!("${{1}}({receiver_text})"))
         .add_to(acc, ctx.db);
+    if let Some(expected_ty) = ctx.expected_type.as_ref() {
+        let ty_name = expected_ty.display(ctx.db, ctx.display_target).to_string();
```
A4-Tacks, line 120: In an ideal situation, this should determine whether there is a `new` method, if there is one parameter, and if the types match


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22132/8954b66d43225e62c92e8bbcc8500191b5cceb1e..a3e74d21c6e0cd9c024f07d8073066f84b3ba1a0)*

A4-Tacks (approved):

PR #22138, opened 2026-04-22
docs: add some intradoc links

ada4a:


crates/syntax/src/lib.rs
```diff
@@ -9,9 +9,9 @@
 //! For more information, see the [RFC]. Current implementation is inspired by
 //! the [Swift] one.
 //!
-//! The most interesting modules here are `syntax_node` (which defines concrete
-//! syntax tree) and `ast` (which defines abstract syntax tree on top of the
-//! CST). The actual parser live in a separate `parser` crate, though the
+//! The most interesting modules here are [`syntax_node`] (which defines concrete
```
ada4a: this one causes
```
warning: public documentation for `syntax` links to private item `syntax_node`
  --> crates/syntax/src/lib.rs:12:45
   |
12 | //! The most interesting modules here are [`syntax_node`] (which defines concrete
   |                                             ^^^^^^^^^^^ this item is private
   |
   = note: this link resolves only because you passed `--document-private-items`, but will break without
   = note: `#[warn(rustdoc::private_intra_doc_links)]` on by default
```
but it looks like the rustdoc CI workflow passes `--document-private-items` anyway -- should I `allow` this lint maybe?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22138/f45c1e29321abf353869ebf753a3b18536dbeb78..a5e63dac4f4feb57de6f9e5b9cdae2ebca7463af)*

ChayimFriedman2: I don't think the link is worth allowing the lint.

ShoyuVanilla (approved):

PR #22139, opened 2026-04-22
minor: Rename unwrap_block to unwrap_branch

A4-Tacks:
Related rust-lang/rust-analyzer#13747

The name `Unwrap block` often seems to cause misunderstandings


ChayimFriedman2:
But it applies to blocks that aren't branches as well.

A4-Tacks:
It can also be understood as a single branch

ChayimFriedman2:
That doesn't make it clearer IMO. Maybe name it "Unwrap branch" only if it is really a branch?

A4-Tacks:
> Maybe name it "Unwrap branch" only if it is really a branch?

Good idea

bit-aloo (approved):

ChayimFriedman2 (approved):

PR #22144, opened 2026-04-22
feat: Feature a "type annotations needed" diagnostic

ChayimFriedman2:
There are two kinds of this diagnostic:

 1 Some constructs, like calling or indexing, require the type to be known *immediately*.
 2. Otherwise, if we cannot resolve the variable at the end of inference, it's an error.

This PR implements only the first kind, which is also simpler to implement. We just need to push an error in `structurally_resolve_type()` (as opposed to `try_structurally_resolve_type()` that never errors) if the type cannot be resolved, and of course also make sure we only call this function where rustc does.

The second case needs to track the origin of infer vars, and it's for another time.

Blocked on https://github.com/rust-lang/rust-analyzer/pull/22141.

### DIFF
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -401,6 +401,9 @@ pub enum InferenceDiagnostic {
     InvalidLhsOfAssignment {
         lhs: ExprId,
     },
+    TypeMustBeKnown {
+        at_point: ExprOrPatId,
+    },
 }
 
 /// A mismatch between an expected and an inferred type.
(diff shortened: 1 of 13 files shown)


Veykril (approved):
Lovely addition

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22144/e7d5b5b5789729501590af74c8d1542ee040f673..4facc672eb45076eff832b8074b69763e67dd838)*

PR #22145, opened 2026-04-22
feat: add diagnostic for E0735

ada4a:

r? @ChayimFriedman2

### DIFF
--- a/crates/hir-ty/src/lower/path.rs
+++ b/crates/hir-ty/src/lower/path.rs
@@ -300,8 +300,10 @@ impl<'a, 'b, 'db> PathLoweringContext<'a, 'b, 'db> {
                 prohibit_generics_on_resolved(GenericArgsProhibitedReason::SelfTy);
 
                 if self.ctx.lowering_param_default.is_some() {
-                    // Generic defaults are not allowed to refer to `Self`.
-                    // FIXME: Emit an error.
+                    let segment = self.current_segment_u32();
+                    self.on_diagnostic(PathLoweringDiagnostic::GenericDefaultRefersToSelf {
+                        segment,
+                    });
                     return false;
                 }
             }
--- a/crates/ide-diagnostics/src/handlers/generic_default_refers_to_self.rs
+++ b/crates/ide-diagnostics/src/handlers/generic_default_refers_to_self.rs
@@ -1,0 +1,43 @@
+use crate::{Diagnostic, DiagnosticCode, DiagnosticsContext};
+
+// Diagnostic: generic-default-refers-to-self
+//
+// This diagnostic is shown when a generic default refers to `Self`
+pub(crate) fn generic_default_refers_to_self(
+    ctx: &DiagnosticsContext<'_>,
+    d: &hir::GenericDefaultRefersToSelf,
+) -> Diagnostic {
+    Diagnostic::new_with_syntax_node_ptr(
+        ctx,
+        DiagnosticCode::RustcHardError("E0735"),
+        "generic parameters cannot use `Self` in their defaults",
+        d.segment.map(Into::into),
+    )
+    .stable()
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::tests::check_diagnostics;
+
+    #[test]
+    fn plain_self() {
+        check_diagnostics(
+            r#"
+struct Foo<T = Self>(T);
+            // ^^^^ error: generic parameters cannot use `Self` in their defaults
+"#,
+        );
+    }
+
+    #[test]
+    fn self_as_generic() {
+        check_diagnostics(
+            r#"
+struct Wrapper<T>(T);
+struct Foo<T = Wrapper<Self>>(T);
+                    // ^^^^ error: generic parameters cannot use `Self` in their defaults
+"#,
+        );
+    }
+}
--- a/crates/hir-ty/src/lower/diagnostics.rs
+++ b/crates/hir-ty/src/lower/diagnostics.rs
@@ -81,6 +81,11 @@ pub enum PathLoweringDiagnostic {
         def: GenericDefId,
         expected_count: u32,
     },
+    /// Generic defaults are not allowed to refer to `Self`.
+    GenericDefaultRefersToSelf {
+        /// Index of the `Self` segment.
+        segment: u32,
+    },
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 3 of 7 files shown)


crates/ide-diagnostics/src/handlers/generic_default_refers_to_self.rs
```diff
@@ -0,0 +1,42 @@
...
+    #[test]
+    fn self_as_generic() {
+        check_diagnostics(
+            r#"
+struct Foo<T = Box<Self>>(T);
+                // ^^^^ error: generic parameters cannot use `Self` in their defaults
+"#,
+        );
+    }
+}
```
ada4a, line 43: This test currently fails. This is because the place where the error is emitted only covers the simplest case of the generic default _being_ `Self` (as shown in `plain_self`). I wasn't able to (easily) find the place where the error for this more complex case ought to be emitted, and because of that, I'm not actually sure the current structure of `struct GenericDefaultRefersToSelf` is future-proof: it refers to _a_ segment index, but if our original `path` is `Box<Self>`, then `segment_idx=0` would be incorrect if applied on the whole path (it would return `Box<Self>` instead of `Self` I imagine). Because of that, I'm not sure if we should merge this as is, or first try to tackle the complex case as well.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22145/322fc570fac21c9ed2de3b0121c24cb87d7088e9..1ff13b94a9daa372bf73d9dbb473348d5d6fdfbf)*

crates/ide-diagnostics/src/handlers/generic_default_refers_to_self.rs
```diff
@@ -0,0 +1,42 @@
...
+    ctx: &DiagnosticsContext<'_>,
+    d: &hir::GenericDefaultRefersToSelf,
+) -> Diagnostic {
+    Diagnostic::new_with_syntax_node_ptr(
+        ctx,
+        DiagnosticCode::RustcHardError("E0735"),
+        "generic parameters cannot use `Self` in their defaults",
+        d.segment.map(Into::into),
+    )
+    .stable()
```
ada4a, line 16: I didn't implement a fixit for this, because rustc also doesn't have one -- is that okay?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22145/322fc570fac21c9ed2de3b0121c24cb87d7088e9..1ff13b94a9daa372bf73d9dbb473348d5d6fdfbf)*

crates/hir-ty/src/lower/path.rs
```diff
@@ -300,8 +300,10 @@ impl<'a, 'b, 'db> PathLoweringContext<'a, 'b, 'db> {
                 prohibit_generics_on_resolved(GenericArgsProhibitedReason::SelfTy);
 
                 if self.ctx.lowering_param_default.is_some() {
-                    // Generic defaults are not allowed to refer to `Self`.
-                    // FIXME: Emit an error.
+                    let segment = self.current_segment_u32();
+                    self.on_diagnostic(PathLoweringDiagnostic::GenericDefaultRefersToSelf {
+                        segment,
+                    });
                     return false;
```
ada4a, line 307: Apparently `ide`'s `hover::tests::hover_struct_default_arg_self` fails if I remove this line. I don't really understand why it's necessary to `return false;` in this case but not after `prohibit_generics_on_resolved`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22145/322fc570fac21c9ed2de3b0121c24cb87d7088e9..2736324207decea37ad5926a7247b4c9bef4fa09)*

crates/ide-diagnostics/src/handlers/generic_default_refers_to_self.rs
ChayimFriedman2: We aren't bound to rustc's fixes, but we can always start without a fix.

crates/ide-diagnostics/src/handlers/generic_default_refers_to_self.rs
ChayimFriedman2: Nested `Self` should work fine, but the test fails because `Box` isn't defined. Our tests don't have access to std, see https://rust-analyzer.github.io/book/contributing/testing.html.

crates/hir-ty/src/lower/path.rs
ChayimFriedman2: The `false` tells us we should throw the resolution and instead put an error type. On items with prohibited generics, nothing happens if we return `true` - since they don't have generic args, the resolution isn't even used. But if we return `true` here the `Self` will actually find its way down the line, and that can cause errors and even panics later when we try to instantiate the defaults.

ChayimFriedman2:
Please don't assign PRs to me just because I have authored their issue. I review PRs when I have time and will. Only assign to me if I had previous interaction with the PR or its contents, or had explicitly expressed interest in reviewing it.

crates/ide-diagnostics/src/handlers/generic_default_refers_to_self.rs
ada4a: Right.. I keep forgetting about that. I wonder if the test suite could detect mistakes like this :thinking:

crates/ide-diagnostics/src/handlers/generic_default_refers_to_self.rs
ada4a: I guess it's also that there isn't a clear fix in this case -- sure, `= Self` we can maybe just remove, but with something more complex, I'm not convinced suggesting a removal is the right call

ada4a:
> Please don't assign PRs to me just because I have authored their issue. I review PRs when I have time and will. Only assign to me if I had previous interaction with the PR or its contents, or had explicitly expressed interest in reviewing it.

Oh, sorry, thanks for letting me know. Should I reroll (`r?`) then? Or, more generally, should I explicitly `r?`, or will someone come look at the PR by themselves?

ChayimFriedman2:
I don't want to be explicitly assigned, I will look at PRs when I want. Other members may have different wishes.

ChayimFriedman2:
But you don't need to reroll this PR.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22145/322fc570fac21c9ed2de3b0121c24cb87d7088e9..f71d1fe76b45c5736ca1f16984b0ed8d82855877)*

PR #22148, opened 2026-04-23
minor: filter default bounds for introduce_named_type_parameter

A4-Tacks:
Example
---
```rust
fn foo(bar: $0impl Sized) {}
```

**Before this PR**

```rust
fn foo<$0S: Sized>(bar: S) {}
```

**After this PR**

```rust
fn foo<$0S>(bar: S) {}
```

### DIFF
--- a/crates/ide-assists/src/handlers/introduce_named_type_parameter.rs
+++ b/crates/ide-assists/src/handlers/introduce_named_type_parameter.rs
@@ -48,7 +48,8 @@ pub(crate) fn introduce_named_type_parameter(
             )
             .for_impl_trait_as_generic(&impl_trait_type);
 
-            let type_param = make.type_param(make.name(&type_param_name), Some(type_bound_list));
+            let type_bound_list = non_default_bounds(&type_bound_list).then_some(type_bound_list);
+            let type_param = make.type_param(make.name(&type_param_name), type_bound_list);
             let new_ty = make.ty(&type_param_name);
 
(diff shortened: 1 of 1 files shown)


bit-aloo (approved):

crates/ide-assists/src/handlers/introduce_named_type_parameter.rs
```diff
@@ -63,6 +64,10 @@ pub(crate) fn introduce_named_type_parameter(
     )
 }
 
+fn non_default_bounds(bounds: &ast::TypeBoundList) -> bool {
+    bounds.bounds().collect_array().is_none_or(|[bound]| bound.syntax().text() != "Sized")
```
ChayimFriedman2, line 68: A bit better to resolve the path and not do it textually.

ChayimFriedman2 (approved):
LGTM if you prefer the current way.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22148/7bf346ebb7ab5e64782dac47426ebd14e1cef29b..5e48ae4ebdc29229317a00ab3e971b8ac27dd120)*

crates/ide-assists/src/handlers/introduce_named_type_parameter.rs
A4-Tacks: I think this can solve 99% of cases and is very lightweight

PR #22157, opened 2026-04-23
Fix: Correct quickfix position for if-else type mismatch

### LINKED ISSUE #22152: TypeMismatch quickfix position error
**rust-analyzer version**: rust-analyzer 0.3.2870-standalone

**rustc version**: rustc 1.97.0-nightly (913e4bea8 2026-04-22)

**editor or extension**: Vim

```rust
fn main() {
    if true { "".to_owned() } else { "" };
    if true { "".to_owned() } else { "" }.to_owned(); // fixed
    if false { "" } else { "".to_owned() };
    if false { "" } else &{ "".to_owned() }; // fixed
}
```

Similar to #19428, but not a macro, which may be easier to fix

ghost:
## Problem
When a type mismatch occurs in if-else expressions, the quickfix was incorrectly pointing to the position of the if token instead of the actual source of the mismatch in the else branch.

This led to misleading diagnostics, as the reported location did not match the code causing the error.

## Solution
- Use the existing else_branch() (when available) to determine the correct location of the mismatch
- Update the diagnostic range to else_branch().syntax().text_range() for more accurate highlighting
- Retain the previous behavior (using the if token) when no else branch is present

## Testing
- All 49 type_mismatch tests pass ✅
- cargo build -p ide-diagnostics: Success ✅

## AI Disclosure
Used GitHub Copilot for assistance

Fixes rust-lang/rust-analyzer#22152

### DIFF
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -30,18 +30,18 @@ pub(crate) fn type_mismatch(
     let display_range = adjusted_display_range(ctx, d.expr_or_pat, &|node| {
         let Either::Left(expr) = node else { return None };
         let salient_token_range = match expr {
-   ast::Expr::IfExpr(it) => {
-    // Determine the appropriate text range for if-expression type mismatch diagnostics.
-    // In if-else expressions, type mismatches commonly arise from branches with
-    // incompatible types. By prioritizing the else branch range when present,
-    // we ensure diagnostics and quickfixes highlight the actual error location
-    // rather than just the `if` keyword, improving user experience.
-    if let Some(else_branch) = it.else_branch() {
-        else_branch.syntax().text_range()
-    } else {
-        it.if_token()?.text_range()
-    }
-},
+            ast::Expr::IfExpr(it) => {
+                // Determine the appropriate text range for if-expression type mismatch diagnostics.
+                // In if-else expressions, type mismatches commonly arise from branches with
+                // incompatible types. By prioritizing the else branch range when present,
+                // we ensure diagnostics and quickfixes highlight the actual error location
+                // rather than just the `if` keyword, improving user experience.
+                if let Some(else_branch) = it.else_branch() {
+                    else_branch.syntax().text_range()
+                } else {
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
Please note that we require disclosure of AI tools, if you used them. See https://github.com/rust-lang/rust-analyzer/blob/master/CONTRIBUTING.md#use-of-ai-tools.

crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -30,7 +30,18 @@ pub(crate) fn type_mismatch(
     let display_range = adjusted_display_range(ctx, d.expr_or_pat, &|node| {
         let Either::Left(expr) = node else { return None };
         let salient_token_range = match expr {
-            ast::Expr::IfExpr(it) => it.if_token()?.text_range(),
+            ast::Expr::IfExpr(it) => {
+                // Determine the appropriate text range for if-expression type mismatch diagnostics.
+                // In if-else expressions, type mismatches commonly arise from branches with
+                // incompatible types. By prioritizing the else branch range when present,
+                // we ensure diagnostics and quickfixes highlight the actual error location
+                // rather than just the `if` keyword, improving user experience.
```
ChayimFriedman2: This comment is redundant, or at least could be expressed much more shortly.

ChayimFriedman2 (changes requested):
Please add a test.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22157/4c57da3b27bf5dd3fdcbddd4ae82e1a27e5dc776..91aed1718e46df534b11aee780916227caeac312)*

crates/ide-diagnostics/src/handlers/type_mismatch.rs
ghost: Done

ChayimFriedman2:
The test added shows complete misunderstanding, common for AI tools. You are apparently not reviewing the AI output well enough, despite me linking to our contributing guidelines that require doing so.

I'm closing this PR. If you will repeat this behavior we'll reach out to moderators.

PR #22158, opened 2026-04-23
Remove ted from pretify macro expansion

bit-aloo:

### DIFF
--- a/crates/syntax-bridge/src/prettify_macro_expansion.rs
+++ b/crates/syntax-bridge/src/prettify_macro_expansion.rs
@@ -3,8 +3,7 @@ use syntax::{
     NodeOrToken,
     SyntaxKind::{self, *},
     SyntaxNode, SyntaxToken, T, WalkEvent,
-    ast::make,
-    ted::{self, Position},
+    syntax_editor::{Position, SyntaxEditor},
 };
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


crates/syntax-bridge/src/prettify_macro_expansion.rs
```diff
@@ -93,24 +104,42 @@ pub fn prettify_macro_expansion(
...
+                    .prev_token()
+                    .filter(|it| it.kind() == WHITESPACE)
+                    .is_some_and(|it| it.text().contains('\n'));
+                if !has_newline_before {
+                    if indent > 0 {
+                        mods.push(do_indent(before, tok, indent));
+                    }
+                    mods.push(do_nl(before, tok));
                 }
-                mods.push(do_nl(before, tok));
```
A4-Tacks, line 99: Why not maintain the original logic? Is this necessary for migration?
This function has hardly been tested, so is not easy to verify breaking

BTW, I added some tests on #22058


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22158/4c57da3b27bf5dd3fdcbddd4ae82e1a27e5dc776..4e795f3e4e830306db3bffdbec889fa3ff9485fe)*

bit-aloo: Its not a logic change mostly whitespace handling, which we could probably improve. I added those adjustments based on errors I encountered in tests while working on this.

I rebased my changes on top of your PR, and it works. I also noticed we’re using `ted` in `check_pretty` in the tests we should update that to use the editor variant. I will mention this on the PR.

bit-aloo:
closing in favor of: https://github.com/rust-lang/rust-analyzer/pull/22058

PR #22177, opened 2026-04-25
fix: surface generic substitution when hovering inside a call's arguments

### LINKED ISSUE #22164: Generic substitution hover broken for functions with arguments
**rust-analyzer version**: (eg. output of "rust-analyzer: Show RA Version" command, accessible in VSCode via <kbd>Ctrl/⌘</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd>) rust-analyzer version: 0.3.2870-standalone

**rustc version**: (eg. output of `rustc -V`) rustc 1.97.0-nightly (9ec5d5f32 2026-04-21)

**editor or extension**: (eg. VSCode, Vim, Emacs, etc. For VSCode users, specify your extension version; for users of other editors, provide the distribution if applicable) VS Code extension

----

There's a very handy feature for showing inferred generic args. 
It's kind of hard to find, because you have to hover not over function that's called, but right over `()` in that function and it shows that compiler has figured `Field` type for the generic variable.

Image

<details>
<summary>Working example: generic function with no parameters</summary>

```rust
trait Discriminator {
    fn make() -> Self;
}

fn get_pair<V: Discriminator>() -> Result<(V, i32), String> {
    Ok((V::make(), 0))
}

enum Field {
    A,
    B,
}

impl Discriminator for Field {
    fn make() -> Self {
        Field::A
    }
}

fn main() -> Result<(), String> {
    match get_pair()? {
        (Field::A, _) => println!("got A"),
        (Field::B, _) => println!("got B"),
    }
    Ok(())
}
```
</details>

But where it breaks down, if function has an argument, then it seems impossible to bring up this diagnostic. Even if I place the cursor exactly after `(` and hit Ctrl-K Ctrl-I (VS Code hotkey to trigger hovering from the keyboard), I'll get `i32` - argument type.

A workaround - insert `let l = get_pair(42)?; match l` and hover over `l` instead.

Image

<details>
<summary>Broken example: generic function with parameters</summary>

```rust
trait Discriminator {
    fn make() -> Self;
}

fn get_pair<V: Discriminator>(seed: i32) -> Result<(V, i32), String> {
    Ok((V::make(), seed))
}

enum Field {
    A,
    B,
}

impl Discriminator for Field {
    fn make() -> Self {
        Field::A
    }
}

fn main() -> Result<(), String> {
    match get_pair(42)? {
        (Field::A, _) => println!("got A"),
        (Field::B, _) => println!("got B"),
    }
    Ok(())
}
```
</details>

BootstrapperSBL:
Fixes rust-lang/rust-analyzer#22164.

Hovering inside the parentheses of a generic function call only surfaced the generic substitution (e.g. `T = i32`) when the call had no arguments — the cursor would land on `(` or `)` and `pick_best_token` would pick the paren. Once the call had arguments, the cursor landed on the argument literal/expression instead, and the substitution information became unreachable from inside the parens (the user's reported workaround was assigning the call to a `let` binding and hovering on the binding).

This adds a small post-step in `hover_offset`: when the cursor token is descended from an `ArgList` of a `CallExpr` or `MethodCallExpr`, also produce the hover for the called function itself (the same hover as if the cursor were on the function's name). It's gated on the call having a non-trivial generic substitution (at least one substituted type that isn't `unknown`), so the existing `hover_call_parens` style behaviour for non-generic calls is preserved. The per-argument hover (`i32` for `42`, etc.) is preserved and merged with the new block via the standard `\n\n---\n` joiner.

Tested with new cases in `crates/ide/src/hover/tests.rs::subst_inside_call_arg_list` covering:
- cursor inside `()` of a no-argument generic call (existing behaviour preserved)
- cursor on an argument literal of a generic call with arguments (the bug)
- cursor right before `)` after the last argument
- nested calls (inner call's substitution wins)
- generic method calls

### DIFF
--- a/crates/ide/src/hover.rs
+++ b/crates/ide/src/hover.rs
@@ -361,6 +361,64 @@ fn hover_offset(
             };
             render::type_info_of(sema, config, &Either::Left(call_expr), edition, display_target)
         };
+        // When the cursor is inside a call's argument list, also surface the
+        // hover for the called function itself if the call has a generic
+        // substitution. Without this, hovering on an argument like `42` only
+        // shows `i32`, hiding the resolved generic types of the enclosing
+        // call.
+        let enclosing_call_subst = || {
+            let arg_list = token
+                .parent_ancestors()
+                .take_while(|n| !ast::Item::can_cast(n.kind()))
+                .find_map(ast::ArgList::cast)?;
+            let name_ref = match arg_list.syntax().parent().and_then(|p| {
+                syntax::match_ast! {
+                    match p {
+                        ast::CallExpr(call) => {
+                            call.expr().and_then(|e| match e {
+                                ast::Expr::PathExpr(p) => p
+                                    .path()
+                                    .and_then(|p| p.segment())
+                                    .and_then(|s| s.name_ref()),
+                                _ => None,
+                            })
+                        },
+                        ast::MethodCallExpr(call) => call.name_ref(),
+                        _ => None,
+                    }
+                }
+            }) {
+                Some(it) => it,
+                None => return None,
+            };
+            IdentClass::classify_node(sema, name_ref.syntax())?
+                .definitions()
+                .into_iter()
+                .filter(|(_, subst)| {
+                    subst
+                        .as_ref()
+                        .is_some_and(|s| s.types(sema.db).iter().any(|(_, ty)| !ty.is_unknown()))
+                })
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
This is really not needed. If you want to see the generic arguments, hover over the function name and they'll show clearly. If you want to see the return type, select the whole call and hover over it.

PR #22185, opened 2026-04-26
TypeMismatch quickfix position error: `to_owned()`

### LINKED ISSUE #22152: TypeMismatch quickfix position error

WeiTheShinobi:
Fixed rust-lang/rust-analyzer#22152

### DIFF
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -327,8 +327,16 @@ fn str_ref_to_owned(
     let hir::FileRange { file_id, range } = ctx.sema.original_range_opt(expr.syntax())?;
 
     let to_owned = ".to_owned()".to_owned();
-
-    let edit = TextEdit::insert(range.end(), to_owned);
+    let mut edit_expr = expr.clone();
+    while let ast::Expr::BlockExpr(block) = edit_expr.clone() {
+        match block.tail_expr() {
+            Some(tail) => edit_expr = tail,
+            None => break,
+        }
+    }
+    let hir::FileRange { range: edit_range, .. } =
+        ctx.sema.original_range_opt(edit_expr.syntax())?;
+    let edit = TextEdit::insert(edit_range.end(), to_owned);
     let source_change = SourceChange::from_text_edit(file_id.file_id(ctx.sema.db), edit);
     acc.push(fix("str_ref_to_owned", "Add .to_owned() here", source_change, range));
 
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -327,8 +327,16 @@ fn str_ref_to_owned(
     let hir::FileRange { file_id, range } = ctx.sema.original_range_opt(expr.syntax())?;
 
     let to_owned = ".to_owned()".to_owned();
-
-    let edit = TextEdit::insert(range.end(), to_owned);
+    let mut edit_expr = expr.clone();
+    while let ast::Expr::BlockExpr(block) = edit_expr.clone() {
```
A4-Tacks: This is not a real fix
We should fix the emit of diagnostic because `d: &hir::TypeMismatch<'_>` points to `{ "" }` instead of `""`


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22185/9006feed5f7160a2f87710e93a4a3c18faf21df7..94f431ceae5686451954c4f414012931b7146815)*

A4-Tacks (changes requested):

crates/ide-diagnostics/src/handlers/type_mismatch.rs
WeiTheShinobi: I moved the logic to where `TypeMismatch` is created and both cases have been solved,
Or should I fix this further upstream?

crates/hir/src/lib.rs
```diff
@@ -2149,7 +2149,12 @@ impl DefWithBody {
 
         for (pat_or_expr, mismatch) in infer.type_mismatches() {
             let expr_or_pat = match pat_or_expr {
-                ExprOrPatId::ExprId(expr) => source_map.expr_syntax(expr).map(Either::Left),
+                ExprOrPatId::ExprId(mut expr) => {
+                    while let Expr::Block { tail: Some(tail), .. } = &body[expr] {
+                        expr = *tail;
```
A4-Tacks: @ChayimFriedman2 I don't know if it's appropriate this, do you think?


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22185/9006feed5f7160a2f87710e93a4a3c18faf21df7..bde7a6a59336eb8e5f63d60cd0c4aae0542639e0)*

ChayimFriedman2: I don't think this is the place. If we have adjustments to do and we can't do them when emitting the type mismatch, we should do them in ide-diagnostics.

crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -68,7 +68,16 @@ fn fixes(ctx: &DiagnosticsContext<'_, '_>, d: &hir::TypeMismatch<'_>) -> Option<
     let mut fixes = Vec::new();
 
     if let Some(expr_ptr) = d.expr_or_pat.value.cast::<ast::Expr>() {
-        let expr_ptr = &InFile { file_id: d.expr_or_pat.file_id, value: expr_ptr };
+        let file_id = d.expr_or_pat.file_id;
+        let root = ctx.sema.db.parse_or_expand(file_id);
+        let mut expr = expr_ptr.to_node(&root);
+        while let ast::Expr::BlockExpr(block) = expr.clone() {
```
A4-Tacks, line 74: I still hope to change `fn emit_type_mismatch()` or its calling site


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22185/9e457da896bb870275abe7b599ba92072548fc00..4c20a4d196a8549e37808e8034eb21df72c2cc2d)*

PR #22186, opened 2026-04-26
feat: Diagnose trait errors 🎉

ChayimFriedman2:
I.e. type does not implement trait.

There are a lot of other solver errors to be emitted, and the error is pretty simple currently, but this is a huge step!

Watch this wonderful video:

https://github.com/user-attachments/assets/ad562403-3174-42d7-9fcb-aab444946860

### DIFF
--- a/crates/hir-ty/src/tests/regression.rs
+++ b/crates/hir-ty/src/tests/regression.rs
@@ -2363,7 +2363,6 @@ fn test() {
 }
 "#,
         expect![[r#"
-            46..49 'Foo': Foo<N>
             93..97 'self': Foo<N>
             108..125 '{     ...     }': usize
             118..119 'N': usize
--- a/crates/hir-def/src/attrs.rs
+++ b/crates/hir-def/src/attrs.rs
@@ -265,6 +265,12 @@ fn match_attr_flags(attr_flags: &mut AttrFlags, attr: ast::Meta) -> ControlFlow<
                         }
                         _ => {}
                     },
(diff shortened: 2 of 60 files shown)


crates/hir-ty/src/tests/regression.rs
```diff
@@ -2363,7 +2363,6 @@ fn test() {
 }
 "#,
         expect![[r#"
-            46..49 'Foo': Foo<N>
```
ShoyuVanilla, line 2366: I don't think this should be a severe regression/blocker but just out of curiosity, do you happen to know what makes this change?

ShoyuVanilla (approved):
Really exciting, huge progress 🎉 
I'm quite sure the implementation is done right, but I delegate the merge at your disposal as you might want second eyes for this or this might cause some false positives so might be a little bit overwhelming to be merged just before the stable release 😅 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22186/c25feec3dd3ffaef55d02a3cfb383dc7727bd880..26ae4ac4a56f707036d5408b33415b7d1cbf9d64)*

ChayimFriedman2:
The diagnostic is experimental, so false positives are ~fine.

crates/hir-ty/src/tests/regression.rs
ChayimFriedman2: Hmm, no. Weird 😕

PR #22187, opened 2026-04-26
perf: Do not intern `AdtDef`

ChayimFriedman2:
rustc uses this struct to store all information about an ADT, but we have other types and are mainly using it for the solver. So store only the information we need.

### DIFF
--- a/crates/hir-def/src/attrs.rs
+++ b/crates/hir-def/src/attrs.rs
@@ -724,52 +724,56 @@ impl AttrFlags {
             return None;
         }
 
-        return repr(db, owner);
+        Self::repr_assume_has(db, owner)
+    }
 
-        #[salsa::tracked]
-        fn repr(db: &dyn DefDatabase, owner: AdtId) -> Option<ReprOptions> {
-            let mut result = None;
-            collect_attrs::<Infallible>(db, owner.into(), |attr| {
-                let mut current = None;
-                if let ast::Meta::TokenTreeMeta(attr) = &attr
-                    && let Some(path) = attr.path()
-                    && let Some(tt) = attr.token_tree()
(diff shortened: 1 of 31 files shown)


crates/hir-def/src/attrs.rs
```diff
@@ -724,52 +724,56 @@ impl AttrFlags {
...
-                let mut current = None;
-                if let ast::Meta::TokenTreeMeta(attr) = &attr
-                    && let Some(path) = attr.path()
-                    && let Some(tt) = attr.token_tree()
+    /// Only call this when you've verified the type indeed has a `#[repr]` attribute!
+    ///
+    /// Prefer [`AttrFlags::repr()`] in non-perf-sensitive places as it also has a check that
+    /// that the ADT has repr.
+    #[salsa::tracked]
+    pub fn repr_assume_has(db: &dyn DefDatabase, owner: AdtId) -> Option<ReprOptions> {
```
Veykril, line 735: Can we just keep the previous name? `repr_assume_has` as a name does not really serve well as a deterrent imo, and it still does what it did before, handing you the rear either way

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22187/3aaa35b49ef27e15144952aa4f7ba3eecd36fbb4..043a333cfdb6ec0afa4c0fc318ba573a71d657a0)*

Veykril (approved):

crates/hir-def/src/attrs.rs
ChayimFriedman2: We have two methods now, so we can't keep the same name for both. Do you have a better name?

Veykril: Oh, I did not see the nesting from before so I missed the top function. Nevermind then my concern is void sorry

PR #22191, opened 2026-04-27
perf: Provide access to `RootDatabase`'s `LineIndex` for the proc macro protocol

ChayimFriedman2:
Instead of recreating it.

### DIFF
--- a/crates/load-cargo/src/lib.rs
+++ b/crates/load-cargo/src/lib.rs
@@ -609,11 +609,9 @@ impl ProcMacroExpander for Expander {
             SubRequest::LineColumn { file_id, ast_id, offset } => {
                 let range =
                     resolve_sub_span(db, file_id, ast_id, TextRange::empty(TextSize::from(offset)));
-                let source = db.file_text(range.file_id.file_id(db)).text(db);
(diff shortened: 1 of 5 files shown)


crates/load-cargo/src/lib.rs
bit-aloo, line 608: This line can be removed now.

bit-aloo (commented):
LGTM. One suggestion

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22191/3a1d2ce9d20083396c90130ce56ef2b1bd1c8518..b8d2768f4784773c826af7afe29f1d6fe0d56185)*

crates/load-cargo/src/lib.rs
Veykril: Well, its still not incremental tbf

bit-aloo: TBH, I didn’t read the ‘not incremental’ part :P. I was referring to the second part of the statement, considering we’re no longer rebuilding it every time. 

Non-blocker though.

Veykril (approved):

PR #22134, opened 2026-04-22
fix: Port pattern inference from rustc

ChayimFriedman2:
Some additional changes:

- Have `Path` instead of `Option<Path>` for paths in hir. This means when desugaring some constructs that require lang item and the lang item is not present we might miss expressions/patterns, which requires adding minicore items in some tests, but it simplifies the code, so worth it IMO.
 - Port pattern adjustments from rustc (allowing overloaded deref adjustments for deref patterns)
 - Have the infrastructure in place for deref patterns, although they aren't even parsed yet but why remove them when we port from rustc. On the other hand, I removed support for pin patterns, since the feature is more experimental and it was impacting the code more.

Blocked on https://github.com/rust-lang/rust-analyzer/pull/22129.

Fixes https://github.com/rust-lang/rust-analyzer/issues/20149, fixes https://github.com/rust-lang/rust-analyzer/issues/20928, fixes https://github.com/rust-lang/rust-analyzer/issues/19500 (at least, removes one instance of this problem, see the added test).

PR #22195, opened 2026-04-27
fix: Make `InferenceResult::binding_mode()` fallible

ChayimFriedman2:
The pattern can be missing a binding mode when there was a cycle, and the `InferenceResult` is empty.

This should fix the nightly releases.


lnicola:
changelog fixup #22134

PR #22198, opened 2026-04-27
fix: Do not infer signatures, instead infer anon consts in them

ChayimFriedman2:
See [the Zulip discussion](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Anon.20consts.20for.20everything/with/587684630) for details, but in short:
 - This allows us to easily set the expected type.
 - They can have proper MIR an eval, preventing the need for hacks like `fixme_resolve_all_clone()`.

This also fixes a bunch of old inference failures caused by us unable to evaluate constants.

In order to make them visible to the IDE layer, each query that can create anon consts now keeps a list of the anon consts it created. The expression store also gains an ability to find the root expression of something. Then, in `Semantics`, when wanting to find inference info for something, we find its root expr, then search it in the list of anon consts defined by the body.

This should reduce our diagnostics on self by at least two (they're caused by failure to evaluate consts on `to_le_bytes()`, we even have a test for that).

Blocked on https://github.com/rust-lang/rust-analyzer/pull/22194.


crates/hir-ty/src/tests/regression/new_solver.rs
```diff
@@ -34,7 +34,6 @@ impl Space for [u8; 1] {
             223..227 'iter': IntoIter<u8>
             230..231 'a': Vec<u8>
             230..243 'a.into_iter()': IntoIter<u8>
-            322..323 '1': usize
```
ChayimFriedman2, line 37: Previously we inferred all expressions in the signature. Now we skip literals and paths (we don't create an anon const for them) to save time and space. The net effect is that they aren't shown in tests and to IDE, although it's possibly to support them for IDE in little effort.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22198/c2a4de0fe8e7808f07b2db1ecc87238c16bf09ee..0a2eb35cce6b6f1785f20618aca4204952a9a13b)*

crates/ide-assists/src/handlers/add_explicit_type.rs
```diff
@@ -208,14 +208,17 @@ fn main() {
 }
 "#,
         );
-        // note: this may break later if we add more consteval. it just needs to be something that our
-        // consteval engine doesn't understand
-        check_assist_not_applicable(
```
ChayimFriedman2, line 213: It's not easy to support this now, since the type does not contain an error type anymore: it just has an anon const that fails to evaluate, which causes it to fail to normalize. But I've realized this test is no longer needed, since `_` in arrays is supported by now.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22198/c2a4de0fe8e7808f07b2db1ecc87238c16bf09ee..0a2eb35cce6b6f1785f20618aca4204952a9a13b)*

crates/hir-ty/src/lower.rs
```diff
@@ -281,128 +299,45 @@ impl<'db, 'a> TyLoweringContext<'db, 'a> {
...
+            self.interner,
+            self.def,
+            self.store,
+            const_ref.expr,
+            self.resolver,
+            const_type,
+            &|| self.generics(),
+            None,
+            self.forbid_params_after,
+        );
```
nicolas-guichard, line 312: In the `Expr::Underscore` case, because this doesn't provide `infcx` here, `create_anon_const` returns an `Err` which will reach the `unwrap_or` case below. I believe we want to create a new const var instead?
(I tripped on this when rebasing my [related PR](https://github.com/rust-lang/rust-analyzer/pull/22211/changes/763baef5e7ad8b57f9e1c3201c60da72f34c1184) on top of this.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22198/a28bec8594f8a059643d843341ca961bd6a9464c..8e45e71864dbe437be79c26b0e72b927a8db7322)*

ChayimFriedman2: We don't have access to the `InferCtxt` here, only after https://github.com/rust-lang/rust-analyzer/pull/22237 we will.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -3106,6 +3087,31 @@ fn pat_literal_to_hir(lit: &ast::LiteralPat) -> Option<(Literal, ast::Literal)>
 }
 
 impl ExprCollector<'_> {
+    fn with_non_body_expr_root(&mut self, f: impl FnOnce(&mut Self) -> ExprId) -> ExprId {
```
Veykril: Would like a different name here something like
```suggestion
    fn with_fresh_binding_expr_root(&mut self, f: impl FnOnce(&mut Self) -> ExprId) -> ExprId {
```
?
non-body isn't too telling imo

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22198/a28bec8594f8a059643d843341ca961bd6a9464c..8e45e71864dbe437be79c26b0e72b927a8db7322)*

crates/hir-ty/src/next_solver/interner.rs
```diff
@@ -1965,36 +1964,54 @@ impl<'db> Interner for DbInterner<'db> {
...
+            match expr {
+                // The repeat is an anon const.
+                &hir_def::hir::Expr::Array(hir_def::hir::Array::Repeat {
+                    initializer,
+                    repeat: _,
+                }) => append_coroutines_in_expr(db, owner, store, initializer, result),
+                _ => store.walk_child_exprs(expr_id, |expr_id| {
+                    append_coroutines_in_expr(db, owner, store, expr_id, result)
+                }),
+            }
```
Veykril, line 2013: This sounds like we should have a walk_child_exprs version that exhaustively matches and skips anon const places. Otherwise it is easy for us to miss this if a new case arises

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22198/a28bec8594f8a059643d843341ca961bd6a9464c..8e45e71864dbe437be79c26b0e72b927a8db7322)*

crates/hir-ty/src/tests/regression/new_solver.rs
Veykril: How big are the savings here? Imo we should definitely make these work for IDE analysis

crates/hir-def/src/expr_store.rs
```diff
@@ -246,7 +267,7 @@ pub struct ExpressionStoreBuilder {
     pub types: Arena<TypeRef>,
     block_scopes: Vec<BlockId>,
     ident_hygiene: FxHashMap<ExprOrPatId, HygieneId>,
-    pub inference_roots: Option<SmallVec<[(ExprId, RootExprOrigin); 1]>>,
+    inference_roots: Option<SmallVec<[ExprRoot; 1]>>,
```
Veykril, line 270: I think this is always `Some` now?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22198/a28bec8594f8a059643d843341ca961bd6a9464c..8e45e71864dbe437be79c26b0e72b927a8db7322)*

crates/hir-ty/src/db.rs
```diff
@@ -337,3 +409,112 @@ impl InternedCoroutineClosureId {
         Self::new_impl(db, loc)
     }
 }
+
+/// An anonymous const expression that appears in a type position (e.g., array lengths,
+/// const generic arguments like `{ N + 1 }`). Unlike named constants, these don't have
```
Veykril: Aren't const param defaults also anon consts as well as field defaults?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22198/a28bec8594f8a059643d843341ca961bd6a9464c..8e45e71864dbe437be79c26b0e72b927a8db7322)*

Veykril (approved):

crates/hir-ty/src/next_solver/interner.rs
ChayimFriedman2: I want to change the visitors to be based on traits anyway (have a branch for that), that'll be a good place to do this.

crates/hir-def/src/expr_store.rs
ChayimFriedman2: No it is not, we still take it when we're inside an expr.

crates/hir-ty/src/db.rs
ChayimFriedman2: Const param defaults, yes. Field defaults, I don't think we lower those yet.

ChayimFriedman2:
> How big are the savings here? Imo we should definitely make these work for IDE analysis

They are big, 117mb on self. But as I said we can still support them in IDE, I'll work on that in a follow-up PR.

crates/hir-ty/src/db.rs
Veykril: We are given the IDE layer now supports it 
https://github.com/rust-lang/rust-analyzer/blob/1cab8f71386d91aef901b4f1dd29ea81f55a6fa3/crates/hir/src/source_analyzer.rs#L89-L94

PR #22206, opened 2026-04-28
feat(diagnostics): add handler for base expression in struct pattern

ada4a:

### DIFF
--- a/crates/ide-diagnostics/src/handlers/base_expr_in_struct_pattern.rs
+++ b/crates/ide-diagnostics/src/handlers/base_expr_in_struct_pattern.rs
@@ -1,0 +1,65 @@
+use crate::{Diagnostic, DiagnosticCode, DiagnosticsContext};
+
+// Diagnostic: base-expr-in-struct-pattern
+//
+// This diagnostic is triggered if a struct pattern contains a base expression (`..base`).
+pub(crate) fn base_expr_in_struct_pattern(
+    ctx: &DiagnosticsContext<'_>,
+    d: &hir::BaseExprInStructPattern,
+) -> Diagnostic {
+    Diagnostic::new_with_syntax_node_ptr(
+        ctx,
+        DiagnosticCode::SyntaxError,
+        "base expressions aren't allowed in struct patterns",
+        d.node.map(Into::into),
+    )
+    .stable()
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::tests::check_diagnostics;
+
+    #[test]
+    fn spread_variable() {
+        check_diagnostics(
+            r#"
+struct Foo { bar: u32, baz: u32 }
+fn test(f: Foo, g: Foo) {
+    if let Foo { ..g } = f {}
+                // ^ error: base expressions aren't allowed in struct patterns
+    if let Foo { bar: 0, ..g } = f {}
+                        // ^ error: base expressions aren't allowed in struct patterns
+    if let Foo { bar: 0, baz: 0, ..g } = f {}
+                                // ^ error: base expressions aren't allowed in struct patterns
+}
+        "#,
+        );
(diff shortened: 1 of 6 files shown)


crates/ide-diagnostics/src/handlers/base_expr_in_struct_pattern.rs
```diff
@@ -0,0 +1,65 @@
...
+        ctx,
+        DiagnosticCode::SyntaxError,
+        "base expressions aren't allowed in struct patterns",
+        d.node.map(Into::into),
+    )
+    .stable()
+}
+
+#[cfg(test)]
+mod tests {
```
ada4a, line 20: Oh, it turns out I've been running the wrong tests while working on this... But I'm actually not sure what to do with these extraneous errors:
- I can't really ignore syntax errors completely, since the error I'm testing is a syntax error itself
- I guess I could give in and add annotations for them to the test?
 
It also looks like no other handlers are for syntax errors, so nowhere to draw inspiration from..

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22206/c2a4de0fe8e7808f07b2db1ecc87238c16bf09ee..227e177525daf29325ea65aae82ea2d1f4defb08)*

ChayimFriedman2: I think that if we already emit syntax errors here, we don't need another error.

ada4a: fwiw the error from rustc is not quite ideal (imo):
```rs
error: expected `}`, found `g`
 --> src/lib.rs:4:36
  |
4 |     if let Foo { bar: 0, baz: 0, ..g } = f {}
  |            ---                     ^ expected `}`
  |            |
  |            while parsing the fields for this pattern
```

but I don't have a strong opinion on this.

In this case, I'd open a V2 PR which updates the fixme comment to say something like "we wanted to emit a diagnostic on this but decided to rely on rustc instead, since this is a syntax error"

lnicola: @ada4a

I don't think the point is to rely on `rustc`. What happens right now is:

Image

ada4a: Sorry @lnicola, I don't see where you're getting at with this

lnicola: By not adding the diagnostic, we're not relying on `rustc`, we're relying on our parser error (`expected COMMA`).

ChayimFriedman2:
Wait a minute: I just realized that we do have to emit a diagnostic here, just the test is incorrect. This code path handles destructuring assignment, not `let` declaration.

ada4a:
Sorry..

It looks like the case with destructuring assignment is still a syntax error (I think? It doesn't have an error code, anyway), though with a better error message, and even a fix:

code:
```rs
#![allow(unused)]

struct Foo { bar: u32, baz: u32 }
fn test(f: Foo, mut bar: u32) {
    Foo { bar, ..Default::default() } = f;
}
```
diagnostic:
```
error: functional record updates are not allowed in destructuring assignments
 --> src/lib.rs:5:18
  |
5 |     Foo { bar, ..Default::default() } = f;
  |                  ^^^^^^^^^^^^^^^^^^ help: consider removing the trailing pattern

error: could not compile `playground` (lib) due to 1 previous error
```

Should I still add a handler despite this being a syntax error?

ChayimFriedman2:
In rustc, sure, but in rust-analyzer, not yet, and this is where we should add it.

PR #22215, opened 2026-04-30
fix: Fix stack overflow on projection display

### LINKED ISSUE #21697: Stack overflow when rendering completions for impl Trait type
A stack overflow seems to be making rust-analyzer crash when rendering completions for certain impl Trait types.  Increasing `RUST_MIN_STACK` doesn't seem to avoid the problem, and my current stack limit is 8MiB.  So I wonder if there might be some infinite recursion going on for some reason.

Here's a snippet that can hopefully reproduce the problem

```rust
trait SuperTrait {
    type AssocTy;
}

trait SubTrait<T = <Self as SuperTrait>::AssocTy>: SuperTrait {}

// The crash seems to happen when I try using `SubTrait` as an impl Trait type:
fn tryme(param: impl SubTrait) {
    // param
    // ^--- Try typing this
}
```

```
thread 'Worker4' (2877308) has overflowed its stack
fatal runtime error: stack overflow, aborting
```

I first noticed this in NeoVim with the RA version below.  After trying it in VSCode, I believe the issue is also present in at least `0.3.2803-standalone` -- which means I don't think I can blame my NeoVim config *this time* 😅.  I also built a few different RA releases from source and captured some backtraces at the point of the `SIGSEGV`, which I hope will be helpful:

```bash
# Built with rustc 1.93.1 (01f6ddf75 2026-02-11)
cargo build --config profile.dev.debug=2
```

- [2026-02-23](https://github.com/rust-lang/rust-analyzer/releases/tag/2026-02-23): [backtrace-2026-02-23.txt](https://github.com/user-attachments/files/25502649/backtrace-2026-02-23.txt)
- [2026-02-16](https://github.com/rust-lang/rust-analyzer/releases/tag/2026-02-16): [backtrace-2026-02-16.txt](https://github.com/user-attachments/files/25502225/backtrace-2026-02-16.txt)
- [2025-01-06](https://github.com/rust-lang/rust-analyzer/releases/tag/2025-01-06): [backtrace-2025-01-06.txt](https://github.com/user-attachments/files/25502262/backtrace-2025-01-06.txt)

Judging by these, I'm guessing that the infinite recursion is happening in code from `hir_ty::display` and consists of:

1. `<Ty<'_> as HirDisplay<'_>>::hir_fmt()`: https://github.com/rust-lang/rust-analyzer/blob/0c746f699034e70e1c3f11036b8c2895a0b1071c/crates/hir-ty/src/display.rs#L1465-L1472
2. `write_bounds_like_dyn_trait_with_prefix()`: https://github.com/rust-lang/rust-analyzer/blob/0c746f699034e70e1c3f11036b8c2895a0b1071c/crates/hir-ty/src/display.rs#L1783
3. `write_bounds_like_dyn_trait()`: https://github.com/rust-lang/rust-analyzer/blob/0c746f699034e70e1c3f11036b8c2895a0b1071c/crates/hir-ty/src/display.rs#L1890
4. `hir_fmt_generic_arguments()`: https://github.com/rust-lang/rust-analyzer/blob/0c746f699034e70e1c3f11036b8c2895a0b1071c/crates/hir-ty/src/display.rs#L1682
5. `<GenericArg<'_> as HirDisplay<'_>>::hir_fmt()`: https://github.com/rust-lang/rust-analyzer/blob/0c746f699034e70e1c3f11036b8c2895a0b1071c/crates/hir-ty/src/display.rs#L683
6. `<Ty<'_> as HirDisplay<'_>>::hir_fmt()` (different branch): https://github.com/rust-lang/rust-analyzer/blob/0c746f699034e70e1c3f11036b8c2895a0b1071c/crates/hir-ty/src/display.rs#L1272
7. `write_projection()`: https://github.com/rust-lang/rust-analyzer/blob/0c746f699034e70e1c3f11036b8c2895a0b1071c/crates/hir-ty/src/display.rs#L668

   This calls `<Ty<'_> as HirDisplay<'_>>::hir_fmt()` and restarts the cycle at step 1.

Hopefully this is enough (but not too much) info.  If I was more familiar with the code base, I'd happily submit a pull request myself.  But since I'm not, any help would be much appreciated!

---

**rust-analyzer version**: `rust-analyzer 1.93.1 (01f6ddf 2026-02-11)`

**rustc version**: `rustc 1.93.1 (01f6ddf75 2026-02-11)`

**editor or extension**: NeoVim (and VSCode)

ChayimFriedman2:
As it turns out, displaying projections can always cause an infinite loop, even when not recursing into a type param's predicates.

This does regress the previous test of infinite loops a bit, which is unfortunate. Maybe we should have a more fine-grained tracking?

Fixes rust-lang/rust-analyzer#21697.

### DIFF
--- a/crates/hir-ty/src/display.rs
+++ b/crates/hir-ty/src/display.rs
@@ -10,8 +10,8 @@ use std::{
 use base_db::{Crate, FxIndexMap};
 use either::Either;
 use hir_def::{
-    ExpressionStoreOwnerId, FindPathConfig, GenericDefId, GenericParamId, HasModule, LocalFieldId,
-    Lookup, ModuleDefId, ModuleId, TraitId,
+    ExpressionStoreOwnerId, FindPathConfig, GenericDefId, GenericParamId, HasModule,
+    ItemContainerId, LocalFieldId, Lookup, ModuleDefId, ModuleId, TraitId,
     expr_store::{ExpressionStore, path::Path},
     find_path::{self, PrefixKind},
     hir::{
@@ -134,7 +134,15 @@ pub struct HirFormatter<'a, 'db> {
     display_lifetimes: DisplayLifetime,
     display_kind: DisplayKind,
     display_target: DisplayTarget,
-    bounds_formatting_ctx: BoundsFormattingCtx<'db>,
+    /// We can have recursive bounds like the following case:
+    /// ```ignore
+    /// where
+    ///     T: Foo,
+    ///     T::FooAssoc: Baz<<T::FooAssoc as Bar>::BarAssoc> + Bar
+    /// ```
+    /// So, record the projection types met while formatting bounds and
+    /// prevent recursing into their bounds to avoid infinite loops.
+    currently_formatting_bounds: FxHashSet<AliasTy<'db>>,
     /// Whether formatting `impl Trait1 + Trait2` or `dyn Trait1 + Trait2` needs parentheses around it,
     /// for example when formatting `&(impl Trait1 + Trait2)`.
     trait_bounds_need_parens: bool,
@@ -154,34 +162,6 @@ pub enum DisplayLifetime {
     Never,
 }
 
-#[derive(Default)]
-enum BoundsFormattingCtx<'db> {
-    Entered {
-        /// We can have recursive bounds like the following case:
-        /// ```ignore
-        /// where
-        ///     T: Foo,
-        ///     T::FooAssoc: Baz<<T::FooAssoc as Bar>::BarAssoc> + Bar
-        /// ```
-        /// So, record the projection types met while formatting bounds and
-        //. prevent recursing into their bounds to avoid infinite loops.
-        projection_tys_met: FxHashSet<AliasTy<'db>>,
-    },
-    #[default]
-    Exited,
-}
-
-impl<'db> BoundsFormattingCtx<'db> {
-    fn contains(&self, proj: &AliasTy<'db>) -> bool {
-        match self {
-            BoundsFormattingCtx::Entered { projection_tys_met } => {
-                projection_tys_met.contains(proj)
-            }
-            BoundsFormattingCtx::Exited => false,
-        }
-    }
-}
-
 impl<'db> HirFormatter<'_, 'db> {
     pub fn start_location_link(&mut self, location: ModuleDefId) {
         self.fmt.start_location_link(location);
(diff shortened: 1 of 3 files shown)


crates/hir-ty/src/display.rs
```diff
@@ -134,7 +134,15 @@ pub struct HirFormatter<'a, 'db> {
...
     display_target: DisplayTarget,
-    bounds_formatting_ctx: BoundsFormattingCtx<'db>,
+    /// We can have recursive bounds like the following case:
+    /// ```ignore
+    /// where
+    ///     T: Foo,
+    ///     T::FooAssoc: Baz<<T::FooAssoc as Bar>::BarAssoc> + Bar
+    /// ```
+    /// So, record the projection types met while formatting bounds and
+    //. prevent recursing into their bounds to avoid infinite loops.
```
jplatte: This line starts with `//.` instead of `///`, looks to only have moved but should probably be changed ^^

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22215/34f1c5f1c600f3373589868e9416a0236ccfb07d..41dbcdf2bb92836f28c106e403b67a7b96561764)*

crates/hir-ty/src/display.rs
```diff
@@ -195,27 +175,17 @@ impl<'db> HirFormatter<'_, 'db> {
...
-                self.bounds_formatting_ctx = BoundsFormattingCtx::Exited;
-                res
-            }
+    ) -> Result {
+        if self.currently_formatting_bounds.insert(target) {
+            format_bounds(self)?;
+            self.currently_formatting_bounds.remove(&target);
+            Ok(())
+        } else {
+            self.write_str(if self.display_kind.is_source_code() { "_" } else { "…" })
```
ShoyuVanilla: I guess we could print the alias just without bounds as our best effort here but I'm not sure 🤔 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22215/692043f8c5de9fa731ecba6331295f109a05f070..91fccf29072f72492dee76d7401a1bc571ebea7b)*

ShoyuVanilla (approved):

crates/hir-ty/src/display.rs
ChayimFriedman2: I'm not sure it's better, but let's try.

ChayimFriedman2: Done that.

ShoyuVanilla (approved):

PR #22216, opened 2026-04-30
minor: Show running tests on CI

ChayimFriedman2:
So we can debug the test we have that gets stuck.


bit-aloo (approved):

ChayimFriedman2:
Nice, we figured the hanging test 😄

PR #22221, opened 2026-04-30
internal: fix incorrect offset on multiple indel

A4-Tacks:


crates/ide-db/src/text_edit.rs
```diff
@@ -133,7 +133,7 @@ impl TextEdit {
         let mut res = offset;
         for indel in &self.indels {
             if indel.delete.start() >= offset {
-                break;
+                continue;
```
ChayimFriedman2, line 136: This looks correct to me; this function should return the new location of `offset`. if `indel.delete.start() >= offset`, it means we've already passed it so we should return what we've accumulated.

OTOH the `return None` in the next branch looks incorrect to me - it'll lead to early-exit when we shouldn't just because the next indel doesn't came yet. It should be `if indel.delete.contains(offset)`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22221/0f5cbd54634357539d9a5d12453d57d35cee280d..823ac90e6c6cf056776e3d98f8214d77e1ad0cd9)*

ChayimFriedman2 (changes requested):

crates/ide-db/src/text_edit.rs
A4-Tacks: This is actually this:
```
if indel.delete.start() < offset {
    if offset < indel.delete.end() {
        return None; // indel.delete.start() < offset && offset < indel.delete.end()
    }
    res += TextSize::of(&indel.insert);
    res -= indel.delete.len();
}
```
But I agree that the readability is very low

ChayimFriedman2: Right, I got confused.

ChayimFriedman2 (approved):

PR #22225, opened 2026-04-30
fix: check original type for replace_arith_op

A4-Tacks:
Fixup rust-lang/rust-analyzer#22180

The old test did not include minicore, which resulted in the inability to actually apply it

Example
---
```rust
//- minicore: add, builtin_impls
fn main() {
    let mut x = 1;
    x $0+= 2;
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
fn main() {
    let mut x = 1;
    x = x.saturating_add(2);
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/replace_arith_op.rs
+++ b/crates/ide-assists/src/handlers/replace_arith_op.rs
@@ -108,7 +108,7 @@ fn replace_arith(acc: &mut Assists, ctx: &AssistContext<'_, '_>, kind: ArithKind
 
 fn is_primitive_int(ctx: &AssistContext<'_, '_>, expr: &ast::Expr) -> bool {
     match ctx.sema.type_of_expr(expr) {
-        Some(ty) => ty.adjusted().is_int_or_uint(),
+        Some(ty) => ty.adjusted().strip_references().is_int_or_uint(),
         _ => false,
     }
 }
@@ -195,6 +195,7 @@ mod tests {
         check_assist(
             replace_arith_with_checked,
             r#"
+//- minicore: add, builtin_impls
 fn main() {
     let x = 1 $0+ 2;
 }
@@ -212,6 +213,7 @@ fn main() {
         check_assist(
             replace_arith_with_saturating,
             r#"
+//- minicore: add, builtin_impls
 fn main() {
     let x = 1 $0+ 2;
 }
@@ -229,6 +231,7 @@ fn main() {
         check_assist(
             replace_arith_with_wrapping,
             r#"
+//- minicore: add, builtin_impls
 fn main() {
     let x = 1 $0+ 2;
 }
@@ -246,6 +249,7 @@ fn main() {
         check_assist(
             replace_arith_with_wrapping,
             r#"
+//- minicore: add, builtin_impls
 fn main() {
     let x = 1*3 $0+ 2;
 }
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/replace_arith_op.rs
```diff
@@ -195,6 +195,7 @@ mod tests {
         check_assist(
             replace_arith_with_checked,
             r#"
+//- minicore: add, builtin_impls
```
Wilfred, line 198: Dumb question: shouldn't the test behaviour actually change? Why is it sufficient to to just change the minicore comments here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22225/43e333ab4f863c7e83d0c1483d906b25258dc511..364c23eeed84cec868c5aca973928b8a8bb9a075)*

A4-Tacks:
```rust
struct Foo;
struct Bar;
impl std::ops::AddAssign<i8> for Foo {}
fn main() {
    let foo = Foo;
    let bar = Bar;
    foo += 2;
  //^^^ original: Foo, adjusted: Some(&mut Foo)
    bar += 2;
  //^^^ original: Bar, adjusted: None
}
```

crates/ide-assists/src/handlers/replace_arith_op.rs
```diff
@@ -108,7 +108,7 @@ fn replace_arith(acc: &mut Assists, ctx: &AssistContext<'_, '_>, kind: ArithKind
 
 fn is_primitive_int(ctx: &AssistContext<'_, '_>, expr: &ast::Expr) -> bool {
     match ctx.sema.type_of_expr(expr) {
-        Some(ty) => ty.adjusted().is_int_or_uint(),
+        Some(ty) => ty.adjusted().strip_references().is_int_or_uint(),
```
ChayimFriedman2:
```suggestion
        Some(ty) => ty.adjusted().strip_reference().is_int_or_uint(),
```
You cannot add `&&int` to `&&int`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22225/43e333ab4f863c7e83d0c1483d906b25258dc511..364c23eeed84cec868c5aca973928b8a8bb9a075)*

ChayimFriedman2: Oh I see this is because `adjusted` is the input to `add()` - I think this means we need to take `original()` instead (but also strip one reference since `Add<&int> for &int` is implemented). And please add a test for adding references.

ChayimFriedman2 (changes requested):

crates/ide-assists/src/handlers/replace_arith_op.rs
A4-Tacks: > I think this means we need to take `original()`

Are there some strange cases where converting from non-integer or non-integer-references to integers has been rejected

ChayimFriedman2: Adjustments can only change due to coercion (besides taking a reference for `add()`), which is not applicable here.

A4-Tacks: > And please add a test for adding references.

Does #23109 include these tests?

ChayimFriedman2: Yes.

ChayimFriedman2 (approved):

PR #22228, opened 2026-04-30
minor: Clarify that `SmolStr` doesn't allocate when cloning

Enyium:
Technically, `O(1)` doesn't imply this, but it may also not be clear to everyone that it makes this very probable.


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22228/692043f8c5de9fa731ecba6331295f109a05f070..fc9cc17131eeac3373aeca8fe2f31bff12ae4a94)*

PR #22231, opened 2026-04-30
feat: supports multi arms for replace_match_with_if_let

A4-Tacks:
Example
---
```rust
fn main() {
    match$0 Some(0) {
        Some(n) if n % 2 == 0 && n != 6 => (),
        Some(n) if valided(n) => foo(),
        _ => code(),
    }
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
fn main() {
    if let Some(n) = Some(0) && (n % 2 == 0 && n != 6) {
        ()
    } else if let Some(n) = Some(0) && valided(n) {
        foo()
    } else {
        code()
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/replace_if_let_with_match.rs
+++ b/crates/ide-assists/src/handlers/replace_if_let_with_match.rs
@@ -4,7 +4,7 @@ use std::iter::successors;
 
 use ide_db::{RootDatabase, defs::NameClass, ty_filter::TryEnum};
 use syntax::{
-    AstNode, Edition, SyntaxKind, T, TextRange,
+    AstNode, SyntaxKind, T, TextRange,
     ast::{
         self, HasName,
         edit::{AstNodeEdit, IndentLevel},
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/replace_if_let_with_match.rs
```diff
@@ -241,27 +241,19 @@ pub(crate) fn replace_match_with_if_let(
         return None;
     }
 
-    let [first_arm, second_arm] = match_arm_list.arms().collect_array()?;
-    if second_arm.guard().is_some() {
-        return None;
-    }
-    if first_arm.guard().is_some() && ctx.edition() < Edition::Edition2024 {
```
ShoyuVanilla, line 248: Are we okay without this edition check?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22231/0d381ca097a8e0375a19387874d952c0a230ac4f..0c4bb983b095b3808917639e9ae7487fb2ce1218)*

A4-Tacks: I think it may be more reasonable to use compilation errors to drive edition migration

ShoyuVanilla (approved):

PR #22237, opened 2026-05-01
fix: Provide an InferCtxt to TyLoweringContext

ChayimFriedman2:
So that it can create infer vars, instead of inserting them after the fact. We gain:

 - Accurate spans for the vars.
 - Being able to distinguish between valid infer vars the user wrote (or implied) and infer vars caused by errors, which should use dummy spans to prevent false diagnostics. See the removed FIXME in incorrect_generics_len.rs.
 - No need to try and match the infer var to the `TypeRefId` for `type_of_type_placeholder`, which can fail. Instead we just register them when we create the var.

Of course, not every context lowering types has an `InferCtxt`, so it's optional.

And some more small changes.

Blocked on rust-lang/rust-analyzer#22198.

Best reviewed commit by commit, I can (and maybe should) put the last commit (which is the main work) in a separate PR.

### DIFF
--- a/crates/hir-ty/src/tests/simple.rs
+++ b/crates/hir-ty/src/tests/simple.rs
@@ -2697,7 +2697,6 @@ fn generic_default_in_struct_literal() {
 
 #[test]
 fn generic_default_depending_on_other_type_arg() {
-    // FIXME: the {unknown} is a bug
     check_infer(
         r#"
         struct Thing<T = u128, F = fn() -> T> { t: T }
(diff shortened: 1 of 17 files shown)


crates/hir-ty/src/tests/simple.rs
ShoyuVanilla:
```suggestion
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22237/e26f9eee3c9f3c579377e148f8ee41b1916ccb3f..b46476c407f8a2b3cd044c1f8a8b93dd675ac926)*

ShoyuVanilla (approved):

PR #22240, opened 2026-05-01
fix: improve whitespaces for trait item complete

A4-Tacks:
Example
---
```rust
trait SomeTrait<T> {}

trait Foo<T> {
    fn function()
    where Self: SomeTrait<T>;
}
impl Foo<u32> for Bar {
    fn f$0
}
```

**Before this PR**

```rust
impl Foo<u32> for Bar {
    fn function()
        where Self: SomeTrait<u32> {
        $0
    }
}
```

**After this PR**

```rust
impl Foo<u32> for Bar {
    fn function()
    where Self: SomeTrait<u32>
    {
        $0
    }
}
```

### DIFF
--- a/crates/ide-completion/src/completions/item_list/trait_impl.rs
+++ b/crates/ide-completion/src/completions/item_list/trait_impl.rs
@@ -40,7 +40,9 @@ use ide_db::{
 use syntax::ast::HasGenericParams;
 use syntax::{
     AstNode, SmolStr, SyntaxElement, SyntaxKind, T, TextRange, ToSmolStr,
-    ast::{self, HasGenericArgs, HasTypeBounds, edit_in_place::AttrsOwnerEdit, make},
+    ast::{
+        self, HasGenericArgs, HasTypeBounds, edit::AstNodeEdit, edit_in_place::AttrsOwnerEdit, make,
+    },
     format_smolstr, ted,
 };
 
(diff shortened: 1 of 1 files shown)


crates/ide-completion/src/completions/item_list/trait_impl.rs
```diff
@@ -1269,13 +1272,14 @@ trait SomeTrait<T> {}
...
     fn function()
-        where Self: SomeTrait<T>;
+    where Self: SomeTrait<T>;
 }
 struct Bar;
 
 impl Foo<u32> for Bar {
     fn function()
-        where Self: SomeTrait<u32> {
+where Self: SomeTrait<u32>
```
ChayimFriedman2, line 1281: Do you think we should implement in tests the editor functionality that inserts indentation?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22240/64cdaeb06f69b6b769a492edd88b022ae88e8ca2..1e07f4e122456053be5b9edce4305d22c181d871)*

ChayimFriedman2 (approved):

crates/ide-completion/src/completions/item_list/trait_impl.rs
A4-Tacks: This may require adding the concept of 'main edit' for test

I don't think it's necessary

Issue #13047, opened 2022-08-17
"Add a reference" assist adds an extra ampersand when fixing to a mutable ref


yedayak:
**rust-analyzer version**: (eg. output of "rust-analyzer: Show RA Version" command, accessible in VSCode via <kbd>Ctrl/⌘</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd>)
0.3.1162-standalone (634cfe3d7 2022-08-07)

**rustc version**: (eg. output of `rustc -V`)
rustc 1.63.0 (4b91a6ea7 2022-08-08)
**relevant settings**: (eg. client settings, or environment variables like `CARGO`, `RUSTUP_HOME` or `CARGO_HOME`)
None

When passing a reference to a function that is mutable, RA gives an assist `Add reference here`, which outputs incorrect code:

```rust
fn main() {
    let mut a = 0;
    foo(<|>&a);
}
fn foo(a: &mut i32) -> () {}
```
-> 
```rust
fn main() {
    let mut a = 0;
    foo(&mut &a);
}

fn foo(a: &mut i32) -> () {}
```

I looked at the [relevant code](https://github.com/rust-lang/rust-analyzer/blob/82ff740501e1da239a6c9ff94dddf7ae1ca9aeb6/crates/ide-diagnostics/src/handlers/type_mismatch.rs#L57-L80), and it seems this can be fixed by checking if d.actual is a reference before suggesting the fix. We might also want to add an assist for fixing just the missing mut problem?

PR #22248, opened 2026-05-02
refactor: migrate `add_reference_here` to `SyntaxEditor`

ada4a:


crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -89,14 +90,18 @@ fn add_reference(
...
-    let edit = TextEdit::insert(range.range.start(), ampersands);
-    let source_change = SourceChange::from_text_edit(range.file_id, edit);
+    let mut builder =
+        SourceChangeBuilder::new(expr_ptr.file_id.original_file(ctx.db()).file_id(ctx.db()));
+    let expr = expr_ptr.to_node(ctx.db());
+    // TODO: should this be `with_mappings` instead?
+    let syntax_factory = SyntaxFactory::without_mappings();
+    let expr_with_ref = syntax_factory.expr_ref(expr.clone_for_update(), mutability.is_mut());
+    builder.replace_ast(expr, expr_with_ref);
+    let source_change = builder.finish();
```
bit-aloo: lets use `syntax_editor` for this

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22248/fb2288fb5d808686fc8c9452181b7c2cb4cc8d54..4ade6a5cf103f16a1243a11cf5291b9729a87381)*

ada4a: Uhh, I tried, see last commit[^1]. Really not sure if it's right though. And the test still doesn't pass :sweat_smile: 

[^1]: I think I'll squash it all down once I figure it out, but for now it should make review easier I hope..

ChayimFriedman2 (commented):
This assist does not use mutable tree editing, why do you want to migrate it?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22248/fb2288fb5d808686fc8c9452181b7c2cb4cc8d54..2af2d4cc96a438726aa18612bd280a0362255e20)*

ada4a:
> This assist does not use mutable tree editing, why do you want to migrate it?

Because to fix #13047, I'd need to add a new assist, which [I think would not work as a text edit](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/.2313047.3A.20.60.26mut.20.26T.60.20is.20coerced.20to.20.60.26mut.20T.60/near/590173154), so I thought it'd make sense to migrate this one as well. Besides, I thought we're trying to migrate away from text edits anyway? Do correct me if I'm wrong

ChayimFriedman2:
Migrating existing assists that do position-based editing is something worth considering if it makes them clearer, but I resist doing that now. Migration of assists that use mutable editing is more important, and the editor API is not set in stone and will likely require further changes. If we migrate them now we're just adding overhead for when the API will change.

PR #22258, opened 2026-05-03
internal: Do not drop `ExpansionSpanMap` and `AstIdMap` asynchronously

ChayimFriedman2:
They do not contain things that require looping when dropping, only `Vec`/`Arena`/`HashTable` of non-`Drop` types. Therefore sending to an external channel is likely to be slower than just dropping.

r? Veykril

### DIFF
--- a/crates/span/src/ast_id.rs
+++ b/crates/span/src/ast_id.rs
@@ -810,48 +810,6 @@ impl AstIdMap {
     }
 }
 
-#[cfg(not(no_salsa_async_drops))]
-impl Drop for AstIdMap {
-    fn drop(&mut self) {
-        let arena = std::mem::take(&mut self.arena);
-        let ptr_map = std::mem::take(&mut self.ptr_map);
-        let id_map = std::mem::take(&mut self.id_map);
-        static AST_ID_MAP_DROP_THREAD: std::sync::OnceLock<
-            std::sync::mpsc::Sender<(
-                Arena<(SyntaxNodePtr, ErasedFileAstId)>,
-                hashbrown::HashTable<ArenaId>,
-                hashbrown::HashTable<ArenaId>,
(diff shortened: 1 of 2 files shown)


Veykril:
What do you mean with looping? The idea behind this was that salsa cancellation was blocking for a fair bit at times due to the amount of data dropped here, as these are used as LRU'd query results

ChayimFriedman2:
What I mean is, e.g. `Vec<Symbol>` needs to loop over the elements to drop each. But the drop for these types is just one deallocation of `SpanMap` or 3 deallocations for `AstIdMap`. I'm not even sure sending to a channel will be faster.

Veykril:
That is not the point I am getting at, we are not merely dropping one instance of these. We are dropping a bunch, especially after cache priming because the queries of these are LRU'd, so bigger invalidations tend to cause a lot of evictions here. This is in fact faster given I implemented this only after I noticed these things in profiling

ChayimFriedman2:
My point is: I haven't benchmarked, but I suspect dropping one instance is cheaper than sending it to a channel, or at least around the same cost. So we don't need to send it to a background thread.

lnicola:
Sending them over a channel is a pair of copies, an atomic increment and waking the receiving thread, but deallocating memory can still take a bunch of time. I'd expect the channel to be faster, but I also didn't benchmark.

ChayimFriedman2:
Apparently you're right. Criterion:
```rust
channel                 time:   [17.709 ns 17.885 ns 18.070 ns]

drop                    time:   [31.758 ns 32.915 ns 34.112 ns]
```

lnicola:
Oh, these aren't even bounded channels. I think I actually tried to figure out some reasonable sizes, but some of them had way too many items. So we still allocate for every item.

PR #22259, opened 2026-05-03
feat: complete `::` on module def

A4-Tacks:
Example
---
```rust
fn main() {
    c$0
}
```

**Before this PR**

```text
md core
...
```

**After this PR**

```text
md core::
...
```

### DIFF
--- a/crates/ide-completion/src/completions/dot.rs
+++ b/crates/ide-completion/src/completions/dot.rs
@@ -1093,7 +1093,7 @@ impl Foo { fn foo(&mut self) { let _: fn(&mut Self) = |this| { $0 } } }"#,
                 me this.foo() fn(&mut self)
                 lc self            &mut Foo
                 lc this            &mut Foo
-                md core
+                md core::
                 sp Self                 Foo
                 st Foo                  Foo
                 tt Fn
@@ -1117,7 +1117,7 @@ impl Foo { fn foo(&self) { let _: fn(&Self) = |foo| { $0 } } }"#,
                 me self.foo() fn(&self)
                 lc foo             &Foo
                 lc self            &Foo
-                md core
+                md core::
                 sp Self             Foo
                 st Foo              Foo
                 tt Fn
@@ -1137,7 +1137,7 @@ impl Foo { fn foo(&self) { let _: fn(&Self) = || { $0 } } }"#,
                 fd self.field       i32
                 me self.foo() fn(&self)
                 lc self            &Foo
-                md core
+                md core::
                 sp Self             Foo
                 st Foo              Foo
                 tt Fn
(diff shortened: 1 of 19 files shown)


ChayimFriedman2 (commented):
Seems worth gating it behind a setting.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22259/9e457da896bb870275abe7b599ba92072548fc00..409059c746ab5ae671035e838ed6f03f329e647b)*

A4-Tacks:
Do we need to add a config?

ChayimFriedman2:
Yes, I think. It can be on by default.

ChayimFriedman2 (approved):

ThriceCola:
This merge caused me to lose a habit I've always had. Originally, after typing a module name, I could then type `::` and rust-analyzer would show me its submodules. But now, I have to press backspace and then type `:` for it to show the submodules.

ChayimFriedman2:
@ThriceCola There's a config to go back to the old behavior, but I agree we should show autocompletion for the children immediately, if we can.

ThriceCola:
Oh, one more thing — I've written quite a few macros that take a module name as an argument. If it's set to true, it means I have to backspace twice when using this macro. There is a config option for it, but I think having it default to true is a bit unfortunate :(

A4-Tacks:
> But now, I have to press backspace and then type `:` for it to show the submodules.

Another option is to trigger completion in place, such as tab or ctrl+space etc

But `":"` was indeed in the trigger character and did not continue to trigger after completion. Is this a client difference or LSP standard behavior?

> I've written quite a few macros that take a module name.

Oh, maybe you can change `$i:ident` to `$i:ident $(::)?`

I realized that this seems to be fixable

ThriceCola:
> But ":" was indeed in the trigger character and did not continue to trigger after completion. Is this a client difference or LSP standard behavior?

I use zed and kate. After typed 'mod' and Tab, r-a would give me 'mod::', but it didn't show submodules

> such as tab or ctrl+space etc

As a non English environment, our input method usually has three switching options: type `shift` `ctrl` `ctrl+space` I chose 'ctrl+space', so this shortcut key was blocked by the input method. I know most people have 'ctrl+space', but I can't choose it I can only choose to set this feature to false

-----

In short, it lost the ability to type a module name and then press Tab to complete a simple word. so it default to true is a bit unfortunate

A4-Tacks:
> In short, it lost the ability to type a module name and then press Tab to complete a simple word. so it default to true is a bit unfortunate

No, these are two questions. You can take a look at the PR I just opened, to solve you macro case

A4-Tacks:
About trigger submodules, if you use tab to trigger completion, it is very fast

example

If there are button conflicts, the editor should have the ability to customize

PR #22261, opened 2026-05-03
Migrate missing fields to syntax editor

bit-aloo:


crates/ide-diagnostics/src/handlers/missing_fields.rs
```diff
@@ -159,35 +164,69 @@ fn fixes(ctx: &DiagnosticsContext<'_, '_>, d: &hir::MissingFields) -> Option<Vec
...
+    field_list: &ast::RecordExprFieldList,
+    fields: Vec<ast::RecordExprField>,
+) -> Option<SyntaxNode> {
+    let old_range = field_list.syntax().text_range();
+    let kind = field_list.syntax().kind();
+    field_list.add_fields(&editor, fields);
+    let edit = editor.finish();
+    edit.new_root()
+        .descendants()
+        .find(|it| it.kind() == kind && it.text_range().start() == old_range.start())
```
ChayimFriedman2: Can't we have a for the range since the editor roots things?

Also it's worth considering whether we should generalize the process of finding the mapped node to a method in the editor.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22261/9e457da896bb870275abe7b599ba92072548fc00..82c640505cecc965b9bf295e23a40d88bde97dc2)*

bit-aloo: I didn't get the first suggestion.

We can opt for range based finding on new root.

ChayimFriedman2: What I was saying is that since the editor `clone_subtree()`s its node, the ranges won't be equal.

bit-aloo: ~~We don't really need this, considering we are inserting only the new fields, it should be present in changed_elements in final edit, which we can directly use that. Added here: 8a83ee6.~~ Tests are failing with this

bit-aloo: Added `find_element`  in syntax_edit, which respects the relative positioning of the element.

crates/syntax/src/syntax_editor/edits.rs
```diff
@@ -241,6 +237,113 @@ impl ast::AssocItemList {
...
+    let next_after_insert;
+    let position = match last_field {
+        Some(last_field) => match comma_after(&last_field) {
+            Some(comma) => {
+                next_after_insert = comma.next_sibling_or_token();
+                Position::after(comma)
+            }
+            None => {
+                next_after_insert = last_field.next_sibling_or_token();
+                elements.push(make.token(T![,]).into());
```
ChayimFriedman2, line 309: Shouldn't this be `insert(0, ...)`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22261/13bbf74923453c8b3f76273ff1a8610007936dec..f80347b32581566ffe17896e383d276966b0e19d)*

bit-aloo: The token will always on first index if ever invoked. I don't see the difference.

ChayimFriedman2: Right, my mistake.

ChayimFriedman2 (approved):

PR #22269, opened 2026-05-04
Fix nested edits in SyntaxEditor

bit-aloo:

### DIFF
--- a/crates/syntax/src/syntax_editor/edit_algo.rs
+++ b/crates/syntax/src/syntax_editor/edit_algo.rs
@@ -111,11 +111,8 @@ pub(super) fn apply_edits(editor: SyntaxEditor) -> SyntaxEdit {
         // Check if this change is dependent on another change (i.e. it's contained within another range)
         if let Some(index) = changed_ancestors
             .iter()
-            .rev()
-            .position(|ancestor| ancestor.affected_range().contains_range(change.target_range()))
+            .rposition(|ancestor| ancestor.affected_range().contains_range(change.target_range()))
         {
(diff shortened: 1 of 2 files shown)


crates/syntax/src/syntax_editor/edit_algo.rs
```diff
@@ -114,6 +114,8 @@ pub(super) fn apply_edits(editor: SyntaxEditor) -> SyntaxEdit {
             .rev()
             .position(|ancestor| ancestor.affected_range().contains_range(change.target_range()))
         {
+            let index = changed_ancestors.len() - 1 - index;
```
ChayimFriedman2:
```suggestion
            .enumerate()
            .rev()
            .find(|ancestor| ancestor.affected_range().contains_range(change.target_range()))
        {
```
Clearer this way.

ChayimFriedman2 (approved):
Awesome! Just one nit.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22269/13bbf74923453c8b3f76273ff1a8610007936dec..275c5c4e8df37c74e3913210802413657ff3c0fd)*

crates/syntax/src/syntax_editor/edit_algo.rs
bit-aloo: Used `rposition`, since we can’t use the `ancestor` from `enumerate` directly, we need a mutable reference to `changed_ancestors`.

ChayimFriedman2 (approved):

PR #22283, opened 2026-05-04
feat: early late classification of lifetimes

dfireBird:
Some things to note:
1. We still don't handle the HRTB Lifetimes (in both where and fn ptr) and will be done in a future PR and that PR should also remove the `in_binders` and fully replace it with `bound_vars` stack.
2. I haven't added a new tests, not sure how to test this particularly, so any approach to test them are helpful (this should be done before merging)
3. There are some IDE features which lose information that I consider to be essential, need to find a way to not have that loss of information (see `FIXME`  in ide tests)
4. I have implemented this to best of my understanding of both r-a and rustc, so if there are panics or data mismatch, I'm really sorry that happened, I'll try to fix it as soon as I get to know about it.


crates/hir-def/src/hir/generics.rs
```diff
@@ -33,6 +33,20 @@ pub struct TypeParamData {
 #[derive(Clone, PartialEq, Eq, Debug, Hash)]
 pub struct LifetimeParamData {
     pub name: Name,
+    pub bound_type: LifetimeBoundType,
+    pub is_opaque_captured: bool,
```
ChayimFriedman2: This is unused currently, and also useless: you could have many opaques.

crates/hir-def/src/hir/generics.rs
```diff
@@ -33,6 +33,20 @@ pub struct TypeParamData {
 #[derive(Clone, PartialEq, Eq, Debug, Hash)]
 pub struct LifetimeParamData {
     pub name: Name,
+    pub bound_type: LifetimeBoundType,
```
ChayimFriedman2, line 36: I wonder if we should have different arenas for the early and late bound lifetimes, so locating a lifetime param's index can be O(1).

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -545,6 +553,50 @@ impl BindingList {
...
+    lifetimes_in_output: FxIndexSet<Name>,
+    opaque_captured_lifetimes: FxIndexSet<Name>,
+}
+
+#[derive(Debug)]
+enum LifetimeBoundScope {
+    Argument,
+    Return,
+    WhereClause,
+    ImplTrait(bool),
```
```suggestion
    ImplTrait { is_argument_scope: bool },
```
Unnamed `bool` field doesn't tell a lot.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -545,6 +553,50 @@ impl BindingList {
     }
 }
 
+#[derive(Debug, Default)]
+pub struct NamedLifetimeStore {
+    lifetime_bound_scope: Option<LifetimeBoundScope>,
+    lifetimes_in_where_clause: FxIndexSet<Name>,
+    lifetimes_constraint_by_input: FxIndexSet<Name>,
```
```suggestion
    lifetimes_constrained_by_input: FxIndexSet<Name>,
```
?

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -774,6 +835,10 @@ impl<'db> ExprCollector<'db> {
         lifetime_ref: LifetimeRef,
         node: LifetimePtr,
     ) -> LifetimeRefId {
+        if let LifetimeRef::Named(name) = &lifetime_ref {
```
ChayimFriedman2, line 842: Unnamed lifetimes can also be late bound, and this code won't account for them. I think we need to switch `push_named_lifetime()` and the maps to store a `LifetimeRefId`.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -3220,6 +3285,98 @@ impl ExprCollector<'_> {
...
+            let alias = loc.source(db);
+
+            let mut expr_collector = ExprCollector::new(db, module, alias.file_id);
+            alias.value.ty().map(|ty| {
+                // Note: This is specific for function lowering, re-using it for type alias because rustc
+                // determines lifetimes by visiting the type itself
+                expr_collector.with_lifetime_bound_scope(
+                    LifetimeBoundScope::Argument,
+                    |expr_collector| {
+                        expr_collector.lower_type_ref(ty, &mut ExprCollector::impl_trait_allocator)
```
ChayimFriedman2: Impl traits should be allowed here.

In general it seems risky to relower things here. Instead we should visit the lowered type and determine what `LifetimeRefId`s are late bound.

crates/hir-def/src/expr_store/lower/path.rs
```diff
@@ -54,6 +54,17 @@ pub(super) fn lower_path(
         ast_segments.push(_segment.clone());
         segments.push(name);
     };
+
+    let type_alias_constrained_lifetimes = collector.get_constrained_lifetimes_if_type_alias(&path);
+    let mut old_lifetime_bound_scope = None;
+    if collector.is_argument_lt_bound_scope()
+        && let Some(lifetimes) = type_alias_constrained_lifetimes
```
ChayimFriedman2: The lifetime names (or IDs if it will be changed to that) here are of the type alias. We cannot reuse them here. Instead we need to find the mapping to the user's lifetimes.

Also, this is called in the body too and there's no reason to put the expensive `get_constrained_lifetimes_if_type_alias()` check there. Please move it inside the `if`.

Also, why re-lower the path in `get_constrained_lifetimes_if_type_alias()` if we'll already have a `ModPath` by the end of this method.

crates/hir-def/src/expr_store/lower/path.rs
```diff
@@ -54,6 +54,17 @@ pub(super) fn lower_path(
...
+    {
+        collector.extend_type_alias_lifetime(lifetimes.into_iter());
+        // we are disabling lifetime bound collector for the entire path
```
ChayimFriedman2: Why, actually?

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -3220,6 +3285,98 @@ impl ExprCollector<'_> {
...
+        bound_scope: LifetimeBoundScope,
+        f: impl FnOnce(&mut Self) -> T,
+    ) -> T {
+        let old = self.named_lifetime_store.lifetime_bound_scope.replace(bound_scope);
+        let res = f(self);
+        self.named_lifetime_store.lifetime_bound_scope = old;
+        res
+    }
+
+    fn for_path_type_projection<T>(&mut self, f: impl FnOnce(&mut Self) -> T) -> T {
```
ChayimFriedman2, line 3436: What is the role of this method? Why are projections treated specifically?

crates/hir-ty/src/next_solver/generic_arg.rs
```diff
@@ -547,7 +555,11 @@ impl<'db> GenericArgs<'db> {
         mk_kind: F,
     ) -> GenericArgs<'db>
     where
-        F: FnMut(u32, GenericParamId, &[GenericArg<'db>]) -> GenericArg<'db>,
+        F: FnMut(
+            u32,
+            (GenericParamId, GenericParamDataRef<'db>),
```
ChayimFriedman2: I see no reason to pass a tuple here. You should flatten it.

crates/hir-ty/src/display.rs
```diff
@@ -2316,8 +2316,8 @@ impl<'db> HirDisplay<'db> for Region<'db> {
                 }
             }
             RegionKind::ReErased => write!(f, "'<erased>"),
-            RegionKind::RePlaceholder(_) => write!(f, "<placeholder>"),
-            RegionKind::ReLateParam(_) => write!(f, "<late-param>"),
+            RegionKind::RePlaceholder(_) => write!(f, "'<placeholder>"),
+            RegionKind::ReLateParam(_) => write!(f, "'<late-param>"),
```
ChayimFriedman2: Given this can be shown to the user, I don't think showing `<late-param>` is appropriate. You need to look up the lifetime name instead.

crates/hir-ty/src/dyn_compatibility.rs
```diff
@@ -417,9 +417,8 @@ fn receiver_is_dispatchable<'db>(
         return true;
     }
 
-    let Some(&receiver_ty) = sig.inputs().skip_binder().first() else {
-        return false;
-    };
+    // FIXME: sig.input(0) has unwrap and might panic
```
ChayimFriedman2: This function is private and only called when the function has a self param.

crates/hir-ty/src/lower.rs
```diff
@@ -230,10 +234,12 @@ impl<'db, 'a> TyLoweringContext<'db, 'a> {
         generic_def: GenericDefId,
         generics: &'a OnceCell<Generics<'db>>,
         lifetime_elision: LifetimeElisionKind<'db>,
+        lifetime_lowering_mode: LifetimeLoweringMode,
     ) -> Self {
         let impl_trait_mode = ImplTraitLoweringState::new(ImplTraitLoweringMode::Disallowed);
         let in_binders = DebruijnIndex::ZERO;
         let interner = DbInterner::new_with(db, resolver.krate());
+        let bound_vars = Vec::from(&[Self::bound_vars(db, interner, generic_def)]);
```
ChayimFriedman2:
```suggestion
        let bound_vars = vec![Self::bound_vars(db, interner, generic_def)];
```

crates/hir-ty/src/lower.rs
```diff
@@ -575,16 +674,21 @@ impl<'db, 'a> TyLoweringContext<'db, 'a> {
             args.push(ctx.lower_ty(ret_ty));
         });
         self.lifetime_elision = old_lifetime_elision;
+
+        let binder = BoundVarKinds::new_from_slice(self.bound_vars.last().unwrap().as_slice());
```
ChayimFriedman2:
```suggestion
        let binder = *self.bound_vars.last().unwrap();
```

crates/hir-ty/src/lower.rs
```diff
@@ -728,6 +832,8 @@ impl<'db, 'a> TyLoweringContext<'db, 'a> {
         let mut clause = None;
         match bound {
             &TypeBound::Path(path, TraitBoundModifier::None) | &TypeBound::ForLifetime(_, path) => {
+                let binder =
+                    BoundVarKinds::new_from_slice(self.bound_vars.last().unwrap().as_slice());
```
ChayimFriedman2:
```suggestion
                    *self.bound_vars.last().unwrap();
```

ChayimFriedman2: Seems worth having a method for this given the `unwrap()`.

crates/hir-ty/src/lower.rs
ChayimFriedman2: Also, now that we handle HRTB, you need to handle `TypeBound::ForLifetime` and push it to this `Vec`.

crates/hir-ty/src/lower.rs
```diff
@@ -575,16 +674,21 @@ impl<'db, 'a> TyLoweringContext<'db, 'a> {
...
+            Binder::bind_with_vars(
+                FnSig {
+                    fn_sig_kind: FnSigKind::new(
+                        fn_.abi,
+                        if fn_.is_unsafe { Safety::Unsafe } else { Safety::Safe },
+                        fn_.is_varargs,
+                    ),
+                    inputs_and_output: Tys::new_from_slice(&args),
+                },
+                binder,
```
ChayimFriedman2, line 764: This is wrong. The binder should include the HRTB lifetimes generated *inside* the fn pointer, not outside it. E.g. the `'_` in `fn(&'_ ())`, or the `'foo` in `for<'foo> fn()`.

ChayimFriedman2: Also for the same reason you don't need to push the top-level bound vars into it.

Generally, when you start to lower HRTB constructs (fn pointers, trait bounds, dyn traits) you should put an empty `Vec` into the stack, and each time we generate a lifetime we'll put it there. Then when finishing lowering the construct, you should pop the stack and create a `BoundVarKinds` from it.

crates/hir-ty/src/next_solver/generics.rs
```diff
@@ -33,40 +35,54 @@ pub(crate) fn generics(interner: DbInterner<'_>, def: SolverDefId) -> Generics<'
...
 
     pub(crate) fn from_generic_def_plus_one(
         db: &'db dyn HirDatabase,
         def: GenericDefId,
-        additional_param: GenericParamId,
+        additional_param: TypeParamId,
+        consider_late_bound: bool,
     ) -> Generics<'db> {
+        let generics = crate::generics::generics(db, def);
+        let data_ref = generics.type_or_const_param(additional_param);
```
ChayimFriedman2: This is entirely wrong. The additional param won't be found in this generics, because it is fake. You need to the `GenericParamDataRef` to this method as well.

crates/hir-ty/src/generics.rs
```diff
@@ -275,12 +331,28 @@ impl<'db> Generics<'db> {
...
+        }
+
+        if let Some(late_bound_idx) = owner.params.late_bound_lifetime_idx(&param.local_id) {
+            return (late_bound_idx as u32, true);
+        }
         let has_trait_self = matches!(owner.def, GenericDefId::TraitId(_));
-        owner.preceding_params_len
-            + u32::from(has_trait_self)
-            + param.local_id.into_raw().into_u32()
+        match owner.params.early_bound_lifetime_idx(&param.local_id) {
```
ChayimFriedman2: Instead of looping twice, just have a single method `lifetime_idx()` on `GenericParams`.

ChayimFriedman2 (dismissed):
Awesome work!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22283/1a68212c5683555ad80f0eab71db9715c6d52145..b2f078f886d3e5b9402c099313395b740229e352)*

crates/hir-def/src/expr_store/lower.rs
dfireBird: By unnamed lifetimes, do you mean `'_` and elided lifetimes, if so then they can never be early bound (from what I have seen).

I considered `LifetimeRefId` as well, but it doesn't work well due to nature we create a new one every time we refer a lifetime and we need to index into `ExprCollector` to determine if they our `LifetimeRefId` refers to same lifetime or different.

crates/hir-def/src/expr_store/lower.rs
dfireBird: Do you mean shouldn't?

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: > The lifetime names (or IDs if it will be changed to that) here are of the type alias. We cannot reuse them here. Instead we need to find the mapping to the user's lifetimes

What do you mean by this?

> Also, this is called in the body too and there's no reason to put the expensive get_constrained_lifetimes_if_type_alias() check there. Please move it inside the if.

We need to know, if the path resolves to an type alias at the very least, so that we can disable the named lifetime store recording.

> Also, why re-lower the path in get_constrained_lifetimes_if_type_alias() if we'll already have a ModPath by the end of this method.

For the reason stated above, I need to keep this before we start any lowering.

crates/hir-def/src/expr_store/lower.rs
ChayimFriedman2: Ah right, it is allowed. Still, it's risky to relower things here.

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: Why we are disabling? If so, once we know what lifetimes are constrained in the type alias, we don't have a need to collect the lifetimes if any in that path. If we don't disable then lifetime in that path could be added as constrained by input, and that's an issue if the lifetime is not constrained in the type alias.

crates/hir-def/src/expr_store/lower.rs
dfireBird: Lifetimes in projection are not constrained by input, as such we disable the lifetime collector in projections.

crates/hir-ty/src/next_solver/generic_arg.rs
dfireBird: Do you mean, adding it as a parameter, instead of in tuple?

crates/hir-def/src/expr_store/lower/path.rs
ChayimFriedman2: > What do you mean by this?

E.g. in:
```rust
type Alias<'a> = ...;

fn foo<'b>(_: Alias<'b>) {}
```
The name will be `'a`, but it should be `'b`.

> We need to know, if the path resolves to an type alias at the very least, so that we can disable the named lifetime store recording.

That is solvable by taking the current constrained lifetime, and at the end append them to the new lifetimes or override the new lifetimes.

crates/hir-ty/src/display.rs
dfireBird: As discussed in Zulip before, `ReLateParam` currently don't have necessary information to find the name of lifetime.

crates/hir-ty/src/dyn_compatibility.rs
dfireBird: So it shouldn't cause an panic and we don't need that FIXME?

crates/hir-ty/src/lower.rs
dfireBird: Yes, I'm thinking of doing that in separate PR, since this PR is already big enough as is.

crates/hir-ty/src/next_solver/generic_arg.rs
ChayimFriedman2: Yes.

crates/hir-ty/src/dyn_compatibility.rs
ChayimFriedman2: Yes.

crates/hir-ty/src/lower.rs
dfireBird: You're right, it's currently wrong (I added this knowing it) but once we start handling the HRTB lifetimes, this will work out correctly.

> Also for the same reason you don't need to push the top-level bound vars into it.

We need the top-level bound vars since we are going to be using it here: https://github.com/rust-lang/rust-analyzer/pull/22283/changes#diff-c6bb274739383bc0ccc767673d4bfcedd83ba7b156b81cecf80676989ad4c010R2734-R2745

crates/hir-ty/src/next_solver/generics.rs
dfireBird: I was actually not sure what to do specifically for this case, I thought this would be correct. Should I instead pass `Option<GenericParamDataRef>` here?

crates/hir-def/src/expr_store/lower.rs
dfireBird: I'll rename it :sweat_smile:

crates/hir-ty/src/lower.rs
ChayimFriedman2: > You're right, it's currently wrong (I added this knowing it) but once we start handling the HRTB lifetimes, this will work out correctly.

No, it'll remain wrong unless you change the `BoundVarKinds`. If you intend to do that, just push empty kinds for now.

> We need the top-level bound vars since we are going to be using it here

GitHub doesn't let me where you point at. If you mean in the callable signature query, that doesn't justify putting it in `TyLoweringContext` for everything.

crates/hir-ty/src/next_solver/generics.rs
ChayimFriedman2: It doesn't seem worth it to change every caller to `Option` just for this, especially given this won't be a lifetime param anyway. So either change it to take a `Option<&LifetimeParamData>` or provide the data here too.

crates/hir-ty/src/display.rs
ChayimFriedman2: That's kinda sad, I'll see what we can do here. Still showing `'<late-param>` which isn't a valid Rust syntax is weird, I'd show `'_` instead, at least when formatting for source code.

crates/hir-def/src/expr_store/lower.rs
dfireBird: We couldn't do what rustc is doing, which is visit the type in the `hir-ty` (see [here](https://github.com/rust-lang/rust/blob/main/compiler/rustc_hir_analysis/src/collect/resolve_bound_vars.rs#L2469-L2471)), but they are entirely driven by queries as such it won't be lowering won't be done twice.

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: You're right, I missed it. I actually need to think how to solve this. We could maybe return the index of the constrained lifetimes and we see what lifetimes the user has used in that index?

> That is solvable by taking the current constrained lifetime, and at the end append them to the new lifetimes or override the new lifetimes.

Do you mean using a temporary store only for path lowering, then decided based on `ModPath` below? Then I can work on that.

crates/hir-ty/src/next_solver/generic_arg.rs
dfireBird: Got it. Thanks I'll work on this along with other changes requested.

crates/hir-ty/src/display.rs
dfireBird: > I'd show '_ instead, at least when formatting for source code.

I'll change it for now then.

> I'll see what we can do here.

I do have a idea for this tho, add an index or name field inside the struct that enum variants contains, we own the struct definition after all not rustc: https://github.com/rust-lang/rust-analyzer/blob/master/crates/hir-ty/src/next_solver/region.rs#L166-L169

crates/hir-def/src/expr_store/lower.rs
ChayimFriedman2: We can visit the `TypeRef`.

It is lowering twice, once in the signature query and once here, and while that's something we sometimes do it's generally not a good idea. For example, we might want to use tracked structs instead of interneds, and that will stand in our way.

crates/hir-def/src/expr_store/lower/path.rs
ChayimFriedman2: > We could maybe return the index of the constrained lifetimes and we see what lifetimes the user has used in that index?

Can work indeed.

> Do you mean using a temporary store only for path lowering, then decided based on ModPath below? Then I can work on that.

A whole store is unneeded, we need just the constrainted_by_input map.

crates/hir-ty/src/display.rs
ChayimFriedman2: We can't put a name because it needs to be `Copy`, an index will work but please do so in a separate PR.

crates/hir-ty/src/lower.rs
dfireBird: > No, it'll remain wrong unless you change the BoundVarKinds. If you intend to do that, just push empty kinds for now.

Sure, I'll add it, but I'm not sure if we panic. If we indeed panic in some place, then I'll have to add that in this PR itself.

> GitHub doesn't let me where you point at. If you mean in the callable signature query, that doesn't justify putting it in TyLoweringContext for everything.

Yes there's where I meant but I get why you mean, I'll remove it then and use the method that I use for populating it instead.

crates/hir-ty/src/next_solver/generics.rs
dfireBird: > So either change it to take a Option<&LifetimeParamData>

You mean the type of second element of tuple (which I'm going to change it as parameter, after your comments)?

crates/hir-def/src/expr_store/lower.rs
dfireBird: > We can visit the TypeRef.

Isn't that what I'm essentially doing here?

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: Sure, I'll work on both.

Just to clarify on the latter, you mean to say, take out the input map value before lowering the path right and replace it at the end, right?

crates/hir-ty/src/display.rs
dfireBird: Sure, got it. There seems to be a lot of follow up PRs :smile:

crates/hir-def/src/expr_store/lower.rs
ChayimFriedman2: No, we visit the AST.

crates/hir-def/src/expr_store/lower/path.rs
ChayimFriedman2: Replace it or extend it, depending on whether we want to keep the constraints that arose from this path.

crates/hir-def/src/expr_store/lower.rs
dfireBird: Oh so visiting the `TypeRef` is kinda like hir-ty lowering?

ChayimFriedman2: Kinda, yes.

dfireBird: There was a [discussion in zulip](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Lifetime.20elision/near/595844645) regarding how to properly visit the generic arg in the type ref of the type alias.

TLDR is we are moving the `on_generic_arg` from `StoreVisitorExt` to `StoreVisitor`

crates/hir-def/src/hir/generics.rs
dfireBird: Coming to this (I kept this to the last), we can't do that with what we have. Since we don't know if lifetime is late bound or early bound until after we finish our `hir-def` lowering. We can not mutate the arena itself that will make the `Idx` of that arena to be invalidated.

ChayimFriedman2: True, but we don't need the Idx during lowering. We can just make the arenas after we finish lowering.

dfireBird: If that's the case, I'll try doing something.

dfireBird: It was fairly simple to get it to compile without errors, the only thing I find problematic is implementation of `std::ops::Idx<LocalLifetimeParamId>`, which lifetime arena to choose? We are currently (with my changes ofc) indexing with both early and late lifetimes.

ChayimFriedman2: They need to have a different ID, otherwise many things will be problematic.

dfireBird: I feel like that's unnecessary complexity for an optimization that would only benefit two methods, the new `lifetime_param_idx` and `len_late_bound_lifetimes`. Both for which can be handled through pre-computing and storing those value.

ChayimFriedman2: I didn't say we should do this, I was only wondering whether it's better. But do note that these methods may be fairly hot.

dfireBird: Yeah I get it. 

I was just voicing out my thoughts 😅 .

crates/ide-assists/src/handlers/extract_type_alias.rs
```diff
@@ -358,6 +358,7 @@ mod m {
         );
     }
 
+    // FIXME: This assists lost some information with early and late boud and this should be fixed(?)
```
Veykril: not a useful comment i'd say as its not clear what even was lost, let's just remove this. We will stumble upon this once we flesh out lifetime handling more

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22283/166910828f46696c96e04fede7180bc96015b2c3..cb21b17e36473582b4972402dcf1005a7900751a)*

crates/ide-completion/src/tests/special.rs
```diff
@@ -1573,6 +1573,7 @@ fn check_signatures(src: &str, kind: CompletionItemKind, reduced: Expect, full:
     full.assert_eq(completion[0].detail.as_ref().unwrap());
 }
 
+// FIXME: This assists lost some information with early and late boud and this should be fixed(?)
```
Veykril: same thing here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22283/166910828f46696c96e04fede7180bc96015b2c3..cb21b17e36473582b4972402dcf1005a7900751a)*

crates/hir-def/src/expr_store/lower/generics.rs
```diff
@@ -62,7 +65,9 @@ impl GenericParamsCollector {
             self.lower_param_list(ec, params)
         }
         if let Some(where_clause) = where_clause {
-            self.lower_where_predicates(ec, where_clause);
+            ec.with_lifetime_bound_scope(LifetimeBoundScope::WhereClause, |ec| {
+                self.lower_where_predicates(ec, where_clause)
+            });
```
Veykril: nit, I think `lower_where_predicates` itself should do the `with_lifetime_bound_scope` call.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22283/166910828f46696c96e04fede7180bc96015b2c3..cb21b17e36473582b4972402dcf1005a7900751a)*

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -545,6 +552,48 @@ impl BindingList {
...
+                self.lifetimes_in_where_clause.insert(lifetime);
+            }
+            Some(LifetimeBoundScope::ImplTrait { is_argument_scope }) => {
+                if is_argument_scope {
+                    self.lifetimes_in_where_clause.insert(lifetime);
+                } else {
+                    self.lifetimes_in_output.insert(lifetime);
+                }
+            }
+            _ => (),
```
```suggestion
            None => (),
```
better to exhaustively match here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22283/166910828f46696c96e04fede7180bc96015b2c3..cb21b17e36473582b4972402dcf1005a7900751a)*

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -3347,6 +3409,129 @@ impl ExprCollector<'_> {
...
+                })
+                .collect();
+            Some(res)
+        } else {
+            None
+        };
+        return res;
+
+        #[salsa::tracked(returns(clone), cycle_result = get_constrained_lifetimes_cycle_result)]
+        fn get_constrained_lifetimes(db: &dyn DefDatabase, type_alias_id: TypeAliasId) -> Vec<u32> {
```
```suggestion
        #[salsa::tracked(returns(as_deref), cycle_result = get_constrained_lifetimes_cycle_result)]
        fn get_constrained_lifetimes(db: &dyn DefDatabase, type_alias_id: TypeAliasId) -> Box<[u32]> {
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22283/166910828f46696c96e04fede7180bc96015b2c3..cb21b17e36473582b4972402dcf1005a7900751a)*

dfireBird: I'd like understand the difference. I saw both were possible and used clone for simplicity.

Veykril: Clone means every access to the query will clone out the result to the caller, which is a lot more expensive. as_deref changes the return type to become `&[u32]` effectively which is just a reference copy

crates/ide-assists/src/handlers/extract_type_alias.rs
dfireBird: It was mostly comment to myself :sweat_smile: . I'll remove it and keep a note of TODO on someplace else.

crates/hir-def/src/expr_store/lower.rs
dfireBird: Oh. It makes sense now. I wasn't thinking correctly then. My bad.

Veykril:
Assuming there is nothing else wrong with this PR, let's try getting this merged after release on monday I'd say?

dfireBird:
> Assuming there is nothing else wrong with this PR, let's try getting this merged after release on monday I'd say?

Yeah, we can do that. 
I'm also ready with a new HRTB lowering PR as well :P


I also found a mismatch between of working between rust and what I did while doing the PR and fixed it for this as well. 
I'll push after resolving the conflicts.

crates/hir-ty/src/next_solver/generics.rs
```diff
@@ -33,40 +38,63 @@ pub(crate) fn generics(interner: DbInterner<'_>, def: SolverDefId) -> Generics<'
...
+        self.generics
+            .iter(self.consider_late_bound)
+            .map(|(id, data)| {
+                if let GenericParamDataRef::LifetimeParamData(lt_param) = data {
+                    (id, Some(lt_param))
+                } else {
+                    (id, None)
+                }
+            })
+            .chain(self.additional_param.zip(None))
```
Veykril, line 97: cant annotate the lines below but

```rust
impl<'db> rustc_type_ir::inherent::GenericsOf<DbInterner<'db>> for Generics<'db> {
    fn count(&self) -> usize {
        self.generics.len() + usize::from(self.additional_param.is_some())
    }
}
```
is this still right with respect to `iter` now doing filtering depending on `self.consider_late_bound`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22283/b6790572b3651d20c5885908577c212de19a378b..0241515bd6c454c9e3a5572270aa33685fecdceb)*

Veykril:
also i just opened a pr against your pr with some fixes https://github.com/dfireBird/rust-analyzer/pull/1 (do check over them, stuff might still be wrong there). Feel free to just cherry-pick those or w/e.

crates/hir-ty/src/next_solver/generics.rs
dfireBird: I think you're right.  We need to consider them or not based on that field.

Thanks for noticing.

ChayimFriedman2:
I'd like to review the changes after my review here, but not sure I'll find the time for that. So if I don't review until Monday feel free to merge.

Veykril:
Didn't mean to rush, was mainly worried about the need for rebasing :)

dfireBird:
We can wait, rebasing is mostly conflicts on imports, except a few instances, which is not a worry for me :)

Veykril:
I'd like to land this on monday now, even if Chayim is still a bit busy and can't necessarily review it at the moment. They can always review it after wards again still.

Veykril (approved):

Veykril:
needs a rebase due to semantic conflict

dfireBird:
> needs a rebase due to semantic conflict

Done.

Veykril:
not quite it seems :)

dfireBird:
Oh well, I thought it had no conflict, it would work :)

My bad.

dfireBird:
I think it's done for real this time :D

Note to self for Nth time: Run all cargo commands after rebase.

dfireBird:
I'll create the follow up PR in few days.

PR #22284, opened 2026-05-04
Backwards compatible CompletionImport#as_underscore

xitep:
Commit [f42e02aa](https://github.com/rust-lang/rust-analyzer/commit/f42e02aaff84d7b613ab9057036e93010c23b93f#diff-1f818ad963401fe13a8a29e158447fa51390f3026ecba9de2e38a1cbdcaf9f4cR95) introduced a new, required field which breaks auto-import in emacs in the very same fashion as issue rust-lang/rust-analyzer#18767.

This PR makes the field optional; it does not, however, fix the issue per se for emacs users since `lsp-mode` still sends a `null`.

### DIFF
--- a/crates/rust-analyzer/src/lsp/ext.rs
+++ b/crates/rust-analyzer/src/lsp/ext.rs
@@ -858,6 +858,7 @@ pub struct InlayHintResolveData {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct CompletionImport {
     pub full_import_path: String,
+    #[serde(default)]
     pub as_underscore: bool,
 }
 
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (approved):
I believe lsp-mode should fix their code, but this change is small enough that I don't resist.

Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22284/cd24e6e23d2aafd4d5ac20b6ed7e6d255bf2fa44..c6f661a09d974a757cd7169b2a52281b0566fbce)*

Veykril:
not the first time we had to fix this emacs bug on our side

PR #22285, opened 2026-05-04
fix: ref match uses unified type

A4-Tacks:
Example
---
```rust
struct S<T>(T);
fn foo<T>(s: &mut S<T>) {}
fn main() {
    let mut ssss = S(2u32);
    foo($0);
}
```

**Before this PR**

```text
st S(…) S(T) []
st &mut S(…) [type]
lc ssss S<u32> [local]
```

**After this PR**

```text
st S(…) S(T) []
st &mut S(…) [type]
lc ssss S<u32> [local]
lc &mut ssss [type+local]
```

### DIFF
--- a/crates/ide-completion/src/render.rs
+++ b/crates/ide-completion/src/render.rs
@@ -713,7 +713,9 @@ fn compute_ref_match(
     if let Some(expected_without_ref) = &expected_without_ref
         && (completion_without_ref.is_none()
             || completion_ty.could_unify_with(ctx.db, expected_without_ref))
-        && completion_ty.autoderef(ctx.db).any(|ty| ty == *expected_without_ref)
+        && completion_ty
+            .autoderef(ctx.db)
+            .any(|ty| ty.could_unify_with(ctx.db, expected_without_ref))
     {
         cov_mark::hit!(suggest_ref);
         let mutability = if expected_type.is_mutable_reference() {
@@ -2914,6 +2916,26 @@ fn main() {
                 lc &ssss [type+local]
                 st S S []
                 fn foo(…) fn(&&S) []
+                fn main() fn() []
+            "#]],
+        );
+        check_relevance(
+            r#"
+struct S<T>(T);
+fn foo<T>(s: &mut S<T>) {}
+fn main() {
+    let mut ssss = S(2u32);
+    foo($0);
+}
+            "#,
+            expect![[r#"
+                st S(…) S(T) []
+                st &mut S(…) [type]
+                lc ssss S<u32> [local]
+                lc &mut ssss [type+local]
+                st S S<{unknown}> []
+                st &mut S [type]
+                fn foo(…) fn(&mut S<T>) []
                 fn main() fn() []
             "#]],
         );


ChayimFriedman2 (approved):

A4-Tacks:
I'm a little worried, this seems to have caused too many ref match items (especially auto import) to be completed by `fn <T>(&T)`, maybe we need a function between `could_unify_with_deeply` and `could_unify_with`

ChayimFriedman2:
How could there be such function? The difference between `could_unify_with_deeply()` and `could_unify_with()` is that the former requires all obligations to success, causing a failure if there are unresolved infer vars.

A4-Tacks:
If using `could_unify_with_deeply()`:
- `i32` != `T`
- `&i32` != `&T`
- `Option<i32>` != `T`
- `Option<i32>` != `Option<T>`

If using `could_unify_with()`:
- `i32` == `T`
- `&i32` == `&T`
- `Option<i32>` == `T`
- `Option<i32>` == `Option<T>`

If using i expected function like:
- `i32` != `T`
- `&i32` != `&T`
- `Option<i32>` != `T`
- `Option<i32>` == `Option<T>`

Although this may not seem reasonable, it seems necessary to suppress a large number of ref matches in shallow generic parameters

```rust
fn foo<T>(_: &T) {}
fn main() {
    foo(a$0);
}
```

Completes:

```text
&align_of()~ const fn() -> usize
&align_of_val(…)~ const fn(&T) -> usize
&args()~(use std::env::args) fn() -> Args
&alloc(…)~(use std::alloc::alloc) unsafe fn(Layout) -> *mut u8
&abort()~(use std::intrinsics::abort) fn() -> !
&abort()~(use std::process::abort) fn() -> !
&args_os()~(use std::env::args_os) fn() -> ArgsOs
&addr_eq(…)~(use std::ptr::addr_eq) fn(*const T, *const U) -> bool
&align_of()~(use std::intrinsics::align_of) const fn() -> usize
&autodiff(…)~(use std::intrinsics::autodiff) const fn(F, G, T) -> R
&absolute(…)~(use std::path::absolute) fn(P) -> Result<PathBuf, Error>
&atomic_or(…)~(use std::intrinsics::atomic_or) unsafe fn(*mut T, U) -> T
&atomic_and(…)~(use std::intrinsics::atomic_and) unsafe fn(*mut T, U) -> T
&atomic_max(…)~(use std::intrinsics::atomic_max) unsafe fn(*mut T, T) -> T
&atomic_min(…)~(use std::intrinsics::atomic_min) unsafe fn(*mut T, T) -> T
&atomic_xor(…)~(use std::intrinsics::atomic_xor) unsafe fn(*mut T, U) -> T
&atomic_load(…)~(use std::intrinsics::atomic_load) unsafe fn(*const T) -> T
&atomic_nand(…)~(use std::intrinsics::atomic_nand) unsafe fn(*mut T, U) -> T
&atomic_umax(…)~(use std::intrinsics::atomic_umax) unsafe fn(*mut T, T) -> T
&atomic_umin(…)~(use std::intrinsics::atomic_umin) unsafe fn(*mut T, T) -> T
&atomic_xchg(…)~(use std::intrinsics::atomic_xchg) unsafe fn(*mut T, T) -> T
&alloc_zeroed(…)~(use std::alloc::alloc_zeroed) unsafe fn(Layout) -> *mut u8
```

Some possible solutions:

- Go back to `could_unify_with_deeply` or `Type::eq`
- Do not trigger `ref_match` on item with imports
- Reduce the item score of `import + ref_match`

ChayimFriedman2:
This does seem unreasonable. If `T` can be anything, it can as well be `Option<i32>`.

I also don't see the problem with those suggestions. They all seem correct.

A4-Tacks:
But this is full of completion list headers, refmatch and import have high scores

These items are rarely used, but are always suggested, making it difficult to see truly useful completion items

ChayimFriedman2:
If we don't have a more specific criteria, as it seems to be the case here, then showing them first is correct.

PR #22295, opened 2026-05-06
Use runnable kind to disambiguate between cargo and shell commands

sunshowers:
_See https://github.com/zed-industries/zed/pull/54011 for the original fix to Zed. That project had copied over these type definitions from rust-analyzer, so we should fix that here too, even though I don't believe r-a deserializes these in production._

Found this bug while investigating why configuring nextest based on the instructions at https://github.com/rust-lang/rust-analyzer/issues/21137#issuecomment-4254611341 wasn't working within Zed.

Previously, we'd use `serde(untagged)`, preferring cargo over shell commands. The problem is that every instance of a shell command is a valid instance of a cargo command. For example, the shell command:

```json
{
    "label": "test my_test",
    "kind": "shell",
    "args": {
        "environment": {"RUSTC_TOOLCHAIN": "/path/to/toolchain"},
        "cwd": "/project",
        "program": "cargo",
        "args": ["nextest", "run", "--package", "my-crate", "--lib", "--", "my_test", "--exact", "--include-ignored"]
    }
}
```

would end up getting deserialized as a Cargo command, silently dropping `program` and `args`.

With this fix, we now use the provided `kind` as a tag. We do have to introduce a `#[serde(flatten)]` unfortunately, which has a few side effects due to internal buffering, but `#[serde(untagged)]` also does internal buffering so this doesn't make things worse.

### DIFF
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -1049,7 +1049,6 @@ pub(crate) fn handle_runnables(
                         all_targets = if all_targets { " --all-targets" } else { "" }
                     ),
                     location: None,
-                    kind: lsp_ext::RunnableKind::Cargo,
                     args: lsp_ext::RunnableArgs::Cargo(lsp_ext::CargoRunnableArgs {
                         workspace_root: Some(spec.workspace_root.clone().into()),
                         cwd: cwd.into(),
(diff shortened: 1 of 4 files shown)


sunshowers:
I'm wondering though if r-a should just drop the `Deserialize` impl for these types entirely, or restrict it to tests.

sunshowers:
(going to follow up on the test failure tomorrow)

Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22295/43e333ab4f863c7e83d0c1483d906b25258dc511..626137b8380dea147eecb550d4ea0aa7fca1736e)*

Veykril:
> I'm wondering though if r-a should just drop the `Deserialize` impl for these types entirely, or restrict it to tests.

I think restricting to tests would make sense. I don't think r-a would ever need to deserialize these.

Veykril:
Turns out we can't test cfg these due to the Request trait requirements on its assoc types

PR #22299, opened 2026-05-06
fix: Generally fix derive helper resolution in semantics

Veykril:


crates/rust-analyzer/src/reload.rs
```diff
@@ -440,7 +440,7 @@ impl GlobalState {
                     let expansion_res = match client {
                         Some(Ok(client)) => match res {
                             Ok((crate_name, path)) => {
-                                progress(format!("loading proc-macros: {path}"));
```
Veykril, line 443: drive by fix as we duplicate the `Loading ...` part in the label 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22299/06aba0c8d642a9c637079f19bb5e2ae88f3ec035..6792359dbee9aeaec6b69f23a2e474eb247eaf0b)*

PR #22303, opened 2026-05-06
Fix: renaming mut vars removed `mut` in patterns generated by macro

### LINKED ISSUE #22301: Renaming variables in a pattern generated by a macro consumes mut keyword
Renaming a pattern which is generated by a macro removes the `mut` keyword.

**rust-analyzer version**: 1.95.0
**rustc version**: 1.95.0

```rust

macro_rules! pat_macro {
    ($pat:pat) => {
        $pat
    };
}

pub fn main() {
    let option = Some(1);
    match option {
        pat_macro!(Some(mut old)) => {
            println!("Some: {}", old);
            old += 1;
            println!("Some +1: {}", old);
        }
        None => {
            println!("None");
        }
    }
}


```

When renaming the `old` in the match statement to something like `new`, the `mut` keyword is removed.

```diff
-        pat_macro!(Some(mut old)) => {
+        pat_macro!(Some(new)) => {
```

BramVerb:
Renaming a pattern which is generated by a macro removed the `mut` keyword.
This also fixes it for `ref` (see tests).

Fixes rust-lang/rust-analyzer#22301

### DIFF
--- a/crates/ide/src/rename.rs
+++ b/crates/ide/src/rename.rs
@@ -4085,19 +4085,16 @@ impl<'a, T> Foo<'a, T> {}
         check(
             "new",
             r#"
+//- minicore: option
 macro_rules! pat_macro {
     ($pat:pat) => {
         $pat
     };
 }
-enum CustomOption<T> {
-    None,
-    Some(T),
-}
 
 pub fn main() {
-    match CustomOption::None {
-        pat_macro!(CustomOption::Some(mut old$0)) => {
+    match None {
+        pat_macro!(Some(mut old$0)) => {
             old += 1,
         }
         None => {}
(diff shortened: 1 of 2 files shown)


crates/ide/src/rename.rs
```diff
@@ -4076,6 +4076,99 @@ impl<T> Foo<'_$0, T> {}
...
+    fn test_rename_mut_pattern_with_macro() {
+        check(
+            "new",
+            r#"
+macro_rules! pat_macro {
+    ($pat:pat) => {
+        $pat
+    };
+}
+enum CustomOption<T> {
```
A4-Tacks: You can use `//- minicore: option`


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22303/b19be37a8ef6f4cd4b5e26ed8c78584adaf4ea5e..2d300e04fa356f7de8c0000511c2b4e7e0bfae3f)*

A4-Tacks (changes requested):

A4-Tacks (approved):

PR #22308, opened 2026-05-07
Migrate split import to syntax editor

bit-aloo:

### DIFF
--- a/crates/syntax/src/ast/edit_in_place.rs
+++ b/crates/syntax/src/ast/edit_in_place.rs
@@ -147,6 +147,35 @@ impl ast::UseTree {
         }
     }
 
+    /// Editor variant of `split_prefix`
+    pub fn split_prefix_with_editor(&self, editor: &SyntaxEditor, prefix: &ast::Path) {
+        debug_assert_eq!(self.path(), Some(prefix.top_path()));
+
+        let make = editor.make();
+        let path = self.path().unwrap();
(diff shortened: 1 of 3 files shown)


crates/syntax/src/ast/edit_in_place.rs
```diff
@@ -147,6 +147,35 @@ impl ast::UseTree {
...
+                make.use_tree(self_path, None, None, false)
+            }
+        } else {
+            let suffix_segments = path.segments().skip(prefix.segments().count());
+            let suffix_path = make.path_from_segments(suffix_segments, false);
+            make.use_tree(
+                suffix_path,
+                self.use_tree_list(),
+                self.rename(),
+                self.star_token().is_some(),
```
ChayimFriedman2, line 170: Are you sure we should use `self`'s elements here, and not `prefix`'s?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22308/6137693d2756c6c69aa8035748b9b72b751b250e..5ccf69fa704394df93e7649e67de75788e8bb8fe)*

bit-aloo: Yes, because those belong to the original `UseTree` and after engulfing the list we prepend the prefix at the end. Prefix at this point is just a path?

ChayimFriedman2 (approved):

PR #22315, opened 2026-05-07
fix: add whitespaces on postfix completion in macro

A4-Tacks:
Example
---
```rust
macro_rules! id { ($($t:tt)*) => ($($t)*); }
fn main() { id!(&raw const x.db$0) }
```

**Before this PR**

```rust
fn main() { id!(dbg!(&rawconstx)) }
```

**After this PR**

```rust
fn main() { id!(dbg!(&raw const x)) }
```

### DIFF
--- a/crates/ide-completion/src/completions/postfix.rs
+++ b/crates/ide-completion/src/completions/postfix.rs
@@ -12,7 +12,7 @@ use ide_db::{
     text_edit::TextEdit,
     ty_filter::TryEnum,
 };
-use itertools::Itertools;
+use itertools::{Either, Itertools};
 use stdx::never;
 use syntax::{
     SmolStr,
@@ -468,6 +468,11 @@ fn include_references(initial_element: &ast::Expr) -> (ast::Expr, String) {
                 .syntax()
                 .children_with_tokens()
                 .filter(|it| Some(it) != last_child_or_token.as_ref())
+                .flat_map(|it| {
+                    let has_ws = it.next_sibling_or_token().is_some_and(|it| it.kind().is_trivia());
+                    let need_ws = !has_ws && it.kind().is_any_identifier();
+                    itertools::chain([Either::Left(it)], need_ws.then_some(Either::Right(" ")))
+                })
                 .format("")
                 .to_smolstr()
                 .as_str(),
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (changes requested):
The problem with prettyifying like here is that it lose original whitespaces and comments. Which is why I prefer fetching the original file text at the macro-mapped range, like we do in several places.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22315/e26f9eee3c9f3c579377e148f8ee41b1916ccb3f..e1f82d45efa92f194633a6c2ff3eea3c55de2375)*

A4-Tacks:
Due to the need to exclude receiver, it needs to be obtained range twice, Is it worth doing this?

ChayimFriedman2:
What do you mean by "it needs to be obtained range twice"?

If you think it's not worth it, I'll merge this, but I do think it is worth it.

A4-Tacks:
> What do you mean by "it needs to be obtained range twice"?

For each `RefExpr` inside receiver, call `sema.original_range_opt` twice

I want to know if this method is fast enough

A4-Tacks:
And, simply adding whitespace to idents is very lightweight

And, after obtaining the original range, it is necessary to index it in the source code text.
Although `get_receiver_text` did the same thing, it doesn't look very good

ChayimFriedman2:
> For each `RefExpr` inside receiver, call `sema.original_range_opt` twice

Why twice? Anyway, I stand by my saying that I think it is worth it, and reminding that if you disagree I agree to merge this PR.

A4-Tacks:
```
&mut foo
^^^^^^^^ a
     ^^^ b
^^^^^    a-b
```
Pseudo code:
```
let ref_range = sema.original_range_opt(ref_expr.syntax());
let receiver_range = sema.original_range_opt(receiver.syntax());
let range = TextRange::new(ref_range.range.start(), receiver_range.range.start())
let text = sema.db.file_text(ref_range.file_id.file_id(sema.db))[range];
prefix.insert_str(0, text);
```

In summary, I believe that a direct merge is the best option

ChayimFriedman2 (approved):

PR #22320, opened 2026-05-08
Migrate expand macro to syntax editor

bit-aloo:

### DIFF
--- a/crates/ide/src/expand_macro.rs
+++ b/crates/ide/src/expand_macro.rs
@@ -6,7 +6,8 @@ use ide_db::{
 };
 use span::{SpanMap, TextRange, TextSize};
 use stdx::format_to;
-use syntax::{AstNode, NodeOrToken, SyntaxKind, SyntaxNode, T, ast, ted};
+use syntax::syntax_editor::SyntaxEditor;
+use syntax::{AstNode, NodeOrToken, SyntaxKind, SyntaxNode, T, ast};
 
 use crate::FilePosition;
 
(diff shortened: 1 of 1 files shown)


crates/ide/src/expand_macro.rs
```diff
@@ -175,6 +176,7 @@ fn expand(
     result_span_map: &mut SpanMap,
     mut offset_in_original_node: i32,
 ) -> SyntaxNode {
+    let (editor, _) = SyntaxEditor::new(expanded.clone());
```
ChayimFriedman2: Just noting that this won't work with lifetime'd nodes. Why can't we use one editor?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22320/ecc571a4138e719a37af19dd8666e99a2887ee09..8871bad09b8ddd1d8334ead66a901a0e561ff1e0)*

ChayimFriedman2 (approved):

crates/ide/src/expand_macro.rs
bit-aloo: Each recursive expansion produces a separate detached syntax tree, and that tree is not part of the parent expansion’s syntax tree until we replace the corresponding macro call with it. SyntaxEdito` can only target nodes that belong to the root it was created for, so one editor for the initial tree cannot edit inside nested expansion trees before they have been inserted. Using a single editor can be difficult.

That said, I do get your concern about lifetime’d nodes. This won't work for sure. I have note'd this down. Let me think and get back. We have a number so such different instances.

ChayimFriedman2: What I imagine is that e.g. in expanding:
```rust
nested1! {
    nested2! { nested3!() }
}
```
We replace the `nested3!()` call in the editor with its expansion in the expansion of `nested2! {}`, then replace the `nested2! {}` in the expansion of `nested1! {}`, and finally the `nested1! {}` in the source file. Then, the `SyntaxEditor` takes care of composing the final node.

I don't know if this can work with today's `SyntaxEditor`, but IMO it should.

PR #22322, opened 2026-05-08
perf: Provide better incrementality for modules

ChayimFriedman2:
Previously module ID reuse was based on order of creation solely. Now we store the parent module and name, and so Salsa will reuse the ID for modules that have the same name and parent.

This is important as many interned IDs contain modules, so if the module is invalidated they too are.

### DIFF
--- a/crates/hir-def/src/lib.rs
+++ b/crates/hir-def/src/lib.rs
@@ -48,7 +48,7 @@ pub mod find_path;
 pub mod import_map;
 pub mod visibility;
 
-use intern::Interned;
+use intern::{Interned, sym};
 use rustc_abi::ExternAbi;
 use thin_vec::ThinVec;
 
@@ -457,6 +457,11 @@ pub struct ModuleIdLt<'db> {
     /// `BlockId` of that block expression. If `None`, this module is part of the crate-level
     /// `DefMap` of `krate`.
     pub block: Option<BlockId>,
+    /// The parent module of this module, or `None` if this is the root module inside the def
+    /// map (including for block def maps).
+    pub containing_module_inside_def_map: Option<ModuleIdLt<'db>>,
+    /// The name of this module, or [`sym::__empty`] for the root module.
+    name_or_empty: Name,
 }
 pub type ModuleId = ModuleIdLt<'static>;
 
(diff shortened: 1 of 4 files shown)


crates/hir-def/src/lib.rs
```diff
@@ -457,6 +457,11 @@ pub struct ModuleIdLt<'db> {
     /// `BlockId` of that block expression. If `None`, this module is part of the crate-level
     /// `DefMap` of `krate`.
     pub block: Option<BlockId>,
+    /// The parent module of this module, or `None` if this is the root module inside the def
+    /// map (including for block def maps).
+    pub containing_module_inside_def_map: Option<ModuleIdLt<'db>>,
```
Veykril, line 462: why not just call this `parent`?

crates/hir-def/src/lib.rs
```diff
@@ -502,21 +507,23 @@ impl ModuleId {
     }
 
     pub fn name(self, db: &dyn DefDatabase) -> Option<Name> {
-        let def_map = self.def_map(db);
-        let parent = def_map[self].parent?;
```
Veykril, line 506: Do we need the `parent` field on the `ModuleData` now or can we discard this in favor of the IDs field?

crates/hir-def/src/lib.rs
```diff
@@ -457,6 +457,11 @@ pub struct ModuleIdLt<'db> {
     /// `BlockId` of that block expression. If `None`, this module is part of the crate-level
     /// `DefMap` of `krate`.
     pub block: Option<BlockId>,
+    /// The parent module of this module, or `None` if this is the root module inside the def
+    /// map (including for block def maps).
+    pub containing_module_inside_def_map: Option<ModuleIdLt<'db>>,
+    /// The name of this module, or [`sym::__empty`] for the root module.
+    name_or_empty: Name,
```
Veykril, line 464: Would also call this just `name`, keeping the doc comment.

Veykril (commented):
Nice catch

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22322/eefa15819364cc01d3d499cef8db5fce354cd679..406af24b55e56d2e2f2ce0ba077edc93dd26ae0b)*

crates/hir-def/src/lib.rs
ChayimFriedman2: It's still used in one place but we can derive it from the module instead of the def map.

crates/hir-def/src/lib.rs
ChayimFriedman2: This will clash with the method (because Salsa generates a method), a frankly I'm happier with the current name. It removes any chances of mistake.

crates/hir-def/src/lib.rs
ChayimFriedman2: I wanted to be explicit. But I can change that, if you want.

ChayimFriedman2:
@Veykril what's with this?

Veykril (approved):

PR #22323, opened 2026-05-08
Migrate derive macro to syntax editor

bit-aloo:

### DIFF
--- a/crates/hir-expand/src/builtin/derive_macro.rs
+++ b/crates/hir-expand/src/builtin/derive_macro.rs
@@ -25,7 +25,6 @@ use syntax::{
         HasName, HasTypeBounds, make,
     },
     syntax_editor::{GetOrCreateWhereClause, SyntaxEditor},
-    ted,
 };
 
 macro_rules! register_builtin {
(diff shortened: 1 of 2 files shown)


crates/hir-expand/src/builtin/derive_macro.rs
```diff
@@ -1403,37 +1382,84 @@ fn coerce_pointee_expand(
     }
 
     /// Returns true if any substitution was performed.
-    fn substitute_type_in_bound(ty: ast::Type, param_name: &str, replacement: &str) -> bool {
+    fn substitute_type_bound(
+        bound: ast::TypeBound,
+        param_name: &str,
+        replacement: &str,
+    ) -> (ast::TypeBound, bool) {
```
ChayimFriedman2: Better to return `Option<ast::TypeBound>`.

crates/hir-expand/src/builtin/derive_macro.rs
```diff
@@ -1403,37 +1382,84 @@ fn coerce_pointee_expand(
...
+        let (editor, bound) = SyntaxEditor::with_ast_node(&bound);
```
ChayimFriedman2, line 1396: What I said in the past about nested editors applies here too. I imagine a `SyntaxEditor` that allows us to make changes to `bound.clone_subtree()`, then insert it into the top editor.

ChayimFriedman2 (approved):
One nit.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22323/73ca1d4ef136e2367627d698f00db94882b49440..ae9486a141da8c8cdf719024ff38192931fbe9e2)*

crates/hir-expand/src/builtin/derive_macro.rs
bit-aloo: This can be done. We just need to make some changes to the mapping so that we can allow intermediate output nodes inside an editor to be mapped.

ChayimFriedman2 (approved):

PR #22326, opened 2026-05-09
ide-diagnostics: emit error for duplicate field in record pattern

MavenRain:
Resolves the second of the FIXMEs called out in rust-lang/rust-analyzer#22140, parallel to rust-lang/rust-analyzer#22235 which handled the record-expression case.  The diagnostic infrastructure was added in rust-lang/rust-analyzer#22235; this PR only needs the push at the new site plus the two pattern tests.


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22326/73ca1d4ef136e2367627d698f00db94882b49440..af1588462290467792df36244bdaa784b3b6b49d)*

PR #22332, opened 2026-05-10
fix: offer on path_type for add_lifetime_to_type

A4-Tacks:
Example
---
```rust
struct Foo {
    a: &'_ i32,
    b: Foo<'_$0>,
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
struct Foo<'a> {
    a: &'a i32,
    b: Foo<'a>,
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/add_lifetime_to_type.rs
+++ b/crates/ide-assists/src/handlers/add_lifetime_to_type.rs
@@ -1,6 +1,8 @@
+use either::Either;
 use syntax::{
-    SyntaxKind, SyntaxNode, SyntaxToken,
+    SyntaxKind, SyntaxNode,
     ast::{self, AstNode, HasGenericParams, HasName},
+    match_ast,
 };
 
 use crate::{AssistContext, AssistId, Assists};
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/add_lifetime_to_type.rs
```diff
@@ -67,6 +65,16 @@ pub(crate) fn add_lifetime_to_type(acc: &mut Assists, ctx: &AssistContext<'_, '_
     })
 }
 
+fn trigger_assist(ctx: &AssistContext<'_, '_>) -> bool {
+    ctx.find_node_at_offset::<ast::RefType>()
```
ChayimFriedman2: I think it's better to check either for `ast::Lifetime` or for a without-lifetime `ast::RefType`. Technically it will be nice if the assist will also trigger for missing lifetime params in ADTs, but I think this is better done by having a diagnostic for this with a quickfix (that can supersede this assist).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22332/4e97512c6ef1b375b202b6bf5d3a9b1ac54723cb..ea9b42dfb9fdf81ab3e9641b74b85d80c4e12457)*

A4-Tacks: > I think it's better to check either for `ast::Lifetime` or for a without-lifetime `ast::RefType`.

The text range of `'_` is too small, which may not be conducive to user use

> but I think this is better done by having a diagnostic for this with a quickfix (that can supersede this assist).

After the diagnostics stabilize, consider migration

> Technically it will be nice if the assist will also trigger for missing lifetime params in ADTs

Good idea

ChayimFriedman2: > The text range of '_ is too small, which may not be conducive to user use

IMO it's confusing if we offer this on unrelated parts of the path.

A4-Tacks:
This assist always needs to be applied or not triggered because it is unlikely to have a false positive

Perhaps it's a good idea for the entire Adt application, what do you think?

ChayimFriedman2:
> This assist always needs to be applied or not triggered because it is unlikely to have a false positive

What do you mean?

A4-Tacks:
For an Adt in development, exists a missing lifetime argument:

- When exists a lifetime parameter: assist not applicable
- When non-exists a lifetime parameter: you always need to add lifetime parameters and arguments
  - uses this assist: At most, it's just an additional renaming
  - not uses this assist:
    - add a lifetime parameter: assist not applicable because exists a lifetime parameter
    - add a lifetime arguments: assist not applicable because not-exists missing lifetime

A4-Tacks:
Anyway, let's implement it this way first. I have placed an XXX comment

PR #22334, opened 2026-05-10
feat: add diagnostic for E0529

WaterWhisperer:

### DIFF
--- a/crates/hir-ty/src/infer/pat.rs
+++ b/crates/hir-ty/src/infer/pat.rs
@@ -1620,7 +1620,10 @@ https://doc.rust-lang.org/reference/types.html#trait-objects";
             TyKind::Slice(element_ty) => (element_ty, Some(expected), expected),
             // The expected type must be an array or slice, but was neither, so error.
             _ => {
-                // FIXME: Emit an error: expected an array or a slice.
+                self.push_diagnostic(InferenceDiagnostic::ExpectedArrayOrSlicePat {
+                    pat,
+                    found: expected.store(),
+                });
                 let err = self.types.types.error;
                 (err, Some(err), err)
             }
--- a/crates/hir-ty/src/mir/lower/pattern_matching.rs
+++ b/crates/hir-ty/src/mir/lower/pattern_matching.rs
@@ -248,9 +248,18 @@ impl<'db> MirLowerCtx<'_, 'db> {
                 (current, current_else)
             }
             Pat::Slice { prefix, slice, suffix } => {
+                let pat_ty = self.infer.pat_ty(pattern);
+                // FIXME: MIR lowering should be skipped for bodies with inference errors. Once
+                // that happens, this recovery for invalid slice patterns can be removed.
+                if !matches!(pat_ty.kind(), TyKind::Array(..) | TyKind::Slice(_)) {
+                    return Err(MirLowerError::TypeError(
+                        "non array or slice type matched with slice pattern",
+                    ));
+                }
+
                 if mode == MatchingMode::Check {
                     // emit runtime length check for slice
-                    if let TyKind::Slice(_) = self.infer.pat_ty(pattern).kind() {
+                    if let TyKind::Slice(_) = pat_ty.kind() {
                         let pattern_len = prefix.len() + suffix.len();
                         let place_len: Place = self
                             .temp(Ty::new_usize(self.interner()), current, pattern.into())?
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -295,6 +295,11 @@ pub enum InferenceDiagnostic {
         #[type_visitable(ignore)]
         has_rest: bool,
     },
+    ExpectedArrayOrSlicePat {
+        #[type_visitable(ignore)]
+        pat: PatId,
+        found: StoredTy,
+    },
     DuplicateField {
         #[type_visitable(ignore)]
         field: ExprOrPatId,
--- a/crates/hir-ty/src/infer/unify.rs
+++ b/crates/hir-ty/src/infer/unify.rs
@@ -578,6 +578,7 @@ pub(super) mod resolve_completely {
                 self.resolve_completely(diagnostic);
 
                 if let InferenceDiagnostic::ExpectedFunction { found: ty, .. }
+                | InferenceDiagnostic::ExpectedArrayOrSlicePat { found: ty, .. }
                 | InferenceDiagnostic::UnresolvedField { receiver: ty, .. }
                 | InferenceDiagnostic::UnresolvedMethodCall { receiver: ty, .. } = diagnostic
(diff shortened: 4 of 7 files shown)


crates/hir-ty/src/infer/pat.rs
```diff
@@ -1620,7 +1620,12 @@ https://doc.rust-lang.org/reference/types.html#trait-objects";
             TyKind::Slice(element_ty) => (element_ty, Some(expected), expected),
             // The expected type must be an array or slice, but was neither, so error.
             _ => {
-                // FIXME: Emit an error: expected an array or a slice.
+                if !expected.references_error() {
```
ChayimFriedman2: `references_error()` is a bad idea to use in r-a as we have many lifetime errors. You can use `references_non_lt_error()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22334/0dc46bfea9b17fed93ab57393971fb932e7a5dd4..ec418834c4c625e41065a2f941395a99350a41cf)*

crates/hir-ty/src/mir/lower/pattern_matching.rs
```diff
@@ -248,9 +248,16 @@ impl<'db> MirLowerCtx<'_, 'db> {
                 (current, current_else)
             }
             Pat::Slice { prefix, slice, suffix } => {
+                let pat_ty = self.infer.pat_ty(pattern);
+                if !matches!(pat_ty.kind(), TyKind::Array(..) | TyKind::Slice(_)) {
```
ChayimFriedman2, line 254: Why is this here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22334/0dc46bfea9b17fed93ab57393971fb932e7a5dd4..ec418834c4c625e41065a2f941395a99350a41cf)*

crates/hir-ty/src/infer/pat.rs
```diff
@@ -1696,7 +1701,7 @@ https://doc.rust-lang.org/reference/types.html#trait-objects";
             // We have a variable-length pattern and don't know the array length.
             // This happens if we have e.g.,
             // `let [a, b, ..] = arr` where `arr: [T; N]` where `const N: usize`.
-            // FIXME: Emit an error: cannot pattern-match on an array without a fixed length.
+            self.push_diagnostic(InferenceDiagnostic::ArrayPatWithUnknownLen { pat });
```
ChayimFriedman2: This is tricky as our const eval is incomplete and can fail for many reasons. So either you check if evaluation failed, which is something `try_const_usize()` currently does not allow you to do, or you cannot emit this error.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22334/0dc46bfea9b17fed93ab57393971fb932e7a5dd4..ec418834c4c625e41065a2f941395a99350a41cf)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/infer/pat.rs
WaterWhisperer: Thanks. I'll update it.

crates/hir-ty/src/mir/lower/pattern_matching.rs
WaterWhisperer: For the E0529 test `let [_a, _b] = x`, without it, the test still reaches `Pat::Slice` MIR lowering after inference has produced the type error, then panics with `Overloaded index is not a projection`.

ChayimFriedman2: We should not lower MIR when there are inference errors. Don't do that in this PR, but please leave a FIXME for that here.

crates/hir-ty/src/infer/pat.rs
WaterWhisperer: I see. I'll remove this part.

crates/hir-ty/src/mir/lower/pattern_matching.rs
WaterWhisperer: Understood, thanks!

crates/hir-ty/src/infer/pat.rs
```diff
@@ -1620,7 +1620,12 @@ https://doc.rust-lang.org/reference/types.html#trait-objects";
             TyKind::Slice(element_ty) => (element_ty, Some(expected), expected),
             // The expected type must be an array or slice, but was neither, so error.
             _ => {
-                // FIXME: Emit an error: expected an array or a slice.
+                if !expected.references_non_lt_error() {
```
ChayimFriedman2: Actually you shouldn't do the filtering here, as there can be unresolved variables. You should do it in `resolve_completely()`, more specifically in `WriteBackCtxt::resolve_diagnostics()`, near the other functions it handles.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22334/0dc46bfea9b17fed93ab57393971fb932e7a5dd4..b9600b67387dacc4b77fde85cd33aa0e679ed083)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/mir/lower/pattern_matching.rs
```diff
@@ -248,6 +248,8 @@ impl<'db> MirLowerCtx<'_, 'db> {
                 (current, current_else)
             }
             Pat::Slice { prefix, slice, suffix } => {
+                // FIXME: MIR lowering should be skipped for bodies with inference errors. Otherwise
+                // invalid slice patterns can reach this code with an error type.
```
ChayimFriedman2: I meant, still put the `if`, we need the tests after all, but also put the FIXME to remove it when we no longer lower MIR with errors.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22334/0dc46bfea9b17fed93ab57393971fb932e7a5dd4..60125572b743f730010ee1351424706e58af93df)*

ChayimFriedman2 (changes requested):

ChayimFriedman2:
Ignore the CI failure; it happens sporadically (we need to figure out why).

crates/hir-ty/src/mir/lower/pattern_matching.rs
WaterWhisperer: Oh, sorry for misunderstanding

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22334/0dc46bfea9b17fed93ab57393971fb932e7a5dd4..6c46dc62be951a7bf1e5b566277db0bcb72faceb)*

PR #22335, opened 2026-05-10
fix: add param on result methods for replace_method_eager_lazy

A4-Tacks:
- **Add some Result methods to minicore**
- **Migrate replace_method_eager_lazy to SyntaxEdutor**

Example
---
```rust
fn foo() {
    let foo = Ok(1);
    return foo.unwrap_$0or(2);
}
```

**Before this PR**

```rust
fn foo() {
    let foo = Ok(1);
    return foo.unwrap_or_else(|| 2);
}
```

**After this PR**

```rust
fn foo() {
    let foo = Ok(1);
    return foo.unwrap_or_else(|${0:_}| 2);
}

### DIFF
--- a/crates/ide-assists/src/handlers/replace_method_eager_lazy.rs
+++ b/crates/ide-assists/src/handlers/replace_method_eager_lazy.rs
@@ -1,5 +1,5 @@
 use hir::Semantics;
-use ide_db::{RootDatabase, assists::AssistId, defs::Definition};
+use ide_db::{RootDatabase, assists::AssistId, defs::Definition, famous_defs::FamousDefs};
 use syntax::{
     AstNode,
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (changes requested):
I think it's better to name the param `_` given it's not used, and also place a snippet on it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22335/0dc46bfea9b17fed93ab57393971fb932e7a5dd4..a9e371fb4435cac982897e729a5d61c8e18a4278)*

A4-Tacks:
It seems good, I hope this won't break some habits

ChayimFriedman2 (approved):

PR #22336, opened 2026-05-10
Emit missing-fields diagnostics from inference

Twil3akine:
Implements one diagnostic FIXME from rust-lang/rust-analyzer#22140.

This changes record-literal missing-fields diagnostics to be emitted from inference, while reusing the existing `MissingFields` diagnostic.

`MissingFields` already represents record literal/pattern missing field diagnostics and is rendered as rustc hard error `E0063`, so adding a separate diagnostic would duplicate behavior.

This PR:
- reuses the existing `MissingFields` diagnostic
- moves record-literal missing-fields emission from body validation into inference
- keeps the existing record-pattern body validation path intact
- does not add a new ide-assist or quick fix

Validated with:

```bash
cargo +nightly-2026-04-30 test -p ide-diagnostics missing_fields
cargo +nightly-2026-04-30 test -p ide-diagnostics
cargo +nightly-2026-04-30 test -p hir-ty
cargo +nightly-2026-04-30 test -p hir
```

AI assistance disclosure:
I used ChatGPT to help review the implementation plan and draft this PR description. I reviewed the generated suggestions before submitting.

### DIFF
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -1096,7 +1096,11 @@ impl<'db> InferenceContext<'_, 'db> {
                     }
                 }
                 if !missing_mandatory_fields.is_empty() {
-                    // FIXME: Emit an error: missing fields.
+                    self.push_diagnostic(InferenceDiagnostic::MissingFields {
+                        expr,
+                        fields: missing_mandatory_fields,
+                        variant,
+                    });
                 }
             }
             RecordSpread::Expr(base_expr) => {
@@ -1188,7 +1192,20 @@ impl<'db> InferenceContext<'_, 'db> {
                 {
                     debug!(?remaining_fields);
 
-                    // FIXME: Emit an error: missing fields.
+                    let missing_fields = variant_fields
+                        .fields()
+                        .iter()
+                        .filter_map(|(field_idx, field)| {
+                            remaining_fields.contains_key(&field.name).then_some(field_idx)
+                        })
+                        .collect::<Vec<_>>();
+                    if !missing_fields.is_empty() {
+                        self.push_diagnostic(InferenceDiagnostic::MissingFields {
+                            expr,
+                            fields: missing_fields,
+                            variant,
+                        });
+                    }
                 }
             }
         }
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -1192,13 +1192,7 @@ impl<'db> InferenceContext<'_, 'db> {
                 {
                     debug!(?remaining_fields);
 
-                    let missing_fields = variant_fields
-                        .fields()
-                        .iter()
-                        .filter_map(|(field_idx, field)| {
-                            remaining_fields.contains_key(&field.name).then_some(field_idx)
-                        })
-                        .collect::<Vec<_>>();
+                    let missing_fields = remaining_fields.values().copied().collect();
                     if !missing_fields.is_empty() {
                         self.push_diagnostic(InferenceDiagnostic::MissingFields {
                             expr,
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -1193,13 +1193,12 @@ impl<'db> InferenceContext<'_, 'db> {
                     debug!(?remaining_fields);
 
                     let missing_fields = remaining_fields.values().copied().collect();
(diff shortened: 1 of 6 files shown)


crates/hir-ty/src/infer/expr.rs
```diff
@@ -1182,7 +1186,20 @@ impl<'db> InferenceContext<'_, 'db> {
...
                     debug!(?remaining_fields);
 
-                    // FIXME: Emit an error: missing fields.
+                    let missing_fields = variant_fields
+                        .fields()
+                        .iter()
+                        .filter_map(|(field_idx, field)| {
+                            remaining_fields.contains_key(&field.name).then_some(field_idx)
+                        })
+                        .collect::<Vec<_>>();
```
ChayimFriedman2:
```suggestion
                    let missing_fields = remaining_fields.values().copied().collect();
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22336/0dc46bfea9b17fed93ab57393971fb932e7a5dd4..793d5435d779848131339805b37d878d63e086ec)*

crates/hir-ty/src/infer/expr.rs
```diff
@@ -1182,7 +1186,20 @@ impl<'db> InferenceContext<'_, 'db> {
...
+                    if !missing_fields.is_empty() {
```
ChayimFriedman2: This `if` is redundant, we already checked that above.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22336/0dc46bfea9b17fed93ab57393971fb932e7a5dd4..793d5435d779848131339805b37d878d63e086ec)*

crates/hir/src/diagnostics.rs
```diff
@@ -786,6 +786,37 @@ impl<'db> AnyDiagnostic<'db> {
...
+                        )
+                    })
+                    .collect();
+
+                let record = expr_syntax(*expr)?;
+                let file = record.file_id;
+                let root = record.file_syntax(db);
+                let Either::Left(ast::Expr::RecordExpr(record_expr)) = record.value.to_node(&root)
+                else {
+                    return None;
```
ChayimFriedman2, line 888:
```suggestion
                    never!("should always map to a `ast::RecordExpr`");
                    return None;
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22336/0dc46bfea9b17fed93ab57393971fb932e7a5dd4..793d5435d779848131339805b37d878d63e086ec)*

crates/hir/src/lib.rs
```diff
@@ -2191,7 +2191,12 @@ impl DefWithBody {
         expr_store_diagnostics(db, acc, source_map);
 
         let infer = InferenceResult::of(db, id);
+        let mut delayed_missing_fields_diagnostics = Vec::new();
```
ChayimFriedman2: Why do you delay this? I see no reason for that.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22336/0dc46bfea9b17fed93ab57393971fb932e7a5dd4..793d5435d779848131339805b37d878d63e086ec)*

ChayimFriedman2 (changes requested):

ChayimFriedman2:
@Twil3akine, what's the status of this?

Twil3akine:
@ChayimFriedman2 

Thanks for the follow-up.

I've rebased the PR and resolved the conflicts.

Regarding the diagnostic ordering change mentioned above: should I proceed with updating the tests so they don't depend on the diagnostic order? I'd appreciate your thoughts.

ChayimFriedman2:
Diagnostics order should not affect anything. If there are tests that depend on it, that's weird and maybe a bug. Feel free to update them. Definitely don't complicate the code to change ordering, anyway.

Twil3akine:
Sorry for the late reply.

This may be due to my implementation, but removing the delay causes some tests to fail. Would it be okay to proceed by removing the delay and updating the tests accordingly?

ChayimFriedman2:
Remove it, update the tests, and I'll see if there is a bug in them.

Twil3akine:
I done, pushed the update.

I removed the delayed emission and updated the diagnostic fix test helper so it selects the matching fix result instead of relying on diagnostic ordering.

Could you take another look when you have time?

PR #22339, opened 2026-05-11
feat: add diagnostic for E0040

WaterWhisperer:

### DIFF
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -384,6 +384,10 @@ pub enum InferenceDiagnostic {
         call_expr: ExprId,
         found: StoredTy,
     },
+    ExplicitDestructorCall {
+        #[type_visitable(ignore)]
+        expr: ExprId,
+    },
     TypedHole {
         #[type_visitable(ignore)]
         expr: ExprId,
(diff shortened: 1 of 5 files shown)


WaterWhisperer:
I noticed rustc also reports E0040 for forms like `<Foo as Drop>::drop(&mut x)`, but that maybe outside the scope of this PR.

ChayimFriedman2:
@ada4a reported starting to work on this in https://github.com/rust-lang/rust-analyzer/issues/22140, and I don't want to steall their work. @ada4a are you still working on this? If yes, I'll close this PR.

ada4a:
Oh, this is unfortunate... My first instinct would be to say that the contents of this PR seem reasonable, so it'd make sense to merge this one, and then I could add support for the `Drop::drop` and `<Foo as Drop>::drop` in a follow-up PR.

On the other hand, given our discussion in https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Emit.20diagnostic.20when.20lowering.20a.20call.20expr, I wonder whether we should be emitting this diagnostic at this location at all – maybe we should instead only do it in path lowering. But this might be a somewhat more complicated refactoring? So maybe we could, again, merge this PR now and I can look into refactoring later (I was planning to do it this week)

WaterWhisperer:
Oh sorry, this part is folded so I didn't notice it at the first time

image

ChayimFriedman2:
Method resolution does not use path lowering, so we need both.

From my own side, I agree to close this PR as I said given you've declared you're working on this. Of course, if you want to be extra nice and not care, you can.

ada4a:
Okay, this PR has some small changes I want to suggest to change, but also a few of things that I notice were missing from my PR -- let me combine everything into my PR, and continue the work from there. Apologies for the back-and-forth @WaterWhisperer; I'll make sure to add you as a co-author, if that's any consolation

PR #22340, opened 2026-05-11
ide-diagnostics: emit error for explicit destructor calls

MavenRain:
Resolves a FIXME in check_for_illegal_method_calls at hir-ty/src/method_resolution/confirm.rs that previously allowed expressions like `s.drop()` to type-check without error.  Adds an InferenceDiagnostic::ExplicitDestructorCall variant, plumbs the cooked diagnostic through hir, and renders it as RustcHardError E0040 ("explicit use of destructor method").

This catches the method-call form only; the UFCS form `Drop::drop(&mut s)` goes through a different resolution path and is out of scope for this FIXME.

Tests: explicit drop call on a Drop-implementing type errors; the free `drop(s)` function does not; an inherent method named `drop` on a non-Drop type does not.

Part of rust-lang/rust-analyzer#22140.

### DIFF
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -437,6 +437,10 @@ pub enum InferenceDiagnostic {
         #[type_visitable(ignore)]
         has_self_arg: bool,
     },
+    ExplicitDestructorCall {
+        #[type_visitable(ignore)]
+        expr: ExprId,
+    },
     InvalidLhsOfAssignment {
         #[type_visitable(ignore)]
         lhs: ExprId,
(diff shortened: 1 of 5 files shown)


ChayimFriedman2:
Wow, we got *two* pull requests (https://github.com/rust-lang/rust-analyzer/pull/22339) at the same time for work that have been already claimed! That's exceptional.

Given that this PR appeared 1.5 hours after https://github.com/rust-lang/rust-analyzer/pull/22339, I'll close this one and let @ada4a decide what to do with the other one. Sorry for your work and thank you for your interest in contribution!

MavenRain:
> Wow, we got _two_ pull requests (#22339) at the same time for work that have been already claimed! That's exceptional.
> 
> Given that this PR appeared 1.5 hours after #22339, I'll close this one and let @ada4a decide what to do with the other one. Sorry for your work and thank you for your interest in contribution!

All good . . . just glad that someone covered it

PR #22342, opened 2026-05-11
fix(assists/add_reference_here): _modify_ the reference type when dealing with &T->&mut T

### LINKED ISSUE #13047: "Add a reference" assist adds an extra ampersand when fixing to a mutable ref

ada4a:
..instead of adding another layer of reference, i.e. &T -> &mut &T

Fixes https://github.com/rust-lang/rust-analyzer/issues/13047

### DIFF
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -103,13 +103,9 @@ fn add_or_fix_reference(
         let expr = ctx.sema.original_ast_node(expr)?;
         let expr_without_ref = RefExpr::cast(expr.syntax().clone())?.expr()?;
 
-        let file_id = expr_ptr.file_id.original_file(ctx.db());
-        let mut builder = SourceChangeBuilder::new(file_id.file_id(ctx.db()));
-        let editor = builder.make_editor(expr.syntax());
-        let make = editor.make();
-        let new_expr = make.expr_ref(expr_without_ref, true);
-        builder.replace_ast(expr, new_expr);
-        let source_change = builder.finish();
+        let pos = expr_without_ref.syntax().text_range().start();
+        let edit = TextEdit::insert(pos, expected_mutability.as_keyword_for_ref().to_owned());
+        let source_change = SourceChange::from_text_edit(range.file_id, edit);
         acc.push(fix(
             "make_reference_mutable",
             "Make reference mutable",
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -51,10 +51,10 @@ pub(crate) fn type_mismatch(
             format!(
                 "expected {}, found {}",
                 d.expected
-                    .display(ctx.sema.db, ctx.display_target)
+                    .display(ctx.db(), ctx.display_target)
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -79,21 +79,48 @@ fn fixes(ctx: &DiagnosticsContext<'_, '_>, d: &hir::TypeMismatch<'_>) -> Option<
...
+
+        let expr = expr_ptr.to_node(ctx.db());
+        // If the node comes from a macro expansion, then we shouldn't assist,
+        // as the suggestion would overwrite the macro _definition_ position
+        let expr = ctx.sema.original_ast_node(expr)?;
+        let expr_without_ref = RefExpr::cast(expr.syntax().clone())?.expr()?;
+
+        let pos = expr_without_ref.syntax().text_range().start();
+        let edit = TextEdit::insert(pos, "mut ".to_string());
+        let source_change = SourceChange::from_text_edit(range.file_id, edit);
```
ada4a, line 108: I initially wanted to do this with `SyntaxEditor`, but it looks like a text edit works just fine... Let me know which one I should go with.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22342/7c3fc8671f83f6e46305358b98354f0611ebb3cd..fdfffdb1584d4dc3e90c9ff39e5e84b1f40749a2)*

Veykril: SyntaxEditor would likely be better in the future, but this is fine for now

Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22342/d2cf9ceecda33bda5bd30b4d2503bd7fa301053e..05bc639bbf9530d417f3f387caa9b31ba0aa000b)*

crates/ide-diagnostics/src/handlers/type_mismatch.rs
ada4a: I had actually implemented it with SyntaxEditor initially, see the second commit.. I'll open a PR which switches to that then (EDIT: done, see https://github.com/rust-lang/rust-analyzer/pull/22467)

PR #22347, opened 2026-05-12
feat: add diagnostic for E0029

WaterWhisperer:

### DIFF
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -302,6 +302,10 @@ pub enum InferenceDiagnostic {
         pat: PatId,
         found: StoredTy,
     },
+    InvalidRangePatType {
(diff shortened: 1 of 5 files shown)


WaterWhisperer:
Only handles the early range-endpoint check in `infer_range_pat`.

Left the later FIXME because it interacts with type mismatch diagnostics after endpoint unification, cases like `match bool { 1..=2 => {} }` should remain `E0308` rather than gaining an extra `E0029`.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22347/7c3fc8671f83f6e46305358b98354f0611ebb3cd..4524dd8f4808494a1fc45f2f946fbb9b929936c7)*

PR #22351, opened 2026-05-12
Migrate merge imports to editor

bit-aloo:


crates/syntax/src/ast/edit.rs
```diff
@@ -263,6 +265,163 @@ pub fn indent(node: &SyntaxNode, level: IndentLevel) -> SyntaxNode {
...
+            {
+                u.parent_use_tree().remove_recursive(editor);
+                return;
+            }
+            self.remove_with_editor(editor);
+            u.remove_unnecessary_braces(editor);
+        }
+    }
+
+    /// Editor variant of `split_prefix`
```
ChayimFriedman2:
```suggestion
    /// Splits off the given prefix, making it the path component of the use tree,
    /// appending the rest of the path to all UseTreeList items.
    ///
    /// # Examples
    ///
    /// `prefix$0::suffix` -> `prefix::{suffix}`
    ///
    /// `prefix$0` -> `prefix::{self}`
    ///
    /// `prefix$0::*` -> `prefix::{*}````

File: crates/ide-assists/src/handlers/merge_imports.rs [hunk 2]
```diff
@@ -28,16 +28,19 @@ use Edit::*;
...
         let target = tree.syntax().text_range();
 
         let use_item = tree.syntax().parent().and_then(ast::Use::cast)?;
-        let mut neighbor = next_prev().find_map(|dir| neighbor(&use_item, dir)).into_iter();
-        let edits = use_item.try_merge_from(&mut neighbor, &ctx.config.insert_use);
-        (target, edits?)
+        let neighbor = next_prev().find_map(|dir| neighbor(&use_item, dir))?;
+        let parent_node = use_item.syntax().parent()?;
+        let (editor, _) =
+            SyntaxEditor::new(parent_node.ancestors().last().unwrap_or(parent_node.clone()));
```
ChayimFriedman2:
```suggestion
        let (editor, _) =
            SyntaxEditor::new(use_item.syntax().ancestors().last().unwrap());
```

crates/ide-assists/src/handlers/merge_imports.rs
```diff
@@ -50,104 +53,81 @@ pub(crate) fn merge_imports(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -> O
         });
 
         let first_selected = selected_nodes.next()?;
-        let edits = match_ast! {
+        let (editor, _) =
+            SyntaxEditor::new(parent_node.ancestors().last().unwrap_or(parent_node.clone()));
```
ChayimFriedman2:
```suggestion
            SyntaxEditor::new(parent_node.ancestors().last().unwrap());
```

crates/ide-db/src/imports/merge_imports.rs
```diff
@@ -701,14 +690,161 @@ fn get_single_subtree(use_tree: &ast::UseTree) -> Option<ast::UseTree> {
...
+) -> Option<ast::UseTreeList> {
+    let use_tree_list = make.use_tree_list(use_trees);
+    match style_source.and_then(use_tree_list_style) {
+        Some(style) => apply_use_tree_list_style(&use_tree_list, style),
+        _ => Some(use_tree_list),
+    }
+}
+
+fn use_tree_list_style(use_tree_list: &ast::UseTreeList) -> Option<(String, String, bool)> {
+    let text = use_tree_list.syntax().text().to_string();
```
ChayimFriedman2: I really dislike this text manipulation. Please do this with proper AST.

crates/ide-db/src/imports/merge_imports.rs
```diff
@@ -98,56 +100,66 @@ fn try_merge_trees_mut(lhs: &ast::UseTree, rhs: &ast::UseTree, merge: MergeBehav
...
+    let mut lhs_list = lhs.use_tree_list()?.use_trees().collect::<Vec<_>>();
+    let mut use_trees = lhs_list
+        .iter()
+        .enumerate()
         // We use Option here to early return from this function(this is not the
         // same as a `filter` op).
-        .map(|tree| merge.is_tree_allowed(&tree).then_some(tree))
-        .collect::<Option<_>>()?;
+        .map(|(idx, tree)| merge.is_tree_allowed(tree).then_some(idx))
+        .collect::<Option<Vec<_>>>()?;
```
ChayimFriedman2: Why did you change this to be a second `Vec` of indices? The previous setup was clearer IMO.

ChayimFriedman2 (dismissed):
I have not reviewed all of the code yet, only most of it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22351/7c28934677b1e7a1c6ef952422e6ef730540f85f..935f981bfc0b90a8f4af26140bf245817a97bcde)*

crates/syntax/src/ast/edit.rs
bit-aloo: Oopsie!!! Added

crates/ide-assists/src/handlers/merge_imports.rs
bit-aloo: We can use "?" instead of unwrap. Added.

crates/ide-assists/src/handlers/merge_imports.rs
bit-aloo: We can use "?" instead of unwrap. Added.

crates/ide-db/src/imports/merge_imports.rs
bit-aloo: I initially wanted to avoid using SyntaxEditor later on, but if you look at the `apply_use_tree_list` method, it ended up using the editor internally anyway. I should have gone with AST-based manipulation instead of text-based mutation. Thanks, I have now removed the text mutation and switched to AST-based handling

crates/ide-db/src/imports/merge_imports.rs
bit-aloo: Ah!!!!....I think we never needed this, I thought we should preserve the order considering we need to map it correctly (so thats the second list of indices), but doesn't seem the case. Removed it. Thanks.

crates/ide-assists/src/handlers/merge_imports.rs
ChayimFriedman2: I actually prefer the unwrap. A `?` signals a possible failure.

bit-aloo: We would not show the assist, if we return early?

ChayimFriedman2: What I mean is, seeing `unwrap()` I know it cannot fail.

bit-aloo: Ah! That makes sense... I will add an expect

ChayimFriedman2: I prefer an `unwrap()`. While an `expect()` should generally be preferred, the reasoning here is obvious and we do this in many places. The real fix is to have a `tree_top()` method in `SyntaxNode`, but this is not urgent.

bit-aloo: Added unwrap, noted tree_top API in my todo list, will open a PR for it later. Thanks.

crates/ide-assists/src/handlers/merge_imports.rs
bit-aloo: Added unwrap

Veykril (approved):
Changed some minor nits but nothing big

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22351/1edcb122f3766bd2079f7964f75f488146f8b6c9..1ace40df7e8581d62b7fb1b857f665cd959dd2e7)*

PR #22357, opened 2026-05-14
fix: show Run lens for fn main in bench targets

### LINKED ISSUE #21948: Benchmarks don't have a "▶Run" button

MavenRain:
`crates/rust-analyzer/src/handlers/request.rs` enumerates `TargetKind::Bin | TargetKind::Example | TargetKind::Test` in two places (code-lens annotation config at line 1660 and `should_skip_target` at line 2341) but omits `TargetKind::Bench`.  The existing analogous list in `target_spec.rs:255` does include `Bench`, so this looks like a copy-paste oversight.  Result: a `fn main()` in `benches/*.rs` with `harness = false` produces a `RunnableKind::Bin` runnable, but the annotation layer's `should_skip_runnable` skips it because the LSP config passes `binary_target = false` for bench targets.

Closes rust-lang/rust-analyzer#21948.

### DIFF
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -1662,7 +1662,10 @@ pub(crate) fn handle_code_lens(
                 .map(|spec| {
                     matches!(
                         spec.target_kind(),
-                        TargetKind::Bin | TargetKind::Example | TargetKind::Test
+                        TargetKind::Bin
+                            | TargetKind::Example
+                            | TargetKind::Test
+                            | TargetKind::Bench
                     )
                 })
                 .unwrap_or(false),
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22357/5dbf5d55ed0c1dc957728a5a5f5d00c804b9859c..790a0fc475cdf0d2a62423fa271d8a704268f051)*

PR #22362, opened 2026-05-14
feat: add inlayHints.typeHints.renderColons option

MavenRain:
Adds a granular config option requested in rust-lang/rust-analyzer#22305: suppress the leading colon for type hints when `inlayHints.typeHints.location` is `end_of_line`, without disabling colons globally via `inlayHints.renderColons`.

Default is `true`, so current behaviour is preserved unchanged.  Inline-placed type hints are unaffected.

### DIFF
--- a/crates/ide/src/inlay_hints.rs
+++ b/crates/ide/src/inlay_hints.rs
@@ -302,6 +302,7 @@ fn hints(
 #[derive(Clone, Debug)]
 pub struct InlayHintsConfig<'a> {
     pub render_colons: bool,
+    pub type_hints_render_colons: bool,
     pub type_hints: bool,
     pub type_hints_placement: TypeHintsPlacement,
     pub sized_bound: bool,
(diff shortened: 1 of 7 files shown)


MavenRain:
I pushed the redesign.  `inlayHints.typeHints.renderColons` is now a placement-independent override of the global `inlayHints.renderColons` for type hints specifically; when `false`, type-hint colons are suppressed regardless of `typeHints.location`.  Default `true` preserves existing behaviour.

I appreciate the steer.  If you'd still prefer to close this in favour of just using the global, please say so and I'll drop it; no offence taken.

ChayimFriedman2:
@rust-lang/rust-analyzer What is your opinion on this? I personally think we should reject this on the basis that the global `inlayHints.renderColons` is enough and that we have too many settings.

A4-Tacks:
The setting effect is negligible and overlaps with existing settings, so I tend to close it

Veykril:
Yea I don't think this setting pulls its weight either. If anything imo we should just strip the colons from the end of line rendering variant entirely, they do not seem to desirable in the first place if one uses that? 

Thanks for looking into this either way though!

PR #22363, opened 2026-05-14
fix: no lint unsized adt self_ty missing bounded assoc

### LINKED ISSUE #22353: False positive "not all trait items implemented" (E0046) for items with Self: Sized for non-primitive unsized types
**rust-analyzer version**: rust-analyzer version: 0.3.2896-standalone (e266f5cab8 2026-05-10)

**rustc version**: rustc 1.95.0 (59807616e 2026-04-14)

**editor or extension**: VSCode & VSCode Insiders

```rust
#[allow(dead_code)]
fn main() {
    trait State {
        type Repr;

        fn into_repr(self) -> Self::Repr
        where
            Self: Sized;
    }

    struct Stateless([()]);

    impl State for Stateless {
       //^^^^^ not all trait items implemented, missing: `fn into_repr` (E0046)
        type Repr = ();
    }
}
```

It compiles on Rust 1.95.0 Stable. This was previously raised in #19962, but the fix in #19963 only adds filtering [for primitives](https://github.com/rust-lang/rust-analyzer/blob/7c28934677b1e7a1c6ef952422e6ef730540f85f/crates/hir/src/lib.rs#L955-L958). I attempted to implement a proper fix myself, but I’m still new to RA and unfortunately wasn’t able to finish it.

A4-Tacks:
Fixes rust-lang/rust-analyzer#22353

Example
---
```rust
trait Trait {
    fn item() where Self: Sized;
}
impl Trait for Adt {}
struct Adt([i32]);
```

**Before this PR**

```
error: not all trait items implemented, missing: `fn item`
```

**After this PR**

no lint

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -946,14 +946,17 @@ impl Module {
                     .collect();
 
                 if !missing.is_empty() {
+                    let env = ParamEnvAndCrate {
+                        param_env: db.trait_environment(GenericDefId::from(impl_id).into()),
+                        krate: self.id.krate(db),
+                    };
                     let self_ty = db.impl_self_ty(impl_id).instantiate_identity().skip_norm_wip();
-                    let self_ty = structurally_normalize_ty(
-                        &infcx,
-                        self_ty,
-                        db.trait_environment(GenericDefId::from(impl_id).into()),
-                    );
+                    let self_ty = structurally_normalize_ty(&infcx, self_ty, env.param_env);
+                    let tail_ty = struct_tail_raw(db, interner, self_ty, |ty| {
+                        structurally_normalize_ty(&infcx, ty, env.param_env)
+                    });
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (changes requested):
Calculating layout to check sizedness is pretty wasteful and also incorrect since layout can fail for unrelated reasons. You can look at how rustc does that here:

https://github.com/rust-lang/rust/blob/7c3c88f42ad444f4688b865591d84660be4ece2f/compiler/rustc_ty_utils/src/ty.rs#L352

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22363/463396f44de42e1ab2236232978693205a707784..16804e71b04a50429becaeaedacde849c7d9c137)*

ChayimFriedman2 (approved):

PR #22367, opened 2026-05-15
fix: only ref match non-unknown value items

### LINKED ISSUE #22324: Strangly refmatch type
**rust-analyzer version**: 
rust-analyzer 0.4.2892-standalone

**rustc version**: 
rustc 1.97.0-nightly (36ba2c771 2026-04-23)

**editor or extension**: Vim

Related: #22285

`Type::could_unify_with` is a bit strange, cc @ChayimFriedman2

```rust
struct S<T>(T);
fn foo<T>(s: &mut S<T>) {}
fn main() {
    let mut ssss = S(2u32);
    foo($0);
}
```

**Before**:

```
None~ None
S(…)~ S(T)
```

**After**:

```
&mut S(…)~ S(T)
&mut ssss~ S<u32>
&mut S~ S<{unknown}>
&mut Box~ Box<{unknown}, {unknown}>
None~ None
S(…)~ S(T)
```

`Box` should not appear, it may be strange for `S~` to be added with 'ref match' and preferred

**Expect**:

```
&mut S(…)~ S(T)
&mut ssss~ S<u32>
None~ None
S(…)~ S(T)
```

A4-Tacks:
Fixes rust-lang/rust-analyzer#22324

Example
---
```rust
struct S<T>(T);
impl<T> core::ops::Deref for S<T> {
    type Target = T;
}
fn foo<T>(s: &u32) {}
fn main() {
    let ssss = S();
    foo($0);
}
```

**Before this PR**

```text
lc ssss S<{unknown}> [local]
lc &ssss [type+local]
st S S<{unknown}> []
st &S [type]
...
```

**After this PR**

```text
lc ssss S<{unknown}> [local]
st S S<{unknown}> []
...
```

### DIFF
--- a/crates/ide-completion/src/render.rs
+++ b/crates/ide-completion/src/render.rs
@@ -492,7 +492,13 @@ fn render_resolution_path(
             ..CompletionRelevance::default()
         });
 
-        path_ref_match(completion, path_ctx, &ty, &mut item);
+        match resolution {
+            ScopeDef::Local(_)
+            | ScopeDef::ModuleDef(ModuleDef::Const(_) | ModuleDef::Static(_)) => {
+                path_ref_match(completion, path_ctx, &ty, &mut item)
+            }
+            _ => (),
+        }
     };
 
     match resolution {
@@ -720,7 +726,7 @@ fn compute_ref_match(
             || completion_ty.could_unify_with(ctx.db, expected_without_ref))
         && completion_ty
             .autoderef(ctx.db)
-            .any(|ty| ty.could_unify_with(ctx.db, expected_without_ref))
+            .any(|ty| !ty.is_unknown() && ty.could_unify_with(ctx.db, expected_without_ref))
     {
         cov_mark::hit!(suggest_ref);
         let mutability = if expected_type.is_mutable_reference() {
@@ -2639,7 +2645,6 @@ fn go(world: &WorldSnapshot) { go(w$0) }
                 st WorldSnapshot {…} WorldSnapshot { _f: () } []
                 st &WorldSnapshot {…} [type]
                 st WorldSnapshot WorldSnapshot []
-                st &WorldSnapshot [type]
                 fn go(…) fn(&WorldSnapshot) []
             "#]],
         );
@@ -2865,7 +2870,6 @@ fn main() {
                 st S S []
                 st &mut S [type]
                 st S S []
-                st &mut S [type]
                 fn foo(…) fn(&mut S) []
                 fn main() fn() []
             "#]],
@@ -2879,6 +2883,7 @@ fn main() {
     foo(&mut $0);
 }
             "#,
+            // FIXME: There are many `S` here
             expect![[r#"
                 lc s S [type+name+local]
                 st S S [type]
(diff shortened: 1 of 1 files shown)


A4-Tacks:
r? @ChayimFriedman2 This needs to be fixed as soon as possible, otherwise any Deref to generic will be auto imported

Example

```
fn foo(_: &u32) {}
fn main() {
    foo(M$0);
}
```

Completes

```
&MutexGuard~(use std::sync::MutexGuard)
&MutexGuard~(use std::sync::nonpoison::MutexGuard)
&ManuallyDrop~(use std::mem::ManuallyDrop)
&MappedMutexGuard~(use std::sync::MappedMutexGuard)
&MappedMutexGuard~(use std::sync::nonpoison::MappedMutexGuard)
&MappedRwLockReadGuard~(use std::sync::MappedRwLockReadGuard)
&MappedRwLockReadGuard~(use std::sync::nonpoison::MappedRwLockReadGuard)
&MappedRwLockWriteGuard~(use std::sync::MappedRwLockWriteGuard)
&MappedRwLockWriteGuard~(use std::sync::nonpoison::MappedRwLockWriteGuard)
```

crates/ide-completion/src/render.rs
```diff
@@ -492,7 +492,13 @@ fn render_resolution_path(
             ..CompletionRelevance::default()
         });
 
-        path_ref_match(completion, path_ctx, &ty, &mut item);
+        match resolution {
+            ScopeDef::Local(_)
+            | ScopeDef::ModuleDef(ModuleDef::Const(_) | ModuleDef::Static(_)) => {
+                path_ref_match(completion, path_ctx, &ty, &mut item)
```
ChayimFriedman2, line 498: Why do you know filter to only consts/statics/locals?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22367/6eb7777057081649fc30c86843f869e0bcab5d1b..49e1d8806169ba47fe897663e2c2bff431269941)*

A4-Tacks: For struct paths, etc., the assoc fn may not necessarily return Self, and it is suggest to reference noise

And if its method is still ref match, it can be done when completing the method e.g `f(Foo::ne)` -> `f(&Foo::new())`

ChayimFriedman2: But this will also prevent struct literals from adding a ref, e.g. `Foo { ... }` or `Foo(...)`.

A4-Tacks: No, this is not. The test was not broken, and it was rendered by literal instead of path

ChayimFriedman2: Which test?

crates/ide-completion/src/render.rs
```diff
@@ -2639,7 +2645,6 @@ fn go(world: &WorldSnapshot) { go(w$0) }
                 st WorldSnapshot {…} WorldSnapshot { _f: () } []
                 st &WorldSnapshot {…} [type]
                 st WorldSnapshot WorldSnapshot []
-                st &WorldSnapshot [type]
```
A4-Tacks, line 2642: @ChayimFriedman2 Look at this test


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22367/6eb7777057081649fc30c86843f869e0bcab5d1b..49e1d8806169ba47fe897663e2c2bff431269941)*

ChayimFriedman2 (approved):

PR #22372, opened 2026-05-15
method-resolution: emit error for method calls with illegal Sized b…

MavenRain:
Resolves a FIXME at hir-ty/src/method_resolution.rs:151 in
  lookup_method_including_private that previously allowed calls
  like `x.cant_call()` on a `&dyn Foo` receiver to type-check
  without diagnostic when `cant_call(&self) where Self: Sized`,
  even though the `Self: Sized` predicate cannot be satisfied
  by the trait object.

  The existing predicates_require_illegal_sized_bound check in
  confirm.rs already correctly identifies this case (it iterates
  elaborated predicates and looks for `Self: Sized` clauses with
  a `TyKind::Dynamic` self type).  Adds an
  InferenceDiagnostic::MethodCallIllegalSizedBound variant,
  plumbs the cooked diagnostic through hir, and renders it as
  RustcHardError E0277.

  Tests: trait-object call with Sized-bounded method errors;
  trait-object call to dyn-safe method does not; concrete-type
  call to a Sized-bounded method does not.

  Part of rust-lang/rust-analyzer#22140.

### DIFF
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -438,6 +438,10 @@ pub enum InferenceDiagnostic {
         #[type_visitable(ignore)]
         def: GenericDefId,
     },
+    MethodCallIllegalSizedBound {
+        #[type_visitable(ignore)]
+        call_expr: ExprId,
+    },
     MethodCallIncorrectGenericsOrder {
         #[type_visitable(ignore)]
         expr: ExprId,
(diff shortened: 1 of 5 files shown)


ChayimFriedman2 (approved):
Interesting that this is not a regular trait bound check.

Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22372/6eb7777057081649fc30c86843f869e0bcab5d1b..e83fb4c4c2b8ef1bd9ee89cdc847bb7dc69caa41)*

PR #22378, opened 2026-05-16
feat(diagnostics): add handler for E0040

ada4a:

cc @WaterWhisperer

### DIFF
--- a/crates/hir-ty/src/infer/path.rs
+++ b/crates/hir-ty/src/infer/path.rs
@@ -11,7 +11,7 @@ use rustc_type_ir::inherent::{SliceLike, Ty as _};
 use stdx::never;
 
 use crate::{
-    InferenceDiagnostic, Span, ValueTyDefId,
+    ExplicitDropMethodUseKind, InferenceDiagnostic, Span, ValueTyDefId,
     infer::{
         InferenceTyLoweringVarsCtx, diagnostics::InferenceTyLoweringContext as TyLoweringContext,
     },
@@ -32,6 +32,14 @@ impl<'db> InferenceContext<'_, 'db> {
         id: ExprOrPatId,
     ) -> Option<(ValueNs, Ty<'db>)> {
         let (value, self_subst) = self.resolve_value_path_inner(path, id, false)?;
+
+        if let ValueNs::FunctionId(f) = value
+            && self.lang_items.Drop_drop.is_some_and(|drop_fn| drop_fn == f)
+        {
+            self.push_diagnostic(InferenceDiagnostic::ExplicitDropMethodUse {
+                kind: ExplicitDropMethodUseKind::Path(id),
+            });
+        }
 
         let (value_def, generic_def, substs) =
             match self.resolve_value_path(path, id, value, self_subst)? {
--- a/crates/ide-diagnostics/src/handlers/explicit_drop_method_use.rs
+++ b/crates/ide-diagnostics/src/handlers/explicit_drop_method_use.rs
@@ -1,0 +1,426 @@
+use either::Either;
+use hir::InFile;
+use ide_db::assists::Assist;
+use ide_db::source_change::{SourceChange, SourceChangeBuilder};
+use ide_db::text_edit::TextEdit;
+use itertools::Itertools;
+use syntax::{
+    AstNode, AstPtr,
+    ast::{self, HasArgList},
+};
+
+use crate::{Diagnostic, DiagnosticCode, DiagnosticsContext, adjusted_display_range, fix};
+
+// Diagnostic: explicit-drop-method-use
+//
+// This diagnostic is triggered when the `Drop::drop` method is called (or named) explicitly.
+pub(crate) fn explicit_drop_method_use(
+    ctx: &DiagnosticsContext<'_, '_>,
+    d: &hir::ExplicitDropMethodUse,
+) -> Diagnostic {
+    match d.expr_or_path {
+        Either::Left(expr) => {
+            let display_range = adjusted_display_range(ctx, expr, &|node| {
+                Some(node.name_ref()?.syntax().text_range())
+            });
+            Diagnostic::new(
+                DiagnosticCode::RustcHardError("E0040"),
+                "explicit use of destructor method",
+                display_range,
+            )
+            .stable()
+            .with_main_node(expr.map(Into::into))
+            .with_fixes(fix_method_call(ctx, expr))
+        }
+        Either::Right(path) => Diagnostic::new_with_syntax_node_ptr(
+            ctx,
+            DiagnosticCode::RustcHardError("E0040"),
+            "explicit use of destructor method",
+            path.map(Into::into),
+        )
+        .stable()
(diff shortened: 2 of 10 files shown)


crates/ide-diagnostics/src/handlers/explicit_drop_method_use.rs
```diff
@@ -0,0 +1,377 @@
...
+        // `path` could be the `Foo::drop` in `let d = Foo::drop;`
+        // -- replace the path with `drop`
+        let range = path.syntax().text_range();
+
+        let mut builder = SourceChangeBuilder::new(file_id.original_file(db).file_id(db));
+        let editor = builder.make_editor(path.syntax());
+        let make = editor.make();
+        let new_path = make.path_from_text("drop");
+        builder.replace_ast(path, new_path);
+        let source_change = builder.finish();
```
ada4a: This one should be pretty easy to replace with a text edit, as it just replaces one path with another -- should I go on and do that?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22378/1a68212c5683555ad80f0eab71db9715c6d52145..276dcd5a312982537ccc67f365cc8ddb8feef8dd)*

crates/hir-ty/src/infer/path.rs
```diff
@@ -33,6 +33,15 @@ impl<'db> InferenceContext<'_, 'db> {
     ) -> Option<(ValueNs, Ty<'db>)> {
         let (value, self_subst) = self.resolve_value_path_inner(path, id, false)?;
 
+        if let ValueNs::FunctionId(f) = value
+            && let ItemContainerId::TraitId(trait_) = f.lookup(self.db).container
+            && self.lang_items.Drop.is_some_and(|drop_trait| drop_trait == trait_)
```
ChayimFriedman2: In method_resolution/confirm.rs we now instead compare to `Drop_drop`, do that here too as it's faster.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22378/1dd030d6d8a0af508ce76093fcae3ee0d8b64fe5..61fae9ced8e7f7ae7be007edb258548ddb7532e2)*

crates/ide-diagnostics/src/handlers/explicit_drop_method_use.rs
```diff
@@ -0,0 +1,377 @@
...
+    mcall_ptr: InFile<AstPtr<ast::MethodCallExpr>>,
+) -> Option<Vec<Assist>> {
+    let db = ctx.db();
+
+    let file_id = mcall_ptr.file_id;
+    let mcall = mcall_ptr.to_node(db);
+    // TODO: is this more correct?
+    // let range = InFile::new(file_id, mcall.syntax().text_range())
+    //     .original_node_file_range_rooted_opt(db)?
+    //     .range;
```
ChayimFriedman2: We do need to handle macros and therefore the current code isn't correct, but if we handle them we no longer have an AST (only token trees) and can no longer use `SyntaxEditor`. So you have two options: either just disable the fix if `file_id.is_macro_file()`, or upmap the range of the receiver and the range of the whole call, delete everything outside the receiver and replace it with `drop(` ... `)`, using range edits only.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22378/1dd030d6d8a0af508ce76093fcae3ee0d8b64fe5..61fae9ced8e7f7ae7be007edb258548ddb7532e2)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/infer/path.rs
ada4a: Completely missed that `Drop_drop` was a lang item itself, thank you for pointing it out!

crates/ide-diagnostics/src/handlers/explicit_drop_method_use.rs
ada4a: I think I'll leave the macro things as a future improvement.. I'll add a TODO for now

crates/ide-diagnostics/src/handlers/explicit_drop_method_use.rs
```diff
@@ -0,0 +1,424 @@
...
+    } else {
+        // `path` could be the `Foo::drop` in `let d = Foo::drop;`
+        // -- replace the path with `drop`
+
+        let range = InFile::new(file_id, path.syntax().text_range())
+            .original_node_file_range_rooted_opt(db)?;
+
+        let edit = TextEdit::replace(range.range, "drop".to_owned());
+        let source_change = SourceChange::from_text_edit(range.file_id.file_id(db), edit);
+        Some(vec![fix("use-drop-function", "Use `drop` function", source_change, range.range)])
```
ada4a, line 124: So I thought I'd turn this one into a text edit because (a) I was already thinking to, and (b) that would allow us to handle macros correctly. However, this doesn't seem to work, as `original_node_file_range_rooted_opt` returns `None` when inside a macro. If you see an obvious problem here, do let me know, otherwise I'd just ~~switch to bailing out here like everywhere else~~ leave this as is and accept the missing fixes in macros for now..

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22378/ace47025ddfc7fec399e11594318e64443c0e1c2..3276d9471a0e070a94ba38e95a7f711f3e54c690)*

ChayimFriedman2: If the code is written in the macro definition, it will return `None` and you should not change the macro definition anyway. When we talk about macro support, we mean code inside macro call.

crates/ide-diagnostics/src/handlers/explicit_drop_method_use.rs
```diff
@@ -0,0 +1,426 @@
...
+
+fn main(mut a: A) {
+    let d = A::drop;
+         // ^^^^^^^ 💡 error: explicit use of destructor method
+    d(&mut a);
+}
+"#,
+            // Because of the error, the code isn't analyzed further (?), and so `d` is warned on as unused.
+            // Arguably a bug in r-a (rustc doesn't emit a warning in this case)
+            // FIXME: remove this once r-a no longer warns
```
ChayimFriedman2, line 306: Our borrowck errors (unused stuff, mutability errors) are highly incorrect and have false positives.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22378/5a5a4b568dfb979715cb9ef298de38d4c87d2cce..00cf4fa63058dd85972b95c13eb5f7240fab184d)*

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22378/5a5a4b568dfb979715cb9ef298de38d4c87d2cce..00cf4fa63058dd85972b95c13eb5f7240fab184d)*

PR #22391, opened 2026-05-17
internal: Ignore workspace/diagnostic/refresh in slow tests

lnicola:
Not sure about this, but might fix https://github.com/rust-lang/rust-analyzer/pull/22389#issuecomment-4470210564.

### DIFF
--- a/crates/rust-analyzer/tests/slow-tests/flycheck.rs
+++ b/crates/rust-analyzer/tests/slow-tests/flycheck.rs
@@ -76,7 +76,6 @@ fn main() {
 }
 
 #[test]
-#[ignore = "this test tends to stuck, FIXME: investigate that"]
 fn test_flycheck_diagnostic_with_override_command() {
     if skip_slow_tests() {
         return;
@@ -113,7 +112,6 @@ fn main() {}
 }
 
 #[test]
-#[ignore = "this test tends to stuck, FIXME: investigate that"]
 fn test_flycheck_diagnostics_with_override_command_cleared_after_fix() {
     if skip_slow_tests() {
         return;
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
Hmm, the timeout for macOS is higher (https://github.com/rust-lang/rust-analyzer/pull/5555), this might be why it always ever failed on macOS...

ChayimFriedman2 (approved):
Seems convincing enough to at least try.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22391/5b101aafa1f8d6d9fbd37909e6eadf11bf625e5b..c328a923a37abc6328720d341711bf032794057f)*

ChayimFriedman2:
Wait, this alone won't save the timeout, just the panic, but you can try to also fix this method?

https://github.com/rust-lang/rust-analyzer/blob/ace47025ddfc7fec399e11594318e64443c0e1c2/crates/rust-analyzer/tests/slow-tests/support.rs#L514-L521

So that it won't be `Ok(msg.ok())`, but `msg.map_err(|_| Timeout)`? This will make sure we error when the channel is closed, instead of hang.

lnicola:
Hmm, that means we never return `Ok(None)`, so we can't report that the channel got closed, which then affects https://github.com/rust-lang/rust-analyzer/blob/ace47025ddfc7fec399e11594318e64443c0e1c2/crates/rust-analyzer/tests/slow-tests/support.rs#L370 and two more places.

ChayimFriedman2:
Oh, I missed that we already handle closed channels.

Well, let's merge this and see.

PR #22392, opened 2026-05-17
fix: Do not autoref in method probe in path mode

ChayimFriedman2:
Fixes the first problem in https://github.com/rust-lang/rust-analyzer/issues/22387.

See also [#t-types > Method resolution probe autorefs on path](https://rust-lang.zulipchat.com/#narrow/channel/144729-t-types/topic/Method.20resolution.20probe.20autorefs.20on.20path/with/595607623).

### DIFF
--- a/crates/hir-ty/src/tests/traits.rs
+++ b/crates/hir-ty/src/tests/traits.rs
@@ -3027,13 +3027,13 @@ fn test() {
             140..146 'IsCopy': IsCopy
             140..153 'IsCopy.test()': bool
             159..166 'NotCopy': NotCopy
-            159..173 'NotCopy.test()': bool
+            159..173 'NotCopy.test()': {unknown}
             179..195 '(IsCop...sCopy)': (IsCopy, IsCopy)
             179..202 '(IsCop...test()': bool
             180..186 'IsCopy': IsCopy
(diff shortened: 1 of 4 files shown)


crates/hir-ty/src/tests/traits.rs
```diff
@@ -3027,13 +3027,13 @@ fn test() {
             140..146 'IsCopy': IsCopy
             140..153 'IsCopy.test()': bool
             159..166 'NotCopy': NotCopy
-            159..173 'NotCopy.test()': bool
+            159..173 'NotCopy.test()': {unknown}
```
ChayimFriedman2, line 3030: The other test regressions are in path mode (where the trait doesn't hold) so expected. The ones in this file is in method call mode, so less expected. The reason for it is that if we fail to resolve with method call mode we retry with path mode.

rustc fails to resolve these methods as well.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22392/5b101aafa1f8d6d9fbd37909e6eadf11bf625e5b..dbb0c46221c1e1f04d3adf291868ca45161f6e92)*

Veykril (approved):

PR #22393, opened 2026-05-17
feat: add diagnostic for E0422

WeiTheShinobi:
Related to https://github.com/rust-lang/rust-analyzer/issues/22140


crates/hir-ty/src/infer/expr.rs
```diff
@@ -964,7 +964,7 @@ impl<'db> InferenceContext<'_, 'db> {
     ) -> Ty<'db> {
         // Find the relevant variant
         let (adt_ty, Some(variant)) = self.resolve_variant(expr.into(), path, false) else {
-            // FIXME: Emit an error.
+            self.push_diagnostic(InferenceDiagnostic::UnresolvedRecordExpr { expr });
```
ChayimFriedman2: I know there was a FIXME, but I think the error should go inside `resolve_variant()`, which will also fix it for patterns.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22393/5a5a4b568dfb979715cb9ef298de38d4c87d2cce..3c921e1f07b0390254bdc7734dd90dbf1024298e)*

crates/ide-diagnostics/src/handlers/unresolved_record_expr.rs
```diff
@@ -0,0 +1,45 @@
+use either::Either;
+use syntax::{AstNode, ast::Expr};
```
ChayimFriedman2: By convention we don't `use` specific ast nodes, instead we refer to them as `ast::Node`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22393/5a5a4b568dfb979715cb9ef298de38d4c87d2cce..3c921e1f07b0390254bdc7734dd90dbf1024298e)*

crates/ide-diagnostics/src/handlers/unresolved_record_expr.rs
```diff
@@ -0,0 +1,45 @@
...
+pub(crate) fn unresolved_record_expr(
+    ctx: &DiagnosticsContext<'_, '_>,
+    d: &hir::UnresolvedRecordExpr,
+) -> Diagnostic {
+    Diagnostic::new(
+        DiagnosticCode::RustcHardError("E0422"),
+        "cannot find struct, variant or union type in this scope".to_owned(),
+        crate::adjusted_display_range(ctx, d.expr, &|expr| match expr {
+            Either::Left(Expr::RecordExpr(it)) => it.path().map(|p| p.syntax().text_range()),
+            _ => None,
```
ChayimFriedman2: If you emit the error in `resolve_variant()`, you also need to account for `ast::Pat::{RecordPat,TupleStructPat}`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22393/5a5a4b568dfb979715cb9ef298de38d4c87d2cce..3c921e1f07b0390254bdc7734dd90dbf1024298e)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/infer.rs
```diff
@@ -2293,6 +2278,8 @@ impl<'body, 'db> InferenceContext<'body, 'db> {
         let interner = DbInterner::conjure();
         let (resolution, unresolved) = if value_ns {
             let Some(res) = path_ctx.resolve_path_in_value_ns(HygieneId::ROOT) else {
+                drop(ctx);
+                self.push_diagnostic(InferenceDiagnostic::UnresolvedVariant { id: node });
```
ChayimFriedman2, line 2282: Sorry about this, but IMO this is still not the right place. We should emit an error in `resolve_path_in_X()`, and I don't know if we should do it yet.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22393/2ba9e2e191b05ddef383d3b702b2aa4695d12d75..783426a1580591a6a519274f587b0a241bb92a1a)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/infer.rs
ChayimFriedman2: Also there are more cases in this method after resolution.

WeiTheShinobi: > Sorry about this, but IMO this is still not the right place. We should emit an error in `resolve_path_in_X()`, and I don't know if we should do it yet.
> 
> _[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22393/2ba9e2e191b05ddef383d3b702b2aa4695d12d75..783426a1580591a6a519274f587b0a241bb92a1a)_

I found that if I emit diagnostics in `resolve_path_in_value_ns` and `resolve_path_in_type_ns`, different call sites need different error code (for example, another `resolve_path_in_type_ns` call site need `E0425` rather than `E0422`). I feel like there is no proper place for now.

ChayimFriedman2: First, the error codes are not set in stone. The code organization is more important. Second, we can pass this information from the outside.

But as I said, I am not sure we should do this yet.

PR #22394, opened 2026-05-17
feat: add `simplify_match` assist

### LINKED ISSUE #10443: Feature request: simplify match expression
When I'm refactoring and inlining stuff all over the place, I often end up with cases like this:
```rust
match Some(foo()) {
  None => { do_none() }
  Some(0) => { do_0() }
  Some(1) => { do_1() }
  _ => { do_wild() }
}
```
I would love an assist that simplifies this to
```rust
match foo() {
  0 => { do_0() }
  1 => { do_1() }
  _ => { do_wild() }
}
```
It would also make sense for `if let`s and even plain `if`s.

I realize this is unlikely to be actionable until match checking is librarified, but that's not too far in the future hopefully.

Albab-Hasan:
when the scrutinee of a `match` is a tuple struct enum variant call the outer constructor wrapper can be stripped:

 - arms matching the same constructor are simplified by removing the outer pattern wrapper (`Some(n)` → `n`, `Pair::Both(a, b)` → `(a, b)`)
- arms for other constructors of the same enum are removed as unreachable dead code
- wildcard and binding arms are unchanged

the assist bails on guards, OR patterns, unit variant scrutinees and any arm pattern that doesnt resolve to the same enum.

single segment variant paths like `None` are syntactically parsed as `IdentPat` not `PathPat`. `resolve_bind_pat_to_const` is used to differentiate them from variable bindings.

closes https://github.com/rust-lang/rust-analyzer/issues/10443

### DIFF
--- a/crates/ide-assists/src/handlers/simplify_match.rs
+++ b/crates/ide-assists/src/handlers/simplify_match.rs
@@ -1,0 +1,379 @@
+use hir;
+use syntax::{
+    AstNode,
+    ast::{self, HasArgList, edit::AstNodeEdit},
+};
+
+use crate::{AssistContext, AssistId, Assists};
+
+// Assist: simplify_match
+//
+// Simplifies a `match` expression where the scrutinee is a constructor call.
+// Unreachable arms for other constructors are removed, and arms matching the
+// constructor have the outer wrapper stripped.
+//
+// ```
+// # //- minicore: option
+// fn f(x: i32) -> i32 {
+//     $0match Some(x + 1) {
+//         Some(n) => n,
+//         None => 0,
+//     }
+// }
+// ```
+// ->
+// ```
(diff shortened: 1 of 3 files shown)


ChayimFriedman2:
Note that we have a feature freeze on assists so this PR might get closed.

Albab-Hasan:
@ChayimFriedman2 my bad, forgot abt that

Nadrieril (approved):
Simple enough, looks good as far as I can tell. Turns out this doesn't make use of match exhaustiveness information, which is nice.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22394/5a5a4b568dfb979715cb9ef298de38d4c87d2cce..cabd0d28a2bf4afbfa1553401dbd26947bfe1ac7)*

PR #22398, opened 2026-05-18
fix: parse field expr keyword field to field-expr node

A4-Tacks:
This makes the completion more coherent and helpful for climbing trees

Example
---
```rust
const _=x.ref
```

**Before this PR**

```
CONST
  CONST_KW "const"
  WHITESPACE " "
  UNDERSCORE "_"
  EQ "="
  FIELD_EXPR
    PATH_EXPR
      PATH
        PATH_SEGMENT
          NAME_REF
            IDENT "x"
    DOT "."
ERROR
  REF_KW "ref"
ERROR
  SEMICOLON ";"
```

**After this PR**

```rust
CONST
  CONST_KW "const"
  WHITESPACE " "
  UNDERSCORE "_"
  EQ "="
  FIELD_EXPR
    PATH_EXPR
      PATH
        PATH_SEGMENT
          NAME_REF
            IDENT "x"
    DOT "."
    ERROR
      REF_KW "ref"
  SEMICOLON ";"
```

---
```rust
struct S { indent: u8 }
fn foo(_: &mut u8) {}
fn main() {
    let mut s = S { indent: 2 };
    foo(s.in$0)
}
```

**Before this PR**

`ty: (), name: ?`

```
indent
```

**After this PR**

`ty: &mut u8, name: ?`

```
&mut indent
indent
```

### DIFF
--- a/crates/ide-completion/src/context/tests.rs
+++ b/crates/ide-completion/src/context/tests.rs
@@ -810,3 +810,39 @@ fn foo() {
         expect![[r#"ty: bool, name: ?"#]],
     );
 }
+
+#[test]
+fn expected_type_some_incomplete_ast() {
+    check_expected_type_and_name(
+        r#"
+fn foo(x: u32) {
+    foo(y.$0)
+}
+"#,
+        expect!["ty: u32, name: x"],
+    );
+    check_expected_type_and_name(
+        r#"
+fn foo(x: u32) {
+    foo(y.in$0)
+}
+"#,
+        expect!["ty: u32, name: x"],
+    );
+    check_expected_type_and_name(
+        r#"
+fn foo(x: u32) {
+    foo(y::$0)
+}
+"#,
+        expect!["ty: u32, name: x"],
+    );
+    check_expected_type_and_name(
+        r#"
+fn foo(x: u32) {
+    foo(crate::$0)
+}
+"#,
+        expect!["ty: u32, name: x"],
+    );
+}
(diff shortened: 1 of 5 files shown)


ChayimFriedman2:
I feel this is a wrong step to take. The reason completion fails is a bug in completion (specifically in the expansion phase, which speculative execution will fix). We shouldn't touch parsing to fix that.

A4-Tacks:
I added a bug fix example

ChayimFriedman2:
I'm aware of this bug and that this fixes it, that's why I said:

> The reason completion fails is a bug in completion (specifically in the expansion phase, which speculative execution will fix). We shouldn't touch parsing to fix that.

A4-Tacks:
But some steps are on the original tree, not speculative tree, like expected type and name

And this does make incomplete parsing more reasonable, it won't get worse

ChayimFriedman2:
> But some steps are on the original tree, not speculative tree, like expected type and name

I don't understand what you mean by that. What I'm saying is that once we have speculative execution, this bug will be fixed without any change to the parser.

> And this does make incomplete parsing more reasonable, it won't get worse

Changing parsing to accept more code has many implications, and I'm opposed to that without a very good reason, especially not to aid a specific IDE feature such as completion.

A4-Tacks:
Okay, I've come up with a worse case:

```rust
const _={
    x.
    let y = 2;
}
```

PR #22399, opened 2026-05-18
fix: complete ref_match in macro

A4-Tacks:
Why isn't `original_range_opt()` used in rust-lang/rust-analyzer#12800 rust-lang/rust-analyzer#12830 rust-lang/rust-analyzer#13092?

Example
---
```rust
macro_rules! id { ($($t:tt)*) => ($($t)*); }
fn foo(data: &i32) {}
fn main() {
    let indent = 2i32;
    id!(foo(i$0))
}
```

**Before this PR**

```
indent
```

**After this PR**

```
&indent
indent
```

### DIFF
--- a/crates/ide-completion/src/render.rs
+++ b/crates/ide-completion/src/render.rs
@@ -181,9 +181,9 @@ pub(crate) fn render_field(
 
         if !expected_fn_type
             && let Some(receiver) = &dot_access.receiver
-            && let Some(receiver) = ctx.completion.sema.original_ast_node(receiver.clone())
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (approved):
I'm not sure if we had `original_range_opt()` back then.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22399/ce81cf65580fb7a52d9f9a896092746356f39320..7def0e8612d77180a100a5e27eebcbbdc1df5337)*

PR #22403, opened 2026-05-19
minor: offer on r_curly for generate_impl

A4-Tacks:
Example
---
```rust
struct Foo {
    x: i32
}$0
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
struct Foo {
    x: i32
}

impl Foo {$0}
```

### DIFF
--- a/crates/ide-assists/src/handlers/generate_impl.rs
+++ b/crates/ide-assists/src/handlers/generate_impl.rs
@@ -50,7 +50,7 @@ pub(crate) fn generate_impl(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -> O
     let name = nominal.name()?;
     let target = nominal.syntax().text_range();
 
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (changes requested):
I support changing the assist to only trigger on the struct name instead.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22403/cdfe408d4b436e806ff525cb3e67588a6a009ed1..fd98f0cb2f7dae12af46379d8c7f4e2e43863246)*

A4-Tacks:
Triggering at the tail is very meaningful, especially for longer structures

ChayimFriedman2:
We already have too many assists. Shrinking their range, while making them less discoverable, also makes them less distracting.

A4-Tacks:
Just triggered in braces, it's also good to trigger in name+braces

PR #22404, opened 2026-05-19
feat: add diagnostic for E0608

WaterWhisperer:

### DIFF
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -730,13 +730,10 @@ impl<'db> InferenceContext<'_, 'db> {
                         trait_element_ty
                     }
                     None => {
-                        if self.lang_items.Index.is_some() && self.lang_items.Index_index.is_some()
-                        {
-                            self.push_diagnostic(InferenceDiagnostic::CannotIndexInto {
-                                expr: tgt_expr,
-                                found: base_t.store(),
-                            });
-                        }
+                        self.push_diagnostic(InferenceDiagnostic::CannotIndexInto {
+                            expr: tgt_expr,
+                            found: base_t.store(),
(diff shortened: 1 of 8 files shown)


crates/hir-ty/src/infer/expr.rs
```diff
@@ -729,8 +729,16 @@ impl<'db> InferenceContext<'_, 'db> {
                         self.table.select_obligations_where_possible();
                         trait_element_ty
                     }
-                    // FIXME: Report an error.
-                    None => self.types.types.error,
+                    None => {
+                        if self.lang_items.Index.is_some() && self.lang_items.Index_index.is_some()
```
ChayimFriedman2: I see no reason for this `if`. In real world the lang items will always be defined, and tests aren't worth it (especially given it will only manifest in diagnostic tests).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22404/cdfe408d4b436e806ff525cb3e67588a6a009ed1..767a04b84b85f91034b5a6f4106a39e231787ebd)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/infer/expr.rs
WaterWhisperer: Yes, I added it for tests. I'll remove it

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22404/cdfe408d4b436e806ff525cb3e67588a6a009ed1..f798d0b8f16aa0f29c954370074f8ee3e52fd7d8)*

PR #22405, opened 2026-05-19
hir-ty: walk container exprs for unused_must_use

MavenRain:
ExprValidator::check_unused_must_use previously matched only Expr::Call and Expr::MethodCall directly on the expression of a Statement::Expr with semicolon.  This left several stmt-with-semi cases unwarned: blocks whose tail is a #[must_use] call (`{ produces() };`), unsafe-blocks (`unsafe { produces() };`), if/else arms, match arms, and const blocks.

Refactors check_unused_must_use into a recursive collector that walks into Expr::Block { tail, .. }, Expr::Unsafe { tail, .. }, Expr::If { then_branch, else_branch, .. }, Expr::Match { arms, .. }, and Expr::Const(inner) before applying the existing leaf check (callee or method #[must_use] attribute, or #[must_use] ADT return type). The diagnostic stays at the leaf call so the span points to the actual must_use producer rather than the wrapping block.

Adds ide-diagnostics tests covering each new container case (block tail, unsafe block tail, nested blocks, if/else branches, match arms, const block, #[must_use] ADT through a block) and lock-in tests confirming that the existing RustcLint plumbing already honors #[allow(unused_must_use)], #[allow(unused)], and #[deny(unused_must_use)] attribute resolution.

Follow-up to rust-lang/rust-analyzer#22239.

### DIFF
--- a/crates/hir-ty/src/diagnostics/expr.rs
+++ b/crates/hir-ty/src/diagnostics/expr.rs
@@ -333,17 +333,20 @@ impl<'db> ExprValidator<'db> {
         let pattern_arena = Arena::new();
         let cx = MatchCheckCtx::new(self.owner.module(self.db()), &self.infcx, self.env);
         for stmt in &**statements {
-            let diag = match *stmt {
+            match *stmt {
                 Statement::Expr { expr: stmt_expr, has_semi: true } if self.validate_lints => {
-                    self.check_unused_must_use(stmt_expr)
+                    let mut diags = Vec::new();
+                    self.check_unused_must_use(stmt_expr, &mut diags);
+                    self.diagnostics.extend(diags);
                 }
                 Statement::Let { pat, initializer, else_branch: None, .. } => {
-                    self.check_non_exhaustive_let(&cx, &pattern_arena, pat, initializer)
+                    if let Some(diag) =
+                        self.check_non_exhaustive_let(&cx, &pattern_arena, pat, initializer)
+                    {
+                        self.diagnostics.push(diag);
+                    }
                 }
-                _ => None,
-            };
-            if let Some(diag) = diag {
-                self.diagnostics.push(diag);
+                _ => {}
             }
         }
     }
@@ -415,7 +418,37 @@ impl<'db> ExprValidator<'db> {
         pattern
     }
 
-    fn check_unused_must_use(&self, expr: ExprId) -> Option<BodyValidationDiagnostic<'db>> {
+    fn check_unused_must_use(
+        &self,
+        mut expr: ExprId,
+        acc: &mut Vec<BodyValidationDiagnostic<'db>>,
+    ) {
+        // Walk through container expressions so that the value-producing leaf is
+        // checked even when wrapped in a block, `unsafe { .. }`, `if`/`match`, or
+        // a `const { .. }` block.  Single-tail chains are followed by reassigning
+        // `expr`; branching containers (`if`/`match`) recurse on each arm.
+        loop {
+            match &self.body[expr] {
+                Expr::Block { tail: Some(tail), .. }
+                | Expr::Unsafe { tail: Some(tail), .. }
+                | Expr::Const(tail) => expr = *tail,
+                Expr::If { then_branch, else_branch, .. } => {
+                    self.check_unused_must_use(*then_branch, acc);
+                    if let Some(else_branch) = else_branch {
+                        self.check_unused_must_use(*else_branch, acc);
+                    }
+                    return;
+                }
+                Expr::Match { arms, .. } => {
+                    for arm in arms.iter() {
+                        self.check_unused_must_use(arm.expr, acc);
+                    }
+                    return;
+                }
+                _ => break,
+            }
+        }
+
         let fn_def = match &self.body[expr] {
             Expr::Call { callee, .. } => {
                 let callee_ty = self.infer.expr_ty(*callee);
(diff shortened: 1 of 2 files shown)


crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -415,7 +416,45 @@ impl<'db> ExprValidator<'db> {
...
+    fn check_unused_must_use(&self, expr: ExprId) -> Vec<BodyValidationDiagnostic<'db>> {
+        let mut diags = Vec::new();
+        self.collect_unused_must_use(expr, &mut diags);
+        diags
+    }
+
+    fn collect_unused_must_use(
+        &self,
+        expr: ExprId,
+        diags: &mut Vec<BodyValidationDiagnostic<'db>>,
```
ChayimFriedman2: You can just push to `self.diagnostics`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22405/cdfe408d4b436e806ff525cb3e67588a6a009ed1..68efe3e9d491f74b3172360cfa369b0a52e70738)*

crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -415,7 +416,45 @@ impl<'db> ExprValidator<'db> {
...
+    ) {
+        // Walk through container expressions so that the value-producing leaf is
+        // checked even when wrapped in a block, `unsafe { .. }`, `if`/`match`, or
+        // a `const { .. }` block.
+        match &self.body[expr] {
```
ChayimFriedman2: Or better, there is no need for a recursion here, given a block has up to one tail expression. Instead loop until you reach it.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22405/cdfe408d4b436e806ff525cb3e67588a6a009ed1..68efe3e9d491f74b3172360cfa369b0a52e70738)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/diagnostics/expr.rs
```diff
@@ -330,19 +330,20 @@ impl<'db> ExprValidator<'db> {
         let (Expr::Block { statements, .. } | Expr::Unsafe { statements, .. }) = expr else {
             return;
         };
+        if self.validate_lints {
```
ChayimFriedman2: Why did you separate to a new loop? And why this `if`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22405/7f916ab1b1f669cec017960c2d91a9a87b4b7bae..ffe5641d332363c7a361e4ca9038834c07c7f791)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/diagnostics/expr.rs
MavenRain: Good catch . . . the split was a workaround: making `check_unused_must_use` take &mut self so it could push to self.diagnostics directly meant it could no longer run in the same loop as the MatchCheckCtx (which borrows &self.infcx).  I've reverted it to &self with a small accumulator the caller drains, so both checks share the single loop again and `validate_lints` is back on the match arm.

crates/ide-diagnostics/src/handlers/unused_must_use.rs
```diff
@@ -126,6 +126,198 @@ fn main() {
...
+        produces()
+      //^^^^^^^^^^ warn: unused return value that must be used
+    };
+}
+"#,
+        );
+    }
+
+    #[test]
+    fn allow_attribute_on_fn_silences_lint() {
```
ChayimFriedman2: We already have tests for lint attributes, we don't need to test them for every lint. Please remove those tests.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22405/7f916ab1b1f669cec017960c2d91a9a87b4b7bae..6a6de48ee1a71ce630a8562d09b5e55f00ef30d8)*

ChayimFriedman2 (changes requested):

MavenRain:
Done: removed the four allow_*/deny_* lock-in tests;  kept only the container-expression behavior tests.  Agreed the generic lint-attribute handling is already covered and doesn't need re-testing per lint.

ChayimFriedman2 (approved):
Please squash.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22405/7f916ab1b1f669cec017960c2d91a9a87b4b7bae..2632230adae37992795e7d072de478e352c28f52)*

MavenRain:
@ChayimFriedman2 CI is red solely from the cancelled powerpc + wasm32); all 13 real jobs are green; could you re-run the cross job or merge past it (contributors can't re-trigger upstream CI)?

MavenRain:
Thanks again for the review @ChayimFriedman2, and the squash is in (head is `c859808`).

One heads-up on the red CI: the failing rollup is purely an infra flake, not anything in the diff.  The `Rust Cross (x86_64-unknown-linux-musl)` job died ~17s in with curl `[16] Error in the HTTP2 framing layer` while downloading the toolchain, and that cancelled the `powerpc-unknown-linux-gnu` and `wasm32-unknown-unknown` cross jobs.  All 13 substantive jobs (rustfmt, clippy, miri, the ubuntu/windows/macos test matrices, analysis-stats, coverage, rustdoc, proc-macro-srv) are green.

Since I can't re-trigger upstream CI as a contributor, would you be able to re-run the cross job or merge past it when you get a moment?  I'd be happy to rebase for a fresh run if you'd rather.  Thanks again!

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22405/d1fc9cfbbed8945d70ef0ca8313fb978fcd2542c..c859808cf526f0ebc7e794b06fd49ac62b1c3c07)*

PR #22409, opened 2026-05-20
fix: no suggest ref match when expected generic ref

A4-Tacks:
Example
---
```rust
fn foo<T>(s: &T) {}
fn main() {
    let ssss = &mut 2i32;
    foo($0);
}
```

**Before this PR**

```text
lc ssss &mut i32 [type_could_unify+local]
lc &ssss [type+local]
md core::  []
fn foo(…) fn(&T) []
fn &foo(…) [type]
fn main() fn() []
fn &main() [type]
```

**After this PR**

```text
lc ssss &mut i32 [type_could_unify+local]
md core::  []
fn foo(…) fn(&T) []
fn &foo(…) [type]
fn main() fn() []
fn &main() [type]
```

### DIFF
--- a/crates/ide-completion/src/render.rs
+++ b/crates/ide-completion/src/render.rs
@@ -718,12 +718,13 @@ fn compute_ref_match(
     ctx: &CompletionContext<'_, '_>,
     completion_ty: &hir::Type<'_>,
 ) -> Option<CompletionItemRefMode> {
+    if compute_type_match(ctx, completion_ty).is_some() {
+        return None;
+    }
     let expected_type = ctx.expected_type.as_ref()?;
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
Worth thinking whether we should check coercion instead, which will also fix that.

Also, we might want to omit the `&foo()` and `&main()` suggestions, given that they return `()` (and in general any `&()` suggestion)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22409/987ea33645ab1c709b1df6823038abcb2fe8973e..3c7db605fed7a4431e47a07e8c34a2bb74c12352)*

PR #22410, opened 2026-05-20
feat: emit async modifier for functions returning impl Future

### LINKED ISSUE #22407: Emit semantic token modifier `async` for impl Future
I randomly discovered `async` token modifiers getting emitted and thought it might be an interesting thing to set up to get highlighted using a specific color.  
However, I noticed that some futures did not get that token modifier, leading to a confusing result, so I disabled it for now. Libraries like embassy seem to use a lot of manually implemented `impl Future` fns or structs that `impl Future`.

It would be great if the `async` token modifier could also get emitted for functions that don't use the async keyword but return `impl Future`, as well as ones that return a struct that impls `Future`.

Examples, note the different highlighting where the greenish color comes from the async token modifier and the yellowish one is the usual one for functions/methods:

Image

Albab-Hasan:
the `async` semantic token modifier was only emitted for `async fn`. functions in embedded async crates like embassy commonly return `impl Future` or a concrete type that implements `Future` without the `async` keyword.

`returns_impl_future` in `hir` is extended to fall through to an `impls_trait` check when the return type is not `impl Trait` syntax covering concrete structs and type aliases that implement `Future`. the `Sized` lang item is optional in the impl Trait path; not a hard failure so its absence in minimal crate environments dosent silently return false.

closes rust-lang/rust-analyzer#22407

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -2790,25 +2790,26 @@ impl Function {
         }
 
         let ret_type = self.ret_type(db);
-        let Some(impl_traits) = ret_type.as_impl_traits(db) else { return false };
         let lang_items = hir_def::lang_item::lang_items(db, self.krate(db).id);
         let Some(future_trait_id) = lang_items.Future else {
             return false;
         };
-        let Some(sized_trait_id) = lang_items.Sized else {
-            return false;
-        };
 
-        let mut has_impl_future = false;
-        impl_traits
-            .filter(|t| {
-                let fut = t.id == future_trait_id;
-                has_impl_future |= fut;
-                !fut && t.id != sized_trait_id
-            })
-            // all traits but the future trait must be auto traits
-            .all(|t| t.is_auto(db))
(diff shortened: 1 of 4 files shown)


ChayimFriedman2:
I object. These functions are not declared async, and hover should be faithful.

Albab-Hasan:
fair

Albab-Hasan:
@ChayimFriedman2 can you take a look at this pls

https://github.com/rust-lang/rust-analyzer/pull/21908

ChayimFriedman2:
Hmm, now I see that this is not about hover but about highlighting. So maybe, I need to form an opinion on this.

ChayimFriedman2:
@rust-lang/rust-analyzer, my personal opinion now is that we should do this. What do you think?

lnicola:
Both seems like defensible positions, but I'm generally in favor of DWIM, so it sounds fine to me.

lnicola:
@Albab-Hasan can you rebase?

PR #22415, opened 2026-05-21
fix: allow wildcard params in foreign fn declarations

youngspe:
A wildcard `_` param in an extern block produces a [E0130](https://doc.rust-lang.org/stable/error_codes/E0130.html) diagnostic, but compiles just fine.

- accept wildcards in addition to identifiers for parameters in `extern` blocks
- add tests to verify diagnostics are not generated for the simple ident and wildcard case

See https://doc.rust-lang.org/reference/items/external-blocks.html#r-items.extern.fn.param-patterns

### DIFF
--- a/crates/hir-def/src/expr_store/lower.rs
+++ b/crates/hir-def/src/expr_store/lower.rs
@@ -2570,9 +2570,9 @@ impl<'db> ExprCollector<'db> {
     }
 
     fn collect_extern_fn_param(&mut self, pat: Option<ast::Pat>) -> PatId {
-        // `extern` functions cannot have pattern-matched parameters, and furthermore, the identifiers
-        // in their parameters are always interpreted as bindings, even if in a normal function they
-        // won't be, because they would refer to a path pattern.
+        // parameters of functions in `extern` blocks can only be simple identifiers and wildcards.
+        // Furthermore, the identifiers in their parameters are always interpreted as bindings, even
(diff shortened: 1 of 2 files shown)


crates/hir-def/src/expr_store/lower.rs
```diff
@@ -2588,6 +2588,7 @@ impl<'db> ExprCollector<'db> {
                 self.add_definition_to_binding(binding, pat);
                 pat
             }
+            ast::Pat::WildcardPat(_) => self.collect_pat_top(Some(pat)),
```
```suggestion
            ast::Pat::WildcardPat(_) => self.alloc_pat(Pat::Wild, AstPtr::new(&pat)),
```
Please inline this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22415/7f916ab1b1f669cec017960c2d91a9a87b4b7bae..e8c944789d40da8fe7d981a6e2e4310f45ba0c68)*

ChayimFriedman2 (changes requested):

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22415/7f916ab1b1f669cec017960c2d91a9a87b4b7bae..4bb5ca397199f3dbb8c3d0a3d7af94f437b262ce)*

PR #22416, opened 2026-05-21
feat: completions imports exclude supports sub items

### LINKED ISSUE #20468: How to ignore some path for auto-import
```rust
use core::convert::identity;

fn main() {
    identity(i$0)
}
```

Output completions:

```
identity(…)~ const fn(T) -> T           f [LS]
i8 i8                                   S [LS]
if~                                     k [LS]
i16 i16                                 S [LS]
i32 i32                                 S [LS]
i64 i64                                 S [LS]
i128 i128                               S [LS]
isize isize                             S [LS]
if let~                                 k [LS]
include                                 r [LS]
include_str                             r [LS]
include_bytes                           r [LS]
io(use std::os::unix::io)               M [LS]
id()~(use std::process::id) fn() -> u32 f [LS]
i16(use core::i16)                      M [LS]
i64(use core::i64)                      M [LS]
i8x2(use core::simd::i8x2)              S [LS]
isize(use core::isize)                  M [LS]
i16x1(use core::simd::i16x1)            S [LS]
i16x2(use core::simd::i16x2)            S [LS]
i16x8(use core::simd::i16x8)            S [LS]
i32x1(use core::simd::i32x1)            S [LS]
i32x8(use core::simd::i32x8)            S [LS]
i64x1(use core::simd::i64x1)            S [LS]
i64x4(use core::simd::i64x4)            S [LS]
i64x8(use core::simd::i64x8)            S [LS]
i8x16(use core::simd::i8x16)            S [LS]
```

How can I make auto-import ignore paths prefixed with `core::simd` in the config file?

A4-Tacks:
Close rust-lang/rust-analyzer#20468

Similar to implementing https://github.com/rust-lang/rust-analyzer/issues/20468#issuecomment-3343893620 , but non-based string

Example
---
**Before this PR**

```json
{
        {"path":"std::intrinsics::AtomicOrdering","type":"always"},
        {"path":"std::intrinsics::ConstParamTy","type":"always"},
        // ... 223 items
    ]
}
```

**After this PR**

```json
{
        {"path":"std::intrinsics","type":"sub_items"},
    ]
}
```

### DIFF
--- a/crates/ide-completion/src/completions/flyimport.rs
+++ b/crates/ide-completion/src/completions/flyimport.rs
@@ -389,7 +389,7 @@ fn filter_excluded_flyimport(ctx: &CompletionContext<'_, '_>, import: &LocatedIm
     }
     let method_imported = import.item_to_import != import.original_item;
     if method_imported
-        && (is_exclude_flyimport.is_some()
+        && (is_exclude_flyimport == Some(AutoImportExclusionType::Methods)
             || ctx.exclude_flyimport.contains_key(&import.original_item.into_module_def()))
     {
         // If this is a method, exclude it either if it was excluded itself (which may not be caught above,
--- a/crates/ide-completion/src/completions/flyimport.rs
+++ b/crates/ide-completion/src/completions/flyimport.rs
@@ -389,7 +389,7 @@ fn filter_excluded_flyimport(ctx: &CompletionContext<'_, '_>, import: &LocatedIm
     }
     let method_imported = import.item_to_import != import.original_item;
     if method_imported
-        && (is_exclude_flyimport == Some(AutoImportExclusionType::Methods)
+        && (is_exclude_flyimport.is_some()
             || ctx.exclude_flyimport.contains_key(&import.original_item.into_module_def()))
     {
         // If this is a method, exclude it either if it was excluded itself (which may not be caught above,
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -662,6 +662,9 @@ config_data! {
         /// For traits the type "methods" can be used to only exclude the methods but not the trait
         /// itself.
         ///
+        /// For modules the type "subItems" can be used to only exclude the sub items but not the module
+        /// itself.
+        ///
         /// This setting also inherits `#rust-analyzer.completion.excludeTraits#`.
         completion_autoimport_exclude: Vec<AutoImportExclusion> = vec![
             AutoImportExclusion::Verbose { path: "core::borrow::Borrow".to_owned(), r#type: AutoImportExclusionType::Methods },
@@ -1938,6 +1941,9 @@ impl Config {
                             AutoImportExclusionType::Methods => {
                                 ide_completion::AutoImportExclusionType::Methods
                             }
+                            AutoImportExclusionType::SubItems => {
+                                ide_completion::AutoImportExclusionType::SubItems
+                            }
                         },
                     ),
                 })
@@ -2997,6 +3003,7 @@ pub enum AutoImportExclusion {
 pub enum AutoImportExclusionType {
     Always,
     Methods,
+    SubItems,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
@@ -4128,10 +4135,11 @@ fn field_props(field: &str, ty: &str, doc: &[&str], default: &str) -> serde_json
                             },
                             "type": {
                                 "type": "string",
-                                "enum": ["always", "methods"],
+                                "enum": ["always", "methods", "subItems"],
                                 "enumDescriptions": [
                                     "Do not show this item or its methods (if it is a trait) in auto-import completions.",
-                                    "Do not show this traits methods in auto-import completions."
+                                    "Do not show this traits methods in auto-import completions.",
+                                    "Do not show this modules sub items in auto-import completions."
                                 ],
                             },
                         }
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -662,8 +662,8 @@ config_data! {
         /// For traits the type "methods" can be used to only exclude the methods but not the trait
         /// itself.
         ///
-        /// For modules the type "subItems" can be used to only exclude the sub items but not the module
-        /// itself.
+        /// For modules the type "subItems" can be used to only exclude the all items in it but not the module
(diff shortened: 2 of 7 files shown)


ChayimFriedman2:
Intrinsics won't be completed anymore after https://github.com/rust-lang/rust-analyzer/pull/22044. Is there another motivation?

A4-Tacks:
Like `std::arch::aarch64::` with >7000 items

ChayimFriedman2:
Fair enough.

A4-Tacks:
> Fair enough.

What do you mean?

crates/ide-completion/src/completions/flyimport.rs
```diff
@@ -389,7 +389,7 @@ fn filter_excluded_flyimport(ctx: &CompletionContext<'_, '_>, import: &LocatedIm
     }
     let method_imported = import.item_to_import != import.original_item;
     if method_imported
-        && (is_exclude_flyimport.is_some()
+        && (is_exclude_flyimport == Some(AutoImportExclusionType::Methods)
```
ChayimFriedman2: While it does not make things less correct, it also does not make them more correct; therefore I see no reason for this change.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22416/7f916ab1b1f669cec017960c2d91a9a87b4b7bae..26395783dd1ba6cf783a786461f9a9670435df0f)*

crates/rust-analyzer/src/config.rs
```diff
@@ -662,6 +662,9 @@ config_data! {
         /// For traits the type "methods" can be used to only exclude the methods but not the trait
         /// itself.
         ///
+        /// For modules the type "subItems" can be used to only exclude the sub items but not the module
```
ChayimFriedman2:
```suggestion
        /// For modules, the type "subItems" can be used to exclude the module and all items in it. This does not include items defined in nested modules.
```
We should also consider whether it should really exclude the module itself, probably not but then the problem is that there is no way to exclude both.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22416/7f916ab1b1f669cec017960c2d91a9a87b4b7bae..26395783dd1ba6cf783a786461f9a9670435df0f)*

crates/rust-analyzer/src/config.rs
```diff
@@ -4128,10 +4135,11 @@ fn field_props(field: &str, ty: &str, doc: &[&str], default: &str) -> serde_json
                             },
                             "type": {
                                 "type": "string",
-                                "enum": ["always", "methods"],
+                                "enum": ["always", "methods", "subItems"],
                                 "enumDescriptions": [
                                     "Do not show this item or its methods (if it is a trait) in auto-import completions.",
-                                    "Do not show this traits methods in auto-import completions."
+                                    "Do not show this traits methods in auto-import completions.",
```
ChayimFriedman2:
```suggestion
                                    "Do not show this trait's methods in auto-import completions.",
```
A preexisting grammatical mistake.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22416/7f916ab1b1f669cec017960c2d91a9a87b4b7bae..26395783dd1ba6cf783a786461f9a9670435df0f)*

crates/rust-analyzer/src/config.rs
```diff
@@ -4128,10 +4135,11 @@ fn field_props(field: &str, ty: &str, doc: &[&str], default: &str) -> serde_json
...
+                                    "Do not show this modules sub items in auto-import completions."
```
ChayimFriedman2:
```suggestion
                                    "Do not show this modules and items defined in it in auto-import completions.",
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22416/7f916ab1b1f669cec017960c2d91a9a87b4b7bae..26395783dd1ba6cf783a786461f9a9670435df0f)*

ChayimFriedman2:
> > Fair enough.
> 
> What do you mean?

I mean that I agree this is something worth having.

crates/rust-analyzer/src/config.rs
A4-Tacks: > but then the problem is that there is no way to exclude both.

~~Both can be excluded, such as:~~

Oh, maybe we can add a `SmallVec<AutoImportExclusiveType>`?

A4-Tacks: Or `ide_completion::AutoImportExclusionType::SubItems { with_self: bool }` and `AutoImportExclusionType::SubItemsAndSelf`?

A4-Tacks: However, I think it is more reasonable not to exclude the module itself.
Taking `std::arch::aarch64::float16x8_t` as an example, not wanting to import it also means that if necessary, we would prefer to have auto import of `std::arch::aarch64` it self, to use `aarch64::float16x8_t`

ChayimFriedman2: > However, I think it is more reasonable not to exclude the module itself.

> We should also consider whether it should really exclude the module itself, **probably not**

That's what I said too.

ChayimFriedman2: "sub items" is not clear enough. That's why I used the term "items defined in this module".

crates/rust-analyzer/src/config.rs
```diff
@@ -662,6 +662,9 @@ config_data! {
         /// For traits the type "methods" can be used to only exclude the methods but not the trait
         /// itself.
         ///
+        /// For modules the type "subItems" can be used to only exclude the all items in it but not the module
+        /// itself. This does not include items defined in nested modules.
```
ChayimFriedman2, line 666: Note that this is not what implemented; currently the module *is* excluded.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22416/987ea33645ab1c709b1df6823038abcb2fe8973e..ef9e665ded7b004c94edda1e9badc81df31ecc77)*

ChayimFriedman2 (changes requested):

crates/rust-analyzer/src/config.rs
A4-Tacks: No, it didn't. I improved the testing

ChayimFriedman2 (approved):

PR #22426, opened 2026-05-23
fix(cfg): correct separator index in CfgDiff disable loop

shulaoda:
The disable loop reused `self.enable.len() - 1` when detecting the last atom, causing wrong "and" placement (and a debug-mode panic when enable was empty). Add a regression test for `disable a and b`.

🤖 Spotted via AI-assisted code review.


Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22426/56dc60dd17ff8f527fdd1732a179e47c4eb244f5..610dd2ac2ba8b38c99132c40e6b6778b771c40db)*

PR #22437, opened 2026-05-24
fix: `extract_module` missing import for macro calls

### LINKED ISSUE #20209: Extract module assist fails to add import for function-like macro
Given the following input:

```rust
use serde_json::json;

// extract this function into a module
fn foo() {
    let _ = json!({});
}
```

rust-analyzer generates a new module lacking the `serde_json::json` import:

```rust
mod modname {
    pub(crate) fn foo() {
        let _ = json!({});
    }
}
```

and doesn't remove the import from the parent module, even if no other uses of `json!` exist (weirdly enough, the existing module import _is_ removed if it is enclosed in curly braces, i.e. `use serde_json::{json};`).

**rust-analyzer version**: 0.4.2532-standalone
**rustc version**: 1.90.0-nightly (0fa4ec6cd 2025-06-25)
**editor or extension**: VSCode, extension version 0.4.2532

@rustbot label +A-assists

Albab-Hasan:
`check_def_in_mod_and_out_sel` had arms for every item kind (`Function`, `Adt`, `Const`, etc.) but no arm for `Definition::Macro`. it fell through to `_ => {}` and returned `(false, false)`, making `def_out_sel` always false for macros. in downstream the condition `(def_out_sel || !is_item) && use_stmt_not_in_sel` was never satisfied so no import was added to the extracted module.

add `Definition::Macro(x) => check_item!(x),` to the match so the macros source location is checked the same way as other items.

closes https://github.com/rust-lang/rust-analyzer/issues/20209

### DIFF
--- a/crates/ide-assists/src/handlers/extract_module.rs
+++ b/crates/ide-assists/src/handlers/extract_module.rs
@@ -733,6 +733,7 @@ fn check_def_in_mod_and_out_sel(
         Definition::Static(x) => check_item!(x),
         Definition::Trait(x) => check_item!(x),
         Definition::TypeAlias(x) => check_item!(x),
+        Definition::Macro(x) => check_item!(x),
         _ => {}
     }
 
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22437/a34f52556aa6171586aa5c6ae9a3f94e12c3223e..8bd24d2f568ebbe249dedf423467046fbddbc335)*

PR #22440, opened 2026-05-24
feat: complete locals in record field

A4-Tacks:
Example
---
```rust
fn outer() {
    let foo = 3;
    {
        let bar = 3;
        struct Foo {
            b$0
        }
        let baz = 3;
        let qux = 3;
    }
    let fez = 3;
}
```

Completes:

```text
bn bar: i32,
bn baz: i32,
bn foo: i32,
kw pub
kw pub(crate)
kw pub(super)
```

### DIFF
--- a/crates/ide-completion/src/completions/field.rs
+++ b/crates/ide-completion/src/completions/field.rs
@@ -1,7 +1,9 @@
 //! Completion of field list position.
 
+use syntax::{AstNode, ast};
+
 use crate::{
-    CompletionContext, Completions,
+    CompletionContext, CompletionItem, CompletionItemKind, Completions,
     context::{PathCompletionCtx, Qualified},
 };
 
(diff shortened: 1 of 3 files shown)


ChayimFriedman2:
I feel like local structs, especially in the middle of a function, are rare enough to not justify this.

A4-Tacks:
This can be temporary, first creating a structure within the function and then move it to the top level

This assist is very useful when refactoring large functions

ChayimFriedman2:
I've never done that and I fail to even imagine how it'll be useful.

I don't want this, but CC @rust-lang/rust-analyzer.

Veykril:
I agree, this seems to have little value and in fact might even be confusing at times due to showing non fields here,

A4-Tacks:
> confusing at times due to showing non fields here,

This was originally only a completion for visibility, I think there's no problem

ChayimFriedman2:
Given that @Veykril seems to agree with me that we don't want this, I'm closing this PR.

PR #22444, opened 2026-05-25
fix: rename schema subItems with sub_items

A4-Tacks:
Fixup rust-lang/rust-analyzer#22416

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -662,7 +662,7 @@ config_data! {
         /// For traits the type "methods" can be used to only exclude the methods but not the trait
         /// itself.
         ///
-        /// For modules the type "subItems" can be used to only exclude the all items in it but not the module
+        /// For modules the type "sub_items" can be used to only exclude the all items in it but not the module
         /// itself. This does not include items defined in nested modules.
         ///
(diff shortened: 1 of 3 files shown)


lnicola (approved):

lnicola:
Hmm, I always write subitem, not sub-item, as if it's a single word.

A4-Tacks:
Or change the renaming config of serde?

`subItems`, `sub_items`, `subitems`

However, considering that it has already been released, it may be necessary to consider breaking

ChayimFriedman2:
It is a breaking change (although the breakage is minimal). If we had come up with this initially, fine, but now, I don't think it's worth the breakage.

lnicola:
I wouldn't be that worried about the breaking change, most users didn't even update yet. There are more people who win the lottery than those who might be affected by this.

ChayimFriedman2:
True, on the other hand I wouldn't be worried about this change either. If it was something clearly better, maybe.

A4-Tacks:
Can we merge now?
It's best to merge as soon as possible, otherwise users will be confused about the schema

ChayimFriedman2:
I still think we should not merge this, but I agree that if we do we should do it now.

A4-Tacks:
Btw, this PR doesn't have *any* breaking, it's just to make the schema correct

PR #22445, opened 2026-05-25
fix: drop lower-ranked tokens from descended set in hover

zrbecker:
## Summary

Filter the descended-token list in `hover_offset` to keep only the highest-ranked tokens. `Ranker` already scores descended tokens, but the rendering loop iterates over the full list, producing spurious hover blocks when an attribute proc macro emits tokens via `quote_spanned!{user_token.span()=> ...}`.

## Reproduction

Using `crates/test-fixture`'s existing `generate_suffixed_type` proc macro:

```rust
//- proc_macros: generate_suffixed_type
#[proc_macros::generate_suffixed_type]
struct Foo$0;
```

Hover on Foo previously rendered both struct Foo and struct FooSuffix. Test added for this.

### In the wild

Same root cause: `#[tokio::main]` (tokio-rs/tokio#5518), `#[async_trait]` (rust-lang/rust-analyzer#14993), `#[tracing::instrument]` (rust-lang/rust-analyzer#19944). For tokio, hover on Ok in `#[tokio::main] async fn main() -> Result<(), String> { Ok(()) }` previously produced 7 hover blocks including ~17KB of unrelated tokio crate docs; now 1.

### DIFF
--- a/crates/ide/src/hover.rs
+++ b/crates/ide/src/hover.rs
@@ -242,6 +242,13 @@ fn hover_offset(
 
     descended.sort_by_cached_key(|tok| !ranker.rank_token(tok));
 
+    // Drop lower-ranked tokens: usually proc-macro-generated content sharing
+    // the cursor's span via `set_span`.
+    if let Some(top_token) = descended.first() {
+        let top_rank = ranker.rank_token(top_token);
+        descended.retain(|tok| ranker.rank_token(tok) == top_rank);
+    }
+
     let mut res = vec![];
     for token in descended {
         let is_same_kind = token.kind() == ranker.kind;
--- a/crates/ide/src/hover/tests.rs
+++ b/crates/ide/src/hover/tests.rs
@@ -2274,6 +2274,35 @@ fn foo$0() {}
 }
 
 #[test]
+fn test_hover_skips_unrelated_proc_macro_generated_tokens_at_same_span() {
+    // `generate_suffixed_type` emits an extra `struct FooSuffix;` whose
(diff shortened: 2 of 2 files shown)


Veykril:
Whats the rationale behind this exactly? We intentionally show more than just one hover result here, we can't really tell whats of interest to the user here and what not, so the best heuristic is sorting but not filtering.

The proper fix for this likely requires def site hygiene, a yet to be stabilized feature in Rust.

zrbecker:
> Whats the rationale behind this exactly? We intentionally show more than just one hover result here, we can't really tell whats of interest to the user here and what not, so the best heuristic is sorting but not filtering.
> 
> The proper fix for this likely requires def site hygiene, a yet to be stabilized feature in Rust.

My rational is a token with different kind or text than what the cursor is on is probably not what the user wants. That's the same implicit assumption behind sorting. We're just taking it one step further.

That said, the status quo is pretty bad: hovering on `Ok(())` in a `#[tokio::main]` function produces 7 hover blocks including 17KB of tokio crate docs. Users filed this against tokio, async-trait, and tracing.

Is there an approach you'd consider acceptable short of def site hygiene? One option: expose this as an opt-in config flag. It's something I'd personally enable.

Happy to close if the answer is "wait for def site hygiene."

ChayimFriedman2:
I agree with @Veykril that's not something we want. I don't want it even behind a config.

The crux of the issue is that the Span APIs and proc macro authors optimize for rustc and not IDEs. And it's very hard to do both.

Perhaps a new method for Span can be added, to disable it for IDE purposes.

PR #22446, opened 2026-05-25
minor: support unwrap single expr stmt with unit type

A4-Tacks:
Example
---
```rust
fn foo() {}
fn main() {
    let f = || {$0 foo(); };
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
fn foo() {}
fn main() {
    let f = || foo();
}
```


ChayimFriedman2:
I'm conflicted about this. On one hand, this *is* useful. On the other hand, this assist is supposed to be "dumb", operating over AST. Adding type information to the mix is not necessarily wise.

PR #22448, opened 2026-05-25
Implement `rust-analyzer/evaluatePredicate` lsp extension

Veykril:
Allows you to have rust-analyzer solve an arbitrary predicate for the cursor position, telling you whether the predicate definitely holds.

https://github.com/user-attachments/assets/8273a170-caa5-4ab9-8f17-a5cd04944f10

Authored with gpt-5.5 and manual reviews / touch ups

### DIFF
--- a/crates/hir-ty/src/consteval.rs
+++ b/crates/hir-ty/src/consteval.rs
@@ -23,6 +23,7 @@ use crate::{
     db::{AnonConstId, AnonConstLoc, GeneralConstId, HirDatabase},
     display::DisplayTarget,
     generics::Generics,
+    lower::LoweringMode,
     mir::{MirEvalError, MirLowerError, pad16},
     next_solver::{
         Allocation, Const, ConstKind, Consts, DbInterner, DefaultAny, GenericArgs, ParamConst,
@@ -305,6 +306,7 @@ pub(crate) enum CreateConstError<'db> {
     DoesNotResolve,
     ConstHasGenerics,
     UnderscoreExpr,
+    AnonConstInterningDisabled,
     TypeMismatch {
         #[expect(unused, reason = "will need this for diagnostics")]
         actual: Ty<'db>,
@@ -355,6 +357,7 @@ pub(crate) fn create_anon_const<'a, 'db>(
     expected_ty: Ty<'db>,
     generics: &dyn Fn() -> &'a Generics<'db>,
     create_var: Option<&mut dyn FnMut(Span) -> Const<'db>>,
+    lowering_mode: LoweringMode,
     forbid_params_after: Option<u32>,
 ) -> Result<Const<'db>, CreateConstError<'db>> {
     match &store[expr] {
@@ -374,6 +377,10 @@ pub(crate) fn create_anon_const<'a, 'db>(
             konst
         }
         _ => {
+            let Some(token) = lowering_mode.allow_tracked_structs() else {
+                return Err(CreateConstError::AnonConstInterningDisabled);
+            };
+
             let allow_using_generic_params = forbid_params_after.is_none();
             let konst = AnonConstId::new(
                 interner.db,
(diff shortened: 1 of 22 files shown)


crates/hir-ty/src/traits.rs
```diff
@@ -153,6 +157,69 @@ pub fn implements_trait_unique_with_infcx<'db>(
...
+    store: &ExpressionStore,
+    def: ExpressionStoreOwnerId,
+    generic_def: GenericDefId,
+    env: ParamEnvAndCrate<'db>,
+    predicate: &WherePredicate,
+) -> WherePredicateEvaluation {
+    let interner = DbInterner::new_with(db, env.krate);
+    let infcx = interner.infer_ctxt().build(TypingMode::PostAnalysis);
+    let generics = OnceCell::<Generics<'db>>::new();
+    let mut ctx = TyLoweringContext::new(
```
ChayimFriedman2, line 184: The problem with lowering here is that it may intern some things (blocks, anon consts, items). This amplifies the problem we already have by lowering in `hir`.

IMO the only solution is to have a mode for TyLoweringContext where it never interns anything.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22448/a7e78201ab9d9830c3cbc7e278af3c2b878c11d6..1f3205639839078f39e6d252218adffcbdac9085)*

crates/hir-ty/src/traits.rs
```diff
@@ -153,6 +157,69 @@ pub fn implements_trait_unique_with_infcx<'db>(
...
+    }
+
+    for clause in clauses {
+        let obligation = Obligation::new(
+            interner,
+            ObligationCause::dummy(),
+            env.param_env,
+            clause.as_predicate(),
+        );
+        if !infcx.predicate_must_hold_modulo_regions(&obligation) {
```
ChayimFriedman2: When you have many obligations to prove you should create an `ObligationCtxt` instead (this is what this method does internally anyway).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22448/a7e78201ab9d9830c3cbc7e278af3c2b878c11d6..1f3205639839078f39e6d252218adffcbdac9085)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/consteval.rs
```diff
@@ -298,13 +298,29 @@ pub fn try_const_isize<'db>(db: &'db dyn HirDatabase, c: Const<'db>) -> Option<i
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AnonConstInterning {
```
ChayimFriedman2: I feel this is a bit too specific given we only have the distinction between lowering during analysis and lowering for IDE, and that there are many forms of lowering we use (e.g. hir-def). Maybe something like `enum LoweringMode { Analysis, Ide }`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22448/96556515cd87746f53d67aec3015775d6c97ef00..3ae6afb8401d2f6c6aa08450d54ac376efda54e9)*

ChayimFriedman2 (approved):

crates/hir-ty/src/consteval.rs
Veykril: Yea definitely, was just not planning into looking into this more than that right now. Will do another pass and generalize that later (or I guess rename for now and add a fixme)

Veykril: Changed this a to single `LoweringMode` enum, currently only tracks anon consts. Will move blocks over to this in a separate PR (and then try turning the blocks back into tracked structs in fact).

crates/hir/src/lib.rs
dfireBird, line 1: Sorry if I'm overstepping my bounds, but don't you think IDE messages are leaking into HIR with having message here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22448/fb9c232bf7a26cda05557b685b3025d2dcdbffdf..b569a4fad4317c20d52b0c57c1324d58edccb913)*

dfireBird (approved):

lnicola:
https://github.com/user-attachments/assets/b7fd7e96-3c53-4943-8f57-95d60bff5e13

PR #22449, opened 2026-05-25
fix: analysis expected ty in enum variant

A4-Tacks:
Fixup rust-lang/rust-analyzer#22425

Example
---
```rust
enum Foo<T> { Var(T) };
fn foo(x: Foo<Option<i32>>) -> Foo {
   match x { Foo::Var($0) => () }
}
```

**Before this PR**

```rust
ty: ?, name: ?
```

**After this PR**

```rust
ty: Option<i32>, name: ?
```

### DIFF
--- a/crates/hir/src/semantics.rs
+++ b/crates/hir/src/semantics.rs
@@ -1947,6 +1947,14 @@ impl<'db> SemanticsImpl<'db> {
         self.analyze(field.syntax())?.resolve_record_pat_field(self.db, field)
     }
 
+    pub fn resolve_tuple_struct_pat_fields(
+        &self,
+        tuple_struct_pat: &ast::TupleStructPat,
+    ) -> Option<Vec<(Field, Type<'db>)>> {
+        self.analyze(tuple_struct_pat.syntax())?
+            .resolve_tuple_struct_pat_fields(self.db, tuple_struct_pat)
+    }
+
     // FIXME: Replace this with `resolve_macro_call2`
     pub fn resolve_macro_call(&self, macro_call: &ast::MacroCall) -> Option<Macro> {
         let macro_call = self.find_file(macro_call.syntax()).with_value(macro_call);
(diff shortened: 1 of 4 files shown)


ChayimFriedman2:
Like I said in https://github.com/rust-lang/rust-analyzer/pull/22452, I object to extending completion analysis now.

A4-Tacks:
This is an unexpected regression introduced in #22425

crates/hir/src/semantics.rs
```diff
@@ -1947,6 +1947,14 @@ impl<'db> SemanticsImpl<'db> {
         self.analyze(field.syntax())?.resolve_record_pat_field(self.db, field)
     }
 
+    pub fn resolve_tuple_struct_pat_fields(
```
ChayimFriedman2, line 1951: We don't need another Semantics helper, we can use `resolve_path()`.

ChayimFriedman2 (changes requested):
Well, I'll merge this now, but please refrain from opening more PRs like this.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22449/a7e78201ab9d9830c3cbc7e278af3c2b878c11d6..6658e7459ae4224f7f9770898785018c4f117acf)*

crates/hir/src/semantics.rs
A4-Tacks: This is regression to #22425

ChayimFriedman2: The real fix is to still use `resolve_path()` and instantiate the generic args with the inferred types. This can probably be done by combining the pattern's type and the resolve path. In case not, I'd be fine with reverting https://github.com/rust-lang/rust-analyzer/pull/22425 for now.

ChayimFriedman2: Although now I see we already have such helper for record patterns. Which is wrong IMO, we should remove it, but I remove my objection to adding this helper for now.

ChayimFriedman2 (approved):

PR #22452, opened 2026-05-25
fix: Add type_match score for struct_pat

A4-Tacks:
Example
---
```rust
struct Foo(Bar);
struct Bar { field: i32 }
fn go(Foo($0): Foo) {}
```

**Before this PR**

```text
st Bar  []
bn Bar {…} Bar { field$1 }$0 []
st Foo  []
bn Foo(…) Foo($1)$0 []
```

**After this PR**

```text
bn Bar {…} Bar { field$1 }$0 [type]
st Bar  []
st Foo  []
bn Foo(…) Foo($1)$0 []
```

### DIFF
--- a/crates/ide-completion/src/item.rs
+++ b/crates/ide-completion/src/item.rs
@@ -181,6 +181,9 @@ pub struct CompletionRelevance {
     /// }
     /// ```
     pub is_local: bool,
+    /// This is missing variant in the patterns.
+    /// Maybe this can also be used for struct fields.
+    pub is_missing: bool,
     /// Populated when the completion item comes from a trait (impl).
     pub trait_: Option<CompletionRelevanceTraitInfo>,
     /// This is set when an import is suggested in a use item whose name is already imported.
@@ -286,6 +289,7 @@ impl CompletionRelevance {
             exact_name_match,
             type_match,
             is_local,
+            is_missing,
             is_name_already_imported,
             requires_import,
             is_private_editable,
@@ -304,6 +308,9 @@ impl CompletionRelevance {
         }
         // slightly prefer locals
         if is_local {
+            score += 1;
+        }
+        if is_missing {
             score += 1;
         }
 
--- a/crates/ide-completion/src/render.rs
+++ b/crates/ide-completion/src/render.rs
@@ -874,6 +874,7 @@ mod tests {
                 exact_name_match,
                 type_match,
                 is_local,
+                is_missing,
                 trait_,
                 is_name_already_imported: _,
                 requires_import,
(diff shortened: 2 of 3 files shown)


ChayimFriedman2:
I've already said that I am not a fan of extending completion analysis until we find a more principled approach, like speculative execution.

A4-Tacks:
> extending completion analysis

This is basically an improvement on pure rendering

ChayimFriedman2:
> This is basically an improvement on pure rendering

What do you mean?

My stance is that completion analysis right now is unmaintainable and implemented the wrong way and extending it is amplifying that. Therefore I'm opposed to *any* extension to it.

A4-Tacks:
This PR is just to supplement the missing type match score

1. Add `type_match` for struct pat (pure render/pattern.rs)
2. Add `type_match` for nested enum variant (bit of context/analysis.rs)

About 2., it shouldn't worsen the situation

ChayimFriedman2:
It does worsen the situation. Just a bit, sure, but a bit here and a bit there and suddenly the situation is a lot more worse than it's already is. As I said, I object to *any* change to analysis. Analysis is pretty fine currently, and nothing will happen if some bugs will stay unfixed until we figure out a better way (and of course, I invite you to be part of this process).

A4-Tacks:
Okay, this is no longer related to the context/analysis

ChayimFriedman2:
This is worse now because this *should* be done via analysis. Adding workarounds around analysis is worse than changing it.

As I said, I have absolutely no problem that this bug will remain unfixed until we figure out a better way to handle analysis. Open an issue and that's it.

A4-Tacks:
> This is worse now because this *should* be done via analysis.

Why? IMO, this is the most correct approach rather than analysis

IMO, the old behavior is the workarounds, this should have always applied type match

ChayimFriedman2:
It's the job of analysis to determine the expected type for what the user will type, which is exactly what we need here.

Framing it differently: if we would use inference's analysis for that, which is what I want us to do eventually, we would get this for free.

A4-Tacks:
> It's the job of analysis to determine the expected type for what the user will type, which is exactly what we need here.

This PR fix: The type that has already been analyzed, but the old solution uses selective addition of type matches to the completion item to change the score

ChayimFriedman2:
Okay, I see now I misunderstood the current change, and you were right (although I can't understand what the intent of the original code was, why only compute type match for missing variants?).

ChayimFriedman2 (approved):

A4-Tacks:
> although I can't understand what the intent of the original code was, why only compute type match for missing variants?

As I said, this is a workaround that can be used to improve scores when type matching occurs

The correct approach is this PR, type matching always occurs, and missing variants is an additional score

PR #22457, opened 2026-05-26
fix: use add deref in assign instead add `&mut` for value

### LINKED ISSUE #22456: Add deref instead of mut ref on assign type mismatch
```rust
fn foo(x: &mut i32) {
    x = 2;
}
```

**Actual**:

```rust
fn foo(x: &mut i32) {
    x = &mut 2;
}
```

**Expect**:

```rust
fn foo(x: &mut i32) {
    *x = 2;
}
```

But it's not that simple:

1. For mutable place, it may indeed be necessary to assign a reference value
2. For immutable variables, perhaps the user wishes to continue adding `mut`

Is it worth catch this? Especially in situations where it is difficult to identify

Perhaps not offer fixes in the assignment is also an option?

A4-Tacks:
Close rust-lang/rust-analyzer#22456

Example
---
```rust
fn test(arg: &mut i32) {
    arg = $02;
}
```

**Before this PR**

```rust
fn test(arg: &mut i32) {
    arg = &mut 2;
}
```

**After this PR**

```rust
fn test(arg: &mut i32) {
    *arg = 2;
}
```

### DIFF
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -120,6 +120,19 @@ fn add_or_fix_reference(
         return None;
     }
 
+    let expr = expr_ptr.to_node(ctx.db());
+    let assign = expr.syntax().parent().and_then(ast::BinExpr::cast).filter(|it| {
+        it.op_kind() == Some(ast::BinaryOp::Assignment { op: None }) && it.rhs() == Some(expr)
+    });
+    if let Some(assign) = assign
+        && expected_mutability.is_mut()
+        && let Some(range) = ctx.sema.original_range_opt(assign.syntax())
+    {
+        let edit = TextEdit::insert(range.range.start(), "*".to_owned());
+        let source_change = SourceChange::from_text_edit(range.file_id.file_id(ctx.db()), edit);
+        acc.push(fix("add_deref_here", "Add deref here", source_change, range.range));
+    }
+
     let ampersands = format!("&{}", expected_mutability.as_keyword_for_ref());
 
     let edit = TextEdit::insert(range.range.start(), ampersands);
@@ -503,6 +516,22 @@ fn main() {
     test((&mut 123));
 }
 fn test(_arg: &mut i32) {}
+            "#,
+        );
+    }
+
+    #[test]
+    fn add_deref_in_assign() {
+        check_fix(
+            r#"
+fn test(arg: &mut i32) {
+    arg = $02;
+}
+            "#,
+            r#"
+fn test(arg: &mut i32) {
+    *arg = 2;
+}
             "#,
         );
     }
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -131,6 +131,7 @@ fn add_or_fix_reference(
         let edit = TextEdit::insert(range.range.start(), "*".to_owned());
         let source_change = SourceChange::from_text_edit(range.file_id.file_id(ctx.db()), edit);
         acc.push(fix("add_deref_here", "Add deref here", source_change, range.range));
+        return Some(());
     }
 
     let ampersands = format!("&{}", expected_mutability.as_keyword_for_ref());
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -121,9 +121,11 @@ fn add_or_fix_reference(
     }
 
     let expr = expr_ptr.to_node(ctx.db());
-    let assign = expr.syntax().parent().and_then(ast::BinExpr::cast).filter(|it| {
-        it.op_kind() == Some(ast::BinaryOp::Assignment { op: None }) && it.rhs() == Some(expr)
-    });
+    let assign = expr
+        .syntax()
+        .parent()
+        .and_then(ast::BinExpr::cast)
+        .filter(|it| it.op_kind() == Some(ast::BinaryOp::Assignment { op: None }));
     if let Some(assign) = assign
         && expected_mutability.is_mut()
         && let Some(range) = ctx.sema.original_range_opt(assign.syntax())


crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -120,6 +120,19 @@ fn add_or_fix_reference(
...
+    });
+    if let Some(assign) = assign
+        && expected_mutability.is_mut()
+        && let Some(range) = ctx.sema.original_range_opt(assign.syntax())
+    {
+        let edit = TextEdit::insert(range.range.start(), "*".to_owned());
+        let source_change = SourceChange::from_text_edit(range.file_id.file_id(ctx.db()), edit);
+        acc.push(fix("add_deref_here", "Add deref here", source_change, range.range));
+    }
+
```
ada4a, line 138: I feel like this could go into its own function? AFAICT it isn't really semantically connected to the rest of this function..

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22457/c3af07ad84d68adc5e652e86f0c20009caa29014..9a12f344f82c887989ff1f80e7f766e943accda2)*

crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -120,6 +120,19 @@ fn add_or_fix_reference(
         return None;
     }
 
+    let expr = expr_ptr.to_node(ctx.db());
+    let assign = expr.syntax().parent().and_then(ast::BinExpr::cast).filter(|it| {
+        it.op_kind() == Some(ast::BinaryOp::Assignment { op: None }) && it.rhs() == Some(expr)
+    });
+    if let Some(assign) = assign
```
ada4a: nit: since we're using a let-chain here already, maybe move the `filter` predicate inside it? Because right now it looks like a bit much..
```suggestion
    if let Some(assign) = expr.syntax().parent().and_then(ast::BinExpr::cast)
        && assign.op_kind() == Some(ast::BinaryOp::Assignment { op: None })
        && assin.rhs() == Some(expr)
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22457/c3af07ad84d68adc5e652e86f0c20009caa29014..9a12f344f82c887989ff1f80e7f766e943accda2)*

ada4a (changes requested):

crates/ide-diagnostics/src/handlers/type_mismatch.rs
A4-Tacks: I am against providing both ref and deref fixes at the same time

Ref: https://github.com/rust-lang/rust-analyzer/issues/22456#issuecomment-4538370206

ada4a: In that case you probably want to early-return after `acc.push();`, no?

A4-Tacks: If this is extracted to another function and called by `fn type_mismatch`, both reference and deref will be triggered

And the code is short enough, there is no need to extract it

ada4a: > If this is extracted to another function and called by fn type_mismatch, both reference and deref will be triggered

I understand. What I mean is: AFAICT, this code, as currently written, will try to execute the `assign` branch, emit the deref fix if it succeeds, _and_ continue on to the rest of the function, thus emitting the reference fix as well. Therefore, I think you need to add `return Some(());` after the `acc.push(..);`, to avoid the function from continuing execution

> And the code is short enough, there is no need to extract it

Maybe renaming it would make sense though, since it no longer deals with just references but also dereferences. Not sure what a good name would be

A4-Tacks: > I think you need to add `return Some(());` after the `acc.push(..);`

Oh, thank you catch, I forgot

ada4a: Glad to have helped:)

crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -120,6 +120,20 @@ fn add_or_fix_reference(
         return None;
     }
 
+    let expr = expr_ptr.to_node(ctx.db());
+    let assign = expr.syntax().parent().and_then(ast::BinExpr::cast).filter(|it| {
+        it.op_kind() == Some(ast::BinaryOp::Assignment { op: None }) && it.rhs() == Some(expr)
```
ChayimFriedman2:
```suggestion
        it.op_kind() == Some(ast::BinaryOp::Assignment { op: None })
```
We have no guarantee the diagnostic will fire on the RHS. It could as well fire on the LHS in the future.

ChayimFriedman2 (approved):
LGTM minus one nit.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22457/c3af07ad84d68adc5e652e86f0c20009caa29014..6d19cc155d3492c5210eb4d0260b6a746edc188c)*

crates/ide-diagnostics/src/handlers/type_mismatch.rs
A4-Tacks: I think we can merge directly.
If triggered on lhs in the future, it will be difficult to solve from this branch

This function adds a reference, and if triggered on lhs, it should be `expected i32, found: &mut i32`, will return earlier

ChayimFriedman2: I disagree. We could change the diagnostic target without swapping expected/found.

My point is that this check does not add correctness to the code, nor does it simplifies it. So it's redundant.

A4-Tacks: Two possibilities:
```rust
fn test(arg: &mut i32) {
    arg = 2;
        //^ expected &mut i32, found i32 (current)
  //^^^ expected i32, found &mut i32
}
```

But due to `d.expected.as_reference()?`, the latter will return earlier

Is it necessary to solve this problem now?

> My point is that this check does not add correctness to the code

Oh right

ChayimFriedman2 (approved):

ChayimFriedman2:
You need to rebase to solve the proc-macro-srv CI.

PR #22460, opened 2026-05-26
fix: `inline_call` produces invalid `Generic<T>` syntax when inlining `Self` across impl blocks

### LINKED ISSUE #19827: Inline function assist produces invalid code when inlining across generic impl blocks

Albab-Hasan:
when a function using `Self` as a constructor (e.g. `Self(x)`) is inlined from a different `impl<T>` block the `Self` token was expanded to the bare type name without generics, producing `Generic(x)`. the original code stripped all generic args to avoid the invalid `Generic<T>(x)` syntax in expression position.

fix by emitting turbofish `Generic::<T>` for type/const params instead of stripping them. lifetime params are still omitted since they are always inferred in expression position.

also fix `PathTransform`'s `SelfType` arm to use `mod_path_to_ast_with_factory` (returning an `ast::Path` not a `PathType`) and attach generic args as turbofish when in expression context. this covers the case where `PathTransform` is applied with an explicit generic arg list at the call site.

fixes rust-lang/rust-analyzer#19827

### DIFF
--- a/crates/ide-assists/src/handlers/inline_call.rs
+++ b/crates/ide-assists/src/handlers/inline_call.rs
@@ -21,9 +21,10 @@ use syntax::{
     ast::{
         self, HasArgList, HasGenericArgs, Pat, PathExpr,
         edit::{AstNodeEdit, IndentLevel},
+        make,
         syntax_factory::SyntaxFactory,
     },
-    syntax_editor::SyntaxEditor,
+    syntax_editor::{Position, SyntaxEditor},
 };
 
 use crate::{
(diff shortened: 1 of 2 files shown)


crates/ide-assists/src/handlers/inline_call.rs
```diff
@@ -1727,6 +1742,45 @@ fn a() -> bool {
         )
     }
 
+    #[test]
+    fn inline_call_across_generic_impl_blocks() {
+        // rust-lang/rust-analyzer#19827: inlining across two separate `impl<T>` blocks for the
+        // same generic type must not produce `Generic<T>(...)` (invalid) — use turbofish instead.
```
A4-Tacks, line 1748: This has already been fixed by #21381


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22460/d2cf9ceecda33bda5bd30b4d2503bd7fa301053e..b0bce5c2c547bfe97a0d9a9ceb5c4ba15c03afb4)*

crates/ide-assists/src/handlers/inline_call.rs
```diff
@@ -1727,6 +1742,45 @@ fn a() -> bool {
...
+        Self(todo!())
+    }
+}
+"#,
+            r#"
+struct Generic<T>(T);
+
+impl<T> Generic<T> {
+    fn new() -> Self {
+        Generic::<T>(todo!())
```
A4-Tacks, line 1771: If it is assoc fn and has default generic parameters, this may not be correct
For `struct Generic<T, U = i32>`:
`Self::method()` -> `<Generic<T>>::method`, instead of `Generic::<T>::method`


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22460/d2cf9ceecda33bda5bd30b4d2503bd7fa301053e..b0bce5c2c547bfe97a0d9a9ceb5c4ba15c03afb4)*

Albab-Hasan:
@A4-Tacks closing this pr

PR #22463, opened 2026-05-26
feat(diagnostics): split `TraitImplIncorrectSafety` into distinct rustc errors

ada4a:


crates/ide-diagnostics/src/handlers/impl_incorrect_safety.rs
ada4a, line 1: should I split the file into 5 as well?

crates/hir/src/lib.rs
```diff
@@ -895,13 +895,45 @@ impl Module {
...
+                    let impl_ = ast_id_map.get(loc.id.value);
+
+                    if let Some(trait_) = trait_ {
+                        // unsafe impl of safe trait
+                        acc.push(UnsafeImplOfSafeTrait { trait_, file_id, impl_ }.into());
+                    } else {
+                        // unsafe inherent impl
+                        acc.push(UnsafeInherentImpl { file_id, impl_ }.into());
+                    }
+                }
```
ada4a: maybe I should combine these into the same struct, e.g. `UnnecessaryUnsafeImpl`? But then `UnsafeNegativeImpl` would be left out...

ada4a (commented):
some questions

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22463/d2cf9ceecda33bda5bd30b4d2503bd7fa301053e..93172535737496f9b344b9da30f42129679283e0)*

ChayimFriedman2 (changes requested):
I feel like this should remain one error but with an enum to distinguish the kind.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22463/d2cf9ceecda33bda5bd30b4d2503bd7fa301053e..21955002739ec395454da34d7e1560209b450910)*

ada4a:
If I make it just an enum, then the fields `file_id` and `impl_` would be repeated everywhere, which I find unfortunate. What do you think about doing something like this?

```rs
struct ImplIncorrectSafety { // since this actually also handles inherent impls
    file_id: HirFileId,
    impl_: AstPtr<ast::Impl>,
    kind: ImplIncorrectSafetyKind,
}

enum ImplIncorrectSafetyKind {
    UnsafeImplOfSafeTrait { trait_: Trait }
    /* ..other variants */
}
```

ada4a:
> What do you think about doing something like this?

I've gone ahead and implemented that, turned out pretty well imo. I've left it as a separate commit for now -- if you like the approach, please let me know before merging so that I can squash the commits

PR #22464, opened 2026-05-26
feat: Hide private fields on hover depending on context

Veykril:
Hides private fields as follows
| Hover location | Definition location | Render private fields? |
|---|---:|---:|
| Workspace crate | Workspace crate | Yes |
| Workspace crate | External/library crate | No |
| External/library crate | Same external/library crate | Yes |
| External/library crate | Different external/library crate | No |
| Anywhere | Same crate as definition | Yes |

### DIFF
--- a/crates/hir-ty/src/display.rs
+++ b/crates/hir-ty/src/display.rs
@@ -130,6 +130,7 @@ pub struct HirFormatter<'a, 'db> {
     pub entity_limit: Option<usize>,
     /// When rendering functions, whether to show the constraint from the container
     show_container_bounds: bool,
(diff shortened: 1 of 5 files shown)


ChayimFriedman2:
FWIW I do find private fields useful but as long as it's configurable I don't object (even if it's off by default).

Veykril:
I am curious about that thought, I think I was of the same opinion before but at this point I've kind of changed my mind. If its a private field in a dep I don't really see how it would be useful in the common case to see since it wont ever be possible for me to change that. And in the very rare case where it is interesting to then see one could always go to def. Hence this PR.

PR #22466, opened 2026-05-27
feat: add diagnostic for E0596

WaterWhisperer:

### DIFF
--- a/crates/hir/src/diagnostics.rs
+++ b/crates/hir/src/diagnostics.rs
@@ -104,7 +104,6 @@ diagnostics![AnyDiagnostic<'db> ->
     AwaitOutsideOfAsync,
     BreakOutsideOfLoop,
     CannotBeDereferenced<'db>,
-    CannotBorrowAsMutable,
     CannotImplicitlyDerefTraitObject<'db>,
     CannotIndexInto<'db>,
     CastToUnsized<'db>,
@@ -134,6 +133,7 @@ diagnostics![AnyDiagnostic<'db> ->
     MissingMatchArms,
     MissingUnsafe,
     MovedOutOfRef<'db>,
+    MutRefInImmRefPat,
     MutableRefBinding,
     NeedMut,
     NonExhaustiveLet,
@@ -337,7 +337,7 @@ pub struct CannotBeDereferenced<'db> {
 }
 
 #[derive(Debug)]
-pub struct CannotBorrowAsMutable {
+pub struct MutRefInImmRefPat {
     pub pat: InFile<ExprOrPatPtr>,
 }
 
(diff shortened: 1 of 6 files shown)


crates/hir/src/diagnostics.rs
```diff
@@ -335,6 +336,11 @@ pub struct CannotBeDereferenced<'db> {
     pub found: Type<'db>,
 }
 
+#[derive(Debug)]
+pub struct CannotBorrowAsMutable {
```
ChayimFriedman2: This name is a bit too general. Which might be OK inside inference but I feel not here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22466/b48cc1083d4fd7264d968ba613553400a30a90a8..de60e47a93a7eae52010a700d21cf867ffdc2e9d)*

crates/ide-diagnostics/src/handlers/mut_ref_in_imm_ref_pat.rs
```diff
@@ -0,0 +1,37 @@
...
+
+#[cfg(test)]
+mod tests {
+    use crate::tests::check_diagnostics;
+
+    #[test]
+    fn cannot_borrow_as_mutable_inside_shared_ref_pattern() {
+        check_diagnostics(
+            r#"
+#![feature(ref_pat_eat_one_layer_2024)]
```
ChayimFriedman2, line 28: Is this necessary? It shouldn't be.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22466/b48cc1083d4fd7264d968ba613553400a30a90a8..de60e47a93a7eae52010a700d21cf867ffdc2e9d)*

ChayimFriedman2 (changes requested):

crates/hir/src/diagnostics.rs
WaterWhisperer: How about `RefMutBindingInSharedRefPattern`? But I’m worried it may be too long.

ChayimFriedman2: That's fine, but maybe `MutRefInImmRefPat`?

WaterWhisperer: Great, that's better! Thanks

crates/ide-diagnostics/src/handlers/mut_ref_in_imm_ref_pat.rs
WaterWhisperer: This test exercises the `WeaklyNot` path: an `&` pattern matching an `&mut` reference under `ref_pat_eat_one_layer_2024`. Without the feature, `let &ref mut x = &mut 0` does not reach this diagnostic and reports a type mismatch instead. I tried `&0`, but that exercises `MutblCap::Not`, not this path.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22466/b48cc1083d4fd7264d968ba613553400a30a90a8..3b5a1b3b37220e7984e45c409959e016eef8dd62)*

PR #22470, opened 2026-05-27
Add `'db` to `NavigationTarget` and beyond

ada4a:
This addresses the FIXME on https://github.com/rust-lang/rust-analyzer/blob/b48cc1083d4fd7264d968ba613553400a30a90a8/crates/ide/src/navigation_target.rs#L53-L54

(cc @ChayimFriedman2 since you were the one who added it)

To be completely honest, I'm not sure if this is worth it, since `NavigationTarget` ends up appearing inside of a lot of structs which in turn are returned by queries in `lsp`, requiring all of them to be turned into their `'static` versions anyway. One could argue that all the _remaining_ uses of `NavigationTarget` benefit from this change, though it's unclear how many of those there are.

### DIFF
--- a/crates/ide/src/annotations.rs
+++ b/crates/ide/src/annotations.rs
@@ -26,13 +26,13 @@ mod fn_references;
 #[derive(Debug, Hash, PartialEq, Eq)]
 pub struct Annotation {
     pub range: TextRange,
-    pub kind: AnnotationKind,
+    pub kind: AnnotationKind<'static>,
 }
 
 #[derive(Debug, Hash, PartialEq, Eq)]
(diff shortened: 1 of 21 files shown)


Veykril:
Hm, yea I am not sure we can even return these ever with a 'db lifetime.

ada4a:
Should I open a Zulip thread to discuss this, or should I just close the PR? (either of those would need to wait until Thursday)

Veykril:
I think closing unfortunately, I do not think there is a way forward here for this one.

ada4a:
oh well.. I'll later open a PR that removes this FIXME then

PR #22473, opened 2026-05-27
fix: Port block and loop inference from rustc

ChayimFriedman2:


Veykril (approved):

ChayimFriedman2:
New Rust, new Clippy!

PR #22474, opened 2026-05-27
refactor: Intern MIR projections using the new ty interning infrastructure

flodiebold:
This allows creating projections from other queries, which is something that later analyses sometimes do in rustc. Also, it aligns the APIs a bit more with rustc's. This should make it easier to port more analyses (like move tracking) from rustc. The performance/memory usage seems to be about the same, at least looking at analysis-stats.

### DIFF
--- a/crates/hir-ty/src/mir.rs
+++ b/crates/hir-ty/src/mir.rs
@@ -1,6 +1,6 @@
 //! MIR definitions and implementation
 
-use std::{collections::hash_map::Entry, fmt::Display, iter};
+use std::{fmt::Display, iter};
 
 use base_db::Crate;
 use either::Either;
@@ -8,10 +8,14 @@ use hir_def::{
     FieldId, StaticId, TupleFieldId, UnionId, VariantId,
     hir::{BindingId, Expr, ExprId, Ordering, PatId},
 };
+use intern::{InternedSlice, InternedSliceRef, impl_slice_internable};
 use la_arena::{Arena, ArenaMap, Idx, RawIdx};
 use rustc_ast_ir::Mutability;
 use rustc_hash::FxHashMap;
-use rustc_type_ir::inherent::{GenericArgs as _, IntoKind, Ty as _};
+use rustc_type_ir::{
+    CollectAndApply, GenericTypeVisitable,
+    inherent::{GenericArgs as _, IntoKind, Ty as _},
+};
 use smallvec::{SmallVec, smallvec};
 use stdx::{impl_from, never};
 
@@ -24,6 +28,7 @@ use crate::{
     next_solver::{
         Allocation, AllocationData, DbInterner, ErrorGuaranteed, GenericArgs, ParamEnv,
         StoredAllocation, StoredConst, StoredGenericArgs, StoredTy, Ty, TyKind,
+        impl_stored_interned_slice,
         infer::{InferCtxt, traits::ObligationCause},
         obligation_ctxt::ObligationCtxt,
     },
@@ -145,7 +150,7 @@ impl<'db> Operand {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub enum ProjectionElem<V: PartialEq> {
     Deref,
     Field(Either<FieldId, TupleFieldId>),
@@ -155,7 +160,7 @@ pub enum ProjectionElem<V: PartialEq> {
     ConstantIndex { offset: u64, from_end: bool },
     Subslice { from: u64, to: u64 },
     //Downcast(Option<Symbol>, VariantIdx),
-    OpaqueCast(StoredTy),
+    OpaqueCast(std::convert::Infallible),
 }
 
 impl<V: PartialEq> ProjectionElem<V> {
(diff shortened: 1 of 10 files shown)


ChayimFriedman2:
I have a dream to share MIR with rustc 😅 At which point we won't need to reimplement move tracking ourselves.

I'll try to look at this later.

flodiebold:
Places are interned in the tcx in rustc, so I think we would need something like this then anyway :sweat_smile: 
I'll fix clippy and format later.

ChayimFriedman2:
To fix Clippy again you just need to rebase (a new Clippy is out).

crates/hir-ty/src/mir.rs
```diff
@@ -260,97 +265,106 @@ impl<V: PartialEq> ProjectionElem<V> {
 
 type PlaceElem = ProjectionElem<LocalId>;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub struct ProjectionId(u32);
+impl<W: crate::next_solver::WorldExposer, V: PartialEq> GenericTypeVisitable<W>
+    for ProjectionElem<V>
+{
+    fn generic_visit_with(&self, _: &mut W) {}
```
ChayimFriedman2, line 269: I'm a bit worried about this non-derived impl, especially with the generic `V` (although IIRC it's only used with one type and we should remove it), given that an incorrect `GenericTypeVisitable` impl can be a soundness problem. Not blocking though.

crates/hir-ty/src/mir.rs
```diff
@@ -260,97 +265,106 @@ impl<V: PartialEq> ProjectionElem<V> {
...
     }
-}
-
-impl ProjectionId {
-    pub const EMPTY: ProjectionId = ProjectionId(0);
 
-    pub fn is_empty(self) -> bool {
-        self == ProjectionId::EMPTY
+    pub fn project(self, projection: PlaceElem) -> Projection<'db> {
+        let mut current = self.as_slice().to_vec();
```
ChayimFriedman2: Reserve capacity for +1? Currently it'll always allocate and copy twice.

ChayimFriedman2: Alternatively use iterators and `chain()`, that may be preferred (non-allocating until some length, one allocation after).

crates/hir-ty/src/mir.rs
```diff
@@ -260,97 +265,106 @@ impl<V: PartialEq> ProjectionElem<V> {
...
-            Entry::Vacant(e) => {
-                let key_clone = e.key().clone();
-                e.insert(new_id);
-                self.id_to_proj.insert(new_id, key_clone);
-                new_id
-            }
-        }
+    #[inline]
+    pub fn as_slice(self) -> &'db [PlaceElem] {
+        &self.interned.get().slice
```
ChayimFriedman2, line 302: Consider implementing `Deref<Target = [T]>` (although `as_slice()` is also useful because `Deref` can only give you a short-lived reference).

crates/hir-ty/src/mir.rs
```diff
@@ -260,97 +265,106 @@ impl<V: PartialEq> ProjectionElem<V> {
...
+        Place { local: self.local, projection: self.projection.store() }
     }
 }
 
-impl From<LocalId> for Place {
+impl<'db> From<LocalId> for PlaceRef<'db> {
     fn from(local: LocalId) -> Self {
-        Self { local, projection: ProjectionId::EMPTY }
+        let empty: &[PlaceElem] = &[];
+        PlaceRef { local, projection: Projection::new_from_slice(empty) }
```
ChayimFriedman2, line 375: Should have an empty projection list in `DefaultAny`, like other lists.

crates/ide/src/inlay_hints/implicit_drop.rs
```diff
@@ -43,8 +43,8 @@ pub(super) fn hints(
 
     for (_, bb) in mir.basic_blocks.iter() {
         let terminator = bb.terminator.as_ref()?;
-        if let TerminatorKind::Drop { place, .. } = terminator.kind {
-            if !place.projection.is_empty() {
+        if let TerminatorKind::Drop { place, .. } = &terminator.kind {
+            if !place.projection.lookup().is_empty() {
```
ChayimFriedman2: Can have an `is_empty()` method for `StoredProjection` to remove the need for that. To your judgement.

crates/hir-ty/src/mir/lower.rs
```diff
@@ -2267,18 +2278,17 @@ pub fn mir_body_for_closure_query<'db>(
...
-                    }
-                    result_projections
-                        .extend(mir_projections[skip_projections_up_to..].iter().cloned());
-                    mir_place.projection = store.intern(result_projections.into());
+                    let maybe_deref: &[PlaceElem] =
+                        if hir_place.is_by_ref() { &[ProjectionElem::Deref] } else { &[] };
+                    mir_place.projection = Projection::new_from_iter(
+                        maybe_deref
+                            .iter()
+                            .cloned()
```
ChayimFriedman2:
```suggestion
                            .copied()
```

crates/hir-ty/src/mir/lower.rs
```diff
@@ -2267,18 +2278,17 @@ pub fn mir_body_for_closure_query<'db>(
...
+                            .chain(mir_projections[skip_projections_up_to..].iter().cloned()),
```
ChayimFriedman2:
```suggestion
                            .chain(mir_projections[skip_projections_up_to..].iter().copied()),
```

ChayimFriedman2 (approved):
LGTM minus a few nits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22474/c77651b22377260ba6412aa7e47452cb30a534e1..3c428a713da36525bfcc014d9feedf4127209f06)*

crates/hir-ty/src/mir.rs
flodiebold: Good point, I've switched the impl to only be for PlaceElem for now

crates/ide/src/inlay_hints/implicit_drop.rs
flodiebold: Sure, why not

crates/hir-ty/src/mir.rs
flodiebold: Added it, can't access it from here though, can I?

ChayimFriedman2: We should have an `empty()` method and a `Default` impl that conjures an interner.

PR #22483, opened 2026-05-28
fix: Do not assume array destructuring assignments with rest pattern are constant-sized

ChayimFriedman2:
We used to set a `slice: None`.

### DIFF
--- a/crates/hir-def/src/expr_store/lower.rs
+++ b/crates/hir-def/src/expr_store/lower.rs
@@ -1844,12 +1844,20 @@ impl<'db> ExprCollector<'db> {
                 }
 
                 let mut elements = e.exprs();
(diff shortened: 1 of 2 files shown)


studyingegret:
Relevant: https://users.rust-lang.org/t/why-can-the-following-code-compile-and-run-successfully-but-the-vs-code-rust-analyzer-extension-reports-an-error/141310

I think `[.., y] = [1, 2]` should be [test](https://github.com/rust-lang/rust-analyzer/pull/22483/changes#diff-586be5a4397030782b4fb65ede70874a44423207af23438be77a901e844d74d5)ed too?

ChayimFriedman2:
The current test is fine.

Veykril (approved):

PR #22485, opened 2026-05-29
remove clone subtree from ide-ssr

bit-aloo:

### DIFF
--- a/crates/ide-ssr/src/fragments.rs
+++ b/crates/ide-ssr/src/fragments.rs
@@ -6,7 +6,7 @@
 //! needs to determine it somehow. We do this in a stupid way -- by pasting SSR
 //! rule into different contexts and checking what works.
 
-use syntax::{AstNode, SyntaxNode, ast};
+use syntax::{AstNode, SyntaxNode, ast, syntax_editor::SyntaxEditor};
 
 pub(crate) fn ty(s: &str) -> Result<SyntaxNode, ()> {
     fragment::<ast::Type>("type T = {};", s)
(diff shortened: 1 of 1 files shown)


bit-aloo:
For some reason, I don’t like how we handle template replacement in SSR. We could have done this cleanly using just make. The only reason we construct a string and then parse it back is because make cannot work directly with a string. All of this should instead be built from AST bits and pieces rather than relying on string-parsing shenanigans.

crates/ide-ssr/src/fragments.rs
```diff
@@ -53,8 +55,9 @@ fn fragment<T: AstNode>(template: &str, s: &str) -> Result<SyntaxNode, ()> {
         return Err(());
     }
     let node = parse.tree().syntax().descendants().find_map(T::cast).ok_or(())?;
-    if node.syntax().text() != s {
+    let (_, node) = SyntaxEditor::new(node.syntax().clone());
```
ChayimFriedman2, line 58: Why the editor here? We don't do any editing.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22485/d1a54ad4f5e71d80b4e67966ef6294ff8eee09b6..48b851f637be467e71cf1273b641d059009f3144)*

bit-aloo: We don't, but we do need the detached node. I want to localize any syntax tree mutations through the editor, including getting deep detached clones obtained via clone_subtree. Thats why we use the editor API, we can come with something better for sure.

ChayimFriedman2: Why do we need the detached node? Ideally by migrating everything to the editor we'll never need to worry about that.

bit-aloo: In this case, I mechanically translated the old sparse uses of clone_subtree to the current variant. To remove it completely, I think we would likely need to make a few additional changes around how SSR interprets the syntax tree.

I also believe we have several places, both in assists and elsewhere, where we currently initialize it specifically to obtain a detached deep clone of a subtree. I'm not entirely sure about the reasoning behind all of those usages, so they may be worth revisiting as well. I am considering of tackling them together making any necessary changes to the editor API.

ChayimFriedman2 (approved):

PR #22486, opened 2026-05-29
fix: don't complete `.await` on receivers of unknown type

### LINKED ISSUE #22478: `.await` completion shown in a non-async function when the receiver is of unknown type
```rust
//- minicore: future
fn foo(t: T) {
    let _ = t.$0;
}
```
`t` is of type `T` which isn't defined anywhere, which in turn causes a completion `t.await` to appear. If we define `T`, e.g. by adding it as a generic to `foo`, then the completion is (correctly) no longer shown.

**rust-analyzer version**: rust-analyzer 1.95.0 (5980761 2026-04-14)

**rustc version**: rustc 1.95.0 (59807616e 2026-04-14)

**editor or extension**: VSCode, extension version 0.3.2913

**relevant settings**: none

@rustbot claim

### LINKED ISSUE #22487: `callable` token modifier emitted on locals of unknown type
```rust
//- minicore: fn
fn foo() {
    let a: T;
}
```
Here, `a` is of type `T` which isn't defined anywhere, which in turn causes a `callable` token modifier to be added to `a`. If we define `T`, e.g. by adding it as a generic to `foo`, then the token modifier is (correctly) no longer added.

Closely related to https://github.com/rust-lang/rust-analyzer/issues/22478 -- in fact, the fix is the same, see https://github.com/rust-lang/rust-analyzer/pull/22486.

**rust-analyzer version**: rust-analyzer 1.95.0 (5980761 2026-04-14)

**rustc version**: rustc 1.95.0 (59807616e 2026-04-14)

**editor or extension**: VSCode, extension version 0.3.2913

**relevant settings**: none

@rustbot claim

ada4a:
Fixes rust-lang/rust-analyzer#22478

This also happens to improve syntax highlighting in presence of error types, and therefore fixes rust-lang/rust-analyzer#22487


crates/hir-ty/src/traits.rs
```diff
@@ -121,6 +121,9 @@ pub fn implements_trait_unique<'db>(
     env: ParamEnvAndCrate<'db>,
     trait_: TraitId,
 ) -> bool {
+    if ty.is_ty_error() {
```
ChayimFriedman2: This should come in `implements_trait_unique_with_infcx()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22486/5ebf65ce47fc047e8a2772297f2ddd988ab86c32..21a8536c34e0701295870513d8117fdc02018a72)*

crates/hir/src/lib.rs
```diff
@@ -5700,9 +5700,7 @@ impl<'db> Type<'db> {
     pub fn into_future_output(&self, db: &'db dyn HirDatabase) -> Option<Type<'db>> {
         let env = self.param_env(db);
         let lang_items = hir_def::lang_item::lang_items(db, env.krate);
-        let (trait_, output_assoc_type) = lang_items
-            .IntoFuture
-            .zip(lang_items.IntoFutureOutput)
+        let (trait_, output_assoc_type) = (lang_items.IntoFuture.zip(lang_items.IntoFutureOutput))
```
ChayimFriedman2: I'm not sure why rustc doesn't warn on the parentheses, but please revert this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22486/5ebf65ce47fc047e8a2772297f2ddd988ab86c32..21a8536c34e0701295870513d8117fdc02018a72)*

crates/ide-completion/src/completions/dot.rs
```diff
@@ -1807,4 +1813,23 @@ fn main() {
             "#]],
         );
     }
+
+    #[track_caller]
+    fn check_await(#[rust_analyzer::rust_fixture] ra_fixture: &'static str, expect: Expect) {
```
ChayimFriedman2: I see no reason to have a separate helper for this. We don't have a helper for any completion kind. Just use an existing helper.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22486/5ebf65ce47fc047e8a2772297f2ddd988ab86c32..21a8536c34e0701295870513d8117fdc02018a72)*

ChayimFriedman2 (dismissed):

crates/hir/src/lib.rs
ada4a: > I'm not sure why rustc doesn't warn on the parentheses

Because they aren't meaningless in this case. My motivation for this change was that it imo makes it easier to see what this expression is saying: "(`IntoFuture` zipped with its output) or (`Future` zipped with its output)". But this is not worth arguing over, I'll revert..

crates/ide-completion/src/completions/dot.rs
```diff
@@ -1807,4 +1807,32 @@ fn main() {
             "#]],
         );
     }
+
+    #[test]
+    fn no_await_on_error_type() {
```
Veykril: lets move this test into `crates\ide-completion\src\tests\expression.rs` (at one point I tried getting rid of the unit tests in the `completions` directory, grouping them by "location" instead, but I never managed to finish that.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22486/55af1774e9e840d3b28d12fcfaa72d2e2caa8ac9..0f3c0ec73d0b9456098fa5002ab6d473e48b808d)*

Veykril (approved):

crates/hir-ty/src/traits.rs
ada4a: This particular bug happens because the self type of the trait ref is an error type -- but in the general case, I think we should check _all_ the types that come up in the trait ref, right? I went with the latter approach, let me know if you think the former would've been sufficient

ada4a:
I was actually waiting for a response to https://github.com/rust-lang/rust-analyzer/pull/22486#discussion_r3427879709, but I can do that as a separate PR if needed

Veykril:
ah my bad. But yea we can address that if Chayim finds t he time look over that.

crates/hir-ty/src/traits.rs
ChayimFriedman2: Actually, no, you should do the `Self` type only. Other args can be too restrictive and is also handled by https://github.com/rust-lang/rust/pull/150453.

ada4a: Done in https://github.com/rust-lang/rust-analyzer/pull/22617

PR #22488, opened 2026-05-29
fix: no generate with default assoc item

A4-Tacks:
- And remove bounds for associated type

Example
---
```rust
trait $0Foo {
    type Item: ?Sized;

    const N: usize = 3;
}
```

**Before this PR**

```rust
impl<T: ?Sized> Foo for $0T {
    type Item: ?Sized;

    const N: usize = 3;
}
```

**After this PR**

```rust
impl<T: ?Sized> Foo for $0T {
    type Item;
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs
+++ b/crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs
@@ -15,7 +15,7 @@ use syntax::{
         self, AssocItem, GenericParam, HasAttrs, HasGenericParams, HasName, HasTypeBounds,
         HasVisibility, edit::AstNodeEdit, syntax_factory::SyntaxFactory,
     },
-    syntax_editor::Position,
+    syntax_editor::{Position, Removable},
 };
 
 // Assist: generate_blanket_trait_impl
@@ -99,10 +99,15 @@ pub(crate) fn generate_blanket_trait_impl(
                             ast::AssocItem::Fn(method) if method.body().is_none() => {
                                 todo_fn(make, &method, ctx.config).into()
                             }
-                            ast::AssocItem::Const(_) | ast::AssocItem::TypeAlias(_) => item,
+                            ast::AssocItem::Const(it) if it.body().is_none() => {
+                                ast::AssocItem::Const(it.reset_indent())
+                            }
+                            ast::AssocItem::TypeAlias(it) if it.ty().is_none() => {
+                                remove_bounds(it.reset_indent()).into()
+                            }
                             _ => return None,
                         };
-                        Some(item.reset_indent().indent(1.into()))
+                        Some(item.indent(1.into()))
                     })
                     .collect::<Vec<_>>();
                 (!items.is_empty()).then(|| make.assoc_item_list(items))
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs
```diff
@@ -284,6 +289,16 @@ fn todo_fn(make: &SyntaxFactory, f: &ast::Fn, config: &AssistConfig) -> ast::Fn
     )
 }
 
+fn remove_bounds(alias: ast::TypeAlias) -> ast::TypeAlias {
+    let (editor, alias) = syntax::syntax_editor::SyntaxEditor::with_ast_node(&alias);
```
ChayimFriedman2, line 293: Just noting that this will not work with lifetime'd nodes.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22488/5ebf65ce47fc047e8a2772297f2ddd988ab86c32..2381dd5a6c437bbef216d2b2f13c463f75a96e10)*

crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs
```diff
@@ -775,7 +817,7 @@ mod foo {
 mod foo {
     mod bar {
         trait $0Foo {
-            type Item: Bar<
+            type Item where Self: Bar<
```
ChayimFriedman2, line 820: I'm not sure that we should not remove `where` clauses on assoc types, while they can also be present on impls many times they are not needed.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22488/5ebf65ce47fc047e8a2772297f2ddd988ab86c32..2381dd5a6c437bbef216d2b2f13c463f75a96e10)*

ChayimFriedman2 (approved):

crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs
A4-Tacks: I don't understand

`type X: 'static;` -> `type X;`
`type X<'a>` -> `type X<'a>;`

It looks great

ChayimFriedman2: I meant when we rewrite `rowan` and syntax nodes will borrow from the tree, then you won't be able to return a node created from a local `SyntaxEditor`.

crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs
A4-Tacks: Users can easily remove them, but adding them requires copying and pasting from the definition

crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs
A4-Tacks: After rewriting, it should be possible to implement 'return a SyntaxEditor', right

crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs
ChayimFriedman2: Makes sense.

PR #22489, opened 2026-05-29
fix: no complete keyword colons before colons

A4-Tacks:
Related rust-lang/rust-analyzer#22479

Example
---
```rust
fn foo() { $0::bar }
```

**Before this PR**

```rust
fn foo() { self::::bar }
```

**After this PR**

```rust
fn foo() { self::bar }
```

### DIFF
--- a/crates/ide-completion/src/completions.rs
+++ b/crates/ide-completion/src/completions.rs
@@ -97,11 +97,30 @@ impl Completions {
         item.add_to(self, ctx.db);
     }
 
+    pub(crate) fn add_keyword_with_colon(
+        &mut self,
+        ctx: &CompletionContext<'_, '_>,
+        keyword: &'static str,
+    ) {
+        debug_assert!(!keyword.ends_with("::"));
+        let label = syntax::format_smolstr!("{keyword}::");
+        let mut item = CompletionItem::new(
+            CompletionItemKind::Keyword,
+            ctx.source_range(),
+            label.clone(),
+            ctx.edition,
+        );
+        if ctx.token.next_token().is_some_and(|it| it.kind() == syntax::T![::]) {
+            item.insert_text(keyword);
+        }
+        item.add_to(self, ctx.db);
+    }
+
     pub(crate) fn add_nameref_keywords_with_colon(&mut self, ctx: &CompletionContext<'_, '_>) {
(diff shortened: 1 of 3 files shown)


ChayimFriedman2 (changes requested):
I think the label should also not include `::`.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22489/5ebf65ce47fc047e8a2772297f2ddd988ab86c32..7c940246d2712a875abfbb966664d4dcedf7dd50)*

A4-Tacks:
> I think the label should also not include `::`.

This will be difficult to distinguish between `self` (local) and `self::` (module)

ChayimFriedman2:
Given they complete exactly the same, I don't think we should show both.

A4-Tacks:
Users can choose whether to complete the colons by selecting the completion item, This feature should not be compromised

ChayimFriedman2:
What I mean is, when we do not complete colons we should not show both.

A4-Tacks:
Cases that do not complete the colons seem to only occur before the colons

This should be irrelevant

ChayimFriedman2:
There, or when the user opted out in the settings. In both cases it's confusing that colons are shown but not completed.

A4-Tacks:
> There

This is irrelevant because colons is exists

> or when the user opted out in the settings. In both cases it's confusing that colons are shown but not completed.

Colons shown and completed

My consideration for this is:
When users want to completely return to their old behavior by changing settings, not completing `::` will break, but perhaps this is a good thing?

A4-Tacks:
```
fn main() {
    s$0
    s$0::bar
}
```
Old: `self::` `self::`
New: `self::` `self`
New `addColonsToModule: false`: `self::` `self`
New `addColonsToModule: false`: `self` `self` (perhaps this is better?)

ChayimFriedman2:
Your table of behaviors is correct. The only thing I said is that when we don't complete the colons, we should not show them in the label either, and that when we do that and also have a `self` completion, we should not show both since it's the same completion.

A4-Tacks:
> The only thing I said is that when we don't complete the colons, we should not show them in the label either,

This will make it difficult to distinguish between local and module in both user and testing scenarios

> and that when we do that and also have a `self` completion, we should not show both since it's the same completion.

This is difficult to implement and has little impact

Although the completion content is the same, `CompletionRelevance` is different

PR #22492, opened 2026-05-30
Better release builds with more debug info

Veykril:


Cargo.toml
```diff
@@ -27,8 +27,9 @@ dissimilar.opt-level = 3
 miniz_oxide.opt-level = 3
 
 [profile.release]
-# Set this to 1 or 2 to get more useful backtraces in debugger.
-debug = 0
+debug = "limited"
+lto = "thin"
+codegen-units = 1
```
Veykril: If one wants to have a fast release build they can just use the dev-rel profile

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22492/5ebf65ce47fc047e8a2772297f2ddd988ab86c32..01551d2b91f7c29deea84bd824bb6a50423d38d6)*

Cargo.toml
```diff
@@ -27,8 +27,9 @@ dissimilar.opt-level = 3
 miniz_oxide.opt-level = 3
 
 [profile.release]
-# Set this to 1 or 2 to get more useful backtraces in debugger.
-debug = 0
+debug = "limited"
```
Veykril: I want to see line info in my profiles when investigating r-a slowness, this makes this easier for me

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22492/5ebf65ce47fc047e8a2772297f2ddd988ab86c32..01551d2b91f7c29deea84bd824bb6a50423d38d6)*

lnicola:
How much larger are these? We could also put it in `xtask dist`, no strong opinion. Might be pretty slow.

Veykril:
Good point, that might be better to put there

xtask/src/dist.rs
```diff
@@ -106,13 +106,12 @@ fn dist_server(
...
 ) -> anyhow::Result<()> {
     let _e = sh.push_env("CFG_RELEASE", release);
+    let _e = sh.push_env("CARGO_PROFILE_RELEASE_DEBUG", "limited");
     let _e = sh.push_env("CARGO_PROFILE_RELEASE_LTO", "thin");
+    let _e = sh.push_env("CARGO_PROFILE_RELEASE_CODEGEN_UNITS", "1");
+    let _e = sh.push_env("CARGO_PROFILE_DEV_REL_DEBUG", "limited");
     let _e = sh.push_env("CARGO_PROFILE_DEV_REL_LTO", "thin");
-
-    // Uncomment to enable debug info for releases. Note that:
-    //   * debug info is split on windows and macs, so it does nothing for those platforms,
```
Veykril, line 113: windows we ship the pdb files, im not sure about the dsyms on macos

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22492/5ebf65ce47fc047e8a2772297f2ddd988ab86c32..37f92c724f109cfa95112c1c31a5217cc8086e3c)*

xtask/src/dist.rs
```diff
@@ -106,13 +106,12 @@ fn dist_server(
...
-    //   * on Linux, this blows up the binary size from 8MB to 43MB, which is unreasonable.
```
Veykril, line 114: presumably `limited` should not blow up as much?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22492/5ebf65ce47fc047e8a2772297f2ddd988ab86c32..37f92c724f109cfa95112c1c31a5217cc8086e3c)*

ChayimFriedman2:
By how much this increased the extensions size?

Veykril:
comparing the last stable release with the last nightly, seems most went from 20mb -> 50-60mb :/ thats a lot more than I'd hoped, so yea we probably want to revert that (except for lto imo)

lnicola:
Oh, wow, that's from 17 to 50 MB compressed.

lnicola:
On the other hand, my toolchain has a 41 MB `cargo` and a 22 MB `clippy-driver`, both with debug info. `librustc_driver` doesn't seem to have debug info. :shrug:, but drives are pretty expensive these days.

ChayimFriedman2:
The difference for me is that I update my Rust installation once a while and mostly per 6 weeks, but r-a I update almost daily.

lnicola:
It's 50 MB, how long could it take to download, 2 seconds? 😄

PR #22495, opened 2026-05-30
Distribute windows builts with mimalloc 

Veykril:
It is vastly faster than the system allocator for our workload on windows. We tend to allocate a lot of small objects on differing threads which causes the system allocator lock to contend and block a lot.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,36 +29,44 @@ jobs:
           - os: windows-latest
             target: x86_64-pc-windows-msvc
             code-target: win32-x64
+            allocator: mimalloc
             pgo: clap-rs/clap@v4.5.36
           - os: windows-latest
             target: i686-pc-windows-msvc
+            allocator: mimalloc
             pgo: clap-rs/clap@v4.5.36
           - os: windows-latest
             target: aarch64-pc-windows-msvc
             code-target: win32-arm64
+            allocator: mimalloc
           - os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


xtask/src/flags.rs
```diff
@@ -309,7 +309,7 @@ impl Install {
         if (self.client || self.proc_macro_server) && !self.server {
             return None;
         }
-        let malloc = if self.mimalloc {
+        let malloc = if self.mimalloc || cfg!(windows) {
```
ChayimFriedman2: I don't think this is the place to do that. Instead we should change how we build our distributed server.

ChayimFriedman2 (commented):
I agree with the goal but I think this PR implements it wrong.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22495/17957d9d6f9fc6ccede4f8537a3bba59e5573e3c..ad69fa2c27888e217b9e6b4be81bd3acd6a7efb5)*

xtask/src/flags.rs
Veykril: Then I need to annoyingly remember to pass the flag every time I cargo xtask install. Especially with us not shipping debug info I will likely run this way more often now than not as I'll likely depend on self-built servers more

ChayimFriedman2: We can make it the default, but not allowing to build without mimalloc seems excessive.

PR #22496, opened 2026-05-30
Re-enable CI cache

Kobzol:
Proposed in https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Moving.20off.20merge.20queue.20to.20new.20bors.3F/with/598734807.

It should now actually work, because CI was switched to run also on the `master` branch. The `coverage` workflow seemed to be quite slow (~5 minutes), so I also enabled the `rust-cache` action for it; the workflow was already running on pushes to `master`.

I removed two commented (previous) usages of `rust-cache`, because the corresponding jobs seem to be quite fast these days (<1m).

I pinned the version to https://github.com/Swatinem/rust-cache/commit/c19371144df3bb44fab255c43d04cbc2ab54d1c4, which is the latest released version of [rust-cache](https://github.com/Swatinem/rust-cache/releases/tag/v2.9.1) as of today, `2.9.1`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,6 +5,9 @@ name: CI
 on:
   pull_request:
   merge_group:
+  push:
+    branches:
+      - master
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
What is the expected speed increase? If it doesn't turn CI into "immediate", and there aren't other reasons to have a cache (save compute for the Project etc.), I don't think this is worth it. Having to deal with caches in CI can be not fun, and 10 minutes is not a lot (r-l/r CI is some hours - *that* is too long).

Kobzol:
Lukas wrote that CI times are atrocious, so I suppose that it depends on the perceived speed :laughing: My goal would be to get all workflows, say, under 2 minutes, so that there isn't any large bottleneck standing out. From my experience, the difference in CI time between 2 minutes and 6-8 minutes is quite large. Not sure if we can get down to 2 minutes though.

ChayimFriedman2:
Maybe it depends on the workflow. I do not depend on CI usually, I run on checks locally (test+Clippy, of course not on all platforms but it's usually enough), unless I have a strong reason to believe I'm doing something simple that will succeed), then publish the PR and turn around to do something else. After some time I check back on the PR.

Veykril (approved):

Veykril:
> Having to deal with caches in CI can be not fun,

Why is that?

Kobzol:
FWIW, as per [#t-compiler/rust-analyzer > Moving off merge queue to new bors? @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Moving.20off.20merge.20queue.20to.20new.20bors.3F/near/598800573), I wasn't actually able to get this to make CI much faster, because compiling local dependencies in the workspace takes most of the time.

Veykril:
ah, unfortunate

ChayimFriedman2:
> > Having to deal with caches in CI can be not fun,
> 
> Why is that?

They tend to cause CI failures from time to time, then you need to debug them or at least clean them.

PR #22500, opened 2026-05-31
Always use crates from sysroot in proc-macro-srv

bjorn3:
Unlike the rest of rust-analyzer it only compiles on nightly anyway. And this makes it easier to add dependencies on other rustc crates like rustc_metadata for parsing crate metadata in the future.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,7 +56,7 @@ jobs:
       - name: Install Rust toolchain
         run: |
           RUSTC_VERSION=$(cat rust-version)
-          rustup-toolchain-install-master ${RUSTC_VERSION} -c cargo -c rust-src -c rustfmt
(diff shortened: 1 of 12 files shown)


bjorn3:
By the way I noticed that the proc-macro-srv tests are broken unless the span/salsa feature is enabled.

Veykril:
> By the way I noticed that the proc-macro-srv tests are broken unless the span/salsa feature is enabled.

Do we not run these on ci? 😱

Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22500/ce1a27e86789d6c3aa626c258e32f71c0dcd2dd0..2ce446852be151fd9ade57524436f82cd6fbd954)*

bjorn3:
In the configuration we run these tests, feature unification enables the salsa feature.

PR #22501, opened 2026-05-31
Fix proc-macro-srv CI job depending on salsa

Veykril:
cc https://github.com/rust-lang/rust-analyzer/pull/22500


lnicola:
This broke `cargo check -p syntax-bridge` (i.e. auto-publishing).

PR #22505, opened 2026-06-01
Adopt uv's AI policy

Veykril:


AI_POLICY.md
```diff
@@ -0,0 +1,36 @@
...
+If you are opening an issue, we expect you to describe the problem in your own
+words.
+
+If you are opening a pull request, we expect you to be able to explain the
+proposed changes in your own words. This includes the pull request body and
+responses to questions. **Do not copy responses from the AI when replying to
+questions from maintainers.**
+
+Due to the foundational nature of our projects, we require a human in the loop
+who understands the work produced by AI. **We do not allow autonomous agents to
```
lnicola: I think something like this would be better, for cases like "I asked Claude to fix a bug, then reviewed the changes and filed a PR myself", where Claude still works autonomously. But it's probably not worth deviating from the original policy, and it can mostly be inferred from context.

> **We do not allow autonomous agents to be used to open pull requests or issues to our projects**

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22505/d1a54ad4f5e71d80b4e67966ef6294ff8eee09b6..29ed84c50b79ff871ed6d45c6fade7222be02816)*

Veykril: Yea I personally already read it with that meaning, but we can clarify this too imo.

AI_POLICY.md
```diff
@@ -0,0 +1,28 @@
+We support using AI (i.e., LLMs) as tools for coding.
```
flodiebold: To me, this sounds too much like encouraging the use of LLMs in general, so I would prefer something like

```suggestion
We allow using AI (i.e., LLMs) as tools for contributing to rust-analyzer.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22505/2eac062855a9606829da1ce9bbe4422c7cb4496b..8c961ad99233834400bdccba71e07289e9fb6e41)*

ChayimFriedman2 (approved):

CONTRIBUTING.md
```diff
@@ -34,10 +34,8 @@ it out of scope (be it due to generally not fitting in with rust-analyzer, or ju
 maintenance capacity). If there already is a feature issue open but it is not clear whether it is
 considered accepted feel free to just drop a comment and ask!
 
-## Use of AI tools
+## Use of AI
 
-AI tool use is not discouraged on the rust-analyzer codebase, as long as it meets our quality standards.
-We kindly ask you to disclose usage of AI tools in your contributions.
```
Wilfred, line 40: Oh, I hadn't noticed this change: are we no longer asking people to disclose whether they've used AI tools?

ChayimFriedman2: Huh, I don't think that was the meaning (it certainly wasn't *my* meaning).

Veykril: Hmm, I could swear this was implied in the new text *somewhere* but now that I am rereading its not?

Issue #15626, opened 2023-09-17
"ra@gennew" appearing in inline suggestion


CarbonSmasher:
**rust-analyzer version**: 0.3.1657-standalone

**rustc version**: 1.71.1

**relevant settings**: none

Steps to reproduce:
1. Open a rust file
2. Create a new function
3. Type `for foo.`

Example:
```
fn bar() {
    for foo.
}
```
Rust analyzer will suggest the text `<ra@gennew>[some number]` as a property
Expected behavior: No suggestion is made
Actual behavior: The text is suggested

HKalbasi:
Duplicate of #14986 but I will close that one since this one is more detailed.

Veykril:
We probably should for now, the proper fix is to probably have desugaring hygiene like rustc and then once we have hygienic name resolution they'd be filtered out on their own.

Paladynee:
any updates on this? it's been bugging me ever since i started using rust analyzer.

Veykril:
Do you have an example snippet at hand that triggers this? The one in the issue does not seem to trigger this behavior anymore

roife:
```rust
fn bar() {
    for i in 0..10 {
        for j in 0..10 {
            r| // cursor
        }
    }
}
```

<img width="596" alt="Image" src="https://github.com/user-attachments/assets/867d036e-b675-4c66-8a25-c0cb54a37cdc" />

bjorn3:
I got this again in an axum project with `rust-analyzer version: 0.4.2796-standalone [/home/bjorn/.vscode/extensions/rust-lang.rust-analyzer-0.4.2796-linux-x64/server/rust-analyzer]`.

```rust
.route("/subscriptions/{id}", get(subscription::get).put(subscription::edit).delete(subscription::delete(<ra@gennew>4, <ra@gennew>5, <ra@gennew>6)))
```

ibg101:
any updates on this? i always get this weird `<ra@gennew>` suggestion

ChayimFriedman2:
We should filter them on `Scope::process_names()` IMO.

PR #22507, opened 2026-06-01
fix: no hint ra@gennew param

A4-Tacks:
Example
---
**Before this PR**

```rust
async fn test_async(foo: i32, _: i32) {}
fn main() {
    test_async(1, 2)
             //^ foo
                //^ <ra@gennew>0
```

**After this PR**

```rust
async fn test_async(foo: i32, _: i32) {}
fn main() {
    test_async(1, 2)
             //^ foo
```

### DIFF
--- a/crates/ide/src/inlay_hints/param_name.rs
+++ b/crates/ide/src/inlay_hints/param_name.rs
@@ -7,7 +7,7 @@
 use std::iter::zip;
 
 use either::Either;
-use hir::{EditionedFileId, Semantics};
+use hir::{EditionedFileId, Semantics, name};
 use ide_db::{RootDatabase, famous_defs::FamousDefs};
 
 use stdx::to_lower_snake_case;
@@ -195,15 +195,16 @@ fn should_hide_param_name_hint(
     // - the argument is a qualified constructing or call expression where the qualifier is an ADT
     // - exact argument<->parameter match(ignoring leading and trailing underscore) or
     //   parameter is a prefix/suffix of argument with _ splitting it off
-    // - param starts with `ra_fixture` or `<ra@`
+    // - param starts with `ra_fixture`
     // - param is a well known name in a unary function
+    // - param is generated name
 
     let param_name = param_name.trim_matches('_');
     if param_name.is_empty() {
         return true;
     }
 
-    if param_name.starts_with("ra_fixture") || param_name.starts_with("<ra@") {
+    if param_name.starts_with("ra_fixture") || name::is_generated(param_name) {
         return true;
     }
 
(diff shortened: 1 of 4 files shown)


crates/ide/src/inlay_hints/param_name.rs
```diff
@@ -195,15 +195,15 @@ fn should_hide_param_name_hint(
...
+    // - param starts with `ra_fixture` or `<ra@`
     // - param is a well known name in a unary function
 
     let param_name = param_name.trim_matches('_');
     if param_name.is_empty() {
         return true;
     }
 
-    if param_name.starts_with("ra_fixture") {
+    if param_name.starts_with("ra_fixture") || param_name.starts_with("<ra@") {
```
ChayimFriedman2: We have `Name::is_generated()`, but it's only for `Name`s. However I don't want the knowledge how generated names look like to be here. So instead you can export a function that does the same on `&str` inside hir-expand/src/name.rs (whether as an associated function of `Name` or not), and call it from `Name::is_generated()` and from here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22507/123c1660f4d80443aea20329cfcc5185b1c233ad..d75df44d2fa41da3a7c7c5f0405df2286f146eb1)*

ChayimFriedman2 (changes requested):

crates/ide/src/inlay_hints/param_name.rs
A4-Tacks: I refer to `src/inlay_hints/implicit_drop.rs`:

```rust
if name.starts_with("<ra@") {
    continue; // Ignore desugared variables
}
```

ChayimFriedman2: Well I think we should change that too 😅

ChayimFriedman2 (approved):

stevenlele:
Related:
- #15626
  - apparently not fully fixed yet
- #18835
  - introduced two occurrences of `name.as_str().starts_with('<')` that you may want to refactor

By the way, `<ra@gennew>` still appears in signature hints of `async fn`s with anonymous params. Should probably be a new issue.

A4-Tacks:
I don't want to add a lot of filtering in the IDE, nor do I want to change the desugar code generation

ChayimFriedman2:
The desugaring should not change, it is correct. We might want depending on hygiene instead of generated names, but that is a complex endeavor infeasible with the current codebase and not sure it's worth it, plus it'll still require filtering.

The correct thing to do will be to filter in the `hir` layer, but I'm not sure we can.

PR #22511, opened 2026-06-01
Stabilize optimize attribute

veluca93:


ChayimFriedman2:
That file is way out of date. I have no problems merging this but we need to figure out a way to generate it automatically anyway.

PR #22513, opened 2026-06-02
Fix parser stack overflow

sunny026:


crates/parser/src/grammar/expressions.rs
```diff
@@ -246,6 +246,23 @@ fn expr_bp(
     m: Option<Marker>,
     r: Restrictions,
     bp: u8,
+) -> Option<(CompletedMarker, BlockLike)> {
+    if !p.check_recursion_limit() {
```
ChayimFriedman2, line 250: This is, of course, not enough. Many functions in the parser recurse.

I'm not yet sure whether we want such thing, though.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22513/55af1774e9e840d3b28d12fcfaa72d2e2caa8ac9..61e626a399319a27968b27f92db50e9f6d1e34cb)*

Veykril: if we want to fix this we should do what rustc does which is growing the stack on demand

ChayimFriedman2: And there are also plenty of places besides parsing that will need that (although parsing is the most important).

ChayimFriedman2:
Wait, did you even check that it is parsing that stack overflows on this code?

crates/parser/src/grammar/expressions.rs
Veykril: Yea hir-def lowering does as well at the very least

ChayimFriedman2:
Given that you haven't answered my question:

> Wait, did you even check that it is parsing that stack overflows on this code?

And that the utility of this in general is doubted, I'm closing this.

PR #22519, opened 2026-06-03
fix: no generate unused generic params in trait sign

A4-Tacks:
Example
---
```rust
struct Foo<T, const N: usize>([T; N]);

impl<T, const N: usize> F$0oo<T, N> {
    fn spec_len(&self) -> usize {
        N
    }
}
```

**Before this PR**

```rust
struct Foo<T, const N: usize>([T; N]);

trait SpecLen<T, const N: usize> {
    fn spec_len(&self) -> usize;
}

impl<T, const N: usize> SpecLen<T, N> for Foo<T, N> {
    fn spec_len(&self) -> usize {
        N
    }
}
```

**After this PR**

```rust
struct Foo<T, const N: usize>([T; N]);

trait SpecLen {
    fn spec_len(&self) -> usize;
}

impl<T, const N: usize> SpecLen for Foo<T, N> {
    fn spec_len(&self) -> usize {
        N
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/generate_trait_from_impl.rs
+++ b/crates/ide-assists/src/handlers/generate_trait_from_impl.rs
@@ -1,7 +1,7 @@
 use crate::assist_context::{AssistContext, Assists};
-use ide_db::assists::AssistId;
+use ide_db::{assists::AssistId, defs::Definition, search::SearchScope};
 use syntax::{
-    AstNode, AstToken, SyntaxKind, T,
+    AstNode, AstToken, SyntaxKind, T, TextRange,
     ast::{
         self, HasDocComments, HasGenericParams, HasName, HasVisibility, edit::AstNodeEdit,
         syntax_factory::SyntaxFactory,
@@ -114,10 +114,11 @@ pub(crate) fn generate_trait_from_impl(
 
             let editor = builder.make_editor(impl_ast.syntax());
             let make = editor.make();
+            let params = used_params(&impl_ast, make, ctx);
             let trait_ast = make.trait_(
                 false,
                 &trait_name(&impl_assoc_items, make).text(),
-                impl_ast.generic_param_list(),
+                params.clone(),
                 impl_ast.where_clause(),
                 trait_items,
             );
@@ -133,7 +134,7 @@ pub(crate) fn generate_trait_from_impl(
                 make.whitespace(" ").into(),
             ];
 
-            if let Some(params) = impl_ast.generic_param_list() {
+            if let Some(params) = params {
                 let gen_args = &params.to_generic_args(make);
                 elements.insert(1, gen_args.syntax().clone().into());
             }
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/generate_trait_from_impl.rs
```diff
@@ -167,6 +168,42 @@ pub(crate) fn generate_trait_from_impl(
...
+        .map(item_in_trait_range)
+        .collect::<Vec<_>>();
+    let used_in_impl = |param: &ast::GenericParam| {
+        let Some(def) = ctx.sema.to_def(param) else { return true };
+        Definition::GenericParam(def)
+            .usages(&ctx.sema)
+            .in_scope(&SearchScope::single_file(ctx.file_id()))
+            .all()
+            .file_ranges()
+            .any(|it| trait_ranges.iter().any(|range| range.contains_range(it.range)))
```
ChayimFriedman2:
```suggestion
            .at_least_one()
```
We don't actually need to check whether they're in the range, since they're declared in it.

ChayimFriedman2 (approved):
LGTM minus one nit.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22519/55af1774e9e840d3b28d12fcfaa72d2e2caa8ac9..b5bad9e7fc2ee9909f31674f9ae95a840578cb7d)*

crates/ide-assists/src/handlers/generate_trait_from_impl.rs
A4-Tacks: > We don't actually need to check whether they're in the range, since they're declared in it.

This is necessary because the references in self-type and fn-body need to be ignored

ChayimFriedman2: Right, then references in the body of consts/type aliases should also be ignored.

A4-Tacks: Because const and type exist in both trait and impl after generation, they are retained

It should be up to the user to decide which one to adopt

crates/ide-assists/src/handlers/generate_trait_from_impl.rs
```diff
@@ -167,6 +168,42 @@ pub(crate) fn generate_trait_from_impl(
...
+            fn_item.body().map_or(range, |body| {
+                TextRange::new(range.start(), body.syntax().text_range().start())
+            })
+        }
+    };
+    let trait_ranges = impl_ast
+        .assoc_item_list()
+        .into_iter()
+        .flat_map(|list| list.assoc_items())
+        .map(item_in_trait_range)
```
bit-aloo: Shouldn't we need to check for where clause as well?

Consider something like this: 

```rust
impl<T> Foo<T> where T: Copy { fn foo() {} } 
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22519/55af1774e9e840d3b28d12fcfaa72d2e2caa8ac9..b5bad9e7fc2ee9909f31674f9ae95a840578cb7d)*

ChayimFriedman2: True. I think it's better to have a negative list, i.e. a list of excluded ranges.

A4-Tacks: Good idea

bit-aloo (approved):
LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22519/55af1774e9e840d3b28d12fcfaa72d2e2caa8ac9..832c8c28cef5c92ec8840532e241df2fad4aa575)*

crates/ide-assists/src/handlers/generate_trait_from_impl.rs
ChayimFriedman2: I don't think they should be retained, but, fair enough.

ChayimFriedman2 (approved):

PR #22528, opened 2026-06-04
internal: Improve panic message when FileSourceRootInput is missing

Wilfred:
I'm seeing several occurrences of errors of the form:

```
Unable to get `FileSourceRootInput` with `vfs::FileId` (FileId(22452)); this is a bug
```

I suspect it occurs when the user has many projects open, but I can't reliably reproduce yet. In the meantime, include the actual file name to aid debugging.

AI disclosure: Partially written with Codex and GPT 5.5.

### DIFF
--- a/crates/base-db/src/lib.rs
+++ b/crates/base-db/src/lib.rs
@@ -170,14 +170,30 @@ impl Files {
         };
     }
 
-    pub fn file_source_root(&self, id: vfs::FileId) -> FileSourceRootInput {
+    pub fn file_source_root(
+        &self,
+        db: &dyn SourceDatabase,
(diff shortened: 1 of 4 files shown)


ChayimFriedman2 (approved):
It will be more reliable to get the path from the VFS (if there is a problem with sysroots, the path might not be there), but we don't have an access to it here and it's still an improvement.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22528/aadee11251e6074c9245ac96bc4bfe6eca05a64c..273abbb6b7c62104a6def5032aefc37f36ccbd75)*

PR #22533, opened 2026-06-06
internal: Remove AST types from FoldKind

Veykril:
We have no business exposing the AST types from this, they are solely needed internally

### DIFF
--- a/crates/test-utils/src/fixture.rs
+++ b/crates/test-utils/src/fixture.rs
@@ -399,6 +399,7 @@ impl MiniCore {
     ///
     /// This is probably over-engineered to support flags dependencies.
     pub fn source_code(mut self, raw_source: &str) -> String {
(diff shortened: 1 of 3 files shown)


crates/test-utils/src/fixture.rs
nicolas-guichard, line 1: I think you included this in this commit by mistake.
(Doesn't look wrong, just letting you know in case you didn't notice.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22533/09b33150cb25cd5a42fe3451272f1cce53afa6a0..fbd7af8b34d72e3d5555e418c06fca48ca0c4aeb)*

Veykril: yea that was intentional

Issue #10913, opened 2021-12-03
unit-returning macros in statement position should be completed with a semicolon


matklad:
When completing assert, assert_eq, println and friends in statement context, we should insert `;` automatically.

jhgg:
Is there a good way to programmatically detect this, or should we just hard code a list of known macros?

PR #22536, opened 2026-06-06
fix: add macro semicolon in item position

A4-Tacks:
Example
---
```rust
macro_rules! foo { () => {} }
f$0
```

**Before this PR**

```rust
macro_rules! foo { () => {} }
foo!($0)
```

**After this PR**

```rust
macro_rules! foo { () => {} }
foo!($0);
```

### DIFF
--- a/crates/ide-completion/src/config.rs
+++ b/crates/ide-completion/src/config.rs
@@ -27,6 +27,7 @@ pub struct CompletionConfig<'a> {
     pub callable: Option<CallableSnippets>,
     pub add_colons_to_module: bool,
     pub add_semicolon_to_unit: bool,
+    pub add_semicolon_to_macro: MacroSemicolonStyle,
     pub snippet_cap: Option<SnippetCap>,
     pub insert_use: InsertUseConfig,
     pub prefer_no_std: bool,
@@ -46,6 +47,15 @@ pub enum AutoImportExclusionType {
     Methods,
     SubItems,
     Variants,
+}
+
+// FIXME: Implement 'Auto' mode
+// https://github.com/rust-lang/rust-analyzer/pull/22536#issuecomment-4834287685
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum MacroSemicolonStyle {
+    None,
+    Item,
+    Always,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
--- a/crates/ide-completion/src/context.rs
+++ b/crates/ide-completion/src/context.rs
@@ -886,22 +886,22 @@ impl<'a, 'db> CompletionContext<'a, 'db> {
         );
 
         // FIXME: This should be part of `CompletionAnalysis` / `expand_and_analyze`
-        let complete_semicolon = if !config.add_semicolon_to_unit {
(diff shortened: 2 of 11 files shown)


ChayimFriedman2 (approved):
Should probably be controlled by a setting like `addSemicolonToUnit`. Other than that, LGTM.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22536/a9a66c4077fedd375c0f1e61dc39a8f3243ce0ef..ac149cb24b49dd1cececebb2c84673af0ccbb989)*

A4-Tacks:
I don't think it should be controlled by `addSemicolonToUnit`

1. Macros not result unit
2. Not adding semicolons is always a syntax error

ChayimFriedman2:
Not by this setting, what I meant is that it should have its own setting like that.

A4-Tacks:
Currently, it only supports completion at the item-place

I don't think it's harmful, nor does it break the habit

I think that adding settings may be necessary after supporting completion at stmt-place in the future

ChayimFriedman2:
I'm not sure I understand what you mean. The setting can be on by default (just like `addSemicolonToUnit`), but I do think some people won't like this behavior and will want an opt-out.

A4-Tacks:
There is a difference between functions and macros: `foo()$0` and `foo!($0)`

For empty functions, users may be accustomed to manually adding semicolons after confirmation, but macros do not

I think needs moving the cursor is enough to avoid habitually adding semicolons

Besides, as you mentioned before, there are too many setting options

ChayimFriedman2:
> For empty functions, users may be accustomed to manually adding semicolons after confirmation, but macros do not

Why? I'm pretty sure that's incorrect.

I agree we have too many settings, but when we change behavior users are accustomed to, I feel it's fundamentally different than someone requesting a new setting, and we're more obliged to support it.

A4-Tacks:
> Why? I'm pretty sure that's incorrect.

Although adding semicolons is definitely necessary, I don't think it will form a "habit" or "muscle memory"

ChayimFriedman2:
I do think it will, and it's not different than `addSemicolonToUnit`.

A4-Tacks:
Okay, I hope this is good

ChayimFriedman2:
I said that we shouldn't reuse `addSemicolonToUnit` for this, it should be a new setting.

A4-Tacks:
Like `"addSemicolonToMacro": "item"`?

enumerate "item", "auto" (unimplemented), "always" (similar fn, in next PR), "none"

ChayimFriedman2:
What's `auto`? Why is `always` useful?

`"addSemicolonToMacro": true/false` should be enough.

A4-Tacks:
| code                        | always | auto | item | none |
| ---                         | ---    | ---  | ---  | ---  |
| `fn main() { _ = vec$0 }`   | -      | -    | -    | -    |
| `fn main() { vec$0 }`       | y      | -    | -    | -    |
| `fn main() { assert_eq$0 }` | y      | y    | -    | -    |
| `mod foo { any_macro$0 }`   | y      | y    | y    | -    |

For 'auto', about #10913

PR #22546, opened 2026-06-08
fix: alias subItems to sub_items

A4-Tacks:
Alternative to rust-lang/rust-analyzer#22444

Example
---
```rust
{
        {"path": "std::intrinsics", "type": "subItems"},
    }
}
```

**Before this PR**

```
invalid config value: /completion/autoimport/exclude: data
did not match any variant of untagged enum
AutoImportExclusion;
```

**After this PR**

no errors

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -3003,6 +3003,7 @@ pub enum AutoImportExclusion {
 pub enum AutoImportExclusionType {
     Always,
     Methods,
+    #[serde(alias = "subItems")]
     SubItems,
 }
 


ChayimFriedman2:
This is not breaking and also a tiny change so I don't object, but may I ask: *why* is this so important to you? Do you think the current name is inconsistent with our other settings?

A4-Tacks:
> why is this so important to you?

The user is using schema, but r-a cannot parse it, so this is a bug

Bugs should be fixed

ChayimFriedman2:
If the user is using the *wrong* schema, and r-a cannot parse it, this is very obviously not a bug of r-a.

A4-Tacks:
As an alternative to #22444, there is always one that needs to be merged

ChayimFriedman2:
You still didn't answer my question: what do you consider so bad with the current `sub_items` that you feel must support `subItems`?

A4-Tacks:
Because the current schema is `subItems` and #22444 has not been merged, I have opened an alternative PR

Some plans:

1. Change variant deserialize name to `subItems` (some breaking, but clean)
2. Add variant deserialize alias `subItems` (this PR)
3. Change schema to `sub_items` (#22444)

> what do you consider so bad with the current `sub_items`

Note: Alias allows both `subItems` and `sub_items` to be parsed

ChayimFriedman2:
Only now I understand what the problem is, and that https://github.com/rust-lang/rust-analyzer/pull/22444 does not actually change parsing and only the presented JSON schema.

With that understood, I think I prefer the other PR (which is not a breaking change, now I understand) over this PR, which adds a duplicate spelling for no reason. So I'll close this one and merge the other one.

PR #22548, opened 2026-06-08
internal: Add a lot more database lifetimes

Veykril:


ChayimFriedman2:
To be honest I'm not a fan of all of our IDs gaining lifetimes. I'd much rather a permanent option for disabling the lifetime for both interneds and tracked structs, considering it isn't required for safety.

But, at the very least, I assume we want to wait for @dfireBird's PR to be merged.

Veykril:
> I'd much rather a permanent option for disabling the lifetime for both interneds and tracked structs, considering it isn't required for safety.

Huh? How is that not required for safety? The entire point of the lifetimes is to prevent usage of stale things

Veykril:
Also the entire lifetime escape mechanism was only introduced to ease our transition, it was never meant to be a permanent thing.

flodiebold:
FWIW, I agree with Chayim, the lifetimes clutter everything up for questionable gain.

ChayimFriedman2:
> Huh? How is that not required for safety? The entire point of the lifetimes is to prevent usage of stale things

I might be wrong, but things I've heard and my knowledge of the Salsa codebase imply this is false.

Furthermore you fundamentally cannot rely on the lifetimes for safety (without branding) since you can use the same ID to access a different database.

It can help correctness a bit but I don't think this is worth it.

> Also the entire lifetime escape mechanism was only introduced to ease our transition, it was never meant to be a permanent thing.

I know that, I suggest we keep it (and extend it to tracked structs).

Veykril:
Fair, I do agree its a bit ugly to have the lifetime infect everything. 

> I might be wrong, but things I've heard and my knowledge of the Salsa codebase imply this is false.

> Furthermore you fundamentally cannot rely on the lifetimes for safety (without branding) since you can use the same ID to access a different database.

I do think salsa relies on the lifetime for safety though fwiw (with the intern gc and what not, at this point), though you are right that the mixing of databases might be problematic 🤔 Will look into that actually.

Though yea, maybe adding the attribute for tracked structs would be good either way (the attr should also be marked unsafe for that matter though)

ChayimFriedman2:
So, you were right. I was able to construct a UB example when using the interneds GC.

Given that, I do support this. IMO preventing unsafety and the need for global reasoning is worth the lifetimes.

Veykril:
Is it a small self contained example, curious to see in that case.

> But, at the very least, I assume we want to wait for dfireBird's PR to be merged.

Also naturally will wait on the lifetime PR to land first, rebasing this PR is a lot easier than other PRs on top of this

ChayimFriedman2:
> Is it a small self contained example, curious to see in that case.

```rust
use salsa::Setter;

#[salsa::interned(revisions = 1, no_lifetime)]
struct Interned {
    #[returns(ref)]
    s: String,
}

#[salsa::input(debug)]
struct Input {
    i: String,
}

#[salsa::tracked]
fn foo(db: &dyn salsa::Database, i: Input) -> Interned {
    Interned::new(db, i.i(db))
}

#[test]
fn test() {
    let mut db = salsa::DatabaseImpl::default();
    let input = Input::builder("0".to_owned()).new(&db);
    let foo_input = Input::new(&db, "abc def dsgdf".to_owned());
    let foo_input2 = Input::new(&db, String::new());
    let interned = foo(&db, foo_input);

    input.set_i(&mut db).to("123".to_owned());

    let s = &**interned.s(&db);
    foo(&db, foo_input2);
    dbg!(s);
}
```

crates/hir-ty/src/infer.rs
```diff
@@ -1042,7 +1058,7 @@ impl InferenceResult {
         }
     }
 
-    pub fn method_resolution<'db>(&self, expr: ExprId) -> Option<(FunctionId, GenericArgs<'db>)> {
+    pub fn method_resolution<'a>(&self, expr: ExprId) -> Option<(FunctionId, GenericArgs<'a>)> {
```
ada4a: I _think_ this is identical to:
```suggestion
    pub fn method_resolution(&self, expr: ExprId) -> Option<(FunctionId, GenericArgs<'static>)> {
```
which is imo clearer than introducing another lifetime generic?

ada4a (commented):
Oops, I've been working on implementing what's apparently a subset of this PR^^ Glad to see that my changes seem to match up with what was done here

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22548/4fde9dd9822df453dc6f20149b8e375efbbe66c0..b3a6adacc8855f2dea974a7dbad2e384f871babf)*

crates/hir-ty/src/infer.rs
Veykril: Oh dear, the fact this compiles is not good. This is a pretty severe soundness bug as that lifetime is just straight up wrong

ada4a: Oh... In that case you would probably be unhappy to learn that the same pattern occurs throughout that impl block

Veykril: yea i put a fixme on the actual method that does the bad transmute

ChayimFriedman2: We already talked about that, this is pretty much unavoidable and it only makes GC potentially unsound, which is why we marked it unsafe.

Veykril: Right, that was that thing.

crates/hir-ty/src/next_solver/def_id.rs
```diff
@@ -384,23 +456,172 @@ macro_rules! declare_id_wrapper {
...
+}
+
+impl<'db> inherent::DefId<DbInterner<'db>, SolverDefId<'db>> for CoroutineClosureIdWrapper<'db> {
+    fn as_local(self) -> Option<SolverDefId<'db>> {
+        Some(self.into())
+    }
+    fn is_local(self) -> bool {
+        true
+    }
+}
```
ada4a, line 543: It's imo unfortunate that we add so much code which, by the looks of it, can still be pretty easily generated by a macro -- especially because `InternedOpaqueTyId` is probably going to get a lifetime as well at some point, adding to this mess. I did try to adapt `declare_id_wrapper` to accept lifetimes, but that proved to be difficult -- what do you think about having a similar but separate macro instead?

<details><summary>Patch</summary>
<p>

```patch
Commit ID: b60e24e2897ed644ade475a97ce3a7be275a4dda
Change ID: kpoytytuxxnrkmlzvvqlwtysmlrvquwo
Author   : Ada Alakbarova <ada.alakbarova@ferrous-systems.com> (2026-06-17 14:51:04)
Committer: Ada Alakbarova <ada.alakbarova@ferrous-systems.com> (2026-06-17 14:51:19)

    add `declare_id_wrapper_with_lt`

diff --git a/crates/hir-ty/src/next_solver/def_id.rs b/crates/hir-ty/src/next_solver/def_id.rs
index afffd45f9c..98ea16e2a6 100644
--- a/crates/hir-ty/src/next_solver/def_id.rs
+++ b/crates/hir-ty/src/next_solver/def_id.rs
@@ -454,160 +454,74 @@
     };
 }
 
+macro_rules! declare_id_wrapper_with_lt {
+    ($name:ident, $wraps:ident) => {
+        declare_id_wrapper_with_lt!($name, $wraps, SolverDefId<'db>);
+    };
+
+    ($name:ident, $wraps:ident, $local:ty) => {
+        declare_id_wrapper_with_lt!($name, $wraps, $local, no_try_from);
+
+        impl<'db> TryFrom<SolverDefId<'db>> for $name<'db> {
+            type Error = ();
+
+            #[inline]
+            fn try_from(value: SolverDefId<'db>) -> Result<Self, Self::Error> {
+                match value {
+                    SolverDefId::$wraps(it) => Ok(Self(it)),
+                    _ => Err(()),
+                }
+            }
+        }
+    };
+
+    ($name:ident, $wraps:ident, $local:ty, no_try_from) => {
+        #[derive(Clone, Copy, PartialEq, Eq, Hash)]
+        pub struct $name<'db>(pub $wraps<'db>);
+
+        impl std::fmt::Debug for $name<'_> {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                std::fmt::Debug::fmt(&SolverDefId::from(self.0), f)
+            }
+        }
+
+        impl<'db> From<$name<'db>> for $wraps<'db> {
+            #[inline]
+            fn from(value: $name<'db>) -> $wraps<'db> {
+                value.0
+            }
+        }
+
+        impl<'db> From<$wraps<'db>> for $name<'db> {
+            #[inline]
+            fn from(value: $wraps<'db>) -> $name<'db> {
+                Self(value)
+            }
+        }
+
+        impl<'db> From<$name<'db>> for SolverDefId<'db> {
+            #[inline]
+            fn from(value: $name<'db>) -> SolverDefId<'db> {
+                value.0.into()
+            }
+        }
+
+        impl<'db> inherent::DefId<DbInterner<'db>, $local> for $name<'db> {
+            fn as_local(self) -> Option<$local> {
+                Some(self.into())
+            }
+            fn is_local(self) -> bool {
+                true
+            }
+        }
+    };
+}
+
 declare_id_wrapper!(TraitIdWrapper, TraitId);
 declare_id_wrapper!(TypeAliasIdWrapper, TypeAliasId);
-#[derive(Clone, Copy, PartialEq, Eq, Hash)]
-pub struct ClosureIdWrapper<'db>(pub InternedClosureId<'db>);
-
-impl std::fmt::Debug for ClosureIdWrapper<'_> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        std::fmt::Debug::fmt(&SolverDefId::from(self.0), f)
-    }
-}
-
-impl<'db> From<ClosureIdWrapper<'db>> for InternedClosureId<'db> {
-    #[inline]
-    fn from(value: ClosureIdWrapper<'db>) -> InternedClosureId<'db> {
-        value.0
-    }
-}
-
-impl<'db> From<InternedClosureId<'db>> for ClosureIdWrapper<'db> {
-    #[inline]
-    fn from(value: InternedClosureId<'db>) -> ClosureIdWrapper<'db> {
-        Self(value)
-    }
-}
-
-impl<'db> From<ClosureIdWrapper<'db>> for SolverDefId<'db> {
-    #[inline]
-    fn from(value: ClosureIdWrapper<'db>) -> SolverDefId<'db> {
-        value.0.into()
-    }
-}
-
-impl<'db> TryFrom<SolverDefId<'db>> for ClosureIdWrapper<'db> {
-    type Error = ();
-
-    #[inline]
-    fn try_from(value: SolverDefId<'db>) -> Result<Self, Self::Error> {
-        match value {
-            SolverDefId::InternedClosureId(it) => Ok(Self(it)),
-            _ => Err(()),
-        }
-    }
-}
-
-impl<'db> inherent::DefId<DbInterner<'db>, SolverDefId<'db>> for ClosureIdWrapper<'db> {
-    fn as_local(self) -> Option<SolverDefId<'db>> {
-        Some(self.into())
-    }
-    fn is_local(self) -> bool {
-        true
-    }
-}
-
-#[derive(Clone, Copy, PartialEq, Eq, Hash)]
-pub struct CoroutineIdWrapper<'db>(pub InternedCoroutineId<'db>);
-
-impl std::fmt::Debug for CoroutineIdWrapper<'_> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        std::fmt::Debug::fmt(&SolverDefId::from(self.0), f)
-    }
-}
-
-impl<'db> From<CoroutineIdWrapper<'db>> for InternedCoroutineId<'db> {
-    #[inline]
-    fn from(value: CoroutineIdWrapper<'db>) -> InternedCoroutineId<'db> {
-        value.0
-    }
-}
-
-impl<'db> From<InternedCoroutineId<'db>> for CoroutineIdWrapper<'db> {
-    #[inline]
-    fn from(value: InternedCoroutineId<'db>) -> CoroutineIdWrapper<'db> {
-        Self(value)
-    }
-}
-
-impl<'db> From<CoroutineIdWrapper<'db>> for SolverDefId<'db> {
-    #[inline]
-    fn from(value: CoroutineIdWrapper<'db>) -> SolverDefId<'db> {
-        value.0.into()
-    }
-}
-
-impl<'db> TryFrom<SolverDefId<'db>> for CoroutineIdWrapper<'db> {
-    type Error = ();
-
-    #[inline]
-    fn try_from(value: SolverDefId<'db>) -> Result<Self, Self::Error> {
-        match value {
-            SolverDefId::InternedCoroutineId(it) => Ok(Self(it)),
-            _ => Err(()),
-        }
-    }
-}
-
-impl<'db> inherent::DefId<DbInterner<'db>, SolverDefId<'db>> for CoroutineIdWrapper<'db> {
-    fn as_local(self) -> Option<SolverDefId<'db>> {
-        Some(self.into())
-    }
-    fn is_local(self) -> bool {
-        true
-    }
-}
-
-#[derive(Clone, Copy, PartialEq, Eq, Hash)]
-pub struct CoroutineClosureIdWrapper<'db>(pub InternedCoroutineClosureId<'db>);
-
-impl std::fmt::Debug for CoroutineClosureIdWrapper<'_> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        std::fmt::Debug::fmt(&SolverDefId::from(self.0), f)
-    }
-}
-
-impl<'db> From<CoroutineClosureIdWrapper<'db>> for InternedCoroutineClosureId<'db> {
-    #[inline]
-    fn from(value: CoroutineClosureIdWrapper<'db>) -> InternedCoroutineClosureId<'db> {
-        value.0
-    }
-}
-
-impl<'db> From<InternedCoroutineClosureId<'db>> for CoroutineClosureIdWrapper<'db> {
-    #[inline]
-    fn from(value: InternedCoroutineClosureId<'db>) -> CoroutineClosureIdWrapper<'db> {
-        Self(value)
-    }
-}
-
-impl<'db> From<CoroutineClosureIdWrapper<'db>> for SolverDefId<'db> {
-    #[inline]
-    fn from(value: CoroutineClosureIdWrapper<'db>) -> SolverDefId<'db> {
-        value.0.into()
-    }
-}
-
-impl<'db> TryFrom<SolverDefId<'db>> for CoroutineClosureIdWrapper<'db> {
-    type Error = ();
-
-    #[inline]
-    fn try_from(value: SolverDefId<'db>) -> Result<Self, Self::Error> {
-        match value {
-            SolverDefId::InternedCoroutineClosureId(it) => Ok(Self(it)),
-            _ => Err(()),
-        }
-    }
-}
-
-impl<'db> inherent::DefId<DbInterner<'db>, SolverDefId<'db>> for CoroutineClosureIdWrapper<'db> {
-    fn as_local(self) -> Option<SolverDefId<'db>> {
-        Some(self.into())
-    }
-    fn is_local(self) -> bool {
-        true
-    }
-}
+declare_id_wrapper_with_lt!(ClosureIdWrapper, InternedClosureId);
+declare_id_wrapper_with_lt!(CoroutineIdWrapper, InternedCoroutineId);
+declare_id_wrapper_with_lt!(CoroutineClosureIdWrapper, InternedCoroutineClosureId);
 declare_id_wrapper!(AdtIdWrapper, AdtId);
 declare_id_wrapper!(OpaqueTyIdWrapper, InternedOpaqueTyId, OpaqueTyIdWrapper);
```

</p>
</details> 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22548/5017a50153285a2d9bfa286e6fa96723e997d5b6..a686cebc4f27ed5508e1090f1a0089c77291b4a1)*

Veykril: Yea we should have some macro that allows us to do this again, I just didn't wanna bother until getting this to compile again

PR #22549, opened 2026-06-09
feat: support flyimport exclude variants

A4-Tacks:
Due to rust-lang/rust-analyzer#22385, some large enum may need to be exclude variants, e.g `std::ascii::Char`

Example
---
```json
{
        {"path": "xxx_crate::Foo", "type": "variants"},
    ]
}
```

```rust
enum Foo {
    Variant1,
    Variant2,
}
fn main() {
    V$0
}
```

**Before this PR**

```
ev TupleV(…)                 TupleV(u32)
ev Variant1 (use Foo::Variant1) Variant1
ev Variant2 (use Foo::Variant2) Variant2
bt u32                               u32
```

**After this PR**

```
ev TupleV(…)                 TupleV(u32)
bt u32                               u32
```

### DIFF
--- a/crates/ide-completion/src/config.rs
+++ b/crates/ide-completion/src/config.rs
@@ -45,6 +45,7 @@ pub enum AutoImportExclusionType {
     Always,
     Methods,
     SubItems,
+    Variants,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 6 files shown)


ChayimFriedman2 (approved):
We might want to support the new exclusion types in the attribute as well, e.g. `ascii::Char`'s variants seems like something you never want auto-imported.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22549/4fde9dd9822df453dc6f20149b8e375efbbe66c0..3ffba73c8fb8edd23242e384f429e7cd9deb1772)*

PR #22551, opened 2026-06-09
fix: supports inline variable in macro

A4-Tacks:
- **fix: supports inline variable in macro**
- **fix: do not panic variable def in macro**

Example
---
```rust
macro_rules! m { ($i:expr) => { $i } }
fn f() {
    let xyz = 0;
    m!(xyz$0);
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
macro_rules! m { ($i:expr) => { $i } }
fn f() {
    m!(0);
}
```

---

```rust
macro_rules! i { ($($t:tt)*) => { $($t)* } }
fn f() {
    i!(let xyz = 0;);
    _ = xyz$0;
    _ = xyz;
}
```

**Before this PR**

```
request handler panicked: can't resolve SyntaxNodePtr { kind: LET_STMT, range: 0..9 } with SOURCE_FILE@0..108
```

**After this PR**

```rust
macro_rules! i { ($($t:tt)*) => { $($t)* } }
fn f() {
    i!(let xyz = 0;);
    _ = 0;
    _ = xyz;
}
```

PR #22552, opened 2026-06-09
fix: offer insert explicit type in macro

A4-Tacks:
Supports attribute macro and macro call

Example
---
```rust
#[tokio::main]
async fn main() {
    let $0x = 3;
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
#[tokio::main]
async fn main() {
    let x: i32 = 3;
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/add_explicit_type.rs
+++ b/crates/ide-assists/src/handlers/add_explicit_type.rs
@@ -21,10 +21,18 @@ use crate::{AssistContext, AssistId, Assists};
 // }
 // ```
 pub(crate) fn add_explicit_type(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -> Option<()> {
-    let syntax_node = ctx.find_node_at_offset::<Either<LetStmt, Param>>()?;
+    let syntax_node = ctx.find_node_at_offset_with_descend::<Either<LetStmt, Param>>()?;
     let (ascribed_ty, expr, pat) = if let Either::Left(let_stmt) = syntax_node {
         let cursor_in_range = {
-            let eq_range = let_stmt.eq_token()?.text_range();
+            let let_range = ctx.sema.original_range_opt(let_stmt.syntax())?.range;
+            let first = crate::utils::cover_edit_range(ctx.source_file().syntax(), let_range)
+                .start()
+                .clone();
+            let eq_range = std::iter::successors(Some(first), |it| it.next_sibling_or_token())
+                .take_while(|it| let_range.contains_range(it.text_range()))
+                .find(|it| it.kind() == syntax::T![=])?
+                .text_range();
+
             ctx.offset() < eq_range.start()
         };
         if !cursor_in_range {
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/add_explicit_type.rs
```diff
@@ -21,10 +21,18 @@ use crate::{AssistContext, AssistId, Assists};
...
 pub(crate) fn add_explicit_type(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -> Option<()> {
-    let syntax_node = ctx.find_node_at_offset::<Either<LetStmt, Param>>()?;
+    let syntax_node = ctx.find_node_at_offset_with_descend::<Either<LetStmt, Param>>()?;
     let (ascribed_ty, expr, pat) = if let Either::Left(let_stmt) = syntax_node {
         let cursor_in_range = {
-            let eq_range = let_stmt.eq_token()?.text_range();
+            let let_range = ctx.sema.original_range_opt(let_stmt.syntax())?.range;
+            let first = crate::utils::cover_edit_range(ctx.source_file().syntax(), let_range)
+                .start()
+                .clone();
```
Veykril, line 30:
```suggestion
            let let_range = ctx.sema.original_range_opt(let_stmt.syntax())?;
            if let_range.file_id != ctx.file_id() {
                return None;
            }
            let first = crate::utils::cover_edit_range(ctx.source_file().syntax(), let_range.range)
                .start()
                .clone();
```
Or something along those lines (dont recall the apis here), technically upmapping can land us in a different file I think (thanks to `include!` etc)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22552/587ce15e272b94d4f0a69d695e3718a02c24667e..5f0ef404771bffe3aa29184154319c3655c21b88)*

Veykril (approved):
same file id concern for the other two original_range calls, we need to check the resulting file id. But otherwise lgtm!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22552/587ce15e272b94d4f0a69d695e3718a02c24667e..5f0ef404771bffe3aa29184154319c3655c21b88)*

crates/ide-assists/src/handlers/add_explicit_type.rs
A4-Tacks: I think this is a very niche case, like `let x = include!("expr");`

A4-Tacks: I find it hard to think of a case that would cause this problem. If it's just that, I tend to merge it as it is

Veykril: I'm not sure I understand, that shouldn't be a lot of trouble to guard against? We had issues with mixing wrong text ranges and files in the past so I'd prefer if we'd try to avoid this. Especially if proc-macros end up being able to mix spans from differing files in the future as well (I don't think they can now, but they might in the future) at which point this can very much break again

A4-Tacks: Do you think the `original_range_in` in #22551 is expected?

A4-Tacks: I think this is difficult to solve because macros can mix non adjacent tokens in the same file

In the future, a more consistent solution should be needed to simply run code-action on tokens that are expanded as is in macros

PR #22562, opened 2026-06-10
Create directory for `cargo xtask metrics rustc_tests`

Hoverbear:
In c0f428d55b425c8ba18039a3687cdcdc47e111d1 (https://github.com/rust-lang/rust-analyzer/pull/20018) this code was adjusted to handle a change in `load_workspace`.

The change assumed that the `ra-rustc-test` folder existed in the `std::env::temp_dir`, but this assumption is not always correct.

## Testing

You can test with `cargo xtask metrics rustc_tests`.

Note: If you run the command multiple times you may need to `rm -r target/metrics/rust` to clean the directory.


Veykril (approved):

Veykril:
Thanks! (we should probaly re-enable this on CI at some point, iirc last we stopped it because we kept crashing in it🙈)

PR #22566, opened 2026-06-11
Fix destructuring assignments not introducing moves

### LINKED ISSUE #22565: Wrong closure capture analysis
The snippet below is valid rust but rust-analyzer incorrectly infers a `&wrapper` capture when it should be a `wrapper` capture. 
```rust
struct NonCopy;

struct Wrapper {
    field: NonCopy,
}

impl Wrapper {
    fn wrapped(&self) -> NonCopy {
        NonCopy
    }
}

pub struct Wrapper2 {
    field1: NonCopy,
    field2: NonCopy,
}

fn fun(wrapper: Wrapper) {
    pub fn update<T>(_: impl FnOnce(&mut T)) {
        todo!()
    }

    update::<Wrapper2>(|this| {
        this.field1 = wrapper.wrapped();
        this.field2 = wrapper.field;
     // ^^^^^^^^^^^ cannot move `NonCopy` out of reference
    });
}
```

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/22565

I believe this to be correct as it ought to mirror the binding pat case for by val mode.

### DIFF
--- a/crates/hir-ty/src/infer/closure/analysis/expr_use_visitor.rs
+++ b/crates/hir-ty/src/infer/closure/analysis/expr_use_visitor.rs
@@ -1016,8 +1016,9 @@ impl<'a, 'b, 'db, D: Delegate<'db>> ExprUseVisitor<'a, 'b, 'db, D> {
                     }
                 }
                 Pat::Expr(expr) => {
-                    // Destructuring assignment.
                     this.mutate_expr(expr)?;
+                    // Destructuring assignment moves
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (approved):

ChayimFriedman2:
Yes, rustc lowers this to `let temp = <rhs>; <lhs> = temp;`, so it ought to mirror binding.

PR #22574, opened 2026-06-12
fix: use package id as argument to `--package` if package is not unique

dfireBird:
A simple fix that does not need much changes, not sure how to add a test for this tbrh.

We could a do another fix as well checking if package is "unique" by comparing the versions (I have it in works as well, if this not good, I'll push that one)


Veykril (approved):

Veykril:
Thanks!

PR #22575, opened 2026-06-12
Don't count C-variadic `...` as a parameter for fn pointers

### LINKED ISSUE #22573: False positive E0107 "wrong number of arguments" on variadic extern "C" function pointer calls
## Summary

rust-analyzer reports E0107 ("wrong number of arguments") when calling a
variadic `extern "C"` function *pointer* with only the named arguments.
`rustc` and `cargo check` accept the code without error.

## Reproduction

```rust
use std::os::raw::c_int;

#[repr(C)]
enum NifOption { DelayHalt }

struct NifEnv;

struct Funcs {
    pub set_option: unsafe extern "C" fn(*mut NifEnv, NifOption, ...) -> c_int,
}

fn funcs() -> &'static Funcs { todo!() }

fn main() {
    let env: *mut NifEnv = std::ptr::null_mut();
    unsafe { (funcs().set_option)(env, NifOption::DelayHalt) };
    // rust-analyzer: E0107 expected 3 arguments, found 2
    // rustc: no error
}
```

## Expected behaviour

No diagnostic. The `...` in a C-variadic function pointer is not a required
argument; calling with only the named parameters is valid.

## Actual behaviour

E0107: `expected 3 arguments, found 2` is reported on the call site.
rust-analyzer is counting the `...` slot as a required positional argument.

## Notes

This is distinct from #3088 (variadic function *declarations* parsed
incorrectly), which was fixed. This appears to be in the arity-checking
path for variadic function *pointers* specifically.

## Version
rust analyzer vscode extension 0.3.2929
rust 1.95.0

ldgabbay:
rust-analyzer reported E0107 when calling a C-variadic `extern "C"`
function *pointer* with only its named arguments, e.g.:

```rust
struct Funcs {
    set_option: unsafe extern "C" fn(*mut Env, Opt, ...) -> i32,
}
unsafe { (funcs().set_option)(env, Opt::A) }; // false E0107
```

rustc and `cargo check` accept this.

Root cause: the fn-pointer type lowering in `expr_store/lower.rs`
included the `...` slot in the lowered parameter list, so the arity
check in `check_call_arguments` treated it as a required argument.
The function-*declaration* path already excludes it (the rust-lang/rust-analyzer#3088 fix);
this mirrors that filtering for fn pointers.

Added a regression test in `mismatched_arg_count.rs`.

Fixes rust-lang/rust-analyzer#22573

### DIFF
--- a/crates/hir-def/src/expr_store/lower.rs
+++ b/crates/hir-def/src/expr_store/lower.rs
@@ -671,6 +671,7 @@ impl<'db> ExprCollector<'db> {
                     }
 
                     pl.params()
+                        .filter(|it| it.dotdotdot_token().is_none())
                         .map(|it| {
                             let type_ref = self.lower_type_ref_opt(it.ty(), impl_trait_lower_fn);
                             let name = match it.pat() {
--- a/crates/ide-diagnostics/src/handlers/mismatched_arg_count.rs
+++ b/crates/ide-diagnostics/src/handlers/mismatched_arg_count.rs
@@ -355,6 +355,30 @@ fn f() {
     }
 
     #[test]
+    fn varargs_fn_pointer() {
+        check_diagnostics(
+            r#"
+struct Funcs {
+    f: unsafe extern "C" fn(u8, u8, ...) -> i32,
+    g: unsafe extern "C" fn(...) -> i32,
+}
+
+fn f(funcs: Funcs) {
+    unsafe {
+        (funcs.f)(0, 1);
+        (funcs.f)(0, 1, 2);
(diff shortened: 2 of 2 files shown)


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22575/5c7ce22ac85b8c98bf6ca038aefeba132a30a2ff..e9d9dc1fa049d0039bacbec7afca3ff960635714)*

PR #22576, opened 2026-06-12
internal: decide not to use `'db` in `NavigationTarget.docs`

ada4a:
Supersedes https://github.com/rust-lang/rust-analyzer/pull/22470, as per https://github.com/rust-lang/rust-analyzer/pull/22470#issuecomment-4580150555

cc @ChayimFriedman2 since you were the one who added the FIXME
cc @Veykril since you reviewed the original PR

### DIFF
--- a/crates/ide/src/navigation_target.rs
+++ b/crates/ide/src/navigation_target.rs
@@ -50,7 +50,6 @@ pub struct NavigationTarget {
     pub kind: Option<SymbolKind>,
     pub container_name: Option<Symbol>,
     pub description: Option<String>,
-    // FIXME: Use the database lifetime here.
     pub docs: Option<Documentation<'static>>,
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
I do not want to commit to not doing that, since we may need that anyway because of more lifetime annotations we need to add (e.g. https://github.com/rust-lang/rust-analyzer/pull/22548).

So I agree to remove the FIXME, but not to add another comment instead.

ada4a:
Removed the note

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22576/06070f3ad9824dca321f6407e2dcdd7952d33cd2..55baa899f299cdc3eb990a7ebf53fd15a4fe80d0)*

PR #22577, opened 2026-06-12
fix: handle fn-ptr variadic param

### LINKED ISSUE #22573: False positive E0107 "wrong number of arguments" on variadic extern "C" function pointer calls

A4-Tacks:
Fixes rust-lang/rust-analyzer#22573

Example
---
**Before this PR**

```rust
fn test(varargsfp: unsafe extern "C" fn(i16, ...)) {
     // ^^^^^^^^^ unsafe fn(i16, {unknown}, ...)
}
```

**After this PR**

```rust
fn test(varargsfp: unsafe extern "C" fn(i16, ...)) {
     // ^^^^^^^^^ unsafe fn(i16, ...)
}
```

### DIFF
--- a/crates/hir-def/src/expr_store/lower.rs
+++ b/crates/hir-def/src/expr_store/lower.rs
@@ -361,14 +361,8 @@ pub(crate) fn lower_function(
                     });
                 }
             }
-            let p = param_list
-                .params()
+            let p = non_variadic_params(&mut has_variadic, &param_list)
                 .filter(|param| collector.check_cfg(param))
-                .filter(|param| {
-                    let is_variadic = param.dotdotdot_token().is_some();
-                    has_variadic |= is_variadic;
-                    !is_variadic
-                })
                 .map(|param| param.ty())
                 // FIXME
                 .collect::<Vec<_>>();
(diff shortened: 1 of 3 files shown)


dfireBird:
The OP of issue has created a PR for the issue as well: #22575

A4-Tacks:
Oh, ok

PR #22579, opened 2026-06-13
perf: defer initial workspace flycheck until cache priming completes

tilladam:
On startup the initial cargo check flycheck is launched in the same `became_quiescent` block as cache priming, so the two run concurrently and fight for CPU. Cache priming fans crate work across worker threads and cargo check spawns up to one rustc job per core, so running them together oversubscribes the cores. The effect is felt particularly on macOS, where rust-analyzer's priming threads run at `QOS_CLASS_UTILITY` (`ThreadIntent::Worker`) while the `cargo check` subprocess runs at the default, higher QoS, so the scheduler starves priming (the work that makes the editor usable) in favour of the check.

This patch moves the initial workspace flycheck out of became_quiescent into the PrimeCachesProgress::End handler via a pending_initial_flycheck flag, so the check only starts once priming has finished. We fall back to firing it immediately when cache priming is disabled.

The impact is only felt when the startup cargo check has something to compile, on the first open after clone, git pull,  branch switch, toolchain change, where startup already hurts. 

Time-to-ready (cache priming complete) measured on macOS with a cold check, build scripts and proc-macros disabled to isolate the priming/check contention:

    rust-analyzer    74.8s -> 7.9s   (~9.5x)
    slint           103.1s -> 26.6s  (~3.9x)

Flycheck slow-tests still pass.

Disclosure: analyzed with assistance from Claude (Anthropic); changes reviewed and owned by the committer.

### DIFF
--- a/crates/rust-analyzer/src/main_loop.rs
+++ b/crates/rust-analyzer/src/main_loop.rs
@@ -387,6 +387,16 @@ impl GlobalState {
                             if cancelled {
                                 self.prime_caches_queue
                                     .request_op("restart after cancellation".to_owned(), ());
+                            } else if self.config.check_on_save(None)
+                                && self.config.flycheck_workspace(None)
+                                && !self.fetch_build_data_queue.op_requested()
+                            {
+                                // Priming finished; now run the deferred initial workspace flycheck
+                                // (kept off the critical path so `cargo check` doesn't contend with
+                                // cache priming for CPU).
+                                self.flycheck
+                                    .iter()
+                                    .for_each(|flycheck| flycheck.restart_workspace(None));
                             }
                             if let Some((message, fraction, title)) = last_report.take() {
                                 self.report_progress(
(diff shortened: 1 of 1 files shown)


crates/rust-analyzer/src/main_loop.rs
```diff
@@ -482,20 +490,26 @@ impl GlobalState {
...
+                {
+                    if self.config.prefill_caches() {
+                        // Project loaded; defer the initial workspace flycheck until cache priming
+                        // finishes so the `cargo check` subprocess doesn't steal CPU from priming.
+                        self.pending_initial_flycheck = true;
+                    } else {
+                        // No priming will run (cache priming disabled); kick off flycheck now.
+                        self.flycheck.iter().for_each(|flycheck| flycheck.restart_workspace(None));
+                    }
+                }
```
Veykril, line 514: I don't think we need new state tracked in GlobalState here. We should change things slightly:
1. If prime caching is disabled, just kick off flycheck by the given condition as before
2. If prime caching is enabled, add to the condition that we expect `proc_macros_loaded ` to be true, and `!self.prime_caches_queue.op_in_progress()` (notably that means we still need to move the flycheck start after the `prime_caches_queue.request_op` as was done in the PR) and `!self.prime_caches_queue.op_requested()`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22579/3f92cd1612268995d5667bd04fa03ba2916413d9..6171df6cf2ae2400ea0f8c7ad1511f40a308fd96)*

Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22579/3f92cd1612268995d5667bd04fa03ba2916413d9..d78428779a701433099d99480458b5e2804ae5b7)*

PR #22580, opened 2026-06-13
Avoiding temporary lowercase allocations

HuaGu-Dragon:
## Objective

- Remove unnecessary temporary lowercase allocations from several ASCII-only case-insensitive comparisons.


ChayimFriedman2:
This complicates the code a bit and it isn't perf-sensitive so I reject this. Thank you for interest in contributing!

I will accept, though, converting the `to_lowercase()` into `to_ascii_lowercase()`.

PR #22581, opened 2026-06-13
provide startup time to ready log point and associated benchmark

tilladam:
While profiling rust-analyzer startup and workspace initialization I found it useful to get the timing of completed cache priming, when the workspace is loaded and indexed. This patch provides that log point and a benchmark to track it over time. 

Disclosure: Anthropic Claude 4.7 built the benchmark, a supposedly qualified human reviewed it and takes responsibility for it.

### DIFF
--- a/crates/rust-analyzer/src/main_loop.rs
+++ b/crates/rust-analyzer/src/main_loop.rs
@@ -408,6 +408,9 @@ impl GlobalState {
                                     .iter()
                                     .for_each(|flycheck| flycheck.restart_workspace(None));
                             }
+                            if !cancelled {
+                                tracing::info!("workspace loaded and indexed");
+                            }
                             if let Some((message, fraction, title)) = last_report.take() {
                                 self.report_progress(
                                     title,
--- a/crates/rust-analyzer/src/main_loop.rs
+++ b/crates/rust-analyzer/src/main_loop.rs
@@ -397,19 +397,19 @@ impl GlobalState {
                             if cancelled {
                                 self.prime_caches_queue
                                     .request_op("restart after cancellation".to_owned(), ());
-                            } else if self.config.check_on_save(None)
-                                && self.config.flycheck_workspace(None)
-                                && !self.fetch_build_data_queue.op_requested()
-                            {
-                                // Priming finished; now run the deferred initial workspace flycheck
-                                // (kept off the critical path so `cargo check` doesn't contend with
-                                // cache priming for CPU).
-                                self.flycheck
-                                    .iter()
-                                    .for_each(|flycheck| flycheck.restart_workspace(None));
-                            }
-                            if !cancelled {
-                                tracing::info!("workspace loaded and indexed");
+                            } else {
+                                if self.config.check_on_save(None)
+                                    && self.config.flycheck_workspace(None)
+                                    && !self.fetch_build_data_queue.op_requested()
+                                {
+                                    // Priming finished; now run the deferred initial workspace flycheck
+                                    // (kept off the critical path so `cargo check` doesn't contend with
+                                    // cache priming for CPU).
+                                    self.flycheck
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
I'm sorry but I don't think this is worth the extra complexity when we already have a `prime-caches` subcommand.

tilladam:
Thanks for looking at this. The prime-caches subcommand is great for testing the raw database caching, but it bypasses the  main_loop  completely. It doesn't run the actual LSP server, the VFS progress sync, or the flycheck background tasks. I can be convinced that the benchmark doesn't add a ton of value, since regressions are somewhat more likely to come from the interaction of all those things, rather than the code path the benchmark covers, but I'd argue the log point still makes it quite a bit easier to get timing measurements in a significant spot when profiling the launch experience more holistically, for very modest added complexity.

ChayimFriedman2:
It still loads all files into the VFS and runs Cargo. We don't have much more than that.

But since you disagree, let's ask: @rust-lang/rust-analyzer does someone do want this?

Veykril:
Yea I don't think this is too useful either, having an fino log when done with workspace loading etc seems fine to me though, but anything other than that is unnecessary if you ask me

tilladam:
So you're saying keep the log point and nix the benchmark? I'm ok with that.

Veykril:
Yea but without the time tracking, just have a log for startup, and a log for whenever we load a workspace, the default logger has timestamps already after all

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -387,16 +387,9 @@ impl GlobalState {
...
-                                && !self.fetch_build_data_queue.op_requested()
-                            {
-                                // Priming finished; now run the deferred initial workspace flycheck
-                                // (kept off the critical path so `cargo check` doesn't contend with
-                                // cache priming for CPU).
-                                self.flycheck
-                                    .iter()
-                                    .for_each(|flycheck| flycheck.restart_workspace(None));
+                            }
+                            if !cancelled {
```
ChayimFriedman2:
```suggestion
} else {
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22581/72fee124d23f2a80c2dd3968717276b0c77406a1..ac8291baa2e5f172ee643257f9ede95d96be7762)*

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -387,16 +387,9 @@ impl GlobalState {
...
+                                tracing::info!("workspace loaded and indexed");
```
ChayimFriedman2:
```suggestion
                                tracing::info!("cache priming completed successfully");
```
?

A bit weird to talk about workspace loading here where it was completed a long time ago.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22581/72fee124d23f2a80c2dd3968717276b0c77406a1..ac8291baa2e5f172ee643257f9ede95d96be7762)*

ChayimFriedman2 (changes requested):

crates/rust-analyzer/src/main_loop.rs
tilladam: done

crates/rust-analyzer/src/main_loop.rs
tilladam: done

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22581/28eedd70c04d6036b811e11aaf0629d631cec794..8026695c3efdb5551aa6577b924b91cfd4d5fb54)*

PR #22582, opened 2026-06-13
fix: Only run `Drop::drop` when implemented

Veykril:
Otherwise with the recent `Drop` changes we now get stuck in a recursion as we attempt to drop `Pin<&mut Self`>` which itself then runs a destructor for `Pin<&mut Pin<&mut Self>>` and so on.

Was authored with opus 4.8 for investigation.

### DIFF
--- a/crates/hir-ty/src/mir/eval.rs
+++ b/crates/hir-ty/src/mir/eval.rs
@@ -3036,7 +3036,7 @@ impl<'a, 'db: 'a> Evaluator<'a, 'db> {
         ty: Ty<'db>,
         locals: &Locals<'a>,
         addr: Address,
-        _metadata: &[u8],
+        metadata: &[u8],
         span: MirSpan,
     ) -> Result<'db, ()> {
         let Some(drop_fn) = self.lang_items().Drop_drop else {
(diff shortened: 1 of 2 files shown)


crates/hir-ty/src/mir/eval.rs
```diff
@@ -3046,13 +3046,23 @@ impl<'a, 'db: 'a> Evaluator<'a, 'db> {
...
+            drop_fn,
+            generic_args,
+        );
+        if let Either::Left(drop_impl) = drop_impl
+            && matches!(drop_impl.lookup(self.db).container, ItemContainerId::ImplId(_))
+            && let Ok(body) = self.db.monomorphized_mir_body(
+                drop_impl.into(),
+                drop_args.store(),
+                self.param_env.store(),
+            )
```
Veykril, line 3060: Specifically here we now only run drop when its a concrete impl, not sure if we have a better more concise way of checking this tbh

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22582/3f92cd1612268995d5667bd04fa03ba2916413d9..b09dda00f1b374da9b29acdf565fe2f8b15cec25)*

ChayimFriedman2 (approved):

PR #22584, opened 2026-06-14
fix: offer inline macro in macro call and proc macro

eihqnh:
1. support `inline macro` expansion inside attribute macros and macro calls
2. add two new tests

## example
```rust
#[tokio::main]
async fn main() {
    $0println!();
    //inline macro was not working inside attribute macros
}


#[tokio::main]
async fn main() {
    {
        std::io::_print(std::format_args_nl!("{}",x));
    };
	//after this pr inline macro works
}

//another example
#[esp_rtos::main]
async fn main(spawner: Spawner) {
    $0rtt_target::rtt_init_defmt!();
}

#[esp_rtos::main]
async fn main(spawner: Spawner) {
//after
    {
        use rtt_target::ChannelMode::NoBlockSkip;
        rtt_target::rtt_init_defmt!(NoBlockSkip,1024);
    };
```

### DIFF
--- a/crates/ide-assists/src/handlers/inline_macro.rs
+++ b/crates/ide-assists/src/handlers/inline_macro.rs
@@ -1,8 +1,11 @@
 use hir::db::ExpandDatabase;
 use ide_db::syntax_helpers::prettify_macro_expansion;
-use syntax::ast::{self, AstNode, edit::AstNodeEdit};
+use syntax::ast::{self, AstNode, edit::IndentLevel};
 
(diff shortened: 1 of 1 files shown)


Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22584/5017a50153285a2d9bfa286e6fa96723e997d5b6..7705f06171b8209a3035441d76fb1f7938e68aae)*

PR #22587, opened 2026-06-14
Scoped lazy priming

tilladam:
rust-analyzer currently eagerly primes the whole workspace (all crates) on open. That's generally fine, as def-map computation is mostly fast enough. Proc-macro expansion, however, can be slow. The slint! macro is a particularly egregious (and admittedly a bit pathological) case here. The uses of the slint! macro in the slint project's own workspace cluster around examples and demos, leading to a very long priming phase when opening that project. Not a great experience for slint contributors.

This PR adds the ability to configure whether all crates (as before) should be primed, all open ones, or those representing libraries along with those from open files. The default is set to library, since that seems highly likely to me to be what people would expect and want, but this is a behavior change I'm happy to debate. 

json based workspaces default to open (also debatable) and the CLI flow etc still default to all, which seems sensible to me for those use cases.

With this new knob set to library, the initial priming of the slint workspace goes from 125s to 35s, a real quality of life difference for those of us working on slint. I've benchmarked a few other larger projects and none of them really benefit. So in the end I'm not sure if this is worth it and suggested fixing the workspace layout on the slint side here: https://github.com/slint-ui/slint/issues/12097.

So, if you feel this is useful enough, I'm happy to work on it further, if you feel we should rather just fix the slint workspace layout and not bother adding this complexity to rust-analyzer, my feelings won't be hurt either :).

### DIFF
--- a/crates/ide-db/src/prime_caches.rs
+++ b/crates/ide-db/src/prime_caches.rs
@@ -4,9 +4,8 @@
 //! various caches, it's not really advanced at the moment.
 use std::panic::AssertUnwindSafe;
 
-use base_db::all_crates;
 use hir::{Symbol, import_map::ImportMap, sym};
-use rustc_hash::FxHashMap;
+use rustc_hash::{FxHashMap, FxHashSet};
 use salsa::{Cancelled, Database};
 
 use crate::{FxIndexMap, RootDatabase, base_db::Crate, symbol_index::SymbolIndex};
@@ -23,12 +22,22 @@ pub struct ParallelPrimeCachesProgress {
     pub work_type: &'static str,
 }
 
+/// Warm caches for `scope`.
+///
+/// `scope` must be closed under transitive dependencies: the scheduler only
+/// follows reverse-dep edges within `scope`. Out-of-scope crates can still be
+/// primed by salsa on demand when a scope crate's queries reach into them.
+/// Callers that want to prime everything pass `&all_crates(db)`.
 pub fn parallel_prime_caches(
     db: &RootDatabase,
+    scope: &[Crate],
     num_worker_threads: usize,
     cb: &(dyn Fn(ParallelPrimeCachesProgress) + Sync),
 ) {
-    let _p = tracing::info_span!("parallel_prime_caches").entered();
+    if scope.is_empty() {
+        return;
+    }
+    let _p = tracing::info_span!("parallel_prime_caches", scope_size = scope.len()).entered();
 
     enum ParallelPrimeCacheWorkerProgress {
         BeginCrateDefMap { crate_id: Crate, crate_name: Symbol },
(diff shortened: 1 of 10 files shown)


flodiebold:
If someone has a big binary crate, priming only libraries is probably not a good default. Couldn't we just prime everything except examples? I'd argue that wouldn't even need to be a setting.

ChayimFriedman2:
I agree with @flodiebold, I don't like this setting and I agree not priming examples is a good idea.

However this setting might be useful in a monorepo, which is why maybe we should still include it. Even there, I don't think we should have the `library` option and inferring want-to-work-on crates from the open files is not necessarily a good idea. I'd argue we should just let the user explicitly set the crates in this case.

However I've never worked on a monorepo, so I'd like input from someone who does on whether they use cache priming and if this will help them.

Veykril:
I think defaulting to only libs and bins is good enough, that is I'd even go as far putting integration tests into the exclusion here? I don't think a config is necessary either, tests and examples are always leaf crates so them not being included in case priming isn't that problematic anyways, the main benefit of priming is that we can do analysis on dependencies in parallel after all

tilladam:
> I think defaulting to only libs and bins is good enough, that is I'd even go as far putting integration tests into the exclusion here? I don't think a config is necessary either, tests and examples are always leaf crates so them not being included in case priming isn't that problematic anyways, the main benefit of priming is that we can do analysis on dependencies in parallel after all

Done that in the latest push.

crates/ide-db/src/prime_caches.rs
```diff
@@ -28,7 +28,25 @@ pub fn parallel_prime_caches(
...
+    let all = all_crates(db);
+    parallel_prime_caches_scoped(db, &all, num_worker_threads, cb);
+}
+
+/// Same as [`parallel_prime_caches`], but only enqueues crates in `scope`.
+///
+/// `scope` must be closed under transitive dependencies: the scheduler only
+/// follows reverse-dep edges within `scope`. Out-of-scope crates can still be
+/// primed by salsa on demand when a scope crate's queries reach into them.
+pub fn parallel_prime_caches_scoped(
```
Veykril: I think we can just force the caller to always provide the scope, that is give `parallel_prime_caches` a `crates` parameter and force the caller to call into `all_crates`

crates/rust-analyzer/src/global_state.rs
```diff
@@ -195,6 +225,11 @@ pub(crate) struct GlobalState {
 
     pub(crate) minicore: MiniCoreRustAnalyzerInternalOnly,
     pub(crate) last_gc_revision: Revision,
+
+    /// Server start time, for reporting time-to-ready.
+    pub(crate) start_time: Instant,
+    /// Whether time-to-ready was already reported.
+    pub(crate) initial_ready_logged: bool,
```
Veykril: Unrelated?

crates/rust-analyzer/src/main_loop.rs
```diff
@@ -512,6 +544,15 @@ impl GlobalState {
                         self.flycheck.iter().for_each(|flycheck| flycheck.restart_workspace(None));
                     }
                 }
+            } else if proc_macros_loaded && memdocs_added_or_removed {
+                // Newly-opened files may pull in crates outside the primed
+                // scope; re-prime if so. Salsa skips already-warm crates.
```
Veykril:
```suggestion
            } else if proc_macros_loaded {
```
I don't think that is a good idea, that will cause the prime cache indicator to pop up everytime one opens a file, it also seems not the useful to me

Veykril (commented):
Have not reviewed the actual scope compute stuff yet

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22587/5c9046eff07e7a087043e2f15da98409f71621ea..38b933f3ea4c8daaaa68a23f4962e26d1607689c)*

tilladam:
I've stripped the open document seeding, so we are now only doing libs+bins. rust-project.json projects are now Lang-crates only. To change that, we'd need to elevate is_workspace_member from pub(crate) to pub. I can do that, but did not want to mix that into this PR. Is that worth doing?

Veykril:
Went and cleaned up some stuff a bit and also added better support for rust-project.json

Veykril (approved):

PR #22595, opened 2026-06-16
SCIP: Exclude leading/trailing trivia in definition ranges

Wilfred:
Previously, the navigation position (which we compute SCIP positions from) would include the full CST node position, including trivia. This generally isn't what we want in SCIP: we just want the definition itself.

SCIP positions, unlike LSP, are supposed to exclude trivia. Ensure we drop trivia when computing positions for SCIP.

AI disclosure: Written with help by Codex and GPT-5.5.

### DIFF
--- a/crates/ide/src/static_index.rs
+++ b/crates/ide/src/static_index.rs
@@ -2,6 +2,7 @@
 //! read-only code browsers and emitting LSIF
 
 use arrayvec::ArrayVec;
+use either::Either;
 use hir::{Crate, Module, Semantics, db::HirDatabase};
 use ide_db::{
     FileId, FileRange, FxHashMap, FxHashSet, RootDatabase,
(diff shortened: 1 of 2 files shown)


crates/ide/src/hover/tests.rs
Wilfred, line 4113: The hover position for `S` now no longer includes the comment, which seems correct. The comment doesn't actually relate to `S`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22595/06070f3ad9824dca321f6407e2dcdd7952d33cd2..489078c2553e235c99ea63c7071b06692b15b3e6)*

ChayimFriedman2:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#locationLink

> ```typescript
>   /**
>	 * The full target range of this link. If the target is, for example, a
>	 * symbol, then the target range is the range enclosing this symbol not
>	 * including leading/trailing whitespace but everything else like comments.
>	 * This information is typically used to highlight the range in the editor.
>	 */
>	targetRange: Range;
> ```

So I think our current behavior is actually standards-conformant.

Wilfred:
Interesting, I didn't know that. I'm surprised that leading/trailing comments are explicitly included in LSP, I'm struggling to imagine the use case outside of doc comments.

In any case, SCIP is a little more explicit about not including trivia unless they're doc comments, so I've pushed the comment removal into the SCIP generation logic.

ChayimFriedman2 (approved):

PR #22605, opened 2026-06-17
SCIP: Emit references for symbols referenced in macro-expansion

Wilfred:
Previously, we only emitted references from source tokens. This causes issues with find-references tools built on top of SCIP. For example, the following code would produce a SCIP index with no references to `greeting`.

    use askama::Template;

    #[derive(Template)]
    #[template(source = "{{ crate::greeting() }}", ext = "txt")]
    struct Page;

    fn greeting() -> &'static str {
        "hello from askama"
    }

Instead, walk macro-expanded tokens and emit SCIP references with the Generated role.

Whilst this is strictly more work, SCIP generation time increased by less than 10% when testing on the crates inside rust-analyzer.

(I've split out a commit that reindents the closure a bunch, to hopefully make the actual logic changes easier to read.)

AI disclosure: Partially written by Codex and GPT-5.5.

### DIFF
--- a/crates/ide/src/static_index.rs
+++ b/crates/ide/src/static_index.rs
@@ -6,12 +6,12 @@ use hir::{Crate, Module, Semantics, db::HirDatabase};
 use ide_db::{
     FileId, FileRange, FxHashMap, FxHashSet, RootDatabase,
     base_db::{SourceDatabase, VfsPath},
-    defs::{Definition, IdentClass},
+    defs::{Definition, IdentClass, NameClass},
     documentation::Documentation,
     famous_defs::FamousDefs,
     ra_fixture::RaFixtureConfig,
 };
-use syntax::{AstNode, SyntaxNode, SyntaxToken, TextRange};
+use syntax::{AstNode, NodeOrToken, SyntaxNode, SyntaxToken, TextRange, ast};
 
 use crate::navigation_target::UpmappingResult;
 use crate::{
@@ -98,7 +98,11 @@ pub struct StaticIndexedFile {
     pub file_id: FileId,
     pub folds: Vec<Fold>,
     pub inlay_hints: Vec<InlayHint>,
+    /// Tokens that occur in the file.
     pub tokens: Vec<(TextRange, TokenId)>,
+    /// Tokens that only occur after macro expansion. The text range
+    /// associated is the position of the macro call.
+    pub generated_tokens: Vec<(TextRange, TokenId)>,
 }
 
 fn all_modules(db: &dyn HirDatabase) -> Vec<Module> {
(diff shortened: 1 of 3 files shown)


crates/ide/src/static_index.rs
```diff
@@ -282,12 +324,63 @@ impl StaticIndex<'_> {
...
+                } else if let Some(item) = ast::Item::cast(node.clone()) {
+                    // Attribute proc-macros attached to an item.
+                    let call_site = item.syntax().text_range();
+                    sema.expand_attr_macro(&item)
+                        .map(|er| vec![(er.value.value, call_site)])
+                        .unwrap_or_default()
+                } else {
+                    Vec::new()
+                }
+            });
```
ChayimFriedman2, line 366: I don't think this is the way. No other IDE feature, except "Expand macro recursively", manually expands macros like that. Instead, they descend into each token of interest. Which you already do, except you filter to matching SyntaxKind/text and the first token.

I don't know what exactly SCIP needs. If it needs 100% reliability for finding references then this is indeed the way, but even r-a's own references finding does not do that (albeit due to time constraints).

You also do not recurse into expansions, meaning you'll miss nested macros. And using the macro call's range as the reference range is weird, I'd expect the token range instead (and even that won't be enough in cases like in the OP, since you cannot point in the middle of a string).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22605/abb1301c3c14a40645bb2588b1cc858fe374b527..92674c66b75d573eafc41fc9d4a90c508d2b40dc)*

crates/ide/src/static_index.rs
```diff
@@ -98,7 +98,11 @@ pub struct StaticIndexedFile {
     pub file_id: FileId,
     pub folds: Vec<Fold>,
     pub inlay_hints: Vec<InlayHint>,
+    /// Tokens that occur in the file.
     pub tokens: Vec<(TextRange, TokenId)>,
+    /// Tokens that only occur after macro expansion. The text range
+    /// associated is the position of the macro call.
+    pub generated_tokens: Vec<(TextRange, TokenId)>,
```
ChayimFriedman2, line 105: Why the separation between "real" and expanded tokens?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22605/abb1301c3c14a40645bb2588b1cc858fe374b527..92674c66b75d573eafc41fc9d4a90c508d2b40dc)*

Wilfred: FWIW I was experimenting with doing something different with SCIP and LSIF. I'm really not sure what do with LSIF in general: the standard seems pretty universally replaced by SCIP (e.g. https://lsif.dev/ is archived), so I'd be tempted to remove LSIF support entirely at some point.

Wilfred:
Playing with this some more, the specific repro doesn't actually provide a go-to-references in LSP either. I really should have double-checked that, sorry. Normally LSP works perfectly so I assumed it covered this case too.

Given that this doesn't work with LSP today, I'm going to close this. I think the proper solution is probably to fix in r-a in general, not SCIP specifically.

ChayimFriedman2:
It used to work in LSP, then we explicitly removed support for that. We don't and won't support macros that generate definitions from strings. Most of them (e.g. serde) are only doing that for legacy reasons.

Whether SCIP supports macro expanded code, I don't know but probably yes. It doesn't currently expand all macros like you do here though, so what I said above still applies.

PR #22619, opened 2026-06-19
fix: Exclude impls on the error type from impl enumeration

Wilfred:
Previously, rust-analyzer would include impl methods from `impl Foo for NoSuchType`, which would lead to rust-analyzer reporting nonexistent type errors.

This is noticeable when depending on tokio with feature="fs" and cfg(test) enabled. The following code would produce an invalid type error.

    use std::io::Read;
    fn f<T>(mut o: Option<T>) {
        o.take().unwrap();
    }

rust-analyzer was incorrectly resolving the `.take()` to `Read::take()` instead of `Option::take()`.

Inside tokio, it uses `mockall:mock!` to define a `MockFile`, and then does `impl Read for &'_ MockFile`. However, since `mockall` is a dev dependency of tokio and we don't include dev dependencies of dependencies, rust-analyzer just sees `impl Read for &'_ NoSuchType`.

Instead, exclude traits whose self type is error.

AI disclosure: Partially written by Codex and GPT-5.5

### DIFF
--- a/crates/hir-ty/src/method_resolution.rs
+++ b/crates/hir-ty/src/method_resolution.rs
@@ -728,7 +728,13 @@ impl TraitImpls {
                     {
                         continue;
                     }
+
                     let self_ty = trait_ref.self_ty();
+                    if self_ty_has_error_constructor(self_ty) {
+                        // If we see `impl Foo for NoSuchType`, just ignore it.
+                        continue;
+                    }
+
                     let interner = DbInterner::new_no_crate(db);
                     let entry = map.entry(trait_ref.def_id.0).or_default();
                     match simplify_type(interner, self_ty, TreatParams::InstantiateWithInfer) {
@@ -862,5 +868,24 @@ impl TraitImpls {
             for_each_block(trait_block, type_block).for_each(&mut *for_each);
             for_each_block(type_block, trait_block).for_each(for_each);
         }
+    }
+}
+
+fn self_ty_has_error_constructor<'db>(mut self_ty: Ty<'db>) -> bool {
+    if !self_ty.references_non_lt_error() {
+        return false;
+    }
+
+    loop {
+        self_ty = match self_ty.kind() {
+            TyKind::Error(_) => return true,
+            TyKind::Ref(_, inner, _)
+            | TyKind::RawPtr(inner, _)
+            | TyKind::Array(inner, _)
+            | TyKind::Slice(inner)
+            | TyKind::Pat(inner, _) => inner,
+            TyKind::UnsafeBinder(inner) => inner.skip_binder(),
+            _ => return false,
+        };
     }
 }
--- a/crates/hir-ty/src/tests/method_resolution.rs
+++ b/crates/hir-ty/src/tests/method_resolution.rs
@@ -2258,3 +2258,30 @@ fn main() {
     "#,
     );
 }
+
+#[test]
+fn trait_impl_with_error_self_ty_does_not_match_arbitrary_receiver() {
+    check_types(
+        r#"
+//- minicore: sized
+trait UnrelatedTrait {
+    fn take(self) {}
+}
+
+struct MyOption<T>(T);
+
+impl<T> MyOption<T> {
+    fn take(&mut self) -> MyOption<T> {
+        loop {}
+    }
+}
+
+fn f<T>(mut o: MyOption<T>) {
+    let value = o.take();
+      //^^^^^ MyOption<T>
+}
+
+impl UnrelatedTrait for &'_ MissingType {}
+"#,
+    );
+}


Veykril:
Not opposed to this, it sounds like the right thing to do here, but why is cfg(test) set for you in tokio which I presume is a dependency of your workspace? That asks for weird issues

Wilfred:
Yeah, this is the curse of a monorepo: we don't really know which libraries you're actively working on, so I defensively set `cfg(test)` essentially everywhere so you can always go-to-def on test helpers. I noticed this when using rust-project as a discover command.

It looks like I should modify rust-project to exclude cfg(test) on imported libraries like tokio, since our monorepo (using reindeer) doesn't actually have dev-dependencies for imported libraries so `cfg(test)` is genuinely wrong there.

The bug was a spooky action-at-a-distance issue: adding tokio to the transitive set of dependencies causes spurious red squiggles and it took me a while to get to the bottom of it. This fix does feel appropriately defensive.

I see CI is red, I'll fix that too.

ChayimFriedman2 (changes requested):
Generally that's something that came up multiple times in the past. In each of them we ended up finding the problematic impl and resolving its issue, but I agree this is something we want in general as the solver is quite sensitive to such impls.

However I think `references_non_lt_error()` is too strong condition, checking for error type itself should be enough IMO.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22619/ff5fcc70d9bcdc54db365114e2f7b1c18056ed4f..4b2028300d1efcee3dfe26ac4e1c09a58f7e34f7)*

Wilfred:
@ChayimFriedman2 how about this, using `references_only_ty_error` instead? That seems to be a narrower condition.

I did play with `if matches!(trait_ref.skip_binder().self_ty().kind(), TyKind::Error(_))` but that doesn't work with my original repro, since the impl is on `&NoSuchType`.

ChayimFriedman2:
I'm a bit on the edge about this. On one hand, const errors are indeed more commonly used by bugs in r-a than type errors. On the other hand... removing the impl due to *any* type error feels a bit like going too far. Maybe, and given that this is the most problematic case, only filter it if it has an error constructor? That is, peeling references, raw pointers(?), arrays(?), slices(?), pattern types, and unsafe binders.

But I'm not sure myself.

Wilfred:
That makes sense to me: the only really problematic case is instances of `impl Foo for SomeMissingType` leaking everywhere. I've updated the PR and things like `impl Foo for Option<SomeMissingType>` seem to work correctly.

crates/hir-ty/src/next_solver/interner.rs
```diff
@@ -1755,6 +1779,14 @@ impl<'db> Interner for DbInterner<'db> {
 
         TraitImpls::for_each_crate_and_block(self.db, krate, block, &mut |impls| {
             for &impl_ in impls.blanket_impls(trait_def_id.0) {
+                let Some(trait_ref) = self.db.impl_trait(impl_) else {
+                    continue;
+                };
+                let self_ty = trait_ref.skip_binder().self_ty();
+                if is_error_or_ref_error_ty(self_ty) {
```
ChayimFriedman2: Should check `references_non_lt_error()` before, it's just a flag so faster for the common case.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22619/ef691be44d10f57bae44f9ffb5dbd0c90e6d924f..ab02708a227566c3e7578d637ff4aadd3ed92019)*

ChayimFriedman2 (changes requested):

Wilfred:
Something like this? I've added a `references_non_lt_error()` call and also tweaked the helper to be trait specific, I think it reads better.

crates/hir-ty/src/next_solver/interner.rs
```diff
@@ -1642,6 +1665,13 @@ impl<'db> Interner for DbInterner<'db> {
                     let (regular_impls, builtin_derive_impls) =
                         impls.for_trait_and_self_ty(trait_def_id.0, &simp);
                     for &impl_ in regular_impls {
+                        let Some(trait_ref) = self.db.impl_trait(impl_) else {
```
ChayimFriedman2: I didn't saw this previously, but definitely don't do the filtering in this method. It's super duper hot. Do it when collecting the impls, so an erroneous impl will never reach the interner.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22619/f3652a64211b437ce2e45aca2af165547a0020d1..54bd33349cf3474a76f57c5fd752b3570179f1f9)*

ChayimFriedman2 (changes requested):

crates/test-utils/src/minicore.rs
```diff
@@ -1655,6 +1655,10 @@ pub mod option {
             }
         }
 
+        pub const fn take(&mut self) -> Option<T> {
```
ChayimFriedman2: I don't like adding convenience methods to minicore more than needed. You'll be perfectly fine defining your own type with `take()` in the test.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22619/f3652a64211b437ce2e45aca2af165547a0020d1..54bd33349cf3474a76f57c5fd752b3570179f1f9)*

Wilfred:
OK, moved the logic, and done another pass on making the test small and more self-documenting.

ChayimFriedman2 (approved):

PR #22621, opened 2026-06-20
fix: Don't panic on out-of-range integer literals in const positions

### LINKED ISSUE #22620: Array length literal larger than `usize::MAX` panics `hir-ty` const interning
## Summary

An array-length (or const-generic) position whose integer literal does not fit in the target integer type, for example `[u8; 18446744073709551616]` (`2^64`, one past `usize::MAX`), makes rust-analyzer **panic** while interning the length constant. Because the panic happens inside a salsa query that participates in cycle recovery, salsa keeps re-propagating the stored panic, and the language-server log fills **without bound** with `WARN Propagating panic for cycle head that panicked in an earlier execution in that revision`, repeated on every later revision until the offending line is removed. The editor otherwise stays usable, so the user-visible damage is the unbounded log growth and the wasted work of re-propagating the panic.

## Steps to reproduce

In VS Code:

1. In any Cargo project, add `fn f(_: [u8; 18446744073709551616]) {}` to a source file (`18446744073709551616 == usize::MAX + 1` on a 64-bit target).
2. Open `View > Output` and pick the `rust-analyzer Language Server` channel.
3. As soon as the file is analyzed, the panic is logged, followed by `Propagating panic for cycle head ...` warnings that keep accumulating on every later revision. Deleting the line stops them.

From the CLI (no editor needed), point any affected `rust-analyzer` build at a throwaway crate:

```bash
cargo new --lib /tmp/ra-repro
echo 'fn f(_: [u8; 18446744073709551616]) {}' > /tmp/ra-repro/src/lib.rs
rust-analyzer analysis-stats /tmp/ra-repro
```

The run reports `panics: 1` and prints the `consteval.rs` panic.

## Actual behavior

rust-analyzer panics while interning the array-length constant:

```
thread '...' panicked at crates/hir-ty/src/consteval.rs:96:73:
```

The server log then repeats, on every analysis revision and without limit:

```
WARN Propagating panic for cycle head that panicked in an earlier execution in that revision
WARN Propagating panic for cycle head that panicked in an earlier execution in that revision
... (continues indefinitely)
```

## Expected behavior

The out-of-range literal should be treated as an error constant (no panic), the same way other const-evaluation errors are handled, and ideally surfaced as a "literal out of range for `usize`" diagnostic. No panic, no log spam.

## Environment

- rust-analyzer `0.3.2937-standalone (cc272809a1 2026-06-14)`
- VS Code 1.125.1, extension `rust-lang.rust-analyzer` 0.3.2937
- macOS 26.5.1 (arm64)
- Also reproduces from the CLI via `rust-analyzer analysis-stats`

## Related

Same general area as #22600 / #22601 (panics on array lengths), but a **distinct cause**: an out-of-range integer *literal* rather than a `static` / type alias in length position.

shulaoda:
Writing an array length that doesn't fit in `usize`, like `[u8; 18446744073709551616]`, makes rust-analyzer panic and then spam the log with `Propagating panic for cycle head ...` over and over.

The problem is in `intern_const_ref`: it interns the literal with `ScalarInt::try_from_uint(..).unwrap()`, but `try_from_*` returns `None` when the value doesn't fit the target type, so the `unwrap()` blows up inside a salsa query. I changed those three integer branches (and `usize_const`) to return an error const instead of unwrapping, so an out-of-range literal is just treated as an error rather than crashing.

Also added a regression test. Running `analysis-stats` on the repro reports `panics: 1` before the change and `panics: 0` after.

Closes rust-lang/rust-analyzer#22620

### DIFF
--- a/crates/hir-ty/src/consteval.rs
+++ b/crates/hir-ty/src/consteval.rs
@@ -87,18 +87,24 @@ fn intern_const_ref<'db>(
     let valtree = match (ty.kind(), value) {
         (TyKind::Uint(uint), Literal::Uint(value, _)) => {
             let size = uint.bit_width().map(Size::from_bits).unwrap_or(data_layout.pointer_size());
-            let scalar = ScalarInt::try_from_uint(*value, size).unwrap();
+            let Some(scalar) = ScalarInt::try_from_uint(*value, size) else {
+                return Ok(Const::error(interner));
+            };
             ValTreeKind::Leaf(scalar)
         }
         (TyKind::Uint(uint), Literal::Int(value, _)) => {
             // `Literal::Int` is the default, so we also need to account for the type being uint.
             let size = uint.bit_width().map(Size::from_bits).unwrap_or(data_layout.pointer_size());
-            let scalar = ScalarInt::try_from_uint(*value as u128, size).unwrap();
+            let Some(scalar) = ScalarInt::try_from_uint(*value as u128, size) else {
+                return Ok(Const::error(interner));
+            };
             ValTreeKind::Leaf(scalar)
         }
         (TyKind::Int(int), Literal::Int(value, _)) => {
             let size = int.bit_width().map(Size::from_bits).unwrap_or(data_layout.pointer_size());
-            let scalar = ScalarInt::try_from_int(*value, size).unwrap();
+            let Some(scalar) = ScalarInt::try_from_int(*value, size) else {
+                return Ok(Const::error(interner));
+            };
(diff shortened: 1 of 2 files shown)


Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22621/ead27e60a2f9b5b738005142ba39608a01a8a893..d4f9db048447800c24d07e1815c3e2d2359059c5)*

PR #22623, opened 2026-06-20
minor: Use `std::env::home_dir` instead of home crate

### LINKED ISSUE #22570: Switch from home::home_dir to standard library home_dir
The `toolchain` crate currently has a dependency on the `home` crate, because it uses the `home_dir` function
https://github.com/rust-lang/rust-analyzer/blob/1edcb122f3766bd2079f7964f75f488146f8b6c9/crates/toolchain/src/lib.rs#L122

However, according to the documentation of https://crates.io/crates/home , users are 
> strongly encouraged to switch to using the standard library's [home_dir](https://doc.rust-lang.org/nightly/std/env/fn.home_dir.html) instead.

raushan728:
Closes rust-lang/rust-analyzer#22570

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1003,15 +1003,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "home"
-version = "0.5.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc627f471c528ff0c4a49e1d5e60450c8f6461dd6d10ba9dcd3a61d3dff7728d"
-dependencies = [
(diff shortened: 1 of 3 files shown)


Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22623/ead27e60a2f9b5b738005142ba39608a01a8a893..8b9866c88ed8bfb8dedf5c4d8a1b1eab0c633519)*

PR #22629, opened 2026-06-21
feat(assists): support `super::` paths in `qualify_path` assist

### LINKED ISSUE #21290: Assist idea: replace `super::` with an absolute path
Currently, VSCode doesn't suggest this on `Ctrl+.` I had to manually delete `super::` and then use the `Qualify as ...` assist on the item.

I use `rust-analyzer` VSCode extension, currently at version 0.3.2719

raushan728:
Fixes: rust-lang/rust-analyzer#21290

Adds support for triggering the `qualify_path` assist on `super::` paths, replacing them with their absolute `crate::...` equivalents.

https://github.com/rust-lang/rust-analyzer/issues/21290#issuecomment-3664576887
> IMO it would make sense to make the Qualify assist work on super paths as well.

Previously, `super::foo` was ignored by the assist. The blocker was that `super::foo` is already a valid, fully resolved path. The existing `find_importable_node` logic strictly filters out paths that successfully resolve (`sema.resolve_path(path).is_some`), meaning the assist never even fired for super.

Changes:
- Added a fast-path `qualify_super_path` at the beginning of the `qualify_path` assist
- If the path starts with `SuperKw` we resolve its semantic definition
- We then use `find_use_path` with `PrefixKind::ByCrate` to generate the correct absolute path string
- Ensured this ignores `use` statements which have their own logic
- Added 3 tests `simple`, `nested`, and `cursor positions`

### DIFF
--- a/crates/ide-assists/src/handlers/qualify_path.rs
+++ b/crates/ide-assists/src/handlers/qualify_path.rs
@@ -5,7 +5,10 @@ use hir::AsAssocItem;
 use ide_db::RootDatabase;
 use ide_db::{
     helpers::mod_path_to_ast_with_factory,
-    imports::import_assets::{ImportCandidate, LocatedImport},
+    imports::{
+        import_assets::{ImportCandidate, LocatedImport},
+        insert_use::PrefixKind,
+    },
 };
 use syntax::Edition;
 use syntax::ast::HasGenericArgs;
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
Sorry, but [we have a feature freeze on new assists due to lack of review capacity](https://github.com/rust-lang/rust-analyzer/blob/master/CONTRIBUTING.md). Therefore, I'm closing this PR for now.

Thank you for your interest in contributing!

PR #22637, opened 2026-06-25
internal: Bump salsa

Veykril:
Lots of improvements have landed in this patch release

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -808,15 +808,6 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
-version = "0.16.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


Veykril:
cc @MichaReiser, this cut down r-a's memory usage on itself by 70mb! 1501 -> 1435

MichaReiser:
Nice! You might also see large wins from marking some inputs as `NEVER_CHANGE`. The impact on ty was minimal (1%), but that's only because most of our queries read some setting which has a lower durability. You could see much better results if you have queries that only read `NEVER_CHANGE` inputs.

Veykril:
Yea I was planning to in https://github.com/rust-lang/rust-analyzer/pull/22638 but somethings up with CI there, haven't gotten around looking into that.

MichaReiser:
ohoh :D

PR #22638, opened 2026-06-25
internal: Make use of salsa's new NEVER_CHANGE durability

Veykril:


crates/base-db/src/lib.rs
```diff
@@ -128,9 +128,10 @@ impl Files {
         durability: Durability,
     ) {
         match self.files.entry(file_id) {
-            Entry::Occupied(mut occupied) => {
+            Entry::Occupied(mut occupied) if durability != Durability::NEVER_CHANGE => {
```
MichaReiser: Ignoring the write seems a bit suspicious. Isn't it desired to panic here and, instead, either fix the durability for the file if it is `NEVER_CHANGE` and can change, or change the call site not to call `set_file_text_with_durability` for such a fiel?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22638/cdab17b9fc0b312e2a32c351fbe3cd938557a44a..599005cb48b5e44bce1a011a9c6a826e7f071c1f)*

ChayimFriedman2:
If this has the same disadvantage as my old branch, namely that it cannot reload changed features in dependencies correctly, then I oppose to that. We're not at the same place we were at the time wrt. memory usage, and today I think this is not worth the code complexity and workflow disruption.

MichaReiser:
In ty, we only use it for fields that are truly immutable. That is, we vendor some files, so they can never change. Each file has a `path`, and the path can never change (it's its identity).

ChayimFriedman2:
We don't have such things, but we do have std which I suppose we can assume never changes. It won't be a big win, but given this is already in Salsa I guess it could be worth it.

Veykril:
But we do? Any file in the local crate registry as well as the standard library we can assume to be never changing. That's what this PR is doing. Aside from that I agree, there isn't really anything else (aside from some source roots in theory maybe?).

ChayimFriedman2:
No, like I said back then, changing Cargo features (or cfgs) can change the analysis on those files, even if their content doesn't change.

ChayimFriedman2:
I see now that you don't mark the `Crate` itself as `NEVER_CHANGE`, only the file text, so I believe it should work but the memory saving will be vanishingly small.

Veykril:
Yea, trusting the analysis stats run here its ~4mb on self. Not great but might as well try. I do want to check out how much using on a crate would save (just out of curiosity), its a shame we can't actually do that.

crates/base-db/src/lib.rs
Veykril: You are not wrong, but our file setup is kind of convoluted so I don't even know if we can reliably do that.

Veykril: In fact yea then we'd need to do the filtering somewhere else as we do track changes to these files (which arguably we shouldn't register file watchers for I think)

ChayimFriedman2:
Do note that this is not completely enough, you also need to make the `expand_proc_attr_macros()` setting NEVER_CHANGE.

PR #22639, opened 2026-06-25
internal: remove input queries from `#[query_group]`

ada4a:

- First commit removes the input query `DefDatabase::expand_proc_attr_macros`
- Second commit removes the input query `ExpandDatabase::proc_macros`, and its sister query `ExpandDatabase::proc_macros_for_crate` (more details in the commit description)
- Third commit removes the input query support from `#[query_group]`

cc @Veykril

### DIFF
--- a/crates/hir-def/src/db.rs
+++ b/crates/hir-def/src/db.rs
@@ -4,6 +4,7 @@ use hir_expand::{
     EditionedFileId, HirFileId, InFile, Lookup, MacroCallId, MacroDefId, MacroDefKind,
     db::ExpandDatabase,
 };
+use salsa::{Durability, Setter};
 use triomphe::Arc;
 
 use crate::{
@@ -16,10 +17,6 @@ use crate::{
 
 #[query_group::query_group]
 pub trait DefDatabase: ExpandDatabase + SourceDatabase {
-    /// Whether to expand procedural macros during name resolution.
-    #[salsa::input]
-    fn expand_proc_attr_macros(&self) -> bool;
-
     /// Computes an [`ItemTree`] for the given file or macro expansion.
     #[salsa::invoke(file_item_tree)]
     #[salsa::transparent]
@@ -45,6 +42,26 @@ pub trait DefDatabase: ExpandDatabase + SourceDatabase {
 
     #[salsa::invoke(include_macro_invoc)]
     fn include_macro_invoc(&self, crate_id: Crate) -> Arc<[(MacroCallId, EditionedFileId)]>;
+}
+
+/// Whether to expand procedural macros during name resolution.
+///
+/// Note: this struct shouldn't be exposed to downstream crates -- consider using
+/// [`set_expand_proc_attr_macros`] instead, if possible.
+#[salsa::input(singleton, debug)]
+pub(crate) struct ExpandProcAttrMacros {
+    #[returns(copy)]
+    pub(crate) enabled: bool,
+}
+
+pub fn set_expand_proc_attr_macros(db: &mut dyn DefDatabase, enabled: bool) {
+    if let Some(expand_proc_attr_macros) = ExpandProcAttrMacros::try_get(db) {
+        if expand_proc_attr_macros.enabled(db) != enabled {
+            expand_proc_attr_macros.set_enabled(db).with_durability(Durability::HIGH).to(enabled);
+        }
+    } else {
+        _ = ExpandProcAttrMacros::builder(enabled).durability(Durability::HIGH).new(db);
+    }
 }
 
 // return: macro call id and include file id
--- a/crates/hir-expand/src/db.rs
+++ b/crates/hir-expand/src/db.rs
@@ -17,7 +17,7 @@ use crate::{
     declarative::DeclarativeMacroExpander,
     fixup::{self, SyntaxFixupUndoInfo},
     hygiene::{span_with_call_site_ctxt, span_with_def_site_ctxt, span_with_mixed_site_ctxt},
-    proc_macro::{CrateProcMacros, CustomProcMacroExpander, ProcMacros},
+    proc_macro::CustomProcMacroExpander,
     span_map::{ExpansionSpanMap, RealSpanMap, SpanMap},
     tt,
 };
@@ -50,14 +50,6 @@ pub enum TokenExpander<'db> {
 
 #[query_group::query_group]
 pub trait ExpandDatabase: SourceDatabase {
-    /// The proc macros. Do not use this! Use `proc_macros_for_crate()` instead.
-    #[salsa::input]
-    fn proc_macros(&self) -> Arc<ProcMacros>;
-
-    /// Incrementality query to prevent queries from directly depending on `ExpandDatabase::proc_macros`.
-    #[salsa::invoke(crate::proc_macro::proc_macros_for_crate)]
-    fn proc_macros_for_crate(&self, krate: Crate) -> Option<Arc<CrateProcMacros>>;
-
     #[salsa::invoke(ast_id_map)]
     #[salsa::transparent]
     fn ast_id_map(&self, file_id: HirFileId) -> &AstIdMap;
--- a/crates/hir-expand/src/proc_macro.rs
+++ b/crates/hir-expand/src/proc_macro.rs
@@ -7,6 +7,7 @@ use std::{panic::RefUnwindSafe, sync};
 use base_db::{Crate, CrateBuilderId, CratesIdMap, Env, ProcMacroLoadingError};
 use intern::Symbol;
 use rustc_hash::FxHashMap;
+use salsa::{Durability, Setter};
 use span::Span;
 use triomphe::Arc;
 
@@ -83,14 +84,24 @@ impl ProcMacrosBuilder {
         );
     }
 
-    pub(crate) fn build(self, crates_id_map: &CratesIdMap) -> ProcMacros {
+    /// Builds [`ProcMacros`] and adds id to `db` with the provided `durability`.
+    pub(crate) fn build_in(
+        self,
+        db: &mut dyn ExpandDatabase,
+        durability: Durability,
+        crates_id_map: &CratesIdMap,
+    ) {
         let mut map = self
             .0
             .into_iter()
             .map(|(krate, proc_macro)| (crates_id_map[&krate], proc_macro))
             .collect::<FxHashMap<_, _>>();
         map.shrink_to_fit();
-        ProcMacros(map)
+        ProcMacros::try_get(db)
+            .unwrap_or_else(|| ProcMacros::new(db, Default::default()))
+            .set_by_crate(db)
+            .with_durability(durability)
+            .to(map);
     }
 }
 
@@ -107,11 +118,25 @@ impl FromIterator<(CrateBuilderId, ProcMacroLoadResult)> for ProcMacrosBuilder {
 #[derive(Debug, PartialEq, Eq)]
 pub struct CrateProcMacros(StoredProcMacroLoadResult);
 
-#[derive(Default, Debug)]
-pub struct ProcMacros(FxHashMap<Crate, Arc<CrateProcMacros>>);
+/// The proc macros. Do not use [`Self::get`]! Use [`Self::get_for_crate`] instead.
+#[salsa::input(singleton, debug)]
+pub struct ProcMacros {
+    #[returns(ref)]
+    pub by_crate: FxHashMap<Crate, Arc<CrateProcMacros>>,
+}
+
 impl ProcMacros {
-    fn get(&self, krate: Crate) -> Option<Arc<CrateProcMacros>> {
-        self.0.get(&krate).cloned()
+    pub fn init_default(db: &dyn ExpandDatabase, durability: Durability) {
+        _ = Self::builder(Default::default()).durability(durability).new(db);
+    }
+}
+
+#[salsa::tracked]
+impl ProcMacros {
+    /// Incrementality query to prevent queries from directly depending on [`Self::get`].
+    #[salsa::tracked(returns(as_ref))]
+    pub fn get_for_crate(db: &dyn ExpandDatabase, krate: Crate) -> Option<Arc<CrateProcMacros>> {
+        Self::get(db).by_crate(db).get(&krate).cloned()
     }
 }
 
@@ -276,7 +301,7 @@ impl CustomProcMacroExpander {
                 ExpandError::new(call_site, ExpandErrorKind::MacroDisabled),
             ),
             id => {
-                let proc_macros = match db.proc_macros_for_crate(def_crate) {
+                let proc_macros = match ProcMacros::get_for_crate(db, def_crate) {
                     Some(it) => it,
                     None => {
                         return ExpandResult::new(
@@ -346,11 +371,4 @@ impl CustomProcMacroExpander {
             }
         }
     }
-}
-
-pub(crate) fn proc_macros_for_crate(
-    db: &dyn ExpandDatabase,
-    krate: Crate,
-) -> Option<Arc<CrateProcMacros>> {
-    db.proc_macros().get(krate)
 }
--- a/crates/query-group-macro/tests/hello_world.rs
+++ b/crates/query-group-macro/tests/hello_world.rs
@@ -4,13 +4,13 @@ use query_group_macro::query_group;
 mod logger_db;
 use logger_db::LoggerDb;
 
+#[salsa::input(singleton)]
+struct InputString {
+    inner: String,
+}
+
 #[query_group]
 pub trait HelloWorldDatabase: salsa::Database {
-    // input
-    // // input with no params
-    #[salsa::input]
-    fn input_string(&self) -> String;
-
     // unadorned query
     #[salsa::invoke_interned(length_query)]
     fn length_query(&self, key: ()) -> usize;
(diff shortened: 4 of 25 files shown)


ada4a:
Left some stylistic questions in the code, mainly for future such PRs

crates/rust-analyzer/src/reload.rs
```diff
@@ -109,13 +112,20 @@ impl GlobalState {
...
-        if self.analysis_host.raw_database().expand_proc_attr_macros()
-            != self.config.expand_proc_attr_macros()
-        {
-            self.analysis_host.raw_database_mut().set_expand_proc_attr_macros_with_durability(
-                self.config.expand_proc_attr_macros(),
-                Durability::HIGH,
-            );
+        // TODO: `ide`'s Cargo.toml says that it should avoid depending on `hir-*` crates directly,
+        // and go through `hir` re-exports instead. `rust-analyzer` OTOH does seems to depend on `hir-*`
+        // after all. So, should I import `ExpandProcAttrMacros` through `hir` or `hir_def`?
```
Veykril: Generally speaking we should at the very minimum have a re-export from `hir` instead then. Though even there the question then arises whether we shouldn't have a more "stable" interface for the given thing. We have not been very dilligent with keeping the separation maintained at the moment.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22639/cdab17b9fc0b312e2a32c351fbe3cd938557a44a..c0dfbc45163123b6245ed0ee0919a4e753b39389)*

ada4a:
Let me draft this while I investigate the test failures...

ada4a:
- in `hir_def`'s `macro_expansion_tests::mbe` tests, there are several ones which have `+syntaxctxt` included, and they fail because the numeric SyntaxContext ids (?) have changed. Though the ids all seem to change consistently (17408->16384)
- in `hir_def`'s `hir-def expr_store::tests::body::block::nested_module_scoping`, the salsa id, crate id, and the block id of a module have changed
- in `ide_db'`s `symbol_index::tests`, the module id, `EditionedFileId`s, and struct ids have changed -- again, consistently (everything fell by `0x400`)

While I have no idea why those ids would change as a result of this PR, the fact that the changes are all consistent seems like a good sign?

Veykril:
Its because this change changes the salsa ingredient order which thus changes the page allocator and IDs of the allocation

ada4a:
So.. it's fine? Should I just update the tests?

Veykril:
Oh yea its fine and expected, meant to add that, so just update the tests

crates/rust-analyzer/src/reload.rs
```diff
@@ -109,13 +112,20 @@ impl GlobalState {
...
+        // TODO: should I create `AnalysisHost::{update_,}expand_proc_attr_macros` to clean this up?
+        // TODO: we don't use `raw_db` mutably until we go into the `if` -- is it a problem that I
+        // still get it mutably in an eager manner? I did this to avoid needing two separate calls, to
+        // `raw_database` and `raw_database_mut`. Or maybe I should just inline `raw_db`?
```
ChayimFriedman2: the `rust-analyzer` crate does depend on analysis crates, but we should avoid adding new such dependencies where possible. But I also don't think exposing the struct wholesale is a good idea; instead, export a `fn set_expand_proc_attr_macros(value: bool)`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22639/cdab17b9fc0b312e2a32c351fbe3cd938557a44a..5f7510d511c5f8b5e963d8f2ca8d3f9c96370f49)*

ChayimFriedman2 (changes requested):

crates/rust-analyzer/src/reload.rs
ada4a: Added a setter and a getter, but called the former `set_expand_proc_attr_macros_with_durability`, to match `base_db::set_all_crates_with_durability`. The change is in a separate commit to simplify review, but I'd like to squash the commit before merging

Veykril:
Also coming back to the IDs in tests changing, ideally we wouldn't print those in tests at all / would "normalize" them such that they are stable wrt to the test, not to the salsa databass state, but no one has gotten around to changing that (also unsure how easy that is)

ada4a:
Should I maybe combine this PR and https://github.com/rust-lang/rust-analyzer/pull/22641, and rename this to something like "internal: remove input queries from query_group", to match https://github.com/rust-lang/rust-analyzer/pull/22328? Since that's what these two PRs are really intending to do

crates/rust-analyzer/src/reload.rs
ChayimFriedman2: I think we should have only a setter, with no durability (that is, it sets the durability itself), that checks for equality with the previous value like here. `hir` is a high-level interface.

ada4a: Agree with the proposal to change the setter, but I don't think we can quite get rid of the getter, because it has one legitimate user:

https://github.com/rust-lang/rust-analyzer/blob/c0dfbc45163123b6245ed0ee0919a4e753b39389/crates/hir-def/src/nameres/collector.rs#L619

ChayimFriedman2: That's in `hir-def`, it can access the input struct directly.

Veykril:
yea re-using one of these for both changes seems fine to me

crates/query-group-macro/tests/hello_world.rs
```diff
@@ -32,98 +32,83 @@ pub trait HelloWorldDatabase: salsa::Database {
...
         [
             "salsa_event(WillCheckCancellation)",
-            "salsa_event(WillExecute { database_key: create_data_HelloWorldDatabase(Id(0)) })",
-            "salsa_event(WillCheckCancellation)",
-            "salsa_event(DidValidateMemoizedValue { database_key: create_data_HelloWorldDatabase(Id(0)) })",
-            "salsa_event(WillCheckCancellation)",
-            "salsa_event(WillExecute { database_key: invoke_length_query_shim(Id(800)) })",
+            "salsa_event(WillExecute { database_key: create_data_HelloWorldDatabase(Id(400)) })",
             "salsa_event(WillCheckCancellation)",
+            "salsa_event(WillExecute { database_key: invoke_length_query_shim(Id(c00)) })",
```
ada4a, line 84: are these changes to be expected?..

ada4a (commented):
I went ahead and removed the support for `#[salsa::input]` after all, because it's nice to have the query-group-macro tests reflect the most current way of writing query groups (and just to test my own understanding).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22639/a77548188dbecc332bdc292b01a5077ec770cd98..5834f7a0c3850135086b60db58a038a6aca14f73)*

crates/query-group-macro/tests/hello_world.rs
Veykril: Yea the query group macro does a bit extra work i think thats redundant

crates/hir-expand/src/db.rs
```diff
@@ -50,10 +50,6 @@ pub enum TokenExpander<'db> {
 
 #[query_group::query_group]
 pub trait ExpandDatabase: SourceDatabase {
-    /// The proc macros. Do not use this! Use `proc_macros_for_crate()` instead.
```
ada4a, line 53: I think it'd make sense to move this comment.. somewhere. Not sure where though. Maybe on the `ProcMacros` struct?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22639/a77548188dbecc332bdc292b01a5077ec770cd98..5834f7a0c3850135086b60db58a038a6aca14f73)*

crates/hir-expand/src/db.rs
```diff
@@ -50,10 +50,6 @@ pub enum TokenExpander<'db> {
...
-    #[salsa::input]
-    fn proc_macros(&self) -> Arc<ProcMacros>;
-
     /// Incrementality query to prevent queries from directly depending on `ExpandDatabase::proc_macros`.
```
ada4a: Well, this function no longer exists. Should I replace this with `ProcMacros`, or `ProcMacros::get` maybe?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22639/a77548188dbecc332bdc292b01a5077ec770cd98..5834f7a0c3850135086b60db58a038a6aca14f73)*

ada4a:
> Also coming back to the IDs in tests changing, ideally we wouldn't print those in tests at all / would "normalize" them such that they are stable wrt to the test, not to the salsa databass state, but no one has gotten around to changing that (also unsure how easy that is)

I would be keen to work on this, but I'd need guidance on how to perform this normalization. Do you have any pointers?

crates/hir-expand/src/db.rs
Veykril: This function we should make an associated function of `ProcMacros` as well I think. The equivalent of that previous `proc_macros` function here is now `ProcMacros::get` fyi, that's the one that should not be used directly.

I think we can just turn `ProcMacros::get_for_crate` into a query, remove `proc_macros_for_crate` and replace all callers with that `get_for_crate` instead. 

```rust
#[salsa::tracked]
impl ProcMacros {
	#[salsa::tracked(returns(as_ref)]
    fn get_for_crate(&self, db: &dyn ExpandDatabase, krate: Crate) -> Option<Arc<CrateProcMacros>> {
        self.by_crate(db).get(&krate).cloned()
    }
}
```

ada4a: Got it. Should I do it in this PR, or leave as a follow-up? (both are fine by me)

Veykril:
> I would be keen to work on this, but I'd need guidance on how to perform this normalization. Do you have any pointers?

I am not sure tbh, I have not really looked into that. It might be that in some tests we are just printing unnecessary info, though I reckon in most we are interested in some kind of identification. 

Generally the thing I envisioned is to instead somehow map these page based IDs (salsa uses a page allocator which means not all IDs of one type of thing is sequential, as in things will not start with id 0, 1, 2, ..., instead they all are prefixed at some page offset), into numbers that are sequential. Unsure how we could best do that though.

crates/hir-expand/src/db.rs
ada4a: Now I kind of regret having merged these two PRs... I think I could've done one cleaning up `proc_macros{,_for_crate}`, and another one removing `expand_proc_attr_macros`+input query support

ada4a: But wait: as written, `ProcMacros::get_for_crate` still requires the caller to create an instance of `ProcMacros` to pass as `self`, i.e. call `ProcMacros::get`. Should I maybe change it like this?
```rs
#[salsa::tracked]
impl ProcMacros {
	#[salsa::tracked(returns(as_ref)]
    fn get_for_crate(/* no more self */ db: &dyn ExpandDatabase, krate: Crate) -> Option<Arc<CrateProcMacros>> {
        Self::get(db).by_crate(db).get(&krate).cloned()
     // ^^^^^^^^^^^^^ changed
    }
}
```
That would call `Self::get` inside the query, which should encapsulate it, and thus ensure incrementality? If my understanding is correct

Veykril: Yes the call should be in there, so it should be an associated function but not a method correct

ada4a: So I tried implementing it, and it somehow looks less confusing as one single commit removing both `proc_macros` and `proc_macros_for_crate`, since we don't need to the intermediary dance with `proc_macros_for_crate` (the free-standing fn) -- I would maybe just squash those changes in?

<details><summary>The combined diff</summary>
<p>

```patch
Commit ID: 9b29d271f846dfa125f5e65dd8b3ba694c760e8c
Change ID: qzrnpumvvorsmrvyyustppzqrruuqqso
Author   : Ada Alakbarova <ada.alakbarova@ferrous-systems.com> (2026-06-25 17:40:19)
Committer: Ada Alakbarova <ada.alakbarova@ferrous-systems.com> (2026-06-28 13:16:29)

    move `ExpandDatabase::{proc_macros,proc_macros_for_crate}`  out of the query group

diff --git a/crates/hir-expand/src/change.rs b/crates/hir-expand/src/change.rs
index 3959741e6f..22dc2cbd71 100644
--- a/crates/hir-expand/src/change.rs
+++ b/crates/hir-expand/src/change.rs
@@ -2,7 +2,6 @@
 //! state. Changes are transactional.
 use base_db::{CrateGraphBuilder, FileChange, SourceRoot, salsa::Durability};
 use span::FileId;
-use triomphe::Arc;
 
 use crate::{db::ExpandDatabase, proc_macro::ProcMacrosBuilder};
 
@@ -16,12 +15,13 @@
     pub fn apply(self, db: &mut impl ExpandDatabase) {
         let crates_id_map = self.source_change.apply(db);
         if let Some(proc_macros) = self.proc_macros {
-            let proc_macros = proc_macros.build(
+            proc_macros.build_in(
+                db,
+                Durability::HIGH,
                 crates_id_map
                     .as_ref()
                     .expect("cannot set proc macros without setting the crate graph too"),
             );
-            db.set_proc_macros_with_durability(Arc::new(proc_macros), Durability::HIGH);
         }
     }
 
diff --git a/crates/hir-expand/src/db.rs b/crates/hir-expand/src/db.rs
index 33672d10fa..74e27be5a9 100644
--- a/crates/hir-expand/src/db.rs
+++ b/crates/hir-expand/src/db.rs
@@ -17,7 +17,7 @@
     declarative::DeclarativeMacroExpander,
     fixup::{self, SyntaxFixupUndoInfo},
     hygiene::{span_with_call_site_ctxt, span_with_def_site_ctxt, span_with_mixed_site_ctxt},
-    proc_macro::{CrateProcMacros, CustomProcMacroExpander, ProcMacros},
+    proc_macro::CustomProcMacroExpander,
     span_map::{ExpansionSpanMap, RealSpanMap, SpanMap},
     tt,
 };
@@ -50,14 +50,6 @@
 
 #[query_group::query_group]
 pub trait ExpandDatabase: SourceDatabase {
-    /// The proc macros. Do not use this! Use `proc_macros_for_crate()` instead.
-    #[salsa::input]
-    fn proc_macros(&self) -> Arc<ProcMacros>;
-
-    /// Incrementality query to prevent queries from directly depending on `ExpandDatabase::proc_macros`.
-    #[salsa::invoke(crate::proc_macro::proc_macros_for_crate)]
-    fn proc_macros_for_crate(&self, krate: Crate) -> Option<Arc<CrateProcMacros>>;
-
     #[salsa::invoke(ast_id_map)]
     #[salsa::transparent]
     fn ast_id_map(&self, file_id: HirFileId) -> &AstIdMap;
diff --git a/crates/hir-expand/src/lib.rs b/crates/hir-expand/src/lib.rs
index 9c5714be2d..3f94f03647 100644
--- a/crates/hir-expand/src/lib.rs
+++ b/crates/hir-expand/src/lib.rs
@@ -53,7 +53,7 @@
     },
     db::ExpandDatabase,
     mod_path::ModPath,
-    proc_macro::{CustomProcMacroExpander, ProcMacroKind},
+    proc_macro::{CustomProcMacroExpander, ProcMacroKind, ProcMacros},
     span_map::{ExpansionSpanMap, SpanMap},
 };
 
@@ -176,7 +176,7 @@
                 kind: RenderedExpandError::DISABLED,
             },
             &ExpandErrorKind::MissingProcMacroExpander(def_crate) => {
-                match db.proc_macros_for_crate(def_crate).as_ref().and_then(|it| it.get_error()) {
+                match ProcMacros::get_for_crate(db, def_crate).and_then(|it| it.get_error()) {
                     Some(e) => RenderedExpandError {
                         message: e.to_string(),
                         error: e.is_hard_error(),
diff --git a/crates/hir-expand/src/proc_macro.rs b/crates/hir-expand/src/proc_macro.rs
index 467eae3122..0ec0a7179a 100644
--- a/crates/hir-expand/src/proc_macro.rs
+++ b/crates/hir-expand/src/proc_macro.rs
@@ -7,6 +7,7 @@
 use base_db::{Crate, CrateBuilderId, CratesIdMap, Env, ProcMacroLoadingError};
 use intern::Symbol;
 use rustc_hash::FxHashMap;
+use salsa::{Durability, Setter};
 use span::Span;
 use triomphe::Arc;
 
@@ -83,14 +84,24 @@
         );
     }
 
-    pub(crate) fn build(self, crates_id_map: &CratesIdMap) -> ProcMacros {
+    /// Builds [`ProcMacros`] and adds id to `db` with the provided `durability`.
+    pub(crate) fn build_in(
+        self,
+        db: &mut dyn ExpandDatabase,
+        durability: Durability,
+        crates_id_map: &CratesIdMap,
+    ) {
         let mut map = self
             .0
             .into_iter()
             .map(|(krate, proc_macro)| (crates_id_map[&krate], proc_macro))
             .collect::<FxHashMap<_, _>>();
         map.shrink_to_fit();
-        ProcMacros(map)
+        ProcMacros::try_get(db)
+            .unwrap_or_else(|| ProcMacros::new(db, Default::default()))
+            .set_by_crate(db)
+            .with_durability(durability)
+            .to(map);
     }
 }
 
@@ -107,11 +118,24 @@
 #[derive(Debug, PartialEq, Eq)]
 pub struct CrateProcMacros(StoredProcMacroLoadResult);
 
-#[derive(Default, Debug)]
-pub struct ProcMacros(FxHashMap<Crate, Arc<CrateProcMacros>>);
-impl ProcMacros {
-    fn get(&self, krate: Crate) -> Option<Arc<CrateProcMacros>> {
-        self.0.get(&krate).cloned()
+#[salsa::input(singleton, debug)]
+pub struct ProcMacros {
+    #[returns(ref)]
+    pub by_crate: FxHashMap<Crate, Arc<CrateProcMacros>>,
+}
+
+impl ProcMacros {
+    pub fn init_default(db: &dyn ExpandDatabase, durability: Durability) {
+        _ = Self::builder(Default::default()).durability(durability).new(db);
+    }
+}
+
+#[salsa::tracked]
+impl ProcMacros {
+    /// Incrementality query to prevent queries from directly depending on [`Self::get`].
+    #[salsa::tracked(returns(as_deref))]
+    pub fn get_for_crate(db: &dyn ExpandDatabase, krate: Crate) -> Option<Arc<CrateProcMacros>> {
+        Self::get(db).by_crate(db).get(&krate).cloned()
     }
 }
 
@@ -276,7 +300,7 @@
                 ExpandError::new(call_site, ExpandErrorKind::MacroDisabled),
             ),
             id => {
-                let proc_macros = match db.proc_macros_for_crate(def_crate) {
+                let proc_macros = match ProcMacros::get_for_crate(db, def_crate) {
                     Some(it) => it,
                     None => {
                         return ExpandResult::new(
@@ -347,10 +371,3 @@
         }
     }
 }
-
-pub(crate) fn proc_macros_for_crate(
-    db: &dyn ExpandDatabase,
-    krate: Crate,
-) -> Option<Arc<CrateProcMacros>> {
-    db.proc_macros().get(krate)
-}
diff --git a/crates/ide-db/src/lib.rs b/crates/ide-db/src/lib.rs
index 87cbc7c533..04d9c79a8f 100644
--- a/crates/ide-db/src/lib.rs
+++ b/crates/ide-db/src/lib.rs
@@ -65,7 +65,7 @@
 };
 use hir::{
     FilePositionWrapper, FileRangeWrapper,
-    db::{DefDatabase, ExpandDatabase, HirDatabase},
+    db::{DefDatabase, HirDatabase},
 };
 use triomphe::Arc;
 
@@ -203,7 +203,7 @@
         // This needs to be here otherwise `CrateGraphBuilder` will panic.
         set_all_crates_with_durability(&mut db, std::iter::empty(), Durability::HIGH);
         CrateGraphBuilder::default().set_in_db(&mut db);
-        db.set_proc_macros_with_durability(Default::default(), Durability::MEDIUM);
+        hir::ProcMacros::init_default(&db, Durability::MEDIUM);
         _ = base_db::LibraryRoots::builder(Default::default())
             .durability(Durability::MEDIUM)
             .new(&db);
```

</p>
</details> 

On a somewhat unrelated note: it would be great if we could mark `ProcMacros::get` as not recommended to use, and suggested `ProcMacros::get_for_crate` instead -- does salsa offer something for that? I guess, again, putting a comment on `ProcMacros` itself would be the next best thing...

Veykril: we can just use `#[deprecated = "reason"]` for this, then `#[expect]` on the allowed call sites, we already do that for other things

ada4a: Sure, but how do I put an attribute on `get`, given that it's generated by the proc macro?

Veykril: Right, I don't think it offers the knobs necessary there, so yea doc comment it is for now

crates/hir-def/src/db.rs
```diff
@@ -47,6 +44,26 @@ pub trait DefDatabase: ExpandDatabase + SourceDatabase {
     fn include_macro_invoc(&self, crate_id: Crate) -> Arc<[(MacroCallId, EditionedFileId)]>;
 }
 
+/// Whether to expand procedural macros during name resolution.
+///
+/// Note: this struct shouldn't be exposed to downstream crates -- consider using
```
ChayimFriedman2, line 49: "Downstream crates" is not entirely accurate, if `hir` or `hir-ty` need it (currently not but maybe in the future) this isn't a big problem. It shouldn't be exposed to **IDE** crates.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22639/8d38942ce80559c09548f6f88a0564d8c1fff6d2..a042ea8dc5079b5d18521515fc0412314b47d5db)*

crates/hir-expand/src/proc_macro.rs
```diff
@@ -83,14 +84,24 @@ impl ProcMacrosBuilder {
         );
     }
 
-    pub(crate) fn build(self, crates_id_map: &CratesIdMap) -> ProcMacros {
+    /// Builds [`ProcMacros`] and adds id to `db` with the provided `durability`.
+    pub(crate) fn build_in(
+        self,
+        db: &mut dyn ExpandDatabase,
+        durability: Durability,
```
ChayimFriedman2, line 91: If we never call this with a durability other than `HIGH`, it might make sense to remove this parameter.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22639/8d38942ce80559c09548f6f88a0564d8c1fff6d2..a042ea8dc5079b5d18521515fc0412314b47d5db)*

crates/query-group-macro/tests/hello_world.rs
```diff
@@ -4,13 +4,13 @@ use query_group_macro::query_group;
 mod logger_db;
 use logger_db::LoggerDb;
 
+#[salsa::input(singleton)]
```
ChayimFriedman2, line 7: When we remove features of the query_group macro, we should remove their tests, not change them to the idiomatic Salsa way. We shouldn't test Salsa.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22639/8d38942ce80559c09548f6f88a0564d8c1fff6d2..a042ea8dc5079b5d18521515fc0412314b47d5db)*

ChayimFriedman2 (approved):

crates/query-group-macro/tests/hello_world.rs
ada4a: AFAIU these test files don't just test the `#[salsa::input]` feature, but several others:
- `hello_world.rs`: `#[salsa::invoke_interned]` and `#[salsa::transparent]` queries
- `lru.rs`: `#[salsa::lru(n)]`
- `multiple_dbs.rs`: having multiple dbs
- `result.rs`: queries returning `Result`s?..
- `tuples.rs`: queries returning tuples

and so removing those tests completely wouldn't be advisable?..

I think the reason they all contain an input query is that you can't really call any other queries until you've provided some input for them to work on.

ChayimFriedman2: Okay, fine then.

ChayimFriedman2:
Thanks!

ada4a:
Really happy with how this turned out, thank you two for all the help:)

@ChayimFriedman2 I was planning to address your other two comments before merging though.. should I make a small follow-up PR?

ChayimFriedman2:
Feel free.

Issue #13, opened 2018-01-27
Get rid of `curly_block` method


matklad:
We have a `curly_block` method over here: https://github.com/matklad/libsyntax2/blob/357cd3358167daa38f3ff34d225e1501faff6015/src/parser/event_parser/parser.rs#L179.

This is a high-order function which eats a balanced paren sequence inside of `{}`, which helps quite a lot with error recovery, because all errors are contained in a block. However, this high order function does not help much, and I suggest just removing it and writing `while` loops by hand, like in here:

https://github.com/matklad/libsyntax2/blob/357cd3358167daa38f3ff34d225e1501faff6015/src/parser/event_parser/grammar/items.rs#L108-L114

Ideally, the error block should handle `{` specially (by parsing a balanced sequence), like it is done in Kotlin:
https://github.com/JetBrains/kotlin/blob/4d951de616b20feca92f3e9cc9679b2de9e65195/compiler/frontend/src/org/jetbrains/kotlin/parsing/KotlinParsing.java#L438

Issue #11, opened 2018-01-20
Make TokenSet faster


matklad:
We have a `TokenSet` abstraction over here: https://github.com/matklad/libsyntax2/blob/c8cf1d8cdac48f48caf9505bd5dc20dd2b962317/src/parser/event_parser/parser.rs#L70-L92

It is a set of `SyntaxKind`s, which is used during parsing for error recovery: the idea is to define TokenSets for things like `FIRST(item)` and then, when an error occurs, skip tokens until a token from this set is encountered.

- [ ] move `TokenSet` to a separate file
- [ ] use `[u64; 4]` to represet a bit-set of tokens
- [ ] add macros to declare and union `const TokenSet`

Plasticcaz:
Not sure if I'm biting off more than I can chew here, but I'd be willing to give this a go. I'm forking this and giving it my best shot!

matklad:
Looks like most of it was already fixed: https://github.com/matklad/libsyntax2/blob/751562d2f77a7bc3eeeffbf8e53c9f3515fa9653/crates/libsyntax2/src/parser_api.rs#L7-L32.

It would still be nice to move this to a separate file though!

Plasticcaz:
ah, okay. That should be fairly straight forward.

Plasticcaz:
Should we close this issue, or is there something else to do?

Issue #10, opened 2018-01-13
Should empty nodes be allowed?


matklad:
In principle, it is possible to have nodes of length zero: this would be an internal node without children.

However such nodes are annoying to work with. For example, with empty nodes there maybe arbitrary number of nodes at a given offset (as opposed to at most two for non-empty nodes). So it seems a great idea to forbid this node. 

That means that care must be taken to parse stuff like `use foo::92;` without empty nodes: we *know* that there should be a path segment after `::`, but we shouldn't create a node for it unconditionally.

A funny edge case here is an empty file: We'll have to create a file node for it, and it will be empty.

matklad:
For the time being, only the file node can be empty, and this works out OK

PR #22640, opened 2026-06-25
fix: Crash when hovering on anonymous consts

Wilfred:
Previously, the following code would cause rust-analyzer to infinitely recurse when hovering on `x`.

    trait Tr<const N: usize> {}
    pub fn f(x: impl Tr<{ 0 }>) {}

Attaching gdb and printing a stack trace shows that the problem is in HirDisplay for an anonymous const.

    #9  hir_ty::display::hir_fmt_generic_arguments () at crates/hir-ty/src/display.rs:1920
    #10 0x00005555563b9f94 in hir_ty::display::hir_fmt_generics () at crates/hir-ty/src/display.rs:1843
    #11 0x00005555563bd0ea in <hir_ty::next_solver::consts::Const as hir_ty::display::HirDisplay>::hir_fmt () at crates/hir-ty/src/display.rs:768
    #12 0x00005555563ba964 in <hir_ty::next_solver::generic_arg::GenericArg as hir_ty::display::HirDisplay>::hir_fmt () at crates/hir-ty/src/display.rs:716
    #13 hir_ty::display::hir_fmt_generic_arguments () at crates/hir-ty/src/display.rs:1920

Instead, just render anonymous consts as `{const}` rather than trying to render the synthetic type parameters too.

(We could track cycles like we do for recursive bounds, but that just produces a hover of `x: impl Tr<{const}<impl …>> + ?Sized` which doesn't seem terribly useful.)

AI disclosure: Initial implementation done with Codex and GPT-5.5.

### DIFF
--- a/crates/hir-ty/src/display.rs
+++ b/crates/hir-ty/src/display.rs
@@ -746,26 +746,33 @@ impl<'db> HirDisplay<'db> for Const<'db> {
             ConstKind::Value(value) => render_const_scalar_from_valtree(f, value.ty, value.value),
             ConstKind::Unevaluated(unev) => {
                 let c = unev.def.0;
-                match c {
-                    GeneralConstId::ConstId(id) => match &ConstSignature::of(f.db, id).name {
-                        Some(name) => {
-                            f.start_location_link(id.into());
-                            write!(f, "{}", name.display(f.db, f.edition()))?;
-                            f.end_location_link();
+                let generic_def = match c {
+                    GeneralConstId::ConstId(id) => {
+                        match &ConstSignature::of(f.db, id).name {
+                            Some(name) => {
+                                f.start_location_link(id.into());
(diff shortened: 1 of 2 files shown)


Wilfred:
FWIW I minimised this from a bigger codebase that was using smithay-client-toolkit 0.19.2.

ChayimFriedman2 (approved):
Please fix the issue link, then LGTM.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22640/cdab17b9fc0b312e2a32c351fbe3cd938557a44a..03aa33939cd13545cb404c466954a2ae8599e89b)*

Wilfred:
I think the bot was confused by the `#9` in the indented code block? I've fixed it anyway.

PR #22641, opened 2026-06-25
internal: move `ExpandDatabase::proc_macros` out of the query group

ada4a:
This required several non-trivial changes:
- Rename the existing `ProcMacros::get` to `ProcMacros::get_for_crate` (feel free to suggest a better name), to avoid clashing with the new getter created by salsa.
- Change `ProcMacroBuilder::build` to take `db` and `durability`, as it will need to build `ProcMacros` inside the db. Also rename it to `build_in`, to reflect the change in functionality.
- `ProcMacros` is now stored in the database directly, instead of being wrapped in an `Arc`. This should be fine though, since its `by_crate` field is marked with `#[returns(ref)]`.

@rustbot blocked on https://github.com/rust-lang/rust-analyzer/pull/22639, but mostly for easier rebasing later

cc @Veykril

### DIFF
--- a/crates/hir-expand/src/proc_macro.rs
+++ b/crates/hir-expand/src/proc_macro.rs
@@ -7,6 +7,7 @@ use std::{panic::RefUnwindSafe, sync};
 use base_db::{Crate, CrateBuilderId, CratesIdMap, Env, ProcMacroLoadingError};
 use intern::Symbol;
 use rustc_hash::FxHashMap;
+use salsa::{Durability, Setter};
 use span::Span;
 use triomphe::Arc;
 
@@ -83,14 +84,24 @@ impl ProcMacrosBuilder {
         );
     }
 
-    pub(crate) fn build(self, crates_id_map: &CratesIdMap) -> ProcMacros {
+    /// Builds [`ProcMacros`] and adds id to `db` with the provided `durability`.
+    pub(crate) fn build_in(
+        self,
+        db: &mut dyn ExpandDatabase,
+        durability: Durability,
+        crates_id_map: &CratesIdMap,
+    ) {
         let mut map = self
             .0
             .into_iter()
             .map(|(krate, proc_macro)| (crates_id_map[&krate], proc_macro))
             .collect::<FxHashMap<_, _>>();
         map.shrink_to_fit();
-        ProcMacros(map)
+        ProcMacros::try_get(db)
+            .unwrap_or_else(|| ProcMacros::new(db, Default::default()))
+            .set_by_crate(db)
+            .with_durability(durability)
+            .to(map);
     }
 }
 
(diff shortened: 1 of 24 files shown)


crates/hir-expand/src/proc_macro.rs
```diff
@@ -83,14 +84,24 @@ impl ProcMacrosBuilder {
...
             .into_iter()
             .map(|(krate, proc_macro)| (crates_id_map[&krate], proc_macro))
             .collect::<FxHashMap<_, _>>();
         map.shrink_to_fit();
-        ProcMacros(map)
+        ProcMacros::try_get(db)
+            .unwrap_or_else(|| ProcMacros::new(db, Default::default()))
+            .set_by_crate(db)
+            .with_durability(durability)
+            .to(map);
```
ada4a, line 104: The reason for this ugliness is twofold:
- I can't just do `ProcMacros::builder(map).durability(durability).new(db)`, because `ProcMacros` is marked as a singleton, and salsa complains (panics) if I create a new instance of it.
- I can't just use `ProcMacros::get`, because some tests panic with "`Option::unwrap` called on `None`", which I assume means that `ProcMacros` is sometimes not yet set by the time we call this method.

Let me know if I should add this justification as a comment to the code maybe

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22641/cdab17b9fc0b312e2a32c351fbe3cd938557a44a..e644310dfce2788b8b9ee6e12646bf8b169b300d)*

crates/ide-db/src/lib.rs
```diff
@@ -203,7 +200,7 @@ impl RootDatabase {
         // This needs to be here otherwise `CrateGraphBuilder` will panic.
         set_all_crates_with_durability(&mut db, std::iter::empty(), Durability::HIGH);
         CrateGraphBuilder::default().set_in_db(&mut db);
-        db.set_proc_macros_with_durability(Default::default(), Durability::MEDIUM);
+        _ = hir::ProcMacros::builder(Default::default()).durability(Durability::MEDIUM).new(&db);
```
Veykril: I think we can expose another function on `ProcMacros` here  like `init_default(&db, Durability::MEDIUM)` or so, otherwise looks good to me

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22641/cdab17b9fc0b312e2a32c351fbe3cd938557a44a..e644310dfce2788b8b9ee6e12646bf8b169b300d)*

ada4a: Done:)

ada4a:
Moved into https://github.com/rust-lang/rust-analyzer/pull/22639

PR #22645, opened 2026-06-26
feat: add fixes add '.await' for type_mismatch

A4-Tacks:
Example
---
```rust
async fn foo() -> u32 { 2 }
async fn test() {
    let x: u32 = foo()$0;
}
```

**Before this PR**

No available fixes

**After this PR**

```rust
async fn foo() -> u32 { 2 }
async fn test() {
    let x: u32 = foo().await;
}
```


Abusalah0:
issue: The `// XXX: maybe should check if ctx is async` note looks necessary for correctness.

I tested both a synchronous function and a non async closure inside an async function:

```rust
async fn foo() -> u32 {
    2
}

fn test() {
    let x: u32 = foo();
}
```

In both  invalid contexts, the `E0308` diagnostic offers `AssistId("add_await")`, producing `foo().await`, which is still invalid Rust.

an async block control case correctly receives the fix, so the future-output coercion check appears to work as intended; the missing piece seems to be validating that `.await` is legal at the expression’s location.

ChayimFriedman2:
@Abusalah0 Our assists are not required to emit correct code. The current behavior seems the best to me. Sometimes you'll want to add the async block after.

ChayimFriedman2 (approved):

Issue #22341, opened 2026-05-11
rust-analyzer panic: should have `ExpressionStore::expr_only`


kpqi5858:
**rust-analyzer version**: rust-analyzer 1.95.0 (5980761 2026-04-14)

**rustc version**: rustc 1.95.0 (59807616e 2026-04-14)

**editor or extension**: neovim

**relevant settings**:

**repository link (if public, optional)**: 

**code snippet to reproduce**:
```rust
struct Foo<T>(T);

impl<T> Foo<T> {
    fn foo<const A: = >() {}
}
```

```
thread 'Worker6' (702867) panicked at src/tools/rust-analyzer/crates/hir-ty/src/lower.rs:283:36:
should have `ExpressionStore::expr_only`
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: <hir_ty::lower::TyLoweringContext>::lower_const
   4: <<_ as hir_ty::db::HirDatabase>::generic_defaults_with_diagnostics::generic_defaults_with_diagnostics_shim::generic_defaults_with_diagnostics_shim_Configuration_ as salsa::function::Configuration>::execute
   5: <salsa::function::IngredientImpl<<_ as hir_ty::db::HirDatabase>::generic_defaults_with_diagnostics::generic_defaults_with_diagnostics_shim::generic_defaults_with_diagnostics_shim_Configuration_>>::execute
   6: <_ as hir_ty::db::HirDatabase>::generic_defaults_with_diagnostics::generic_defaults_with_diagnostics_shim
   7: <hir::GenericDef>::diagnostics
   8: <hir::AssocItem>::diagnostics
   9: <hir::Module>::diagnostics
   10: ide_diagnostics::semantic_diagnostics
   11: ide_diagnostics::full_diagnostics
   12: <ide::Analysis>::full_diagnostics
   13: rust_analyzer::handlers::request::handle_document_diagnostics
   14: <<stdx::thread::pool::Pool>::spawn<<rust_analyzer::task_pool::TaskPool<rust_analyzer::main_loop::Task>>::spawn<<rust_analyzer::handlers::dispatch::RequestDispatcher>::on_with_thread_intent<false, false, lsp_types::request::DocumentDiagnosticRequest>::{closure#0}>::{closure#0}>::{closure#0} as core::ops::function::FnOnce<()>>::call_once::{shim:vtable#0}

[...]

Issue #22175, opened 2026-04-24
rust-analyzer cries (sometimes) when it sees unclosed braces while typing


Lyude:
~~Since upgrading from rust-analyzer 1.90 to 1.94~~, trying to input an opening brace anywhere in rust source files appears to cause rust-analyzer to be upset and spit out this message:

(Update: it's not a regression. It's just completely broken on only this specific file apparently)

```
rust-analyzer: -32603: request handler panicked: should have `ExpressionStore::expr_only`
```

**rust-analyzer version**: rust-analyzer-1.94.1-1.fc43
**editor or extension**: neovim-0.11.6-1.fc43
**rustc version**: rustc 1.94.1 (e408947bf 2026-03-25) (Fedora 1.94.1-1.fc43)
**repository link**: https://gitlab.freedesktop.org/lyudess/linux/-/tree/rust-upstream/r-a-bug-20260428
**code snippet to reproduce**: just a brace :)

# Reproducer instructions

* Setup rust-analyzer with the linked kernel repository:

(to be clear, I committed the .config I'm using - so no need to run through kernel config stuff if you run everything below)

```
make rustavailable
make olddefconfig
make -j$(nproc) CLIPPY=1
make rust-analyzer
```

* Open rust/kernel/drm/gem/shmem.rs
* Try to input a opening brace anywhere in the file

# Additional info

~~The 1.90 version of rust-analyzer definitely doesn't seem to have this issue~~ yes it does

Lyude:
Oh dear, oh boy, this isn't fun.
turns out it's not a regression, and it's -just- this file that is doing this. I will update the bug info to reflect this, but it turns out it only happens on rust/kernel/drm/gem/shmem.rs despite the fact it seems to compile just fine.

ChayimFriedman2:
Can you give a repo containing a normal Cargo project, or at least a ready rust-project.json, without the need to download the whole kernel and `make` it? It will definitely ease the investigation (and make it more likely to actually happen).

Lyude:
@ChayimFriedman2 I'm not sure because I haven't tried to actually figure out what precisely in the file is causing it but I'll do my best and put this on my todo list, I probably can come up with something

debuggerdragon311:
I traced through the source and have a reasonable picture of what's going on, though I'd need a reproducer to confirm the exact trigger. Happy to take a crack at it if nobody else is already on it.

[...]

PR #22646, opened 2026-06-26
fix: Crash on lowering consts with associated types

Wilfred:
Previously we'd modify TyLoweringContext::store without modifying TyLoweringContext::def. This causes a crash later when we try to look up the expression in the wrong store:

    thread 'main' panicked at crates/hir-ty/src/infer/expr.rs:326:
    should have `ExpressionStore::expr_only`

Instead, ensure we update both consistently so we can find the expression.

Add a regression test (based on code originally seen in the bitmaps crate version 3.2.1).

AI disclosure: Bisected and coded with help from Codex and GPT-5.5.

### DIFF
--- a/crates/hir-ty/src/generics.rs
+++ b/crates/hir-ty/src/generics.rs
@@ -143,7 +143,7 @@ impl<'db> Generics<'db> {
         self.chain.iter()
     }
 
-    fn owner(&self) -> &SingleGenerics<'db> {
+    pub(crate) fn owner(&self) -> &SingleGenerics<'db> {
         self.chain.last().expect("must have an owner params")
     }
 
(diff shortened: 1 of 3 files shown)


Wilfred:
Issues that look related: #22175, #22341.

This was found with me bisecting another SCIP issue on some large codebases, it keeps finding ~cursed~ interesting patterns :)

The code looks sensible, I think the invariant it's preserving is the right thing to do, but let me know if there's a better way.

ChayimFriedman2 (commented):
Maybe have a `TyLoweringContext::set_owner(owner: &SingleGenerics)`?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22646/a77548188dbecc332bdc292b01a5077ec770cd98..58aded7bcb06c385a36e45d0ffa055fa2512b7b2)*

Wilfred:
Done. I originally wanted to keep the change as small as possible, but I agree a helper is probably nicer API.

ChayimFriedman2 (approved):

PR #22655, opened 2026-06-28
fix: resolve each definition when a use statement imports multiple items

### LINKED ISSUE #22632: Rust hover shows `mod draw` for `use super::draw` that resolves to function
# code:
```Rust
mod outer {
    // This is the function that `super::draw` should resolve to.
    pub fn draw(_: i32) {}

    mod inner {
        use super::draw;

        pub fn test() {
            draw(1);
        }
    }
}

// This module has the same name, but it is not the target of `super::draw`.
mod draw {}
```
# explanation:
Expected: hover on `draw` in `use super::draw` shows the function `outer::draw`. Actual: hover shows `mod draw`. Go to definition resolves correctly.

I'm not sure whether this is a Zed issue or a rust-analyzer issue, but the hover information looks inconsistent with go-to-definition.

Hmikihiro:
When it finds definitions on a NameRef of a use statement, it returns multiple definitions as NameRefClass::DefinitionPerNs. Each definition is registered in the vector of definiotns.

In goto_definition, hover and highlights, It show multiple items in use statement.

Fixes rust-lang/rust-analyzer#22632

AI-disclosure: Antigravity

スクリーンショット 2026-06-28 13 31 19
スクリーンショット 2026-06-28 13 31 50

### DIFF
--- a/crates/hir/src/source_analyzer.rs
+++ b/crates/hir/src/source_analyzer.rs
@@ -1484,7 +1484,7 @@ impl<'db> SourceAnalyzer<'db> {
         let hir_path =
             collector.lower_path(path.clone(), &mut ExprCollector::impl_trait_error_allocator)?;
         let (store, _) = collector.store.finish();
-        Some(resolve_hir_path_(
+        let PathResolutionPerNs { type_ns, value_ns, macro_ns } = resolve_hir_path_(
             db,
             &self.resolver,
             self.infer_body,
@@ -1493,7 +1493,11 @@ impl<'db> SourceAnalyzer<'db> {
             name_hygiene(db, InFile::new(self.file_id, path.syntax())),
             Some(&store),
             true,
-        ))
+        );
+        if type_ns.is_none() && value_ns.is_none() && macro_ns.is_none() {
+            return None;
+        }
+        Some(PathResolutionPerNs { type_ns, value_ns, macro_ns })
     }
 
     pub(crate) fn record_literal_missing_fields(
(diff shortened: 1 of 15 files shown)


crates/hir/src/source_analyzer.rs
```diff
@@ -1956,11 +1960,13 @@ fn resolve_hir_path_(
     };
 
     if resolve_per_ns {
-        PathResolutionPerNs {
-            type_ns: types().or_else(items),
-            value_ns: values(),
-            macro_ns: macros(),
+        let type_ns = types().or_else(items);
+        if let Some(PathResolution::Def(ModuleDef::Module(module))) = type_ns
+            && module.is_crate_root(db)
```
ChayimFriedman2, line 1965: That `if` is definitely incorrect. Why did you add it?

ChayimFriedman2 (changes requested):
The right way, I'm pretty sure about that, is **not** to do add a different variant to `NameRefClass`. What it is - I don't know. It might be changing `Definition` to include all namespaces, or returning multiple `NameRefClass`es from `classify()`, or having a new method, or something else entirely.

Also note that @dfireBird has already claimed https://github.com/rust-lang/rust-analyzer/issues/19421, which is highly related (although not the same), you might want to discuss their approach with them.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22655/5d15d9f5fd939bd8f121ae878d55b30e6a992a32..58ad2b7d3c674b5d2eb5846bebec5807792a8c4f)*

crates/hir/src/source_analyzer.rs
Hmikihiro: Sorry. I think that fix was just a stopgap measure. The issue is that the system is searching for and finding the `value` inside the crate at the same level as `foo`, which represents the crate, so I'll look into it and fix that part as well.

https://github.com/rust-lang/rust-analyzer/blob/ef691be44d10f57bae44f9ffb5dbd0c90e6d924f/crates/ide/src/hover/tests.rs#L6869-L6887

````
*foo*

```rust
extern crate foo
```

---

```rust
foo
```

```rust
pub fn foo()
```
````

Comment (ChayimFriedman2) [reply crates/hir/src/source_analyzer.rs:1965]: This is interesting. In analysis we won't resolve this to the function because it wasn't recorded yet in the module, but in IDE resolution we have everything recorded.

I don't know if we can solve this, but I don't think this justifies this `if`. It can also lead to wrong resolution, e.g.:
```rust
use crate as foo;
fn foo() {}

mod bar {
    use super::foo$0;
}
```

ChayimFriedman2: Although, how do we even get to this function in this case? It should call `resolve_hir_path_qualifier()` instead.

PR #22658, opened 2026-06-28
internal: move `ExpandDatabase::syntax_context` out of the query group

ada4a:

I thought that, rather than removing the whole query group in one big PR (which could complicate reviewing), I'd chip away at a small number of related queries at a time.

This PR changes `ExpandDatabase::syntax_context` to be a method on `HirFileId`, as it's a mapping from `HirFileId` to `SyntaxContext`, which uses `Edition` as an additional input.

cc @Veykril

### DIFF
--- a/crates/hir-expand/src/db.rs
+++ b/crates/hir-expand/src/db.rs
@@ -2,7 +2,7 @@
 
 use base_db::{Crate, SourceDatabase};
 use mbe::MatchedArmIndex;
-use span::{AstIdMap, Edition, Span, SyntaxContext};
+use span::{AstIdMap, Span};
 use std::borrow::Cow;
 use syntax::{AstNode, Parse, SyntaxError, SyntaxNode, SyntaxToken, T, ast};
 use syntax_bridge::{DocCommentDesugarMode, syntax_node_to_token_tree};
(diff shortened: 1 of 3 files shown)


Veykril:
Yea chipping away is definitely preferred

crates/hir-expand/src/lib.rs
```diff
@@ -1107,6 +1107,7 @@ impl From<MacroCallId> for HirFileId {
     }
 }
 
+#[salsa::tracked]
```
Veykril: One nit, lets split out a new `impl HirFileId` just for tracked queries. Reason being that the proc-macro has to parse less then. That is a separate 

```
#[salsa::tracked]
impl HirFileId {
     // only #[salsa::tracked] functons in this impl blocks
}
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22658/5d15d9f5fd939bd8f121ae878d55b30e6a992a32..3192a949d75c1409e5c9a3d0a1396210d6c6a667)*

ada4a: Oh, okay, I was wondering which way is preferred. Done!

Veykril (approved):

crates/hir-expand/src/db.rs
```diff
@@ -136,19 +136,6 @@ pub trait ExpandDatabase: SourceDatabase {
         &self,
         macro_call: MacroCallId,
     ) -> Option<ExpandResult<Arc<[SyntaxError]>>>;
-
-    #[salsa::transparent]
```
Veykril, line 140: Oh, good thing the test caught this cause I missed this. This is a "transparent query", in otherwords, this itself isn't actually a query. Its just a function, so we don't need to put `salsa::tracked` on the associated function now. (the actual query there is the call to `macro_arg_considering_derives`. Reason for this itself not being one is becuase the `SyntaxContext::root(edition)` call itself is unnecessary to cache.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22658/5d15d9f5fd939bd8f121ae878d55b30e6a992a32..54c6ab74e9f5ed8c6fd26b6487fc5f7f9c6b45dd)*

ada4a: Right, I was confused about which things we do/do not mark as tracked, but the explanation makes sense. Done!

Veykril (approved):

PR #22664, opened 2026-06-29
vscode: nest submodules under mod.rs by default

mansiverma897993:
This PR configures VS Code's file nesting to group other Rust files under their corresponding `mod.rs` file.

When `explorer.fileNesting.enabled` is active, this effectively sorts `mod.rs` to the top of directories by nesting its submodules under it.

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -3344,7 +3344,8 @@
                 }
             },
             "explorer.fileNesting.patterns": {
-                "Cargo.toml": "Cargo.lock"
+                "Cargo.toml": "Cargo.lock",
+                "mod.rs": "*.rs"
             }
         },
         "problemPatterns": [
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -3344,8 +3344,7 @@
                 }
             },
             "explorer.fileNesting.patterns": {
-                "Cargo.toml": "Cargo.lock",
-                "mod.rs": "*.rs"
+                "Cargo.toml": "Cargo.lock"
             }
         },
         "problemPatterns": [


ChayimFriedman2:
Nesting is controversial (I for example won't like it), the issue talked about sorting, which should be uncontroversial.

flodiebold:
Yeah, I think nesting is more for secondary files to avoid clutter (e.g. files generated out of one source file), not for something like this.

mansiverma897993:
Hi! Since VS Code does not expose any Extension API to customize the native file explorer's sort order, it is not possible to implement a native sorting rule specifically for `mod.rs` files within the `rust-analyzer` extension.

We initially attempted to use VS Code's native file nesting (`explorer.fileNesting.patterns`) to group files under `mod.rs` to place it at the top of the group. However, as pointed out, file nesting is controversial and not intended for primary source files/module entry points.

Given these platform limitations, there is no viable solution to natively sort `mod.rs` to the top of directories in the standard VS Code Explorer. I have reverted the nesting changes and will close this PR.

As workarounds, developers can:
1\. Adopt the modern Rust 2018 module structure (using `foo.rs` and `foo/` subdirectories instead of `foo/mod.rs`), which naturally avoids the `mod.rs` sorting clutter.
2\. Use VS Code's `workbench.editor.customLabels.patterns` to make open `mod.rs` files identifiable in editor tabs:
   ```json
   "workbench.editor.customLabels.patterns": {
       "**/mod.rs": "${dirname}/mod.rs"
   }
   ```
3\. Use a third-party VS Code extension like 'Custom Explorer Sort' if custom explorer sorting is required.

mansiverma897993:
@flodiebold @ChayimFriedman2 

> "Thanks for the feedback! Since native sorting isn't supported by VS Code and nesting is controversial, I've reverted the changes and closed the PR. I'll stick to the recommended workarounds (like custom labels or the modern module structure)."

ChayimFriedman2:
I admit I don't know if it's possible, but I believe there are extensions for custom sorting, so it should be possible.

PR #22668, opened 2026-06-29
Add tree top method to Syntax node

bit-aloo:
This PR adds tree top method to syntax node, to fetch the root, instead of letting the user do that everytime.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2359,9 +2359,9 @@ checksum = "dc897dd8d9e8bd1ed8cdad82b5966c3e0ecae09fb1907d58efaa013543185d0a"
 
 [[package]]
 name = "rowan"
-version = "0.15.18"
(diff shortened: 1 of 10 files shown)


ChayimFriedman2:
I actually prefer this to be a method in rowan (on both `SyntaxNode` and `SyntaxToken`), feel free to send a PR there (including bumping the version then I'll release it).

ChayimFriedman2:
Ah and you need to base this on the 0.15 version series, 0.16 has some changes that break us.

bit-aloo:
PR on rowan with  version bump: https://github.com/rust-analyzer/rowan/pull/213

bit-aloo:
I will update this PR, once we publish v0.15.19

ChayimFriedman2 (approved):

PR #22669, opened 2026-06-29
project-model: Pass config flags to cargo metadata / cargo config commands

emilio:
See https://bugzilla.mozilla.org/show_bug.cgi?id=2051235 for some of the discussion. If you want to point cargo to some extra configuration, this makes extraArgs able to do it.

### DIFF
--- a/crates/project-model/src/cargo_config_file.rs
+++ b/crates/project-model/src/cargo_config_file.rs
@@ -16,12 +16,14 @@ impl CargoConfigFile {
     pub(crate) fn load(
         manifest: &ManifestPath,
         extra_env: &FxHashMap<String, Option<String>>,
+        extra_args: &[String],
         sysroot: &Sysroot,
     ) -> Option<Self> {
         let mut cargo_config = sysroot.tool(Tool::Cargo, manifest.parent(), extra_env);
         cargo_config
             .args(["-Z", "unstable-options", "config", "get", "--format", "toml", "--show-origin"])
             .env("RUSTC_BOOTSTRAP", "1");
+        cargo_config.args(extra_args);
         if manifest.is_rust_manifest() {
             cargo_config.arg("-Zscript");
         }
--- a/crates/project-model/src/cargo_workspace.rs
+++ b/crates/project-model/src/cargo_workspace.rs
@@ -672,15 +672,15 @@ impl FetchMetadata {
 
         let mut other_options = vec![];
         // cargo metadata only supports a subset of flags of what cargo usually accepts, and usually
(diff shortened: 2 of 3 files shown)


ChayimFriedman2:
There is a complex (IMO too complex) machinery for handling the extraArgs for each command, I don't want to complicate it further. Especially here given that there is a simpler and better way: provide a setting to manually choose the Cargo config path.

emilio:
@chayimfriedman2 fwiw, I tend to agree, that said, extraArgs is documented to apply to all `cargo` commands, right? Is that not the intention? This certainly builds on that assumption.

In any case, I'd be happy to write a PR for some sort of `cargo.configPath` option I guess? Or should it be `cargo.extraConfigPaths` or so? Either way works for me I suppose. That introduces a different set of interactions, tho...

ChayimFriedman2:
> extraArgs is documented to apply to all `cargo` commands, right?

It should in theory, yes, but reality is more complicated that that. Cargo commands don't have the same set of supported flags, and that setting is older than more specific alternatives (applying to specific commands). So if we don't want to break people's setups, we need to do some filtering. We already have a very crude approximation for `cargo metadata` (only unstable flags are passed), and it causes issues. If we would be starting from scratch, I wouldn't have this setting to begin with, only settings for specific commands (see also [#t-compiler/rust-analyzer > Cargo extraArgs](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Cargo.20extraArgs/with/607252632)). I don't want to add any more filtering.

ChayimFriedman2:
So I'm closing this and please open the PR as you proposed. As for whether to support multiple configs, I'd say let's not begin with that. We can always do so later if needed. So `cargo.configPath: null | string`.

ChayimFriedman2:
Note also that to maintain consistency, we'd want to not just put this config path in `CargoConfigFile::load()` (which is our own extraction of data we need from the config), but also to all Cargo commands via the `--config` parameter.

emilio:
@ChayimFriedman2 yeah, agreed. While looking into adding `cargo.configPath` I realized:

 * I _kinda_ can get away without any changes by using `metadataExtraArgs`.
 * Passing the config file everywhere messes up with the sysroot discovery.

So a more narrow change would be #22676 (we're already not passing `extra_args`). But I'm happy to also add `cargo.configPath` if you want. Looking at #21964, it's unlikely that you would want to mess with the sysroot config, so I think #22676 would satisfy my use case.

emilio:
But let me know if you want `cargo.configPath` instead of / in addition to tweaking `metadataExtraArgs`.

PR #22676, opened 2026-06-30
project-model: Don't pass metadata extra args to sysroot.

emilio:
The same way we don't pass extraArgs. The sysroot config is not likely to work with the workspace config options.

### DIFF
--- a/crates/project-model/src/workspace.rs
+++ b/crates/project-model/src/workspace.rs
@@ -1953,7 +1953,7 @@ fn sysroot_metadata_config(
         features: Default::default(),
         targets,
         extra_args: Default::default(),
-        metadata_extra_args: config.metadata_extra_args.clone(),
+        metadata_extra_args: Default::default(),
         extra_env: config.extra_env.clone(),
         toolchain_version,
         kind: "sysroot",


emilio:
Context is that when trying to implement `cargo.configPath` as discussed in #22669, I realized that passing it down to the sysroot config would be wrong. E.g. for Firefox it'd try to do something like:

```
cd "/home/emilio/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library" && CARGO_RESOLVER_LOCKFILE_PATH="/tmp/rust-analyzer1852f-1/Cargo.lock" CARGO_TARGET_DIR="/home/emilio/src/moz/firefox/obj-debug" RUSTUP_AUTO_INSTALL="0" RUSTUP_TOOLCHAIN="/home/emilio/.rustup/toolchains/stable-x86_64-unknown-linux-gnu" __CARGO_TEST_CHANNEL_OVERRIDE_DO_NOT_USE_THIS="nightly" "/home/emilio/.cargo/bin/cargo" "metadata" "--format-version" "1" "--manifest-path" "/home/emilio/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/Cargo.toml" "--config" "/home/emilio/src/moz/firefox/obj-debug/.cargo/config.toml" "--filter-platform" "x86_64-unknown-linux-gnu" "-Zlockfile-path" "-Zunstable-options"
```

Where the `obj-debug/.cargo/config.toml` would use vendored sources which are obviously not present for the sysroot.

ChayimFriedman2 (approved):
This does seem better, yeah.

Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22676/ffc51e283a93c507e7a61b99a3149c1cf6d47192..e2073fbbd014955aef27bd5d186b26e41fc40845)*

emilio:
Thank you for the super fast turnaround!

PR #22677, opened 2026-06-30
fix: Assertion failure on closure with unbound function

Wilfred:
Previously that we asserted that callee inference always succeeded when FnOnce is defined. This isn't true when we have a closure that references an unbound function.

I originally noticed this when investigating a configuration where rust-analyzer couldn't see generated files from a library (e.g. a build.rs or buck equivalent), but rust-analyzer should be tolerant of unbound functions in general.

Remove the failing assertion. I've also added a unit test.

AI disclosure: Bisected by Opus 4.8, initial implementation by GPT 5.5.

### DIFF
--- a/crates/hir-ty/src/infer/callee.rs
+++ b/crates/hir-ty/src/infer/callee.rs
@@ -585,13 +585,7 @@ impl<'a, 'db> DeferredCallResolution<'db> {
                     method_callee.args,
                 );
             }
-            None => {
-                assert!(
-                    ctx.lang_items.FnOnce.is_none(),
-                    "Expected to find a suitable `Fn`/`FnMut`/`FnOnce` implementation for `{:?}`",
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
I must admit I don't understand (after little investigation) why rustc, which this code was copied from, does not suffer from the same ICE, but anyway even in rustc this code was changed recently (3 days ago) to not panic. So I'm approving this, but I think the `if` and the logging are redundant.

Wilfred:
Aha, I didn't realise it was recently fixed in rustc. Sound like it was an ICE there too, at least in some circumstances (my unit test doesn't cause an ICE though): https://github.com/rust-lang/rust/issues/157951

Wilfred:
Removed the logging.

ChayimFriedman2 (approved):

PR #22678, opened 2026-06-30
project-model: Introduce cargo.configExtraArgs.

emilio:
Much like cargo.metadataExtraArgs. This is necessary so that rust-analyzer can have a consistent view of the configuration in the project.


ChayimFriedman2:
I thought that I explicitly said that I *don't* want that. If you need a custom config, provide a setting to set its path.

emilio:
Yeah sorry. The point of the `extraMetadataFlags` etc was to provide a variety of `--config` args. But I guess we don't need to follow that pattern.

PR #22679, opened 2026-06-30
project-model: Introduce cargo.configPath

emilio:
As discussed in https://github.com/rust-lang/rust-analyzer/pull/22669 / https://github.com/rust-lang/rust-analyzer/pull/22678.

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -1688,6 +1688,8 @@ pub struct RunnablesConfig {
     pub override_cargo: Option<String>,
     /// Additional arguments for the `cargo`, e.g. `--release`.
     pub cargo_extra_args: Vec<String>,
+    /// Path to an extra config file passed to cargo via `--config`.
+    pub config_path: Option<AbsPathBuf>,
     /// Additional arguments for the binary being run, if it is a test or benchmark.
     pub extra_test_binary_args: Vec<String>,
     /// Subcommand used for doctest runnables instead of `test`.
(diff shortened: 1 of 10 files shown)


emilio:
@ChayimFriedman2 PTAL, I think it's useful to propagate it to flycheck but existing extra_args options don't, so lmk if you want to revert the second commit

ChayimFriedman2:
I do think flycheck should use it as well, yes. And FWIW it does use `extraArgs`.

emilio:
Oh right, I thought it was just using `check.extraArgs`, but it's using `check.extraArgs` on top of `cargo.extraArgs`. So should be fine to also use `cargo.configPath` here.

emilio:
The other place where we could plumb this is `RunnablesConfig`. That currently _does not_ use `cargo.extraArgs`, should I plumb this there?

emilio:
I added two commits for those but let me know if you want the first, both, or none :)

crates/rust-analyzer/src/config.rs
```diff
@@ -2333,6 +2339,12 @@ impl Config {
         extra_args
     }
 
+    pub fn runnables_extra_args(&self, source_root: Option<SourceRootId>) -> Vec<String> {
+        let mut extra_args = self.extra_args(source_root).clone();
```
ChayimFriedman2: I don't want to apply `extra_args()` to places they weren't used before. Especially runnables can have distinct commands. Like I said, in retrospect this setting was a mistake and we should have only had specific per-command settings.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22679/5300ee266534f8c68065285de005759c58ac7883..11436d2bf3d1cbd7db23b626778c0aa36567ace6)*

ChayimFriedman2 (changes requested):

crates/rust-analyzer/src/config.rs
emilio: Great, I removed that commit (2727462d58) from the stack. Assuming that you want to keep the configPath plumbing through the runnables, but happy to remove that as well if you want.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22679/2ce932fe628ee5c1d9504fa565e8367603423d9a..daddd891bb9dbb4bea05c19382df19e8b806692b)*

PR #22683, opened 2026-06-30
fix: Support Cargo 1.97.0 lockfile path setting

ChayimFriedman2:
The `resolver.lockfile_path` setting was stabilized, so we can omit the `-Zlockfile-path` argument.

Due to the recent stuff I checked that.

### DIFF
--- a/crates/project-model/src/build_dependencies.rs
+++ b/crates/project-model/src/build_dependencies.rs
@@ -474,8 +474,14 @@ impl WorkspaceBuildScripts {
                                 cmd.arg("--lockfile-path");
                                 cmd.arg(lockfile_copy.path.as_str());
                             }
-                            LockfileUsage::WithEnvVar => {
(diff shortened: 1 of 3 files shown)


lnicola:
Will we need https://github.com/rust-lang/rust-analyzer/pull/21763 again?

ChayimFriedman2:
I just left it without any prerelease qualification. Adding `-Zlockfile-path` only triggers a warning, so we can wait until 1.97.0 becomes stable.

ShoyuVanilla (approved):
`1.97.0` is stable now, so let's merge this 😄 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22683/58e1b7aba7f89031b5c38f8ab1719b74c8f99177..0525a562168019d8d0785f1f479cc976bc1c7e68)*

PR #22684, opened 2026-06-30
internal: Document some little-known features

ChayimFriedman2:
We have a big discoverability problem with our features, but if we don't document them, how can we even expect people to find them?!

### DIFF
--- a/crates/ide/src/rename.rs
+++ b/crates/ide/src/rename.rs
@@ -121,6 +121,17 @@ pub(crate) fn prepare_rename(
 // | VS Code | <kbd>F2</kbd> |
 //
 // ![Rename](https://user-images.githubusercontent.com/48062697/113065582-055aae80-91b1-11eb-8ade-2b58e6d81883.gif)
+//
+// #### Magic Renames
+//
+// rust-analyzer supports some special renames that do additional magic:
+//
+//  - **Anonymous lifetime renames**. You can rename `'_` to any lifetime name (the new name must start with `'`),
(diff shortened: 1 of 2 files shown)


crates/ide/src/rename.rs
```diff
@@ -121,6 +121,17 @@ pub(crate) fn prepare_rename(
 // | VS Code | <kbd>F2</kbd> |
 //
 // ![Rename](https://user-images.githubusercontent.com/48062697/113065582-055aae80-91b1-11eb-8ade-2b58e6d81883.gif)
+//
+// #### Magic Renames
+//
+// rust-analyzer supports some special renames that do additional magic:
+//
+//  - **Anonymous lifetime renames**. You can rename `'_` to any lifetime name (the new name must start with `'`),
```
Wilfred, line 129: Most of these make sense to me -- I only learnt about e.g. `return` go-to-def relatively recently. That said, I wonder if these cases are too detailed? It feels like "we do the right thing when you rename a lifetime" which is great, but maybe doesn't need calling out?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22684/58e1b7aba7f89031b5c38f8ab1719b74c8f99177..a85fdcfd4fec17d54de4ff441668adfacd4a2397)*

ChayimFriedman2: It is "the right thing", which is why it was added, but it definitely isn't *obvious*. It is more than a pure rename.

PR #22686, opened 2026-06-30
internal: Remove postcard default features

workingjubilee:
Everything seems to work fine with this change.

This addresses rust-lang/rust-analyzer#22622 but seems worth it even without since it removes a bunch of (individually small) leaf crates and compiling less never hurts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -66,15 +66,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
-name = "atomic-polyfill"
-version = "1.0.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (approved):

ChayimFriedman2:
The thing I'm worried from the most is that it'll change the serialization format as that'll break communication with old proc macro servers, but the docs don't say anything so let's take the risk 😅 

Thanks!

workingjubilee:
Hm. I have faith that @jamesmunns would make any hypothetical features-that-change-postcard's-serialization-format brightly labeled and non-default. It seems the defaults just are for some kinda "heapless-cas" feature which I assume benefits no-std usage more than what r-a has going on

jamesmunns:
There are not and will not be any changes to postcards "on the wire" format based on features (or at all, as far as I plan). The format [is specified](https://postcard.jamesmunns.com/), and a wire spec change would be *at least* a major semver change.

edit: I even wrote this out specifically in the spec docs: https://postcard.jamesmunns.com/wire-format#stability

PR #22689, opened 2026-07-01
internal: inline `ExpandDatabase::expand_proc_macro` into its only caller

ada4a:

The query was only called by one function (`macro_expand`), and the function backing the query (`expand_proc_macro`) is located in the same module as `macro_expand`, so I think it's fine to call `expand_proc_macro` directly, instead of going through a query.

cc @Veykril


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22689/2ce932fe628ee5c1d9504fa565e8367603423d9a..8e0f324e6d658f83f2d37c71306556433b08d576)*

PR #22690, opened 2026-07-01
fix UB in smol_str borsh_non_utf8 test cases

Manishearth:
This was discovered by an AI-driven unsafe code analysis.

This is test code, so it's not too important to fix, but it impacts the miri-ability of this crate. I _think_ all the same behaviors are being tested, the only thing this doesn't test is the SmolStr serialization path, which shouldn't be being tested with UB strings anyway.

I think just testing the deserialization path is the way to go here, and the test is better as a result.

### DIFF
--- a/lib/smol_str/tests/test.rs
+++ b/lib/smol_str/tests/test.rs
@@ -443,11 +443,10 @@ mod borsh_tests {
     }
     #[test]
     fn borsh_non_utf8_stack() {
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22690/cc5f0f8dadfc3297aebe4026d99b6d8f85fb1eb8..c16a72abce3bfd0a6c7b8c48f80c0dec345425df)*

PR #22698, opened 2026-07-01
Fix loop label rename

mansiverma897993:
This PR restricts the search scope of Definition::Label to its containing body and adds regression tests for renaming labels from both definition and reference sites.


ChayimFriedman2:
The PR title is wrong, it does not fix anything. The test is unnecessary, we do not add tests just because. The only thing that makes sense is the search scope changes, but it does not belong in this PR.

PR #22699, opened 2026-07-01
internal: remove `ExpandDatabase::proc_macro_span` in favor of the underlying fn

ada4a:

The query was only called by two functions, and they are located in the same module as the function backing the query (`proc_macro_span`), so I think it's fine to call the latter directly, instead of going through a query.

The transformation is made somewhat ugly by the query having been defined as `#[salsa::invoke_interned]`, ~~but this seems to be the way to do things: https://github.com/rust-lang/rust-analyzer/pull/22063/files#diff-5828430af308cfce455219d40dac23f272baf89378ada6febee5d6a186ffe0fbL255-R268~~ (EDIT: I was made aware of a nicer solution: https://github.com/rust-lang/rust-analyzer/pull/22699#discussion_r3509149402).

cc @Veykril

### DIFF
--- a/crates/hir-expand/src/db.rs
+++ b/crates/hir-expand/src/db.rs
@@ -600,8 +600,8 @@ fn macro_expand<'db>(
 
 fn proc_macro_span(db: &dyn ExpandDatabase, ast: AstId<ast::Fn>) -> Span {
     let root = db.parse_or_expand(ast.file_id);
-    let ast_id_map = &db.ast_id_map(ast.file_id);
-    let span_map = &db.span_map(ast.file_id);
+    let ast_id_map = db.ast_id_map(ast.file_id);
+    let span_map = db.span_map(ast.file_id);
 
     let node = ast_id_map.get(ast.value).to_node(&root);
     let range = ast::HasName::name(&node)
--- a/crates/hir-expand/src/db.rs
+++ b/crates/hir-expand/src/db.rs
@@ -107,14 +107,6 @@ pub trait ExpandDatabase: SourceDatabase {
         id: AstId<ast::Macro>,
     ) -> &DeclarativeMacroExpander;
 
-    /// Retrieves the span to be used for a proc-macro expansions spans.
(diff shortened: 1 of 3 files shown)


crates/hir-expand/src/db.rs
```diff
@@ -598,15 +590,29 @@ fn macro_expand<'db>(
...
-    let span_map = &db.span_map(ast.file_id);
-
-    let node = ast_id_map.get(ast.value).to_node(&root);
-    let range = ast::HasName::name(&node)
-        .map_or_else(|| node.syntax().text_range(), |name| name.syntax().text_range());
-    span_map.span_for_range(range)
+    #[salsa::interned]
+    struct InternedAstIdFn {
+        id: AstId<ast::Fn>,
+    }
```
ada4a: My initial instinct was to do
```rs
struct InternedAstId<N> {
    id: AstId<N>,
}
```
instead, but salsa didn't seem to like that

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22699/28ccb8be3ed3cb6281fca31e9faf46185d9b6117..22bf4b3e12f7d63a4c65113ffb5057a908602eeb)*

ChayimFriedman2: A better way is to accept a dummy `()` parameter, then Salsa will intern the parameters automatically.

ada4a: That's... cursed. It does look nicer though -- switched to it

ChayimFriedman2: You didn't seem to switch to it?

ada4a: Sorry, forgot to push^^

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22699/28ccb8be3ed3cb6281fca31e9faf46185d9b6117..580545161401c437e8d790303abf970fd51438d6)*

PR #22702, opened 2026-07-01
internal: remove `#[salsa::invoke_interned]` support from `#[query_group]`

ada4a:

Turns out the query removed in https://github.com/rust-lang/rust-analyzer/pull/22699 was the last one of this kind.

RE: tests: I replaced some of the occurrences of `#[salsa::invoke_interned]` with `#[salsa::transparent]`. I think this technically means those functions are no longer queries, and so I would need to also rewrite their underlying functions to be compatible with `#[salsa::tracked]`, as was done in https://github.com/rust-lang/rust-analyzer/pull/22699, but I could not be bothered...

cc @Veykril

### DIFF
--- a/crates/query-group-macro/tests/arity.rs
+++ b/crates/query-group-macro/tests/arity.rs
@@ -2,16 +2,16 @@ use query_group_macro::query_group;
 
 #[query_group]
 pub trait ArityDb: salsa::Database {
-    #[salsa::invoke_interned(one)]
+    #[salsa::transparent]
     fn one(&self, a: ()) -> String;
 
-    #[salsa::invoke_interned(two)]
+    #[salsa::transparent]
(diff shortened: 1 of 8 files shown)


crates/query-group-macro/tests/arity.rs
ChayimFriedman2, line 1: You should just delete this file, it does not test the `query_group` macro anymore.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22702/8e3f3c7539bc3f8fec0068b88000643cb5c8141f..42199a5f7dcfe8622fa2cb20e6ad245a3768080b)*

ChayimFriedman2 (approved):

ada4a:
Wait, could you cancel the merge then, so that I can delete the file in this PR?

PR #22705, opened 2026-07-02
fix: Crash on extract_variable when selecting unresolved macro call

Wilfred:
Be defensive against the range form the token tree not being in the token tree node being edited.

Includes a unit test that previously panicked with:

    Bad range: node range 11..12, range 9..12

AI disclosure: Repro constructed with Fable 5, implementation with GPT 5.5.

### DIFF
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -110,6 +110,9 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
             .skip_while(|it| !it.text_range().contains_range(range))
             .find_map(valid_target_expr(ctx))?;
         let original_range = ctx.sema.original_range(expr.syntax());
+        if !node.text_range().contains_range(original_range.range) {
+            return None;
+        }
         (cover_edit_range(&node, original_range.range), expr)
     } else {
         let expr = node
@@ -3025,5 +3028,10 @@ fn main() {
 "#,
             "Extract into variable",
         );
+    }
+
+    #[test]
+    fn repro_bad_range_unresolved_macro_paren() {
+        check_assist_not_applicable(extract_variable, "fn f() { m!$0($0 }");
     }
 }


Wilfred:
FWIW ~70% of LSP request panics that I see (from the users I support) is "Bad range: node range" of some form. I'm doubtful this fix covers all of them, but it's a good start.

crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -110,6 +110,9 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
             .skip_while(|it| !it.text_range().contains_range(range))
             .find_map(valid_target_expr(ctx))?;
         let original_range = ctx.sema.original_range(expr.syntax());
+        if !node.text_range().contains_range(original_range.range) {
```
ChayimFriedman2, line 113: This will prevent extraction if a macro maps the selected range into something outside its containing node. It's possible to support this by replacing `node` with an ancestor of `original_range`, but this case is weird enough I don't think we ought to.

FYI you can debug such issues more easily (or make it easier for us to debug them) if you ship a r-a build with debug info (`--profile=dev-rel`). If you control r-a installations company-wide this might be easier for you than it is for us.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22705/490c8f22a4095f8db59a020a57129ea163f8de2e..a507205f7ae69832fdf8649ce6ad4624cafabb85)*

ChayimFriedman2 (approved):

crates/ide-assists/src/handlers/extract_variable.rs
Wilfred: Yep, I've been setting `debug = 1` in Cargo.toml for a while now, and we still carry #16231 as a patch although I haven't looked whether it's still benefitting us.

It's still really hard to investigate an LSP error though. There's so much state: the file on disk, the IDE buffer, where the cursor is, any selected text, whether rust-analyzer is fully ready, which files/projects are open, and whether they have changed any VS Code settings. Even asking users nicely is rarely sufficient for getting a repro that I can work with.

SCIP errors are much nicer in this respect: it's only the state on the disk, so I can usually find a repro to start minimising.

If you have any thoughts about how to make debugging easier for me or you I'm super interested :)

ChayimFriedman2: This is of course true, debugging stateful systems is always harder. Sometimes the backtrace is enough for someone who knows the codebase, but many times a repro is needed.

PR #22707, opened 2026-07-02
fix: Crash when computing diagnostics with MIR and error types

Wilfred:
Previously MIR would panic if we tried to access out-of-bounds tuple elements or fields in a value with an error type. This would produce an error of the form:

     field FieldIndex(1) out of range: (i32,)
     can't project out of {type error}
     field FieldIndex(99) out of range: (&'{region error} str, u32)

Instead, handle error types gracefully, and add unit tests.

AI disclosure: Some Fable 5 and GPT 5.5 usage, any errors are of course mine :)

### DIFF
--- a/crates/hir-ty/src/mir/lower/pattern_matching.rs
+++ b/crates/hir-ty/src/mir/lower/pattern_matching.rs
@@ -137,7 +137,8 @@ impl<'db> MirLowerCtx<'_, 'db> {
             }
             Pat::Wild => (current, current_else),
             Pat::Tuple { args, ellipsis } => {
-                let subst = match self.infer.pat_ty(pattern).kind() {
+                let place_ty = cond_place.ty(&self.result, &self.infcx, self.env).ty;
+                let subst = match place_ty.kind() {
                     TyKind::Tuple(s) => s,
                     _ => {
                         return Err(MirLowerError::TypeError(
--- a/crates/hir-ty/src/mir/lower.rs
+++ b/crates/hir-ty/src/mir/lower.rs
@@ -1404,11 +1404,14 @@ impl<'a, 'db> MirLowerCtx<'a, 'db> {
         expr_id: ExprId,
     ) -> Result<'db, ()> {
         if let Expr::Field { expr, name } = &self.store[expr_id] {
-            if let TyKind::Tuple(..) = self.expr_ty_after_adjustments(*expr).kind() {
+            if let TyKind::Tuple(tys) = self.expr_ty_after_adjustments(*expr).kind() {
                 let index =
                     name.as_tuple_index().ok_or(MirLowerError::TypeError("named field on tuple"))?
                         as u32;
-                *place = place.project(ProjectionElem::Field(FieldIndex(index)))
+                if tys.get(index as usize).is_none() {
+                    return Err(MirLowerError::TypeError("tuple field index out of range"));
+                }
+                *place = place.project(ProjectionElem::Field(FieldIndex(index)));
             } else {
                 let field = self
                     .infer
--- a/crates/hir-ty/src/mir/lower/tests.rs
+++ b/crates/hir-ty/src/mir/lower/tests.rs
@@ -121,3 +121,16 @@ fn tuple_field() {
     "#,
     );
 }
+
+#[test]
+fn borrowck_alias_projection_recovery_does_not_panic() {
+    check_borrowck(
+        r#"
+//- minicore: sized
+trait Tr { type A; }
+fn alias<T: Tr>(x: T::A) {
+    let (a, b) = x;
+}
+    "#,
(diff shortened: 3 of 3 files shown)


Wilfred:
This has been the second most common LSP request panic I've seen recently (after #22705).

ChayimFriedman2 (changes requested):
I prefer this to be done where we build the projection and can also return a lowering error, i.e. in `push_field_projection()`.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22707/490c8f22a4095f8db59a020a57129ea163f8de2e..34d8146fb781862e08926d99b89392597374584f)*

Wilfred:
OK, I've pushed the validation into MIR lowering, does this make sense? I've also split the PR into two commits, this is arguably two distinct fixes.

crates/hir-ty/src/mir/lower/pattern_matching.rs
```diff
@@ -137,7 +140,14 @@ impl<'db> MirLowerCtx<'_, 'db> {
             }
             Pat::Wild => (current, current_else),
             Pat::Tuple { args, ellipsis } => {
-                let subst = match self.infer.pat_ty(pattern).kind() {
+                let place_ty = cond_place.ty(&self.result, &self.infcx, self.env).ty;
+                let mut ocx = ObligationCtxt::new(&self.infcx);
+                let normalized_ty = ocx
+                    .structurally_normalize_ty(&ObligationCause::dummy(), self.env, place_ty)
```
ChayimFriedman2: I don't see why we should normalize here. Inference already deeply normalizes at the end, and we already fail lowering if the type isn't a tuple. Do you have a test case where we fail only without normalization?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22707/81d8b1d74a368bb32d75dd696c9a2783df48b6ae..319ddaadc39019f2f36515d5d732e27ae3270ff3)*

crates/ide-diagnostics/src/tests.rs
```diff
@@ -281,6 +281,48 @@ pub(crate) fn check_diagnostics_with_config(
     }
 }
 
+#[test]
+fn mir_tuple_field_projection_recovery_does_not_panic() {
```
ChayimFriedman2: We shouldn't add more tests in this file where possible. Add them to the borrowck diagnostics instead.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22707/81d8b1d74a368bb32d75dd696c9a2783df48b6ae..319ddaadc39019f2f36515d5d732e27ae3270ff3)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/mir/lower/pattern_matching.rs
Wilfred: Yep, the test on this commit fails prior to this change:

```
thread 'tests::mir_alias_projection_recovery_does_not_panic' (58661203) panicked at crates/hir-ty/src/mir.rs:1277:22:
can't project out of Alias(AliasTy { args: [#0], kind: Projection { def_id: TypeAliasId("A") }, .. })
```

crates/ide-diagnostics/src/tests.rs
Wilfred: Moved to borrowck tests.

crates/hir-ty/src/mir/lower/pattern_matching.rs
ChayimFriedman2: This is a projection on an alias, something we cannot normalize before monomorphization. It's not normalization that solves this, and it's not the lack of it that causes the problem. The root of the problem is mismatch between the place type and the pattern type.

The code has a type error; we should not try to lower MIR for such code, but unfortunately that's something we do today. Still, no normalization is needed: just fetching the place type is enough.

Wilfred: OK, so the problem AIUI is just that we're projecting based on the type inferred from the pattern, whereas we should actually be projecting based on type from the place (the RHS in `let (x, y) = not_a_tuple`) and emitting a lowering error if that type isn't a tuple.

I've updated the code and rewritten the commit message based on my understanding here. Let me know if it makes sense.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22707/ed623f271b2ed5a7bcf8375350885ab652c87eea..94332509ea1cf23cf437adc47c5bc57b6d784543)*

PR #22709, opened 2026-07-02
perf: Reduce parser joint-token allocation

mingley:
## Summary
- Reserve parser joint-token storage by bitset word count, not byte count.
- This avoids over-allocating the `joint` side table when parser input is created.

## Measurement
- 1,000,000-token capacity: `125,000` -> `15,625` `u64` words.
- Allocation reduction: `875,000` bytes.

## Testing Done
- [x] `cargo test -p parser --quiet` (`319 passed, 0 failed`)

### DIFF
--- a/crates/parser/src/input.rs
+++ b/crates/parser/src/input.rs
@@ -27,7 +27,7 @@ impl Input {
     pub fn with_capacity(capacity: usize) -> Self {
         Self {
             kind: Vec::with_capacity(capacity),
-            joint: Vec::with_capacity(capacity / size_of::<bits>()),
+            joint: Vec::with_capacity(capacity.div_ceil(bits::BITS as usize)),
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22709/490c8f22a4095f8db59a020a57129ea163f8de2e..a2843dd4a0e25b75cb11eeff3c7923f889bc7e8b)*

ChayimFriedman2:
Do note that this doesn't really matter, as this is a very short-lived allocation, but it's still nice :)

mingley:
> Do note that this doesn't really matter, as this is a very short-lived allocation, but it's still nice :)

Agree.

I still feel like many good things are built upon a big collection of small things done well.

PR #22710, opened 2026-07-02
perf: Avoid allocating tool env var names

mingley:
## Summary
- Use static env var names for Rust tool lookup.
- Avoids allocating a new uppercase `String` for each lookup.
- Adds coverage for the tool-to-env-var mapping.

## Measurement
- Isolated env-key generation median: `15.63ns` -> `0.46ns` per lookup.

## Testing Done
- [x] `cargo test -p toolchain --lib --quiet` (`1 passed, 0 failed`)

### DIFF
--- a/crates/toolchain/src/lib.rs
+++ b/crates/toolchain/src/lib.rs
@@ -20,7 +20,7 @@ pub enum Tool {
 
 impl Tool {
     pub fn proxy(self) -> Option<Utf8PathBuf> {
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
This code is not perf-sensitive at all, so complicating it even just a bit to improve its performance is not worth it.

Thank you anyway for your interest in contributing!

PR #22712, opened 2026-07-02
test: Cover path fragment generic transcription

mingley:
## Summary
- Add coverage for `$path:path` transcription into expression context.
- Verifies that `::` is inserted before generic arguments when needed.

## Testing Done
- [x] `cargo test -p mbe path_fragment_followed_by_generic_args --quiet` (`1 passed, 0 failed`)

### DIFF
--- a/crates/mbe/src/tests.rs
+++ b/crates/mbe/src/tests.rs
@@ -213,6 +213,45 @@ fn main() {
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
We don't add tests just because, if there are no issues fixed.

mingley:
> We don't add tests just because, if there are no issues fixed.

no issues fixed yet ;)

PR #22717, opened 2026-07-03
internal: simplify `fn_macro::register_builtin!`

ada4a:
Instead of constructing both kinds of expanders in the same macro call, construct them in two. This allows for some deduplication in the macro definition.

AFAICT having everything in the same macro was necessary in order to implement `find_by_name`, but I've changed it to be a method on each of the expanders, with `find_builtin_macro` calling both and combining their results.

Also the macro now accepts an optional trailing comma.

All of this makes the macro more similar to the ones in `attr_macro` and `derive_macro`.

### DIFF
--- a/crates/hir-expand/src/builtin/fn_macro.rs
+++ b/crates/hir-expand/src/builtin/fn_macro.rs
@@ -25,41 +25,27 @@ use crate::{
 };
 
 macro_rules! register_builtin {
-    ( $LAZY:ident: $(($name:ident, $kind: ident) => $expand:ident),* , $EAGER:ident: $(($e_name:ident, $e_kind: ident) => $e_expand:ident),*  ) => {
+    ( $EXPANDER:ident: $(($name:ident, $kind: ident) => $expand:ident),* $(,)? ) => {
         #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-        pub enum $LAZY {
+        pub enum $EXPANDER {
             $($kind),*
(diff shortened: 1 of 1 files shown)


crates/hir-expand/src/builtin/fn_macro.rs
```diff
@@ -27,41 +27,27 @@ use crate::{
...
             }
-        }
 
-        impl EagerExpander {
-            fn expander(&self) -> fn (&dyn ExpandDatabase, MacroCallId, &tt::TopSubtree, Span) -> ExpandResult<tt::TopSubtree>  {
-                match *self {
-                    $( EagerExpander::$e_kind => $e_expand, )*
+            fn find_by_name(ident: &name::Name) -> Option<Self> {
+                match ident {
+                    $( id if id == &sym::$name => Some(Self::$kind), )*
```
ChayimFriedman2:
```suggestion
                    $( id if *id == sym::$name => Some(Self::$kind), )*
```
Comparing with references is almost never a good idea. It optimizes worse and can sometimes cause lifetime errors.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22717/693e516795941b029ecd958dd426f94a251c5156..a1bd2c36c8198ff48c406d1718af2630cd70bbb0)*

ChayimFriedman2 (approved):

Veykril (approved):

PR #22719, opened 2026-07-04
internal: remove some more `ExpandDatabase` queries

ada4a:

cc @Veykril


crates/hir-expand/src/span_map.rs
```diff
@@ -54,7 +54,7 @@ pub(crate) fn real_span_map(
 ) -> RealSpanMap {
     use syntax::ast::HasModuleItem;
     let mut pairs = vec![(syntax::TextSize::new(0), span::ROOT_ERASED_FILE_AST_ID)];
-    let ast_id_map = db.ast_id_map(editioned_file_id.into());
+    let ast_id_map = HirFileId::from(editioned_file_id).ast_id_map(db);
```
ada4a, line 57: This `from` is slightly unfortunate -- I could implement `ast_id_map` for `EditionedFileId` as well to avoid it. Though both users of `HirFileId::ast_id_map` have `HirFileId` in scope anyway, so this doesn't require an extra import or anything.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22719/693e516795941b029ecd958dd426f94a251c5156..307571a726c5d2af0647abfd5ee6a59da5244ec5)*

Veykril (approved):

crates/hir-expand/src/span_map.rs
Veykril: That's fine its a single use only after all

crates/hir-expand/src/lib.rs
```diff
@@ -537,6 +485,357 @@ impl MacroCallId {
...
+    }
+
+    pub fn parse_macro_expansion_error(
+        self,
+        db: &dyn ExpandDatabase,
+    ) -> Option<ExpandResult<Arc<[SyntaxError]>>> {
+        let e: ExpandResult<Arc<[SyntaxError]>> =
+            self.parse_macro_expansion(db).as_ref().map(|it| Arc::from(it.0.errors()));
+        if e.value.is_empty() && e.err.is_none() { None } else { Some(e) }
+    }
```
ada4a, line 516: From what I've seen, when `do_something` is a query, then `do_something_error` is usually a query as well (and it's usually referred to as a firewall query, which I assume means "a 'don't break incrementality' query"). Does that logic apply here as well?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22719/693e516795941b029ecd958dd426f94a251c5156..307571a726c5d2af0647abfd5ee6a59da5244ec5)*

Veykril: No, the sole caller of this is in `hir` or so iirc, we don't need  the incrementality at that layer anymore. It's only useful if called within another query.

PR #22722, opened 2026-07-04
fix: fix some assist indent

A4-Tacks:
For exprs like `non_newline.call(newline)` `non_newline(newline)` etc, is it worth adding some `if` in rust-lang/rust-analyzer#21719 to exclude newline?

# Example
**Before this PR**:
```
fn main() {
    {
        {
            loop {
                if let Some(_) = foo(
                            "xxx"
                        )
            {
                    bar(
                        "xxx",
                    );
                } else {
                    break;
                }
            }
        }
    }
}
```
**After this PR**:
```
fn main() {
    {
        {
            loop {
                if let Some(_) = foo(
                    "xxx"
                )
                {
                    bar(
                        "xxx",
                    );
                } else {
                    break;
                }
            }
        }
    }
}
```

### DIFF
--- a/crates/ide-assists/src/handlers/convert_bool_then.rs
+++ b/crates/ide-assists/src/handlers/convert_bool_then.rs
@@ -188,7 +188,7 @@ pub(crate) fn convert_bool_then_to_if(
             let mapless_make = SyntaxFactory::without_mappings();
             let closure_body = match closure_body.reset_indent() {
                 ast::Expr::BlockExpr(block) => block,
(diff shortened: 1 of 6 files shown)


ChayimFriedman2:
@A4-Tacks, I already said many times I think we should **not** explicitly fixing whitespaces in assists. You seem to disagree. Therefore, I would like to set a policy wrt. this:

@rust-lang/rust-analyzer, do you think we should accept code changes that fix whitespaces in assists?

PR #22726, opened 2026-07-05
suggest code action fixes produced from diagnostics under cursor, even if they have effects elsewhere

bend-n:
the point of this, at least for me, is mostly the 
"feature(x) is unstable" errors.
this allows you to code-action fix them from the point of feature usage.

### DIFF
--- a/crates/rust-analyzer/src/diagnostics.rs
+++ b/crates/rust-analyzer/src/diagnostics.rs
@@ -60,7 +60,7 @@ pub(crate) struct DiagnosticCollection {
 #[derive(Debug, Clone)]
 pub(crate) struct Fix {
     // Fixes may be triggerable from multiple ranges.
-    pub(crate) ranges: SmallVec<[lsp_types::Range; 1]>,
+    pub(crate) ranges: SmallVec<[lsp_types::Range; 2]>,
     pub(crate) action: lsp_ext::CodeAction,
 }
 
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
I cannot get rustc to emit fixes for unstable features used; did you verify this patch works?

bend-n:
yeah

ChayimFriedman2:
What was the code?

bend-n:
image

```rs
const fn x() -> &'static [u8] {
    unsafe { core::intrinsics::transmute_unchecked::<u32, u8>(4) };
    todo!()
}
```

Veykril:
You only get those on nightly toolchains I think

ChayimFriedman2 (changes requested):
Alright. I accept the idea, but please put the new range inside `Fix::ranges` instead (and increase it's `SmallVec` size to 2).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22726/28eedd70c04d6036b811e11aaf0629d631cec794..63eb51de8227a29684a23582c205ffd9f91adefc)*

bend-n:
should i just like. `fix.ranges[1] = diagnostic.range`?

ChayimFriedman2:
No, of course not. You need to `push()` it.

bend-n:
yeah but like. would there never be two things in the fix.ranges? was it basically just an option<range> before?

ChayimFriedman2:
No, that's why it's a vector.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22726/d30a8c780af1fdae94c707f46567017899e3bb82..bbdca9aec1f0708033a12994e4115b878db55262)*

PR #22734, opened 2026-07-07
feat: add fixes for array length for type_mismatch

A4-Tacks:
Example
---
```rust
const VALS: [i32; 2$0] = [1, 2, 3];
```

**Before this PR**

No available fixes

**After this PR**

```rust
const VALS: [i32; 3] = [1, 2, 3];
```

### DIFF
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -429,9 +429,10 @@ fn array_length(
     };
     let ast::Type::ArrayType(ty) = ty else { return None };
     let len = ty.const_arg()?.expr()?;
-    let hir::FileRange { file_id, range } = ctx.sema.original_range_opt(len.syntax())?;
+    let hir::FileRange { range: len_range, .. } = ctx.sema.original_range_opt(len.syntax())?;
+    let hir::FileRange { range, file_id } = ctx.sema.original_range_opt(&container)?;
 
-    let edit = TextEdit::replace(range, actual.to_string());
+    let edit = TextEdit::replace(len_range, actual.to_string());
     let source_change = SourceChange::from_text_edit(file_id.file_id(ctx.db()), edit);
     let label = &format!("Change array length to {actual}");
     acc.push(fix("array_length", label, source_change, range));
@@ -1539,6 +1540,16 @@ static VALS: [i32; 2$0] = {[1, 2, 3]};
             "#,
             r#"
 static VALS: [i32; 3] = {[1, 2, 3]};
+            "#,
+        );
+
+        // Convenient trigger range
+        check_fix(
+            r#"
+static VALS: [i32; 2] = [$01, 2, 3];
+            "#,
+            r#"
+static VALS: [i32; 3] = [1, 2, 3];
             "#,
         );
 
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -402,6 +403,58 @@ fn add_await(
...
+fn array_length(
+    ctx: &DiagnosticsContext<'_, '_>,
+    d: &hir::TypeMismatch<'_>,
+    expr_ptr: &InFile<AstPtr<ast::Expr>>,
+    acc: &mut Vec<Assist>,
+) -> Option<()> {
+    let (ty1, expected) = d.expected.as_array(ctx.db())?;
+    let (ty2, actual) = d.actual.as_array(ctx.db())?;
+
+    if ty1 != ty2 || expected == actual {
```
ChayimFriedman2: Maybe check for unification instead of type equality?

ChayimFriedman2 (approved):
LGTM.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22734/89716ef32899552fed84d909415e4ef5e3e38753..c997654a194487aebf92efc83b1308f990a7e36e)*

crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -402,6 +403,58 @@ fn add_await(
...
+        }
+    };
+    let ast::Type::ArrayType(ty) = ty else { return None };
+    let len = ty.const_arg()?.expr()?;
+    let hir::FileRange { file_id, range } = ctx.sema.original_range_opt(len.syntax())?;
+
+    let edit = TextEdit::replace(range, actual.to_string());
+    let source_change = SourceChange::from_text_edit(file_id.file_id(ctx.db()), edit);
+    let label = &format!("Change array length to {actual}");
+    acc.push(fix("array_length", label, source_change, range));
```
A4-Tacks, line 438: Maybe we should use the `container`'s range to trigger quick fix more effectively?


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22734/0e3f6f8728865fa87a8df81e9003e33d1557cae3..7e8e0086ab488eb0aaae3fceb4389ccaccc51e08)*

ChayimFriedman2: I think we should use the range of the array type. The range of the length is too small, but the range of the container is too large.

A4-Tacks: But the editing point is outside the range of the array, I don't know if this is appropriate

ChayimFriedman2: No? We're editing the length of the array which is inside the array.

A4-Tacks: Oh, you said array type, I thought it was an array expression

But I think the array type range is still too small

ChayimFriedman2: I think it's not, but I'll let you decide.

PR #22737, opened 2026-07-08
feat: cross-parent module move in willRenameFiles

### LINKED ISSUE #8872: Request: move modules in project structure refactoring
Will be good to see a following feature:
When I'm moving a folder with one module to another place (for example to a directory of a higher level - like from src/ui/views/ to src/ui/) using drag-and-drop in VS Code file explorer it should automatically: delete this mod from mods declared in old place, add this mod to a mod where selected mod was dropped, update imports - both in moved mod (for ex. when `super::*` imports become invalid because mod parent changed) and in all other files where this mod was used.
This feature for me seems similar to currently presented ability to rename modules - which automatically renames all imports of renamed mod and this mod declaration in structure.

enomado:
## Summary

Implements the request in rust-lang/rust-analyzer#8872: moving a module to a **new parent** — dragging a
`.rs` file or a module folder into a *different* directory — now fixes the module
tree and rewrites references, instead of leaving the code broken.

`workspace/willRenameFiles` previously handled only same-directory renames (rust-lang/rust-analyzer#7009).
This adds the cross-parent case.

## What it does

A new `ide` entry point `Analysis::will_move_module`, backed by
`ide_db::rename::move_mod`, produces a single `SourceChange`:

- **mod-tree surgery** — drop the old `mod X;` (with its trailing newline) and
  re-declare it, preserving visibility, at the end of the new parent's item list;
- **reference rewrite** — for each usage, splice the leading `ast::Path` ending at
  the module segment with the module's new absolute `crate::…` path, leaving any
  trailing `::Item` untouched (HIR still describes the pre-move tree, so the
  old/new absolute paths are computed textually);
- **`super::` re-anchoring inside the moved file** — a leading `super` run is
  rewritten to the `crate::…` path it denotes today, since after the move the same
  text would resolve against a different parent (`self::` is left alone);
- **grouped `use` imports** — a member importing through the moved module is pulled
  out into its own `use`; an emptied group is replaced wholesale, otherwise moved
  members are deleted as overlap-free runs;
- **cross-crate references** keep the extern crate name (`foo::a::m` → `foo::b::m`),
  honouring `package = "…"` renames;
- **`mod.rs` / directory modules** — the same surgery generalizes; the editor moves
  the directory physically, so only text edits are emitted.

`handle_will_rename_files` gained a cross-parent arm that maps the destination
directory to the new parent's crate-relative module segments and routes both dropped
files and dragged folders (resolved via their `mod.rs`) to `will_move_module`.
File-system edits stay cleared — the editor owns the physical move.

## Boundaries (documented in code, left to the compiler)

- macro-expanded references (edits would have to land in the macro *input*);
- a `super::super…` chain inside a *child* file of a moved directory module that
  climbs *above* the module (only the moved module's own file is re-anchored);
- `pub(super)` / `pub(in super::x)` visibility.

## Tests

- `will_move_module_*` unit tests in `ide` cover each case exhaustively via
  `expect_test`;
- two `rust-analyzer` slow-tests pin the LSP wiring:
  `test_will_rename_files_cross_parent` (file drag) and
  `test_will_rename_files_cross_parent_folder` (folder drag).

## Second commit: moving a single item (`experimental/moveItemToModule`)

The commit on top adds the sibling operation: relocate one **item** — a function, a
type, a const — into a module that **already exists**. It is a separate story, kept as
its own commit so it can be reviewed on its own; it lives here only because it builds
on the reference-splicing pass introduced above. It can be split into a follow-up PR if
that reads better — it just cannot land first.

Both ends are named by absolute path (`my_crate::a::foo` into `my_crate::b`) rather
than pointed at: there is no editor gesture that picks an item *and* a destination
module the way dragging a file picks a module and its new parent, and a caller running
a batch of moves has paths, not offsets. Hence a request rather than an assist.

Tearing an item away from its neighbours breaks three different things at once, and
each is fixed:

- paths **inside** the item resolved against the source module's scope — they are
  re-resolved against the destination via `PathTransform`, so nothing is imported
  there and nothing can collide with what the destination already has;
- qualified references **to** the item — spliced by the same pass the module move
  uses, `use`-group splitting included;
- bare references that resolved by proximity — every file left holding one gets an
  import, since the item is not a neighbour there anymore.

Two visibility widenings fall out of the same reasoning: an item that kept its callers
by living next to them is widened to `pub(crate)`, and so is a private neighbour the
moved body calls — the destination cannot otherwise name it, and leaving the call as
written would emit code that does not compile.

Kinds whose meaning is not carried by their text are refused with a message saying so:
`macro_rules!` (visible by textual order), macro-generated items, enum variants,
modules (that is the module move), and `pub(super)`/`pub(in …)` visibility. Two
boundaries are deliberate rather than refused: `impl` blocks do not travel with a type
(an inherent impl is legal from anywhere in the crate, and half the impls moved would
be worse than none), and a `use` left unused by the departure is not removed (a
warning, not an error).

Tests: 10 `move_item_*` unit tests in `ide` via `expect_test`, plus a `rust-analyzer`
slow-test `test_move_item_to_module` pinning the LSP wiring.

The branch is rebased on current `master`. Two small fixes for upstream API changes are
folded into the first commit: `ast::NameRef::text()` now yields `&str`, and
`FileRename`'s uris are `Uri` rather than `String`.

## Disclosure

Both commits were written **end-to-end by an AI coding agent** (Anthropic's Claude) —
it is entirely AI-generated ("vibe-coded"). A human directed the work, reviewed the
diff, and ran the tests, but did not hand-write the implementation. Flagging this
explicitly so reviewers can weigh it accordingly; happy to iterate on anything.

Closes rust-lang/rust-analyzer#8872.

### DIFF
--- a/crates/ide-db/src/rename.rs
+++ b/crates/ide-db/src/rename.rs
@@ -28,21 +28,26 @@ use crate::{
 };
 use base_db::AnchoredPathBuf;
 use either::Either;
-use hir::{FieldSource, FileRange, HasCrate, InFile, ModuleSource, Name, Semantics, sym};
+use hir::{
+    FieldSource, FileRange, HasCrate, InFile, ModPath, ModuleSource, Name, PathKind, Semantics,
+    sym,
+};
 use itertools::Itertools;
 use rustc_hash::FxHashSet;
 use span::{Edition, FileId, SyntaxContext};
 use stdx::{TupleExt, never};
 use syntax::{
-    AstNode, SyntaxKind, T, TextRange,
-    ast::{self, HasName},
+    AstNode, SyntaxElement, SyntaxKind, SyntaxNode, T, TextRange, TextSize,
+    ast::{self, HasAttrs, HasName, HasVisibility},
+    syntax_editor::{Position, Removable, SyntaxEditor},
 };
 
 use crate::{
     RootDatabase,
     defs::Definition,
+    helpers::mod_path_to_ast_with_factory,
     search::{FileReference, FileReferenceNode},
-    source_change::{FileSystemEdit, SourceChange},
+    source_change::{FileSystemEdit, SourceChange, SourceChangeBuilder},
     syntax_helpers::node_ext::expr_as_name_ref,
     traits::convert_to_def_in_trait,
 };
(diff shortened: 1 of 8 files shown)


ChayimFriedman2:
> This change was written end-to-end by an AI coding agent (Anthropic's Claude) —
it is entirely AI-generated ("vibe-coded"). A human directed the work, reviewed the
diff, and ran the tests, but did not hand-write the implementation.

A human reviewing the AI-generated code is not enough; said human must have the ability to fully understand the code, to the point that they could have written it themselves and only use the AI as a way to cut development time. Given that this is only your second contribution to rust-analyzer, and the first one was nowhere close to the size of this one, I heavily doubt this is the case.

Also, I suspect you also used AI to author the pull request description, and that is plain forbidden by our [AI policy](https://github.com/rust-lang/rust-analyzer/blob/master/AI_POLICY.md).

enomado:
i need feature - i asked claude to implement this feature. you are free to play your  ai-denial games. i did pr because hundreds of people did the same but as scripts.

ChayimFriedman2:
You are, of course, free to merge and personally maintain any feature or code you want in your rust-analyzer fork, AI-generated or not. What you cannot do is asking us to merge it into the main repo. If you want your code to be merged, you need to obey our rules.

ChayimFriedman2:
With that said: I don't want to review hundreds of unsupervised AI-generated code lines. If you confirm that you think you completely understand this patch, I'll take your word for it and review this. Otherwise, I will close this PR. Note that this PR has *a lot* of comments, which is typical for AI-generated code and usually not a good sign.

That is unrelated to the fact you used AI to generate the PR description. Given that you probably didn't know that is forbidden that is acceptable this time, but if you'll continue doing that we'll reach out to moderation.

enomado:
Well. Fair enough. I'm not sure this is solid right now. I need some time

ChayimFriedman2:
Converted to draft, please convert this back when it's ready.

enomado:
I've been using this for a few months now. Seems to work.

And about the PR being so big — I could actually have made five small PRs, starting with the plain rename, then the path fixing, but I just decided to do it thoroughly in one go. There isn't really that much code there, half of it is tests. It barely touches the architecture.

ChayimFriedman2:
To be clear: neither big PRs nor lack of testing (if not too severe) are blocking. We're merging large PRs and PRs we did not even manually test all the time.

What matters is the implementation. If the implementation is AI-generated, you *must* be able to understand it to the point you could've write it yourself. If you don't, it cannot be merged, no matter how tested it is.

enomado:
Sorry, what exactly does "buying into" the code entail?

I’ve been programming for 15 years. Of course, if I were already in this domain and spent three months on it, I could write this myself.

The issue is something else entirely: I’m adding what is essentially a core feature—one I’ve missed throughout my entire career and that would have personally saved me months of time spent on manual renaming. Yet, for some reason, ai-deniers treat this like a game or a piece of postmodern performance art—as if all the world's problems were solved and we were just role-playing a retro-computing scenario.

Just calculate the energy wasted on manual renaming versus doing it algorithmically.

I also don't want to win this argument through sheer stubbornness. I respect your work and understand that many projects likely end up without maintainers.

If there’s no one to handle this, and the extra code would make maintenance more burdensome, then fair enough—I’ll probably just keep it in my own fork.

ChayimFriedman2:
> Sorry, what exactly does "buying into" the code entail?

It's as simple as: you could have written the code yourself with your current knowledge of the project and it. Meaning, using AI saved you time (the time to physically write the code) but not understanding. Of course you cannot *know* that for sure, but as a programmer for 15 years you can be expected to know when you know the code.

The key point is:

> if I were already in this domain and spent three months on it, I could write this myself.

And I'm afraid this is not enough. A simple test is: *after* you've looked at the AI generated code, delete it all and re-write it yourself (I don't require you to do that, but metaphorically). If you can do that immediately, great, this is in good shape. If this still requires "entering into the domain and spending three months on it", then we cannot accept this PR.

I see you've chosen to close the PR which is also fine. I still found it important to clarify and not just for you.

PR #22738, opened 2026-07-08
fix: Cancel ongoing build scripts/workspace discovery on exit

### LINKED ISSUE #22732: Build process keeps running in the background after closing vscode
When working in a rustc checkout, some process spawned by RA kept running in the background even after closing vscode, holding the build lock, and preventing me from finishing the `git commit` call. I had to do `killall curl` to fix that (looks like bootstrap was stuck trying to download rustc over a way too slow train internet connection).

Before quitting vscode, I also tried "cancel running flychecks", which did not help either. The stuck command wasn't a flycheck, it was the build script / proc macro building command. But I would expect that, too, to be canceled when quitting vscode.

The status bar looks like this while this happens (i.e. just before I quite vscode):

Image

**rust-analyzer version**: rust-analyzer version: 0.3.2955-standalone [/home/r/.vscode-oss/extensions/rust-lang.rust-analyzer-0.3.2955-linux-x64/server/rust-analyzer]

**rustc version**:  rustc dev tree

**editor or extension**: VSCodium, extension version 0.3.2963

**relevant settings**: a lot :)
```json
{
    "rust-analyzer.linkedProjects": [
        "Cargo.toml",
        "library/Cargo.toml",
        "src/bootstrap/Cargo.toml",
        "compiler/rustc_codegen_cranelift/Cargo.toml",
        "compiler/rustc_codegen_gcc/Cargo.toml",
    ],
    "rust-analyzer.check.invocationStrategy": "once",
    "rust-analyzer.check.overrideCommand": [
        "./ra-check", "--json-output"
    ],
    "rust-analyzer.rustfmt.overrideCommand": [
        "${workspaceFolder}/build/host/rustfmt/bin/rustfmt",
        "--edition=2024",
    ],
    "rust-analyzer.procMacro.server": "${workspaceFolder}/build/host/stage0/libexec/rust-analyzer-proc-macro-srv",
    "rust-analyzer.cargo.buildScripts.invocationStrategy": "once",
    "rust-analyzer.cargo.buildScripts.overrideCommand": [
        "python3", "x.py", "check",
        "library", "compiler", "miri",
        "--json-output",
        "--compile-time-deps",
        "--set", "rust.download-rustc=false",
    ],
    "rust-analyzer.server.extraEnv": {
        "RUSTC": "${workspaceFolder}/build/host/stage0/bin/rustc",
        "CARGO": "${workspaceFolder}/build/host/stage0/bin/cargo",
    },
    "rust-analyzer.rustc.source": "./Cargo.toml",
    "rust-analyzer.cargo.sysrootSrc": "./library",
    "rust-analyzer.cargo.extraEnv": {
        "RUSTC_BOOTSTRAP": "1"
    },
    // Some diagnostics don't work properly here
    "rust-analyzer.diagnostics.enable": false, // due to <https://github.com/rust-lang/rust-analyzer/issues/13504>, we have too many errors
    //
    "files.watcherExclude": {
        "build": true,
    },
    "files.exclude": {
        "build": true
    },
}
```

**repository link (if public, optional)**: https://github.com/rust-lang/rust/

ChayimFriedman2:
Flycheck and proc macro server process handles are stored in the `GlobalState`, using wrappers (`JodChild` or `CommandHandle`) that kill the process on drop. So when we exit and `GlobalState` is dropped, they're automatically cancelled.

However, build scripts and workspace discovery are triggered inside thread pool tasks. When `GlobalState` is dropped, they're left untouched, and when r-a process exits and the handles are destroyed, the processes are detached.

The way I went to solve this is by adding an option to cancel all tasks on a thread pool, by injecting a panic into it. Of course you cannot force a thread to panic (this has the same problems as killing a thread), so instead the thread must check some state periodically. Most of our tasks are have short execution time before executing the next database query, so we can cancel the database, making Salsa inject a panic. Except those that invoke child processes.

So I went and changed `JodChild` to also check periodically for cancellation. This is non-trivial because the standard library has neither "read with timeout" nor "wait with timeout" primitives for pipes, so we're forced into platform-specific code. I also introduced a `JodCommand` wrapper around `Command` to make it harder to use waiting operations by mistake, and added `Command` and `Child` to clippy::disallowed_types.

This was still not enough, because the extension itself has not delivered the shutdown event correctly, and also had a too short timeout for it (200ms). So I changed it to send a shutdown request in `deactivate()`, and increased the timeout to 2s.

I have no idea how to test this automatically, but I tested this manually by setting a build command that waits forever then closing VSCode. Before those changes, the build process is still there; after them, it is killed properly.

Fixes https://github.com/rust-lang/rust-analyzer/issues/22732.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3016,6 +3016,7 @@ name = "toolchain"
 version = "0.0.0"
 dependencies = [
  "camino",
+ "stdx",
 ]
 
 [[package]]
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,9 +1,12 @@
 disallowed-types = [
     { path = "std::collections::HashMap", reason = "use FxHashMap" },
     { path = "std::collections::HashSet", reason = "use FxHashSet" },
-    { path = "std::collections::hash_map::RandomState", reason = "use BuildHasherDefault<FxHasher>"}
+    { path = "std::collections::hash_map::RandomState", reason = "use BuildHasherDefault<FxHasher>" },
+    { path = "std::process::Command", replacement = "stdx::process::JodCommand" },
+    { path = "std::process::Child", replacement = "stdx::process::JodChild" },
 ]
 
 disallowed-methods = [
     { path = "std::process::Command::new", reason = "use `toolchain::command` instead as it forces the choice of a working directory" },
+    { path = "stdx::process::JodCommand::new", reason = "use `toolchain::command` instead as it forces the choice of a working directory" },
 ]
--- a/crates/ide/src/expand_macro.rs
+++ b/crates/ide/src/expand_macro.rs
@@ -255,7 +255,7 @@ fn _format(
     let edition = crate_id.data(db).edition;
 
     #[allow(clippy::disallowed_methods)]
-    let mut cmd = std::process::Command::new(toolchain::Tool::Rustfmt.path());
+    let mut cmd = stdx::process::JodCommand::new(toolchain::Tool::Rustfmt.path());
     cmd.arg("--edition");
     cmd.arg(edition.to_string());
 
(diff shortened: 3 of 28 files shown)


Veykril:
I don't think this is the right way to handle this. All the major platforms should have some ways for us to do this at the operating system. I am taking a look.

PR #22740, opened 2026-07-08
fix: avoid unbounded recursion in the parser

### LINKED ISSUE #9824: Avoid unbounded recursion in the parser
The parser should track the "depth" of the parce, and fail gracefully when depth is larger than the limit (128 or whereabouts feels about right). Might be best implemented via unwinding :0)

See #9358 for a test case and motivation

Albab-Hasan:
deeply nested input overflows the parser stack and takes the whole process down with it (the 100k nested parens reproducer from https://github.com/rust-lang/rust-analyzer/pull/9358 .

track recursion depth on `Parser` and bail out with an `ERROR` node past 128 levels. each guard wraps the deepest function of its grammar family and delegates to an `_inner` fn so no return path can skip `leave_recursion`. counter instead of unwinding since unwinding past an armed `Marker` trips its `DropBomb`. the limit also bounds the syntax tree depth that everything downstream recurses over.

fixes https://github.com/rust-lang/rust-analyzer/issues/9824

### DIFF
--- a/crates/parser/src/grammar/attributes.rs
+++ b/crates/parser/src/grammar/attributes.rs
@@ -60,6 +60,15 @@ fn cfg_attr_meta(p: &mut Parser<'_>, m: Marker) {
 const CFG_PREDICATE_FIRST_SET: TokenSet = TokenSet::new(&[T![true], T![false], T![ident]]);
 
 fn cfg_predicate(p: &mut Parser<'_>) {
+    if p.enter_recursion() {
+        p.bail_recursion();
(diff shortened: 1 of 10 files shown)


ChayimFriedman2:
The real fix is probably expanding the stack. [T-compiler has just accepted a MCP to remove `stacker` in favor of letting the OS grow the stack](https://github.com/rust-lang/compiler-team/issues/1011). We can do the same - that requires almost zero changes to existing code.

PR #22743, opened 2026-07-09
fix: resolve path on all namespace on resolve_path

dfireBird:
Thanks @ChayimFriedman2 for the initial work.

### DIFF
--- a/crates/hir/src/source_analyzer.rs
+++ b/crates/hir/src/source_analyzer.rs
@@ -24,8 +24,9 @@ use hir_def::{
     hir::{BindingId, Expr, ExprId, ExprOrPatId, Pat, PatId, generics::GenericParams},
     lang_item::LangItems,
     nameres::MacroSubNs,
-    resolver::{Resolver, TypeNs, ValueNs, resolver_for_scope},
+    resolver::{ResolveValueResult, Resolver, TypeNs, ValueNs, resolver_for_scope},
     type_ref::{Mutability, TypeRefId},
+    visibility::Visibility,
 };
 use hir_expand::{
     HirFileId, InFile,
@@ -1060,7 +1061,7 @@ impl<'db> SourceAnalyzer<'db> {
         };
 
         let store_owner = self.resolver.expression_store_owner();
-        let res = resolve_hir_value_path(
+        let (res, _) = resolve_hir_value_path(
             db,
             &self.resolver,
             store_owner,
@@ -1643,7 +1644,8 @@ impl<'db> SourceAnalyzer<'db> {
                         Some(name.clone()),
                     )),
                     hygiene,
-                ),
+                )
+                .map(|(it, _)| it),
             )
         })
     }
@@ -1683,7 +1685,8 @@ impl<'db> SourceAnalyzer<'db> {
                         Some(name.clone()),
                     )),
                     hygiene,
-                ),
+                )
+                .map(|(it, _)| it),
             )
         }))
     }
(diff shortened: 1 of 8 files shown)


ChayimFriedman2:
I don't think "half work" is a descriptive commit name 😅 Feel free to squash.

ChayimFriedman2:
Also for Clippy you need to rebase on master.

dfireBird:
> I don't think "half work" is a descriptive commit name 😅 Feel free to squash.

I just pushed didn't think much with that :sweat_smile: 

> Also for Clippy you need to rebase on master.

I rebased a bit early lol.

crates/hir/src/source_analyzer.rs
```diff
@@ -1956,41 +1963,63 @@ fn resolve_hir_path_<'db>(
...
-        if res.any().is_some() {
-            res
-        } else if let Some(type_ns) = items() {
-            PathResolutionPerNs::new(Some(type_ns), None, None)
-        } else {
-            PathResolutionPerNs::new(None, None, macros())
+    let mut types_is_visible = false;
+    let mut values_is_visible = false;
+    let mut macros_is_visible = false;
+    let mut types = types().or_else(items).map(|(res, vis)| {
```
ChayimFriedman2: I know I wrote this, but this can be `inspect()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22743/af1fe01c98cfb006b3e12bcca4941551f1c96273..ed59858aba2b2cc12110b4d19ca8f0d07df64f6d)*

crates/hir/src/source_analyzer.rs
```diff
@@ -1956,41 +1963,63 @@ fn resolve_hir_path_<'db>(
...
-            macro_ns: macros(),
-        }
-    } else {
-        let res = if prefer_value_ns {
-            values()
-                .map(|value_ns| PathResolutionPerNs::new(None, Some(value_ns), None))
-                .unwrap_or_else(|| PathResolutionPerNs::new(types(), None, None))
+    if !resolve_per_ns {
+        if prefer_value_ns {
+            if let Some((res, vis)) = values()
```
ChayimFriedman2: This means we can resolve twice when `!resolve_per_ns`. So I think we should hoist the declaration of `types`, `values` and `macros`, and `X_is_visible`, and set them initially to `None as <Option<Option<_>>>` and `None as Option<bool>`. Then if `!resolve_per_ns`, return with them if they resolve to something visible, or otherwise set the corresponding namespace to `Some(resolution)` and `X_is_visible` to `Some(is_visible)`. Then after the `if`, we do:

```rust
let types = types.unwrap_or_else(|| ...);
...
let types_is_visible = types_is_visible.unwrap_or_else(|| types.inspect(...));
...
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22743/af1fe01c98cfb006b3e12bcca4941551f1c96273..ed59858aba2b2cc12110b4d19ca8f0d07df64f6d)*

ChayimFriedman2 (changes requested):

crates/hir/src/source_analyzer.rs
dfireBird: I did something in this lines. I think it can now resolve at most once but correct me if I'm wrong.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22743/85e85df195c58fa2c887d4fc09efec8250d6fcdb..4bc9b21c70d0d939c8b07e1c2a8814f4a51ae55c)*

PR #22745, opened 2026-07-09
internal: remove `DefDatabase`

ada4a:

The migrations are in the order of increasing controversiality:
- `macro_def`, `assoc_item_visibility`: I could turn them into methods on items defined in `hir_def`, so I did.
- `crate_supports_no_std`: The query has regressed during the attribute rewrite and is now useless (see https://github.com/rust-lang/rust-analyzer/pull/22745#discussion_r3550765343). Removed it completely.
- `file_item_tree`: Can't be a method on `HirFileId` because orphan rule. Had a user in `ide`, so I re-exported it via `hir::db`. See [#t-compiler/rust-analyzer > &#96;DefDatabase::file_item_tree&#96;](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/.60DefDatabase.3A.3Afile_item_tree.60/with/608982206) for discussion.
- `crate_notable_traits`, `include_macro_invoc`: Can't be a method on `Crate` because orphan rule. Had only one user in `hir` though, so I exposed them as free-standing functions.

Let me know if the PR is too big and you'd prefer to have it split up.


crates/hir-def/src/find_path.rs
```diff
@@ -653,6 +654,13 @@ fn find_local_import_locations(
     });
 }
 
+#[salsa::tracked]
+fn crate_supports_no_std(db: &dyn DefDatabase, crate_id: Crate) -> bool {
+    let root_module = crate_def_map(db, crate_id).root_module_id();
+    let attrs = AttrFlags::query(db, AttrDefId::ModuleId(root_module));
+    attrs.contains(AttrFlags::IS_NO_STD)
+}
+
```
Veykril: side note (irrelevant to the pr), this is incorrectly implemented imo. This draws a dependency on the root module attrs, the no std state should be encoded in the crate def map and be looked up from there imo.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22745/e7e17b692a073ca9820d1822626646b9cc045153..f83c49bf2b112996ae88565e2b05cf048f680b20)*

crates/hir-def/src/lib.rs
```diff
@@ -652,11 +676,68 @@ pub enum MacroId {
...
+    pub fn is_attribute(self, db: &dyn SourceDatabase) -> bool {
         matches!(self, MacroId::ProcMacroId(it) if it.lookup(db).kind == ProcMacroKind::Attr)
     }
 }
 
+#[salsa::tracked]
+impl MacroId {
+    /// Turns a MacroId into a MacroDefId, describing the macro's definition post name resolution.
+    #[salsa::tracked]
+    pub fn macro_def(self, db: &dyn SourceDatabase) -> MacroDefId {
```
ada4a: should I rename this to just `def` maybe?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22745/e7e17b692a073ca9820d1822626646b9cc045153..f83c49bf2b112996ae88565e2b05cf048f680b20)*

Veykril (approved):

crates/hir-def/src/lib.rs
Veykril: I;d rather go with `definition` then I think, but yea

crates/hir-def/src/find_path.rs
ada4a: > the no std state should be encoded in the crate def map and be looked up from there imo

As a matter of fact, it already is: https://github.com/rust-lang/rust-analyzer/blob/e7e17b692a073ca9820d1822626646b9cc045153/crates/hir-def/src/nameres.rs#L224

I'll open a follow-up PR to change the query to use this

Veykril: oh actually, we changed behavior here with the attribute rewrite. This query used to poke the attributes for whether they contain cfg_attr gated no_std attributes. We regressed here, this query meant to be a heuristic. That is to sayh, the way this query is written right now is uselss since we can just lookup the crate def map. So lets get rid of this query entirely

ada4a: Should I add a FIXME talking about how we'd like to eventually implement that heuristic again (and at that point wrap that logic into a query)?

Veykril: Hmm, I am actually confused. We have a test `prefer_core_paths_over_std` for the heuristic, so where is that covered?

Veykril: Ah I think the test is written incorrectly so we never caught this

Veykril: Yea I'd say lets remove this query, I'll have a brief look at fixing this up.

PR #22747, opened 2026-07-09
fix: Reimplement `crate_supports_no_std` syntactic heuristic

Veykril:
This was never correctly tested which caused us  to lose the correct implementation.

### DIFF
--- a/crates/hir-def/src/attrs.rs
+++ b/crates/hir-def/src/attrs.rs
@@ -380,6 +380,29 @@ pub fn parse_extra_crate_attrs(db: &dyn SourceDatabase, krate: Crate) -> Option<
     Some(p.tree())
 }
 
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (commented):
Is there a specific reason for not simply querying the def map? The approach in this PR is faster as it avoids computing the def map, but this doesn't matter where it's used.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22747/2b4c2c3bbda55930a97ae6bf1bfd82e4f385da6d..c5ac92036d8ca68938acac6eaf38c0deec679411)*

Veykril:
The def map stores the `no_std`ness for the current cfg set. This function wants to know whether the crate can be no_std under some arbitrary cfg configuration

ChayimFriedman2 (approved):

PR #22753, opened 2026-07-10
fix!: make lsp-server `Response` type closer aligned to JSON-RPC

ribru17:
From [the JSON-RPC 2.0 protocol](https://www.jsonrpc.org/specification#response_object):

> result
>  This member is REQUIRED on success.
>  This member MUST NOT exist if there was an error invoking the method.
>  The value of this member is determined by the method invoked on the Server.
> error
>  This member is REQUIRED on error.
>  This member MUST NOT exist if there was no error triggered during invocation.
>  The value for this member MUST be an Object as defined in section 5.1.
>
> ...
>
> Either the result member or error member MUST be included, but both members MUST NOT be included.

The current typing of the `Response` object allows both the `result` and `error` members to be present, or for neither to be present. This commit strengthens the typing of the response to prevent these cases from being expressed. This is a breaking change.

The response kind definition is borrowed from the
[tower-lsp](https://github.com/ebkalderon/tower-lsp) definition.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1455,7 +1455,7 @@ dependencies = [
 
 [[package]]
 name = "lsp-server"
-version = "0.8.0"
+version = "0.9.0"
 dependencies = [
  "anyhow",
  "crossbeam-channel",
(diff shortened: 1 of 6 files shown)


Veykril (approved):

Veykril:
Thanks!

ChayimFriedman2:
Shouldn't we, then, just use `Result`?

ribru17:
Good point- I'll draft a PR

PR #22755, opened 2026-07-10
internal: match_ast macro supports guard

A4-Tacks:
Example
---
```rust
match_ast! {
    match parent {
        ast::Expr(it) if !it.is_block_like() => replace(it),
        _ => return,
    }
}
```


ChayimFriedman2 (approved):
This is so simple that I'm not opposed even though it's not used currently.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22755/888b2b3409a72e4e1c7251976e52edc3a610a8f4..3027de45f05f5257c949c6164081b845a02b8e14)*

PR #22758, opened 2026-07-10
fix: Recover from MIR field projections on slices

### LINKED ISSUE #22757: MIR borrow checker panics on a field projection from a slice
**rust-analyzer version**: rust-analyzer 1.99.0-nightly (af3d9558 2026-07-09)

**rustc version**: rustc 1.99.0-nightly (af3d95584 2026-07-09)

**editor or extension**: VsCode extension. rust-analyzer  v0.3.2963. Observed through LSP `textDocument/diagnostic` and `textDocument/codeAction` requests. The behavior should not specific to an editor.

**relevant settings**: N/A

**repository link (if public, optional)**: [Overseer](https://github.com/northaxosky/overseer)

```rust
#[derive(Clone, Copy)]
enum Row {
    A { x: usize },
    B { x: usize, y: usize, z: bool, w: usize },
}

struct S {
    rows: Vec<Row>,
}

impl S {
    fn f(&mut self, idx: Option<usize>) -> bool {
        let rows = &self.rows;
        let Some(row) = idx.and_then(|i| rows.get(i)).copied() else {
            return false;
        };

        match row {
            Row::B { y, .. } => {
                let _ = y;
                false
            }
            Row::A { x } => {
                let _ = x;
                false
            }
        }
    }
}
```

The reduced standalone source does not preserve every aspect of the original inference context. The panic itself can be reproduced directly in the `hir-ty` test harness by applying a MIR `Field` projection to a slice-typed `PlaceTy`.

The panic is:

​```text
thread 'Worker3' panicked at crates/hir-ty/src/mir.rs:1277:22:
can't project out of [Row]
​```

`borrowck_query` reaches `PlaceTy::field_ty` with a `ProjectionElem::Field`, no downcast variant, and a slice base type. `field_ty` only accepts non-enum ADTs, closures, and tuples in this case.

A field projection does not contain the element index needed to project through an array or slice, so it cannot derive a meaningful element-field type. This malformed MIR should not crash diagnostic or code-action requests. Propagating an error type allows MIR diagnostic recovery to stop safely.

This is rust-analyzer's MIR-based diagnostic borrow checker, not rustc's borrow checker, so the panic does not prevent `cargo build`.

northaxosky:
Fixes rust-lang/rust-analyzer#22757

MIR diagnostic borrow checking can encounter a `Field` projection whose base place has an array or slice type. `PlaceTy::field_ty` previously treated that as an invariant violation and panicked with:

​```text
can't project out of [T]
​```

A `Field` projection contains a field index but no array or slice element index, so it cannot produce a meaningful element-field type. This change treats array and slice bases as failed MIR type recovery and propagates an error type. Other unexpected base types retain the existing invariant check.

The regression test exercises the production `PlaceTy::projection_ty` path with a slice base and a `Field` projection. Before the fix it panics with `can't project out of [bool]`; afterward it returns the error type.

Tests:

​```text
cargo test -p hir-ty
​```

All 998 `hir-ty` tests pass.

### DIFF
--- a/crates/hir-ty/src/mir.rs
+++ b/crates/hir-ty/src/mir.rs
@@ -1274,6 +1274,9 @@ impl<'db> PlaceTy<'db> {
                     .get(f.0 as usize)
                     .cloned()
                     .unwrap_or_else(|| panic!("field {f:?} out of range: {self_ty:?}")),
+                TyKind::Array(..) | TyKind::Slice(..) => {
+                    Ty::new_error(infcx.interner, ErrorGuaranteed)
+                }
                 _ => panic!("can't project out of {self_ty:?}"),
             }
         }
--- a/crates/hir-ty/src/mir/lower/tests.rs
+++ b/crates/hir-ty/src/mir/lower/tests.rs
@@ -1,7 +1,14 @@
 use hir_def::DefWithBodyId;
+use rustc_type_ir::inherent::Ty as _;
 use test_fixture::WithFixture;
 
-use crate::{db::HirDatabase, setup_tracing, test_db::TestDB};
+use crate::{
+    db::HirDatabase,
+    mir::{FieldIndex, PlaceElem, PlaceTy, ProjectionElem},
+    next_solver::{DbInterner, ParamEnv, Ty, TypingMode, infer::DbInternerInferExt},
+    setup_tracing,
+    test_db::TestDB,
+};
 
 fn lower_mir(#[rust_analyzer::rust_fixture] ra_fixture: &str) {
     let _tracing = setup_tracing();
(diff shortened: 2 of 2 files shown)


ChayimFriedman2:
This is the wrong way to fix. Instead of patching the error, we should ask why we create a field projection for slices and fix *that*.

Also your test does not use the helpers in the same file as it should.

northaxosky:
Thanks for the feedback. It was pretty short sighted of me to PR with this band aid fix. I have not contributed a project of this scale before, and I am not quite confident enough in myself to propose a root-cause fix. The original panic came from the following code segment:
https://github.com/northaxosky/overseer/blob/621da8c34bedfc95e8c69e4887a6ecd279fca0c3/crates/overseer-tui/src/app/input/actions.rs#L92-L119
around `rows.get(index).copied()`, followed byt the matching on `ModPaneRow`. The logs showed a Field projection eventually reaching a `[ModPaneRow]`, but I havent traced where that projection is introduced. 
Ill step back from the implementation rather than submitting a speculative fix. If the reproduction or logs could be helpful, or if there is a particular lowering path worth investigating, I would be happy to look into it

PR #22759, opened 2026-07-10
Fix crashes in assists due to .unwrap() calls in SyntaxFactory

Wilfred:
These functions would panic when they try to re-parse code that isn't
valid in a top level function.

For parameters, this is `..`, which is valid in a closure but not a
function (requires an annotation). For tail block expressions, this is
`let var = expr`, which is valid in a tail position but not a function
body (requires a semicolon).

I've checked both quote! invocations against the rust.ungram file and
I think the structures are correct.

Include a test of an assist that previously caused a panic.

AI disclosure: Written with a little help by GPT-5.5.

### DIFF
--- a/crates/ide-assists/src/handlers/convert_closure_to_fn.rs
+++ b/crates/ide-assists/src/handlers/convert_closure_to_fn.rs
@@ -739,6 +739,25 @@ fn main() {
     }
 
     #[test]
+    fn handles_closures_with_unannotated_rest_patterns() {
+        check_assist(
+            convert_closure_to_fn,
+            r#"
+fn main() {
+    let closure = |$0..| ();
+}
+"#,
+            r#"
+fn main() {
+    fn closure(..: _) {
+        ()
+    }
+}
+"#,
+        );
+    }
+
+    #[test]
     fn multiple_capture_usages() {
         check_assist(
             convert_closure_to_fn,
--- a/crates/ide-assists/src/handlers/unwrap_branch.rs
+++ b/crates/ide-assists/src/handlers/unwrap_branch.rs
@@ -832,6 +832,25 @@ fn main() {
     }
 
     #[test]
+    fn regression_22759() {
+        check_assist(
+            unwrap_branch,
+            r#"
+fn main() {
+    match () {
+        () $0=> let x = (),
+    }
+}
+"#,
+            r#"
+fn main() {
+    let x = ()
+}
+"#,
+        );
+    }
+
+    #[test]
     fn simple_if_in_while_bad_cursor_position() {
         check_assist_not_applicable(
             unwrap_branch,
--- a/crates/syntax/src/ast/make.rs
+++ b/crates/syntax/src/ast/make.rs
@@ -565,7 +565,17 @@ pub fn async_move_block_expr(
 }
 
 pub fn tail_only_block_expr(tail_expr: ast::Expr) -> ast::BlockExpr {
-    ast_from_text(&format!("fn f() {{ {tail_expr} }}"))
+    quote! {
+        BlockExpr {
+            StmtList {
+                ['{']
+                " "
+                #tail_expr
+                " "
+                ['}']
+            }
+        }
+    }
 }
 
 /// Ideally this function wouldn't exist since it involves manual indenting.
@@ -1038,7 +1048,14 @@ pub fn untyped_param(pat: ast::Pat) -> ast::Param {
 }
 
 pub fn param(pat: ast::Pat, ty: ast::Type) -> ast::Param {
-    ast_from_text(&format!("fn f({pat}: {ty}) {{ }}"))
+    quote! {
+        Param {
+            #pat
+            [:]
+            " "
+            #ty
+        }
+    }
 }
 
 pub fn self_param() -> ast::SelfParam {


Veykril (commented):
I think this is the wrong way to guard agains this. The code rightfully assumes that the `make` module constructs well formed code if given well formed code. So I guess the issue we are running into now is the way the make module works, by stringifying and then re-parsing, which makes this entire assumption fairly brittle as different parser recovery states can kick in.

I wonder how far away we are from finally changing this now that we have removed the mutable APIs

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22759/9482996a746236c56c3333ef1bf796bd2d5f7272..d6bec934a217e85bbd7a7f309457dbf0d261b3a7)*

A4-Tacks (commented):
IMO, this should be fixed in the assists


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22759/9482996a746236c56c3333ef1bf796bd2d5f7272..d6bec934a217e85bbd7a7f309457dbf0d261b3a7)*

crates/ide-assists/src/handlers/unwrap_branch.rs
```diff
@@ -831,6 +831,25 @@ fn main() {
...
     }
 
+    #[test]
+    fn unwrap_match_arm_let_expr_without_crashing() {
+        check_assist(
+            unwrap_branch,
+            r#"
+fn main() {
+    match () {
+        _ $0=>$0 let x = (),
```
A4-Tacks: Maybe we should change `tail_only_block_expr()` to like `block_expr()` uses `quote! {}`


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22759/9482996a746236c56c3333ef1bf796bd2d5f7272..d6bec934a217e85bbd7a7f309457dbf0d261b3a7)*

crates/ide-assists/src/handlers/convert_closure_to_fn.rs
```diff
@@ -738,6 +738,44 @@ fn main() {
...
+        check_assist(
+            convert_closure_to_fn,
+            r#"
+fn main() {
+    let closure: fn(()) = |$0..| ();
+}
+"#,
+            r#"
+fn main() {
+    fn closure(..:) {
```
A4-Tacks: This should be fixed in assists to get `closure(..: _)`


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22759/9482996a746236c56c3333ef1bf796bd2d5f7272..d6bec934a217e85bbd7a7f309457dbf0d261b3a7)*

crates/ide-assists/src/handlers/convert_closure_to_fn.rs
```diff
@@ -738,6 +738,44 @@ fn main() {
...
+        );
+    }
+
+    #[test]
+    fn handles_untyped_param_pattern_mapping() {
+        check_assist(
+            convert_closure_to_fn,
+            r#"
+fn main() {
+    let closure: fn(i32) -> i32 = |$0x| x;
```
A4-Tacks: What's the use of this test


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22759/9482996a746236c56c3333ef1bf796bd2d5f7272..d6bec934a217e85bbd7a7f309457dbf0d261b3a7)*

A4-Tacks (changes requested):

ChayimFriedman2:
I agree with @Veykril that we should not make `make` constructors fallible. FWIW a possible future goal for the GSoC project of @Shourya742 (if we'll have time) was to implement auto-generation of make constructors without textual parsing, which will solve this.

bit-aloo:
> I agree with @Veykril that we should not make `make` constructors fallible. FWIW a possible future goal for the GSoC project of @Shourya742 (if we'll have time) was to implement auto-generation of make constructors without textual parsing, which will solve this.

We will, the only pre-requisite (AFAICT), for this gonna be moving trivia inside the tokens, and making our grammar generator smarter. Currently, working on removing mutable rowan API's.

crates/ide-assists/src/handlers/convert_closure_to_fn.rs
```diff
@@ -86,10 +86,15 @@ pub(crate) fn convert_closure_to_fn(acc: &mut Assists, ctx: &AssistContext<'_, '
             match node.ty() {
                 Some(_) => Some(node),
                 None => {
-                    let ty = param_ty
-                        .display_source_code(ctx.db(), module.into(), true)
-                        .unwrap_or_else(|_| "_".to_owned());
-                    Some(make.param(node.pat()?, make.ty(&ty)))
+                    let pat = node.pat()?;
+                    let ty = if matches!(&pat, ast::Pat::RestPat(_)) {
```
A4-Tacks: You should use something like `param_ty.is_unknown()`


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22759/268267dc6f1e55dc650bda3226cbc53b531a5146..4d014d3245e720a89f84b8c381f215c4b09a0c67)*

crates/ide-assists/src/handlers/unwrap_branch.rs
```diff
@@ -86,9 +86,13 @@ pub(crate) fn unwrap_branch(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -> O
         let mut replacement = extract_statements(replacement);
         let container = prefer_container.unwrap_or(container);
 
-        if ast::ExprStmt::can_cast(container.kind())
-            && block.tail_expr().is_some_and(|it| !it.is_block_like())
-        {
+        if block.tail_expr().is_some_and(|it| {
+            (ast::ExprStmt::can_cast(container.kind()) && !it.is_block_like())
+                || (matches!(it, ast::Expr::LetExpr(_))
```
A4-Tacks: Why changing this condition, is it not enough to fix it with `quote! {}` alone?


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22759/268267dc6f1e55dc650bda3226cbc53b531a5146..4d014d3245e720a89f84b8c381f215c4b09a0c67)*

A4-Tacks (changes requested):

crates/ide-assists/src/handlers/unwrap_branch.rs
Wilfred: Oops, sorry, I pushed work-in-progress experiments to this PR. I've dropped this commit because I think it's out of scope.

Whilst it's true that code like the following:

```
fn main() {
    match () {
        _ => let x = (),
    }
}
```

would be like this after the assist, which isn't syntactically valid:

```
fn main() {
    let x = () // needs ; here
}
```

I think it's way beyond fixing the crashes that started me investigating this issue, so I'm ignoring it for now. It seems relatively benign.

Wilfred:
OK, I've done another pass on this: using `quote!` fixes the crash cases I've seen, round-tripping properly. I've also tidied up the assist fixes, just focusing on the closure case.

I think this small and sensible now, but let me know if you'd like other changes.

crates/ide-assists/src/handlers/unwrap_branch.rs
A4-Tacks: I think you have solved the issue with the panic (uses `quote! {}`), try add a regression test:

```
fn main() {
    match () {
        _ $0=>$0 let x = (),
    }
}
```
->
```
fn main() {
    let x = ()
}
```

Note: This without semicolons, bare let-expr is not a valid rust code, and assistant does not need to guarantee it

Wilfred: The test in the second commit did crash prior to the `quote!` changes. Would you like me to move the test to the `quote!` commit, to show the old behaviour (no longer crashing, just invalid)?

(Edit: the above sounds kinda like an LLM, but I assure you I wrote it!)

crates/ide-assists/src/handlers/convert_closure_to_fn.rs
```diff
@@ -86,10 +86,15 @@ pub(crate) fn convert_closure_to_fn(acc: &mut Assists, ctx: &AssistContext<'_, '
...
-                    Some(make.param(node.pat()?, make.ty(&ty)))
+                    let ty = if param_ty.is_unknown() {
+                        make.ty_placeholder()
+                    } else {
+                        let ty = param_ty
+                            .display_source_code(ctx.db(), module.into(), true)
+                            .unwrap_or_else(|_| "_".to_owned());
+                        make.ty(&ty)
+                    };
+                    Some(make.param(node.pat()?, ty))
```
Veykril: Actually, this change makes it seem like `display_source_code` creates an invalid source type output, thats a bug in display_source_code. Not something we should work around in one assist

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22759/b669719e5a68446a9bf1676615e56ae888f7359d..918f5a6254e09b15228283909d8954780e380411)*

Veykril:
yea that looks better to me :)

A4-Tacks:
There seems to be a similar PR: #21578

crates/ide-assists/src/handlers/convert_closure_to_fn.rs
Wilfred: You're saying that I shouldn't call `.unwrap_or_else()` here, or you're saying that the `.is_unknown()` should be unnecessary here?

Veykril: I am saying that I am surprised `let ty = param_ty
                            .display_source_code(ctx.db(), module.into(), true)
                            .unwrap_or_else(|_| "_".to_owned());` isn't enough. So yea, the `is_unknown` branch should not be necessary imo

Wilfred:
OK, done another pass on this. I've got one regression test, and just changed the logic to use `quote!` in two places. I've also rewritten with the commit message with my current understanding.

As with #22782, I am seeing this cause crashes for a few users (8 people recently) even without invoking the assist, so I think it's worth porting to `quote!` even before we have the proper generated code for `quote!` usage.

Let me know what you think.

crates/ide-assists/src/handlers/unwrap_branch.rs
A4-Tacks: Commit this patch (I have tested it on 6faf4c and it passed):

```diff
diff --git a/crates/ide-assists/src/handlers/unwrap_branch.rs b/crates/ide-assists/src/handlers/unwrap_branch.rs
index a582af4e2..b5b72c34a 100644
--- a/crates/ide-assists/src/handlers/unwrap_branch.rs
+++ b/crates/ide-assists/src/handlers/unwrap_branch.rs
@@ -1293,4 +1293,23 @@ fn main() {
             "Unwrap block",
         );
     }
+
+    #[test]
+    fn regression_22759() {
+        check_assist(
+            unwrap_branch,
+            r#"
+fn main() {
+    match () {
+        () $0=> let x = (),
+    }
+}
+"#,
+            r#"
+fn main() {
+    let x = ()
+}
+"#,
+        );
+    }
 }
```

A4-Tacks:
> I've got one regression test, and just changed the logic to use `quote!` in two places.

What about your second regression test? I mentioned it in https://github.com/rust-lang/rust-analyzer/pull/22759#discussion_r3579992294

Wilfred:
Added the second regression test, sorry for the slow response.

A4-Tacks (approved):

PR #22763, opened 2026-07-11
internal: Funnel all child process spawning through stdx::process

Veykril:
And enforce it via clippy. This allows us to easily manage process lifetimes and ensure tear down of child processes when r-a exits.

Alternative to https://github.com/rust-lang/rust-analyzer/pull/22738, note this does nothing for macos though, as macos seems to lack a nice way of doing this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2357,7 +2357,6 @@ dependencies = [
  "parser",
  "paths",
  "proc-macro-api",
- "process-wrap",
  "profile",
  "project-model",
  "ra-ap-rustc_type_ir",
@@ -2695,6 +2694,7 @@ dependencies = [
  "jod-thread",
  "libc",
  "miow",
+ "process-wrap",
  "tracing",
  "windows-sys 0.61.2",
 ]
--- a/clippy.toml
+++ b/clippy.toml
@@ -6,4 +6,7 @@ disallowed-types = [
 
 disallowed-methods = [
     { path = "std::process::Command::new", reason = "use `toolchain::command` instead as it forces the choice of a working directory" },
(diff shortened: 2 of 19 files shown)


ChayimFriedman2:
This does seem like a better way :cry: (sad because I've spent quite a lot of time on solving and debugging that PR), but I have some notes on the implementation.

Veykril:
Did you mean to send those notes as a review that you didn't submit by accident? Or are those still coming? Not too sure how happy I am with the current impl here either

ChayimFriedman2:
I didn't send them or reviewed the code thoroughly yet; one comment, however, is that it seems to be possible to also do this on macOS (or any Unix), by associating the child processes with a process group then using `atexit()` to kill this group on exit.

Veykril:
Ah yea no worries then if you just haven't found the time yet, just sounded like you might not have sent the review by accident (has happened to me before)

ChayimFriedman2:
So, without diving deep into the code yet, I have two problems with the way this is implemented currently:

 1. This reimplements too much of the process API. We should only control spawning, and after that use standard stdlib APIs. In Windows, create a job in a static `LazyLock`, call `creation_flags(CREATE_SUSPENDED)` on the `Command` before spawning, spawn to get a `Child`, attach it to the job, then resume - resumption is a bit problematic because [`main_thread_handle()`](https://doc.rust-lang.org/nightly/std/os/windows/process/trait.ChildExt.html#tymethod.main_thread_handle) is currently unstable, so it seems the best way is to enumerate child threads and resume them. On Linux, use `pre_exec()` like you do; on other Unixes, use `atexit()` like I said above. We should still return `JodChild` to ensure termination on drop as well, but all APIs should just forward to std.
 2. I think that we should actually have our own `JodCommand` like in my PR; it both minimizes the risk for mistake as well as simplifies code that wants to spawn processes.

PR #22768, opened 2026-07-11
Issue resolved #22596

### LINKED ISSUE #22596: Enhanced enter key / onEnter erroneously deletes/interprets `$foo` occurring on the same line
**rust-analyzer version**: 0.4.2938-standalone (888f5810a4 2026-06-15)

**editor or extension**: VSCode, extension version  0.4.2938

```rust
macro_rules! foo {
    ($bar:tt) => {
        const { $bar }
    };
}
```

1. Place the cursor immediately to the right of `const {` on the third line.
2. Press Enter (invoke `rust-analyzer.onEnter`)

The result is:

```rust
macro_rules! foo {
    ($bar:tt) => {
        const {
            
        }
    };
}
```

The tokens `$bar` have been deleted, but should not be. This appears to happen to any occurrences of the *text* `$` followed by an identifier or number which are to the right of a `{`. Also, if it is the text `$0`, then an additional cursor appears there. This suggests that the problem is incorrect escaping of replacement text passed to VS Code.

j-ricardo-goncalves:
Solves rust-lang/rust-analyzer#22596 

by reusing the function escape_snippet_bits used in postfix.rs, the issue seems to be gone completely. made two test cases, using $bar and $0, as per the issue rust-lang/rust-analyzer#22596 by kpreid. a possible improvement would be moving escape_snippet_bits to place where both postfix.rs and on_enter.rs can see it, but im not sure where I would do that...

### DIFF
--- a/crates/ide/src/typing/on_enter.rs
+++ b/crates/ide/src/typing/on_enter.rs
@@ -1,7 +1,7 @@
 //! Handles the `Enter` key press, including comment continuation and
 //! indentation in brace-delimited constructs.
 
-use ide_db::{FilePosition, RootDatabase};
+use ide_db::{FilePosition, RootDatabase, source_change::SnippetEdit};
 use syntax::{
     AstNode, SmolStr, SourceFile,
     SyntaxKind::*,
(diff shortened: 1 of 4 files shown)


j-ricardo-goncalves:
Ok, well, it seems to fix the $bar one but its still failing the $0 test which I added without much thought
Edit: removed the $0 test case

crates/ide/src/typing/on_enter.rs
```diff
@@ -108,12 +108,19 @@ fn on_enter_in_comment(
     Some(edit)
 }
 
+// for lack of a better place to put this function
+fn escape_snippet_bits(text: &mut String) {
```
ChayimFriedman2: We already have this function in ide-completion, so it makes sense to unite them, e.g. in ide-db.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22768/5be5e89cf0145d73a85c805726821b4adfc3af48..9b1bffd176a52888f87f885502d27f78596eb4cd)*

ChayimFriedman2 (changes requested):

crates/ide/src/typing/on_enter.rs
j-ricardo-goncalves: Ok, move it to ide-db/source_change.rs in the implementation of SnippetEdit. 
Thought it was the most appropriate place where it should go.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22768/5be5e89cf0145d73a85c805726821b4adfc3af48..316f8afc1ed60623ae7452ced068cc109325009a)*

PR #22770, opened 2026-07-12
fix: clamp cttz const-eval result to type width

qti3e:
`0u64.trailing_zeros()` is showing the wrong result in the const eval hover/preview.

image

```
cargo run    
   Compiling ra-tzcnt-bug v0.1.0 (/home/qti3e/Code/Challenge/ra-tzcnt-bug)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/ra-tzcnt-bug`
[src/main.rs:4:5] X = 64
```

detail: the impl was converting any input into a u128 to handle generics, which was making it return `128` regardless of the input's type, the fix clamps the result to the number of bits in the input.

### DIFF
--- a/crates/hir-ty/src/consteval/tests/intrinsics.rs
+++ b/crates/hir-ty/src/consteval/tests/intrinsics.rs
@@ -691,6 +691,33 @@ fn cttz() {
         "#,
         3,
     );
+    check_number(
+        r#"
+        #[rustc_intrinsic]
+        pub fn cttz<T: Copy>(x: T) -> T;
+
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (approved):
Yeah I guess we have many more such bugs in MIR.

Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22770/5be5e89cf0145d73a85c805726821b4adfc3af48..349bda8a8fea14dcc248b56d96efdb98c493ea15)*

PR #22786, opened 2026-07-14
perf - Reduce flyimport ordering allocations

mingley:
## Summary
fix https://github.com/rust-lang/rust-analyzer/blob/63a6f0d4bcfd3bbcf36383fcbcbcd93456ed1653/crates/ide-completion/src/completions/flyimport.rs#L443

Instead of recalculating and allocating a lowercase string for every sort comparison, each key is now calculated once per candidate using allocation-free matching.

- Compute fly-import ordering keys once per candidate.
- Avoid allocating lowercase strings during ordering.
- Preserve existing completion ordering.

## Testing Done

- `cargo test`
- `cargo clippy`
- Added additional unit tests to verify case-insensitive ranking, import-path tie-breaking, later matches, and fuzzy-only matches through actual completion output

Local 100-candidate microbenchmark produced:

- Allocations: `1,339` → `1`
- Median time: `41.8µs` → `2.0µs`

### DIFF
--- a/crates/ide-completion/src/completions/flyimport.rs
+++ b/crates/ide-completion/src/completions/flyimport.rs
@@ -275,15 +275,15 @@ fn import_on_the_fly<'db>(
                 && ctx.check_stability(original_item.attrs(ctx.db).as_ref())
         })
         .filter(|import| filter_excluded_flyimport(ctx, import))
-        .sorted_by(|a, b| {
-            let key = |import_path| {
-                (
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
Despite what the comment says, this code is not really perf sensitive. Your code is more complicated and also O(N^2). Therefore I'm closing this as not needed.

Also, did you use AI tools to author the PR description? Please note that our [AI policy](https://github.com/rust-lang/rust-analyzer/blob/master/AI_POLICY.md) forbids doing so.

PR #22790, opened 2026-07-14
fix: Assertion failure on dyn trait in closure

Wilfred:
Similarly to rust-lang/rust-analyzer#22677, it's possible to get an assertion failure in HIR inference after error recovery.

The best repro I've found is actually the sample code from Rust issue
157951. I do think it's a legitimate fix, I've seen two real users hit this assertion.

Give up if we don't have a closure kind in
DeferredCallResolution::resolve(), and add a regression test that previously asserted.

AI disclosure: Implemented with some help by GPT-5.5.


ChayimFriedman2:
Unlike the previous PR, this code was not changed in rustc and yet does not ICE. I want to understand why before removing the assertion.

PR #22791, opened 2026-07-14
Enable auto trait inference

Veykril:


ChayimFriedman2:
What is the impact on analysis-stats times?

Veykril:
Was waiting on CI for that. Surprisingly, none? 

Times from https://github.com/rust-lang/rust-analyzer/pull/22789
```
Inference:           96.47s, 832mb
Total:               139.50s, 1426mb
```
vs this PR
```
Inference:           97.78s, 840mb
Total:               141.06s, 1426mb
```

ChayimFriedman2:
CI runs in debug, it's not a reliable benchmark.

Also, benchmarking even the same machine in different times is not a good idea, doubly so a CI machine.

Veykril:
Fair enough, it does work as a decent indicator usually though. Can't check the timings for this today

Veykril:
baseline
Inference:           37.07s, 956mb
this PR
Inference:           37.89s, 969mb

So yea its within noise (with slight memory increase, since presumably we fill out a couple more query slots)

PR #22792, opened 2026-07-14
Fix extract variable preserving whitespace from macro input

### LINKED ISSUE #22767: Whitespace not respected during extract variable from macro
Firstly, thanks for the Rust Analyzer, I don't know how I'd survive without it. 

**rust-analyzer version**:  cecae478915cf3602a81038dd47301e58a5c6484 (git hash)

**rustc version**:  rustc 1.97.0 (2d8144b78 2026-07-07)

**repository link (if public, optional)**: https://github.com/rust-lang/rust-analyzer

```rust
//- minicore:assert
fn check(value: &mut usize) -> bool {
    false
}
fn foo(mut bar: usize) {
    assert!(check($0&mut bar$0));
}
```
```rust
//- minicore:assert
fn main() {
  assert_eq!($0if true {1} else {0}$0, 3);
}
```
With most expressions eg `1 + 1`, it doesn't make a difference, but it matters when two "words" follow each other.

**Notes**
I had a look into fixing this, but the issue seems to be here
https://github.com/rust-lang/rust-analyzer/blob/5be5e89cf0145d73a85c805726821b4adfc3af48/crates/ide-assists/src/handlers/extract_variable.rs#L98-L99
`first` and `last` have whitespace tokens but `first_descend` and `last_descend` do not.
When it tries to collect the tokens from the macro, it doesn't include whitespace. 
I couldn't understand what the problem was, but hopefully this knowledge is helpful.

cuishuang:
When extracting an expression selected inside a macro token tree, the assist descends into the macro expansion for semantic analysis. The expanded syntax may not preserve whitespace and comments from the original macro input, so the generated initializer could contain invalid text such as `&mutbar` or `iftrue`.

This keeps using the expanded expression for semantic analysis and replacement mapping, while building the initializer from the original source text covered by the final replacement range. 

Regression tests cover required whitespace in `&mut bar` and `if true`, source formatting inside token-tree macros, and parenthesized expressions.

Fixes rust-lang/rust-analyzer#22767.

### DIFF
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -1,4 +1,4 @@
-use hir::{HirDisplay, TypeInfo};
+use hir::{HirDisplay, Module, TypeInfo};
 use ide_db::{
     assists::GroupLabel,
     syntax_helpers::{LexedStr, suggest_name},
@@ -10,6 +10,7 @@ use syntax::{
         self, AstNode,
         edit::{AstNodeEdit, IndentLevel},
     },
+    hacks::parse_expr_from_str,
     syntax_editor::{Element, Position},
 };
 
@@ -92,7 +93,7 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
     let node = node.ancestors().take_while(|anc| anc.text_range() == node.text_range()).last()?;
     let range = node.text_range();
 
-    let (to_replace, analysis) = if node.kind() == SyntaxKind::TOKEN_TREE {
+    let (to_replace, analysis, original_expr) = if node.kind() == SyntaxKind::TOKEN_TREE {
         let (first, last) = extract_token_range_of(&node, ctx.selection_trimmed())?;
 
         let first_descend = ctx.sema.descend_into_macros_single_exact(first.clone());
@@ -111,14 +112,15 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
         if !node.text_range().contains_range(original_range.range) {
             return None;
         }
-        (cover_edit_range(&node, original_range.range), expr)
+        let original_expr = original_expr(ctx, original_range.file_id, original_range.range);
+        (cover_edit_range(&node, original_range.range), expr, original_expr)
     } else {
         let expr = node
             .descendants()
             .take_while(|it| range.contains_range(it.text_range()))
             .find_map(valid_target_expr(ctx))?;
         let to_extract = expr.syntax().syntax_element();
-        (to_extract.clone()..=to_extract, expr)
+        (to_extract.clone()..=to_extract, expr, None)
     };
     let place = match to_replace.start() {
         NodeOrToken::Node(node) => node.clone(),
@@ -217,6 +219,10 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
                     editor.add_annotation(pat_name.syntax().clone(), tabstop);
                 }
 
+                let to_extract_no_ref = original_expr
+                    .clone()
+                    .map(peel_parens)
+                    .unwrap_or_else(|| prettify_macro_expr(ctx, module, to_extract_no_ref.clone()));
                 let initializer = match ty.as_ref().filter(|_| needs_ref) {
                     Some(receiver_type) if receiver_type.is_mutable_reference() => {
                         make.expr_ref(to_extract_no_ref.clone(), true)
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -111,14 +112,15 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
         if !node.text_range().contains_range(original_range.range) {
             return None;
         }
-        (cover_edit_range(&node, original_range.range), expr)
+        let original_expr = original_expr(ctx, original_range.file_id, original_range.range);
+        (cover_edit_range(&node, original_range.range), expr, original_expr)
```
A4-Tacks: I think relying solely on `to_replace` is sufficient,
`slice(to_replace.start.text_range().cover(to_replace.end.text_range()))`

And for `peel_parens()`, can create a `fn (RangeInclusive<SyntaxElement>) -> RangeInclusive<SyntaxElement>` version
For `identity!($0(1+2)$0)` expected `let var_name = 1+2;`, but you outputs `let var_name = (1+2);`. however, it is acceptable


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22792/229e3bffd15badac120119acb796d5cbe85dcc56..7cc09005f584ccaeb5b93733b94aed6ef77d1e52)*

crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -217,6 +219,10 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
                     editor.add_annotation(pat_name.syntax().clone(), tabstop);
                 }
 
+                let to_extract_no_ref = original_expr
+                    .clone()
+                    .map(peel_parens)
+                    .unwrap_or_else(|| prettify_macro_expr(ctx, module, to_extract_no_ref.clone()));
```
A4-Tacks: No need to prettify, just use the source code text


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22792/229e3bffd15badac120119acb796d5cbe85dcc56..7cc09005f584ccaeb5b93733b94aed6ef77d1e52)*

A4-Tacks (changes requested):

crates/ide-assists/src/handlers/extract_variable.rs
cuishuang: Thanks for your suggesstion, updated to rely on the final `to_replace` range when reading the original source text. I also added a regression test for `identity!($0(1+2)$0)`. It currently preserves the parentheses, as you noted is acceptable.

crates/ide-assists/src/handlers/extract_variable.rs
cuishuang: Updated. I removed the `prettify_macro_expansion()` fallback and now parse the initializer directly from the source text covered by `to_replace`. Please review it again when you have time.

crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -217,6 +220,12 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
                     editor.add_annotation(pat_name.syntax().clone(), tabstop);
                 }
 
+                let to_extract_no_ref = if prefer_source_expr {
```
A4-Tacks: I think we can just `let to_extract_no_ref = source_expr(ctx, to_replace.clone()).unwrap()` directly


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22792/229e3bffd15badac120119acb796d5cbe85dcc56..a2775d068f3962b8a4ec8569d9d90ff0f91de27e)*

cuishuang: I tried that simplification locally. It works for the token-tree path, but applying it unconditionally breaks existing non-token-tree behavior.

For field shorthand, the shadowed `to_replace` can cover an entire `RecordExprField`, so `source_expr()` returns `None` and the unwrap panics. It also bypasses the existing parenthesis and dereference adjustments, causing `generates_no_ref_for_deref` and the generated extract-variable doctest to fail.

I therefore changed the token-tree path to unwrap the source expression directly, while keeping the existing adjusted expression for non-token-tree cases. I also renamed the flag to `use_source_expr`.

A4-Tacks: > It also bypasses the existing parenthesis and dereference adjustments, causing `generates_no_ref_for_deref` and the generated extract-variable doctest to fail.

You are right.

A4-Tacks (approved):

PR #22793, opened 2026-07-14
fix!: use `Result` for the lsp-server `Response` payload type

ribru17:
We don't need a new type here thanks to serde's `remote`. This is a follow-up to https://github.com/rust-lang/rust-analyzer/pull/22753#issuecomment-4948514165.

### DIFF
--- a/lib/lsp-server/examples/minimal_lsp.rs
+++ b/lib/lsp-server/examples/minimal_lsp.rs
@@ -82,9 +82,7 @@ use toolchain::command; // clippy-approved wrapper
 
 #[allow(clippy::print_stderr, clippy::disallowed_types, clippy::disallowed_methods)]
 use anyhow::{Context, Result, anyhow, bail};
-use lsp_server::{
-    Connection, Message, Request as ServerRequest, RequestId, Response, ResponseKind,
-};
+use lsp_server::{Connection, Message, Request as ServerRequest, RequestId, Response};
 
 // =====================================================================
 // main
@@ -306,8 +304,7 @@ fn full_range(text: &str) -> Range {
 }
 
 fn send_ok<T: serde::Serialize>(conn: &Connection, id: RequestId, result: &T) -> Result<()> {
-    let resp =
-        Response { id, response_kind: ResponseKind::Ok { result: serde_json::to_value(result)? } };
+    let resp = Response { id, response_result: Ok(serde_json::to_value(result)?) };
     conn.sender.send(Message::Response(resp))?;
     Ok(())
 }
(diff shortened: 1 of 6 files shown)


lib/lsp-server/examples/minimal_lsp.rs
```diff
@@ -306,8 +304,7 @@ fn full_range(text: &str) -> Range {
 }
 
 fn send_ok<T: serde::Serialize>(conn: &Connection, id: RequestId, result: &T) -> Result<()> {
-    let resp =
-        Response { id, response_kind: ResponseKind::Ok { result: serde_json::to_value(result)? } };
+    let resp = Response { id, response_result: Result::Ok(serde_json::to_value(result)?) };
```
ChayimFriedman2: `Result::Ok` and `Result::Err` are weird to see in Rust, we just use `Ok` and `Err`. Same in the other code.

lib/lsp-server/src/msg.rs
```diff
@@ -365,4 +364,27 @@ mod tests {
...
+    fn serialize_err_response() {
+        let msg = Message::Response(Response::new_err(
+            RequestId::from(3),
+            -32600,
+            String::from("bad response message"),
+        ));
+        let serialized = serde_json::to_string(&msg).unwrap();
+
+        assert_eq!(
+            "{\"id\":3,\"error\":{\"code\":-32600,\"message\":\"bad response message\"}}",
```
ChayimFriedman2: Use a raw string here?

ChayimFriedman2 (changes requested):
Two nits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22793/229e3bffd15badac120119acb796d5cbe85dcc56..2885e3934431e461673592c61aa7bf84ef4a6158)*

Veykril:
I am confused, what is the benefit of reusing rust's Result here? We need a duplicate anyways for serde derive. Imo we should stick with the original change here

ChayimFriedman2:
Because it is a public field and `Result` is nice and well known.

ribru17:
another benefit i can think of: being able to use `?`

lib/lsp-server/examples/minimal_lsp.rs
ribru17: Done.

lib/lsp-server/src/msg.rs
ribru17: Done, applied to the other JSON test strings there as well.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22793/76d092ea9d27a66c53aedf630c07e3dae42db1c1..99d6571707b77560c846965a05dca6c7630e0a6a)*

PR #22794, opened 2026-07-14
tidy - flyimport lowercase buffer (last-ditch attempt)

mingley:
## Summary

Last ditch attempt at addressing a FIXME comment https://github.com/rust-lang/rust-analyzer/blob/63a6f0d4bcfd3bbcf36383fcbcbcd93456ed1653/crates/ide-completion/src/completions/flyimport.rs#L443

Goals:
- Simple.
- Better.
- No more fixme.

Previous beneficial perf implications on https://github.com/rust-lang/rust-analyzer/pull/22786 were overstated.

While https://github.com/rust-lang/rust-analyzer/pull/22786 did "fix" the allocation thrashing mentioned in the FIXME, it naively increased worst case time complexity while adding additional implementation complexity. This is not good.

Thank you @ChayimFriedman2 for keeping me honest. https://github.com/rust-lang/rust-analyzer/pull/22786#issuecomment-4965017126

So now, I attempt to preserve existing sorting and substring-search behavior, while adding a lowercase string buffer to support reuse of allocated capacity instead of allocating a new string for each key computation. This should address the fixme while also not increasing time complexity, and still keeps it pretty similar to original implementation behavior.

## Testing Done
- `cargo test`
- `cargo clippy`
- Added one additional test covering case-insensitive ranking and path tie-breaking behavior.

### DIFF
--- a/crates/ide-completion/src/completions/flyimport.rs
+++ b/crates/ide-completion/src/completions/flyimport.rs
@@ -262,6 +262,7 @@ fn import_on_the_fly<'db>(
         }
     };
     let user_input_lowercased = potential_import_name.to_lowercase();
+    let mut import_name_buffer = String::new();
 
     let import_cfg = ctx.config.import_path_config();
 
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (approved):
Thanks, now it is good.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22794/0a4d1675878a339f7ef6578e55adb904970817db..a3aa30fa410eb492ed32163ad431d36200e3a51d)*

PR #22808, opened 2026-07-15
remove clone_for_update / mutable rowan API's

bit-aloo:
This PR removes the rowan mutable APIs and pave a path for future rowan changes.

### DIFF
--- a/crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs
+++ b/crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs
@@ -141,7 +141,7 @@ pub(crate) fn generate_blanket_trait_impl(
             if let Some(cap) = ctx.config.snippet_cap
                 && let Some(self_ty) = impl_.self_ty()
             {
-                builder.add_tabstop_before(cap, self_ty);
+                editor.add_annotation(self_ty.syntax(), builder.make_tabstop_before(cap));
             }
             builder.add_file_edits(ctx.vfs_file_id(), editor);
         },
--- a/crates/syntax/src/syntax_editor/edit_algo.rs
+++ b/crates/syntax/src/syntax_editor/edit_algo.rs
@@ -200,6 +200,9 @@ impl EditPlan {
     }
 
     /// Checks that replacement at the same tree depth do not overlap
+    ///
+    /// `changes` is sorted by range start, so overlap is a single comparison against the
+    /// last range at that key, and `insert` can throw away the range it evicts.
     fn replacements_are_disjoint(
         changes: &[Change],
         mut node_depth: impl FnMut(SyntaxNode) -> usize,
--- a/crates/syntax/src/syntax_editor/edit_algo.rs
+++ b/crates/syntax/src/syntax_editor/edit_algo.rs
@@ -1,415 +1,697 @@
 //! Implementation of applying changes to a syntax tree.
 
-use std::{
-    cmp::Ordering,
-    collections::VecDeque,
-    ops::{Range, RangeInclusive},
-};
+use std::{cmp::Ordering, ops::Range};
 
 use rowan::TextRange;
 use rustc_hash::FxHashMap;
 use stdx::format_to;
 
-use crate::{
-    SyntaxElement, SyntaxNode, SyntaxNodePtr,
-    syntax_editor::{Change, ChangeKind, PositionRepr, mapping::MissingMapping},
+use crate::{NodeOrToken, SyntaxElement, SyntaxNode};
+
+use super::{
+    Change, ChangeKind, PositionRepr, SyntaxAnnotation, SyntaxEdit, SyntaxEditor, SyntaxMapping,
+    mapping::MissingMapping,
 };
 
-use super::{SyntaxEdit, SyntaxEditor};
+/// A validated batch of changes in the exact order in which it must execute.
+///
+/// Planning is deliberately separate from tree mutation. Once an `EditPlan`
+/// exists, execution does not need to reason about overlaps, dependencies, or
+/// source ordering.
+struct EditPlan {
+    changes: Vec<PlannedChange>,
+}
 
+/// A change whose target tree and output tracking are fully known.
+struct PlannedChange {
+    tree: SyntaxNode,
+    change: Change,
+    record_as_changed: bool,
+}
+
+impl PlannedChange {
+    /// Returns the immutable source elements that this change will slice in.
+    fn replacement_elements(&self) -> &[SyntaxElement] {
+        match &self.change {
+            Change::Insert(_, element) | Change::Replace(_, Some(element)) => {
+                std::slice::from_ref(element)
+            }
+            Change::InsertAll(_, elements)
+            | Change::ReplaceWithMany(_, elements)
+            | Change::ReplaceAll(_, elements) => elements,
+            Change::Replace(_, None) => &[],
+        }
+    }
+}
+
+/// The dependency info accumulated from one source ordered changes.
+///
+/// `parent` is an edge to the nearest containing node replacement. A discarded
+/// entry has no executable graph node because an ancestor deletion or ambiguous
+/// range replacement has mde its target unavailable.
+#[derive(Clone, Copy, Default)]
+struct PlanEntry {
+    parent: Option<usize>,
+    discarded: bool,
+}
+
+/// Planning failure containing the source-ordered changes use for diag.
+struct InvalidEditPlan {
+    changes: Vec<Change>,
+}
+
+impl EditPlan {
+    /// Validates raw editor changes and turns them into an execution schedule.
+    ///
+    /// The input is first sorted in source order, dependent targets are then
+    /// rewritten from their input trees into ancestor replacement trees. Finally,
+    /// discarded changes are removed and the dependency forest is traversed in
+    /// postorder.
+    /// Independent roots and sibling changes are prioritized right to left.
+    fn build(
+        mut changes: Vec<Change>,
+        mappings: &SyntaxMapping,
+        mut node_depth: impl FnMut(SyntaxNode) -> usize,
+    ) -> Result<Self, InvalidEditPlan> {
+        changes.sort_by(|left, right| {
+            left.target_range()
+                .start()
+                .cmp(&right.target_range().start())
+                .then_with(|| {
+                    let left_target = left.target_parent();
+                    let right_target = right.target_parent();
+                    if left_target == right_target {
+                        Ordering::Equal
+                    } else {
+                        node_depth(left_target).cmp(&node_depth(right_target))
+                    }
+                })
+                .then(left.change_kind().cmp(&right.change_kind()))
+        });
+
+        if !Self::replacements_are_disjoint(&changes, &mut node_depth) {
+            return Err(InvalidEditPlan { changes });
+        }
+
+        let mut entries = vec![PlanEntry::default(); changes.len()];
+        let mut regions_by_tree = FxHashMap::<SyntaxNode, Vec<ChangedRegion>>::default();
+
+        for (index, change) in changes.iter().enumerate() {
+            let target_tree = change.target_parent().tree_top();
+            let regions = regions_by_tree.entry(target_tree).or_default();
+            if let Some(region_index) = regions
+                .iter()
+                .rposition(|region| region.range.contains_range(change.target_range()))
+            {
+                regions.truncate(region_index + 1);
+                match regions[region_index].nested_changes {
+                    NestedChanges::Remap => {
+                        entries[index].parent = Some(regions[region_index].change_index);
+                    }
+                    NestedChanges::Discard => entries[index].discarded = true,
+                }
+            } else {
+                regions.clear();
+            }
+
+            if let Some(region) = ChangedRegion::for_change(change, index, entries[index].discarded)
+            {
+                regions.push(region);
+            }
+        }
+
+        // Work from the innermost dependency towards the outermost one. This
+        // lets a chain A -> B -> C rewrite C into B before B itself is mapped
+        // into A's replacement tree.
+        for (child, entry) in entries.iter().enumerate().rev() {
+            if let Some(parent) = entry.parent {
+                Self::rewrite_dependent_target(&mut changes, parent, child, mappings);
+            }
+        }
+
+        let mut children = vec![Vec::new(); changes.len()];
+        for (child, entry) in entries.iter().enumerate() {
+            if let Some(parent) = entry.parent {
+                children[parent].push(child);
+            }
+        }
+        for siblings in &mut children {
+            siblings.sort_by(|&left, &right| {
+                Self::execution_priority(&changes[left], &changes[right], &mut node_depth)
+            });
+        }
+
+        let mut roots = entries
+            .iter()
+            .enumerate()
+            .filter_map(|(index, entry)| {
+                (!entry.discarded && entry.parent.is_none()).then_some(index)
+            })
+            .collect::<Vec<_>>();
+        roots.sort_by(|&left, &right| {
+            Self::execution_priority(&changes[left], &changes[right], &mut node_depth)
+        });
+
+        let mut planned = changes
+            .into_iter()
+            .zip(entries)
+            .map(|(change, entry)| {
+                (!entry.discarded).then_some(PlannedChange {
+                    tree: change.target_parent().tree_top(),
+                    change,
+                    record_as_changed: entry.parent.is_none(),
+                })
+            })
+            .collect::<Vec<_>>();
+
+        let mut ordered = Vec::new();
+        for root in roots {
+            Self::append_postorder(root, &children, &mut planned, &mut ordered);
+        }
+
+        Ok(Self { changes: ordered })
+    }
+
+    /// Orders disjoint changes from right to left, with deeper ties first.
+    fn execution_priority(
+        left: &Change,
+        right: &Change,
+        node_depth: &mut impl FnMut(SyntaxNode) -> usize,
+    ) -> Ordering {
+        right
+            .target_range()
+            .start()
+            .cmp(&left.target_range().start())
+            .then_with(|| node_depth(right.target_parent()).cmp(&node_depth(left.target_parent())))
+            .then(right.change_kind().cmp(&left.change_kind()))
+    }
+
+    /// Appends a dependency subtree in post order fashion.
+    fn append_postorder(
+        index: usize,
+        children: &[Vec<usize>],
+        planned: &mut [Option<PlannedChange>],
+        ordered: &mut Vec<PlannedChange>,
+    ) {
+        for &child in &children[index] {
+            Self::append_postorder(child, children, planned, ordered);
+        }
+        ordered.push(planned[index].take().expect("reachable plan nodes are not discarded"));
+    }
+
+    /// Checks that replacement at the same tree depth do not overlap
+    fn replacements_are_disjoint(
+        changes: &[Change],
+        mut node_depth: impl FnMut(SyntaxNode) -> usize,
+    ) -> bool {
+        let mut previous = FxHashMap::<(SyntaxNode, usize), TextRange>::default();
+        for change in changes {
+            if !matches!(change.change_kind(), ChangeKind::Replace | ChangeKind::ReplaceRange) {
+                continue;
+            }
+
+            let parent = change.target_parent();
+            let key = (parent.tree_top(), node_depth(parent));
+            if previous
+                .insert(key, change.target_range())
+                .is_some_and(|range| range.end() > change.target_range().start())
+            {
+                return false;
+            }
+        }
+        true
+    }
+
+    /// Maps one dependent change into its ancestor replacement tree.
+    fn rewrite_dependent_target(
+        changes: &mut [Change],
+        parent: usize,
+        child: usize,
+        mappings: &SyntaxMapping,
+    ) {
+        let (input_ancestor, output_ancestor) = match &changes[parent] {
+            Change::Replace(
+                SyntaxElement::Node(target),
+                Some(SyntaxElement::Node(replacement)),
+            ) => (target.clone(), replacement.clone()),
+            _ => unreachable!("only node replacements can own dependent changes"),
+        };
+
+        let upmap_node = |target: &SyntaxNode| {
+            mappings.upmap_child(target, &input_ancestor, &output_ancestor).unwrap_or_else(
+                |MissingMapping(current)| {
+                    panic!(
+                        "no mappings exist between {current:?} (ancestor of {input_ancestor:?}) and {output_ancestor:?}"
+                    )
+                },
+            )
+        };
+        let upmap_element = |target: &SyntaxElement| {
+            mappings.upmap_child_element(target, &input_ancestor, &output_ancestor).unwrap_or_else(
+                |MissingMapping(current)| {
+                    panic!(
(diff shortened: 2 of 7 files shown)


ChayimFriedman2:
@Shourya742 Is this ready for review?

bit-aloo:
> @Shourya742 Is this ready for review?

Writing a huge comment explaining the changes, give me a few minutes. I had to revamp the edit_algo a bit, but it's now much more structured than (aligns with my mental model) before and works like a charm without the mutable API.

bit-aloo:
Processing pipeline did change a bit, though initial processing kinda remain the same.

Edits are processed in four phases:

1. Order and validate changes. Changes are sorted by source offset, target depth, and change kind. Overlapping replacements at the same depth are rejected because their result would depend on application order. (This is same as before)

2. Build dependency forest. A change contained in a replaced node becomes a child of that replacement. Its target is mapped from the input node into the generated replacement tree using `SyntaxMapping`. Changes inside deleted nodes, or inside a replaced range that has no unambiguous one-to-one mapping, are omitted from the plan. (This is also same but with better terminology for mind mapping)

3. Execute the dependency forest. Independent changes form roots and dependent changes form child edges. A postorder traversal applies every dependent edit before the ancestor that consumes its replacement tree. Disjoint roots and siblings execute from right to left so child indices remain stable.
 
4. Resolve tracked results. Paths for changed elements and annotations are resolved against the final root. Every returned element consequently belongs to SyntaxEdit::new_root.

 ```text
 Vec<Change>
     |
     v
 +------------------+
 | order + validate |  reject ambiguous same-depth replacement overlap
 +------------------+
     |
     v
 +--------------------------+
 | build dependency forest  |  parent edge = nearest containing replacement
 +--------------------------+
     |
     v
 +--------------------------+
 | postorder execution plan |  children before parent; siblings right to left
 +--------------------------+
     |
     v
 +--------------------------+
 | execute in TreeStore     |  one immutable TreeState per source root
 +--------------------------+
     |
     v
 SyntaxEdit { new_root, changed_elements, annotations }
 ```

 A nested edit is represented by an edge (previously we had that weird sorting rule). For example, suppose `R` replaces node `A` with generated node `A'`, and `D1` and `D2` edit descendants mapped into `A'`:

 ```text
 dependency forest                 required postorder

          R        I                      D2
          |                                |
         D1                               v
          |                              D1
         D2                                |
                                           v
                                          R

 I is an independent sibling to R by source position.
 ```

Executing `R` before `D1` would copy `A'` into its destination too early. Postorder makes the dependency explicit: `D2` updates the payload used by `D1`, and the resulting payload is then consumed by `R`.


 Every `SyntaxElement` belongs to one syntax-tree root. Calling `tree_top()` gives us that root, and we use it as the tree's identity:

 ```text
 editor element                         factory element
       |                                      |
       v                                      v
   tree_top()                             tree_top()
       |                                      |
       v                                      v
 T0: editor root                      T1: detached factory root
 ```

There is no single tree containing the editor root plus everything produced by `SyntaxFactory`. Factory constructors produce detached trees. A mapping says that an input node corresponds to a copied node in one of those trees; it does not merge the trees or change which root owns either node. (This will help later with terminologies)

Walkthrough

Say the editor starts with this tree, which we will call `T0`:

 ```text
 fn demo() {
     let result = x + y;
 }
```

An assist uses the editor's `SyntaxFactory` to build `check(x + y)`. The final call lives in another detached tree, `T1`. The factory may create a few intermediate detached fragments while building it, but the mappings are followed outward to the copy in the final generated tree:

```text
T0: editor tree

 BinExpr `x + y`
      |
      `-------------------- mapping --------------------.
                                                          |
 T1: final generated tree                                 |
                                                          |
 CallExpr `check(x + y)`                                  |
 `-- ArgList                                              |
     `-- BinExpr `x + y` <--------------------------------'
         `-- NameRef `y`

The mapped pair is BinExpr -> BinExpr. The CallExpr merely owns the output copy; mappings do not map a BinExpr to a node of a different kind.
 ```

 Now suppose three changes were recorded:

 ```text
 R: replace T0's `x + y` with T1's `check(x + y)`
 D: replace T0's `y` with a detached `z`
 I: insert `let traced = true;` after the let statement in T0
 ```
 Step 1: order and validate

`EditPlan::build` sorts by source start, then target depth, then change kind. An ancestor consequently appears before a descendant that starts at the same position. For this example the order is:

 ```text
 T0 source:

 fn demo() {
     let result = [x + [y] D] R;
                                  ^ I
 }

 source-ordered vector: [R, D, I]

 R and D overlap in text, but they are at different tree depths:

 depth n:       [------------- R -------------]
 depth n + 1:                    [-- D --]

 This structural order is valid. Two partially overlapping replacements at the same depth would be ambiguous and reject the plan.
 ```

 Step 2: discover containment

The planner maintains one stack of active replacement regions per source tree. Walking `[R, D, I]` produces:

 ```text
 visit R:
   active regions in T0: []
   R replaces one Node with one Node, so descendants can be mapped
   active regions in T0: [R(Remap)]
   entry[R] = { parent: None, discarded: false }

 visit D:
   D.range is inside R.range
   entry[D] = { parent: Some(R), discarded: false }
   dependency edge: R -> D

 visit I:
   I is outside R.range, so the active T0 stack is cleared
   entry[I] = { parent: None, discarded: false }

 resulting forest:

       R                    I
       |
       D
 ```

If `R` were a deletion, a range replacement, its region would be `Discard`. `D` would then be omitted because there would be no unique generated descendant to edit.

 Step 3: rewrite dependent targets

Before planning, `D` targets `T0::y`. The factory mapping lets us rewrite that target to the copied `T1::y`:

 ```text
 D before rewriting                                        D after rewriting

 target = T0::NameRef("y")                        target = T1::NameRef("y")
              |                                                     ^
              `-------- SyntaxMapping -------------'

 target.tree_top() == T0              target.tree_top() == T1
 ```
 Step 4: produce the execution schedule

`R` and `I` are independent roots. Because `I` is farther right in `T0`, it is ordered first. Postorder expands `R` as its child `D`, followed by `R`:

 ```text
 roots, right to left:       [I, R]
                                  |
 postorder R:                    [D, R]
                                  |
 final EditPlan:             [I, D, R]
 ```

A global depth or offset sort cannot express this requirement. `D` must run before `R` even though its target is in a different tree after rewriting, because `R` consumes the tree produced by `D`.

Step 5: initialize tree states

`TreeStore` lazily maintains an independent state for each immutable root:

 ```text
 TreeStore
 +-- T0 -> TreeState {
 |            root: original source,
 |            edits: [],
 |            changed: [],
 |            annotations: [...]
 |        }
 `-- T1 -> created when D executes
 ```

A `TreeState` never mutates a tree. Each applied change constructs a new root and appends enough path information to translate later targets.

 Step 6: execute `I` in `T0`

 `I` has no dependency. Its insertion position is mapped into the current T0 root, and the statement is inserted with a zero-width child range:

 ```text
 splice(parent = BlockExpr, deleted = statement_index..statement_index)

 T0 before I                         T0 after I

 fn demo() {                        fn demo() {
     let result = x + y;                let result = x + y;
 }                                      let traced = true;
                                    }

 PathEdit recorded for T0:
     Splice {
         parent: path(BlockExpr),
         deleted: k..k,
         inserted: 1,
     }
 ```

 Paths to children at or after `k` move right by one. Paths before `k`, such as the path to `R`'s target, remain valid.

 Step 7: execute `D` in generated tree `T1`

 The mapped target belongs to T1, so `TreeStore` creates a state for T1 and replaces its `y`:

 ```text
 T1 before D                         T1 after D

 check(x + y)                        check(x + z)
           ^                                   ^
           target                              replacement

 TreeStore
 +-- T0 -> root containing `let traced = true;`
 `-- T1 -> root containing `check(x + z)`
 ```

 Step 8: execute `R` using the updated payload

 `R` still stores the original T1 `check(x + y)` as its payload source. `prepare_element` finds T1's state, maps that original path through T1's recorded edits, and captures the current `check(x + z)` instead:

 ```text
 R payload source: T1 path [] -> map through T1 edits -> current T1 root
                                                        `check(x + z)`
                                                               |
                                                               v
 T0 before R                         final T0

 fn demo() {                         fn demo() {
     let result = x + y;                 let result = check(x + z);
     let traced = true;                  let traced = true;
 }                                   }
 ```

Step 9: resolve the public result

 ```text
 final T0 TreeState
     |
     +-- resolve changed paths against final root
     +-- resolve and group annotation paths against final root
     `-- discard states for detached trees such as T1
     |
     v
 SyntaxEdit
 +-- old_root: original T0
 +-- new_root: final T0
 +-- changed_elements:
 |      +-- `check(x + z)` from independent change R
 |      `-- `let traced = true;` from independent change I
 `-- annotations: mapped elements that all belong to final T0
 ```

`D` is not reported separately in `changed_elements`: it is a dependent modification already contained in `R`'s reported payload.

 What `TreeStore` actually stores

 `TreeStore` is a map from a tree's baseline root to the state we have built for it:

 ```text
 TreeStore.states
 |
 +-- key: original T0 root
 |   `-- TreeState
 |       +-- root: latest immutable T0 root
 |       +-- edits: path translations made in T0
 |       +-- changed: paths to report from final T0
 |       `-- annotations: paths currently alive in T0
 |
 `-- key: original T1 root
     `-- TreeState
         +-- root: latest immutable T1 root (`check(x + z)`)
         +-- edits: path translations made in T1
         +-- changed: []
         `-- annotations: paths currently alive in T1
 ```

 The key stays the same because recorded elements still belong to their baseline tree. `TreeState::root` changes after every immutable splice. That stable-key/current-value split lets an old payload such as `T1::check(x + y)` find the newer `T1::check(x + z)`.

States are created only when they are useful: for a tree containing a planned target, or for a tree containing a tracked annotation. Merely using an element as a payload does not automatically create a state. If its tree has never been edited or annotated, `TreeStore::prepare_element` can use the original immutable element directly.

 How the trees come back together 
 
 A generated tree moves into another tree only when a change uses one of its elements as a payload.

 One other small detail: states do not own separate lists of changes. `EditPlan` is still one ordered list. Each `PlannedChange` simply says which tree it should run against:

 ```text
 EditPlan
 |
 +-- I  destination = T0
 +-- D  destination = T1
 `-- R  destination = T0, payload = original element from T1
 ```

 `TreeStore::execute` walks the list and sends each change to the right state. Just before applying a change, it gets the payload ready:

 ```text
 PlannedChange
 +-- destination tree
 +-- target in that tree
 `-- zero or more payload elements
                    |
                    v
          TreeStore::prepare_element
                    |
                    v
 PreparedElement
 +-- latest syntax for that payload
 `-- annotations below it, expressed relative to the payload
                    |
                    v
       destination TreeState::apply
 ```

 Here is that handoff for `R` from the walkthrough:

 ```text
 R's recorded payload
 original T1 element: `check(x + y)`
 original path in T1: []
          |
          | 1. use payload.tree_top() to find the T1 state
          v
 T1 TreeState
 root:  `check(x + z)`
 edits: [replace the old `y` slot]
          |
          | 2. replay edits over the original payload path
          | 3. resolve the translated path in T1's current root
          v
 PreparedElement
 syntax: `check(x + z)`
 annotations: paths relative to `check(x + z)`
          |
          | 4. splice the prepared syntax into R's destination, T0
          | 5. rebase relative annotation paths into their new T0 location
          v
 new T0 root
 fn demo() {
     let result = check(x + z);
     let traced = true;
 }
 ```

The empty path just means that the payload is the root of `T1`. If the payload were nested, its path might be `[2, 1, 0]` instead. We replay `T1`'s `PathEdit`s to shift those indices around earlier inserts and replacements, then use the updated path to grab the element from the current root.

The important bit is that only the prepared subtree crosses from `T1` to `T0`. The bookkeeping does not tag along: we do not copy `T1.edits`, `T1.changed`, or its absolute annotation paths into `T0`.

 ```text
 T1 TreeState                                  T0 TreeState

 root: check(x + z)                            root: fn demo() { ... }
 edits: [D]                                    edits: [I]
       |                                              |
       | prepare current payload                      | apply R
       `----------> PreparedElement ------------------'
                    syntax + relative annotations
                               |
                               v
 T1 stays separate                         T0 gets one new immutable root
 ```

As far as `T0` is concerned, `R` is just one splice inserting the already updated `check(x + z)`. `T0` records that splice in its own edit ledger. The smaller edit that changed `y` to `z` stays in `T1` because it was only needed to prepare the payload.

This is also why `D` is not reported as a separate changed element. `D` updates the temporary payload, and `R` puts that finished payload into `T0`. The public result reports `check(x + z)` once; it does not report the wrapper and the inner `z` as two separate edits.

So by the time execution ends, every generated edit we care about has already made its way into `T0`. That leaves `TreeStore::finish` with a pretty small job:

 ```text
 TreeStore::finish(original T0)
          |
          +-- remove T0's state from the map
          +-- take T0.root as SyntaxEdit::new_root
          +-- resolve T0.changed against that root
          +-- resolve and group T0.annotations against that root
          `-- drop the remaining temporary states
 ```
 The useful generated subtrees are already inside `T0`; `finish` just packages that final state as the public `SyntaxEdit`.
 
 Disclaimer: I structured the ascii diagram first, then used an LLM to generate better ascii (should've just gone with Excalidraw 😅). It still took me hours to get the right ascii diag😅.

**Comment by ChayimFriedman2** [comment]
I only looked at your comment (well-explained!), not at the code yet, but from what I understand:

 - This allows us to change nodes that are descendants of already-replaced nodes, right? This is something I've wanted for a long time, I've said that I think this is the only way to proceed with lifetime'd nodes.
 - I have no problems with not doing it now and doing it as a next step, but once we want to remove mutable editing completely (instead rebuild the new tree from scratch), do I understand correctly that we only need to change the "postorder execution plan" step, and basically reverse it, doing it parent-to-child and left-to-right?

File: crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs [hunk 1]
```diff
@@ -141,7 +141,7 @@ pub(crate) fn generate_blanket_trait_impl(
             if let Some(cap) = ctx.config.snippet_cap
                 && let Some(self_ty) = impl_.self_ty()
             {
-                builder.add_tabstop_before(cap, self_ty);
```
ChayimFriedman2, line 144: We should remove the snippet methods from `SourceChangeBuilder`, right?

crates/syntax/src/syntax_editor/mapping.rs
```diff
@@ -127,55 +127,23 @@ impl SyntaxMapping {
...
+            let node = match &current {
+                SyntaxElement::Node(node) => node.clone(),
+                SyntaxElement::Token(token) => token.parent().unwrap(),
+            };
+            let Some(input_ancestor) =
+                node.ancestors().find(|ancestor| self.upmap_node_single(ancestor).is_some())
+            else {
+                return current;
+            };
+            let output_ancestor = self.upmap_node_single(&input_ancestor).unwrap();
```
ChayimFriedman2:
```suggestion
            let Some((input_ancestor, output_ancestor)) =
                node.ancestors().find(|ancestor| self.upmap_node_single(ancestor.clone()).map(|output| (ancestor, output)))
            else {
                return current;
            };
```

crates/syntax/src/syntax_editor/edit_algo.rs
```diff
@@ -1,415 +1,697 @@
 //! Implementation of applying changes to a syntax tree.
```
ChayimFriedman2, line 1: As a more high-level note, this does *a lot* of allocations and they're only used during the editing process. It makes sense to use an arena (e.g. `bumpalo`) for this. Not necessarily for this PR though.

crates/syntax/src/syntax_editor/edit_algo.rs
```diff
@@ -1,415 +1,697 @@
...
     }
 }
 
+/// Plans and executes all changes recorded by a SyntaxEditor.
+pub(super) fn apply_edits(editor: SyntaxEditor) -> SyntaxEdit {
+    let SyntaxEditor { root, changes, annotations, make } = editor;
+    let mappings = make.take();
+    let mut node_depths = FxHashMap::<SyntaxNode, usize>::default();
+    let mut node_depth = |node: SyntaxNode| {
+        *node_depths.entry(node).or_insert_with_key(|node| node.ancestors().count())
```
ChayimFriedman2, line 673: Should we instead call this recursively? This will populate the cache for the ancestors as well, meaning more memory usage but perhaps faster access.

crates/syntax/src/syntax_editor/edit_algo.rs
```diff
@@ -1,415 +1,697 @@
...
+    /// rewritten from their input trees into ancestor replacement trees. Finally,
+    /// discarded changes are removed and the dependency forest is traversed in
+    /// postorder.
+    /// Independent roots and sibling changes are prioritized right to left.
+    fn build(
+        mut changes: Vec<Change>,
+        mappings: &SyntaxMapping,
+        mut node_depth: impl FnMut(SyntaxNode) -> usize,
+    ) -> Result<Self, InvalidEditPlan> {
+        changes.sort_by(|left, right| {
```
ChayimFriedman2, line 76:
```suggestion
        changes.sort_unstable_by(|left, right| {
```

crates/syntax/src/syntax_editor/edit_algo.rs
```diff
@@ -1,415 +1,697 @@
...
+                continue;
             }
-            Change::ReplaceAll(range, _) => {
-                changed_ancestors.push_back(ChangedAncestor::multiple(range, change_index))
+
+            let parent = change.target_parent();
+            let key = (parent.tree_top(), node_depth(parent));
+            if previous
+                .insert(key, change.target_range())
+                .is_some_and(|range| range.end() > change.target_range().start())
```
ChayimFriedman2, line 220: Perhaps add a comment that because the array is sorted we don't need to check intersection and it's also fine to throw the previous node from the map?

crates/syntax/src/syntax_editor/edit_algo.rs
```diff
@@ -1,415 +1,697 @@
...
+            {
+                regions.truncate(region_index + 1);
+                match regions[region_index].nested_changes {
+                    NestedChanges::Remap => {
+                        entries[index].parent = Some(regions[region_index].change_index);
+                    }
+                    NestedChanges::Discard => entries[index].discarded = true,
+                }
+            } else {
+                regions.clear();
```
ChayimFriedman2, line 114: I understand that we clear here and truncate before so that the text ranges won't be messed up, but won't this be problematic for future changes?

crates/syntax/src/syntax_editor/edit_algo.rs
```diff
@@ -1,415 +1,697 @@
...
-        // Check if this change is dependent on another change (i.e. it's contained within another range)
-        if let Some(index) = changed_ancestors
+        let mut children = vec![Vec::new(); changes.len()];
+        for (child, entry) in entries.iter().enumerate() {
+            if let Some(parent) = entry.parent {
+                children[parent].push(child);
+            }
+        }
+        for siblings in &mut children {
+            siblings.sort_by(|&left, &right| {
```
ChayimFriedman2, line 139:
```suggestion
            siblings.sort_unstable_by(|&left, &right| {
```

crates/syntax/src/syntax_editor/edit_algo.rs
```diff
@@ -1,415 +1,697 @@
...
-            .rposition(|ancestor| ancestor.affected_range().contains_range(change.target_range()))
-        {
-            // Pop off any ancestors that aren't applicable
-            changed_ancestors.drain((index + 1)..);
+            .enumerate()
+            .filter_map(|(index, entry)| {
+                (!entry.discarded && entry.parent.is_none()).then_some(index)
+            })
+            .collect::<Vec<_>>();
+        roots.sort_by(|&left, &right| {
```
ChayimFriedman2, line 151:
```suggestion
        roots.sort_unstable_by(|&left, &right| {
```

ChayimFriedman2 (approved):
Awesome work!!

Just some nits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22808/857dec62c43dd8df13a8f628ef5b9d59bb28b0c0..06fa6f6565389ae50675816156603bdfd9d42994)*

bit-aloo:
> I only looked at your comment (well-explained!), not at the code yet, but from what I understand:
> 
> * This allows us to change nodes that are descendants of already-replaced nodes, right? This is something I've wanted for a long time, I've said that I think this is the only way to proceed with lifetime'd nodes.
> * I have no problems with not doing it now and doing it as a next step, but once we want to remove mutable editing completely (instead rebuild the new tree from scratch), do I understand correctly that we only need to change the "postorder execution plan" step, and basically reverse it, doing it parent-to-child and left-to-right?

The current approach does allow us to change nodes that are descendants of already replaced nodes, but only if it is mapped correctly to editor tree, and not the disjoint tree. For that (nested editor case), we need to track such individual trees (calling them secondary/tertiary tree), and then at the end of editing of edit cycle merging them in the editor tree (in specific order tertiary trees -> secondary trees -> primary tree). We should be able to extend current implementation, but would require a corresponding change to mapping to denote where the change belong to the editor tree or secondary tree or .... There might be some small tweaks to be made to current implementation but apart from that it should be doable. I will open a follow up for this, and migrate a couple of such cases we have in assist.

crates/ide-assists/src/handlers/generate_blanket_trait_impl.rs
bit-aloo: I can do that in a follow-up. We still have a few snippet methods that are being used, so we can't remove them all yet, but we can definitely remove some of them.

crates/syntax/src/syntax_editor/edit_algo.rs
bit-aloo: Hmm... I didn't really think much about allocations or performance. I was mostly trying to make the design more intuitive. Maybe I should benchmark it and see where we can improve things.

crates/syntax/src/syntax_editor/edit_algo.rs
bit-aloo: I will run some benchmarks and see if we can improve things with this strategy here.

crates/syntax/src/syntax_editor/edit_algo.rs
bit-aloo: Unstable sort are kinda creating problems, as for equal ordering, it doesn't respect the initial order of elements, making them kinda randomly organised, I was seeing a failure case for this, where elements with similar priority are ordered in different order from initial ordering, causing issues.

crates/syntax/src/syntax_editor/edit_algo.rs
ChayimFriedman2: This isn't *really* perf-sensitive (it'll only done for invoked assists etc.), but using an arena is an easy and major win. You don't have to do that in this PR, as I said.

crates/syntax/src/syntax_editor/edit_algo.rs
bit-aloo: Not really. regions here is just the stack of currently active ordered rangess that the current change falls inside. If a change comes along that isn't contained in any region we know about, we clear it and start fresh. Or I didn't get your question, and considering all the changes are already either disjoint or enclosing, this should not cause any problem in grouping the nesting, I guess.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22808/857dec62c43dd8df13a8f628ef5b9d59bb28b0c0..e1ab77119a8593efcc842da7590507b95d54c2f1)*

PR #22813, opened 2026-07-15
feat: Add handler for E0572

Yashb404:
rust-lang/rust-analyzer#22140 

This pull request addresses two FIXME comments in rust-analyzer/crates/hir-ty/src/infer/expr.rs.

Previously, using a return or become statement outside of a function block returned an generic error type 

This PR:

- adds ReturnOutsideFunction to InferenceDiagnostic with a boolean flag to differentiate between return and become.
- changes the signature of "infer_expr_become" to accept tgt_expr so the diagnostic can point to the entire statement.
- exposes the diagnostic through hir.
- adds an IDE handler (E0572) and tests that verify cases for keywords.

AI Disclosure : I only used AI to find the fixme and navigate the codebase , for the code I followed the principles of my previous merged PR and similiar error diagnostics such as "break-outside-of-loop" and codebase's existing patterns.

### DIFF
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -524,6 +524,18 @@ pub enum InferenceDiagnostic {
         #[type_visitable(ignore)]
         expr: ExprId,
     },
+    ReturnOutsideFunction {
+        #[type_visitable(ignore)]
+        expr: ExprId,
+        #[type_visitable(ignore)]
+        kind: ReturnKind,
+    },
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum ReturnKind {
+    ReturnExpr,
+    BecomeExpr,
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -43,7 +43,7 @@ use crate::{
 };
 
 use super::{
-    BreakableContext, Diverges, Expectation, InferenceContext, InferenceDiagnostic,
+    BreakableContext, Diverges, Expectation, InferenceContext, InferenceDiagnostic, ReturnKind,
     cast::CastCheck, find_breakable,
 };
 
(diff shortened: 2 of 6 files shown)


crates/hir-ty/src/infer.rs
```diff
@@ -524,6 +524,12 @@ pub enum InferenceDiagnostic {
         #[type_visitable(ignore)]
         expr: ExprId,
     },
+    ReturnOutsideFunction {
+        #[type_visitable(ignore)]
+        expr: ExprId,
+        #[type_visitable(ignore)]
+        is_return: bool,
```
ChayimFriedman2: If it's a return, surely it's `return`, no? A much better way will be to have an `enum ReturnKind { ReturnExpr, BecomeExpr }` then `kind: ReturnKind`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22813/b0fec0dbee399b2768de3d0f2b15be64d74424ef..91275326c5a63f9eb04b3ea1acd5e1c0e4d323ab)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/infer.rs
Yashb404: I implemented the requested changes . My original inspiration was from break-outside-of-loop as it utilized the same convention for differentiating between break and continue . This is much cleaner and less confusing . Thank you for your feedback.

ChayimFriedman2 (approved):
Please squash.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22813/b0fec0dbee399b2768de3d0f2b15be64d74424ef..f74d59e4ba05e9bf713968cf4403e80c28ee7949)*

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22813/76d092ea9d27a66c53aedf630c07e3dae42db1c1..d777f0c379424858e547bae1e4d97d13ae1402fa)*

PR #22814, opened 2026-07-15
internal: migrate `HirDatabase::borrowck`

ada4a:

Split out from the PR removing `HirDatabase` completely, as that was getting too big.

### DIFF
--- a/crates/hir-ty/src/mir/borrowck.rs
+++ b/crates/hir-ty/src/mir/borrowck.rs
@@ -133,7 +133,7 @@ fn all_mir_bodies<'db>(
     }
 }
 
-#[salsa_macros::tracked(returns(ref), lru = 2024)]
+#[salsa_macros::tracked(returns(as_deref), lru = 2024)]
 pub fn borrowck_query(
     db: &dyn HirDatabase,
     def: InferBodyId,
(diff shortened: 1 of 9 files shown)


crates/hir-ty/src/mir/borrowck.rs
```diff
@@ -133,59 +133,66 @@ fn all_mir_bodies<'db>(
...
-                        match (&*parent_mol, &child.mutability_of_locals[*child_local]) {
-                            (MutabilityReason::Mut { .. }, _) => {}
-                            (_, MutabilityReason::Mut { .. }) => {
-                                // FIXME: Fix the child spans.
-                                *parent_mol = MutabilityReason::Mut { spans: Vec::new() }
+impl InferBodyId {
+    pub fn borrowck(self, db: &dyn HirDatabase) -> Result<&[BorrowckResult], MirLowerError> {
+        return borrowck_query(db, self).map_err(|e| e.clone());
+
+        #[salsa::tracked(returns(as_deref), lru = 2024)]
```
ChayimFriedman2, line 140: Any reason you didn't mark the impl `#[salsa::tracked]`, then you won't need the inner function?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22814/b0fec0dbee399b2768de3d0f2b15be64d74424ef..9a58ff2dbe6588282441d725cc886f42489efc6a)*

ada4a: Yes: even with `returns(as_deref)`, the inner function returns `Result<&[BorrowckResult], &MirLowerError>`, but what we want to expose to callers is `Result<&[BorrowckResult], MirLowerError>` (this is what `HirDatabase::borrowck` used to return), hence the `.map_err(|err| err.clone()` in the outer function.

Worth a comment perhaps?

ChayimFriedman2 (approved):

PR #22816, opened 2026-07-15
better error message

ashishk15678:
Related https://github.com/rust-lang/rust-analyzer/issues/13256

Just changed the error message from 
`Failed to create perf counter`

to 
`Failed to create perf counter , you can suppress this warning by setting RA_DISABLE_PERF=1 {err}`

UX related small change.


lnicola:
This feels a bit pointless. I think "Failed to create perf counter, executed instructions metrics will not be available", but TBH, this command is used for benchmarking and debugging, so a "normal user" will have no reason to use it anyway,

ashishk15678:
No worries @lnicola , I am new to the repo and was trying out easy issues . Close it or do as you prefer . Thanks .

PR #22831, opened 2026-07-16
internal: Migrate `HirDatabase` away from `#[query_group]`; remove the latter

### LINKED ISSUE #22657: Tracking issue for migrating away from `#[query_group]`
@Veykril [said](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Do.20you.20need.20anything.20cleaned.20up.3F/near/590703741):
> One thing we still need to "clean up" are the left over tasks from the salsa migration. That is we currently have this `query_group` macro that mimics the old `salsa` API which we used to make the migration easier. We still need to get rid of all the invocations of that macro and replace them by the newer salsa APIs

### Steps/History
- [x] remove features from the `#[query_group]` macro:
  - https://github.com/rust-lang/rust-analyzer/pull/22328
  - https://github.com/rust-lang/rust-analyzer/pull/22639
  - https://github.com/rust-lang/rust-analyzer/pull/22700
  - https://github.com/rust-lang/rust-analyzer/pull/22702
  - https://github.com/rust-lang/rust-analyzer/pull/22746
  - https://github.com/rust-lang/rust-analyzer/pull/22780
- [x] remove the remaining query groups:
  - [x] `RootQueryDb` (https://github.com/rust-lang/rust-analyzer/pull/21944)
  - [x] `LineIndexDatabase` (https://github.com/rust-lang/rust-analyzer/pull/22063)
  - [x] `ExpandDatabase`
    - https://github.com/rust-lang/rust-analyzer/pull/22658
    - https://github.com/rust-lang/rust-analyzer/pull/22689
    - https://github.com/rust-lang/rust-analyzer/pull/22699
    - https://github.com/rust-lang/rust-analyzer/pull/22719
    - https://github.com/rust-lang/rust-analyzer/pull/22729
    - (https://github.com/rust-lang/rust-analyzer/pull/22733)
    - https://github.com/rust-lang/rust-analyzer/pull/22739
  - [x] `DefDatabase` (https://github.com/rust-lang/rust-analyzer/pull/22745)
  - [x] `HirDatabase`
    - https://github.com/rust-lang/rust-analyzer/pull/22814
    - https://github.com/rust-lang/rust-analyzer/pull/22831

ada4a:
Resolves https://github.com/rust-lang/rust-analyzer/issues/22657 :tada:

See [#t-compiler/rust-analyzer > Migrating &#96;HirDatabase&#96;](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Migrating.20.60HirDatabase.60/with/610845156) for discussion, and individual commits for details.


ChayimFriedman2 (dismissed):
I don't like the many places that need to explicitly coerce to `&dyn SourceDatabase`. I think a better solution is to add `fn as_dyn(&self) -> &dyn SourceDatabase` to `RootDatabase` and `TestDB`. Maybe even implement `Deref`, although that's probably a bad idea.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22831/7a5ecc88dbdfa13bac99fa4e0f38adcfdca1bb0a..5b54e097eb65576dc8c9f6330bf75103c3466326)*

ada4a:
Done. Though I worry that future contributors will have a hard time discovering `as_dyn`, and may end up doing the cast explicitly. Honestly even that I'm not sure about -- one'd need to come up with casting to `&dyn SourceDatabase` just to let the compiler see the `HirDatabase` impl on that (`db as &dyn SourceDatabase as &dyn HirDatabase` doesn't even work)

crates/hir-ty/src/db.rs
```diff
@@ -36,248 +36,290 @@ use crate::{
...
+
     // region:mir
 
     // FIXME: Collapse `mir_body_for_closure` into `mir_body`
     // and `monomorphized_mir_body_for_closure` into `monomorphized_mir_body`
-    #[salsa::transparent]
     fn mir_body(&self, def: InferBodyId) -> Result<&MirBody, MirLowerError> {
-        crate::mir::mir_body_query(self, def).map_err(|err| err.clone())
+        let db = self.as_dyn_source_database();
+        crate::mir::mir_body_query(db, def).map_err(|err| err.clone())
```
Veykril: Given we are keeping the database trait, why do we even need to swap all the trait objects from `HirDatabase` to `SourceDatabase`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22831/1649637a9c68a8f371d2e1326bbd15d36a8a74a8..c110819016f0bfd17761365191d2ba2cbaae6a12)*

ada4a: Wouldn't doing otherwise lead to circular reasoning? `mir_body_query` would require the caller to be `&dyn HirDatabase`, but we'd want to call it inside `HirDatabase::mir_body`, which.. doesn't yet know that the caller implements `HirDatabase`, as we're in the process of implementing it.

Besides, it looks like I can't annotate functions taking `db: &dyn HirDatabase` with `#[salsa::tracked]`, probably because the it's no longer a salsa db:
```text
error[E0576]: cannot find method or associated constant `zalsa_register_downcaster` in trait `HirDatabase`
    --> crates/hir-ty/src/mir/lower.rs:2285:1
     |
2285 | #[salsa_macros::tracked(returns(as_ref), cycle_result = mir_body_cycle_result)]
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `HirDatabase`
     |
     = note: this error originates in the macro `salsa::plumbing::setup_tracked_fn` which comes from the expansion of the attribute macro `salsa_macros::tracked` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Veykril: That salsa error is because you need to annotate the trait with `#[salsa::database]` (or some attribute name).

I don't see a circular reasoning here. Sure, things can end up recursive in calls (with different function arguments etc) but thats the case already today. In general in Rust inside of trait definitions the `Self` generic argument is assumed to implement the trait itself so you can use that fact just fine.

ada4a: If I remove the `let db = self.as_dyn_source_database();` line and try to pass `self` directly, I get back to this error:

```text
error[E0277]: the size for values of type `Self` cannot be known at compilation time
  --> crates/hir-ty/src/db.rs:48:36
   |
48 |         crate::mir::mir_body_query(self, def).map_err(|err| err.clone())
   |                                    ^^^^ doesn't have a size known at compile-time
   |
   = note: required for the cast from `&Self` to `&(dyn HirDatabase + 'static)`
help: consider further restricting `Self`
   |
47 |     fn mir_body(&self, def: InferBodyId) -> Result<&MirBody, MirLowerError> where Self: std::marker::Sized {
   | 
```

Do you know how I'd solve that?...

Veykril: I think we need a `'static` super trait bound? That is `trait HirDatabase: SourceDatabase + 'static`

ada4a: But that would require the implementer to be `dyn SourceDatabase + 'static`, which doesn't sound right? Well, it didn't work anyway...

ada4a: It looks like I can fix this by, once again, moving the method definitions from the trait definition to the impl. But that's ugly...

crates/hir-ty/src/db.rs
```diff
@@ -37,260 +36,296 @@ use crate::{
...
+        crate::variance::variances_of(db, def)
+    }
+}
+
+#[salsa::db]
+impl<T: SourceDatabase> HirDatabase for T {
+    fn as_dyn(&self) -> &dyn HirDatabase {
+        self
+    }
 }
```
Veykril, line 342: Neat that's even better than what i proposed.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22831/081ddc9c1d9cba0b67f72ec3c4b45186d039c28c..2f54dec192fbebc08eeafd4083358668cf934d7f)*

Veykril (approved):

Veykril:
Ah we need another rebase given the PR I just merged it seems

crates/hir-ty/src/db.rs
ada4a: For posterity: we solved this by changing `as_dyn_source_database` to `fn as_dyn(&self) -> &dyn HirDatabase`. Apparently this method _can_ be defined in the `impl HirDatabase`, provided it's for a Sized type

PR #22843, opened 2026-07-17
fix: panic on macro-defined structs with unknown fields

Wilfred:
The no_such_field assist offers to add a field to a struct definition when the user writes a struct literal `Foo { no_such_field_yet: 1 }`.

However, if the struct is defined in a macro, we would look for the struct definition using the offset computed from the expansion. This is wrong, that expansion offset doesn't match the source file.

Fix the assist to handle macro expansion correctly.

(See individual commits for AI disclosures.)

### DIFF
--- a/crates/ide-diagnostics/src/handlers/no_such_field.rs
+++ b/crates/ide-diagnostics/src/handlers/no_such_field.rs
@@ -1,11 +1,11 @@
 use either::Either;
-use hir::{HasSource, HirDisplay, InMacroFile, Semantics, VariantId};
+use hir::{HasSource, HirDisplay, InFile, Semantics, VariantId};
 use ide_db::text_edit::TextEdit;
 use ide_db::{
     EditionedFileId, RootDatabase, helpers::is_editable_crate, source_change::SourceChange,
 };
 use syntax::{
-    AstNode, TextSize,
+    AstNode,
     ast::{self, edit::IndentLevel, make},
 };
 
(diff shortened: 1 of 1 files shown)


A4-Tacks (commented):
I think it might be a good idea to support editing within macro definitions, as long as the crate is editable

At least support as-is expansion (already supported by `private_field.rs`), like:

```
macro_rules! identity { ($($t:tt)*) => ($($t)*); }
identity! {
    struct Foo {
        pub a: i32,
    }
}
```


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22843/1649637a9c68a8f371d2e1326bbd15d36a8a74a8..37122c8485899820d44645bf40266581c1416f60)*

Wilfred:
I did wonder about that, but I worry it'd be hard to get right when macros are less structured that normal source code (just token streams IIRC). I see `private_field.rs` is pretty short and readable though.

I have seen this panic hit real users (another panic triggered by computing assists, not requesting a specific assist is applied), but only a few folks.

Wilfred:
Thanks for the pointer to private_field.rs, I've updated the code to handle macros properly now.

> as long as the crate is editable

Oh, it looks like we weren't checking the editable state of the crate before. I've added a commit fixing that, with a test.

crates/ide-diagnostics/src/handlers/no_such_field.rs
```diff
@@ -97,13 +102,19 @@ fn missing_record_expr_field_fixes(
...
         if let Some(last_field) = record_fields.fields().last() {
             let indent = IndentLevel::from_node(last_field.syntax());
             let offset = last_field.syntax().text_range().end();
             let needs_comma = !last_field.to_string().ends_with(',');
             (indent, offset, String::new(), needs_comma)
         } else {
+            // We don't have enough whitespace information in a macro-defined empty struct
+            // to compute the correct indent.
+            if def_file_id.is_macro() {
+                return None;
```
A4-Tacks: Don't give up fixes just for indentation


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22843/1174734d9c6453d13d2b5e3d578512c579ca37f1..0f0c0fa1b6f8dc11517b1e70683fd8a333a939ef)*

crates/ide-diagnostics/src/handlers/no_such_field.rs
```diff
@@ -449,6 +474,80 @@ fn main() {
...
+fn f() { let _ = S { a: 1, b: false }; }
+"#,
+        );
+    }
+
+    #[test]
+    fn test_add_field_macro_defined_struct_large_offset() {
+        // Regression test: we should handle macro definitions whose offset is larger than
+        // the max position in main.rs.
+        check_fix(
```
A4-Tacks, line 485: After careful consideration and testing, it seems that supporting this case is quite hacky and not as useful
I'm sorry for wasting your time. Not supporting this case should make the implementation simpler, like:
```
let (indent /*u8*/, after /*SyntaxToken*/, postfix, needs_comma) = ...;
let hir::FileRange { file_id, range } = InFile::new(def_file_id, after.clone()).original_file_range_opt(sema.db)?;
let origin = sema.parse(file_id).syntax().covering_element(range);
let indent = IndentLevel::from_element(&origin) + indent;
```


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22843/1174734d9c6453d13d2b5e3d578512c579ca37f1..0f0c0fa1b6f8dc11517b1e70683fd8a333a939ef)*

Wilfred: No worries, it was a useful exercise and didn't take too long :)

Are you suggesting that I go back to the "don't offer any assist if there's a macro" implementation? You're commenting on the specific test case that replicates the original crash I was trying to fix.

A4-Tacks: I feel uneasy about the TextSize calculation here

I have a clean solution, can I push it?

Wilfred: Sure, feel free.

Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22843/1174734d9c6453d13d2b5e3d578512c579ca37f1..062187ce4751f5f17ff46233fd31c0eca981c474)*

PR #22850, opened 2026-07-18
temp : move maybe_collect_expr into different arms

ashishk15678:
By only restructuring the match block in maybe_collect_expr , the peak memory drops by 144mb ( from 2780 to 2639 )
Ran cargo run --release --bin rust-analyzer -- analysis-stats . on rust-analyzer's repo .
But the time also increased by ~400ms . (I think this is a good tradeoff)

One more approach that I think would be quite good . https://github.com/rust-lang/rust-analyzer/issues/5317#issuecomment-5009945497

Related https://github.com/rust-lang/rust-analyzer/issues/5317

### DIFF
--- a/crates/hir-def/src/expr_store/lower.rs
+++ b/crates/hir-def/src/expr_store/lower.rs
@@ -1337,101 +1337,7 @@ impl<'db> ExprCollector<'db> {
                 let expr = self.collect_expr_opt(e.expr());
                 self.alloc_expr(Expr::Let { pat, expr }, syntax_ptr)
             }
-            ast::Expr::BlockExpr(e) => match e.modifier() {
-                Some(ast::BlockModifier::Try { try_token: _, bikeshed_token: _, result_type }) => {
-                    self.desugar_try_block(e, result_type)
-                }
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
The difference cannot come from stack usage. The whole stack is only 8mb.

The only reason I can think of that can maybe bring this saving is change in creation/dropping times for rowan red trees.

However, I absolutely does not agree that saving 144mb in peak (not enduring!) memory usage is worth this. Not because the speed impact like because the code maintainability impact.

The good news are that we're rewriting rowan anyway, and after we finish that, rowan won't have red trees so they will consume zero memory. Peak memory usage as well as enduring memory usage are expected to decrease significantly.

ashishk15678:
No worries @ChayimFriedman2, can you also please point me out to some issue whcih I can work on, an easy one. I have tried solving 2 so far and both got closed. Something good to work with please. Thanks.

ChayimFriedman2:
Look in [#t-compiler/rust-analyzer > Contribution](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Contribution/with/611455982).

Issue #22140, opened 2026-04-22
Address FIXMEs for adding diagnostics


ChayimFriedman2:
We have quite a lot of FIXMEs, mostly in hir-ty, for adding new diagnostics for things (and I'm constantly adding more). The shape of them is like `FIXME: Emit an error ...` (my own FIXMEs usually use these words, others might use similar). Most are pretty easy to do. The steps to resolve one (in inference) are:

 1. Add a variant to `InferenceDiagnostic` with the necessary information.
 2. Call `self.push_diagnostic(diag)` on the place.
 3. In `hir/src/diagnostics.rs`, add a new struct for the diagnostic and convert the enum variant to it in the functions of that module. Note: this struct must not use `hir-ty` types, only `hir` types.
 4. Create a new handler in `ide-diagnostics/src/handlers/` for the diagnostic (follow existing handlers). A simple handler can take a few lines. Also add some tests.
 5. Call the handler from `ide-diagnostics/src/lib.rs`.

That's it! You can also of course customize the error message to be clearer/clever, and/or add a fix, but that is good enough.

ada4a:
I'll start with this one: https://github.com/rust-lang/rust-analyzer/blob/f45c1e29321abf353869ebf753a3b18536dbeb78/crates/hir-ty/src/lower/path.rs#L302-L306

ChayimFriedman2:
@ada4a That one is an error in types as well, so you need to add the diagnostic to `PathLoweringDiagnostic` instead. But the process is similar.

ada4a:
https://github.com/rust-lang/rust-analyzer/blob/7bf346ebb7ab5e64782dac47426ebd14e1cef29b/crates/hir-ty/src/infer/expr.rs#L251-L258

ada4a:
https://github.com/rust-lang/rust-analyzer/blob/c2a4de0fe8e7808f07b2db1ecc87238c16bf09ee/crates/hir-def/src/expr_store/lower.rs#L2478-L2502

ada4a:
https://github.com/rust-lang/rust-analyzer/blob/210340894905b8f6e94aaba2084b03cbbbde4cd3/crates/hir-ty/src/method_resolution/confirm.rs#L574-L581

MavenRain:
[...]

PR #22851, opened 2026-07-18
fix: add diagnostic for struct patterns which don't specify sub-patterns for its fields

TheDoctor314:
Relates to #22140 .

Emit a diagnostic [E0027](https://doc.rust-lang.org/error_codes/E0027.html) when patterns for a struct don't specify sub-patterns for each of its fields.

### Notes
this PR also includes a modification of the `extract_annotations()` to support more than one continuation for a multi-line annotation. This was required for the test which checks diagnostic for more than one missing field. ref: `crates/ide-diagnostics/src/handlers/struct_pat_must_have_all_fields_or_rest_pat.rs`

### DIFF
--- a/crates/ide-diagnostics/src/handlers/missing_fields.rs
+++ b/crates/ide-diagnostics/src/handlers/missing_fields.rs
@@ -274,7 +274,7 @@ fn get_default_constructor(
 
 #[cfg(test)]
 mod tests {
-    use crate::tests::{check_diagnostics, check_fix, check_no_fix};
+    use crate::tests::{check_diagnostics, check_fix, check_has_fix, check_no_fix};
 
     #[test]
     fn missing_record_pat_field_diagnostic() {
@@ -829,7 +829,7 @@ fn f() {
 
     #[test]
     fn test_fill_struct_pat_fields_partial() {
-        check_fix(
+        check_has_fix(
             r#"
 struct S { a: &'static str, b: i32 }
 
--- a/crates/hir-ty/src/diagnostics/expr.rs
+++ b/crates/hir-ty/src/diagnostics/expr.rs
@@ -157,18 +157,6 @@ impl<'db> ExprValidator<'db> {
                 _ => {}
             }
         }
-
-        for (id, pat) in body.pats() {
-            if let Some((variant, missed_fields)) =
-                record_pattern_missing_fields(db, self.infer, id, pat)
-            {
-                self.diagnostics.push(BodyValidationDiagnostic::RecordMissingFields {
-                    record: Either::Right(id),
-                    variant,
-                    missed_fields,
-                });
-            }
-        }
     }
 
     fn validate_call(
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -538,6 +538,14 @@ pub enum InferenceDiagnostic {
         #[type_visitable(ignore)]
         kind: ReturnKind,
     },
+    RecordMissingFields {
+        #[type_visitable(ignore)]
+        record: ExprOrPatId,
+        #[type_visitable(ignore)]
+        variant: VariantId,
+        #[type_visitable(ignore)]
+        missed_fields: Vec<LocalFieldId>,
+    },
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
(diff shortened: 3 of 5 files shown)


ChayimFriedman2 (changes requested):
We already have a `BodyValidationDiagnostic` for this, like you can see in the test. We should not have both. Remove that diagnostic, and change the one emitted from inference to match it. You don't need to change the ide-diagnostics handler at all.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22851/cac0779549328e4bd4b808000c03307f1721f869..07040986fd4d600414aa077885b0daf2d0df2307)*

TheDoctor314:
Hi, just making sure I understand you correctly.
You want me to stop emitting `BodyValidationDiagnostic::RecordMissingFields` diag on https://github.com/rust-lang/rust-analyzer/blob/3dc827887f2ec2c5550412dba7f4192b48d84d8c/crates/hir-ty/src/diagnostics/expr.rs#L161-L171

and to use `InferenceDiagnostic::StructPatMustHaveAllFieldsOrRest` to emit `hir::MissingFields` diag instead, right?

ChayimFriedman2:
Indeed, and not only that, also make the `InferenceDiagnostic` agnostic to coming from pattern or expression (you can still leave expressions unimplemented if you want, although that's not hard to implement).

TheDoctor314:
not sure why CI is failing, clippy does not raise any errors on my machine and all test cases pass.

ChayimFriedman2:
You need to rebase then fix the code.

ChayimFriedman2 (changes requested):
Please revert the changes to test, we already have tests for this from the old implementation, and also squash.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22851/3eaae41d955ac27fca16ea1ef08874cc131ff335..bd7a4213cf0cdbbd84906b7fd19e1d6788cb4f9b)*

crates/ide-diagnostics/src/handlers/missing_fields.rs
```diff
@@ -829,7 +829,7 @@ fn f() {
 
     #[test]
     fn test_fill_struct_pat_fields_partial() {
-        check_fix(
+        check_has_fix(
```
ChayimFriedman2, line 832: Why this change? Are there multiple fixes now? Why?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22851/af1fe01c98cfb006b3e12bcca4941551f1c96273..c30aa3a1e31164d8321414e01f82b501ddbb0e2c)*

crates/test-utils/src/lib.rs
```diff
@@ -261,9 +261,15 @@ pub fn extract_annotations(text: &str) -> Vec<(TextRange, String)> {
                             .iter()
```
ChayimFriedman2, line 261: Please revert the changes to this file as you no longer use them.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22851/af1fe01c98cfb006b3e12bcca4941551f1c96273..c30aa3a1e31164d8321414e01f82b501ddbb0e2c)*

ChayimFriedman2 (changes requested):

crates/ide-diagnostics/src/handlers/missing_fields.rs
TheDoctor314: there were always multiple fixes emitted here - 'Rename unused `a` to `a: _a`' and 'Fill struct fields'. After moving 'MissingFields' diag from body validation to inference diagnostics, this order changed.
Since `check_fix` pops the last diag and asserts the source edits, this test was failing. `check_has_fix` instead worked.

context - https://github.com/TheDoctor314/rust-analyzer/blob/master/crates/hir/src/lib.rs#L2206 and https://github.com/TheDoctor314/rust-analyzer/blob/master/crates/hir/src/lib.rs#L2044

crates/test-utils/src/lib.rs
TheDoctor314: reverted

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22851/1174734d9c6453d13d2b5e3d578512c579ca37f1..bca065857546e090a0f34922e0a208a92818b124)*

PR #22854, opened 2026-07-19
feat: Record obligation chain for unimplemented trait diagnostics and show it

Veykril:
A very simplistic approach.

@ChayimFriedman2 I assume you have some thoughts on this.

### DIFF
--- a/crates/hir-ty/src/next_solver/infer/errors.rs
+++ b/crates/hir-ty/src/next_solver/infer/errors.rs
@@ -32,19 +32,10 @@ use crate::{
 pub struct FulfillmentError<'db> {
     pub obligation: PredicateObligation<'db>,
     pub code: FulfillmentErrorCode<'db>,
-    pub trait_obligation_chain: Vec<Predicate<'db>>,
+    pub parent_trait_obligations: Vec<Predicate<'db>>,
 }
 
-impl<'db> FulfillmentError<'db> {
-    pub fn new(
-        obligation: PredicateObligation<'db>,
-        code: FulfillmentErrorCode<'db>,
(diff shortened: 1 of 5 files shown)


crates/test-utils/src/lib.rs
Veykril, line 1: These are necessary to support multiple annotation continuations

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22854/cac0779549328e4bd4b808000c03307f1721f869..0ce3d413da91036f310a22316358d5c6e43de6f7)*

crates/hir-ty/src/next_solver/infer/errors.rs
```diff
@@ -15,7 +15,7 @@ use crate::{
     db::GeneralConstId,
```
ChayimFriedman2, line 15: I am not that pleased with changing this file, because it was copied from rustc and I'd prefer to keep it like that (I didn't even properly review it). OTOH this *is* a nice feature.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22854/cac0779549328e4bd4b808000c03307f1721f869..0ce3d413da91036f310a22316358d5c6e43de6f7)*

Veykril: I think the diff is small enough for that to be okay in that case, but I understand the thought. Your call since you are likely to touch this more than I am (though we will need to derive this info somehow at some point anyways if we do get the proof tree inspector I imagine)

ChayimFriedman2 (approved):

lnicola:
image

PR #22857, opened 2026-07-19
fix: Handle enum variants in next-solver `generics`

Veykril:


crates/hir-ty/src/next_solver/generics.rs
```diff
@@ -24,6 +22,10 @@ pub(crate) fn generics(interner: DbInterner<'_>, def: SolverDefId) -> Generics<'
         (_, SolverDefId::BuiltinDeriveImplId(id)) => {
             return crate::builtin_derive::generics_of(interner, id);
         }
+        (_, SolverDefId::EnumVariantId(id) | SolverDefId::Ctor(Ctor::Enum(id))) => {
```
ChayimFriedman2, line 25: BTW I was considering whether we should get rid of `Ctor`. It's true that in rustc it's a different DefId, but it might be that they only appear in different type context.

PR #22858, opened 2026-07-19
fix: Fix `typed_hole` panic

### LINKED ISSUE #22659: request handler panicked: failed to unify type owners
```
rust-analyzer 1.91.1 (ed61e7d 2025-11-07)
rustc 1.91.1 (ed61e7d7e 2025-11-07)

Name: rust-analyzer
Id: rust-lang.rust-analyzer
Description: Rust language support for Visual Studio Code
Version: 0.3.2946
Publisher: The Rust Programming Language 
```

```
#[derive(Clone, Copy)]
struct bits {
    inner: u64,
}


// change static to const, rust-analyzer fail 
const  GC_TAG: u64 = 0x8888_8888_0000_0000;
// ^^^^

fn foo() -> bits {
    let x = 10;
    x
        
    
}

fn main() {
}

```

https://github.com/user-attachments/assets/f9f8951a-d455-4abb-90be-c78a7ea017e8

Veykril:
Fixes https://github.com/rust-lang/rust-analyzer/issues/22659

Also makes the hir type unification API more resilient to mistakes in release builds

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -5477,7 +5477,10 @@ impl<'db> Type<'db> {
         db: &'db dyn HirDatabase,
         rebase_into: &Type<'db>,
     ) -> Type<'db> {
-        self.try_rebase_into(db, rebase_into).unwrap_or_else(|| self.instantiate_with_errors())
+        self.try_rebase_into(db, rebase_into).unwrap_or_else(|| Type {
+            owner: rebase_into.owner,
+            ty: self.instantiate_with_errors().ty,
+        })
     }
 
     pub fn try_rebase_into_owner(
@@ -6607,8 +6610,11 @@ impl<'db> Type<'db> {
     /// Note that we consider placeholder types to unify with everything.
     /// For example `Option<T>` and `Option<U>` unify although there is unresolved goal `T = U`.
     pub fn could_unify_with(&self, db: &'db dyn HirDatabase, other: &Type<'db>) -> bool {
-        self.owner.must_unify(other.owner);
-        let env = self.param_env(db);
+        let Some(owner) = self.owner.unify(other.owner) else {
+            never!("callers should not mix owners");
+            return false;
+        };
+        let env = Type { owner, ty: self.ty }.param_env(db);
         let interner = DbInterner::new_no_crate(db);
         let tys = hir_ty::replace_errors_with_variables(
             interner,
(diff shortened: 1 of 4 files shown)


ChayimFriedman2:
How does this relate to https://github.com/rust-lang/rust-analyzer/pull/22662?

crates/hir/src/lib.rs
```diff
@@ -5477,7 +5477,10 @@ impl<'db> Type<'db> {
         db: &'db dyn HirDatabase,
         rebase_into: &Type<'db>,
     ) -> Type<'db> {
-        self.try_rebase_into(db, rebase_into).unwrap_or_else(|| self.instantiate_with_errors())
+        self.try_rebase_into(db, rebase_into).unwrap_or_else(|| Type {
+            owner: rebase_into.owner,
+            ty: self.instantiate_with_errors().ty,
+        })
```
ChayimFriedman2, line 5483: This change makes it less flexible (although it doesn't really matter): `instantiate_with_errors()` returns a `NoParams` owner, which unifies (not just rebases) with any owner.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22858/3dc827887f2ec2c5550412dba7f4192b48d84d8c..dd6ecda2b78a273825f2e03d1a312f837c7e4481)*

crates/hir/src/lib.rs
```diff
@@ -6607,8 +6610,11 @@ impl<'db> Type<'db> {
     /// Note that we consider placeholder types to unify with everything.
     /// For example `Option<T>` and `Option<U>` unify although there is unresolved goal `T = U`.
     pub fn could_unify_with(&self, db: &'db dyn HirDatabase, other: &Type<'db>) -> bool {
-        self.owner.must_unify(other.owner);
-        let env = self.param_env(db);
+        let Some(owner) = self.owner.unify(other.owner) else {
+            never!("callers should not mix owners");
```
ChayimFriedman2, line 6614: Probably a good idea to still have `must_unify()` just for this `never!()`?

We can also have a macro for the `return false` but that's probably not a good idea.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22858/3dc827887f2ec2c5550412dba7f4192b48d84d8c..dd6ecda2b78a273825f2e03d1a312f837c7e4481)*

crates/hir/src/lib.rs
```diff
@@ -6607,8 +6610,11 @@ impl<'db> Type<'db> {
...
+            return false;
+        };
+        let env = Type { owner, ty: self.ty }.param_env(db);
```
ChayimFriedman2, line 6617: Why this change? Given that we return on unification failure (and we must, otherwise we can panic) we can leave this as `self.param_env(db)`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22858/3dc827887f2ec2c5550412dba7f4192b48d84d8c..dd6ecda2b78a273825f2e03d1a312f837c7e4481)*

Veykril:
Ah I missed your PR

ChayimFriedman2:
What's about this? I prefer my PR, it's smaller and nicer IMO.

Veykril:
I still prefer having the API be more robust to mistakes than panicking to be fair but I merged your PR either way just to have things fixed for onw

ChayimFriedman2:
I don't object to making it `never!()` instead of panicking. My objection to this PR was to the changes to term search, not the changes to hir/src/lib.rs.

PR #22861, opened 2026-07-19
Record expressions in types in ExprScope

Veykril:


Veykril:
We might port this over to the visitor infra fully

crates/hir-def/src/expr_store/scope.rs
```diff
@@ -260,6 +261,42 @@ impl ExprScopes {
...
+        let mut scope = *self.const_scope;
+        compute_expr_scopes(expr, self.store, self.scopes, &mut scope, self.const_scope);
+    }
+
+    fn on_pat(&mut self, pat: PatId) {
+        self.store.visit_pat_children(pat, &mut *self);
+    }
+
+    fn on_type(&mut self, ty: TypeRefId) {
+        self.store.visit_type_ref_children(ty, &mut *self);
```
ChayimFriedman2, line 286: This should call `compute_type_scope()`.

And yes I agree that it makes sense to only use the visitor infra.

PR #22863, opened 2026-07-19
fix: Type mismatch with ! on async function (Issue #22853)

### LINKED ISSUE #22853: Type mismatch with `!` on async function
https://github.com/rust-lang/rust-analyzer/pull/22852 fixes https://github.com/rust-lang/rust-analyzer/issues/22847, but a similar code still emit a type mismatch:
```rust
async fn test() -> ! {
    test().await;
}
```
(This code does not compile as-is, you need to `Box::pin(test())`. However for a test in hir-ty it is enough).

When the function calls itself instead of another function we still fail. I investigated this a bit, and it appears that we fail to normalize `<RPIT_opaque as IntoFuture>::Output` into `!` in time (that is, when inferring the `.await`). When calling another function, the solver knows it cannot normalize the opaque and so it only uses its bounds and normalizes the projection from them; but when inside the function, we can normalize the opaque but we don't know it yet, which seems to make the solver fail to normalize instead of normalizing via the bounds.

This code however does compile in rustc with `-Znext-solver=globally`, so it seems the problem is in our code and not in the solver (I didn't check with the same solver version as we use though).

jibin7jose:
Fixes rust-lang/rust-analyzer#22853

**Description:**
This PR fixes a bug where `is_never()` would evaluate to `false` when an `async fn` returning `!` recursively called itself and awaited the result. The trait solver would fail to evaluate `<Opaque as IntoFuture>::Output` inside the opaque type's defining scope, returning an unconstrained inference variable instead of `!`, which led to the block returning `()` and the resulting type mismatch `expected !, got ()`.

**Changes:**
- Added a fallback in `infer_await_expr` inside `crates/hir-ty/src/infer/expr.rs`. When `.await` yields an unconstrained inference variable, we now check if the `awaitee_ty` resolves to an opaque type. If so, we manually fetch its item bounds and extract the `FutureOutput` or `IntoFutureOutput` constraint.
- Added a regression test for Issue rust-lang/rust-analyzer#22853 to `crates/hir-ty/src/tests/never_type.rs`.

### DIFF
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -962,10 +962,46 @@ impl<'db> InferenceContext<'_, 'db> {
             return self.types.types.error;
         };
         self.table.register_bound(awaitee_ty, into_future, ObligationCause::new(expr));
-        self.table.try_structurally_resolve_type(
+        let mut resolved = self.table.try_structurally_resolve_type(
             expr.into(),
             Ty::new_projection(self.interner(), into_future_output.into(), [awaitee_ty]),
-        )
+        );
+
+        if resolved.is_ty_var() {
+            let awaitee_resolved = self.table.resolve_vars_if_possible(awaitee_ty);
+            if let rustc_type_ir::TyKind::Alias(rustc_type_ir::AliasTy {
+                kind: rustc_type_ir::AliasTyKind::Opaque { def_id },
+                args,
+                ..
+            }) = awaitee_resolved.kind()
+            {
+                let item_bounds = def_id.0.predicates(self.db);
+                for predicate in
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
It is entirely clear that you do not understand the bug or how to solve it. Chances are close to 100% that you used a LLM to author the patch (and maybe the PR description as well) without even looking at it, and without disclosing it. Please note that [our AI policy](https://github.com/rust-lang/rust-analyzer/blob/master/AI_POLICY.md) requires disclosure for AI usage, and completely disallows contribution you don't or can't understand as well as using LLMs for authoring GitHub comments.

I'm closing this. If you'll keep this behavior we'll reach out to moderation.

ChayimFriedman2:
Even that apology feels written by an LLM :rofl:

PR #22864, opened 2026-07-19
internal: Port `ExprScopes` over to visitor fully

Veykril:
Follow up to https://github.com/rust-lang/rust-analyzer/pull/22861

### DIFF
--- a/crates/hir-def/src/expr_store/scope.rs
+++ b/crates/hir-def/src/expr_store/scope.rs
@@ -5,9 +5,9 @@ use la_arena::{Arena, ArenaMap, Idx, IdxRange, RawIdx};
 
 use crate::{
     BlockId, DefWithBodyId, ExpressionStoreOwnerId, GenericDefId, VariantId,
-    expr_store::{Body, ExpressionStore, HygieneId, StoreVisitor, body::Param},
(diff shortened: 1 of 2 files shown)


crates/hir-def/src/expr_store/scope.rs
```diff
@@ -182,7 +188,13 @@ impl ExprScopes {
         let root = scopes.root_scope();
         for root_expr in roots {
             let mut scope = scopes.new_scope(root);
-            compute_expr_scopes(root_expr, store, &mut scopes, &mut { scope }, &mut scope);
+            ExprScopeVisitor {
+                store,
+                scopes: &mut scopes,
```
ChayimFriedman2, line 193: At this point you do not need to store references in the visitor anymore, you can store the values directly since the visitor gets a mutable reference to itself.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22864/3eaae41d955ac27fca16ea1ef08874cc131ff335..970ff5482ee99cbc0b7e9dddb73e8bd93602939f)*

Veykril: Yea we could, but it is actually a bit annoying due to how we swap out the active scope and what not so I am gonna leave that be for now (got other things I wanna look into).

PR #22869, opened 2026-07-20
fix: const eval of `*_with_overflow` intrinsics for signed integers

### LINKED ISSUE #22871: Type size / layout not shown for `NonZero<i32>`
**rust-analyzer version**: rust-analyzer version: 0.4.2979-standalone (cac0779549 2026-07-18)

**rustc version**: rustc 1.97.1

```rust
struct A {
  value: NonZero<i32>,
}
```
Size/layout of A isn't shown. Noticed it because type containing `time::Date` size wasn't shown.

This issue may be related and recently got fixed https://github.com/rust-lang/rust-analyzer/issues/21814

MintSoup:
The consteval overflow flag calculation for the signed `with_overflow` functions is not correct.
Example:

```rust
const FLAG: bool = 0i8.overflowing_sub(1).1;
```  
Hover over `FLAG` shows true whereas it should clearly be false.

The blast radius of this bug is much larger than this contrived example, as this also breaks struct layout calculation for signed `NonZero` types. This in turn breaks layout calculation for structs from many popular crates in the ecosystem (notably `chrono` naive types) **and their transitive dependents**.

Fixes rust-lang/rust-analyzer#22871.

AI disclosure: AI has been used to help troubleshoot the issue and write the fix.

### DIFF
--- a/crates/hir-ty/src/mir/eval/shim.rs
+++ b/crates/hir-ty/src/mir/eval/shim.rs
@@ -973,17 +973,35 @@ impl<'a, 'db> Evaluator<'a, 'db> {
                     self.interner(),
                     [lhs.ty, Ty::new_bool(self.interner())].into_iter(),
                 );
+                let is_signed = matches!(lhs.ty.kind(), TyKind::Int(_));
                 let op_size = self.size_of_sized(lhs.ty, locals, "operand of add_with_overflow")?;
-                let lhs = u128::from_le_bytes(pad16(lhs.get(self)?, false));
-                let rhs = u128::from_le_bytes(pad16(rhs.get(self)?, false));
-                let (ans, u128overflow) = match name {
-                    "add_with_overflow" => lhs.overflowing_add(rhs),
-                    "sub_with_overflow" => lhs.overflowing_sub(rhs),
-                    "mul_with_overflow" => lhs.overflowing_mul(rhs),
-                    _ => unreachable!(),
+                let lhs = u128::from_le_bytes(pad16(lhs.get(self)?, is_signed));
+                let rhs = u128::from_le_bytes(pad16(rhs.get(self)?, is_signed));
+                let (ans, is_overflow) = if is_signed {
+                    let (ans, overflow) = match name {
+                        "add_with_overflow" => (lhs as i128).overflowing_add(rhs as i128),
+                        "sub_with_overflow" => (lhs as i128).overflowing_sub(rhs as i128),
+                        "mul_with_overflow" => (lhs as i128).overflowing_mul(rhs as i128),
+                        _ => unreachable!(),
+                    };
+                    // Below 16 bytes the `i128` arithmetic itself cannot overflow, but the
+                    // result may not fit the operand type: check with a sign-extension
+                    // round-trip.
+                    let out_of_range = op_size < 16 && {
+                        let shift = 128 - 8 * op_size as u32;
+                        (ans << shift) >> shift != ans
+                    };
+                    (ans as u128, overflow || out_of_range)
+                } else {
+                    let (ans, overflow) = match name {
+                        "add_with_overflow" => lhs.overflowing_add(rhs),
+                        "sub_with_overflow" => lhs.overflowing_sub(rhs),
+                        "mul_with_overflow" => lhs.overflowing_mul(rhs),
+                        _ => unreachable!(),
+                    };
+                    let out_of_range = ans.to_le_bytes()[op_size..].iter().any(|&it| it != 0);
+                    (ans, overflow || out_of_range)
                 };
-                let is_overflow = u128overflow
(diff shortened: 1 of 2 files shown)


crates/hir-ty/src/mir/eval/shim.rs
```diff
@@ -973,17 +973,35 @@ impl<'a, 'db> Evaluator<'a, 'db> {
...
+                    let (ans, overflow) = match name {
+                        "add_with_overflow" => (lhs as i128).overflowing_add(rhs as i128),
+                        "sub_with_overflow" => (lhs as i128).overflowing_sub(rhs as i128),
+                        "mul_with_overflow" => (lhs as i128).overflowing_mul(rhs as i128),
+                        _ => unreachable!(),
+                    };
+                    // Below 16 bytes the `i128` arithmetic itself cannot overflow, but the
+                    // result may not fit the operand type: check with a sign-extension
+                    // round-trip.
+                    let out_of_range = op_size < 16 && {
```
ChayimFriedman2, line 990: The existing check is good enough (and simpler). You don't need to duplicate it, and certainly not change it. The `if` should only return `(ans, u128overflow, if is_signed { 0xFF } then { 0 })` and then you change the check for 0 in `out_of_range` to check for 0 or 0xFF.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22869/af1fe01c98cfb006b3e12bcca4941551f1c96273..59498ef3003fcca35839ca9c89b6ca76eb3a3546)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/mir/eval/shim.rs
MintSoup: Did you mean something like this?

```rust
let is_signed = matches!(lhs.ty.kind(), TyKind::Int(_));
let op_size = self.size_of_sized(lhs.ty, locals, "operand of add_with_overflow")?;
let lhs = u128::from_le_bytes(pad16(lhs.get(self)?, is_signed));
let rhs = u128::from_le_bytes(pad16(rhs.get(self)?, is_signed));
let (ans, is_overflow) = match name {
    "add_with_overflow" => lhs.overflowing_add(rhs),
    "sub_with_overflow" => lhs.overflowing_sub(rhs),
    "mul_with_overflow" => lhs.overflowing_mul(rhs),
    _ => unreachable!(),
};
let extra = if is_signed { 0xFF } else { 0 };
let out_of_range =
    ans.to_le_bytes()[op_size..].iter().any(|&it| it != 0 && it != extra);
let is_overflow = vec![u8::from(is_overflow || out_of_range)];
```

I'm afraid this won't work for `sub_with_overflow(0i8, 1)`

ChayimFriedman2: No, rather:
```rust
let (ans, is_overflow) = if is_signed {
    let (lhs, rhs) = (lhs as i128, rhs as i128);
    let ans = match name {
        "add_with_overflow" => lhs.overflowing_add(rhs),
        "sub_with_overflow" => lhs.overflowing_sub(rhs),
        "mul_with_overflow" => lhs.overflowing_mul(rhs),
        _ => unreachable!(),
    } as u128;
    (ans, 0xFF)
} else {
    let ans = match name {
        "add_with_overflow" => lhs.overflowing_add(rhs),
        "sub_with_overflow" => lhs.overflowing_sub(rhs),
        "mul_with_overflow" => lhs.overflowing_mul(rhs),
        _ => unreachable!(),
    };
    (ans, 0)
};
```

MintSoup: That doesn't compile, I believe you meant this?
```rust
let (ans, is_overflow, bg) = if is_signed {
    let (lhs, rhs) = (lhs as i128, rhs as i128);
    let (ans, overflow) = match name {
        "add_with_overflow" => lhs.overflowing_add(rhs),
        "sub_with_overflow" => lhs.overflowing_sub(rhs),
        "mul_with_overflow" => lhs.overflowing_mul(rhs),
        _ => unreachable!(),
    };
    (ans as u128, overflow, 0xFF)
} else {
    let (ans, overflow) = match name {
        "add_with_overflow" => lhs.overflowing_add(rhs),
        "sub_with_overflow" => lhs.overflowing_sub(rhs),
        "mul_with_overflow" => lhs.overflowing_mul(rhs),
        _ => unreachable!(),
    };
    (ans, overflow, 0)
};

let out_of_range =
    ans.to_le_bytes()[op_size..].iter().any(|&it| it != bg && it != 0);

```

Still fails, for e.g. `-128 - 1`, reports as no overflow. I believe the background byte needs to depend on the actual value, not just the signedness.

MintSoup: @ChayimFriedman2 is there any update on this?

PR #22870, opened 2026-07-20
i have added the ifle postfix completion for the if let Err

sahkunal:
Addresses rust-lang/rust-analyzer#21440.

i have added a new postfix completion trigger for the Result that matches the Err variant, without bringing the change in the existing one for Ok.
typing .ifle on a result expression expands it into if let Err(...)=..., the same way .ifl has done for Ok.
for a safer side, I have added a test to confirm the changes and ran the complete test suite locally, happy to say, all 747 tests passed . yeeey


ChayimFriedman2:
I don't actually think this carries its weight, when a user wanting that can easily add their own.

But: @rust-lang/rust-analyzer do you agree?

sahkunal:
proc-macro-srv is failing at the toolchain install step, looks like a pre-existing CI infra. issue unrelated to this change. I have confirmed this by rerunning via an empty commit with the same failure.

sahkunal:
> I don't actually think this carries its weight, when a user wanting that can easily add their own.
> 
> But: @rust-lang/rust-analyzer do you agree?

i think you are right, i respect your thought and i am happy to close this .

PR #22883, opened 2026-07-21
internal: add `Module::path_segments`, dedup the path-to-root name walk

enomado:
Five call sites spelled out the same walk by hand — `path_to_root`, reverse, `filter_map` the names — before rendering it as `::`-joined text: `ModuleDef::canonical_path`, `cli::full_name_of_item`, two in `analysis_stats`, and hover's path rendering.

`Module::path_segments` names that walk once and returns `Name`s rather than a finished `String`: the call sites disagree on the edition to display with (the CLI ones pin `Edition::LATEST`, hover takes it as a parameter), so joining inside the helper would have fit none of them.

`ModuleDef::canonical_module_path` is the nearby sibling that yields the `Module`s themselves; it does not serve these sites. Four of the five hold only a `Module`, and it hangs off `ModuleDef` — while routing the fifth, `canonical_path`, through it would put the `filter_map` this change removes straight back. The two are cross-referenced in the docs instead.

Left alone: the two walks in `runnables`, which display each segment with its own module's edition and so need the `Module`, not just its name.

No behavior change, with one wrinkle: `canonical_path` now renders each segment into a `String` before joining. The iterator is lazy and owns its `Name`s, so a borrowed `impl Display` cannot escape the closure — the old code got away with it by collecting into a `Vec` first. The other four sites already did this.

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -438,12 +438,9 @@ impl ModuleDef {
     }
 
     pub fn canonical_path(&self, db: &dyn HirDatabase, edition: Edition) -> Option<String> {
-        let mut segments = vec![self.name(db)?];
-        for m in self.module(db)?.path_to_root(db) {
-            segments.extend(m.name(db))
-        }
-        segments.reverse();
-        Some(segments.iter().map(|it| it.display(db, edition)).join("::"))
+        let name = self.name(db)?;
+        let segments = self.module(db)?.path_segments(db).chain(Some(name));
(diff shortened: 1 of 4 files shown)


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22883/5c1dbddeb36dad603576cf112529788c386f83b5..b633e68db446e300c0952c0bea437cf389ad1a9c)*

PR #22885, opened 2026-07-22
internal: refer to salsa macros using `salsa::`

ada4a:
Sometimes it's `salsa_macros::`, sometimes it's `salsa::` -- the inconsistency was annoying.

### DIFF
--- a/crates/base-db/src/input.rs
+++ b/crates/base-db/src/input.rs
@@ -447,7 +447,7 @@ impl BuiltDependency {
 
 pub type CratesIdMap = FxHashMap<CrateBuilderId, Crate>;
 
-#[salsa_macros::input]
+#[salsa::input]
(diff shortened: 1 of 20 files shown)


ada4a:
Apparently this would allow me to remove `salsa_macros` as a dependency from several crates -- should be fine to do, right?

ada4a:
Hm, https://github.com/rust-lang/rust-analyzer/pull/19716/commits/996d6ed1a3f02bb70c4fa6c16a3a2b0313406c81 makes me think that we actually want to go in the opposite direction? Writing out `salsa_macros::` everywhere gets somewhat verbose though...

cc @Veykril

Veykril:
Yea I tried this for better compile times but I don't think there is a point in this after all as we'd need to get all our dependencies to also agree with this which isn't really feasible

Veykril (approved):

PR #22886, opened 2026-07-22
fix: Fix glob import shadowing bug

### LINKED ISSUE #21668: False `E0308` caused by wrong name resolution
**rust-analyzer version**: 0.3.2795-standalone (00a9173e57 2026-02-15)

**rustc version**: rustc 1.92.0-nightly (9f32ccf35 2025-09-21)

**editor or extension**: Cursor (Extension version v0.3.2795)

**relevant settings**: -

**repository link (if public, optional)**: https://github.com/djeedai/bevy_hanabi (See additional context)

```rust
// cargo add bevy_hanabi
use bevy_hanabi::prelude::*;

fn main() {
    let gradient = bevy_hanabi::Gradient::new();
    let modifier = ColorOverLifetimeModifier::new(gradient);
}
```

**Additional Context**:
This code compiles, but rust-analyzer is reporting an E0308 diagnostic `expected Gradient, found Gradient<{unknown}>`. 

When hovering `Gradient` in `ColorOverLifetimeModifier::new`'s signature in the dependency source, rust-analyzer shows `bevy_ui::gradients::Gradient` (an enum) instead of `bevy_hanabi::gradient::Gradient<T>`.

Based off source from [bevy_hanabi](https://github.com/djeedai/bevy_hanabi/blob/main/src/modifier/output.rs):
```rust
use bevy::prelude::*; // brings in `bevy_ui::gradients::Gradient`
use crate::Gradient; // brings in `bevy_hanabi::gradient::Gradient`

// ...

impl ColorOverLifetimeModifier {
    // rust-analyzer is reporting the type here as `bevy_ui::gradients::Gradient`
    fn new(gradient: Gradient<Vec4>) -> Self {
        // ...
    }
}
```

### LINKED ISSUE #21679: False positive error E0308 due to type confusion from a glob use and use statement ordering
**rust-analyzer version**: `rust-analyzer 1 (7cb789d982 2026-02-01)` (from `rust-analyzer --version`)

**rustc version**: `rustc 1.93.1 (01f6ddf75 2026-02-11) (Arch Linux rust 1:1.93.1-1.1)`

**editor or extension**: happens with any including commandline

**relevant settings**: defaults

**repository link (if public, optional)**: https://archlinux.org/packages/extra/x86_64/rust-analyzer/

**code snippet to reproduce**: `cargo new something`, then put this into `src/main.rs`
```rust
pub use std::cmp::*;

pub mod evil {
    pub struct Ordering(pub i32);
}

pub mod oblivious {
    use crate::Ordering;

    pub fn what() -> Ordering {
        Ordering(2)
    }
}
pub use evil::Ordering;


fn main() {
    println!("Hello, world!");
}

```

This snippet causes a false positive error from rust-analyzer that cargo check doesn't report:

```
architector4@AGOGUS:/tmp/something$ cargo check
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.00s
architector4@AGOGUS:/tmp/something$ rust-analyzer diagnostics .
at crate ass, file /tmp/something/src/main.rs: Error RustcHardError("E0308") from LineCol { line: 10, col: 8 } to LineCol { line: 10, col: 19 }: expected Ordering, found Ordering

diagnostic scan complete

Error: diagnostic error detected

Stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: start_thread
             at /usr/src/debug/glibc/glibc/nptl/pthread_create.c:454:8
   5: __GI___clone3
             at /usr/src/debug/glibc/glibc/misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:78:0
```

Not sure if that diagnostic error error or stack backtrace should be there normally, but I guess that's there too

Note the position of that last `pub use` being _after_ the two modules were defined. Moving it before `oblivious` resolves the false positive.

(I sprinkled `pub` everywhere to get rid of "unused" warnings, but that doesn't seem to be necessary to reproduce the bug.)

### LINKED ISSUE #22783: Glob-imported tuple variant shadows a same-named struct in type position (false "missing structure fields")
**rust-analyzer version**: 0.3.2971
**rustc version**: rustc 1.97.0 (2d8144b78 2026-07-07)
**editor or extension**: VS Code 1.128.0, rust-analyzer extension 0.3.2971
**relevant settings**: none

rust-analyzer resolves a *type* path to a glob-imported enum *variant* instead of the same-named struct, and reports a false `E0063` on the struct literal. `cargo check --all-targets` is clean and warning-free (also passes `rustc -D warnings`). The diagnostic is rust-analyzer's own, not forwarded from `cargo check`: it also reproduces against a minimal LSP client with `checkOnSave` disabled, and its `source` is `rust-analyzer`.

No dependencies; the whole crate (edition 2024):

```rust
#[derive(Debug)]
pub enum E {
    S(u32),
}

use E::*;

pub fn use_variant(e: E) -> u32 {
    match e {
        S(v) => v,
    }
}

pub struct S {}

pub mod inner {
    use super::*;

    pub fn f() -> S {
        S {} // rust-analyzer: "missing structure fields: - 0"
    }
}
```

The missing field it names is `0` -- the tuple field of `E::S(u32)` -- so it is checking the struct literal against the variant rather than against `pub struct S {}`. rustc keeps types and values in separate namespaces, so `S` in type position unambiguously refers to the struct.

**Minimal trigger.** Three ingredients, each required -- removing any one clears the diagnostic:

1. **A derive on the enum.** `#[derive(Debug)]` suffices; with no derive at all, resolution is correct. A proc-macro derive is not needed.
2. **A tuple variant** colliding with the struct name. A unit variant does not trigger it.
3. **A chain of glob imports** to the use site: `use E::*;` at the crate root *plus* `use super::*;` in the child module. Importing `S` explicitly in `inner`, or moving `f` to the crate root, clears it.

**Notes.**

- `rust-analyzer diagnostics .` (the CLI) does not surface this; it only appears over LSP.
- Downstream this surfaces as false `E0308`s on unrelated expressions once the binding becomes `{unknown}`, with no hint that a name collision caused it.
- Possibly the same root cause as #21837 (variants colliding with same-named structs, derive-generated `use super::*`), though that repro needs a proc macro and this one does not.

kristoff3r:
Fixes rust-lang/rust-analyzer#21668
Fixes rust-lang/rust-analyzer#21679
Fixes rust-lang/rust-analyzer#22783

In most cases overriding glob imports work. However, a single-name import that first resolves indeterminately and later re-resolves to a shadowed import could never overwrite its own stale type binding. To fix we check for that case explicitly.

This was triaged and the initial patch was written by AI, then heavily modified by me. The existing logic is quite complicated, so I'm not sure if this has unintended side effects. The existing tests pass though.

### DIFF
--- a/crates/hir-def/src/item_scope.rs
+++ b/crates/hir-def/src/item_scope.rs
@@ -621,7 +621,11 @@ impl ItemScope {
                             // for that.
                         }
                         _ => {
-                            if glob_imports.types.remove(&lookup) {
+                            // A non-glob import either shadows a glob import of the same
+                            // name, or re-resolves a stale binding it recorded earlier.
+                            if glob_imports.types.remove(&lookup)
+                                || entry.get().is_reresolved_by(&fld.def, import)
+                            {
                                 let prev = std::mem::replace(&mut fld.import, import);
                                 if let Some(import) = import {
                                     self.use_imports_types.insert(
@@ -659,19 +663,22 @@ impl ItemScope {
                     changed = true;
                 }
                 Entry::Occupied(mut entry)
-                    if !matches!(import, Some(ImportOrExternCrate::Glob(..)))
-                        && glob_imports.values.remove(&lookup) =>
+                    if !matches!(import, Some(ImportOrExternCrate::Glob(..))) =>
                 {
-                    cov_mark::hit!(import_shadowed);
-
                     let import = import.and_then(ImportOrExternCrate::import_or_glob);
-                    let prev = std::mem::replace(&mut fld.import, import);
-                    if let Some(import) = import {
-                        self.use_imports_values
-                            .insert(import, prev.map_or(ImportOrDef::Def(fld.def), Into::into));
+                    if glob_imports.values.remove(&lookup)
+                        || entry.get().is_reresolved_by(&fld.def, import)
+                    {
+                        cov_mark::hit!(import_shadowed);
+
+                        let prev = std::mem::replace(&mut fld.import, import);
+                        if let Some(import) = import {
+                            self.use_imports_values
+                                .insert(import, prev.map_or(ImportOrDef::Def(fld.def), Into::into));
+                        }
+                        entry.insert(fld);
+                        changed = true;
                     }
-                    entry.insert(fld);
-                    changed = true;
                 }
                 _ => {}
             }
@@ -699,7 +706,8 @@ impl ItemScope {
                 }
                 Entry::Occupied(mut entry)
                     if !matches!(import, Some(ImportOrExternCrate::Glob(..)))
-                        && glob_imports.macros.remove(&lookup) =>
+                        && (glob_imports.macros.remove(&lookup)
+                            || entry.get().is_reresolved_by(&fld.def, import)) =>
                 {
                     cov_mark::hit!(import_shadowed);
                     let prev = std::mem::replace(&mut fld.import, import);
--- a/crates/hir-def/src/per_ns.rs
+++ b/crates/hir-def/src/per_ns.rs
@@ -35,6 +35,16 @@ pub struct Item<Def, Import = ImportId> {
     pub import: Option<Import>,
 }
 
+impl<Def: PartialEq, Import: Copy + PartialEq> Item<Def, Import> {
+    /// Whether `import` is the same import that produced `self`, now resolving to a different
+    /// `def`. This happens when an import is first recorded as an indeterminate resolution
+    /// (e.g. only one namespace was available at the time) and later re-resolves to another
+    /// def, such as an explicit import that shadows a glob only after the glob has been seen.
+    pub(crate) fn is_reresolved_by(&self, def: &Def, import: Option<Import>) -> bool {
+        import.is_some() && self.import == import && self.def != *def
+    }
+}
+
 pub type TypesItem = Item<ModuleDefId, ImportOrExternCrate>;
 pub type ValuesItem = Item<ModuleDefId, ImportOrGlob>;
 // May be Externcrate for `[macro_use]`'d macros
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -1813,4 +1813,29 @@ fn main() {
 "#,
         );
     }
+
+    #[test]
+    fn regression_21668() {
+        check_diagnostics(
+            r#"
+mod std {
+    pub enum Ordering { Less }
+}
+pub use std::*;
+
+pub mod evil {
+    pub struct Ordering(pub i32);
+}
+
+pub mod oblivious {
+    use crate::Ordering;
+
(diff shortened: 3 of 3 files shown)


ChayimFriedman2:
Without looking at the code yet, name resolution in Rust is an extremely complicated process (and we definitely don't have extensive enough testing for it). Unfortunately, I don't think we can allow a (non-trivial) patch to it to be written by AI by someone that does not know it and the r-a codebase well enough - it's just too easy to miss something.

kristoff3r:
> Without looking at the code yet, name resolution in Rust is an extremely complicated process (and we definitely don't have extensive enough testing for it). Unfortunately, I don't think we can allow a (non-trivial) patch to it to be written by AI by someone that does not know it and the r-a codebase well enough - it's just too easy to miss something.

That's fair. My hope was that since it is short and (to me) seems like a reasonably scoped fix that it might be helpful. Feel free to close if it's not.

Veykril:
I think if this fixes some issues it's fine to accept this here. The import resolution code is not gonna get prettier until we rewrite it wholesale anyways, so reducing some of the bugs here seems beneficial either way. Given I landed a weird special case some months ago as well to fix a specific issue with the windows crate.

kristoff3r:
> I think if this fixes some issues it's fine to accept this here. The import resolution code is not gonna get prettier until we rewrite it wholesale anyways, so reducing some of the bugs here seems beneficial either way. Given I landed a weird special case some months ago as well to fix a specific issue with the windows crate.

Thanks!

I'll try to run it through some big code bases tomorrow for more testing, if I find anything I can add more regression tests. That should be helpful for the rewrite as well.

ChayimFriedman2:
> I think if this fixes some issues it's fine to accept this here. The import resolution code is not gonna get prettier until we rewrite it wholesale anyways, so reducing some of the bugs here seems beneficial either way. Given I landed a weird special case some months ago as well to fix a specific issue with the windows crate.

Just fixing the issue is not enough, e.g. we know one way to fix it that has an unknown and perhaps significant impact on performance and might even make name resolution unable to progress in more cases.

ChayimFriedman2 (approved):
I'm quite surprised (I expected this to be harder, apparently even I do not fully understand our name resolution 😄), but having reviewed your PR thoroughly, this fix is indeed correct. We might even want to look at the hacky way @Veykril fixed the old variant of this issue, it might no longer be needed.

Please squash and then LGTM.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22886/1174734d9c6453d13d2b5e3d578512c579ca37f1..6f64ed36bfd5b4f2c1b891c2f3e4282b02c82a7f)*

kristoff3r:
> I'm quite surprised (I expected this to be harder, apparently even I do not fully understand our name resolution 😄), but having reviewed your PR thoroughly, this fix is indeed correct. We might even want to look at the hacky way @Veykril fixed the old variant of this issue, it might no longer be needed.
> 
> Please squash and then LGTM.

Done!

PR #22892, opened 2026-07-23
fix: merge_imports panic on invalid paths

Wilfred:
When computing code actions, rust-analyzer would previously panic if a path had malformed segments. We already fixed empty path segments in 343d5ff383, which is the most frequent case, but I've seen crashes with malformed paths too (most commonly a trailing `:` or `#`).

Ensure that handle invalid path segments too, and add a test.

AI disclosure: Code partly written by GPT-5.6 Sol

### DIFF
--- a/crates/ide-assists/src/handlers/merge_imports.rs
+++ b/crates/ide-assists/src/handlers/merge_imports.rs
@@ -838,4 +838,15 @@ use foo::$0;
 ",
         );
     }
+
+    #[test]
+    fn test_merge_with_malformed_colon_path_segment() {
+        check_assist_not_applicable(
+            merge_imports,
+            r"
+use foo::bar;
+use foo::$0:;
+",
+        );
+    }
 }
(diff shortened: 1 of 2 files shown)


Veykril:
Do you have the concrete panic on hand that this runs into? Mainly curious if there is a more general fix that needs to be done here maybe

Wilfred:
The test reproduces the panic, did you want anything else specifically?

Veykril:
The panic message and location, but I am home now so i can briefly check that myself

Veykril:
Ah okay, this is just the make api being brittle once again

Veykril (approved):
This special case seems fine in that case, even when we fix the make api fully

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22892/1174734d9c6453d13d2b5e3d578512c579ca37f1..fad5d20d574679fee1c6668a81abcd4d351665f1)*

PR #22896, opened 2026-07-24
fix: uses bool instead pat ty in guard

A4-Tacks:
Example
---
```rust
enum E { V }
fn foo() {
    match E::V { _ if a$0 => {} }
}
```

**Before this PR**

```rust
ty: E, name: ?
```

**After this PR**

```rust
ty: bool, name: ?
```

### DIFF
--- a/crates/ide-completion/src/context/analysis.rs
+++ b/crates/ide-completion/src/context/analysis.rs
@@ -794,6 +794,10 @@ fn expected_type_and_name<'db>(
                     }.map(TypeInfo::original);
                     (ty, None)
                 },
+                ast::MatchGuard(it) => {
+                    let ty = it.condition().and_then(|e| sema.type_of_expr(&e)).map(TypeInfo::original);
+                    (ty, None)
+                },
                 ast::IdentPat(it) => {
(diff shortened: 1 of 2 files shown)


crates/ide-completion/src/context/analysis.rs
```diff
@@ -794,6 +794,10 @@ fn expected_type_and_name<'db>(
                     }.map(TypeInfo::original);
                     (ty, None)
                 },
+                ast::MatchGuard(it) => {
+                    let ty = it.condition().and_then(|e| sema.type_of_expr(&e)).map(TypeInfo::original);
```
ChayimFriedman2, line 798: This seems too fragile. Why not just return `bool`, given that this should always be the type?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22896/1174734d9c6453d13d2b5e3d578512c579ca37f1..0272cd6184318cc79b997cb97431f1cfe3f41a9e)*

A4-Tacks: There is no precedent, I repeated the logic of `IfExpr`

ChayimFriedman2 (approved):

PR #22897, opened 2026-07-24
feat: use guard condition instead of bool body if-expr

A4-Tacks:
I'm not sure if this is reasonable, the user may want to add a complex else branch

Example
---
```rust
fn main() {
    match 92 {
        x $0if x > 10 => false,
        _ => true
    }
}
```

**Before this PR**

```rust
fn main() {
    match 92 {
        x => if x > 10 {
            false
        },
        _ => true
    }
}
```

**After this PR**

```rust
fn main() {
    match 92 {
        x => x <= 10,
        _ => true
    }
}
```


Veykril:
I think that is too surprising and specific. It also likely goes against the idea of what the assist is meant to do, it strictly moves the guard to an if expression

A4-Tacks:
Okay

PR #22900, opened 2026-07-24
fix: Fix inverted arm-count limit check in `add_missing_match_arms`

ChihweiLHBird:
### Description

The assist caps generated arms at 256, but the array guard computes the count with the operands swapped, `len ^ variants` instead of `variants ^ len`. A `match` on `[bool; 12]` therefore eagerly builds 4096 patterns on every code-action request (up to 65,536 at length 16), while small valid cases like a 9-variant `[E; 2]` (81 arms) are wrongly rejected. The tuple branch also counts arms with an unchecked multiplication that overflows on a 64-element `bool` tuple.

This PR computes the count in the correct order with checked arithmetic (overflow counts as over-limit) and caps the array length, which bounds the size of each generated pattern. Behavior changes only for cases that were over the intended limit or wrongly blocked by the swapped operands. Adds five regression tests.

AI disclosure: developed with AI assistance; I reviewed, verified, and understand the changes.

### DIFF
--- a/crates/ide-assists/src/handlers/add_missing_match_arms.rs
+++ b/crates/ide-assists/src/handlers/add_missing_match_arms.rs
@@ -140,11 +140,11 @@ pub(crate) fn add_missing_match_arms(acc: &mut Assists, ctx: &AssistContext<'_, 
             .iter()
             .any(|enum_def| enum_def.is_non_exhaustive(ctx.db(), module.krate(ctx.db())));
 
-        let mut n_arms = 1;
+        let mut n_arms = Some(1usize);
         let variants_of_enums: Vec<Vec<ExtendedVariant>> = enum_defs
             .into_iter()
             .map(|enum_def| enum_def.variants(ctx.db()))
-            .inspect(|variants| n_arms *= variants.len())
+            .inspect(|variants| n_arms = n_arms.and_then(|n| n.checked_mul(variants.len())))
             .collect();
 
         // When calculating the match arms for a tuple of enums, we want
@@ -155,7 +155,7 @@ pub(crate) fn add_missing_match_arms(acc: &mut Assists, ctx: &AssistContext<'_, 
 
         // A number of arms grows very fast on even a small tuple of large enums.
         // We skip the assist beyond an arbitrary threshold.
-        if n_arms > 256 {
+        if n_arms.is_none_or(|n_arms| n_arms > 256) {
             return None;
         }
 
@@ -187,7 +187,7 @@ pub(crate) fn add_missing_match_arms(acc: &mut Assists, ctx: &AssistContext<'_, 
         let is_non_exhaustive = enum_def.is_non_exhaustive(ctx.db(), module.krate(ctx.db()));
         let variants = enum_def.variants(ctx.db());
 
-        if len.pow(variants.len() as u32) > 256 {
+        if len > 256 || variants.len().checked_pow(len as u32).is_none_or(|n_arms| n_arms > 256) {
             return None;
         }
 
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
First, I already said multiple times that I am rewriting this assist (I really should finish that). That said, this change is really small so I would not be opposed...

...**If it was implemented correctly**. Which it is clearly not. You (or more correctly, the AI) added a bunch of completely unnecessary checked arithmetic and unnecessary and unrelated tests. Which throws a heavy shade on your words that you "reviewed, verified, and understand the changes", and makes me outright reject this PR, without even asking you to fix it.

If you will repeat this behavior, we'll reach out to moderation.

ChihweiLHBird:
> First, I already said multiple times that I am rewriting this assist (I really should finish that). That said, this change is really small so I would not be opposed...
> 
> ...**If it was implemented correctly**. Which it is clearly not. You (or more correctly, the AI) added a bunch of completely unnecessary checked arithmetic and unnecessary and unrelated tests. Which throws a heavy shade on your words that you "reviewed, verified, and understand the changes", and makes me outright reject this PR, without even asking you to fix it.
> 
> If you will repeat this behavior, we'll reach out to moderation.

I am sorry to hear that... I did look into this issue and the diff, and I thought I understood it. At least I believe I tried my best and spent hours on this. I never intended to spam with AI slop. I am sorry if this was an incorrect fix and bothered you.

For the tests, maybe one or two of them are redundant, but if you tried you can see the added tests would fail with the existing implementation.

I responsibly disclose my AI usage according to the policy.

My motivation for working on this is the VS code extension often work less ideal in my environment and I wanted to help to stabilize it. It could be my environment's issue too.

crates/ide-assists/src/handlers/add_missing_match_arms.rs
```diff
@@ -187,7 +187,7 @@ pub(crate) fn add_missing_match_arms(acc: &mut Assists, ctx: &AssistContext<'_,
         let is_non_exhaustive = enum_def.is_non_exhaustive(ctx.db(), module.krate(ctx.db()));
         let variants = enum_def.variants(ctx.db());
 
-        if len.pow(variants.len() as u32) > 256 {
+        if len > 256 || variants.len().checked_pow(len as u32).is_none_or(|n_arms| n_arms > 256) {
```
ChihweiLHBird, line 190: If this `len > 256` "authentic check" looks unnecessary and bothered you, I can tell you there is a reason for having it.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22900/ae631b9fffa2e47b7b8e0caa5effdc574472cfdd..920ad1843bba0df83d32796764dc0d705b676942)*

ChayimFriedman2:
You have added 5 new tests for a change that has *at most* 2 new behaviors: disabling the assist on some places it was previously offered on and enabling it on others. That requires at most 2 new tests. And adding redundant tests is a thing that AI tools are known to love to do.

You also changed a whole bunch of operations to be checked when that's completely unnecessary - when was the last time you had a 64-elements tuple? The only operation that might realistically overflow is the power (and anyway using saturating arithmetic is better here).

ChihweiLHBird:
> You have added 5 new tests for a change that has *at most* 2 new behaviors: disabling the assist on some places it was previously offered on and enabling it on others. That requires at most 2 new tests. And adding redundant tests is a thing that AI tools are known to love to do.
> 
> You also changed a whole bunch of operations to be checked when that's completely unnecessary - when was the last time you had a 64-elements tuple? The only operation that might realistically overflow is the power (and anyway using saturating arithmetic is better here).

2 tests were for the observable behavior change, and 1 test for the panic case. I admit that the 2 tests left are not strictly necessary because they only guard the `len>256` check.

The `len>256` is there to defend list with 1 million size, which will then cause creation of vec of 1 million items to calculate in the patterns matching block below.

There are indeed not many normal use cases for 64 elements tuple, but I consider this as a very critical project because it's the only official plugin of the most popular editor in the world for Rust. It may have to handle extraordinary cases and maybe also malicious code that's designed to crash the IDE as a part of DoS attack effort. Imagine if someone hides a large turtle or 1 million size list with a match in your codebase, and you didn't realize. The extension might keep crashing (if not other guards in place) and it's hard to figure out why.

Again, I may be wrong because I am not familiar with the full picture of the project, and I am not a developer with years of Rust experience. But I can assure you that I always responsibly read, understand, and verify the stuff I submitted.

ChayimFriedman2:
rust-analyzer does not defend against malicious code. If you run r-a on malicious code, you have much, much bigger problems than DoS - it's very easy to gain arbitrary RCE using it.

> 2 tests were for the observable behavior change, and 1 test for the panic case

Which panic? If you mean the overflow, as I said it isn't needed.

ChihweiLHBird:
> rust-analyzer does not defend against malicious code. If you run r-a on malicious code, you have much, much bigger problems than DoS - it's very easy to gain arbitrary RCE using it.
> 
> > 2 tests were for the observable behavior change, and 1 test for the panic case
> 
> Which panic? If you mean the overflow, as I said it isn't needed.

Yes I mean the overflow.

Understood, thanks for the info.

If you changed mind I can restore the branch with unnecessary tests and overflow test removed. For the overflow, personally I would still prefer the safer way in the non-test code (like `checked_pow`). I can also get `len>256` check removed.

And if you wouldn't like to deal with all these we can just leave it there.

PR #22901, opened 2026-07-24
Fix: strip leading asterisk decoration from block doc comments

### LINKED ISSUE #1759: Doc comment block should strip leading asterisk (*)
In looking at `struct ClientCapabilities` of lsp-types defined as

```rust
/**
 * Where ClientCapabilities are currently empty:
 */
pub struct ClientCapabilities {
    // ...
}
```

I noticed that hover shows the comment as

> * Where ClientCapabilities are currently empty: 

In looking at https://docs.rs/lsp-types/0.61.0/lsp_types/struct.ClientCapabilities.html we can see  that it's rendered without the leading asterisk (or bullet in markdown).

My initial thought was that maybe we need to strip leading `*` off of block dock comments if present in `DocCommentsOwner::doc_comment_text` but that could also end up stripping markdown. Maybe it should be done in `format_docs` instead when we look for markdown... but by then we've lost all knowledge of the comment except for it's string contents.

kivancgnlp:
## Summary

- Strip a consistent leading `*` decoration from multiline block doc comments, matching rustdoc behavior.
- Preserve existing behavior for line docs, `#[doc = "..."]` attributes, `*foo`, and blocks without a consistent star column.
- Adjust per-line source-map offsets by the stripped byte count so doc links and syntax highlighting continue to map to the correct source ranges.

## Tests Passed :

- `cargo test -p hir-def`
- `cargo test -p ide -- hover doc_links`
- `cargo test -p ide -- syntax_highlighting`
- `cargo fmt -p hir-def -p ide -- --check`
- `cargo clippy -p hir-def -p ide --all-targets -- --cap-lints warn`

Fixes rust-lang/rust-analyzer#1759

### DIFF
--- a/crates/hir-def/src/attrs/docs.rs
+++ b/crates/hir-def/src/attrs/docs.rs
@@ -201,7 +201,8 @@ impl Docs {
 
     fn extend_with_doc_comment(&mut self, comment: ast::Comment, indent: &mut usize) {
         let Some((doc, offset)) = comment.doc_comment() else { return };
-        self.extend_with_doc_str(doc, comment.syntax().text_range().start() + offset, indent);
+        let offset = comment.syntax().text_range().start() + offset;
+        self.extend_with_doc_str(doc, offset, indent, comment.kind().shape);
     }
 
     fn extend_with_doc_attr(&mut self, value: ast::String, indent: &mut usize) {
@@ -209,7 +210,9 @@ impl Docs {
         let value_offset = value_offset.start();
         let Ok(value) = value.value() else { return };
         // FIXME: Handle source maps for escaped text.
-        self.extend_with_doc_str(&value, value_offset, indent);
+        //
+        // rustc passes `CommentKind::Line` for desugared `#[doc = "..."]` attributes.
+        self.extend_with_doc_str(&value, value_offset, indent, ast::CommentShape::Line);
     }
 
     pub(crate) fn extend_with_doc_str(
@@ -217,30 +220,94 @@ impl Docs {
         doc: &str,
         offset_in_ast: TextSize,
         indent: &mut usize,
+        shape: ast::CommentShape,
     ) {
-        self.push_doc_lines(doc, Some(offset_in_ast), indent);
+        self.push_doc_lines(doc, Some(offset_in_ast), indent, shape);
     }
 
     fn extend_with_unmapped_doc_str(&mut self, doc: &str, indent: &mut usize) {
-        self.push_doc_lines(doc, None, indent);
+        // Macro-expanded doc strings are desugared, so pass `CommentShape::Line` matching
+        // rustc's `CommentKind::Line`.
+        self.push_doc_lines(doc, None, indent, ast::CommentShape::Line);
     }
 
-    fn push_doc_lines(&mut self, doc: &str, mut ast_offset: Option<TextSize>, indent: &mut usize) {
-        for line in doc.split('\n') {
-            self.docs_source_map
-                .push(DocsSourceMapLine { string_offset: TextSize::of(&self.docs), ast_offset });
-            if let Some(ref mut offset) = ast_offset {
-                *offset += TextSize::of(line) + TextSize::of("\n");
-            }
-
-            let line = line.trim_end();
-            if let Some(line_indent) = line.chars().position(|ch| !ch.is_whitespace()) {
-                // Empty lines are handled because `position()` returns `None` for them.
-                *indent = std::cmp::min(*indent, line_indent);
-            }
-            self.docs.push_str(line);
-            self.docs.push('\n');
+    /// Beautifies `doc` and appends the result to `self.docs`, one line at a time via
+    /// [`Docs::push_doc_line`]. Mirrors rustc's [`beautify_doc_string`], delegating to
+    /// [`get_vertical_trim`] and [`get_horizontal_trim`] for the multi-line case.
+    ///
+    /// Individual `///` line comments always reach us as a single-line `doc`, so the
+    /// `!doc.contains('\n')` fast path fires and the multi-line logic never runs on them.
+    /// Desugared `#[doc = "..."]` strings and macro-expanded docs also route through here
+    /// with `shape = CommentShape::Line`, matching rustc.
+    ///
+    /// Unlike rustc's version, which joins the beautified lines into a new interned `Symbol`,
+    /// this port pushes each line individually via [`Docs::push_doc_line`] and pairs it with
+    /// its byte offset relative to `doc`'s start so the source-map records accurate per-line
+    /// offsets.
+    ///
+    /// [`beautify_doc_string`]: https://github.com/rust-lang/rust/blob/16a623ad672a92409b5c04beb303583c6cf72a7e/compiler/rustc_ast/src/util/comments.rs#L37
+    fn push_doc_lines(
+        &mut self,
+        doc: &str,
+        ast_offset: Option<TextSize>,
+        indent: &mut usize,
+        shape: ast::CommentShape,
+    ) {
+        if !doc.contains('\n') {
+            self.push_doc_line(doc, ast_offset, indent);
+            return;
         }
+
+        let doc_start = doc.as_ptr() as usize;
+        let mut lines: Vec<(&str, TextSize)> = doc
+            .lines()
+            .map(|line| {
+                let offset = TextSize::new((line.as_ptr() as usize - doc_start) as u32);
+                (line, offset)
+            })
+            .collect();
+
+        let raw_lines: Vec<&str> = lines.iter().map(|(l, _)| *l).collect();
+        let lines = match get_vertical_trim(&raw_lines) {
+            Some((i, j)) => &mut lines[i..j],
+            None => &mut lines[..],
+        };
+
+        let raw_lines: Vec<&str> = lines.iter().map(|(l, _)| *l).collect();
+        if let Some(horizontal) = get_horizontal_trim(&raw_lines, shape) {
+            let horizontal_len = TextSize::of(horizontal.as_str());
+            // Strip `"[ \t]*\*"` from each line where present, exactly like rustc.
+            for (line, line_offset) in lines.iter_mut() {
+                if let Some(rest) = line.strip_prefix(horizontal.as_str()) {
+                    *line = rest;
+                    *line_offset += horizontal_len;
+                    if shape == ast::CommentShape::Block
+                        && (*line == "*" || line.starts_with("* ") || line.starts_with("**"))
+                    {
+                        *line = &line[1..];
+                        *line_offset += TextSize::of("*");
+                    }
+                }
+            }
+        }
+
+        for (line, line_offset) in lines.iter().copied() {
+            self.push_doc_line(line, ast_offset.map(|it| it + line_offset), indent);
+        }
+    }
+
+    /// Appends a single beautified line to `self.docs` and records its source-map row.
+    fn push_doc_line(&mut self, line: &str, ast_offset: Option<TextSize>, indent: &mut usize) {
+        self.docs_source_map
+            .push(DocsSourceMapLine { string_offset: TextSize::of(&self.docs), ast_offset });
+
+        let line = line.trim_end();
+        if let Some(line_indent) = line.chars().position(|ch| !ch.is_whitespace()) {
+            // Empty lines are handled because `position()` returns `None` for them.
+            *indent = std::cmp::min(*indent, line_indent);
+        }
+        self.docs.push_str(line);
+        self.docs.push('\n');
     }
 
     fn remove_indent(&mut self, indent: usize, start_source_map_index: usize) {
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
Did you write the PR description using AI? I'm not sure you did, but if yes, please note that [our AI policy](https://github.com/rust-lang/rust-analyzer/blob/master/AI_POLICY.md) forbids doing that.

ChayimFriedman2:
Looking at the code, it's pretty clear that you used AI for the code as well without disclosing it, also contrary to our AI policy. And I'm not sure it's properly reviewed either: I am pretty sure it's possible to implement the same thing using at most half the lines, and probably with better perf.

kivancgnlp:
> Did you write the PR description using AI? I'm not sure you did, but if yes, please note that [our AI policy](https://github.com/rust-lang/rust-analyzer/blob/master/AI_POLICY.md) forbids doing that.

Yes I used Claude for draft message but then I edit the message.

ChayimFriedman2:
So, this is also disallowed. Please take notice for the next time.

kivancgnlp:
> Looking at the code, it's pretty clear that you used AI for the code as well without disclosing it, also contrary to our AI policy. And I'm not sure it's properly reviewed either: I am pretty sure it's possible to implement the same thing using at most half the lines, and probably with better perf.

Claude generated the initial code and tests. I reviewed the changes and tests by myself. I'm sorry that I discovered the policy afterward. I also evaluated possible pitfalls and alternative solutions. I’ll simplify the code.

kivancgnlp:
- Implementation simplified
- Hover documentation was tested using a local rust-analyzer server and VS Code.
- The following behaviors were checked and verified: stars disappeared, formatting remained correct, and the documentation link worked.

kivancgnlp:
Hi @ChayimFriedman2, just checking in on this one. As noted above, I simplified the implementation and removed the AI-generated parts entirely. The current version reflects my own review and manual testing (verified with a local rust-analyzer server and VS Code, doc-comment stars are stripped correctly, formatting stays intact, and the documentation link still resolves).

Happy to make further changes if the simplified approach still isn't the right shape. Let me know what you'd like adjusted.

ChayimFriedman2 (changes requested):
I don't want us to reinvent the wheel. The code to do that in rustc is here:

https://github.com/rust-lang/rust/blob/16a623ad672a92409b5c04beb303583c6cf72a7e/compiler/rustc_resolve/src/rustdoc.rs#L210

The `beautify_doc_string()` does this and perhaps more things. We can copy it almost as-is, perhaps with minor changes.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22901/857dec62c43dd8df13a8f628ef5b9d59bb28b0c0..de72f8fa237c35be6a34eb2bd40038820c49f42e)*

kivancgnlp:
Thanks. Adopted `beautify_doc_string` in [877fbfa](https://github.com/rust-lang/rust-analyzer/commit/877fbfa14aa309b526dd88a66cf653541688ad65) (on top of a rebase onto current master). Kept it as a separate commit so the diff against the original approach is easy to see.

A few notes on the adaptation:

- The changes live in `compute_block_doc_trim` and return a `BlockDocTrim { keep, horizontal }` value instead of interning a new `Symbol`. That way `push_doc_lines` can still compute per-line source-map offsets, lines dropped by the vertical trim keep advancing `ast_offset` by `source_len + "\n"`, so surviving lines map back to their true source bytes. Added a citation to the rustc revision (`16a623ad`) in the doc comment.
- Restricted the update to `CommentKind::Block`. Enabling the `Line` branch (which strips a uniform leading-space prefix from `///` comments) would change hover rendering more broadly than the issue calls for. Happy to do that in a follow-up PR if you'd like, but wanted to keep this one scoped to #1759.
- Dropped my earlier `block_star_prefix`; the horizontal-trim logic in the patch supersedes it and also handles the `* \t` prefix cases my version rejected.
- One test case I initially wrote to exercise the vertical trim (`/**\n****\n * foo\n * bar\n****\n */`) doesn't actually hit that path, after `doc_comment()` strips the `/**` prefix, the fence lines aren't at index 0 or N-1, so rustc's `get_vertical_trim` wouldn't fire on it either. Left the vertical-trim code in place for spec fidelity but didn't ship a test that can't reach it. Let me know if you'd prefer I remove that branch as dead code in this crate's usage.

All the tests from the PR body still pass (`cargo test -p hir-def`, `cargo test -p ide -- hover doc_links`, `cargo test -p ide -- syntax_highlighting`), plus `cargo fmt --check` and `cargo clippy` are clean.

ChayimFriedman2 (changes requested):
This is more complicated that it needs to be. `get_vertical_trim()` and `get_horizontal_trim()` you can copy verbatim. Current `push_doc_lines()` should be renamed to `push_doc_line()` and handle only one line, and a new `push_doc_lines()` should be created, that is equivalent to rustc's `beautify_doc_string()` except that if `!data_s.contains('\n')` it just calls `push_doc_line()`, and if not, instead of collecting into `Vec<&str>` it'll collect into `Vec<(&str, TextSize)>` where the `TextSize` is the offset from the string start (to get it you can use `data_s.substr_range(line).unwrap().start`). Then at the end instead of `join("\n")` you iterate over `lines` and call `push_doc_line()`.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22901/014d54b685ccc00f0774d755d1a4c889f7e8e512..877fbfa14aa309b526dd88a66cf653541688ad65)*

kivancgnlp:
Refactored in [b393f3d](https://github.com/rust-lang/rust-analyzer/commit/b393f3d4bc) along the shape you sketched:

- `push_doc_lines` is now the single-line pusher, renamed to `push_doc_line`.
- New `push_doc_lines` mirrors `beautify_doc_string`: `!doc.contains('\n')` calls `push_doc_line` directly; otherwise it builds `Vec<(&str, TextSize)>`, runs `get_vertical_trim` / `get_horizontal_trim`, applies the horizontal strip (plus the additional `*` strip when it's block decoration), then iterates the survivors and calls `push_doc_line(line, ast_offset + line_offset, indent)`.
- The trim helpers are near-verbatim copies of rustc's. Two adaptations I want to flag:
  - Signatures take `&[(&str, TextSize)]` instead of `&[&str]`, the offset is ignored inside the helpers, but keeping it paired avoids projecting a temporary `Vec<&str>` at each call site. Happy to switch to `&[&str]` + a projection if you'd prefer stricter parity with upstream; it makes future copy-paste updates from `rustc_ast` trivially mechanical at the cost of one small allocation per call.
  - `get_vertical_trim` gets one extra guard: the empty-first-line case is skipped. rustc feeds this helper the output of `data_s.lines()`; I feed it `doc.split('\n')` so I can track per-line byte offsets, and unlike `.lines()`, `.split('\n')` can produce a leading empty entry that `chars().all(...)` matches vacuously. The deviation is called out inline.
- Byte offsets are computed manually with a running cursor rather than `str::substr_range`, since that's not stable at the workspace's MSRV (1.95).

One test expectation moved (`block_doc_comment_stars`). For the `/**\n * foo\n *\n *   * bullet\n *bar\n */` input, my earlier `block_star_prefix` version left `" *bar"` intact when the line didn't match the `* ` / `**` / `*` guard, which pinned the block's min indent at 1 and let downstream indent-stripping shift the surrounding lines left. Rustc's rule strips the horizontal `" "` from every line unconditionally, so `*bar` becomes `*bar` (no leading space), min indent drops to 0, and the neighbouring lines keep their leading whitespace. That matches rustdoc's rendering for the same input, which is what we want. Comment on the test spells this out.

All the tests from the PR body still pass (`cargo test -p hir-def`, `cargo test -p ide -- hover doc_links`, `cargo test -p ide -- syntax_highlighting`) plus `cargo fmt --check` and `cargo clippy` are clean.

crates/hir-def/src/attrs/docs.rs
```diff
@@ -218,31 +219,92 @@ impl Docs {
...
+    ) {
+        if !doc.contains('\n') {
+            self.push_doc_line(doc, ast_offset, indent);
+            return;
+        }
+
+        // Each entry is `(line, offset_from_doc_start)`. The offset stays in sync with the
+        // string as we strip its prefix, so the caller can add it to `ast_offset` for the
+        // source map. Computed manually rather than via `str::substr_range` to stay compatible
+        // with the workspace's MSRV.
```
ChayimFriedman2: We can upgrade the MSRV (this is quite new and we prefer not to do that so quickly but I prefer that over duplicating the code).

crates/hir-def/src/attrs/docs.rs
```diff
@@ -218,31 +219,92 @@ impl Docs {
...
+        let mut lines: Vec<(&str, TextSize)> = Vec::new();
+        let mut cursor = TextSize::new(0);
         for line in doc.split('\n') {
```
ChayimFriedman2: And crucially, rustc uses `lines()` which has a different behavior wrt. `\r`, which is why I said to use `substr_range()`.

crates/hir-def/src/attrs/docs.rs
```diff
@@ -218,31 +219,92 @@ impl Docs {
...
+    /// this port carries each line's byte offset relative to `doc`'s start so
+    /// [`Docs::push_doc_line`] can record accurate per-line source-map offsets.
+    ///
+    /// [`beautify_doc_string`]: https://github.com/rust-lang/rust/blob/16a623ad672a92409b5c04beb303583c6cf72a7e/compiler/rustc_ast/src/util/comments.rs#L37
+    fn push_doc_lines(
+        &mut self,
+        doc: &str,
+        ast_offset: Option<TextSize>,
+        indent: &mut usize,
+        shape: Option<ast::CommentShape>,
```
ChayimFriedman2: rustc passes `CommentKind::Line` for desugared comments, please do the same here.

ChayimFriedman2 (changes requested):
Please also squash.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22901/014d54b685ccc00f0774d755d1a4c889f7e8e512..ae0774e2a113af64ea6fd2b0d45a27eba95bb419)*

kivancgnlp:
Addressed the three inline comments plus the squash in [7f44856](https://github.com/rust-lang/rust-analyzer/commit/7f4485696) (rebase-amended into the single squashed commit).

- **MSRV**: bumped `rust-version` from 1.95 to 1.96 in the workspace `Cargo.toml`. Called out in the commit message so it's visible in the log.
- **`.lines()` + `substr_range`**: switched from the manual `.split('\n')` + running-cursor loop to `doc.lines()` with `doc.substr_range(line).unwrap().start` for the byte offset, matches rustc exactly, including `\r\n` handling.
- **`CommentShape::Line` for desugared docs**: `push_doc_lines` no longer takes `Option<CommentShape>`. `extend_with_doc_str` (doc attributes) and `extend_with_unmapped_doc_str` (macro-expanded docs) now pass `CommentShape::Line`, matching rustc's `CommentKind::Line` for those cases. Inline comment cites the parity.
- **`get_vertical_trim` / `get_horizontal_trim`** are now byte-for-byte with rustc, taking `&[&str]`. The caller projects a temporary `Vec<&str>` view around the `Vec<(&str, TextSize)>` before each helper call. The `!line.is_empty()` guard I'd added on the first-line check in the previous round is gone, no longer needed with `.lines()`.
- **Test expectations moved**: `.lines()` no longer emits a leading empty entry for the `\n` right after `/**`, so `block_doc_comment_source_map` now expects `"foo\nbar\n"` (offsets shift by 1) and `block_doc_comment_stars` drops the leading blank row. Both match rustdoc's actual rendering more faithfully than the previous port did. Comments on the tests explain why.
- **Squashed** the whole thing into a single commit and force-pushed.

Local checks: `cargo test -p hir-def` (502 pass), `cargo test -p ide -- hover doc_links` (308 pass), `cargo test -p ide -- syntax_highlighting` (38 pass), `cargo fmt --check`, `cargo clippy`, and `RUSTDOCFLAGS="-D warnings" cargo doc -p hir-def --no-deps --document-private-items` all clean.

crates/hir-def/src/attrs/docs.rs
```diff
@@ -218,29 +219,95 @@ impl Docs {
         offset_in_ast: TextSize,
         indent: &mut usize,
     ) {
-        self.push_doc_lines(doc, Some(offset_in_ast), indent);
+        // Desugared `#[doc = "..."]` strings and macro-expanded docs behave like line comments
+        // in rustc's `beautify_doc_string`, so pass `CommentShape::Line` here (matches rustc
+        // passing `CommentKind::Line` for the desugared case).
+        self.push_doc_lines(doc, Some(offset_in_ast), indent, ast::CommentShape::Line);
```
ChayimFriedman2: This should take the `CommentShape` as a parameter.

crates/hir-def/src/attrs/docs.rs
```diff
@@ -201,7 +201,8 @@ impl Docs {
 
     fn extend_with_doc_comment(&mut self, comment: ast::Comment, indent: &mut usize) {
         let Some((doc, offset)) = comment.doc_comment() else { return };
-        self.extend_with_doc_str(doc, comment.syntax().text_range().start() + offset, indent);
+        let offset = comment.syntax().text_range().start() + offset;
+        self.push_doc_lines(doc, Some(offset), indent, comment.kind().shape);
```
ChayimFriedman2: And then pass it here, avoiding the switch to `push_doc_lines()`.

crates/hir-def/src/attrs/docs.rs
```diff
@@ -218,29 +219,95 @@ impl Docs {
...
+        shape: ast::CommentShape,
+    ) {
+        if !doc.contains('\n') {
+            self.push_doc_line(doc, ast_offset, indent);
+            return;
+        }
+
+        // Use `str::lines()` (matching rustc) so `\r\n` line endings behave correctly, and
+        // `str::substr_range` (matching rustc) to recover each line's byte offset from `doc`'s
+        // start. Both were stabilized by our workspace MSRV.
```
ChayimFriedman2: Redundant comment, please remove.

ChayimFriedman2 (changes requested):
Some final nits (and remember to squash).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22901/8f07bcc8cb2ec2f3702fdf02fc66bed01744944a..7f44856965474d493fbfb8545bc269c61acb9cee)*

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22901/8f07bcc8cb2ec2f3702fdf02fc66bed01744944a..93d98164ab4255035502e9e06e2b86dde06f0141)*

kivancgnlp:
Two things landed on top of [7f44856965](https://github.com/rust-lang/rust-analyzer/commit/7f44856965), squashed into [93d98164a](https://github.com/rust-lang/rust-analyzer/commit/93d98164ab):

**CI failures on the MSRV bump.** Reverted to `rust-version = "1.95"`.
- Clippy failed with `incompatible_msrv`. I had the stability wrong: `str::substr_range` is stable since **1.98**, not 1.96. My local `rustc 1.98.0` didn't flag it, but clippy reads the declared workspace MSRV and correctly errored. Sorry for the wrong number.
- Miri failed with `rustc 1.95.0-nightly is not supported`, the `miri` job pins a nightly whose version prefix is `1.95.0-nightly`, so any MSRV bump breaks it until the toolchain pin is bumped in a separate PR.

Given "prefer not to do that so quickly," I don't think escalating to a 1.98 bump plus a miri-toolchain bump is right for this PR. Per-line offsets are now computed by pointer arithmetic (`line.as_ptr() as usize - doc.as_ptr() as usize`), same result as `substr_range`, no MSRV cost. `.lines()` and correct `\r\n` handling stay.

**Nits.** `extend_with_doc_str` now takes `CommentShape`, and `extend_with_doc_comment` passes `comment.kind().shape` through it instead of calling `push_doc_lines` directly. `extend_with_doc_attr` also routes through `extend_with_doc_str` with `CommentShape::Line`. Removed the redundant comment on the `.lines()` block.

Single commit, all tests + fmt + clippy + rustdoc green locally.

ChayimFriedman2:
I know it was stable in 1.98 (in fact I didn't realize that's not what you said). We keep policy of latest-stable MSRV, but we try to not update so quickly. However like I said, here I think it's justified.

The Miri CI should be unpinned anyway.

kivancgnlp:
Thanks for the context, good to know. Happy to leave this on the pointer-arithmetic version so the PR doesn't take a dependency on the MSRV policy or the miri pin moving.

ChayimFriedman2:
I did that myself, https://github.com/rust-lang/rust-analyzer/pull/23239.

We suspect you're writing your comments via an AI. Please note that per [our AI policy](https://github.com/rust-lang/rust-analyzer/blob/master/AI_POLICY.md), this is forbidden, and AI usage must be disclosed.

boozook:
How to disable this breaking behavior?
Relative: https://github.com/rust-lang/rust-analyzer/issues/1759#issuecomment-3321031847

ChayimFriedman2:
@boozook You can't. This is a bugfix and we won't reenable spacebar heating ([reference](https://xkcd.com/1172/)).

boozook:
This isn't bugfix. This is rust spec disrespect and an error. It could be fix it is java.

ChayimFriedman2:
The Rust spec is the implementation. The Reference is explicitly non-normative. In this case, the implementation is rustdoc.

We're not going to argue over it. If you can convince the rustdoc teams to change their behavior, we will follow suit. Until then, discussion here is not useful.

ChayimFriedman2:
Also, please remember that this project is open source and developed by volunteers. Even if you disagree with the maintainers' decisions, respect is warranted. And of course, remember to always follow [the Code of Conduct](https://rust-lang.org/policies/code-of-conduct/).

PR #22903, opened 2026-07-24
fix: Discover protocol should only parse stdout

Wilfred:
Currently r-a expects JSON on both stdout and stderr from the discover command. This means that the discover command can't log any warnings, and any accidental stderr logs breaks discovery entirely.

Instead, only parse stdout, and update `JsonLinesParser` to require implementors to be explicit.


Wilfred:
cc @davidbarsky I'm pretty sure the old behaviour was entirely accidental due to the old trait API, but let me know what you think.

crates/rust-analyzer/src/discover.rs
```diff
@@ -136,6 +136,11 @@ impl JsonLinesParser<DiscoverProjectMessage> for DiscoverProjectParser {
     fn from_eof(&self) -> Option<DiscoverProjectMessage> {
         None
     }
+
+    fn from_stderr_line(&self, line: &str, _error: &mut String) -> Option<DiscoverProjectMessage> {
+        tracing::warn!(%line, "discover command stderr");
```
ChayimFriedman2: I'm on the edge about this. Is every stderr line really a warning? I see three options:

 1. Do nothing, leave it as `warn!()`. I don't like this.
 2. Downgrade to `info!()` or even `debug!()`.
 3. Expect JSON with only the level and the message.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22903/ae631b9fffa2e47b7b8e0caa5effdc574472cfdd..ed5af79dcb5180cd346c08c56b513b3960d5f81d)*

Wilfred: stderr output is definitely rare, or we'd have hit this issue sooner! :)

I picked `warn!` because I thought it was nice to see the stderr output by default. That said, `info!` also seems fine: it solves my issue with unexpected logging from the rust-project binary breaking people's rust-analyzer setup.

Switched to `info!` and amended the commit.

davidbarsky:
> cc @davidbarsky I'm pretty sure the old behaviour was entirely accidental due to the old trait API, but let me know what you think.

I _think_ the intention was have *all* errors go over JSON, but unlike Chayim, I don't think I have concerns about using `stderr` for error collection. 

(side note: I got some rust-project changes, if you got a moment: https://github.com/facebook/buck2/pull/1398)

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22903/ae631b9fffa2e47b7b8e0caa5effdc574472cfdd..ae9fa8c52faf81e1d3ba141c82c356c3d4c1d6a4)*

ChayimFriedman2:
You need to `UPDATE_EXPECT=1 cargo test`.

Wilfred:
Oops, fixed.

PR #22905, opened 2026-07-24
fix: attach db on worker threads in parallel analysis-stats inference

tilladam:
The next-generation trait solver stores the `HirDatabase` in a thread-local, set via `hir::attach_db`. `analysis-stats` attaches it once on the main thread, but the `--parallel` inference path runs `InferenceResult::of` on rayon worker threads that never attach it, so every parallel run panicked with "Try to use attached db, but not db is attached" at `next_solver/interner.rs`. Wrap the per-body inference in `hir::attach_db` so each worker has the database attached.


Veykril:
Damn have we really not run that one since the migration? 😅 Thanks!

Veykril (approved):

PR #22906, opened 2026-07-25
internal: remove dead `#[allow]` attributes

weiscadence:
Each of these 20 `#[allow(..)]` attributes suppresses a lint that no longer
fires at that site, so the attribute currently has no effect. The change is
purely subtractive — 20 deletions, no statement, signature or behaviour change.

### How they were found

1. Rewrote every `#[allow(..)]` under `crates/` to `#[expect(..)]` and collected
   `unfulfilled_lint_expectations`.
2. Discarded every diagnostic whose span came from a macro expansion. Those are
   **not** dead: the suppression is still needed by other expansions of the same
   macro. This was the large majority (134 of 153 sites).
3. Confirmed each remaining candidate by actually **deleting** the attribute and
   rebuilding.

Step 3 was not redundant. One candidate that step 1 reported as unfulfilled —
`#[allow(unused_mut)]` in `crates/hir-def/src/expr_store/lower/path.rs` — makes
`unused_mut` fire when removed, because that binding is only mutated under
`#[cfg(test)]`. It is left untouched here. An unfulfilled expectation is not by
itself proof that an `allow` is dead.

### Verification

- `cargo check --workspace --all-targets` — clean, no diagnostics.
- Same under CI's `RUSTFLAGS="-W unreachable-pub --cfg no_salsa_async_drops"`.
- `Cargo.lock` is deliberately untouched. (It does churn locally on a pristine
  checkout too — `salsa-macros` — so that is unrelated to this change.)

### Not covered

I could not build the `in-rust-tree` feature (needs a nightly toolchain), so
these were not checked under it. As far as I can tell CI only builds that
feature for `proc-macro-srv`, `proc-macro-srv-cli` and `proc-macro-api`, none of
which are touched here — but flagging it rather than implying coverage I don't
have.

Happy to split this per-crate if that is easier to review.

### DIFF
--- a/crates/hir-expand/src/lib.rs
+++ b/crates/hir-expand/src/lib.rs
@@ -531,6 +531,12 @@ impl MacroCallId {
     /// Lowers syntactic macro call to a token tree representation. That's a firewall
     /// query, only typing in the macro call itself changes the returned
     /// subtree.
+    ///
+    /// Calling this is incorrect, call [`Self::macro_arg_considering_derives`] instead.
+    /// This used to be enforced by `#[deprecated]`, which is no longer expressible here:
+    /// `#[salsa::tracked]` expands the method into a trait impl, and rustc rejects
+    /// `#[deprecated]` on trait methods in impl blocks. The `#[allow(deprecated)]` at the
+    /// call sites is kept for the same reason, so please leave it in place.
     #[salsa::tracked(returns(ref))]
     fn macro_arg(self, db: &dyn SourceDatabase) -> MacroArgResult {
         let loc = self.loc(db);
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -108,12 +108,7 @@ use crate::{
     utils::TargetFeatureIsSafeInTarget,
 };
 
-// This lint has a false positive here. See the link below for details.
-//
-// https://github.com/rust-lang/rust/issues/57411
-#[allow(unreachable_pub)]
 pub use coerce::could_coerce;
-#[allow(unreachable_pub)]
 pub use unify::{could_unify, could_unify_deeply};
 
 use cast::{CastCheck, CastError};
@@ -701,7 +696,6 @@ pub enum PointerCast {
     /// Go from a mut raw pointer to a const raw pointer.
     MutToConstPointer,
 
-    #[allow(dead_code)]
     /// Go from `*const [T; N]` to `*const T`
     ArrayToPointer,
 
--- a/crates/hir-expand/src/fixup.rs
+++ b/crates/hir-expand/src/fixup.rs
@@ -343,7 +343,6 @@ pub(crate) fn reverse_fixups(tt: &mut TopSubtree, undo_info: &SyntaxFixupUndoInf
     let top_subtree = tt.top_subtree();
     let open_span = top_subtree.delimiter.open;
     let close_span = top_subtree.delimiter.close;
-    #[allow(deprecated)]
     if never!(
         close_span.anchor.ast_id == FIXUP_DUMMY_AST_ID
             || open_span.anchor.ast_id == FIXUP_DUMMY_AST_ID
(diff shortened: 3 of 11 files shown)


ChayimFriedman2:
Please note that [our AI policy](https://github.com/rust-lang/rust-analyzer/blob/master/AI_POLICY.md) disallows using AI to write GitHub comments, including PR descriptions.

crates/hir-ty/src/infer.rs
```diff
@@ -111,9 +111,7 @@ use crate::{
 // This lint has a false positive here. See the link below for details.
```
ChayimFriedman2: Please remove this comment as well.

crates/hir-expand/src/eager.rs
```diff
@@ -60,7 +60,6 @@ pub fn expand_eager_macro_input(
         ctxt: call_site,
     };
     let arg_id = MacroCallId::new(db, loc);
-    #[allow(deprecated)] // builtin eager macros are never derives
     let (_, _, span) = arg_id.macro_arg(db);
```
ChayimFriedman2, line 64: This is incorrect, the `#[deprecated]` was just lost during the Salsa transition, we should add it back.

crates/hir-expand/src/lib.rs
```diff
@@ -514,7 +514,6 @@ impl MacroCallId {
     /// This is not connected to the database so it does not cache the result. However, the inner [macro_arg] query is
     ///
     /// [macro_arg]: Self::macro_arg
-    #[allow(deprecated)] // we are macro_arg_considering_derives
```
ChayimFriedman2, line 517: Here too.

ChayimFriedman2 (changes requested):
Please switch all other `allow`s to `expect`s, or a comment explaining why they can't be.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22906/ca8fba298406608be4003978f15ecb72a3ddf481..0733e9d1abec250a6451f198c1e8d1e858b40e4f)*

weiscadence:
infer.rs - I removed the comment 

eager.rs + lib.rs - you’re right they aren’t dead but your fix doesn’t compile. #[salsa::tracked] expand the method into a trait impl. I tried both attribute orders and got the same error. So the depreciation wasn’t dropped by oversight in the salsa transition, it stopped being expressable.

rustc rejects it:
“error: #[deprecated] attribute cannot be used on trait methods in impl blocks”

There is a workaround:

rename the tracked query to macro_arg_ and put #[deprecated] on a thin wrapper. The same _-suffix pattern already in cli/flags.rs. I confirmed it compiles and that the warning genuinely fires (stripped the allow back out and the deprecation warning appeared). But it renames a salsa query so do you want that or leave those two allows deleted?

Allow-expect - for the whole codebase or just the touched files? 134 of 153 allow sites sit inside macro expansions where expect breaks because other expansions of the same macro still need the suppression.

ChayimFriedman2 (approved):
> eager.rs + lib.rs - you’re right they aren’t dead but your fix doesn’t compile. #[salsa::tracked] expand the method into a trait impl. I tried both attribute orders and got the same error. So the depreciation wasn’t dropped by oversight in the salsa transition, it stopped being expressable.

Ugh. Okay then I guess a comment above the function to not remove it will be enough, you don't need to insert an inner function.

> Allow-expect - for the whole codebase or just the touched files? 134 of 153 allow sites sit inside macro expansions where expect breaks because other expansions of the same macro still need the suppression.

For the whole codebases, but only outside of macros. You can do that in a separate PR if you want.

Please also squash.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22906/ca8fba298406608be4003978f15ecb72a3ddf481..456a5a2893d249f2757cde747249e68ef23b3e61)*

ChayimFriedman2:
Do you intend to replace with `#[expect]` in this PR?

PR #22907, opened 2026-07-25
proc-macro-api: Implement client side timeout handling

Veykril:
Was meaning to do this but never got around to it.

Written in part with gpt-5.6 and Fable


ChayimFriedman2:
Why, actually? Assuming we kill the server when we exit and cancels ongoing requests when cancelling LSP requests, what will a timeout give us? Especially if then we need retrying etc..

PR #22910, opened 2026-07-25
perf: avoid re-running the trait-solver fast path on unchanged obligations

tilladam:
`FulfillmentCtxt::try_evaluate_obligations` re-processes every pending ambiguous obligation on every call. For obligations resolved by `compute_goal_fast_path`, the `Certainty::Maybe` branch discarded the obligation's `GoalStalledOn` (re-registering with `None`), which meant every subsequent call fully re-ran the fast path's predicate resolution (`shallow_resolve` + `is_trivially_wf`) instead of being able to skip it, even though the real solver's own internal stall-check exists precisely to avoid this.

I measured the speedup:
  - rust-analyzer: whole-workspace inference was about 1.3–1.4× faster (24.4s → 17–18s); the worst prefill body was 1.79× faster (7.79s → 4.35s).
  - Slint: the worst generate_enums_pyi body was roughly 4.7× faster (~17s → 3.64s).

The bigger problem is that frequent fulfillment calls × rescanning every pending obligation = O(n²), but I'll address that in a separate PR, to keep this minimal.

Tokens were burned in the analysis, profiling and implementation of this PR.

### DIFF
--- a/crates/hir-ty/src/next_solver/fulfill.rs
+++ b/crates/hir-ty/src/next_solver/fulfill.rs
@@ -8,15 +8,17 @@ use rustc_next_trait_solver::{
     solve::{GoalEvaluation, GoalStalledOn, HasChanged, SolverDelegateEvalExt},
 };
 use rustc_type_ir::{
-    Interner, TypeSuperVisitable, TypeVisitable, TypeVisitableExt, TypeVisitor,
-    inherent::IntoKind,
-    solve::{Certainty, NoSolution},
+    InferConst, InferCtxtLike, InferTy, Interner, TyVid, TypeSuperVisitable, TypeVisitable,
+    TypeVisitableExt, TypeVisitor,
+    inherent::{IntoKind, OpaqueTypeStorageEntries as _},
+    solve::{Certainty, Goal, NoSolution},
 };
 
 use crate::{
     Span,
     next_solver::{
-        DbInterner, SolverContext, SolverDefId, Ty, TyKind, TypingMode,
+        Const, ConstKind, DbInterner, GenericArg, Predicate, SolverContext, SolverDefId, Ty,
+        TyKind, TypingMode,
         infer::{
             InferCtxt,
             traits::{PredicateObligation, PredicateObligations},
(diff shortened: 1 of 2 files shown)


ChayimFriedman2:
As I said in https://github.com/rust-lang/rust-analyzer/pull/22911#issuecomment-5079934167:

> This code is copied from rustc and we will absolutely not change it from what rustc does. If rustc was already changed and we haven't followed, please point at the rustc code. If not, the only way for this to be accepted is to first suggest this to rust-lang/rust.

ChayimFriedman2:
Also note that the rustc people may not be happy if someone that may not understand the type system well suggest LLM changes to the solver, and also, if you used a LLM to author the PR description, that [we disallow that](https://github.com/rust-lang/rust-analyzer/blob/master/AI_POLICY.md).

tilladam:
> As I said in [#22911 (comment)](https://github.com/rust-lang/rust-analyzer/pull/22911#issuecomment-5079934167):
> 
> > This code is copied from rustc and we will absolutely not change it from what rustc does. If rustc was already changed and we haven't followed, please point at the rustc code. If not, the only way for this to be accepted is to first suggest this to rust-lang/rust.

Turns out this issue was fixed similarly in https://github.com/rust-lang/rust/pull/158249 in rustc. I guess we can drop this and wait for that to be pulled in?

tilladam:
> Also note that the rustc people may not be happy if someone that may not understand the type system well suggest LLM changes to the solver, and also, if you used a LLM to author the PR description, that [we disallow that](https://github.com/rust-lang/rust-analyzer/blob/master/AI_POLICY.md).

I'm aware of the policy and was not attempting to circumvent it. I did copy and paste the benchmark numbers, assuming that was acceptable.

ChayimFriedman2:
> Turns out this issue was fixed similarly in [rust-lang/rust#158249](https://github.com/rust-lang/rust/pull/158249) in rustc. I guess we can drop this and wait for that to be pulled in?

Looks so, yes. This doesn't seem to require additional changes on our side.

PR #22911, opened 2026-07-25
Perf: fulfillment wake by stalled var

tilladam:
Problem: try_evaluate_obligations drained/re-touched the whole pending set per call; body inference calls it ~O(n) times -> O(n²) per body. This hits large generated bodies particularly as they exist in both rust-analyzer and slint (and I'm sure elsewhere also).

UndoLogs::in_snapshot for InferCtxtUndoLogs deliberately returns true so ena reports all mutations (its gates skip push outside snapshots). Real undo entries stay gated internally, an inherent in_snapshot provides the real check. This could/should later be replaced by a small upstream ena API (an "always observe" hook), if you guys agree this is the right approach.

Impact measurements (macOS M5-series, release):

synthetic 8k-stmt body (one fn)  23.6s (down from 86.8s after rust-lang/rust-analyzer#22910) -> 0.38s

RA self-analysis, total inference: 24.4s -> 14.0s (across this and rust-lang/rust-analyzer#22910)
Slint generate_enums_pyi body:  3.64s ->  0.49s
Slint workspace inference: 40.6s → 27.4s

Correctness: full self-analysis differential (unknown types, type mismatches, pattern variants, MIR failed bodies, failed const evals) identical before/after, peak RSS unchanged, 1018/0 hir-ty tests, clippy clean.

This prepares better parallelization because making those bodies near-linear:

  - shortens that serial critical path
  - reduces workload imbalance between threads
  - raises the maximum useful speedup from future body-level parallelization

LLM help was used in the analysis, profiling and implementation of this change.


ChayimFriedman2:
This code is copied from rustc and we will absolutely not change it from what rustc does. If rustc was already changed and we haven't followed, please point at the rustc code. If not, the only way for this to be accepted is to first suggest this to rust-lang/rust.

tilladam:
> This code is copied from rustc and we will absolutely not change it from what rustc does. If rustc was already changed and we haven't followed, please point at the rustc code. If not, the only way for this to be accepted is to first suggest this to rust-lang/rust.

Fair enough. It's still a problem in rustc and will hit them once the new solver stabilizes. I'll attempt to find someone to point the quadratic behavior out to on zulip or via an issue then.

tilladam:
Filed against rustc here: https://github.com/rust-lang/rust/issues/159933.

PR #22914, opened 2026-07-25
internal: Cleanup code for old Rust versions

ChayimFriedman2:
That we no longer support after https://github.com/rust-lang/rust-analyzer/pull/22876.


Veykril (approved):

Veykril:
That bump might've been a bit too aggressive by us as 1.94 is only from march.

ChayimFriedman2:
Maybe, but I'm actually quite happy about it. It lets us cleanup a lot of code. Since Rust 1.94.0 was released more than 4 months have passed. If you have a legitimate reason to use such old version (and remember we only officially support the last version), you can very well downgrade your r-a. I believe most people using this toolchain just didn't see a reason to upgrade, so we gave them it :smile_cat:

Veykril:
Yea, in this case the code we get rid of is nice. I am fine with merging either way (since otherwise we'd need to revert some PRs anyways), just something to note down.

PR #22918, opened 2026-07-26
fix: prefer `alloc` over `std` paths when `preferNoStd` is set

### LINKED ISSUE #19053: imports.preferNoStd for alloc ignored when "extern crate std" is present
When a crate contains `extern crate std;` rust-analyzer always want to import e.g. `String` and other alloc imports from `std`, even when preferNoStd is set and/or the crate is `#![no_std]`. For imports from core it seems to suggest both, but with core first.

I discovered this while working on a bevy crate that has `extern crate std` behind a feature flag, but it seems to be a problem regardless of the feature flag.

**rust-analyzer version**: rust-analyzer version: 0.4.2281-standalone 

**rustc version**: rustc 1.84.0 (9fc6b4312 2025-01-07)

**editor or extension**: VScode v0.4.2281 (pre-release) 

**relevant settings**: 
`settings.json`:
```
{
  "rust-analyzer.imports.preferNoStd": true
}
```

```rust
#![no_std]

extern crate alloc;

extern crate std;

// ctrl+. on string only suggests std::string::String
pub fn foo() -> String {}
```

shulaoda:
Closes rust-lang/rust-analyzer#19053

With `preferNoStd` on, auto-import still suggests `std::string::String` instead of `alloc::string::String` as soon as the crate contains `extern crate std;`. The setting is documented as preferring imports of the core and alloc crate over std, so the code doesn't do what the docs promise.

### DIFF
--- a/crates/hir-def/src/find_path.rs
+++ b/crates/hir-def/src/find_path.rs
@@ -407,6 +407,10 @@ fn find_in_sysroot<'db>(
         if matches!(best_choice, Some(Choice { stability: Stable, .. })) {
             return;
         }
+        search(LangCrateOrigin::Alloc, best_choice);
+        if matches!(best_choice, Some(Choice { stability: Stable, .. })) {
+            return;
+        }
         search(LangCrateOrigin::Std, best_choice);
         if matches!(best_choice, Some(Choice { stability: Stable, .. })) {
             return;
@@ -417,6 +421,10 @@ fn find_in_sysroot<'db>(
             return;
         }
         search(LangCrateOrigin::Core, best_choice);
+        if matches!(best_choice, Some(Choice { stability: Stable, .. })) {
+            return;
+        }
+        search(LangCrateOrigin::Alloc, best_choice);
         if matches!(best_choice, Some(Choice { stability: Stable, .. })) {
             return;
         }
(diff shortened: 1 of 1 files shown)


crates/hir-def/src/find_path.rs
```diff
@@ -407,6 +407,10 @@ fn find_in_sysroot<'db>(
         if matches!(best_choice, Some(Choice { stability: Stable, .. })) {
             return;
         }
+        search(LangCrateOrigin::Alloc, best_choice);
```
ChayimFriedman2, line 410: This should also be in the `else` branch.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22918/857dec62c43dd8df13a8f628ef5b9d59bb28b0c0..1bf5bd928544ae8dfbc2e4dd4e21a3d2dc590811)*

ChayimFriedman2 (changes requested):

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22918/b59f59fa685a3e575ffc6beebf0e98a7979b88dc..902eb9e456bec53b101bcd734b15bcf46694f9b9)*

PR #22919, opened 2026-07-26
fix: don't offer `replace_qualified_name_with_use` on an unqualified path

### LINKED ISSUE #22691: Code action `Replace qualified path with use` resolves incorrectly when item name matches crate name
**rust-analyzer version**: rust-analyzer version: 0.3.2955

**rustc version**: rustc 1.96.1 

When using the `Replace qualified path with use` code action, if the crate name and an item directly under that crate share the same name, this action inserts `self` into the use statement instead of importing the specific item.

I think it shoud suggest auto import or qualify path.

"Replace qualified path with use"
```
.
├── Cargo.toml
├── src
│   └── main.rs
└── sub
    ├── Cargo.toml
    └── src
        └── lib.rs
```

```rust
// main.rs
use sub::S; // before
// expect: use sub::{sub, S};
// actual: use sub::{self, S};


fn main() {
    sub();
 // ^^^ "Replace qualified path with use"
}

```
```toml
[package]
name = "ra-bug-repro"
edition = "2024"

[dependencies]
sub = { path = "./sub" }
```
```rust
// sub/lib.rs
pub struct S;
pub fn sub() {}
```

```toml
# sub/Cargo.toml
[package]
name = "sub"
[dependencies]
```

Another Example

```rust

use smallvec::SmallVec;


fn main() {

smallvec!();

}

```

shulaoda:
Closes rust-lang/rust-analyzer#22691

When a crate and an item inside it share a name, `sub()` resolves to the crate in the type namespace, so the assist is offered on a path that has no qualifier at all. It then imports the crate itself (`use sub;`) and leaves the call untouched, which is a no-op edit on code that does not compile either way. There is nothing to replace here, so the assist should not be offered.

### DIFF
--- a/crates/ide-assists/src/handlers/replace_qualified_name_with_use.rs
+++ b/crates/ide-assists/src/handlers/replace_qualified_name_with_use.rs
@@ -40,6 +40,12 @@ pub(crate) fn replace_qualified_name_with_use(
 
     let original_path = target_path(ctx, original_path)?;
 
+    // There is no qualifier to replace, so there is nothing for this assist to do
+    if original_path.qualifier().is_none() {
+        cov_mark::hit!(not_applicable_for_unqualified_path);
+        return None;
+    }
+
     // then search for an import for the first path segment of what we want to replace
     // that way it is less likely that we import the item from a different location due re-exports
     let module = match ctx.sema.resolve_path(&original_path.first_qualifier_or_self())? {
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22919/857dec62c43dd8df13a8f628ef5b9d59bb28b0c0..b546cf0012a000937c0981023b43e66fe2bad6e9)*

PR #22927, opened 2026-07-27
feat: implement lifetime elision

dfireBird:


dfireBird:
Not sure why this fails in CI and passes in my system?

ShoyuVanilla:
> Not sure why this fails in CI and passes in my system?

https://github.com/rust-lang/rust-analyzer/blob/85e85df195c58fa2c887d4fc09efec8250d6fcdb/docs/book/src/contributing/README.md?plain=1#L67

You need to set an extra env var to run slow tests

dfireBird:
Oh I didn't know, we ran slow tests. My bad :sweat_smile:

crates/hir-def/src/expr_store/pretty.rs
```diff
@@ -1324,7 +1329,9 @@ impl Printer<'_> {
             TypeRef::Fn(fn_) => {
                 let ((_, return_type), args) =
                     fn_.params.split_last().expect("TypeRef::Fn is missing return type");
-                if let Some(binder) = &fn_.binder {
+                if let Some(binder) = &fn_.binder
+                    && !binder.is_empty()
```
Veykril: nit, lets drop the `is_empty` check. Ideally we'd just lower to `None` when binder is empty

crates/hir-def/src/lib.rs
```diff
@@ -713,12 +713,9 @@ pub struct LifetimeParamId {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub struct HrtbLifetimeParamId {
-    pub scope: GenericDefId,
-    pub local_id: usize,
-}
+pub struct HrtbLifetimeParamId(pub usize);
```
Veykril:
```suggestion
pub struct HrtbLifetimeParamId(pub u32);
```
We wont ever have that many lifetimes, so being conservative for memory usage won't hurt here

crates/hir-def/src/resolver.rs
```diff
@@ -550,6 +550,7 @@ impl<'db> Resolver<'db> {
             LifetimeRef::Param(lifetime_param_id) => {
                 Some(LifetimeNs::LifetimeParam(*lifetime_param_id))
             }
+            LifetimeRef::HrtbParam(_) => None, // this should be unreachable, should we panic?
```
Veykril:
```suggestion
            LifetimeRef::HrtbParam(_) => {
                never!();
                None
            }
```

crates/hir-ty/src/display.rs
```diff
@@ -2521,7 +2522,9 @@ impl<'db> HirDisplayWithExpressionStore<'db> for TypeRefId {
                 write!(f, "]")?;
             }
             TypeRef::Fn(fn_) => {
-                if let Some(binder) = &fn_.binder {
+                if let Some(binder) = &fn_.binder
+                    && !binder.is_empty()
```
Veykril: same here

crates/ide-completion/src/context/tests.rs
```diff
@@ -154,7 +154,7 @@ fn expected_type_fn_param_deref() {
 fn foo() { bar(*$0); }
 fn bar(x: &u32) {}
 "#,
-        expect!["ty: &'_ &'_ u32, name: x"],
+        expect!["ty: &'_ &'<erased> u32, name: x"],
```
Veykril: we probably need to check our rendering now, we shouldn't show erased lifetimes to the user

crates/ide-completion/src/render.rs
```diff
@@ -2672,8 +2672,8 @@ struct WorldSnapshot { _f: () };
 fn go(world: &WorldSnapshot) { go(w$0) }
 "#,
             expect![[r#"
-                lc world &WorldSnapshot [type+name+local]
-                ex world  [type]
+                lc world &WorldSnapshot [type_could_unify+name+local]
+                ex world  [type_could_unify]
```
Veykril, line 2681: Interesting, so we lose a lot of type equality now which I guess makes sense. Worth to keep in mind if we want t o have something like type-equal modulo lifetimes

crates/ide/src/signature_help.rs
Veykril, line 1: We should definitely not render `'_` in the generic param list 🙈 if we can't yet render the elided lifetimes nicely here we should probably hide them wholesale

crates/hir-def/src/expr_store/lower/generics.rs
```diff
@@ -304,4 +330,34 @@ impl GenericParamsCollector {
             lifetime.bound_type = LifetimeBoundType::LateBound
         }
     }
+
+    fn lower_elided_lifetime(
+        lifetimes: &mut Arena<LifetimeParamData>,
+        parent: GenericDefId,
+        is_impl_header: bool,
```
Veykril: Let's make this an enum (r-a has an assist for this if you put the cursor on the parameter here iirc)

Veykril (commented):
Only a partial review so far

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22927/85e85df195c58fa2c887d4fc09efec8250d6fcdb..88e02b366947413007254a7372dae2e6e7623897)*

Veykril:
Very excited for this, I've been wanting this for so long!

crates/ide-completion/src/context/tests.rs
dfireBird: I could make the `RegionKind::ReErased` to display `'_` for now, would that be fine?

crates/hir-def/src/expr_store/pretty.rs
dfireBird: Yeah I guess, I should've kept `None` if the binder is empty after fn ptr lowering instead of this and that makes sense as well.

crates/hir-def/src/expr_store/lower/generics.rs
dfireBird: My naming might be terrible for this, so if you have any concerns, let me know (once I push ofc).

crates/ide/src/signature_help.rs
dfireBird: Would it be fine, if we still showed them in the argument list?

crates/ide/src/hover/tests.rs
```diff
@@ -678,7 +678,7 @@ fn main() { }
             ```
 
             ```rust
-            pub fn foo(_: &Path)
+            pub fn foo<'_>(_: &'_ Path)
```
ChayimFriedman2: IMO this is worse display. I think we should omit `'_` generics, and omit them for references as well. For non-references - I'm not sure.

crates/hir-expand/src/name.rs
```diff
@@ -112,6 +112,11 @@ impl Name {
         }
     }
 
+    #[inline]
+    pub fn anon_lifetime() -> Name {
+        Self::new_text("'_")
```
ChayimFriedman2: This should be a predefined symbol.

crates/hir-def/src/lib.rs
```diff
@@ -1414,6 +1411,21 @@ impl ModuleDefId {
...
+    /// Returns `None` for variants, modules, builtin types, and macros.
+    pub fn as_generic_def_id(self) -> Option<GenericDefId> {
+        match self {
+            ModuleDefId::FunctionId(id) => Some(id.into()),
+            ModuleDefId::AdtId(id) => Some(id.into()),
+            ModuleDefId::ConstId(id) => Some(id.into()),
+            ModuleDefId::StaticId(id) => Some(id.into()),
+            ModuleDefId::TraitId(id) => Some(id.into()),
+            ModuleDefId::TypeAliasId(id) => Some(id.into()),
+            _ => None,
```
ChayimFriedman2: Use exhaustive matching here.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -689,47 +761,84 @@ impl<'db> ExprCollector<'db> {
...
+
                 let ret_ty = inner
                     .ret_type()
                     .and_then(|rt| rt.ty())
-                    .map(|it| self.lower_type_ref(it, impl_trait_lower_fn))
+                    .map(|it| {
+                        self.lower_type_ref(
+                            it,
+                            impl_trait_lower_fn,
+                            &mut Self::elided_lifetime_placeholder_allocator,
```
ChayimFriedman2: This does not look correct. We should first lower the arguments. If there is exactly one HRTB lifetime in them, any elided lifetime in the return type should refer to it. Otherwise, it's an error lifetime (plus a diagnostic). In fact, I'm pretty sure we should remove `LifetimeRef::Placeholder`.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -689,47 +761,84 @@ impl<'db> ExprCollector<'db> {
         &mut self,
         node: ast::Type,
         impl_trait_lower_fn: ImplTraitLowerFn<'_>,
+        lifetime_elision_fn: LifetimeElisionFn<'_>,
```
ChayimFriedman2: I'm not pleased with passing another callback everywhere. I would prefer an enum field saying what to do: an error, a `'static` lifetime (for `const` and `static` types), a specific lifetime, or a new anonymous lifetime. In fact you can look at the old `LifetimeElisionKind`, which implements this (but only for diagnostics) and was copied from rustc.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -1242,53 +1466,60 @@ impl<'db> ExprCollector<'db> {
...
+                    ec.lower_path_type(&path_type, impl_trait_lower_fn, lifetime_elision_fn)
+                        .map(|p| {
+                            let path = ec.alloc_path(p, AstPtr::new(&path_type).upcast());
+                            let binder = ec.for_type_binder.take();
+                            if let Some((binder, _)) = binder
+                                && !binder.is_empty()
+                            {
+                                TypeBound::ForLifetime(binder, path)
+                            } else {
+                                TypeBound::Path(path, m)
```
ChayimFriedman2: Do we even need `TypeBound::Path` now? After all in rustc_type_ir every predicate has a binder. We could just always produce `ForLifetime`.

ChayimFriedman2 (changes requested):
Round 1 review.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22927/85e85df195c58fa2c887d4fc09efec8250d6fcdb..88e02b366947413007254a7372dae2e6e7623897)*

crates/ide/src/hover/tests.rs
Veykril: Agree'd, this should mimic the lint. So rendering `'_` for paths is good, for references we should skip as its always implied.

dfireBird: I think it comes down to what I was saying before in [zulip chat ](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Lifetime.20elision/near/612839426)[#t-compiler/rust-analyzer > Lifetime elision @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Lifetime.20elision/near/612839426), have a way to determine if a `LifetimeRef` is elided or not, provided we have `ExpressionStore`.

crates/hir-def/src/expr_store/lower.rs
dfireBird: I use `LifetimeRef::Placeholder` for return types, so that I can resolve it correctly during `hir-ty` lowering. But I missed to copy from my old WIP of the lifetime elision.

crates/hir-def/src/expr_store/lower.rs
dfireBird: I would anyways need access to lifetime `Arena` in `GenericParamCollector`.

crates/hir-def/src/expr_store/lower.rs
ChayimFriedman2: I see no reason to resolve it during hir-ty lowering and complicate the code with additional variant when we can resolve it during hir lowering instead.

crates/hir-def/src/expr_store/lower.rs
ChayimFriedman2: What do you mean?

crates/ide/src/hover/tests.rs
ChayimFriedman2: You could check the name for `'_`, or add another way to differentiate.

crates/hir-def/src/expr_store/lower.rs
dfireBird: The callback works by pushing a `'_` and using that `Idx` to create the `LifetimeParamId`.

ChayimFriedman2: So? Why can't this be done via an enum field?

crates/ide/src/hover/tests.rs
dfireBird: I could indeed check for `LifetimeRef::Param`, but if it's going to be used in other places, it's going to be a bit of problem.

crates/hir-def/src/expr_store/lower.rs
dfireBird: You mean to say use a mutable reference of the arena inside one of enum variant.

crates/ide/src/hover/tests.rs
ChayimFriedman2: But only elided lifetimes can have name `'_`.

dfireBird: Oh I realized something that I can do, I'll try that.

crates/hir-def/src/expr_store/lower.rs
ChayimFriedman2: I don't understand what you mean.

What I'm proposing is, instead of accepting a callback for what to do when encountering an elided lifetime, keep a field `lifetime_elision_kind` of type `LifetimeElisionKind` in `ExpressionStore`, similar to (in fact replacing) `LifetimeElisionKind` in `TyLoweringContext`, and act according to it.

dfireBird: The primary purpose of it being a callback here is to have a mutable access to the lifetime arena, but otherwise, I can indeed do what you suggest.

ChayimFriedman2: Hmm, I think it will be better if the `enum LifetimeElisionKind` will have a variant `AsLifetimeParam { collector: &mut GenericParamsCollector, ... }`. This does mean introducing another lifetime to `ExprCollector`, but that's not so bad.

dfireBird: Yeah, I can do that. Let me do it along with some other review comments over the weekend..

dfireBird: I think some form of "callback" is necessary, because I can't neither borrow `GenericParmsCollector`, since we are borrowing it mutably for all lowering, especially when lowering functions, `collect_impl_trait` method also borrows it mutably.

Veykril: yea because of that i went with the callback setup for `collect_impl_trait`

dfireBird: Maybe instead of having multiple callback factories, we can have that a state in `ExprCollector` and based on which a the callback (so a single callback factory) returns appropriate `LifetimeRef`?

ChayimFriedman2: You can borrow just the parts you need, like you do for a callback.

dfireBird: I don't think it would work as well, only when lowering functions, because it would be borrowed mutably twice, once when I borrow it for the `Arena` and the other when it borrows for `collect_impl_trait`.

It might work in other places tho.

ChayimFriedman2: Closures borrow only the fields they use since edition 2024. If they wouldn't the code using callbacks wouldn't compile either.

dfireBird: I found a way to work around it, without callbacks, took a lot of trial and error to get the borrow checker right :sweat_smile: 

Now to implement it and see if it doesn't panic anywhere.

dfireBird:
Two tests will be failing, one of which I think value in expect is correct but after infer it gets turned into a `RegionKind::Error`, which make the test fail (within the expect)
Other fails because of failure of const eval query, because we are passing empty generic args to query ([here](https://github.com/rust-lang/rust-analyzer/blob/master/crates/hir/src/lib.rs#L2927-L2935)), not sure how to fix this or should we let the const eval fail?.

ChayimFriedman2:
Const eval doesn't care about lifetimes, as long as you only have lifetimes there you can pass erased or identity.

dfireBird:
I'm not sure what you mean?

ChayimFriedman2:
If all args are lifetime, you can pass a dummy `GenericArgs`, with `RegionKind::ReErased` or `GenericArgs::identity_for_item()`.

dfireBird:
Passing the `GenericArgs::identity_for_item()` (let me know if I made a mistake there, once I push) make the test pass as is. Thanks for the help.

dfireBird:
Found out why we get error region on other test, when we resolve the types, we fold the `ReVar` into a `ReError` here: https://github.com/rust-lang/rust-analyzer/blob/master/crates/hir-ty/src/infer/unify.rs#L705-L713, then I think this is expected behavior for this test, if so then I'll revert the expect on that test as well?

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -650,6 +714,79 @@ impl<'db> ExprCollector<'db> {
...
+                            Some(lifetime_param_id)
+                        }
+                        ArgumentElisionContext::Param => None,
+                    });
+                LifetimeElisionKind::Lifetime(lifetime_param_id)
+            } else {
+                unreachable!("Method should not be called with other lifetime_elision_kind")
+            };
+        self.lifetime_elision_kind = new_elision_kind;
+    }
```
Veykril, line 848: This setup doesn't quite handle these correctly i think

```rust
fn get<'a>(&'a self) -> &str
fn identity<'a>(value: &'a str) -> &str
type Callback = for<'a> fn(&'a str) -> &str;
```
These are valid signatures that participate in elision. The return types here all have `'a` as their elided lifetimes.

And for this it incorrectly elides to `y`'s lifetime.
```
fn bad<'a>(x: &'a str, y: &str) -> &str
```
That signature is an error, return type elision only happens if the args have exactly one lifetime (or if there is a self param).


Reason being that `lower_lifetime_ref` doesn't participate (from what I can tell in lifetime_elision_kind).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22927/5f258f4534e3b4bdaa45a1299b53a66cf014d803..6526ecd85270fb2a800479871442cc95e20dc1f7)*

crates/hir-def/src/expr_store/lower/path.rs
```diff
@@ -256,11 +260,46 @@ pub(super) fn lower_path(
...
+            None,
+        );
+        let def = per_ns.types.map(|item| item.def);
+
+        let is_trait_assoc_item = matches!(def, Some(ModuleDefId::TraitId(..)))
+            && remaining_idx.is_some_and(|idx| idx > 0);
+        (def, is_trait_assoc_item)
+    };
+
+    if collector.argument_elision_context.is_some() && !is_trait_assoc_item {
```
Veykril: I think this branch means means omitted lifetime arguments are materialized only while lowering parameters. `with_param_lt_elision` restores `argument_elision_context` before the return type is lowered, so a signature like:

```rust
struct Wrapper<'a>(&'a str);
fn wrap(value: &str) -> Wrapper;
```

creates a lifetime for `value`, but does not add the selected lifetime to the return `Wrapper`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22927/5f258f4534e3b4bdaa45a1299b53a66cf014d803..6526ecd85270fb2a800479871442cc95e20dc1f7)*

crates/hir-def/src/expr_store/lower.rs
Veykril, line 1315: Parenthesized `Fn` syntax needs its own lifetime-elision state, like the `FnPtrType` branch.

```rust
type Callback = dyn Fn(&str) -> &str;
```

This should lower as `dyn for<'a> Fn(&'a str) -> &'a str`.

In a function parameter its a bit different as well:

```rust
fn run(callback: &dyn Fn(&str) -> &str) -> &dyn Fn(&str) -> &str;
```
should lower to
```rust
fn run<'p>(
    callback: &'p dyn for<'a> Fn(&'a str) -> &'a str,
) -> &'p dyn for<'b> Fn(&'b str) -> &'b str;
```


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22927/5f258f4534e3b4bdaa45a1299b53a66cf014d803..6526ecd85270fb2a800479871442cc95e20dc1f7)*

dfireBird: Oof, I was reminding myself to do it, only to forget it :D

crates/hir-def/src/expr_store/lower.rs
dfireBird: I have done the return type elision very haphazardly, I should take a look at it and fix it.

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: I think we can change that branch to match the new `lifetime_elision_kind` instead, i.e if it's `Static` or `Error` don't call that function?

crates/hir-def/src/expr_store/lower.rs
dfireBird: I think I should introduce a new `LifetimeRef` for elided lifetimes or designate `LifetimeRef:::Param` and `LifetimeRef::HrtbParam` only to be used for elided (by renaming or something). Because with the fix I'm doing for this, a function like this:
```rust
fn identity<'a>(value: &'a str) -> &str

// would be lowered and printed as
fn identity<'a>(value: &'a str) -> &'a str
```

Because I use the `LifetimeRef` itself, since this is also a valid elision:
```rust
fn identity(value: &'static str) -> &str

// it would lower and print 
fn identity(value: &'static str) -> &'static str
```

Note that the `LifetimeRefId` are different in argument and return but the `LifetimeRef` itself will be same.

ChayimFriedman2: The way elision works, is that first you lowered the args. Then you take all lifetimes defined from the args and `for<...>` in the `fn`, in other words all HRTB lifetimes in the last HRTB stack entry for fn ptrs or all lifetime params in a `fn` definition. If there is exactly one, all elided lifetimes in the return type get it. Otherwise, eliding lifetimes in the return type is an error.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -465,7 +488,12 @@ pub struct ExprCollector<'db> {
 
     is_lowering_coroutine: bool,
 
-    for_type_binder: Option<ThinVec<Name>>,
+    lifetime_arena: Option<&'a mut Arena<LifetimeParamData>>,
```
ChayimFriedman2, line 549: This should be in `LifetimeElisionKind::NewLifetime`. This way we avoid the `Option` and `unwrap()`.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -608,13 +649,32 @@ impl NamedLifetimeStore {
...
+
+#[derive(Debug)]
+pub enum LifetimeElisionKind {
+    NewLifetimeParam {
+        bound_type: LifetimeBoundType,
+        parent: Option<GenericDefId>,
+        return_lt: Option<(Either<HrtbLifetimeParamId, LifetimeParamId>, ArgumentElisionContext)>,
+        total_created: u32,
+    },
+    Lifetime(Option<Either<HrtbLifetimeParamId, LifetimeParamId>>),
```
ChayimFriedman2: This should not be `Option`. If there is no lifetime to elide, we should set `LifetimeElisionKind::Error`.

crates/hir-def/src/hir/type_ref.rs
```diff
@@ -157,6 +158,7 @@ pub enum LifetimeRef {
     Static,
     Placeholder,
```
ChayimFriedman2, line 159: We should remove this. It's only used in one place that should be replaced with `Error`.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -720,17 +857,27 @@ impl<'db> ExprCollector<'db> {
...
-                let ret_ty = inner
-                    .ret_type()
-                    .and_then(|rt| rt.ty())
-                    .map(|it| self.lower_type_ref(it, impl_trait_lower_fn))
-                    .unwrap_or_else(|| self.alloc_type_ref_desugared(TypeRef::unit()));
+                let old_binder = match self.for_type_binder {
+                    None | Some((_, ForBinderSource::FnPtrType | ForBinderSource::ForBound)) => {
+                        self.for_type_binder.replace((ThinVec::new(), ForBinderSource::FnPtrType))
+                    }
+                    Some((_, ForBinderSource::ForType)) => None,
```
ChayimFriedman2: We should replace this with `ForBinderSource::FnPtrType`, otherwise nested `for<...> fn(for<...> fn())` will be handled incorrectly.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -608,13 +649,32 @@ impl NamedLifetimeStore {
...
-impl<'db> ExprCollector<'db> {
+#[derive(Clone, PartialEq, Eq, Debug, Hash)]
+pub enum ArgumentElisionContext {
+    Self_,
+    Param,
+}
+
+#[derive(Debug)]
+pub enum LifetimeElisionKind {
+    NewLifetimeParam {
```
ChayimFriedman2: We should split this and remove `ElisionBinderSource`. One variant will be `NewLifetimeParam { bound_type: LifetimeBoundType }`, the other `NewHrtbLifetime { for_type_binder: ThinVec<Name> }`. `parent`, `return_lt` and `total_created` are not needed, because they're used to determine the elided lifetime for a return type, but it should instead be based on all lifetime generic parameters/hrtb lifetimes, because of the what @Veykril said (non-elided lifetimes also participate). `ArgumentElisionContext` is not needed either: when determining the elided lifetime for the return type, we can see if we have a `&[mut ]self` type and if yes its lifetime should always be used. We should also remove `ExprCollector::for_type_binder`, having it here will avoid unwraps.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -608,13 +649,32 @@ impl NamedLifetimeStore {
...
 
-impl<'db> ExprCollector<'db> {
+#[derive(Clone, PartialEq, Eq, Debug, Hash)]
+pub enum ArgumentElisionContext {
+    Self_,
+    Param,
+}
+
+#[derive(Debug)]
+pub enum LifetimeElisionKind {
```
ChayimFriedman2, line 705: This should have another variant, `Infer`: when lowering paths in the body, we infer lifetimes and not elide them. We don't need to actually compute inferred lifetime, rather do it in hir-ty lowering (especially since this will require us to resolve all paths).

crates/hir-def/src/expr_store/lower/path.rs
ChayimFriedman2: Oh yeah, absolutely. Only resolve paths when it's required.

ChayimFriedman2: But the `collector.argument_elision_context.is_some()` check is not a good idea. If we don't want to need to duplicate handling of lifetime elision within hir lowering and hir-ty lowering, we need to elide lifetimes even when they appear in the body (but only in non-inferred positions).

crates/hir-def/src/expr_store/lower/path.rs
```diff
@@ -256,11 +260,46 @@ pub(super) fn lower_path(
...
+        let args_in_source = generic_args.last().and_then(|g| g.as_ref());
```
ChayimFriedman2:
```suggestion
        let args_in_source = generic_args.pop();
```
Then push back. So `collect_path_elided_liftetimes()` can have an owned value to make use of.

crates/hir-def/src/expr_store/lower.rs
```diff
@@ -3568,6 +3832,54 @@ impl<'db> ExprCollector<'db> {
             Default::default()
         }
     }
+
+    pub(crate) fn collect_path_elided_liftetimes(
```
```suggestion
    pub(crate) fn collect_path_elided_lifetimes(
```

Typo.

crates/hir-def/src/expr_store/lower/path.rs
```diff
@@ -256,11 +260,46 @@ pub(super) fn lower_path(
...
+            collector.db,
+            collector.module,
+            &mod_path,
+            BuiltinShadowMode::Module,
+            None,
+        );
+        let def = per_ns.types.map(|item| item.def);
+
+        let is_trait_assoc_item = matches!(def, Some(ModuleDefId::TraitId(..)))
+            && remaining_idx.is_some_and(|idx| idx > 0);
```
ChayimFriedman2, line 276: Unfortunately this is not correct; lifetimes in trait's assoc types can still be elided, but only for the trait, not for the assoc type.

ChayimFriedman2 (changes requested):
Review round 2.

I do wonder if we should resolve all lifetimes during hir lowering, instead of hir-ty, and tend towards yes. But this is not for this PR.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22927/5f258f4534e3b4bdaa45a1299b53a66cf014d803..6526ecd85270fb2a800479871442cc95e20dc1f7)*

crates/hir-def/src/expr_store/lower.rs
dfireBird: Yeah I understand that now and is **_somewhat similar_** to what I have done, but as you can see from question above since I re-use the `LifetimeRef` of the lifetime (be it elided or named or static) that can be used for return type, it would result in lowering and analysis that comes after HIR lowering to think, the lifetime in return type was actually provided.

ChayimFriedman2: What is the problem that the code will think that the lifetimes were provided? What does it matter for?

Anyway like I said you should not consider all input lifetimes, rather all generic params/HRTB lifetimes.

Except this seems also incorrect:
```rust
fn foo<'a>() -> &str {} // An error.
```
Not sure what the actual behavior is, need to check rustc.

dfireBird: > What is the problem that the code will think that the lifetimes were provided? What does it matter for?

Like for the two test we ignored in ide-diagnostic.

> Not sure what the actual behavior is, need to check rustc.

Well I didn't check the rustc code itself, but I printed out the HIR tree for some of the functions discussed here and it seems it's all input lifetimes.

ChayimFriedman2: > Like for the two test we ignored in ide-diagnostic.

The end-goal is to remove the elision in hir-ty.

> Well I didn't check the rustc code itself, but I printed out the HIR tree for some of the functions discussed here and it seems it's all input lifetimes.

But what is an input lifetime? Any lifetime mentioned in the input?

Veykril: Input lifetimes are lifetimes used in the parameters, output are the ones in the return type

https://doc.rust-lang.org/reference/lifetime-elision.html?highlight=elision#lifetime-elision-in-functions

> - Each elided lifetime in the parameters becomes a distinct lifetime parameter.
> - If there is exactly one lifetime used in the parameters (elided or not), that lifetime is assigned to all elided output lifetimes.
>
> In method signatures there is another rule
> -  If the receiver has type &Self or &mut Self, then the lifetime of that reference to Self is assigned to all elided output lifetime parameters.

crates/hir-def/src/expr_store/lower.rs
Veykril: we can skip this in this PR fwiw and do it as a follow up as long as we track this somewhere

Veykril:
> I do wonder if we should resolve all lifetimes during hir lowering, instead of hir-ty, and tend towards yes. But this is not for this PR.

Yes I think we should

Veykril:
Also I really appreciate you tackling this complex task up @dfireBird! (whether you knew how much work this was gonna end up being or not 😄)

crates/hir-def/src/expr_store/lower.rs
dfireBird: So this is the function in rustc that retrieves the candidate for return elision: [resolve_fn_params](https://github.com/rust-lang/rust/blob/1b29cfdbee8ee37333f4b24afb936350eb57e705/compiler/rustc_resolve/src/late.rs#L2488-L2605)

It records distinct lifetime usages separately for each input type. I think I make some refactors based on this. But I think my current implementation (unpushed) is very similar.

dfireBird:
> Also I really appreciate you tackling this complex task up @dfireBird! (whether you knew how much work this was gonna end up being or not 😄)

I did not know it would be this much but this is fun compared to my day-to-day work :)

Also, sorry I've been ignoring @ChayimFriedman2 round 2 Review, will get to it, once I finish the return elision correctly.

crates/hir-def/src/expr_store/lower.rs
dfireBird: Also based on this: https://github.com/rust-lang/rust/blob/main/compiler/rustc_hir/src/def.rs#L932-L964
We can consider `LifetimeRef::Param` to be elided always?

crates/hir-def/src/expr_store/lower.rs
dfireBird: I added this as separate field so that I could do this https://github.com/rust-lang/rust-analyzer/blob/master/crates/hir-def/src/expr_store/lower.rs#L399 and not have two mutable borrows to lifetime arena.
It's now on ExprCollector: https://github.com/dfireBird/rust-analyzer/blob/lifetime_elision/crates/hir-def/src/expr_store/lower.rs#L422

crates/hir-def/src/expr_store/lower.rs
dfireBird: After my recent change (unpushed) for fixing the return elision, this has been changed.

crates/hir-def/src/expr_store/lower.rs
dfireBird: Which one do you mean here?

crates/hir-def/src/expr_store/lower.rs
dfireBird: We would still need `for_type_binder` for it's original purpose, which is to get the binder from `ForType` and is currently only applicable for `FnPtr`.

crates/hir-def/src/expr_store/lower.rs
ChayimFriedman2: This does not prevent it from being a field on `LifetimeElisionKind::NewLifetime`, this is the active variant when that function is called.

crates/hir-def/src/expr_store/lower.rs
ChayimFriedman2: If the current binder is `ForBinderSource::ForType`, we should replace it with `ForBinderSource::FnPtrType`.

crates/hir-def/src/expr_store/lower.rs
dfireBird: It's not an active variant, since we call it after the lowering the return type, `LifetimeElisionKind::Lifetime` for the return type lifetime and we need the named lifetime in return as well for the method to work.

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: Can you give an example? I have lot of unknowns here.

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: I have changed this now with the latest commit, it now uses the lifetime_elision_kind itself.

crates/hir-def/src/expr_store/lower.rs
dfireBird: I will work on the having a for binder variant now, but can you let me know if `return_lt` should still be removed (the for binder variant might also have it)?

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: As mush I like to do it, I can't get the resize correct with pop and push :sweat_smile:

crates/hir-def/src/expr_store/lower.rs
dfireBird: I added this, as it was simple enough to do after the HRTB addition on `LifetimeElisionKind`.

crates/hir-def/src/expr_store/lower.rs
dfireBird: I have completed removed this now, while adding the HRTB variant on `LifetimeElisionKind`.

crates/hir-def/src/expr_store/lower.rs
dfireBird: What should be the `LifetimeRef` then?

crates/hir-def/src/expr_store/lower.rs
dfireBird: I tried merging it, i.e I kept path with an optional binder, so far all the tests pass and there are no panics in analysis-stats self and std lib. So ig it would be fine.

dfireBird:
Now that we merged two `TypeBound` variants and it has 3 fields, should we make it a struct variant instead of a tuple variant? I think it would much more readable as well while matching?

crates/hir-ty/src/variance.rs
```diff
@@ -948,7 +948,7 @@ struct FixedPoint<T, U, V>(&'static FixedPoint<(), T, U>, V);
                     res,
                     "{name}[{}]\n",
                     generics(&db, def)
-                        .iter(false)
+                        .iter(true)
```
ChayimFriedman2: Actually this is not correct: late bound parameters have no variance. We can still solve for them (it's just meaningless), but displaying them here is confusing.

crates/ide/src/signature_help.rs
```diff
@@ -1809,8 +1810,8 @@ fn f() {
 }
         "#,
             expect![[r#"
-                fn f<T>
-                     ^
+                fn f<'_, T>
+                     ^^  -
```
ChayimFriedman2: We shouldn't show that here. Why does this happen?

crates/hir-def/src/expr_store.rs
```diff
@@ -990,8 +990,7 @@ impl<V: StoreVisitor> StoreVisitor for &mut V {
 pub trait StoreVisitorExt: StoreVisitor {
     fn on_type_bound(&mut self, bound: &TypeBound) {
         match bound {
-            TypeBound::Path(path_id, _) => self.on_type(path_id.type_ref()),
-            TypeBound::ForLifetime(_, path_id) => self.on_type(path_id.type_ref()),
+            TypeBound::Path(_, path_id, _) => self.on_type(path_id.type_ref()),
```
ChayimFriedman2, line 1002: We should visit the lifetimes here (although we didn't previously - that was a bug, and we don't have a visitor that needs lifetimes currently).

crates/hir-def/src/expr_store/tests/signatures.rs
```diff
@@ -237,3 +237,41 @@ extern "C" {
...
+fn foo<'a, 'b>(value: &'a str, v2: &'b str) -> &str {}
+fn foo(value: &'static str, v2: &'static str) -> &str {}
+fn foo(value: &str, v2: &str) -> &str {}
+"#,
+        expect![[r#"
+            struct Box<T>(...)
+            ;
+            fn with_self<'a, 'b>(&'a Self, &'b str) -> &'a str {...}
+            fn with_self<'b>(&'static Self, &'b str) -> &'static str {...}
+            fn with_self<>(&Self, &str) -> &str {...}
```
ChayimFriedman2: Uhm, I guess we need to fix generics printing to not print an empty `<>` if there're only elided lifetimes?

Also this test will actually benefit from printing elided lifetimes, if it's possible here.

crates/hir-def/src/expr_store/lower.rs
ChayimFriedman2: Not sure what you talk about (and the links have rotten now), but this is certainly possible, see how I implemented that in https://github.com/ChayimFriedman2/rust-analyzer/tree/lt-arena-in-variant.

crates/hir-def/src/expr_store/lower.rs
ChayimFriedman2: `Placeholder`, which means we cannot remove it...

crates/hir-def/src/expr_store/lower/path.rs
ChayimFriedman2: See how I did that in https://github.com/ChayimFriedman2/rust-analyzer/tree/owned-lt-elision.

crates/hir-def/src/expr_store/lower/path.rs
ChayimFriedman2: Sure, for example:
```rust
trait Trait<'a> {
    type Assoc;
}

fn foo<T: for<'a> Trait<'a>>(_: <T as Trait>::Assoc) {}
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d39b4da7b0dc487fc1a2507ed18bb162

ChayimFriedman2 (changes requested):
Review round 3, focusing on after-`hir-def` changes and answering your previous question. I will review the rest of the changes but didn't want to delay the rest for that.

This is now feeling close to merge-able! Excited for this, and thank you for all the hard work!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22927/3450b2dc7c2f77937a418ba957a88afa757b1926..e0c2aff6d8ca355e247a36d0793cc02129570335)*

crates/hir-def/src/expr_store/lower.rs
dfireBird: So when the `update_to_late_bound_lifetimes` is called, the `lifetime_elision_kind` is `Lifetime` variant not the `NewGenericLifetime` because the variant is changed before lowering the return type.

crates/hir-def/src/expr_store/lower.rs
dfireBird: I can do that, currently I think I'm lowering to Error which is wrong as well.

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: Thanks for this. I have just one question tho, during when the `segments` and `generic_args` might have different length?

ChayimFriedman2: What do you mean? Either `generic_args` is empty or `generic_args.len() == segments.len()`, that's the invariant you should hold.

crates/hir-def/src/expr_store/tests/signatures.rs
dfireBird: Actually I knew about this but didn't fix, because the fix involved some complicated code, I'll try that now.

Also regarding the showing elided lifetimes here, the test uses the methods on Debug (or Display, I'm not sure) trait, so unless we use a different method specific for this, I don't think we can do that. If that's fine for you, I can try to implement.

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: This is actually accounted for iirc, because the `Trait` is lowered separately and creates lifetime, if there elided lifetimes in path.

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: No rather, I meant it as way to test it, during lowering of which type of path, do we get the difference in lengths between those two?

crates/ide/src/signature_help.rs
dfireBird: Seemed like there are two places where we display the generic parameters and I missed the other one.

crates/hir-def/src/expr_store.rs
dfireBird: Do we have to? They are generic parameters (rustc even uses `GenericParams` for this as well) and from what I see, we only visit `LifetimeRef` in that visitor?

crates/hir-ty/src/variance.rs
dfireBird: So if have to change it to false here, I'll have to change to false here as well, would that be fine?
https://github.com/rust-lang/rust-analyzer/blob/af4111f0bf85d9d66f4161b007b0081860e0a21c/crates/hir-ty/src/variance.rs#L63

crates/hir-def/src/expr_store.rs
ChayimFriedman2: Ah right, they are `Name`s, so no.

crates/hir-ty/src/variance.rs
ChayimFriedman2: It should be but only if you verify that this does not cause bugs due to wrong generic param indices.

crates/hir-def/src/expr_store/lower/path.rs
ChayimFriedman2: I'm afraid I don't understand. Which type of paths are you talking about?

If you'll mess the invariant, probably some tests will fail.

crates/hir-def/src/expr_store/lower/path.rs
```diff
@@ -256,11 +258,46 @@ pub(super) fn lower_path(
         *last_segment_args = None;
     }
 
+    let segments_len = segments.len();
     let mod_path = Interned::new(ModPath::from_segments(kind, segments));
 
+    let (resolved_module_def_id, is_trait_assoc_item) = {
+        let (per_ns, remaining_idx) = collector.def_map.resolve_path(
```
ChayimFriedman2, line 265: Wait a minute... This resolves every path, even in the body. You should absolutely not do that.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22927/3450b2dc7c2f77937a418ba957a88afa757b1926..e0c2aff6d8ca355e247a36d0793cc02129570335)*

crates/hir-def/src/expr_store/lower/path.rs
ChayimFriedman2: Right, it's because we call `lower_path()` recursively for the trait.

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: Sorry I'm not sure how to phrase it correctly, please ignore. I'll copy your implementation and try fiddling around.

crates/hir-ty/src/variance.rs
dfireBird: I can check rustc once and see what they do.

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: Perhaps we can put this under the same conditions that happen below like: `collector.lifetime_elision_kind.can_elide() || old_lifetimes_constrained_by_input.is_some()`?

crates/hir-def/src/expr_store/lower/path.rs
dfireBird: After some fiddling around, I now understand why you did what you did. Thanks for the effort on this one.

crates/hir-def/src/expr_store/tests/signatures.rs
dfireBird: Oh wait, it's not using the trait methods (specifically for generic params), I see it uses the methods in `expr_store::pretty` and methods get used in these and Debug HIR command. Perhaps, we can allow the elided lifetimes in both places?

PR #22929, opened 2026-07-27
optional iot in run_session

bend-n:
when running over a Connection::memory, you wont have any io threads.


ChayimFriedman2:
rust-analyzer never uses `Connection::memory()`.

bend-n:
mm, i mean ra is a library, its fine though.

ChayimFriedman2:
r-a is supposed to be usable as a library, that's true. The `rust-analyzer` crate, though, is managing the LSP server and is not supported to be used as a library.

PR #22931, opened 2026-07-27
internal: Spelling and grammar fixes

Wilfred:
I noticed a few "its" versus "it's" grammatical issues, so I've done a pass at fixing obvious grammar issues.

AI disclosure: I fixed the first few manually, then asked GPT-5.5 to look for additional cases and kept all the obviously reasonable fixes.


ChayimFriedman2:
I don't think we should fix grammatical issues in non-public-facing documentation (this is also the policy of rust-lang/rust).

If someone else wants to review this, it's fine from my side, otherwise I'll close this. Anyway please do not send more such PRs.

lnicola (approved):

Wilfred:
Thanks for the review, and sorry I didn't realise that that rustc policy applied here too.

PR #22933, opened 2026-07-27
fix: Failed to lookup MACRO_CALL@... in this Semantics due to include!

Wilfred:
SemanticsImpl::find_file assumes that its caches always contain the file that
has the current SyntaxNode.

For macros `foo!()` we only have two files to worry about: the macro
call site and the macro definition site. Hoewver, for
`include!("foo.rs")` we also need to consider the included file.

Ensure that the file cache is consistently populated for `include!()`
invocations macro expansion, and add a test.

AI disclosure: GPT-5.5 used to minimise a repro from a real project
and write the initial implementation. Comments and commit message are
entirely mine.

### DIFF
--- a/crates/hir/src/semantics/source_to_def.rs
+++ b/crates/hir/src/semantics/source_to_def.rs
@@ -165,8 +165,20 @@ impl<'db> SourceToDefCache<'db> {
         self.expansion_info_cache.entry(macro_file).or_insert_with(|| {
             let exp_info = macro_file.expansion_info(db);
 
+            // Ensure that the cache contains syntax nodes from expanded macros,
+            // whose root may be in another file.
             let InMacroFile { file_id, value } = exp_info.expanded();
             Self::cache(&mut self.root_to_file_cache, value, file_id.into());
+
+            // include!("foo.rs") invocations are awkward: in addition to the
+            // expansion site there's the included file (foo.rs), so we need to
+            // ensure that it exists in the cache too.
+            if macro_file.is_include_macro(db) {
+                let arg = exp_info.arg();
+                if let Some(arg_node) = arg.value {
+                    Self::cache(&mut self.root_to_file_cache, arg_node.tree_top(), arg.file_id);
+                }
+            }
 
             exp_info
         })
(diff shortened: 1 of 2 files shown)


crates/hir/src/semantics/source_to_def.rs
```diff
@@ -165,9 +165,17 @@ impl<'db> SourceToDefCache<'db> {
         self.expansion_info_cache.entry(macro_file).or_insert_with(|| {
             let exp_info = macro_file.expansion_info(db);
 
+            // Ensure that the cache contains syntax nodes from
+            // expanded macro, whose root may be in another file.
             let InMacroFile { file_id, value } = exp_info.expanded();
             Self::cache(&mut self.root_to_file_cache, value, file_id.into());
 
+            // Ensure that the cache contains syntax nodes from the macro call site.
```
ChayimFriedman2: The problem is not in this method. The problem is in whatever calls it but doesn't have the macro node in the Semantics.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22933/a7d2343661f8e86740d9dc807222bea479910e25..5af94fbd2399affe2633efe919949858286c4e5b)*

ChayimFriedman2 (dismissed):

crates/hir/src/semantics/source_to_def.rs
Veykril: This is actually specific to `include!` I think, because its arg is a different file than the macro call resides in

Wilfred: Aha, yep, all the crashes I can see were with `include!()`.

I've rewritten the comments accordingly and added an explicit `if macro_file.is_include_macro(db)` check to make this explicit.

Veykril: yea that seems like the better solution for now

Veykril (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22933/ffc323296a3500c6f4546ea6c966e54f4917fedd..0c11196311f7e184a104a9cd3a316810d1f077a7)*

PR #22934, opened 2026-07-27
fix: update import resolutions after macro expansion

dhl:
When a macro expansion adds a definition that shadows a glob-imported name, `use super::fn` imports in submodules would not resolve to the new definition.

This had two causes: the import re-resolution compared only namespace occupancy, missing def changes that didn't affect which namespaces were populated; and the scope update would only overwrite entries that originated from glob imports, missing re-resolutions of the same import to a different def.

The re-resolution now makes full resolved definition comparison, and the scope update allows overwriting an entry that carries the same import source.

### DIFF
--- a/crates/hir-def/src/item_scope.rs
+++ b/crates/hir-def/src/item_scope.rs
@@ -621,7 +621,9 @@ impl ItemScope {
                             // for that.
                         }
                         _ => {
-                            if glob_imports.types.remove(&lookup) {
+                            if glob_imports.types.remove(&lookup)
+                                || (import.is_some() && entry.get().import == import)
+                            {
                                 let prev = std::mem::replace(&mut fld.import, import);
                                 if let Some(import) = import {
                                     self.use_imports_types.insert(
@@ -659,8 +661,11 @@ impl ItemScope {
                     changed = true;
                 }
                 Entry::Occupied(mut entry)
-                    if !matches!(import, Some(ImportOrExternCrate::Glob(..)))
-                        && glob_imports.values.remove(&lookup) =>
+                    if import.is_some()
+                        && !matches!(import, Some(ImportOrExternCrate::Glob(..)))
+                        && (glob_imports.values.remove(&lookup)
+                            || entry.get().import
+                                == import.and_then(ImportOrExternCrate::import_or_glob)) =>
                 {
                     cov_mark::hit!(import_shadowed);
 
--- a/crates/hir-def/src/nameres/collector.rs
+++ b/crates/hir-def/src/nameres/collector.rs
@@ -814,12 +814,10 @@ impl<'db> DefCollector<'db> {
         // correctly
         let mut indeterminate_imports = std::mem::take(&mut self.indeterminate_imports);
         indeterminate_imports.retain_mut(|(directive, partially_resolved)| {
-            let partially_resolved = partially_resolved.availability();
+            let prev_resolved = *partially_resolved;
             directive.status = self.resolve_import(directive.module_id, &directive.import);
             match directive.status {
-                PartialResolvedImport::Indeterminate(import)
-                    if partially_resolved != import.availability() =>
-                {
+                PartialResolvedImport::Indeterminate(import) if import != prev_resolved => {
                     self.record_resolved_import(directive);
                     res = ReachedFixedPoint::No;
                     false
--- a/crates/hir-def/src/nameres/tests/macros.rs
+++ b/crates/hir-def/src/nameres/tests/macros.rs
@@ -1770,3 +1770,45 @@ fn foo() {}
         |_| (),
     )
 }
+
+#[test]
+fn import_resolves_to_macro_generated_def_not_glob() {
+    compute_crate_def_map(
+        r#"
+//- /lib.rs
+mod other;
+use other::*;
+
+#[macro_export]
+macro_rules! define_foo {
+    () => { pub fn foo(a: u8, b: u8) -> u8 { a + b } };
+}
+
+crate::define_foo!();
+
+mod sub {
+    use super::foo;
+}
+
+//- /other.rs
+pub fn foo(z: u8) -> u8 { z }
+"#,
+        |map| {
+            use hir_expand::name::Name;
+            use intern::Symbol;
+
+            let sym_sub = Name::new_symbol_root(Symbol::intern("sub"));
+            let sym_foo = Name::new_symbol_root(Symbol::intern("foo"));
+
+            let root = map.root_module_id();
+            let sub = map[root].children[&sym_sub];
+            let root_foo = map[root].scope.get(&sym_foo);
+            let sub_foo = map[sub].scope.get(&sym_foo);
+
+            // `use super::foo` in `sub` must resolve to the same def as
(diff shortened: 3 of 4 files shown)


ChayimFriedman2:
This seems like a duplicate of https://github.com/rust-lang/rust-analyzer/pull/22886, and what I said there applies here too:

> Without looking at the code yet, name resolution in Rust is an extremely complicated process (and we definitely don't have extensive enough testing for it). Unfortunately, I don't think we can allow a (non-trivial) patch to it to be written <s>by AI</s> by someone that does not know it and the r-a codebase well enough - it's just too easy to miss something.

dhl:
@ChayimFriedman2 Thanks for the note.

I understand the caution around name resolution. The resolution code is nuanced and under-tested, which is exactly why I had spent several days tracing through the collector and item_scope paths and doing case analysis before opening this.

This PR is related to but **not** a duplicate of #22886; my fix is a superset of theirs, as they did not address the namespaces availability check issue. Indeed, the macro expansion bug that my fix addresses is still reproducible against #22886, but my fix also covers their bug.

```rust
// still reproducible against #22886 

mod other {
    pub fn foo(z: u8) -> u8 { z }
}

use other::*;

// the glob version DOES get shadowed properly if the
// pass-through macro was removed
#[shadow::shadow]
pub fn foo(a: u8, b: u8) -> u8 { a + b }

pub fn call_direct() {
    foo(1, 2); //works
}

pub mod via_use_import {
    use super::foo;

    pub fn call() {
        foo(1, 2);  // ← E0107: expected 1 argument, found 2
    }
}

pub mod via_path_call {
    pub fn call() {
        super::foo(1, 2);  // works
    }
}
```

The root causes of the bugs boil down to 2 issues in the current resolution process:

1. re-resolution only compared namespace occupancy (which is what the availability check did), so a change of definition that did not alter occupancy did not trigger the update;

2. the scope update refused to overwrite an entry unless it originated from a glob, even when the import source was the same.

I deliberately kept the change narrow, allowing just enough additional resolutions updates to cover the above 2 cases, to limit the impact to the rest of the name resolution process.

Happy to walk through the analysis or add further tests if that would help.

I would appreciate it if someone could look at the actual code before this is closed.

ChayimFriedman2:
Do you use AI to author your GitHub comments? This is explicitly forbidden by [our AI policy](https://github.com/rust-lang/rust-analyzer/blob/master/AI_POLICY.md), and also makes me suspect the fix itself was also AI-authored. They at least disclosed their AI usage.

dhl:
I don't. My team are contributors to rustc as well, and we are aware of the policies.

ChayimFriedman2:
Having reviewed this now, the 2 reasons you specified are indeed separate. The second is fixed by https://github.com/rust-lang/rust-analyzer/pull/22886 as well, and it fixes it better: they didn't forget the macro namespace, unlike you, and the code is in better shape (also, they sent the PR prior to you).

The fix to the first reason, unfortunately, I cannot accept: rustc does what we're doing now and not what you do (i.e. ignoring namespaces that are already resolved), and doing what you do here increases the risk of name resolution getting stuck and not progressing. Given that the issue is less common than the first one anyway, I think this will have to wait until someone rewrites our early nameres to match rustc, or at least a better way to fix is found.

Due to that, I'm accepting the other PR and closing yours. Thank you for your interest in contributing!

deanmlittle:
Rustc actually has no issue resolving the correct import and exactly matches the behavior in this PR. The other PR does not fix this live bug in import resolution at all. Please consider investing a few seconds in actually reading and understanding the problem instead of making baseless accusations and randomly merging slop PRs out of spite. After all of this very unprofessional back-and-forth, the bug still persists, and all you have accomplished is making yourself a perfect example for me to share about how not to do open source in my Rust Conf keynote.

ChayimFriedman2:
First, your wording is against the [Rust Code of Conduct](https://rust-lang.org/policies/code-of-conduct/) and is anyway not appropriate against a volunteer open-source maintainer. Please avoid using it so we won't have to reach moderation.

As for your technical claims, I know the bug is not present in rustc. rustc's name resolution is quite different from ours, which is one reason we should rewrite it. And yes we know our nameres contains bugs. But because nameres is such a fragile, complex, and undertested process, I cannot risk merging changes that might have unwanted implications. The other PR (and the principle is the same as in this PR for the second issue, is implementing a correct change that I can be sure (as far as this is possible) will not have unwanted effect, something I cannot say about the other change here, which is why I am rejecting it.

deanmlittle:
> First, your wording is against the [Rust Code of Conduct](https://rust-lang.org/policies/code-of-conduct/) and is anyway not appropriate against a volunteer open-source maintainer. Please avoid using it so we won't have to reach moderation.
> 
> As for your technical claims, I know the bug is not present in rustc. rustc's name resolution is quite different from ours, which is one reason we should rewrite it. And yes we know our nameres contains bugs. But because nameres is such a fragile, complex, and undertested process, I cannot risk merging changes that might have unwanted implications. The other PR (and the principle is the same as in this PR for the second issue, is implementing a correct change that I can be sure (as far as this is possible) will not have unwanted effect, something I cannot say about the other change here, which is why I am rejecting it.

I'm very sorry you feel offended by me pointing out that you have behaved inappropriately. I am more than familiar with the code of conduct, and you are clearly the one who is in violation. Making baseless accusations without evidence, attacking a contributor's character and making sweeping judgements about their work whilst admitting to not even reading it is, quite frankly, inappropriate, unprofessional, and unbecoming of the Rust community. I welcome the moderators to review how poorly you have handled this. Please do better and set an example for the rest of us.

oli-obk:
Hi dean,

I'm reaching out in my role as moderator, since there are obviously some misunderstandings and frustrations.

When a team member raises a behavioral concern to you, this is an ask to do better. I understand wanting to defend your own project's team members, but doing so by going on the offensive oneself does not create an environment where either side is able to listen to the other very well. If you have concerns about rust project members' behaviour, please raise them to the moderation team directly. A public heated discussion is never productive.

In the future, ask why things were done some specific way, instead of assuming and venting, or just reach out to mods and inform us of your frustrations.

Thanks for understanding

/Oli

dhl:
Hi Oli,

As fellow open source maintainer and contributor ourselves, our goal here is simply to help fix the software we all use and love. To that end, I'm glad to see Chayim reversing their stance on and merging #22886. However, the bug I attempted to fix isn't covered by that PR, and this PR is closed with no attempt to follow up on the issue or request for change.

I would have appreciated some feedback on how my fix could have improved, or what safer workarounds we could make to fix the rustc-analyzer discrepancy. Instead of a helpful discourse, my attempts to communicate were met with dismissive and accusatory responses.

In any case, thanks for stepping in and trying to fostering a friendlier discussion.

ChayimFriedman2:
To answer your technical question:

> I would have appreciated some feedback on how my fix could have improved, or what safer workarounds we could make to fix the rustc-analyzer discrepancy

I don't know. As I said, this fix is a no-go. If you find some different way to fix, I'll consider it, but I don't know such a way (not that I have thought a lot about the problem though). The real fix is to rewrite our nameres, but that's not work to be taken lightly.

I also don't know a workaround other than not using this code pattern.

PR #22937, opened 2026-07-28
Report a config error for postfix snippets with item scope

### LINKED ISSUE #22894: Warn on "scope":item postfix custom completion snippet (it's not supported)
i just updated specifically for this bug report, i hope VSCodium's update button still works, rust-analyzer version: 0.3.2981-standalone.

1.96.0

VSCodium, NixOS Unstable, standard rust-analyzer extension from open-vsix

(rust-analyzer configuration, specifically the "version" used by vscodium/vscode, you'll need to convert to your editor's config format if not using vscodium or vscode)
```rust
    "rust-analyzer.completion.snippets.custom": {
        "plugin": {
            "postfix": "why",
            "body": [
                "pls work"
            ],
            "description": "Generate a Bevy Plugin impl",
            "scope": "item"
        }
    },
```

i added a custom postfix snippet with item scope using the above example,

expected result:
the snippet exists, it appears in my suggestion box when i type a item, i can select it and it replaces the text i wrote
actual result:
nothing, the completion box doesn't contain the snippet

changing scope to `expr` "fixes" the completion, but then it only works in expr scope, which defeats the entire point of the completion if it generates items.
removing the "scope" field entirely makes it only work in `expr` contexts.

this is also happens in helix, i assume its not IDE specific.

this looks related: https://github.com/rust-lang/rust-analyzer/issues/4360#issuecomment-855455336

Joshuaisikah:
when you  use a Postfix custom snippet with item as the scope, rust-analyzer never really considers this combination, so this  setup previously created a snippet that silently never fired and the user had no way to detect this.

To fix this I decided to add a checker in the config file so that the bad combination is caught early on before  disappearing in the postfix.rs.
I added a test to confirm that this works.
Closes rust-lang/rust-analyzer#22894

### DIFF
--- a/crates/rust-analyzer/src/config.rs
+++ b/crates/rust-analyzer/src/config.rs
@@ -2908,6 +2908,7 @@ enum SnippetScopeDef {
 
 #[derive(Serialize, Deserialize, Debug, Clone, Default)]
 #[serde(default)]
+#[serde(try_from = "SnippetDefRepr")]
 pub(crate) struct SnippetDef {
     #[serde(with = "single_or_array")]
     #[serde(skip_serializing_if = "Vec::is_empty")]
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (changes requested):
This is not a good way to implement that (for example because it does not account for the TOML config above). Instead validate during deserialization. There are crates to do validation with serde, but we don't need them just for this - you can use `#[serde(try_from = "...")]` instead.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22937/ffc323296a3500c6f4546ea6c966e54f4917fedd..1f73f3bf6992028f5518112c817b9093037b6b7d)*

Joshuaisikah:
I have moved the validation into the SnippetDef so now it covers the TOML config too,however I realised that this has one tradeoff where one bad entry now causes deserializing the whole FxIndexMap fail causing it to fall into the inbuilt defaults, but this is still consistent to how other malformen fields behave.

ChayimFriedman2 (approved):
Please squash.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22937/ffc323296a3500c6f4546ea6c966e54f4917fedd..53c90a87981e342ed074e2f388291981bce7d701)*

Joshuaisikah:
Done.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22937/a9c0ce4cb5e3b7c41a7888bf4487d9e72b11e277..fbe2a734ef84d66adf16da72233d314abacd4a13)*

miroo-i-dont-want-to-use-gh:
thanks :D

PR #22939, opened 2026-07-28
internal: update next-solver to 0.166

ada4a:
A small PR to see if I'm doing things right.

This update seems to consist of just https://github.com/rust-lang/rust/pull/156246, which changes `Interner::for_each_{blanket,relevant}_impl` to be able to early-return, using the `VisitorResult` type. My adaptation mostly involved:
- Replacing calls to `.for_each` with for-loops, to allow early-returning (`.try_for_each` unfortunately doesn't work, as `VisitorResult` isn't `Try`)
- Copying the `ret!` macro introduced in the upstream PR. Upstream only needs it in one place, while we need it in four, so I put it into `next_solver/util.rs`. 

cc @ChayimFriedman2

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2062,9 +2062,9 @@ checksum = "f8dcc9c7d52a811697d2151c701e0d08956f92b0e24136cf4cf27b57a6a0d9bf"
 
 [[package]]
 name = "ra-ap-rustc_abi"
-version = "0.165.0"
+version = "0.166.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2f25a779e21ca3bba6795193b16508c8ab159f96ee4b07349893fd272065b525"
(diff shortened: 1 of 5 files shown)


ChayimFriedman2 (approved):
I guess you deliberately didn't upgrade to the last version, to keep it small?

Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22939/30e799799fca143f05d28052c7292d9f2ee57386..a3f0bcfad90f3d765f9cf552a55c2a9834f14780)*

ada4a:
> I guess you deliberately didn't upgrade to the last version, to keep it small?

That, but also because of the fact that, if I were to jump across multiple versions, I'd like to have a commit per version, but then I wouldn't be able to link the relevant PRs, because rustbot would complain...

ChayimFriedman2:
You absolutely can put links in commit messages, you just need them to be full (`rust-lang/rust#number` or `https://github.com/rust-lang/rust/issue/number`).

PR #22940, opened 2026-07-28
fix(vfs): use component-based path prefix matching for virtual paths

cuishuang:
## Summary

Fixed a bug in `VirtualPath::starts_with()` where it used string-based prefix matching instead of component-based path prefix matching. This caused incorrect behavior when one path was a substring prefix of another but not a proper path component prefix.

Before this fix:
- `/foobar`.starts_with(`/foo`) → `true` (incorrect: string prefix but not component prefix)
- `/foo/bar`.starts_with(`/foo`) → `true` (correct)

After this fix:
- `/foobar`.starts_with(`/foo`) → `false` (correct: not a component prefix)
- `/foo/bar`.starts_with(`/foo`) → `true` (correct: proper component prefix)

### DIFF
--- a/crates/vfs/src/vfs_path.rs
+++ b/crates/vfs/src/vfs_path.rs
@@ -337,9 +337,9 @@ impl PartialEq<VfsPath> for AbsPath {
 struct VirtualPath(String);
 
 impl VirtualPath {
-    /// Returns `true` if `other` is a prefix of `self` (as strings).
+    /// Returns `true` if `other` is a prefix of `self`.
     fn starts_with(&self, other: &VirtualPath) -> bool {
-        self.0.starts_with(&other.0)
+        <_ as AsRef<paths::Utf8Path>>::as_ref(&self.0).starts_with(&other.0)
     }
 
     fn strip_prefix(&self, base: &VirtualPath) -> Option<&RelPath> {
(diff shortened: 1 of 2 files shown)


crates/vfs/src/vfs_path.rs
```diff
@@ -337,9 +337,9 @@ impl PartialEq<VfsPath> for AbsPath {
 struct VirtualPath(String);
 
 impl VirtualPath {
-    /// Returns `true` if `other` is a prefix of `self` (as strings).
+    /// Returns `true` if `other` is a prefix of `self`.
     fn starts_with(&self, other: &VirtualPath) -> bool {
-        self.0.starts_with(&other.0)
+        <_ as AsRef<paths::Utf8Path>>::as_ref(&self.0).starts_with(&other.0)
```
ChayimFriedman2, line 342: I wonder if we should store `Utf8Path` instead.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22940/30e799799fca143f05d28052c7292d9f2ee57386..3abeb42ca44b9340101611b462c3a91410c2d983)*

crates/vfs/src/vfs_path.rs
cuishuang: Indeed, using Utf8PathBuf is more appropriate.

Thanks for your review.

PR #22944, opened 2026-07-28
add support for build-dir configuration

### LINKED ISSUE #20150: Cargo `build-dir` support
The Cargo team has done a Call for Testing for the unstable `build-dir` and I'm looking into stabilizing it.  This  allows users to move intermediate artifacts out of their `target-dir`.  With how r-a is using `target-dir` (e.g. rust-lang/rust-analyzer#20072), it seems that `build-dir` will be the more appropriate location moving forward.

Note:
- `build-dir` files do not have compatibility guarantees
- In the future we may change the default `build-dir`
- In the future, we may change the layout of `build-dir`

See https://github.com/rust-lang/cargo/issues/14125#issuecomment-2802967106

icewind1991:
Fixes https://github.com/rust-lang/rust-analyzer/issues/20150

Allows for configuring a rust-analyzer specific `build-dir` just as it already allows for `target-dir`.

Changes mostly consist of duplicating the existing `target-dir` logic, only the logic around setting the option for the spawned commands is different, since there is no `--build-dir` cargo argument.

Has been tested by setting `cargo.buildDir = true` and verifying that the expected subdirectory in the globally configured `build-dir` is created, and that I can run `cargo check` in parallel to the RA check without running into a lock.

### DIFF
--- a/crates/rust-analyzer/src/flycheck.rs
+++ b/crates/rust-analyzer/src/flycheck.rs
@@ -53,6 +53,7 @@ pub(crate) struct CargoOptions {
     pub(crate) extra_env: FxHashMap<String, Option<String>>,
     pub(crate) config_path: Option<AbsPathBuf>,
     pub(crate) target_dir_config: TargetDirectoryConfig,
+    pub(crate) build_dir_config: TargetDirectoryConfig,
 }
 
 #[derive(Clone, Debug)]
@@ -68,6 +69,7 @@ impl CargoOptions {
         &self,
         cmd: &mut Command,
         ws_target_dir: Option<&Utf8Path>,
+        ws_build_dir: Option<&Utf8Path>,
         package_repr: Option<&str>,
         toolchain_version: Option<&semver::Version>,
     ) {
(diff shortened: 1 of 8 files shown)


crates/rust-analyzer/src/flycheck.rs
```diff
@@ -115,6 +117,9 @@ impl CargoOptions {
         if let Some(target_dir) = self.target_dir_config.target_dir(ws_target_dir) {
             cmd.arg("--target-dir").arg(target_dir.as_ref());
         }
+        if let Some(build_dir) = self.build_dir_config.target_dir(ws_build_dir) {
+            cmd.env("CARGO_BUILD_BUILD_DIR", build_dir.as_ref());
```
Veykril, line 120: I take it there is no flag equivalent for this? Bit weird to see target dir and build dir being configured differently 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22944/bec66814323579659ffd77c909b3d963af118ece..f4d61487e34d1aa0fd3827b94ff69d354aecadce)*

icewind1991: Yes, there is no flag to set the build dir at the moment

PR #22947, opened 2026-07-29
hir-ty, ide-diagnostics: use E0057/E0061 for arg-count mismatch (was E0107)

kivancgnlp:
The MismatchedArgCount diagnostic previously used code E0107, which is actually 'wrong number of generic arguments'. Split it based on how the call is made:

- E0057 for calls through the Fn/FnMut/FnOnce traits (arguments bundled into a tuple via TupleArgumentsFlag::TupleArguments in the inference code)
- E0061 for regular function calls

This adds an is_fn_trait_call flag on InferenceDiagnostic::MismatchedArgCount and the hir-surface MismatchedArgCount struct, populated from the tuple_arguments flag already tracked by check_call_arguments. The downstream 'if !args_count_matches' push in infer/expr.rs already covers both paths, so the two FIXMEs at the top of the tuple branch are addressed by threading the kind through rather than by adding a new push site. The nightly-only fallback FIXME below (E0059-ish) is left alone per discussion on issue rust-lang/rust-analyzer#22140.

Adds a test 'arg_count_multi_arg_closure' that exercises the multi-argument tuple case via a closure with signature |_a: u8, _b: u8|. This complements the existing 'arg_count_lambda' test (1-tuple case).

Refs rust-lang/rust-analyzer#22140

### DIFF
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -410,6 +410,11 @@ pub enum InferenceDiagnostic {
         expected: usize,
         #[type_visitable(ignore)]
         found: usize,
+        /// True when the call goes through the `Fn`/`FnMut`/`FnOnce` trait
+        /// (i.e. arguments were bundled into a tuple). Determines whether the
+        /// diagnostic surface uses E0057 (Fn-trait call) or E0061 (regular call).
+        #[type_visitable(ignore)]
+        is_fn_trait_call: bool,
     },
     MismatchedTupleStructPatArgCount {
         #[type_visitable(ignore)]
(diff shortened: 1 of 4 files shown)


ChayimFriedman2 (approved):
Please change the commit message like rustbot said.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22947/126e557c870d1b32f2167a587eb6a386b0e4b0f8..9ab5dcc41e9a5ab997c0fa1d87eda1cda385c129)*

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22947/126e557c870d1b32f2167a587eb6a386b0e4b0f8..a779751464a62e2b94852d9a8f6eb461e2e2c4d0)*

kivancgnlp:
Thanks for the review and the E0057/E0061 pointer.

PR #22964, opened 2026-07-30
fix: show qualified paths when type names collide in E0308

rit3sh-x:
A type mismatch between two same-named types from different modules currently renders with no usable information:

```rust
mod foo { pub struct S; }
mod bar { pub struct S; }

fn test(_: foo::S) {
    test(bar::S);
}
```

Before: `expected S, found S`
After: `expected foo::S, found bar::S`

When both sides of the mismatch render to the same string, they are re-rendered with `display_source_code`, which resolves module paths via `find_path`. The extra work only happens when the strings are equal, so the common path is unchanged. If either side has no renderable path, both keep their plain names, half-qualified message would read worse than the current one.

Nesting falls out of the existing recursion, so `Option<foo::S>` vs `Option<bar::S>` is covered too.

### DIFF
--- a/crates/ide-diagnostics/src/handlers/type_mismatch.rs
+++ b/crates/ide-diagnostics/src/handlers/type_mismatch.rs
@@ -45,23 +45,50 @@ pub(crate) fn type_mismatch(
         cov_mark::hit!(type_mismatch_range_adjustment);
         Some(salient_token_range)
     });
+
+    let expected = d
+        .expected
+        .display(ctx.db(), ctx.display_target)
+        .with_closure_style(ClosureStyle::ClosureWithId)
+        .to_string();
+    let actual = d
+        .actual
+        .display(ctx.db(), ctx.display_target)
+        .with_closure_style(ClosureStyle::ClosureWithId)
+        .to_string();
+
+    // The types differ (that's why we're here), yet they render the same, e.g. `foo::S` and
+    // `bar::S` both render as `S`. Retry with qualified paths so the message isn't a useless
+    // "expected S, found S".
+    let (expected, actual) = if expected == actual {
+        qualified_display(ctx, d).unwrap_or((expected, actual))
+    } else {
+        (expected, actual)
+    };
+
     Some(
         Diagnostic::new(
             DiagnosticCode::RustcHardError("E0308"),
-            format!(
-                "expected {}, found {}",
-                d.expected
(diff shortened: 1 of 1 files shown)


crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -1810,6 +1837,216 @@ fn main() {
...
+    pub struct S;
+}
+fn make() -> [bar::S; 3] { loop {} }
+fn test(_: [foo::S; 3]) {
+    test(make());
+       //^^^^^^ error: expected [foo::S; 3], found [bar::S; 3]
+}
+"#,
+        );
+    }
```
ChayimFriedman2: All those tests can be removed, they're not useful.

crates/ide-diagnostics/src/handlers/type_mismatch.rs
```diff
@@ -1810,6 +1837,216 @@ fn main() {
...
+mod bar {
+    pub struct S;
+}
+fn make() -> Result<Option<bar::S>, i32> { loop {} }
+fn test(_: Result<Option<foo::S>, i32>) {
+    test(make());
+       //^^^^^^ error: expected Result<Option<foo::S>, i32>, found Result<Option<bar::S>, i32>
+}
 "#,
         );
```
ChayimFriedman2: As well as this one.

ChayimFriedman2 (changes requested):
The real fix is to only print the ambiguous paths as fully-qualified, but this is complicated to implement and this PR is a reasonable middle-ground.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22964/d600cff735a7a78f195e0059484820cdc32023c5..378bccb8f40563d21778e9ba6853fa5dce7a36a1)*

rit3sh-x:
@ChayimFriedman2 I shortened the test suite for this implementation. Let me know if you think anything could be improved or changed.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22964/61504e12b88ba8b251aabca786a94973e4010a12..f467deb93bf1c5a8b639c7a6b405815f844e5a6b)*

PR #22968, opened 2026-07-31
Do not use `dyn HirDatabase`, only `dyn SourceDatabase`

ChayimFriedman2:
To save a virtual call.

Reviewing commit-by-commit is recommended. The commit messages are also important.


ChayimFriedman2:
The expected perf benefit is tiny to negligible and this is in direct conflict with the work to split hir-ty. I consider this not worth it.

PR #22971, opened 2026-07-31
fix: support on whitespace inside macro for descend_node_at_offset

### LINKED ISSUE #21729: Work on whitespace for find_node_at_offset_with_descend
`AssistContext::find_node_at_offset_with_descend` Return None on the whitespace within the macro, as the whitespace does not have descended

```rust
macro_rules! m { ($($t:tt)*) => {$($t)*} }
m!(match Some(2) {
    None => 3,
    $0
});
```

`$0` cannot descend a token on WHITESPACE, unless `$0` is on the `}` (currently other bug #21728)

## Problem
Avoid descended cases that do not contain the this whitespace, e.g
call `find_node_at_offset_with_descend::<ast::WildcardPat>()` on `$0  _ => 2`

A4-Tacks:
Fixes rust-lang/rust-analyzer#21729

Example
---
```rust
macro_rules! m { ($expr:expr) => {$expr}}
enum Test { A, B, C }
fn foo(t: Test) {
    m!(match t {
        Test::A => (),
        $0
    });
}
```

**Before this PR**

Assist not applicable

**After this PR**

```rust
macro_rules! m { ($expr:expr) => {$expr}}
enum Test { A, B, C }
fn foo(t: Test) {
    m!(match t {
        Test::A => (),
        Test::B => ${1:todo!()},
        Test::C => ${2:todo!()},$0
    });
}
```

### DIFF
--- a/crates/hir/src/semantics.rs
+++ b/crates/hir/src/semantics.rs
@@ -49,7 +49,7 @@ use stdx::{TupleExt, always};
 use syntax::{
     AstNode, AstPtr, AstToken, Direction, SmolStr, SmolStrBuilder, SyntaxElement, SyntaxKind,
     SyntaxNode, SyntaxNodePtr, SyntaxToken, T, TextRange, TextSize,
-    algo::skip_trivia_token,
+    algo::{skip_trivia_token, skip_whitespace_token},
     ast::{self, HasAttrs as _, HasGenericParams},
 };
 
@@ -1534,7 +1534,20 @@ impl<'db> SemanticsImpl<'db> {
         node: &SyntaxNode,
         offset: TextSize,
     ) -> impl Iterator<Item = impl Iterator<Item = SyntaxNode> + '_> + '_ {
-        node.token_at_offset(offset)
+        let (on_whitespace, tokens) = match node.token_at_offset(offset) {
+            syntax::TokenAtOffset::Single(token) if token.kind() == SyntaxKind::WHITESPACE => {
+                let prev = token
+                    .prev_token()
+                    .and_then(|token| skip_whitespace_token(token, Direction::Prev));
+                let next = token
+                    .next_token()
+                    .and_then(|token| skip_whitespace_token(token, Direction::Next));
+                (true, Either::Left(itertools::chain!([token], prev, next)))
+            }
+            tokens => (false, Either::Right(tokens)),
+        };
+        let file_id = self.find_file(node).file_id;
+        tokens
             .map(move |token| self.descend_into_macros_exact(token))
             .map(|descendants| {
                 descendants.into_iter().map(move |it| self.token_ancestors_with_macros(it))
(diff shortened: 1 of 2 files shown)


crates/hir/src/semantics.rs
```diff
@@ -1534,7 +1534,20 @@ impl<'db> SemanticsImpl<'db> {
...
+                    .and_then(|token| skip_whitespace_token(token, Direction::Prev));
+                let next = token
+                    .next_token()
+                    .and_then(|token| skip_whitespace_token(token, Direction::Next));
+                (true, Either::Left(itertools::chain!([token], prev, next)))
+            }
+            tokens => (false, Either::Right(tokens)),
+        };
+        let file_id = self.find_file(node).file_id;
+        tokens
```
ChayimFriedman2, line 1550: Hmm, not a fan of this implementation, and it also doesn't account for comments.

How about:
```rust
let tokens = node.token_at_offset(offset);
let tokens = if tokens.all(|token| token.kind().is_trivia()) {
    Either::Left(tokens.flat_map(|token| {
                let prev = token
                    .prev_token()
                    .and_then(|token| skip_trivia_token(token, Direction::Prev));
                let next = token
                    .next_token()
                    .and_then(|token| skip_trivia_token(token, Direction::Next));
                std::iter::chain(prev, next)
    }))
} else {
    Either::Right(tokens.filter(|token| token.kind().is_trivia()))
};
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22971/a4b5ba5d74fcadcd983736574c167dc2bf03a401..19ac32f27cc3c2302cca1bbf7e983ab5c0ed1736)*

crates/hir/src/semantics.rs
```diff
@@ -1546,6 +1559,15 @@ impl<'db> SemanticsImpl<'db> {
...
             })
+            .map(move |ancestors| {
+                ancestors.filter(move |node| {
+                    if !on_whitespace {
+                        return true;
+                    }
+                    let origin = self.original_range(node);
+                    origin.file_id == file_id && origin.range.contains(offset)
+                })
+            })
```
ChayimFriedman2, line 1570: This should come before the `kmerge_by()`. Also I think it's better to not check `on_whitespace`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22971/a4b5ba5d74fcadcd983736574c167dc2bf03a401..19ac32f27cc3c2302cca1bbf7e983ab5c0ed1736)*

ChayimFriedman2 (changes requested):

crates/hir/src/semantics.rs
A4-Tacks: > and it also doesn't account for comments.

I think triggering on comments is too surprising

crates/hir/src/semantics.rs
A4-Tacks: If filtering is done before `kmerge`, it will result in excessive redundant calculations for `kmerge`

Checking the `is_whitespace` should prevent performance degradation most of the time and may also avoid some regressions (or fail-fast is better?)

crates/hir/src/semantics.rs
A4-Tacks: This is similar to the behavior of `trimmed_range`

PR #22981, opened 2026-08-01
feat: inlay hints on try-expr

A4-Tacks:
Show try-expr inlay hints (e.g `foo()?;` -> `foo().try?;`).

Similar to `inlay_hints/range_exclusive.rs`

Some people complain that the `?` operation in Rust is not obvious and is easily overlooked when reading the code

Although I do not agree with this viewpoint, I have opened a PR

### DIFF
--- a/crates/ide/src/inlay_hints.rs
+++ b/crates/ide/src/inlay_hints.rs
@@ -44,6 +44,7 @@ mod param_name;
 mod placeholders;
 mod ra_fixture;
 mod range_exclusive;
+mod try_expr;
 
 // Feature: Inlay Hints
 //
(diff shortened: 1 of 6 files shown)


Veykril:
> Some people complain that the ? operation in Rust is not obvious and is easily overlooked when reading the code

I feel like this is better served by those folks adjusting their syntax highlighting then?

Wilfred:
Is this syntactically and semantically correct? Is `foo().try?` equivalent to `foo()?` for both option and result?

I have seen users caught out by `?` being too subtle, but I don't see a nice solution beyond syntax highlighting. In textmate-grammar editors (e.g. VS Code) we use `keyword.operator.question.rust` so you could configure colours for exactly that operator if you wanted.

Veykril:
I'll close this given my previous comment

PR #22993, opened 2026-08-01
fix: recognize format arguments after a backslash in raw strings

### LINKED ISSUE #20034: Highlighter doesn't understand escaped brackets in format!(r##)
**rust-analyzer version**: 0.3.2500-standalone (a207299344 2025-06-15)

**rustc version**: 1.87

```rust
const HERE: &str = "this is a string literal"

fn foo() -> String {
    format!(r#"The brackets \{HERE} are load-bearing"#)
}
```

Hi!
In the example above, my rust-analyzer will "brown out" the {HERE} part even though the \ does not escape the bracket when r## is used.

shulaoda:
Closes rust-lang/rust-analyzer#20034

In `format!(r#"\{HERE}"#)` the backslash does not escape anything, so `{HERE}` is a live format argument, but the highlighter showed the whole string as plain text. `escaped_char_ranges()` ran escape interpretation on raw strings too, which swallowed `\{` as one invalid escape and hid the `{` from the format specifier lexer. Raw strings now report every character as itself.

### DIFF
--- a/crates/ide/src/syntax_highlighting/test_data/highlight_raw_string_format_specifiers.html
+++ b/crates/ide/src/syntax_highlighting/test_data/highlight_raw_string_format_specifiers.html
@@ -1,0 +1,49 @@
+
+<style>
+body                { margin: 0; }
+pre                 { color: #DCDCCC; background: #3F3F3F; font-size: 22px; padding: 0.4em; }
+
+.lifetime           { color: #DFAF8F; font-style: italic; }
+.label              { color: #DFAF8F; font-style: italic; }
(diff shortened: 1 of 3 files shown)


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22993/804ee7d794d4677ba17835653a6b032504839466..e7b86f2e18e083e4d0d0b9de805ec44d87e41c93)*

PR #22994, opened 2026-08-01
fix: term_search exclude useless target type

A4-Tacks:
Example
---
```rust
fn f() { let a: () = (); let b: () = todo$0!() }
```

**Before this PR**

```text
1. Inline macro
2. Replace todo!() with a
3. Replace todo!() with ()
4. Replace todo!() with f()
```

**After this PR**

```text
1. Inline macro
```


kilpkonn:
+1 from me

Veykril (approved):

PR #22996, opened 2026-08-01
fix: detect the rust-analyzer component in a multi-line components array

### LINKED ISSUE #22980: Formatting of `rust-toolchain.toml` should not affect detection of the `rust-analyzer` component in vscode extension
Vscode extension rust-analyzer version | 0.4.2994-standalone (e4e67338ea 2026-07-31)

With a `rust-analyzer` component declared in `rust-toolchain.toml` and installed, the extension still ignores it and starts its bundled server. On a pinned toolchain older than the extension's minimum supported version, that server reports "using an outdated toolchain version", which triggers this modal on every window reload:

> Your workspace uses a rust-toolchain file with a toolchain too old for the extension shipped rust-analyzer to work properly. Consider adding the rust-analyzer component to the toolchain file to use a compatible rust-analyzer version. Add the following to your rust-toolchain file's `[toolchain]` section: `components = ["rust-analyzer"]`

The modal is unactionable: the component *is* declared, it *is* installed, and doing exactly what the message asks changes nothing. Nothing in the "Rust Analyzer Client" output explains why the toolchain binary was skipped, so there is no way to tell from the UI that the extension simply did not see the declaration.

What actually decides this is how the `components` array is formatted. A multi-line array is not detected; the same array on a single line is. Both are valid TOML and rustup treats them identically.

### Steps to reproduce

1. Pin a toolchain older than the extension's minimum supported version, with `components` written as a multi-line array:

   ```toml
   [toolchain]
   channel = "1.88"
   components = [
       "cargo",
       "clippy",
       "rust-analyzer",
       "rust-src",
       "rustc",
       "rustfmt",
   ]
   profile = "default"
   ```

2. `rustup toolchain install 1.88 --component rust-analyzer`
3. Confirm the component is available: `rustup which rust-analyzer` → `~/.rustup/toolchains/1.88-<host>/bin/rust-analyzer`
4. Open the folder in VS Code and reload the window.

**Expected:** the extension uses `~/.rustup/toolchains/1.88-<host>/bin/rust-analyzer`, no warning.

**Actual:** the extension uses the bundled server and shows the "toolchain too old" modal on every reload.

### Workaround

Collapse the array to a single line:

```toml
components = ["cargo", "clippy", "rust-analyzer", "rust-src", "rustc", "rustfmt"]
```

This is fragile in a shared repository — any reformatting back to multi-line silently brings the modal back.

Musteab:
Closes rust-lang/rust-analyzer#22980

`hasToolchainFileWithRaDeclared` matched the `components` array with `.` between the
brackets. `.` does not match newlines, so an array written over several lines was never
detected, even though rustup treats it identically to the single line form. Then extension
then fell back to the bundled server, and on a pinned toolchain older than the extension's
minimum supported version that produces  a "toolchain too old" modal on every reload, with
nothing in the output channel explaining why the toolchain binary was skipped.

The elements are now matched with `[^\]]` instead. It crosses newlines but still stops at the
closing bracket, so the match cannot run past the end of the array into an unrelated key. Two
smaller things came with it: TOML literal strings (`'rust-analyzer'`) are accepted, and the
old `?.length === 1 is gone, and a file containing two matching arrays counted as no match
at all.

Selection behaviour is otherwise unchanged. Detection is still opt-in, only when `components`
explicitly names `rust-analyzer`, and precedence between the toolchain server and the bundled
one is untouched. Having the component installed without declaring it still does not displace
the bundled server.

The predicate is split out as `declaresRaComponent` and exposed through the existing `_private`
object so it can be tested as a plain string check without mocking the filesystem. Tests cover
single line, multi line, literal strings, a components array without rust-analyzer, a file with
no components key, and rust-analyzer appearing outside the array.

🤖 AI-assisted: writing the tests and assisting in the fix

### DIFF
--- a/editors/code/src/bootstrap.ts
+++ b/editors/code/src/bootstrap.ts
@@ -176,14 +176,21 @@ async function fileExists(uri: vscode.Uri) {
     );
 }
 
+// Matches a `components` array that lists `rust-analyzer`. The elements are matched with
+// `[^\]]` rather than `.` so that the array may be spread over several lines, which is just
+// as valid TOML as keeping it on one. TOML strings come in both quote flavours.
+const RA_COMPONENT_RE = /components\s*=\s*\[[^\]]*["']rust-analyzer["'][^\]]*\]/;
+
+function declaresRaComponent(toolchainFileContents: string): boolean {
+    return RA_COMPONENT_RE.test(toolchainFileContents);
+}
+
 async function hasToolchainFileWithRaDeclared(uri: vscode.Uri): Promise<boolean> {
     try {
         const toolchainFileContents = new TextDecoder().decode(
             await vscode.workspace.fs.readFile(uri),
         );
-        return (
-            toolchainFileContents.match(/components\s*=\s*\[.*"rust-analyzer".*\]/g)?.length === 1
-        );
+        return declaresRaComponent(toolchainFileContents);
     } catch (_) {
         return false;
     }
@@ -296,6 +303,7 @@ async function patchelf(dest: vscode.Uri): Promise<void> {
 }
 
 export const _private = {
+    declaresRaComponent,
     earliestToolchainPath,
     orderFromPath,
 };
--- a/editors/code/src/bootstrap.ts
+++ b/editors/code/src/bootstrap.ts
@@ -176,13 +176,16 @@ async function fileExists(uri: vscode.Uri) {
     );
 }
 
-// Matches a `components` array that lists `rust-analyzer`. The elements are matched with
-// `[^\]]` rather than `.` so that the array may be spread over several lines, which is just
-// as valid TOML as keeping it on one. TOML strings come in both quote flavours.
-const RA_COMPONENT_RE = /components\s*=\s*\[[^\]]*["']rust-analyzer["'][^\]]*\]/;
+// Captures the elements of a `components` array. They are matched with `[^\]]` rather than `.`
+// so that the array may be spread over several lines, which is just as valid TOML as keeping it
+// on one, while still stopping at the end of the array.
+const COMPONENTS_RE = /components\s*=\s*\[(?<components>[^\]]*)\]/;
+// TOML strings come in both quote flavours.
+const RA_COMPONENT_RE = /["']rust-analyzer["']/;
 
 function declaresRaComponent(toolchainFileContents: string): boolean {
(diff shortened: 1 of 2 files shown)


editors/code/src/bootstrap.ts
```diff
@@ -176,14 +176,21 @@ async function fileExists(uri: vscode.Uri) {
     );
 }
 
+// Matches a `components` array that lists `rust-analyzer`. The elements are matched with
+// `[^\]]` rather than `.` so that the array may be spread over several lines, which is just
+// as valid TOML as keeping it on one. TOML strings come in both quote flavours.
+const RA_COMPONENT_RE = /components\s*=\s*\[[^\]]*["']rust-analyzer["'][^\]]*\]/;
```
ChayimFriedman2: I'd prefer:
```rust
/components\s*=\s*\[(?<array>.*?)\]*\]/s
```
Then checking if the capture group `array` contains `rust-analyzer`. The current regex is wrong when `rust-analyzer` appears in the rest of the file, not instead this array.

Of course my regex is also wrong, e.g. when `rust-analyzer` appears in a comment. Oh well. The only real fix is to pull a TOML parsing library, but I don't want to do this just for that. I guess my regex is enough for practical needs.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22996/e42c4bce9874f63a53f93885b24543eec43bd1e1..c0ff09b5a3916789c3d31a27fc77fadff983625a)*

ChayimFriedman2 (changes requested):

editors/code/src/bootstrap.ts
Musteab: I switched to the named capture group since it reads better. The previous `[^\]]*` also stopped at the closing `]`, so it could not match a `rust-analyzer` elsewhere in the file. I tightened the test to cover a quoted mention after the array, and kept the quote check instead of `includes()` so unquoted mentions do not count. Happy to change that too if you prefer.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/22996/e42c4bce9874f63a53f93885b24543eec43bd1e1..c88336732fc71d0a488a57a56a782f688386eb77)*

PR #23002, opened 2026-08-02
bump gen-lsp-types to 0.11.0

bend-n:
this adds Custom's to every enum (thereby not simply dying if the client sends a custom thing)

### DIFF
--- a/crates/rust-analyzer/src/handlers/request.rs
+++ b/crates/rust-analyzer/src/handlers/request.rs
@@ -780,10 +780,7 @@ pub(crate) fn handle_will_rename_files(
     let source_changes: Vec<SourceChange> = params
         .files
         .into_iter()
-        .filter_map(|file_rename| {
-            let from = Uri::parse(&file_rename.old_uri).ok()?;
-            let to = Uri::parse(&file_rename.new_uri).ok()?;
-
+        .filter_map(|lsp_types::FileRename { new_uri: to, old_uri: from }| {
             let from_path = from.to_file_path().ok()?;
             let to_path = to.to_file_path().ok()?;
 
@@ -2726,6 +2723,7 @@ fn resource_ops_supported(config: &Config, kind: &DocumentChange) -> anyhow::Res
                     ResourceOperationKind::Create => "create",
                     ResourceOperationKind::Rename => "rename",
                     ResourceOperationKind::Delete => "delete",
+                    ResourceOperationKind::Custom(_) => unreachable!(),
                 }
             ),
         )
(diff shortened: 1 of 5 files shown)


ChayimFriedman2:
Why would clients do that?

bend-n:
the client may, potentially, be running on a newer version of the specification than the older rust-analyzer, and the spec says unknown values should be handled gracefully.

crates/rust-analyzer/src/handlers/request.rs
```diff
@@ -2726,6 +2723,7 @@ fn resource_ops_supported(config: &Config, kind: &DocumentChange) -> anyhow::Res
                     ResourceOperationKind::Create => "create",
                     ResourceOperationKind::Rename => "rename",
                     ResourceOperationKind::Delete => "delete",
+                    ResourceOperationKind::Custom(_) => unreachable!(),
```
ChayimFriedman2, line 2726: If the point is to support unknown operations, panicking is not a good idea. So just return an error.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23002/3f98c4dfcf765d096148401d423eabb3a8611f9f..a0b32c69b060116b7bf894504984be5d73f30121)*

ChayimFriedman2 (changes requested):

crates/rust-analyzer/src/handlers/request.rs
bend-n: well, not really. it is literally unreachable.
```rs
    let op = match kind {
        lsp_types::DocumentChange::CreateFile(_) => ResourceOperationKind::Create,
        lsp_types::DocumentChange::RenameFile(_) => ResourceOperationKind::Rename,
        lsp_types::DocumentChange::DeleteFile(_) => ResourceOperationKind::Delete,
        lsp_types::DocumentChange::TextDocumentEdit(_) => return Ok(()),
    };
    if !matches!(config.workspace_edit_resource_operations(), Some(resops) if resops.contains(&op))
    {
        return Err(LspError::new(
            ErrorCode::RequestFailed as i32,
            format!(
                "Client does not support {} capability.",
                match op {
                    ResourceOperationKind::Create => "create",
                    ResourceOperationKind::Rename => "rename",
                    ResourceOperationKind::Delete => "delete",
                    ResourceOperationKind::Custom(_) => unreachable!(), 
                }
            ),
        )
```

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23002/3f98c4dfcf765d096148401d423eabb3a8611f9f..a0b32c69b060116b7bf894504984be5d73f30121)*

PR #23003, opened 2026-08-02
fix: add reference for same name param coerce matches

A4-Tacks:
Example
---
```rust
fn ref_arg(x: &[i32]) {}
fn foo(x: [i32; 2]) {
    ref_ar$0
}
```

**Before this PR**

```rust
fn ref_arg(x: &[i32]) {}
fn foo(x: [i32; 2]) {
    ref_arg(${1:x});$0
}
```

**After this PR**

```rust
fn ref_arg(x: &[i32]) {}
fn foo(x: [i32; 2]) {
    ref_arg(${1:&x});$0
}
```

### DIFF
--- a/crates/ide-completion/src/render/function.rs
+++ b/crates/ide-completion/src/render/function.rs
@@ -233,7 +233,8 @@ pub(super) fn add_call_parens<'b>(
                         Some(n) => {
                             let smol_str = n.display_no_db(ctx.edition).to_smolstr();
                             let text = smol_str.as_str().trim_start_matches('_');
-                            let ref_ = ref_of_param(ctx, text, param.ty());
+                            let ref_ =
+                                ref_of_param(ctx, text, &param.ty().instantiate_with_errors());
                             f(&format_args!("${{{}:{ref_}{text}}}", index + offset))
                         }
                         None => {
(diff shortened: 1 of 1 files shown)


crates/ide-completion/src/render/function.rs
```diff
@@ -288,14 +288,16 @@ pub(super) fn add_call_parens<'b>(
...
+        let ref_prefix = if mutability.is_mut() { "&mut " } else { "&" };
+
         for (name, local) in ctx.locals.iter().sorted_by_key(|&(k, _)| k.clone()) {
             if name.as_str() == arg {
-                return if local.ty(ctx.db) == derefed_ty {
-                    if ty.is_mutable_reference() { "&mut " } else { "&" }
-                } else {
-                    ""
-                };
+                let local_ty = local.ty(ctx.db).instantiate_with_errors();
```
ChayimFriedman2: You shouldn't need `instantiate_with_errors()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23003/3f98c4dfcf765d096148401d423eabb3a8611f9f..0c856a60f27f661c8e405dc8d7dc1e3664c8fd14)*

A4-Tacks: I need, otherwise it will panic

ChayimFriedman2: Hmm, where does it panic? `local.ty()` is inside the context of the completion body, so some *caller* probably needs to instantiate the type it passes.

A4-Tacks: It seems reasonable on param, but I can't seem to use `instantiate()` instead of `instantiate_with_errors()`

I don't know if I understand.
Ideally, if I didn't instantiate the param, I would expect a panic? Why only instantiate local won't cause panic

ChayimFriedman2: Yes I meant `instantiate_with_errors()`.

The param comes from the context of the called function. If it has generics, they are not necessarily the same as the generics in the caller. Which is why we should instantiate it.

ChayimFriedman2 (approved):

PR #23011, opened 2026-08-03
fix: no `unused_variable` diagnostic for never-type locals

### LINKED ISSUE #22975: False positive unused_variables on never-type
**rust-analyzer version**: rust-analyzer 0.4.2991-standalone

**rustc version**: rustc 1.97.0-nightly (36ba2c771 2026-04-23)

**editor or extension**: Vim

```rust
fn main() {
    let foo = loop {};
    _ = foo; // used
}
```

Cursor on `foo` (line 2)

Expect no this quickfix, but actual output `Rename unused foo to _foo`

After apply:

```rust
fn main() {
    let _foo = loop {};
    _ = foo; // used
}
```

Should either 'not applicable' or 'rename all usage'

Albab-Hasan:
fixes https://github.com/rust-lang/rust-analyzer/issues/22975

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -2141,7 +2141,7 @@ impl DefWithBody {
                         // We should report specific diagnostics for these problems, not `need-mut` and `unused-mut`.
                         continue;
                     }
-                    let Some(&local) = mir_body.binding_locals.get(binding_id) else {
+                    let Some(&mir_local) = mir_body.binding_locals.get(binding_id) else {
                         continue;
                     };
                     if source_map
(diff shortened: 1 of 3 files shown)


ChayimFriedman2:
This is a hack and I don't want this.

This diagnostic is hopeless. Its only real solution is to rewrite MIR and borrowck which is not a short-term thing. I'd be very open to disabling it by default or completely, but not to add hacks to it.

Albab-Hasan:
if you dont mind can i reopen a few of my prs that i ghosted and ping you there for review?

ChayimFriedman2:
We removed this diagnostic.

PR #23013, opened 2026-08-03
Update salsa to 0.28

ada4a:
To simplify work on https://github.com/rust-lang/rust-analyzer/issues/22868. Ref: [#t-compiler/rust-analyzer > Updating salsa to 0.28](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Updating.20salsa.20to.200.2E28/with/613878762)

### DIFF
--- a/crates/hir-ty/src/lower.rs
+++ b/crates/hir-ty/src/lower.rs
@@ -8,7 +8,7 @@
 pub(crate) mod diagnostics;
 pub(crate) mod path;
 
-use std::{cell::OnceCell, iter, mem, sync::OnceLock};
+use std::{cell::OnceCell, iter, mem, ops::Deref, sync::OnceLock};
 
 use either::Either;
 use hir_def::{
@@ -50,7 +50,7 @@ use rustc_type_ir::{
     UpcastFrom, elaborate,
     inherent::{Clause as _, GenericArgs as _, IntoKind as _, Region as _, Ty as _},
 };
-use salsa::Update;
+use salsa::SalsaValue;
 use smallvec::SmallVec;
 use stdx::{impl_from, never};
 use thin_vec::ThinVec;
@@ -78,7 +78,7 @@ use crate::{
 
 pub(crate) struct PathDiagnosticCallbackData(pub(crate) TypeRefId);
 
-#[derive(PartialEq, Eq, Debug, Hash)]
+#[derive(PartialEq, Eq, Debug, Hash, SalsaValue)]
 pub struct WithDefinedOpaques<T> {
     value: T,
     impl_traits: Option<Box<Arena<ImplTrait>>>,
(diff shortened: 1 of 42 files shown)


crates/hir-ty/src/lower.rs
```diff
@@ -1370,14 +1370,16 @@ impl<'db, 'a> TyLoweringContext<'db, 'a> {
     }
 }
 
-#[derive(Clone, PartialEq, Eq, Update)]
+#[derive(Clone, PartialEq, Eq, SalsaValue)]
 pub struct TyLoweringResult<'db, T> {
-    #[update(fallback)]
+    // SAFETY: required for `T = la_arena::ArenaMap`.
+    // It's safe to retain as it doesn't have any `'db` lifetimes
+    #[salsa_value(unsafe(prove_safe_to_retain_manually))]
```
Veykril: I think we should drop this, and newtype the one place where we store a la_arena, that is `field_types_with_diagnostics` should wrap the arena in a new type that instead implements SalsaValue unsafely with a `'static` proof

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23013/83a2d22933a94e27b03ce9c88d87f87e21ebe371..4ddf64a71ee311c8e87be78fffb8be3b19ea322e)*

crates/ide-db/src/symbol_index.rs
```diff
@@ -367,22 +367,8 @@ pub struct SymbolIndex<'db> {
...
-        if *this != new_value {
-            *this = new_value;
-            true
-        } else {
-            false
-        }
-    }
-}
+// - FileSymbol<'db>: SalsaValue
+unsafe impl<'db> SalsaValue for SymbolIndex<'db> where FileSymbol<'db>: SalsaValue {}
```
Veykril: Can this be a derive now?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23013/83a2d22933a94e27b03ce9c88d87f87e21ebe371..4ddf64a71ee311c8e87be78fffb8be3b19ea322e)*

crates/hir-ty/src/lower.rs
ada4a: note that `const_param_types_with_diagnostics` uses an la_arena as well. I'll make two newtypes then

Veykril: we might not even need the 'static proof given how the derive works I think if we are using concrete types and no generics in the new types

ada4a: Wait, I just noticed that we own `la_arena` -- can't we just put the derive on `ArenaMap`, in that case?

Veykril: then we would require la_arena to depend on salsa. That's not great as la_arena is meant to be a standalone lib for others to use.

Veykril (approved):

crates/hir-ty/src/lower.rs
ada4a: We could hide it behind an optional feature, like many crates do with serde.. But I get your point

Veykril: we could yea, and I never really like that. So given our case here we can avoid it as it doesn't add that much cruft to our codebase

PR #23014, opened 2026-08-03
Allow `self` as the last segment of a path

### LINKED ISSUE #23000: FP syntax error `use std::fmt::self;`
```rust
use std::fmt::self;
fn main () {}
```
this compiles but ra diagnostics say
```
at crate ra, file /tmp/ra/src/main.rs: Error SyntaxError from LineCol { line: 0, col: 14 } to LineCol { line: 0, col: 18 }: Syntax Error: The `self` keyword is only allowed as the first segment of a path
```

hongjr03:
Closes rust-lang/rust-analyzer#23000.

### DIFF
--- a/crates/hir-expand/src/mod_path.rs
+++ b/crates/hir-expand/src/mod_path.rs
@@ -315,6 +315,7 @@ fn convert_path(
     for segment in segments {
         let name = match segment.kind()? {
             ast::PathSegmentKind::Name(name) => name.as_name(),
+            ast::PathSegmentKind::SelfKw => continue,
             _ => return None,
         };
         mod_path.segments.push(name);
--- a/crates/syntax/src/validation.rs
+++ b/crates/syntax/src/validation.rs
@@ -285,11 +285,11 @@ fn validate_range_expr(expr: ast::RangeExpr, errors: &mut Vec<SyntaxError>) {
 fn validate_path_keywords(segment: ast::PathSegment, errors: &mut Vec<SyntaxError>) {
     let path = segment.parent_path();
     let is_path_start = segment.coloncolon_token().is_none() && path.qualifier().is_none();
-
     if let Some(token) = segment.self_token() {
-        if !is_path_start {
+        let is_last_segment = path.parent_path().is_none();
+        if !is_path_start && !is_last_segment {
             errors.push(SyntaxError::new(
-                "The `self` keyword is only allowed as the first segment of a path",
+                "The `self` keyword is only allowed at the start or at the end of a path",
                 token.text_range(),
             ));
         }
--- a/crates/hir-def/src/nameres/tests/imports.rs
+++ b/crates/hir-def/src/nameres/tests/imports.rs
@@ -41,6 +41,26 @@ mod foo {
 }
 
 #[test]
+fn trailing_self_import() {
+    check(
+        r#"
+mod fmt {
+    pub struct Thing;
+}
+pub use fmt::self as alias;
+"#,
+        expect![[r#"
+            crate
+            - alias : type (import)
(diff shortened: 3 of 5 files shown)


crates/syntax/src/validation.rs
```diff
@@ -285,11 +285,16 @@ fn validate_range_expr(expr: ast::RangeExpr, errors: &mut Vec<SyntaxError>) {
...
     let is_path_start = segment.coloncolon_token().is_none() && path.qualifier().is_none();
-
     if let Some(token) = segment.self_token() {
-        if !is_path_start {
+        let mut outer = path;
+        while let Some(parent) = outer.syntax().parent().and_then(ast::Path::cast) {
+            outer = parent;
+        }
+        let is_last_segment =
+            outer.qualifier().is_some() && outer.segments().clone().last() == Some(segment.clone());
```
ChayimFriedman2:
```suggestion
        let is_last_segment = path.parent_path().is_none();
```
Not only this is less efficient, this is also incorrect as `use self as foo` is valid.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23014/83a2d22933a94e27b03ce9c88d87f87e21ebe371..9453873dcdea8bbefb2b47f74fa8b13acae5e8dd)*

crates/hir-expand/src/mod_path.rs
```diff
@@ -312,12 +312,20 @@ fn convert_path(
...
+                } else {
+                    // `foo::self::bar` is invalid
+                    None
+                };
+            }
             _ => return None,
-        };
-        mod_path.segments.push(name);
+        }
     }
```
ChayimFriedman2:
```suggestion
    for segment in segments {
        let name = match segment.kind()? {
            ast::PathSegmentKind::Name(name) => name.as_name(),
            ast::PathSegmentKind::SelfKw => continue,
            _ => return None,
        };
        mod_path.segments.push(name);
    }
```
Since putting `self` in the middle is invalid (and also has an error), we're free to handle it as we like, and this is simpler.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23014/83a2d22933a94e27b03ce9c88d87f87e21ebe371..9453873dcdea8bbefb2b47f74fa8b13acae5e8dd)*

ChayimFriedman2 (changes requested):

ChayimFriedman2 (approved):
Please squash.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23014/83a2d22933a94e27b03ce9c88d87f87e21ebe371..574d7da78d237bc2a7c73322a2615ba1fb0fc83e)*

crates/hir-expand/src/mod_path.rs
hongjr03: ~~This looks good but wouldn't this swallow the `foo::self::bar` error?~~ Sorry, I misunderstood.

hongjr03:
I also updated the modified snapshot so this line now passes the test.

https://github.com/rust-lang/rust-analyzer/pull/23014/changes#diff-614ce7f9d9bd1688871d8a961d85d217690e09b89d569312d6002aa83a58401aR2

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23014/eedfe66ee6fc8d727bc99e7e2c264226416cfd57..d8d1b9c6b210c41e046e6c59f4ea6d8805d3f10a)*

PR #23016, opened 2026-08-03
fix: resolve assignment lhs in its expression scope 

### LINKED ISSUE #22988: FP `invalid left-hand side of assignment` function and enclosed var in closure same name
this only repros when the `fn` and the var which is then used in the closure have the same name
```rust
fn i() {
    let mut i = 0;
    (|| i += 1 )();
}
fn main() {}
```
^ this compiled but not for ra diagnostics:
```
at crate ra, file /tmp/ra/src/main.rs: Error RustcHardError("E0067") from LineCol { line: 2, col: 8 } to LineCol { line: 2, col: 9 }: invalid left-hand side of assignment
```

so1ve:
fixes rust-lang/rust-analyzer#22988

### DIFF
--- a/crates/hir-ty/src/infer/expr.rs
+++ b/crates/hir-ty/src/infer/expr.rs
@@ -220,15 +220,26 @@ impl<'db> InferenceContext<'db> {
         result
     }
 
-    fn is_syntactic_place_expr(&self, expr: ExprId) -> bool {
+    fn is_syntactic_place_expr(&mut self, expr: ExprId) -> bool {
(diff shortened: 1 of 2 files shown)


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23016/17a08208b732ec9fa497322bcdfd5961e13925e1..92651c1032a66a447ccef2b6c9ff6a174e0085b0)*

PR #23026, opened 2026-08-04
return both matching braces

bend-n:
so, this is technically the same behavior as simply calling a matching brace on the result of itself, but it saves a request, and the result is trivially computable, so you may as well return both? this makes it trivial to implement "remove brackets" behavior.

### DIFF
--- a/crates/ide/src/lib.rs
+++ b/crates/ide/src/lib.rs
@@ -385,7 +385,10 @@ impl Analysis {
 
     /// Returns position of the matching brace (all types of braces are
     /// supported).
-    pub fn matching_brace(&self, position: FilePosition) -> Cancellable<Option<TextSize>> {
+    pub fn matching_brace(
+        &self,
+        position: FilePosition,
+    ) -> Cancellable<Option<(TextSize, TextSize)>> {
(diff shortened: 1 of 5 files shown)


ChayimFriedman2 (changes requested):
You didn't change the TypeScript extension.

Also, I'm not sure we want this. @rust-lang/rust-analyzer?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23026/5f258f4534e3b4bdaa45a1299b53a66cf014d803..5ef04082040202c48ff4623701a6971e955ac749)*

Wilfred:
It doesn't seem terribly useful. If I'm on the open brace, seeing the closing brace is useful, but I already know where the open brace is?

bend-n:
it acts when you arent on a brace

Veykril:
The thought here is for the fallback, if you aren't on a brace we have a fallback implemented that returns the opening brace of the innermost enclosing node that has one. With this we return both of those. I think that's reasonable. It is a breaking change as well though, so we should at least hookup the vscode side here

PR #23027, opened 2026-08-04
feat: new assist `reorder_match_arms`

### LINKED ISSUE #8475: Reorder match arms
It would be nice if there was an assist to reorder match arms based on the order of variants within the enum definition. I.e. change this:

```rust
enum Foo {
    A,
    B,
    C,
}

fn bar(foo: Foo) -> i32 {
    match foo {
        Foo::C => 3,
        Foo::B => 2,
        Foo::A => 1,
    }
}
```

to this:

```rust
enum Foo {
    A,
    B,
    C,
}

fn bar(foo: Foo) -> i32 {
    match foo {
        Foo::A => 1,
        Foo::B => 2,
        Foo::C => 3,
    }
}
```

Alternatively / in addition to this, the "fill match arms" assist could insert new arms into the correct position, rather than always placing them at the end.

so1ve:
resolve rust-lang/rust-analyzer#8475

Tests were conducted under my guidance, with various boundary conditions defined, and were generated using GPT 5.6 Sol.

### DIFF
--- a/crates/ide-assists/src/handlers/reorder_match_arms.rs
+++ b/crates/ide-assists/src/handlers/reorder_match_arms.rs
@@ -1,0 +1,641 @@
+use hir::{Adt, EnumVariant, ModuleDef, PathResolution};
+use ide_db::FxHashMap;
+use syntax::{
+    AstNode, SyntaxElement, SyntaxKind, SyntaxNode, T, TextRange,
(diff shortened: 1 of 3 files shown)


so1ve:
Ah, forgot that RA has now enacted feature freeze for IDE assists 😅 I'll close this PR for now

PR #23028, opened 2026-08-04
fix: preserve trailing text when `InsertReplaceEdit` is unsupported

### LINKED ISSUE #21365: Completion replaces the word cursor is at instead of inserting
**rust-analyzer version**: commit 11d62122ded8349b2790977fd2cc55013cea68f1

**rustc version**: rustc 1.92.0

**editor or extension**: I use NeoVim with vim-lsp but this should affect everything

**relevant settings**: None

**repository link (if public, optional)**: None

Code snippet is based on this test
https://github.com/rust-lang/rust-analyzer/blob/11d62122ded8349b2790977fd2cc55013cea68f1/crates/ide-completion/src/tests/special.rs#L15

You can also reproduce using this test.

```rust
fn main() {
    let wherewolf = 92;
    drop(where$0sheep) // Note that compared to original test, `sheep` is added behind the cursor
}
```

After the text edit from the completion, it's expected that the code becomes
```rust
fn main() {
    let wherewolf = 92;
    drop(wherewolfsheep)
}
```

However, the actual behavior is that, the sheep gets eaten by the wolf:
```rust
fn main() {
    let wherewolf = 92;
    drop(wherewolf)
}
```

I thought this current behavior was the expected behavior, because it's the default behavior of Intellij IDEA (and [has been hated by people for years](https://stackoverflow.com/questions/35424303/intellij-autocompletion-replacing-the-subsequent-word)), but looking at [this comment](https://github.com/rust-lang/rust-analyzer/blob/11d62122ded8349b2790977fd2cc55013cea68f1/crates/ide-completion/src/item.rs#L38-L40) makes me think this might be unintentional:

> `text_edit` should start with what `source_range` points to

 as in this case, `source_range` points to `wheresheep`, but `text_edit` is `wherewolf`.

I think the problem here is that, here

https://github.com/rust-lang/rust-analyzer/blob/11d62122ded8349b2790977fd2cc55013cea68f1/crates/ide-completion/src/render.rs#L508-L510

the text to delete in this completion is decided to be `ctx.source_range()`, which is determined here

https://github.com/rust-lang/rust-analyzer/blob/11d62122ded8349b2790977fd2cc55013cea68f1/crates/ide-completion/src/context.rs#L514

to be the `original_token`'s `text_range()`, and the `original_token` has been decided here

https://github.com/rust-lang/rust-analyzer/blob/11d62122ded8349b2790977fd2cc55013cea68f1/crates/ide-completion/src/context.rs#L724

to be `wheresheep` instead of `where`.

There's also a discussion about this: https://github.com/rust-lang/rust-analyzer/discussions/21029, where an example that better illustrates why this behavior is annoying in practice is given.

so1ve:
fixes rust-lang/rust-analyzer#21365

### DIFF
--- a/crates/rust-analyzer/src/lsp/to_proto.rs
+++ b/crates/rust-analyzer/src/lsp/to_proto.rs
@@ -183,18 +183,21 @@ pub(crate) fn text_edit(line_index: &LineIndex, indel: Indel) -> lsp_types::Text
 
 pub(crate) fn completion_text_edit(
     line_index: &LineIndex,
-    insert_replace_support: Option<lsp_types::Position>,
+    cursor_position: lsp_types::Position,
+    insert_replace_support: bool,
     indel: Indel,
 ) -> lsp_types::CompletionItemTextEdit {
     let text_edit = text_edit(line_index, indel);
-    match insert_replace_support {
-        Some(cursor_pos) => lsp_types::InsertReplaceEdit {
+    let insert = lsp_types::Range { start: text_edit.range.start, end: cursor_position };
+    if insert_replace_support {
+        lsp_types::InsertReplaceEdit {
             new_text: text_edit.new_text,
-            insert: lsp_types::Range { start: text_edit.range.start, end: cursor_pos },
+            insert,
             replace: text_edit.range,
         }
-        .into(),
-        None => text_edit.into(),
+        .into()
+    } else {
+        lsp_types::TextEdit { range: insert, new_text: text_edit.new_text }.into()
     }
 }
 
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
I'm not sure if insertion is better than replacement, but since people seem to think that the answer is yes, it's fine from my side.

Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23028/5f258f4534e3b4bdaa45a1299b53a66cf014d803..f2c57c815607273c9061a330303df9d63ffffc78)*

PR #23030, opened 2026-08-04
feat: rename references in intra-doc links

### LINKED ISSUE #12583: Update comments that reference refactored symbols
In vscode, you can click <kbd>F2</kbd> to rename a module item. In some cases, the module item is referenced with square brackets in one of the comments. For example

```
// ... some where in your crate
//! This module uses [Builder] to do foo.


// ... else where
struct Builder {
 input_path: PathBuf,
 output_path: PathBuf
}
```

Later you decide to refactor `Builder`, for example, by giving it a more special name. You would want the comments that reference the `Builder` to be refactored, ideally. It would be nice for Rust-Analyzer to locate and update those references and update them when the code is refactored.

so1ve:
resolve rust-lang/rust-analyzer#12583 

One point worth noting is that I deliberately added an `include_intra_doc_links` option to control whether intra-doc links are included in usage search results. For now, I have enabled it only for rename, because I suspect that including documentation links in other features, such as Find References, could make their results too noisy. I have also deliberately left renaming links inside macros unsupported for now.


crates/ide-db/src/documentation/intra_doc_links.rs
```diff
@@ -0,0 +1,286 @@
...
+            }) {
+                Some(&s[prefix.len() + 1..])
+            } else {
+                suffixes.iter().find_map(|&suffix| s.strip_suffix(suffix))
+            }
+        })
+        .unwrap_or(s)
+}
+
+/// Extracts all intra-doc link occurrences from Markdown documentation.
```
ChayimFriedman2, line 68:
```suggestion
/// Extracts all links from a given markdown text returning the definition text range, link-text
/// and the namespace if known.
```
You removed this and this is useful.

crates/ide-db/src/documentation/intra_doc_links.rs
```diff
@@ -0,0 +1,286 @@
...
+            }
+            segment_start = segment_end + 2;
+        }
+    }
+
+    res
+}
+
+fn broken_link_clone_cb(link: BrokenLink<'_>) -> Option<(CowStr<'_>, CowStr<'_>)> {
+    Some((link.reference.clone(), link.reference))
```
ChayimFriedman2, line 264: Duplicated with doc_links.rs, you also removed the comment.

crates/ide-db/src/documentation/intra_doc_links.rs
```diff
@@ -0,0 +1,286 @@
...
+use syntax::{AstNode, SyntaxNode, TextRange, TextSize, ast, match_ast};
+
+use crate::{
+    EditionedFileId, RootDatabase,
+    defs::Definition,
+    documentation::{Documentation, HasDocs},
+};
+
+const MARKDOWN_OPTIONS: Options =
+    Options::ENABLE_FOOTNOTES.union(Options::ENABLE_TABLES).union(Options::ENABLE_TASKLISTS);
```
ChayimFriedman2, line 15: Also duplicated with doc_links.rs.

ChayimFriedman2 (changes requested):
This all seems exceedingly complicated way more than it needs to be.

All you need to do is to check for doc comments inside`FindUsages::search()` in the search loop, similar to what we do for goto def.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23030/36896409dc6dbf741f487b84fc084cdf0e7876e6..604d07b05a45add6f732caddf13d7c438c9f3894)*

so1ve:
Hmm. I made some changes locally but I have a question: should I inlcude intra docs in other usages like go to references?

ChayimFriedman2:
I think we should start with not. It's reasonable to do, but we'll want a config for it anyway, so not for this PR.

so1ve:
The current implementation is indeed fairly complicated 🤔 However I noticed one issue: FindUsages has a fast path for short inherent associated unctions such as new and from

After the refactoring, in order to handle intra-doc links in the normal search loop, rename needs to bypass this fast path when intra-doc link search is enabled. Is that performance trade-off acceptable for this PR or should the fast path also be adapted to handle intra-doc links during rename?

ChayimFriedman2:
You don't need to disable this, you just need to also search doc comments (and string literals in `#[doc = "..."]`) in addition. Since text searching and parsing are very fast and not the bottleneck, this will not make the fast path slower.

ChayimFriedman2:
Actually, you don't even need an addition search - the existing short associated function path could find doclinks (textually) just as well.

PR #23033, opened 2026-08-04
remove deprecated authors field from package manifests

BenjaminBrienen:
see: https://rust-lang.github.io/rfcs/3052-optional-authors-field.html


ChayimFriedman2:
Why? I don't think it's deprecated, just optional, and this will prevent compiling with older versions.

BenjaminBrienen:
Oh, the diagnostic was from Tombi, not r-a

PR #23042, opened 2026-08-05
feat: Support `#[rustc_must_implement_one_of]` in the assists

ChayimFriedman2:
We still don't support it in the diagnostics for missing items, that's for later PR.

### DIFF
--- a/crates/hir-def/src/attrs.rs
+++ b/crates/hir-def/src/attrs.rs
@@ -23,6 +23,7 @@ use either::Either;
 use hir_expand::{
     InFile, Lookup,
     attrs::{AstKeyValueMetaExt, AstPathExt, expand_cfg_attr},
+    name::Name,
 };
 use intern::Symbol;
 use itertools::Itertools;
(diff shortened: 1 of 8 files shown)


crates/hir-def/src/attrs.rs
```diff
@@ -1215,6 +1221,36 @@ impl AttrFlags {
...
+                        .map(|name| Name::new_symbol_root(Symbol::intern(name.text())))
+                        .collect::<Box<[_]>>()
+                    && !names.is_empty()
+                {
+                    return ControlFlow::Break(names);
+                }
+                ControlFlow::Continue(())
+            })
+        }
+    }
```
Veykril, line 1253: I've been wondering, feels like salsa would benefit from a first class feature for this

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23042/fb2e61524bac23017ff00da0e1c4d890cf9ed618..70deecb629ebbc2c7e88d847a0ec40e6e8d7d4d7)*

Veykril (approved):

crates/hir-def/src/attrs.rs
ChayimFriedman2: Maybe, although it's not clear what shape it'll take - what you query to check if we should not populate the query?

PR #23047, opened 2026-08-05
internal: use `rustc_type_ir::try_visit`

ada4a:
..instead of manual reimplementation

See https://github.com/rust-lang/rust/pull/160568 which does the same thing in rustc


ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23047/de9c071849e84653dc78efa84a7518aca943294d..d713a19d7a1ae854126812dcc4eced801ccd73a9)*

PR #23056, opened 2026-08-05
perf: Optimize memory usage of the item tree

ChayimFriedman2:
Do not use hash maps (which consume more memory), instead use bare `ThinVec`s and indexing.

This saves 28mb on self and 59mb on buck2. Not a lot, but significant, the impact on maintainability is practically zero, and furthermore, unlike e.g. optimizations to hir-ty, all item trees are always kept in memory even in IDE scenarios, so the save is real.

This has just a little bit impact on incrementality: now if you reorder items the item tree is invalidated (meaning the def map is invalidated, but not other things). However the item tree is easily invalidated, e.g. by adding a `use`. So this is not a problem.

### DIFF
--- a/crates/span/src/ast_id.rs
+++ b/crates/span/src/ast_id.rs
@@ -59,53 +59,13 @@ pub struct ErasedFileAstId(u32);
 
 impl fmt::Debug for ErasedFileAstId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let kind = self.kind();
-        macro_rules! kind {
-            ($($kind:ident),* $(,)?) => {
-                if false {
-                    // Ensure we covered all variants.
-                    match ErasedFileAstIdKind::Root {
-                        $( ErasedFileAstIdKind::$kind => {} )*
-                    }
-                    unreachable!()
-                }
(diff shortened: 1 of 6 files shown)


Veykril:
> Not a lot

I beg to differ, thats quite a lot of savings

Veykril (approved):

crates/span/src/ast_id.rs
```diff
@@ -293,6 +293,11 @@ impl<N> FileAstId<N> {
         FileAstId { raw: self.raw, _marker: PhantomData }
     }
 
+    #[inline]
+    pub fn downcast_unchecked<M: AstIdNode + Into<N>>(self) -> FileAstId<M> {
```
Veykril, line 331: can we add a debug_assert for this (assuming we have the relevant info at this layer)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23056/cf3b451f47c01d76989425d1ace3f1cba70f1fdb..b5d48f7b27a96cb38569bcef3d3427b36da50f06)*

ChayimFriedman2: Okay, done that.

PR #23060, opened 2026-08-06
Add `'db` lifetime to `BuiltinDeriveMacroId`

ada4a:
Part of https://github.com/rust-lang/rust-analyzer/issues/22868


crates/hir-def/src/attrs/docs.rs
```diff
@@ -420,7 +421,21 @@ fn extend_with_attrs<'a, 'db>(
     make_resolver: &dyn Fn() -> Resolver<'db>,
 ) {
     // Lazily initialised when we first encounter a `#[doc = macro!()]`.
-    let mut expander: Option<DocMacroExpander<'db>> = None;
+    let mut expander = LazyCell::new(|| {
```
ChayimFriedman2: Why this change? (`LazyCell` has an additional branch).

crates/hir-def/src/nameres/path_resolution.rs
```diff
@@ -761,11 +761,11 @@ impl DefMap {
 
 /// Given a block module, returns its nearest non-block module and the `DefMap` it belongs to.
 #[inline]
-fn adjust_to_nearest_non_block_module<'db>(
+fn adjust_to_nearest_non_block_module<'db, 'dm>(
     db: &'db dyn SourceDatabase,
-    mut def_map: &'db DefMap,
+    mut def_map: &'dm DefMap<'db>,
```
ChayimFriedman2, line 759: I think you can just use the `'db` lifetime here, after all the DefMap is stored in the db.

crates/hir-def/src/dyn_map.rs
```diff
@@ -25,175 +25,220 @@
...
+{
+    #[inline]
+    fn index_mut(&mut self, _key: Key<K, V>) -> &mut Self::Output {
+        // SAFETY: Safe due to `#[repr(transparent)]`.
+        unsafe { std::mem::transmute::<&mut DynMap<'db>, &mut KeyMap<'db, Key<K, V>>>(self) }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ChayimFriedman2: Why this test? Is it for the `DynMap`? Most of it is type gymnastics, I don't think it needs tests.

crates/hir-def/src/lib.rs
```diff
@@ -354,7 +354,7 @@ pub struct BuiltinDeriveImplLoc {
     pub derive_index: u32,
 }
 
-#[salsa::interned(debug, unsafe(no_lifetime), revisions = usize::MAX)]
+#[salsa::interned(debug)]
```
ChayimFriedman2: I wouldn't remove `revisions = usize::MAX` yet. We need to take a conscious decision to enable interneds GC and in what frequency.

crates/hir-ty/src/lower.rs
```diff
@@ -212,7 +212,7 @@ pub struct TyLoweringContext<'db, 'a> {
     types: &'db crate::next_solver::DefaultAny<'db>,
     lang_items: &'db LangItems,
     resolver: &'a Resolver<'db>,
-    store: &'db ExpressionStore,
+    store: &'a ExpressionStore,
```
ChayimFriedman2, line 215: Why this change? It should remain `'db`.

ChayimFriedman2 (commented):
I'm a bit hesitant to merge this while @dfireBird has its lifetimes PR in flight; this PR is much easier to review but the work there is definitely higher...

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23060/ece721d6cdfdf420d8bcc7b9feb48e3b6dbc1f04..d5387cb7f5e85193024f83cb3f18ce9d04fc0cde)*

ada4a:
Happy to wait for @dfireBird's PR to land first – I honestly doubted I would be able to drive this to completion, but since I now have, fixing it up as needed (by reading) sounds doable

crates/hir-def/src/attrs/docs.rs
ada4a: Mostly code clarity I'd say? `LazyCell` makes it very clear what's happening: a value that is initialized lazily.

> `LazyCell` has an additional branch

You mean to check whether the value has been initialized or not? But I think `Option::get_or_insert_with` does a similar check for `None`, no?

Anyway I don't feel too strongly about this, happy to revert.

crates/hir-def/src/dyn_map.rs
ada4a: It was a leftover from Lukas' code snippet – I'll remove it if you say it's unnecessary

crates/hir-def/src/attrs/docs.rs
ChayimFriedman2: `LazyCell` needs to check for reentrance.

I wouldn't oppose if this was written this way to begin with, but I do oppose changing it in an unrelated PR.

crates/hir-def/src/nameres/path_resolution.rs
ada4a: This causes some errors down the line, because `def_map` becomes invariant over `'db`, so I'd leave this as is

crates/hir-def/src/lib.rs
ada4a: Note that it was only recently added in during salsa upgrade (https://github.com/rust-lang/rust-analyzer/pull/23013/commits/4d59eee403e1866d3b5c8d2336f921dacab5717f#diff-8a8808fa1cbb71d05f842b69e4bda92728af59464406b4aa58df635cea4c2571L357-R357), because salsa started requiring it when `no_lifetime` is present. Still, I'll re-add it for now.

crates/hir-ty/src/lower.rs
ada4a: Basically, because of this:
```
error[E0597]: `store` does not live long enough
   --> crates/hir/src/source_analyzer.rs:402:17
    |
109 |   impl<'db> SourceAnalyzer<'db> {
    |        --- lifetime `'db` defined here
...
383 |           let (store, params, _) = lower_generic_params(
    |                ----- binding `store` declared here
...
399 |               match where_predicate_must_hold(
    |  ___________________-
400 | |                 db,
401 | |                 &self.resolver,
402 | |                 &store,
    | |                 ^^^^^^ borrowed value does not live long enough
...   |
406 | |                 predicate,
407 | |             ) {
    | |_____________- argument requires that `store` is borrowed for `'db`
...
421 |       }
    |       - `store` dropped here while still borrowed
```
here, `lower_generic_params` creates `store: ExpressionStore`, and when we try to pass a reference to that to `where_predicate_must_hold`, the lifetime of that reference doesn't live enough (for `'db`), because `store` is only live inside the function body.

I think it's fair to say that in general, if a struct/enum has a field of type `&T` (like `TyLoweringContext.store: &ExpressionStore`), then the lifetime of that reference shouldn't be `'db`, because that makes things invariant over `'db`.

PR #23062, opened 2026-08-06
fix: report const generic args of the wrong type

### LINKED ISSUE #22953: panic: const generics with N: i64 instead of usize
```rust
struct Struct<const N: i64>(pub [u8; N]); // "const N: usize" works

pub fn function(value: Struct<3>) -> u8 {
    value.0[0]
}

fn main() {}
```

`rust-analyzer analysis-stats . --run-all-ide-things`

```
Database loaded:     611.62ms, 229minstr, 48mb (metadata 304.75ms, 1010kinstr, 151kb; build 32.86ms, 181kinstr, 2064b)
  item trees: 1
  dependency lines of code: 1_467_433, item trees: 2_948
  dependency item stats: traits: 848, impl: 9_686, mods: 3_059, macro calls: 5_580, macro rules: 1_054
Item Tree Collection: 4.23s, 23ginstr, 427mb
  Total Statistics:
    crates: 1, mods: 1, decls: 3, bodies: 2, adts: 1, consts: 0, signatures: 3, variants: 1
  Workspace:
    traits: 0, macro_rules macros: 0, proc_macros: 0
    lines of code: 7, item trees: 1
    usages: traits: 0, impl: 0, mods: 0, macro calls: 0, macro rules: 0
  Dependencies:
    lines of code: 1_467_433, item trees: 2_948
    declarations: traits: 848, impl: 9_686, mods: 3_059, macro calls: 5_580, macro rules: 1_054
Item Collection:     2.48s, 11ginstr, 179mb
Crate lang items:    455.83ms, 2474minstr, 31mb
Expression Store Lowering: 393.13µs, 372kinstr, 7kb
  exprs: 6, ??ty: 0 (0%), ?ty: 0 (0%), !ty: 0
  pats: 1, ??ty: 0 (0%), ?ty: 0 (0%), !ty: 0
  panics: 0
Inference:           346.58ms, 1346minstr, 35mb
MIR lowering:        260.57µs, 215kinstr, 113kb
Mir failed bodies: 0 (0%)
Data layouts:        721.00ns, 25kinstr, 0b
Failed data layouts: 0 (100%)
Const evaluation:    28.46µs, 46kinstr, 192b
Failed const evals: 0 (100%)
0/1 0% diagnostics: /tmp/ra/src/main.rs
thread 'main' (1787056) panicked at src/tools/rust-analyzer/crates/hir-ty/src/mir.rs:1329:82:
stack backtrace:
   0:     0x7ff05f8a6246 - <<std[74eb2aff46e51e81]::sys::backtrace::BacktraceLock>::print::DisplayBacktrace as core[7aff35c028b048a4]::fmt::Display>::fmt
   1:     0x7ff06000ad52 - core[7aff35c028b048a4]::fmt::write
   2:     0x7ff05f8bb38c - <std[74eb2aff46e51e81]::sys::stdio::unix::Stderr as core[7aff35c028b048a4]::io::write::Write>::write_fmt
   3:     0x7ff05f877e7a - std[74eb2aff46e51e81]::panicking::default_hook::{closure#0}
   4:     0x7ff05f8988e3 - std[74eb2aff46e51e81]::panicking::default_hook
   5:     0x7ff05f898d4b - std[74eb2aff46e51e81]::panicking::panic_with_hook
   6:     0x7ff05f877f64 - std[74eb2aff46e51e81]::panicking::panic_handler::{closure#0}
   7:     0x7ff05f86f3e9 - std[74eb2aff46e51e81]::sys::backtrace::__rust_end_short_backtrace::<std[74eb2aff46e51e81]::panicking::panic_handler::{closure#0}, !>
   8:     0x7ff05f87988d - __rustc[49844a0cf2bb6b0d]::rust_begin_unwind
   9:     0x7ff05c633c7c - core[7aff35c028b048a4]::panicking::panic_fmt
  10:     0x7ff05c9f2952 - core[7aff35c028b048a4]::panicking::panic
  11:     0x7ff05d1bda79 - core[7aff35c028b048a4]::option::unwrap_failed
  12:     0x55e11fc7436a - <hir_ty[f58a0abe133f1c00]::mir::PlaceTy>::projection_ty::<la_arena[f716c310310ca8c]::Idx<hir_ty[f58a0abe133f1c00]::mir::Local>>
  13:     0x55e11fda7357 - hir_ty[f58a0abe133f1c00]::mir::borrowck::moved_out_of_ref::{closure#0}
  14:     0x55e11fda9d59 - <<hir_ty[f58a0abe133f1c00]::InferBodyId>::borrowck::_::borrowck_query_Configuration_ as salsa[8774cceebe7971b6]::function::Configuration>::execute::inner_::{closure#0}
  15:     0x55e11ffd5dfb - <<hir_ty[f58a0abe133f1c00]::InferBodyId>::borrowck::_::borrowck_query_Configuration_ as salsa[8774cceebe7971b6]::function::Configuration>::execute
  16:     0x55e11fdf10ba - <salsa[8774cceebe7971b6]::function::IngredientImpl<<hir_ty[f58a0abe133f1c00]::InferBodyId>::borrowck::_::borrowck_query_Configuration_>>::execute
  17:     0x55e11ff05d85 - <hir_ty[f58a0abe133f1c00]::InferBodyId>::borrowck
  18:     0x55e11f878765 - <hir[4ef1887375cd4b44]::DefWithBody>::diagnostics
  19:     0x55e11f82136e - <hir[4ef1887375cd4b44]::ModuleDef>::diagnostics
  20:     0x55e11f835363 - <hir[4ef1887375cd4b44]::Module>::diagnostics
  21:     0x55e1207152ec - ide_diagnostics[22d31180f2d1ac67]::semantic_diagnostics
  22:     0x55e12070f972 - ide_diagnostics[22d31180f2d1ac67]::full_diagnostics
  23:     0x55e120152ba9 - <ide[604e506ad1ac25c5]::Analysis>::full_diagnostics
  24:     0x55e120c00b9b - <rust_analyzer[f1820004a41e425c]::cli::flags::AnalysisStats>::run
  25:     0x55e120e3ffd6 - rust_analyzer[c4fd19d1c3329aa5]::main
  26:     0x55e120e3c373 - std[74eb2aff46e51e81]::sys::backtrace::__rust_begin_short_backtrace::<fn() -> core[7aff35c028b048a4]::result::Result<std[74eb2aff46e51e81]::process::ExitCode, anyhow[f3ff20ce718c24c8]::Error>, core[7aff35c028b048a4]::result::Result<std[74eb2aff46e51e81]::process::ExitCode, anyhow[f3ff20ce718c24c8]::Error>>
  27:     0x55e120e3e0d2 - std[74eb2aff46e51e81]::rt::lang_start::<core[7aff35c028b048a4]::result::Result<std[74eb2aff46e51e81]::process::ExitCode, anyhow[f3ff20ce718c24c8]::Error>>::{closure#0}
  28:     0x7ff061103553 - std[74eb2aff46e51e81]::rt::lang_start_internal
  29:     0x55e120e56b68 - main
  30:     0x7ff05a627741 - <unknown>
  31:     0x7ff05a627879 - __libc_start_main
  32:     0x55e11f73aed9 - <unknown>
  33:                0x0 - <unknown>
    Checking ra v0.1.0 (/tmp/ra)
```

`rust-analyzer 1.99.0-nightly (26ae60a 2026-07-28)`

hongjr03:
This pr fixes rust-lang/rust-analyzer#22953 (supersedes rust-lang/rust-analyzer#23045). The panic from the issue was already fixed upstream by removing borrowck in rust-lang/rust-analyzer#23043 .

The root cause is that although the `ConstArgHasType` constraint is registered when indexing an array, the resulting `SelectionError::ConstArgHasWrongType` was dropped implicitly in `solver_errors::from_fulfillment_error`, so no diagnostic was produced.

### DIFF
--- a/crates/hir-ty/src/tests/regression.rs
+++ b/crates/hir-ty/src/tests/regression.rs
@@ -3122,3 +3122,36 @@ fn main() {
     "#,
     );
 }
+
+#[test]
+fn regression_22953_mir_body() {
+    let _tracing = crate::setup_tracing();
+    use hir_def::DefWithBodyId;
+    use hir_def::ModuleDefId;
+    use test_fixture::WithFixture;
+
+    let (db, file_id) = crate::test_db::TestDB::with_single_file(
+        r#"
+struct Struct<const N: i64>(pub [u8; N]);
+
+pub fn function(value: Struct<3>) -> u8 {
+    value.0[0]
+}
+
+fn main() {}
+    "#,
+    );
+    crate::attach_db(&db, || {
+        let module = db.module_for_file_opt(file_id.file_id(&db)).unwrap();
+        let def_map = module.def_map(&db);
+        let mut defs = Vec::new();
+        super::visit_module(&db, def_map, module, &mut |it| {
+            if let ModuleDefId::FunctionId(it) = it {
+                defs.push(DefWithBodyId::FunctionId(it));
+            }
+        });
+        for def in defs {
+            let _ = crate::db::HirDatabase::mir_body(&db, def.into());
+        }
+    });
+}
--- a/crates/hir-ty/src/tests/regression.rs
+++ b/crates/hir-ty/src/tests/regression.rs
@@ -3122,36 +3122,3 @@ fn main() {
     "#,
     );
 }
-
-#[test]
-fn regression_22953_mir_body() {
-    let _tracing = crate::setup_tracing();
-    use hir_def::DefWithBodyId;
-    use hir_def::ModuleDefId;
-    use test_fixture::WithFixture;
-
-    let (db, file_id) = crate::test_db::TestDB::with_single_file(
-        r#"
-struct Struct<const N: i64>(pub [u8; N]);
-
-pub fn function(value: Struct<3>) -> u8 {
-    value.0[0]
-}
-
-fn main() {}
-    "#,
-    );
-    crate::attach_db(&db, || {
-        let module = db.module_for_file_opt(file_id.file_id(&db)).unwrap();
-        let def_map = module.def_map(&db);
-        let mut defs = Vec::new();
-        super::visit_module(&db, def_map, module, &mut |it| {
-            if let ModuleDefId::FunctionId(it) = it {
-                defs.push(DefWithBodyId::FunctionId(it));
-            }
-        });
-        for def in defs {
-            let _ = crate::db::HirDatabase::mir_body(&db, def.into());
-        }
-    });
-}
--- a/crates/hir/src/diagnostics.rs
+++ b/crates/hir/src/diagnostics.rs
@@ -108,6 +108,7 @@ diagnostics![AnyDiagnostic<'db> ->
     CannotImplicitlyDerefTraitObject<'db>,
     CannotIndexInto<'db>,
     CastToUnsized<'db>,
+    ConstArgHasWrongType<'db>,
     ExpectedArrayOrSlicePat<'db>,
     ExpectedFunction<'db>,
     ExplicitDropMethodUse,
@@ -658,6 +659,14 @@ pub struct UnimplementedTrait<'db> {
 }
 
 #[derive(Debug)]
+pub struct ConstArgHasWrongType<'db> {
+    pub span: SpanSyntax,
+    pub ct: hir_ty::next_solver::Const<'db>,
+    pub ct_ty: hir_ty::next_solver::Ty<'db>,
+    pub expected_ty: hir_ty::next_solver::Ty<'db>,
+}
+
+#[derive(Debug)]
 pub struct MutableRefBinding {
     pub pat: InFile<ExprOrPatPtr>,
 }
@@ -1200,6 +1209,15 @@ impl<'db> AnyDiagnostic<'db> {
                     })
                     .collect();
                 UnimplementedTrait { span, trait_predicate, parent_trait_predicates }.into()
+            }
+            SolverDiagnosticKind::ConstArgHasWrongType { ct, ct_ty, expected_ty } => {
+                ConstArgHasWrongType {
+                    span,
+                    ct: ct.as_ref(),
+                    ct_ty: ct_ty.as_ref(),
+                    expected_ty: expected_ty.as_ref(),
+                }
+                .into()
             }
         })
     }
--- a/crates/hir/src/diagnostics.rs
+++ b/crates/hir/src/diagnostics.rs
@@ -661,9 +661,9 @@ pub struct UnimplementedTrait<'db> {
 #[derive(Debug)]
 pub struct ConstArgHasWrongType<'db> {
     pub span: SpanSyntax,
-    pub ct: hir_ty::next_solver::Const<'db>,
-    pub ct_ty: hir_ty::next_solver::Ty<'db>,
-    pub expected_ty: hir_ty::next_solver::Ty<'db>,
+    pub ct: String,
+    pub ct_ty: Type<'db>,
+    pub expected_ty: Type<'db>,
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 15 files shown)


crates/hir-ty/src/tests/regression.rs
```diff
@@ -3122,3 +3122,36 @@ fn main() {
     "#,
     );
 }
+
+#[test]
+fn regression_22953_mir_body() {
```
ChayimFriedman2: We have a place (with helpers) for tests for MIR lowering in crates/hir-ty/src/mir/lower/tests.rs.

crates/hir/src/diagnostics.rs
```diff
@@ -657,6 +658,14 @@ pub struct UnimplementedTrait<'db> {
     pub parent_trait_predicates: Vec<crate::TraitPredicate<'db>>,
 }
 
+#[derive(Debug)]
+pub struct ConstArgHasWrongType<'db> {
+    pub span: SpanSyntax,
+    pub ct: hir_ty::next_solver::Const<'db>,
```
ChayimFriedman2: You cannot put `hir_ty` types hir, only `hir` types.

ChayimFriedman2 (changes requested):
More diagnostics is always a good thing and I would not oppose it, but... are you sure this really fixes the root cause? We have no logic in MIR lowering to skip lowering if there are diagnostics (we should have though, but it's tricky to decide what errors should prevent lowering).

Also about this diagnostic specifically, rustc does emit a diagnostic here but I'm not personally sure we should: of course we should emit a diagnostic, but when the the mismatch actually occurs (i.e. when you're using a `i64` as an array length or similar) and not when the constant is *used* (rustc emits at both).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23062/ece721d6cdfdf420d8bcc7b9feb48e3b6dbc1f04..ac4fe7fe9e4f01610f7c51627dc6f876d0893588)*

crates/hir-ty/src/tests/regression.rs
hongjr03: I have moved the test to `crates/hir-ty/src/mir/lower/tests.rs`. Note that the previous `lower_mir` discarded the return value of `db.mir_body()`, so it only verified that no panic occurred during lowering. Therefore I added `MirLoweringExpectation` to assert the expected outcome.

hongjr03:
> More diagnostics is always a good thing and I would not oppose it, but... are you sure this really fixes the root cause? We have no logic in MIR lowering to skip lowering if there are diagnostics (we should have though, but it's tricky to decide what errors should prevent lowering).

You're right. The change now just plugs `ConstArgHasWrongType` into the same type-mismatch path that `lower_to_mir_with_store` already uses to bail out — `emit_trait_errors` now reports it and inference marks it in `nodes_with_type_mismatches`. So no new policy needed, it just treats const-arg mismatches as another type mismatch.

> Also about this diagnostic specifically, rustc does emit a diagnostic here but I'm not personally sure we should: of course we should emit a diagnostic, but when the mismatch actually occurs (i.e. when you're using an `i64` as an array length or similar) and not when the constant is _used_ (rustc emits at both).

I dropped the `ConstArgHasWrongType` diagnostic path and instead emit from type lowering, at the point where the constant is lowered against its expected type. So we now get the actual "mismatch occurrence" site.

Thanks for the review!

ChayimFriedman2 (changes requested):
That was not my meaning. Type mismatches are handles elsewhere, the reason they're not currently emitted for this code is unrelated. You should definitely not make this error a type mismatch. As I said rustc reports it twice, but if we don't then we should just not report `ConstArgHasType`.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23062/ece721d6cdfdf420d8bcc7b9feb48e3b6dbc1f04..9b2a6202f1b5a3e4527e469a12a7579652e42b53)*

hongjr03:
Sorry, I hadn't thought this through and got it pretty badly wrong. I've already removed the `ConstArgHasType` part.

PR #23070, opened 2026-08-07
fix: remove extra spaces in full function signatures

### LINKED ISSUE #23069: Completion descriptions contain unexpected spaces inside parentheses when `fullFunctionSignatures` is enabled
**rust-analyzer version**: 0.3.2997-standalone

**rustc version**: rustc 1.96.0 (ac68faa20 2026-05-25)

**editor or extension**: Zed

**relevant settings**: [rust-analyzer.completion.fullFunctionSignatures.enable](https://rust-analyzer.github.io/book/configuration.html#completion.fullFunctionSignatures.enable)

```rust
static PANICKED: core::sync::atomic::AtomicBool = core::sync::atomic::AtomicBool::new(false);

const _:() = PANICKED.ˇ    // trigger completion here

```
Some completion items contain unexpected whitespace in their descriptions. The second and third items in the screenshot below are affected:
Image
There is a space between `(` and `&self` (and also a space before the closing `)`); there shouldn't be one.

The whitespace turns out to be present in the raw LSP messages produced by rust-analyzer itself. Part of the RPC payload for these two items:
```jsonc
[
{
        "label": "compare_exchange_weak(…)",
        "labelDetails": {
          "detail": "(alias compare_and_swap)",
          "description": "pub fn compare_exchange_weak( &self, current: bool, new: bool, success: Ordering, failure: Ordering ) -> Result<bool, bool>"
        },
},
{
        "label": "compare_exchange(…)",
        "labelDetails": {
          "detail": "(alias compare_and_swap)",
          "description": "pub fn compare_exchange( &self, current: bool, new: bool, success: Ordering, failure: Ordering ) -> Result<bool, bool>"
        },
}
]
```
The `labelDetails.description` in the raw payload contains a space right after `(` and right before `)`, so the extra whitespace is introduced by rust-analyzer, not by the editor's rendering.

lingyaochu:
Closes rust-lang/rust-analyzer#23069 

When `rust-analyzer.completion.fullFunctionSignatures.enable` is set to `true`, completion descriptions are generated by `detail_full`, which internally calls `write_function()` to produce the signature. For functions with more than four parameters, the generated signature is multi-line:
https://github.com/rust-lang/rust-analyzer/blob/ece721d6cdfdf420d8bcc7b9feb48e3b6dbc1f04/crates/hir/src/display.rs#L195-L210

Take the `compare_exchange()` function as an example, the final generated signature for this function is like:
```
pub fn compare_exchange(
    &self,
    current: bool,
    new: bool,
    success: Ordering,
    failure: Ordering,
) -> Result<bool, bool>
```
Flattening this output in `detail_full` previously inserted a space right after `(` and right before `)`, e.g. `pub fn compare_exchange( &self, ...  )`.

This PR makes a small change to `detail_full` to skip the whitespace insertion around the parentheses.

### DIFF
--- a/crates/ide-completion/src/render/function.rs
+++ b/crates/ide-completion/src/render/function.rs
@@ -336,7 +336,7 @@ fn detail_full(ctx: &CompletionContext<'_, '_>, func: hir::Function) -> String {
     let mut detail = String::with_capacity(signature.len());
 
     for segment in signature.split_whitespace() {
-        if !detail.is_empty() {
+        if !detail.is_empty() && !detail.ends_with('(') && !segment.starts_with(')') {
             detail.push(' ');
         }
 
--- a/crates/ide-completion/src/tests/special.rs
+++ b/crates/ide-completion/src/tests/special.rs
@@ -1617,6 +1617,19 @@ fn main() {
         expect!("const fn(&'foo mut self, &Foo) -> !"),
         expect!("pub const fn baz<'foo>(&'foo mut self, x: &'foo Foo) -> !"),
     );
+
+    check_signatures(
+        r#"
+struct Foo;
+impl Foo {
+    pub fn method(&self, first: bool, second: bool, third: bool, fourth: bool) {}
+}
+fn main() { Foo.m$0 }
+"#,
+        CompletionItemKind::SymbolKind(SymbolKind::Method),
+        expect!("fn(&self, bool, bool, bool, bool)"),
+        expect!("pub fn method(&self, first: bool, second: bool, third: bool, fourth: bool)"),
+    );
 }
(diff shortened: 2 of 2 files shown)


A4-Tacks (approved):
LGTM

Btw, this bug comes from #21591
Perhaps in the future, we can add newline related style flags to `HirFormatter`


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23070/ece721d6cdfdf420d8bcc7b9feb48e3b6dbc1f04..33e5b55fd1146aee50eb7bca38302cbe16f32ed0)*

PR #23072, opened 2026-08-07
fix: Unresolved type variables shouldn't escape impl selection

Wilfred:
find_matching_impl() called resolve_vars_if_possible(), but we rely on unresolved region inference variables not escaping. Ensure that we substitute these variables.

Without this fix, we can panic when trying to resolve the variable in the wrong inference context. See the new unit test.

This seems to have always been wrong, but the panic only started occurring due when libcore changed in
https://github.com/rust-lang/rust/pull/136006.

AI disclosure: Code partly written by GPT-5 Codex. I've bisected and confirmed the repro, and reviewed all the code, but possibly there's a nicer place to fix this logic.

### DIFF
--- a/crates/hir-ty/src/method_resolution.rs
+++ b/crates/hir-ty/src/method_resolution.rs
@@ -31,7 +31,7 @@ use hir_def::{
 };
 use rustc_hash::{FxHashMap, FxHashSet};
 use rustc_type_ir::{
-    TypeVisitableExt, VisitorResult,
+    TypeFoldable, TypeVisitableExt, VisitorResult,
     fast_reject::{TreatParams, simplify_type},
     inherent::{BoundExistentialPredicates, IntoKind},
     try_visit,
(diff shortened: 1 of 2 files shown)


Wilfred:
I encountered this panic when trying to generate SCIP for the rhai crate.

ChayimFriedman2 (approved):
I'm not entirely sure if we should replace infer regions with error or erased, but this is good enough for now. Except - can you leave a FIXME for that please?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23072/2cfd097a835dca6e760d95ce937927dc94b5a5d1..c12eb5663d581534b4010f6184a7f5e4e6cb662d)*

Wilfred:
Added a FIXME which I think describes the your comment accurately.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23072/c54a08a7ac0deee74990652403b19ca6a3602c6a..d799ff3c0cb3c2e028ee53cde8c6ce8cc9646231)*

PR #23097, opened 2026-08-09
feat: add strict arithmetic replacement assist

### LINKED ISSUE #23080: The replace arithmetic assist should include the `strict_*` variants
The replace arithmetic assist added in #13458 should also have the option of using the `strict_*` methods stabilized in Rust 1.91.

subotac:
Closes rust-lang/rust-analyzer#23080.

  This adds the missing `strict_*` variant to the existing arithmetic operator replacement assists. For example, `1 + 2` can now be replaced with `1.strict_add(2)`.

  The new `replace_arith_with_strict` assist reuses the existing implementation by adding `ArithKind::Strict`, which maps supported operators to the corresponding `strict_*` method names. The test covers
  replacing addition with `strict_add`.

  Validation:

  - Focused tests: 2 passed
  - Full `ide-assists` test suite: 2806 passed, 3 ignored
  - `cargo fmt`
  - Generated assist tests
  - `xtask tidy`
  - `git diff --check`

  Clippy could not complete under Rust 1.95 because of a new baseline lint in the unchanged `stdx` crate; the failure is unrelated to this change

### DIFF
--- a/crates/ide-assists/src/handlers/replace_arith_op.rs
+++ b/crates/ide-assists/src/handlers/replace_arith_op.rs
@@ -31,6 +31,28 @@ pub(crate) fn replace_arith_with_checked(
     replace_arith(acc, ctx, ArithKind::Checked)
 }
 
+// Assist: replace_arith_with_strict
+//
+// Replaces arithmetic on integers with the `strict_*` equivalent.
+//
+// ```
+// fn main() {
+//   let x = 1 $0+ 2;
+// }
+// ```
+// ->
+// ```
+// fn main() {
+//   let x = 1.strict_add(2);
+// }
+// ```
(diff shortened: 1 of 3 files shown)


lnicola:
Thank you for your contribution. It looks like CI ran on the wrong commit, so it should be fine. But can you say if and to what measure you've used AI for it?

ChayimFriedman2:
This was already added in https://github.com/rust-lang/rust-analyzer/pull/23082.

Also, it appears you've used AI without disclosing it, contrary to [our AI policy](https://github.com/rust-lang/rust-analyzer/blob/master/AI_POLICY.md). Please do not do that, next time we'll reach to moderation.

PR #23100, opened 2026-08-09
Base fill-match-arms suggestions on the exhaustiveness checker

### LINKED ISSUE #8129: "Fill match arms" code action creates non-exhaustive match
I have code like this:

```rust
        let mut chars = s.chars().peekable();
        match chars.peek() {
            Some(&'+') => {
                chars.next();
            }
            Some(&'-') => {
                bigint.negative = true;
                chars.next();
            }
            _ => (),
        }
```

Rust analyzer (run through Emacs) suggests the "fill match arms" action which, when applied, changes the fallback match arm to:

```rust
        None => {}
```

However, that's incorrect because as the compiler points out, the patterns `Some(&'\u{0}'..='*')`, `Some(&',')`, `Some(&'.'..='\u{d7ff}')` and 1 more not covered.

rit3sh-x:
Fixes rust-lang/rust-analyzer#8129.

## What changed
- `add_missing_match_arms` now asks the exhaustiveness checker for the patterns actually missing from a `match` (via a new `missing_match_arm_patterns` query added to `hir-ty` and exposed through `Semantics`), instead of only enumerating the scrutinee's top-level enum variants. This covers arms nested inside tuples, structs, references, and `Option`, and no longer produces the invalid catch-all rewrite from rust-lang/rust-analyzer#8129 when a wildcard arm hides a real gap.
- Falls back to the previous variant-enumeration behavior when the checker can't describe the gap concisely (bare wildcards, or a witness whose only content is a full tuple enumeration).
- Added `crates/hir-ty/src/diagnostics/expr.rs::UncoveredPattern<Var>`, a generic witness-shape type shared between `hir-ty` (parameterized over `VariantId`) and `hir` (parameterized over `Variant` via `map_variant`), rather than two independent copies of the same enum.

## Testing
- New coverage in `crates/ide-assists/src/handlers/add_missing_match_arms.rs` exercising nested variants, record fields, or-patterns, guards, hidden variants, self-referential enums, and `Self`-qualified paths.

## AI Usage
- Used in generated test cases and comments.

### DIFF
--- a/crates/hir-ty/src/diagnostics.rs
+++ b/crates/hir-ty/src/diagnostics.rs
@@ -7,7 +7,8 @@ mod unsafe_check;
 pub use crate::diagnostics::{
     decl_check::{CaseType, IncorrectCase, incorrect_case},
     expr::{
-        BodyValidationDiagnostic, record_literal_missing_fields, record_pattern_missing_fields,
+        BodyValidationDiagnostic, UncoveredPattern, missing_match_arm_patterns,
+        record_literal_missing_fields, record_pattern_missing_fields,
     },
     unsafe_check::{
         InsideUnsafeBlock, UnsafetyReason, missing_unsafe, unsafe_operations,
--- a/crates/hir-ty/src/diagnostics/expr.rs
+++ b/crates/hir-ty/src/diagnostics/expr.rs
@@ -45,6 +45,106 @@ pub(crate) use hir_def::{
     hir::{Expr, ExprId, MatchArm, Pat, PatId, RecordSpread, Statement},
 };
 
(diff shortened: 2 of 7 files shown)


ChayimFriedman2:
It's sad that you worked on this since I already have a working branch (and I've said that multiple times).

Given that, if this is already in good shape, I'll merge this, but if not, I'll not engage in in review rounds and just close this.

rit3sh-x:
@ChayimFriedman2 bad timing is my forte 😅, feel free to address my issues.

rit3sh-x:
@ChayimFriedman2 any updates?

ChayimFriedman2:
I will review that when I have time.

PR #23103, opened 2026-08-10
fix: Do not consider locals of `async fn` as upvars of the returned coroutine

### LINKED ISSUE #23102: E0277 false positive in batch diagnostics: `MutexGuard: Send` reported for spawned async fn with no guard held across await
`rust-analyzer diagnostics .` reports an E0277 hard error on code that rustc accepts. No `MutexGuard` is ever held across an await point.

### Minimal reproduction

`Cargo.toml`:

```toml
[package]
name = "repro-mutexguard-send"
version = "0.1.0"
edition = "2021"

[dependencies]
tokio = { version = "1", features = ["rt", "sync", "time", "macros"] }
```

`src/lib.rs`:

```rust
use std::path::PathBuf;
use std::sync::{Arc, Mutex};

pub struct WorkspaceState {
    pub value: u64,
}

pub async fn actor_body(
    path: PathBuf,
    workspace: Arc<Mutex<WorkspaceState>>,
    notify: Arc<tokio::sync::Notify>,
    last_error: Arc<Mutex<String>>,
) {
    let mut previous: Option<Vec<u8>> = None;
    loop {
        notify.notified().await;
        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
        let _ = &path;
        {
            let guard = workspace.lock().unwrap();
            let bytes = guard.value.to_le_bytes().to_vec();
            if previous.as_deref() != Some(bytes.as_slice()) {
                previous = Some(bytes);
            }
        }
        {
            let mut err = last_error.lock().unwrap();
            err.clear();
        }
        let _ = &previous;
    }
}

pub fn spawn_supervisor(
    path: PathBuf,
    workspace: Arc<Mutex<WorkspaceState>>,
    notify: Arc<tokio::sync::Notify>,
    last_error: Arc<Mutex<String>>,
) {
    tokio::spawn(async move {
        loop {
            let inner_path = path.clone();
            let inner_workspace = Arc::clone(&workspace);
            let inner_notify = Arc::clone(&notify);
            let inner_last = Arc::clone(&last_error);
            let handle = tokio::spawn(actor_body(
                inner_path,
                inner_workspace,
                inner_notify,
                inner_last,
            ));
            let _ = handle.await;
        }
    });
}
```

### Actual

```
Error RustcHardError("E0277") from LineCol { line: 52, col: 25 } to LineCol { line: 52, col: 37 }:
the trait bound `MutexGuard<'<erased>, WorkspaceState>: Send` is not satisfied
```

The span is the `Arc::clone(&last_error)` expression inside the supervisor loop (which contains no `lock()` call at all). Every lock scope in `actor_body` ends before the next await point, so the spawned futures are `Send`; `cargo check` and `cargo clippy --all-targets -- -D warnings` both accept the crate.

### Expected

No diagnostic.

### Versions

Reproduced with both:

- `rust-analyzer 1.97.1 (8bab26f4 2026-07-14)` (rustup component, stable toolchain)
- `rust-analyzer 0.3.2997-standalone (b54a82b321 2026-08-02)` (2026-08-03 release)

macOS aarch64. rustc 1.97.1.

### Possibly related symptom

In the larger workspace this was reduced from, the same two analyzer versions also report `the trait bound `Arguments<'<erased>>: Sync` is not satisfied` spanning two `axum` `WebSocketUpgrade::on_upgrade(move |socket| ...)` call chains whose handlers rustc likewise accepts. I could not reduce that one to a standalone crate (it seems to need the cross-crate async call graph), but the `'<erased>` lifetime in an auto-trait bound and the async-capture context match this report, so they may share a root cause in how auto traits are solved for async generator captures.

ChayimFriedman2:
Fixes rust-lang/rust-analyzer#23102.

### DIFF
--- a/crates/hir-def/src/expr_store/lower.rs
+++ b/crates/hir-def/src/expr_store/lower.rs
@@ -1211,13 +1211,36 @@ impl<'db> ExprCollector<'db> {
                     (false, true) => CoroutineKind::Gen,
                     (false, false) => unreachable!(),
                 };
-                this.lower_coroutine_body_with_moved_arguments(
+                let coroutine = this.lower_coroutine_body_with_moved_arguments(
                     self_param,
                     params,
                     body,
                     kind,
                     CoroutineSource::Fn,
-                )
+                );
+                // *All* locals belong to the inner coroutine...
+                this.associate_unowned_bindings_with(0, coroutine);
+                // ...except the formal parameters, which are *not* necessarily what was passed as parameters into `collect()`,
+                // since `lower_coroutine_body_with_moved_arguments()` might have changed them.
+                params
+                    .iter()
+                    .filter_map(|param| {
+                        if let Pat::Bind { id, .. } = this.store.pats[param.formal] {
+                            Some(id)
+                        } else {
+                            never!(
+                                "`lower_coroutine_body_with_moved_arguments()` should make sure \
+                                the coroutine closure only have simple bind args"
+                            );
+                            None
+                        }
+                    })
+                    .chain(self_param.map(|param| param.formal))
+                    .for_each(|param| {
+                        // They are owned by the top-level function, so should not be present in `bindings_owner`.
+                        this.store.binding_owners.remove(&param);
+                    });
+                coroutine
             } else {
                 body
             }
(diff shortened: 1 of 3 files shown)


crates/hir-def/src/expr_store/lower.rs
```diff
@@ -1210,13 +1211,36 @@ impl<'db> ExprCollector<'db> {
...
+                            never!(
+                                "`lower_coroutine_body_with_moved_arguments()` should make sure \
+                                the coroutine closure only have simple bind args"
+                            );
+                            None
+                        }
+                    })
+                    .chain(self_param.map(|param| param.formal))
+                    .for_each(|param| {
+                        // They are owned by the top-level function, so should not bee present in `bindings_owner`.
```
```suggestion
                        // They are owned by the top-level function, so should not be present in `bindings_owner`.
```
🐝 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23103/57411b3a0fb0181ed5965f6541d5ed39a9d79d4c..81114374d3dfc4a4c48794f9be1100ef4f205503)*

Veykril (approved):

PR #23107, opened 2026-08-10
fix: Off-by-one in lifetime binders when lowering `dyn Trait<'a>`

Wilfred:
`dyn Trait<'a>` does not have an explicit `for<'a>`, but it still has a binder. If we don't call with_shifted_in(), we end up with lifetimes whose de Bruijn index points to the wrong level.

This could cause r-a to refer to the wrong lifetime variable, or even panic when the enclosing binder doesn't declare a lifetime variable. See the new unit test.

This bug meant r-a would previously crash when generating a SCIP file for the dachshund crate.

AI disclosure: Code partly written by GPT-5, all comments, commit messages and mistakes are mine.

### DIFF
--- a/crates/hir-ty/src/lower.rs
+++ b/crates/hir-ty/src/lower.rs
@@ -932,7 +932,7 @@ impl<'db, 'a> TyLoweringContext<'db, 'a> {
         bound: &'b TypeBound,
         self_ty: Ty<'db>,
         ignore_bindings: bool,
-    ) -> impl Iterator<Item = (Clause<'db>, GenericPredicateSource)> + use<'b, 'a, 'db> {
+    ) -> impl Iterator<Item = (Clause<'db>, GenericPredicateSource)> + use<'db> {
         let interner = self.interner;
         let meta_sized = self.lang_items.MetaSized;
         let pointee_sized = self.lang_items.PointeeSized;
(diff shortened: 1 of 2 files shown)


Wilfred:
Relevant code snippet that triggered the panic:

https://github.com/facebookresearch/dachshund/blob/b3d4e16c20f2a13371a25a68b8368f22b4fc685f/src/dachshund/algorithms/connectivity.rs#L48-L54

ChayimFriedman2:
I'm hesitant to merge this since I fear this will collide with the lifetime elision work. CC @dfireBird.

crates/hir-ty/src/lower.rs
```diff
@@ -1030,7 +1030,17 @@ impl<'db, 'a> TyLoweringContext<'db, 'a> {
 
             for b in bounds {
                 let db = self.db;
-                self.lower_type_bound(b, dummy_self_ty, false).for_each(|(b, _)| {
+                let clauses: Vec<_> = match b {
+                    TypeBound::Path(_, TraitBoundModifier::None) => {
```
Wilfred, line 1034: If I'm understanding #22927 correctly, I think this can change to `TypeBound::Path(None, _, TraitBoundModifier::None) =>` if that PR merges first. I think the PRs are strictly complementary.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23107/513f60bfe6641ed48072862cf4c1821696e2630d..a59cd0b1a8dafbff99afa3e7d524878d2ce76b4b)*

dfireBird:
I think this would be fine to merge. Thanks for catching this.

crates/hir-ty/src/lower.rs
```diff
@@ -1030,7 +1030,17 @@ impl<'db, 'a> TyLoweringContext<'db, 'a> {
...
+                        // `dyn Trait<'a>` is an existential predicate that introduces a binder.
+                        self.with_shifted_in(&[], |ctx| {
+                            ctx.lower_type_bound(b, dummy_self_ty, false).collect()
```
ChayimFriedman2: I don't like the extra `collect()`. You can avoid it by extracting the `for_each` into a named closure and doing it in both arms,

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23107/513f60bfe6641ed48072862cf4c1821696e2630d..a59cd0b1a8dafbff99afa3e7d524878d2ce76b4b)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/lower.rs
Wilfred: Refactored to a match that returns an iterator.

ChayimFriedman2 (approved):

PR #23108, opened 2026-08-10
internal: Support multiple stable releases in one day

### LINKED ISSUE #20451: Source of tag 2025-08-11 changed content
This caused downstream build failures in hash check: https://github.com/NixOS/nixpkgs/issues/432960
Also discussions: https://github.com/NixOS/nixpkgs/pull/432742#issuecomment-3182252016

I seems that it was an intentional re-release, but I wanna confirm that.

Even if it's a hot fix, I (as the nixpkgs maintainer of rust-analyzer package) would suggest to release under a new tag (2025-08-12.2 or 2025-08-13) instead changing the existing one, which seems too suspicious and be like a supply chain attack.

lnicola:
AI disclosure: this was largely written by Gemini Flash 3.6 with some prodding to use `awk` instead of a loop. I have reviewed and mostly understood it.

Fixes rust-lang/rust-analyzer#20451

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -214,16 +214,28 @@ jobs:
         with:
           node-version: 22
 
-      - run: echo "TAG=$(date --iso -u)" >> $GITHUB_ENV
-        if: github.ref == 'refs/heads/release'
-      - run: echo "TAG=nightly" >> $GITHUB_ENV
-        if: github.ref != 'refs/heads/release'
-      - run: 'echo "TAG: $TAG"'
-
       - name: Checkout repository
         uses: actions/checkout@v6
         with:
           fetch-depth: ${{ env.FETCH_DEPTH }}
+
+      - name: Set release tag
+        run: |
+          if [ "${{ github.ref }}" = "refs/heads/release" ]; then
+            git fetch --tags
+            today=$(date --iso -u)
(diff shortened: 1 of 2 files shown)


lnicola:
FWIW, I'd like to merge this today and cut a new release because the original one is bugged, and it's a good opportunity to test it.

Draft run [here](https://github.com/lnicola/rust-analyzer/releases), without any release assets.

ChayimFriedman2:
> I have reviewed and mostly understood it.

Normally this wouldn't be enough, but nobody fully understands GitHub's YAML and bash, so...

I don't feel qualified to review this myself though.

lnicola:
I _think_ I understand the AWK script, but I've almost never written AWK so..

nicolas-guichard:
If you're not married to awk, I think that's easier to understand:
```
last_tag="$(git tag -l "${today}*" | sort | tail -n1)"
if [ "$last_tag" ]; then
    last_v=$(echo "$last_tag" | grep -oP '(?<=\.)\d+' || echo 0)
    new_v=$(($last_v + 1))
    tag="${today}.${new_v}"
else
    tag="$today"
fi
```

Veykril (approved):

lnicola:
> I think that's easier to understand

I'm not sure, but I switched to your version, changed to `if [ -n ... ]` and `sort -V`. I hope we won't have 10 releases in one day though.

PR #23109, opened 2026-08-10
Offer `replace_arith` on references to ints

ronnodas:
Currently the `Replace arithmetic" assist is only offered when both operands are integers types, but it should be also applicable when they are references to such.

### DIFF
--- a/crates/ide-assists/src/handlers/replace_arith_op.rs
+++ b/crates/ide-assists/src/handlers/replace_arith_op.rs
@@ -118,16 +118,17 @@ fn replace_arith(acc: &mut Assists, ctx: &AssistContext<'_, '_>, kind: ArithKind
             let receiver = wrap_paren(lhs.clone(), make, ast::prec::ExprPrecedence::Postfix);
 
             let mut rhs = rhs;
-            if let ast::Expr::RefExpr(ref_expr) = &rhs
-                && let Some(inner) = ref_expr.expr()
-            {
-                rhs = inner;
-            }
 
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/replace_arith_op.rs
```diff
@@ -116,6 +116,21 @@ fn replace_arith(acc: &mut Assists, ctx: &AssistContext<'_, '_>, kind: ArithKind
             let method_name = kind.method_name(op);
 
             let receiver = wrap_paren(lhs.clone(), make, ast::prec::ExprPrecedence::Postfix);
+
+            let mut rhs = rhs;
+            if let ast::Expr::RefExpr(ref_expr) = &rhs
+                && let Some(inner) = ref_expr.expr()
```
A4-Tacks: I think this should only be done when it comes to `adjusted.strip_reference() != adjusted`, like
```rust
if ast::Expr::RefExpr(ref_expr) = &rhs && let Some(inner) = ref_expr.expr() {
    rhs = inner;
} else {
    rhs = make.expr_prefix(T![*], rhs).into();
}
```

A4-Tacks (commented):
This intersects with #22225, I think this should be merged after #22225 is merged


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23109/513f60bfe6641ed48072862cf4c1821696e2630d..d912ebee87f7c3a9ec29e2d3c962f4db718835c5)*

A4-Tacks (approved):

PR #23151, opened 2026-08-15
fix: emit E0422 for unresolved record expr paths

Yarous:
Part of rust-lang/rust-analyzer#22140 — addresses the [`// FIXME: Emit an error.`](https://github.com/rust-lang/rust-analyzer/blob/baabc5825f3f6640e99fe32887bbeced640f825e/crates/hir-ty/src/infer/expr.rs#L945) in `infer_record_expr`.

`infer_record_expr` returned the error type without reporting anything when the
path didn't resolve to a variant, so typos in struct literals went unnoticed:

```rust
struct Point {}
fn main() { let a = Poin {}; } // no diagnostic
```

Now emits E0422, matching rustc.

Test-only fallout: the `inactive_code::cfg_diagnostics` fixture used `Struct`
without ever declaring it, which only worked because an unresolved path used to
be silent. It now needs the declaration, otherwise E0422 (and `MissingFields`)
fire in a test that has nothing to do with either:

```rust
struct Struct {}
let x = Struct {
    #[cfg(a)] f: 0,
  //^^^^^^^^^^^^^^ weak: code is inactive due to #[cfg] directives: a is disabled
};
```

Covered by new tests in the corresponding `ide-diagnostics` handler; the rest of
the suite passes.

I used Claude Code to navigate the project.

### DIFF
--- a/crates/hir-ty/src/infer.rs
+++ b/crates/hir-ty/src/infer.rs
@@ -373,6 +373,12 @@ pub enum InferenceDiagnostic {
         #[type_visitable(ignore)]
         id: ExprOrPatIdPacked,
     },
+    UnresolvedRecordExpr {
+        #[type_visitable(ignore)]
+        expr: ExprId,
+        #[type_visitable(ignore)]
+        name: Name,
+    },
     // FIXME: This should be emitted in body lowering
     BreakOutsideOfLoop {
         #[type_visitable(ignore)]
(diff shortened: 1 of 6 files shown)


ChayimFriedman2:
The FIXME was here but this is not the place to emit an error. We should do it in one place for all resolves (probably in `PathLoweringContext`). Furthermore I'm not sure we're ready to emit this diagnostic, our name resolution still has bugs and that'll be annoying.

Yarous:
I'm not sure I can figure this out =). So, I'd better close the PR.

PR #23152, opened 2026-08-16
internal: Store token trees in the proc macro server in the representation of the tt crate

ChayimFriedman2:
And not in the representation of the proc macro bridge. We still do not use the contiguous representation the rest of r-a uses, and prefer a tree with `Rc`s (this changed from `Arc` because we don't need thread safety for the bridge) for cheap cloning, but the types are the same.

The reason for that is that conversion to the bridge's types is a lossy conversion, in rustc as well: the internal token tree representation contains details that are lost in this conversion. This is a prerequisite to fix (don't close GH) https://github.com/rust-lang/rust-analyzer/issues/23088 - the only way to fix it properly is to have an additional kind of token tree for doc comments, like rustc does, as evidenced by the fact that if a proc macro would create this macro (that has a doc comment in its matcher), the doc comment would *not* be ignored for matching, but if the macro's matcher would have been passed to the proc macro and it wouldn't have touch it (meaning, not even listing the `TokenTree`s then giving them back), the comment would still be ignored.

It is also required for supporting invisible groups properly for the same reason (there're actually several kinds of invisible delimiters, and the proc macro bridge lossily converts them all into one).

While we're at, I've also tried to untangle the mess of Cargo features of the proc macro server, and reduce the code it contains:

There are now *two* somewhat-orthogonal axes: the `in-rust-tree` feature, and the `in-ra` vs. `in-proc-macro-srv` feature. The former only decides whether we should import rustc crates from crates.io or from the sysroot. The latter decides what code to enable - the proc macro server doesn't need all the code in tt and proc-macro-api. I also moved the proc macro server's `TokenStream` from proc-macro-srv into proc-macro-api and inverted their dependency, as was required for this work.

You cannot enable `in-proc-macro-srv` and disable `in-rust-tree` (in practice this just disables all proc macro server code since such combination can arise when compiling r-a, also proc-macro-api supports this mode for tests), and you are not supposed to enable both `in-ra` and `in-proc-macro-srv`, except for tests. Enabling `in-rust-tree` and disabling `in-proc-macro-srv` is possible, though, and done when compiling r-a in-tree. In the future, this mode may use sysroot crates not just for the proc macro server, like was done in the past.

This depends on https://github.com/rust-lang/rust-analyzer/pull/23079 to not create merge conflicts for myself.


Veykril:
This will severely degrade proc-macro performance. We moved over to the bridge intentionally as concat is a super common operation in proc-macro's which our representation does not lend itself well too https://github.com/rust-lang/rust-analyzer/pull/21097

ChayimFriedman2:
No this will not. We do *not* use the contiguous representation, only the same leaf types.

crates/tt/src/lib.rs
```diff
@@ -8,812 +8,481 @@
 #[cfg(feature = "in-rust-tree")]
 extern crate rustc_driver as _;
 
-#[cfg(not(feature = "in-rust-tree"))]
-extern crate ra_ap_rustc_lexer as rustc_lexer;
-#[cfg(feature = "in-rust-tree")]
-extern crate rustc_lexer;
+stdx::rustc_crates! {
+    extern crate rustc_lexer or ra_ap_rustc_lexer;
+}
```
bit-aloo, line 13: Should this be inside the `in-ra` feature?

crates/tt/src/lib.rs
```diff
@@ -8,812 +8,481 @@
 #[cfg(feature = "in-rust-tree")]
 extern crate rustc_driver as _;
```
bit-aloo, line 9: Seems, like tt doesn't build without in-ra feature.

crates/proc-macro-srv-cli/src/main.rs
```diff
@@ -2,104 +2,13 @@
...
-        match input {
-            "json-legacy" => Ok(ProtocolFormatArg(ProtocolFormat::JsonLegacy)),
-            "bidirectional-postcard-prototype" => {
-                Ok(ProtocolFormatArg(ProtocolFormat::BidirectionalPostcardPrototype))
-            }
-            _ => Err(format!("unknown protocol format: {input}")),
-        }
+    cfg_select! {
+        feature = "in-rust-tree" => proc_macro_srv_cli::main(),
+        _ => Ok(()),
```
bit-aloo, line 12: Should we return an error like before?

```rust
Err(std::io::Error::new(
        std::io::ErrorKind::Unsupported,
        "proc-macro-srv-cli needs to be compiled with the `in-rust-tree` feature to function"
            .to_owned(),
    ))
```

crates/proc-macro-api/src/lib.rs
```diff
@@ -5,30 +5,34 @@
...
 //! processes: Client (RA itself), Server (the external program)
 
-#![cfg_attr(not(feature = "in-rust-tree"), allow(unused_crate_dependencies))]
+#![cfg_attr(feature = "in-rust-tree", feature(rustc_private))]
 #![cfg_attr(
-    feature = "in-rust-tree",
-    feature(proc_macro_internals, proc_macro_diagnostic, proc_macro_span, rustc_private)
+    all(feature = "in-rust-tree", feature = "in-proc-macro-srv"),
+    feature(proc_macro_internals, proc_macro_diagnostic, proc_macro_span)
 )]
```
bit-aloo, line 12: This can be removed now?

crates/proc-macro-api/Cargo.toml
```diff
@@ -22,20 +22,36 @@ indexmap.workspace = true
...
 postcard.workspace = true
 semver.workspace = true
 rayon.workspace = true
 
+[dev-dependencies]
+# Enable both features for test.
+proc-macro-api = { path = "../proc-macro-api", features = [
+    "in-ra",
+    "in-proc-macro-srv",
+] }
```
bit-aloo, line 42: Won't this cause cyclic dep?

bit-aloo (commented):
Just preliminary set of review, need to dig deeper. Is there any way to benchmark this?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23152/83449b2f18064a94f6fea5af041534f516cbb2a7..dc91f05cabb1c41023b9bcf120a962e6f6351cfa)*

crates/proc-macro-api/Cargo.toml
ChayimFriedman2: Yes (for dev-dependencies), but this is not a problem (except for r-a). We already have such cycle for `cfg`.

crates/proc-macro-api/src/lib.rs
ChayimFriedman2: Removed.

crates/proc-macro-srv-cli/src/main.rs
ChayimFriedman2: Done.

crates/tt/src/lib.rs
ChayimFriedman2: Fixed, and added a CI check for that.

crates/tt/src/lib.rs
ChayimFriedman2: No, it's used even without it.

ChayimFriedman2:
> Is there any way to benchmark this?

Shouldn't be necessary. The changes shouldn't affect perf much.

PR #23156, opened 2026-08-16
Update extension packages

BenjaminBrienen:
vscode-languageclient no longer provides `is` as part of its interface, so I vendored it.
It also now supports snippet text edits natively, so I removed the workaround code. This was originally done by @stefnotch in wgsl-analyzer and I am upstreaming it.


stefnotch:
One notable regression is that the new LogOutput will add additional timestamps to logs.

We should look into how to hook into it?

BenjaminBrienen:
https://github.com/microsoft/vscode-languageserver-node/issues/1754

https://github.com/microsoft/vscode-languageserver-node/blob/93d5fe6f443dd6c8fe37f422c271d618edd121af/client/src/node/main.ts#L298-L305

PR #23171, opened 2026-08-17
minor: skip iter excludes 'into_iter' method

A4-Tacks:


ChayimFriedman2 (approved):
This will also skip an inherent method named `into_iter()` (or another trait's method). While uncommon especially on iterators it's possible. That's fine from my side though, I'll leave it to you whether you want to merge as-is or address that.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23171/1af310723edc1fe1917360ed70a2993a42feaf1d..4178939c7791805ceefa0d91af679d7fddc96f73)*

A4-Tacks:
- Inherent: Skip iter completions, focus on `Iterator` based methods
- Other traits: Since `IntoIterator` is a prelude, this is not a problem

PR #23172, opened 2026-08-17
internal: Split VSIX publishing into different jobs and skip duplicates

lnicola:
Four main changes:

 - the Marketplace and OVSX publishing now happens in separate jobs, so retrying one of them doesn't create a new GitHub release
 - we pass `--skip-duplicate`, which will hopefully let us retry when at least one VSIX gets published, but some of them fail (this used to error out because of the duplicate version)
 - we never passed `--pre-release` to `ovsx`, probably because it wasn't supported, but it's available now
 - we only used to set a timeout on the OVSX step, now we always do

AI disclosure: partly written by Gemini Flash 3.6

Dry run at https://github.com/lnicola/rust-analyzer/actions/runs/32037903853.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -199,15 +199,9 @@ jobs:
 
   publish:
     if: ${{ github.repository == 'rust-lang/rust-analyzer' || github.event_name == 'workflow_dispatch' }}
-    name: publish
     runs-on: ubuntu-latest
     needs: ["dist", "dist-x86_64-unknown-linux-musl"]
     steps:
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):

lnicola:
We got to test this right away: Marketplace publishing failed for one VSIX in https://github.com/rust-lang/rust-analyzer/actions/runs/32083869978/job/95557162785#step:6:37, then worked on the second try: https://github.com/rust-lang/rust-analyzer/actions/runs/32083869978/job/95617678003.

PR #23176, opened 2026-08-17
fix: Push a generic params scope for consts

ChayimFriedman2:
Closes https://github.com/rust-lang/rust-analyzer/pull/23174.

Thanks to @jix for the minimization!


ChayimFriedman2:
Technically statics are also a `GenericDefId` (although I think that maybe we should remove that), but they can't have parents so they don't have this problem.

ShoyuVanilla (approved):

PR #23190, opened 2026-08-19
internal: Do not relower the signature again in inference, instead fetch it from other queries

ChayimFriedman2:
Not only this helps perf, this also make it possible to define `AnonConst` as a tracked struct (but this PR doesn't do that yet), because it won't be created twice.

There's a slight regression in tests because I followed rustc and liberated late bounds regions in the signature, and the printing of bound regions is suboptimal. I didn't fix this to not interfere with @dfireBird's work.

For that to work I had to first push the first commit that refactors collection of diagnostics in `hir`. The commit are almost fully separate and can be reviewed separately.


dfireBird:
> followed rustc and liberated late bounds regions in the signature

TBH, we should liberate the late bound regions when doing inference. We could potentially remove this field as well: https://github.com/ChayimFriedman2/rust-analyzer/blob/08383ae310109cda25123fe77fd4521772b746b9/crates/hir-ty/src/lower.rs#L232

ChayimFriedman2:
We still use `LifetimeLoweringMode::LateParam` but we probably can not to; that's for later PR though.

Veykril (approved):

PR #23196, opened 2026-08-19
internal: Do not the same types twice in hir-ty lowering

ChayimFriedman2:
Again for the intent of making `AnonConst` a tracked struct (as well as for perf).

We still have one problematic query: `resolve_type_param_assoc_type_shorthand()`. It is problematic and I need to think what to do with it.

And a drive-by fix in the second commit.

Best reviewed commit by commit. The commits are fully separate.


crates/hir-ty/src/next_solver/predicate.rs
```diff
@@ -272,6 +272,13 @@ impl<'db> std::fmt::Debug for Clauses<'db> {
     }
 }
 
+impl<'db> Default for Clauses<'db> {
+    #[inline]
+    fn default() -> Self {
+        Clauses::empty(DbInterner::conjure())
+    }
+}
+
```
Veykril, line 281: Do we have to add this? I'd prefer not to ideally `conjure` shouldn't exist in the first place, so I'd rather remove uses of it than add them

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23196/32a3346dff5b26e16227affd5af933e15ebc598b..f43fa60cd7c0689ab37830ee5552313712f33c8e)*

crates/hir-ty/src/lower.rs
```diff
@@ -1851,26 +1854,24 @@ pub(crate) fn impl_self_ty_with_diagnostics_cycle_result<'db>(
...
-    type Target = ArenaMap<LocalTypeOrConstParamId, StoredTy>;
-
-    fn deref(&self) -> &Self::Target {
-        &self.map
+impl ConstParamTypes {
+    #[inline]
+    pub fn get(&self, local_id: LocalTypeOrConstParamId) -> Ty<'_> {
+        match self.map.get(local_id) {
+            Some(ty) => ty.as_ref(),
+            None => Ty::new_error(DbInterner::conjure(), ErrorGuaranteed),
```
Veykril, line 1873: likewise threading the database through here shouldn't be much work

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23196/32a3346dff5b26e16227affd5af933e15ebc598b..f43fa60cd7c0689ab37830ee5552313712f33c8e)*

crates/hir-ty/src/next_solver/predicate.rs
ChayimFriedman2: I don't like `conjure()` too much but I wouldn't want to complicate the API just to avoid it.

Alternatively, we don't really need the database in order to intern and only do that for API reasons. We can avoid that and probably get rid of `conjure()` completely.

PR #23197, opened 2026-08-20
internal: A bunch of cleanups and micro-optimizations

ChayimFriedman2:
Details in individual commits.

Reviewing commit-by-commit is recommended; the commits are fully separate.


crates/hir-ty/src/next_solver.rs
```diff
@@ -143,11 +142,11 @@ impl std::fmt::Debug for DefaultAny<'_> {
 }
 
 #[inline]
-pub fn default_types<'db>(db: &'db dyn HirDatabase) -> &'db DefaultAny<'db> {
+pub fn default_types<'db>() -> &'db DefaultAny<'db> {
     static TYPES: OnceLock<DefaultAny<'static>> = OnceLock::new();
 
-    let interner = DbInterner::new_no_crate(db);
     TYPES.get_or_init(|| {
+        let interner = DbInterner::conjure();
```
nicolas-guichard, line 149: `DbInterner::conjure` has the comment `// FIXME(next-solver): remove this method` and I see changes like https://github.com/rust-lang/rust-analyzer/pull/22119 which went the other way. Is this really desirable?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23197/32a3346dff5b26e16227affd5af933e15ebc598b..506b1e4ee91505a79ef857aa2e1d1d2f49129ba8)*

crates/hir-ty/src/next_solver/interner.rs
```diff
@@ -96,6 +96,16 @@ macro_rules! interned_slice {
...
+                if slice.is_empty() {
+                    // Common case: avoid looking up the empty slice.
+                    Self::empty()
+                } else {
+                    Self::new_from_slice_no_empty(slice)
+                }
+            }
+
+            #[inline]
+            pub(crate) fn new_from_slice_no_empty(slice: &[$ty_db]) -> Self {
```
nicolas-guichard: I'd expect an `assert!(slice.is_empty());`, or when is it okay to call `new_from_slice_no_empty` with an actually empty slice?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23197/32a3346dff5b26e16227affd5af933e15ebc598b..506b1e4ee91505a79ef857aa2e1d1d2f49129ba8)*

crates/hir-ty/src/next_solver.rs
```diff
@@ -166,73 +166,73 @@ pub fn default_types<'db>() -> &'db DefaultAny<'db> {
             ty.as_ref()
         };
         let create_generic_args = |slice| {
-            let ty = GenericArgs::new_from_slice(slice);
+            let ty = GenericArgs::new_from_slice_no_empty(slice);
```
nicolas-guichard: The `create_*` functions are explicitly called with the empty slice below to pre-intern it. This is confusing.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23197/32a3346dff5b26e16227affd5af933e15ebc598b..506b1e4ee91505a79ef857aa2e1d1d2f49129ba8)*

crates/hir-ty/src/next_solver.rs
ChayimFriedman2: I... don't know. It would be nice if we could remove it, but I doubt that. I believe it's the lesser evil.

crates/hir-ty/src/next_solver/interner.rs
ChayimFriedman2: This method is intended *specifically* for empty ranges (mostly), since we need to intern them somewhere (see its callers).

crates/hir-ty/src/next_solver.rs
ChayimFriedman2: If we call `new_from_slice()` here, it's deadlock since it accesses what we're interning here.

crates/hir-ty/src/next_solver/interner.rs
nicolas-guichard: Its closest caller just above calls it in the non-empty case and it's called no_empty. The logic looks right but the name is super confusing. I think something like `new_from_slice_force_interned` is clearer?

ChayimFriedman2: Okay I renamed the method and added a doc comment.

nicolas-guichard (approved):
I was not able to tell if the conjure change was good or not (you're certainly better placed than me to tell), but that's gone now and the rest looks good to me.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23197/682a84e95b5a52cf06e9822fcaa4f628738554bf..3496d68dfe270650d7c85df09be6a4aaeba24111)*

PR #23206, opened 2026-08-21
fix: Switch from `temp-dir` to a homemade `NamedTempFile` implementation

### LINKED ISSUE #23203: ra leaves empty `proc-macro-srv` dir behind
After fuzzing for a couple of days, noticed that inside `/tmp/` I had accumulated more than 40.000 directories like `proc-macro-srv15c70d-0`

`rust-analyzer diagnostics .` on an empty crate (`cargo new`) is enough to make the stale folder appear.
I would think that at least in case of the happy path (no ra panic), leftover files are removed before the application finally terminates.

ChayimFriedman2:
That also deletes the file when the process exits without being dropped; this is important especially for the proc macro server since it is killed and does not exit normally.

Linux and the BSDs unlink the file then access it via `/proc/self/fd` or `/dev/fd`; Windows has a dedicated API for that; and macOS unfortunately does not support that (the file is not removed after being unlinked, but you cannot access it anymore via `/dev/fd`, or at least that's what the AI said - I don't have a macOS machine to check).

Fixes https://github.com/rust-lang/rust-analyzer/issues/23203.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1890,7 +1890,7 @@ dependencies = [
  "paths",
  "proc-macro-test",
  "span",
- "temp-dir",
+ "stdx",
 ]
 
 [[package]]
@@ -1966,7 +1966,6 @@ dependencies = [
  "serde_json",
  "span",
  "stdx",
- "temp-dir",
  "toml",
  "toolchain",
  "tracing",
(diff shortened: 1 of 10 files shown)


ShoyuVanilla (approved):
I was the one who has introduced `temp-dir` to our dependency tree because it has pretty simple implementation. But your implementation looks very clever (especially for the unlink part) while being not too verbose 😄 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23206/5f8813d64fff2e293dcc99d54a21f6f4edc88d50..acfdd30e2c4a904ee577f6bc8b7b23821ac89b72)*

ChayimFriedman2:
> But your implementation looks very clever (especially for the unlink part)

Credit where credit is due; this was the AI's idea (not original probably). Of course I verified in the POSIX specification that `unlink()` does not remove the file if it's open.

PR #23209, opened 2026-08-21
internal: Offer `replace_arith` on `.pow()` and unary negation

rootkiller6788:
Extends the `replace_arith_with_*` assists (checked/wrapping/saturating/strict) beyond plain binary `+ - * /` so they are also offered on:

- `x.pow(e)` method calls → `x.checked_pow(e)` / `x.wrapping_pow(e)` / `x.saturating_pow(e)` / `x.strict_pow(e)`
- unary negation `-x` → `x.checked_neg()` / `x.wrapping_neg()` / `x.saturating_neg()` / `x.strict_neg()`

These were the remaining arithmetic forms with overflow-specified variants that the assist did not cover (issue rust-lang/rust-analyzer#23153).

The handler now parses the expression under the cursor into a small `ArithExpr` enum (binary op / `.pow` call / unary negation) so the rewrite logic is shared. Reference operands are dereferenced as before, parentheses are added where needed for precedence (e.g. `--x` → `(-x).wrapping_neg()`), and `.pow` is only matched with exactly one argument.

Tests: `cargo test -p ide-assists` (2818 passed, 0 failed), `cargo fmt --check` clean, `cargo clippy -p ide-assists --all-targets` clean.

### DIFF
--- a/crates/ide-assists/src/handlers/replace_arith_op.rs
+++ b/crates/ide-assists/src/handlers/replace_arith_op.rs
@@ -1,7 +1,7 @@
 use ide_db::assists::{AssistId, GroupLabel};
 use syntax::{
     AstNode, T,
-    ast::{self, ArithOp, BinaryOp},
+    ast::{self, ArithOp, BinaryOp, HasArgList, UnaryOp, syntax_factory::SyntaxFactory},
 };
 
 use crate::{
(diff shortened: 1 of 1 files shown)


oli-obk:
The account has been banned for spamming

PR #23212, opened 2026-08-21
fix: panic on deref of unresolved aliases

Wilfred:
Previously, code like `let &(x, y) = unknown_var;` would produce a panic of the form:

    deref projection of non-dereferenceable ty PlaceTy { ... }

Ensure that we normalise aliases consistently before projection in MIR, and add a test.

AI disclosure: Code partly written by GPT-5.6, commit message and review entirely done by a human.

### DIFF
--- a/crates/hir-ty/src/mir/lower/pattern_matching.rs
+++ b/crates/hir-ty/src/mir/lower/pattern_matching.rs
@@ -491,6 +491,12 @@ impl<'db> MirLowerCtx<'_, 'db> {
                 )?
             }
             Pat::Ref { pat, mutability: _ } => {
+                let ty = cond_place.ty(&self.result, &self.infcx, self.env).ty;
+                if !ty.is_ref() {
+                    return Err(MirLowerError::TypeError(
+                        "non reference type matched with reference pattern",
+                    ));
+                }
                 let cond_place = cond_place.project(ProjectionElem::Deref);
                 self.pattern_match_inner(current, current_else, cond_place, *pat, mode)?
             }
--- a/crates/hir-ty/src/mir/lower/tests.rs
+++ b/crates/hir-ty/src/mir/lower/tests.rs
@@ -49,3 +49,19 @@ fn foo() {
     "#,
     );
 }
+
+#[test]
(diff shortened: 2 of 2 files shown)


Wilfred:
Very similar to #22707, so I've tried to follow the same MIR lowering approach.

crates/hir-ty/src/mir/lower/pattern_matching.rs
```diff
@@ -491,6 +491,13 @@ impl<'db> MirLowerCtx<'_, 'db> {
                 )?
             }
             Pat::Ref { pat, mutability: _ } => {
+                let ty = cond_place.ty(&self.result, &self.infcx, self.env).ty;
+                let ty = PlaceTy::normalize_projection_ty(&self.infcx, self.env, ty);
```
ChayimFriedman2: Normalization shouldn't be required, just like in the other PR. Why did you add it?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23212/d2e55da49132fa70a13dfbdc99122432b02cf464..5ef0dd435e62c4b387e7be89a77b3d3616a1156b)*

ChayimFriedman2 (changes requested):

crates/hir-ty/src/mir/lower/pattern_matching.rs
Wilfred: Oops, since all the cases I'd minimised from contained an alias, so I started from looking at other cases in the code where we handle aliases.

I've simplified the code.

ChayimFriedman2 (approved):
I dislike the fact that MIR building keeps accumulating more and more checks that belong in inference, but we need to fix the panics (and we cannot let go of MIR). I hope one day we'll be able to just not lower MIR for functions that have inference diagnostics.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23212/f2211ef8887701b58fcb6525473ac8d9b4aea416..690251bbd232d808b7ccd1b54a3905eb1d0dd375)*

PR #23213, opened 2026-08-21
feat: Allow “Extract variable” to be invoked on field names in record expressions.

kpreid:
This change allows a “mistake” I make very often to succeed: putting the cursor on “foo” in `Struct { foo: bar() }` when I want to extract `let foo = bar();`.

It is also preparation for being able to extract multiple field expressions at once (rust-lang/rust-analyzer#21863), because it generalizes the analysis of the selection to be able to work on nodes that are not themselves expressions.

### Notes

The added `take_while()` and the tests `dont_extract_in_pattern_*` are an attempt to preserve the behavior added in rust-lang/rust-analyzer#18982, but that PR did not add any tests or clearly specify what effect it was intended to have, so I may have gotten it wrong.

This is only my second contribution to assists, so all feedback is welcome.

### DIFF
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -76,10 +76,16 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
         if let Some(t) = ctx.token_at_offset().find(|it| it.kind() == T![;]) {
             t.parent().and_then(ast::ExprStmt::cast)?.syntax().clone()
         } else {
-            let expr = ancestors_at_offset(ctx.source_file().syntax(), ctx.offset())
-                .next()
-                .and_then(ast::Expr::cast)?;
-            expr.syntax().ancestors().find_map(valid_target_expr(ctx))?.syntax().clone()
+            // Offer the assist only if the nearest syntax node is an expression, or a record
+            // field, or a record field’s name. This prevents the assist from appearing when
+            // it is unlikely to be relevant, such as when the cursor is in a pattern.
+            // (If we did not want to restrict it this way, we could just apply
+            // `valid_target_expr()` to all ancestors.)
+            let expr_or_field = ancestors_at_offset(ctx.source_file().syntax(), ctx.offset())
+                .find(|it| !ast::NameRef::can_cast(it.kind()))
+                .and_then(either::Either::<ast::Expr, ast::RecordExprField>::cast)?;
+
+            expr_or_field.syntax().ancestors().find_map(valid_target_expr(ctx))?.syntax().clone()
         }
     } else {
         match ctx.covering_element() {
@@ -366,6 +372,11 @@ fn valid_target_expr(ctx: &AssistContext<'_, '_>) -> impl Fn(SyntaxNode) -> Opti
             let path_expr = ast::PathExpr::cast(node)?;
             let path_resolution = ctx.sema.resolve_path(&path_expr.path()?)?;
             like_const_value(ctx, path_resolution).then_some(path_expr.into())
+        }
+        SyntaxKind::RECORD_EXPR_FIELD => {
+            // If we are on `k` in `Struct { k: v }`, then extract `v`.
+            let record_field = ast::RecordExprField::cast(node)?;
+            record_field.expr()
         }
         _ => ast::Expr::cast(node),
     }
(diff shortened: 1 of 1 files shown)


crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -76,10 +76,18 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
...
-            expr.syntax().ancestors().find_map(valid_target_expr(ctx))?.syntax().clone()
+            ancestors_at_offset(ctx.source_file().syntax(), ctx.offset())
+                .skip(1)
+                .take_while(|node| {
+                    // Stop ascending if we encounter a node that is not one of the kinds we want.
+                    // This prevents the assist from applying to an undesirably wide containing
+                    // expression.
+                    let kind = node.kind();
+                    ast::Expr::can_cast(kind) || kind == SyntaxKind::RECORD_EXPR_FIELD
+                })
```
A4-Tacks: Why add this? I noticed without this, `dont_extract_in_pattern` would still pass through
Try not to break


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23213/d2e55da49132fa70a13dfbdc99122432b02cf464..af38837a24ba47441505caca3b45053189b9ec8a)*

crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -1586,6 +1604,63 @@ struct S {
...
+fn main() {
+    let $0foo = 1 + 1;
+    S { foo }
+}
+"#,
+            "Extract into variable",
+        )
+    }
+
+    /// Selecting an entire field of a record expression will extract that field’s value expression.
```
A4-Tacks: Redundant (doc)comments


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23213/d2e55da49132fa70a13dfbdc99122432b02cf464..af38837a24ba47441505caca3b45053189b9ec8a)*

crates/ide-assists/src/handlers/extract_variable.rs
kpreid: On revisiting this, there were several problems.

* The `dont_extract_in_pattern` test did not actually test this code path, because the test had a selection and this is the empty-selection code path. I’ve fixed that by adding another test, and also split out those tests to a separate commit to clarify that they are testing existing behavior.
* I don’t actually need to rewrite this as much as I did. I’ve now kept the existing structure, and only changed `ast::Expr::cast` into the broader condition needed to operate on field names.
* The `skip(1)` was entirely my mistake in understanding the existing code, and has been removed.

crates/ide-assists/src/handlers/extract_variable.rs
kpreid: Removed.

crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -76,10 +76,22 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
...
+                    let kind = node.kind();
+                    ast::Expr::can_cast(kind)
+                        || kind == SyntaxKind::RECORD_EXPR_FIELD
+                        || (kind == SyntaxKind::NAME_REF
+                            && node
+                                .parent()
+                                .is_some_and(|p| p.kind() == SyntaxKind::RECORD_EXPR_FIELD))
+                })?;
+
+            node.ancestors().find_map(valid_target_expr(ctx))?.syntax().clone()
```
A4-Tacks: Don't be so complicated, just:
```diff
             let expr = ancestors_at_offset(ctx.source_file().syntax(), ctx.offset())
-                .next()
-                .and_then(ast::Expr::cast)?;
+                .find(|it| !ast::NameRef::can_cast(it.kind()))
+                .and_then(either::Either::<ast::Expr, ast::RecordExprField>::cast)?;
             expr.syntax().ancestors().find_map(valid_target_expr(ctx))?.syntax().clone()
```


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23213/014d54b685ccc00f0774d755d1a4c889f7e8e512..68abcda90b034852e18c5fa9ffddb1690ee027e1)*

kpreid: Done.

crates/ide-assists/src/handlers/extract_variable.rs
```diff
@@ -76,10 +76,16 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
...
         } else {
-            let expr = ancestors_at_offset(ctx.source_file().syntax(), ctx.offset())
-                .next()
-                .and_then(ast::Expr::cast)?;
-            expr.syntax().ancestors().find_map(valid_target_expr(ctx))?.syntax().clone()
+            // Offer the assist only if the nearest syntax node is an expression, or a record
+            // field, or a record field’s name. This prevents the assist from appearing when
+            // it is unlikely to be relevant, such as when the cursor is in a pattern.
+            // (If we did not want to restrict it this way, we could just apply
+            // `valid_target_expr()` to all ancestors.)
```
A4-Tacks, line 83: Redundant comments, we have testing


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23213/014d54b685ccc00f0774d755d1a4c889f7e8e512..ce17ff9b15e8b688f1f4d94bc2588808e7805141)*

A4-Tacks (approved):

crates/ide-assists/src/handlers/extract_variable.rs
kpreid: This comment is not for specifying the desired functionality, but for someone reading the code (like me) to understand what the goal of this piece of the code is. Not having this part explained was a significant obstacle; I had to dig into git-blame to understand what the code was supposed to achieve.

PR #23216, opened 2026-08-22
fix: adds-arrow unmap ranges when fn inside macro

A4-Tacks:
Example
---
```rust
macro_rules! identity { ($($t:tt)*) => {$($t)*}; }
identity! {
    fn foo() u$0
}
```

**Before this PR**

```rust
macro_r-> ules! identity { ($($t:tt)*) => {$($t)*}; }
identity! {
    fn foo() u32
}
```

**After this PR**

```rust
macro_rules! identity { ($($t:tt)*) => {$($t)*}; }
identity! {
    fn foo() -> u32
}
```

### DIFF
--- a/crates/ide-completion/src/context.rs
+++ b/crates/ide-completion/src/context.rs
@@ -105,7 +105,7 @@ impl PathCompletionCtx<'_> {
 
     pub(crate) fn required_thin_arrow(
         &self,
-        unmap: impl Fn(&syntax::SyntaxNode) -> Option<TextRange>,
+        sema: &Semantics<'_, RootDatabase>,
     ) -> Option<(&'static str, TextSize)> {
         let PathKind::Type {
             location:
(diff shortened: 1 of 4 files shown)


crates/ide-completion/src/context.rs
```diff
@@ -103,7 +103,10 @@ impl PathCompletionCtx<'_> {
         )
     }
 
-    pub(crate) fn required_thin_arrow(&self) -> Option<(&'static str, TextSize)> {
+    pub(crate) fn required_thin_arrow(
+        &self,
+        unmap: impl Fn(&syntax::SyntaxNode) -> Option<TextRange>,
```
ChayimFriedman2: Can't we just always use the original range? Why the callback?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23216/d2e55da49132fa70a13dfbdc99122432b02cf464..d7cca6beaecc2c9b8b7759b40c181ecdab92b7af)*

A4-Tacks: One usage is `is_some`, always using original range may result in performance loss

Additionally, I noticed that there are no other APIs in `context.rs` that use `sema` or `ctx: &CompletionContext<'_, '_>` parameters

ChayimFriedman2: The perf loss will be negligible. The callback complicates the code and adds monomorphization costs, I don't think it's worth it.

ChayimFriedman2 (approved):

PR #23219, opened 2026-08-23
Fix auto-import namespace collision merges

### LINKED ISSUE #11278: auto-import on `alloc::vec` macro instead imports `alloc::vec::{Self}` module
If I start a new library, with this code:
```rust
#![no_std]
extern crate alloc;

pub fn foo() -> Vec<u32> {
    vec![]
}
```

I get two unresolved symbols, `Vec` and `vec`. If I allow rust-analyzer to import those symbols, I first select "import \`alloc::vec::Vec\`", and then "import \`alloc::vec\`".  This results in the added line:
```rust
use alloc::vec::{Vec, self};
```

This doesn't work (code will not compile), because `self` refers to the module `alloc::vec`, and this code needs the *macro* `alloc::vec`.

**rust-analyzer version**: 0add6e95e 2021-12-20 stable

**rustc version**: 1.58.0

### LINKED ISSUE #19816: Auto import unresolved module and constant value with similarly name
**rust-analyzer version**: rust-analyzer 1.88.0-nightly (a15cce2 2025-04-17)

**rustc version**: rustc 1.88.0-nightly (a15cce269 2025-04-17)

**editor or extension**: VIM9.1 coc-rust-analyzer

```rust
mod foo {
    pub mod bar {
        pub fn baz() {}
    }
    pub fn bar() {}
}
use foo::bar;

fn main() {
    let _ = bar();
    baz
}
```

On line 11 complete `baz`:

```rust
mod foo {
    pub mod bar {
        pub fn baz() {}
    }
    pub fn bar() {}
}
use foo::bar::{self, baz};

fn main() {
    let _ = bar();
    baz();
}
```

rustc: `` error[E0423]: expected function, found module `bar` ``

**Expect**:

```rust
mod foo {
    pub mod bar {
        pub fn baz() {}
    }
    pub fn bar() {}
}
use foo::{bar, bar::baz};

fn main() {
    let _ = bar();
    baz();
}
```

### LINKED ISSUE #20790: auto-import sometimes breaks with `self` imports
**rust-analyzer version**: 1.90.0 (1159e78 2025-09-14), 1.92.0-nightly (5c7ae0c 2025-10-02)

**rustc version**: 1.90.0 (1159e78c4 2025-09-14), 1.92.0-nightly (5c7ae0c7e 2025-10-02)

**editor or extension**: Zed

**relevant settings**: none, as far as I can tell

**repository link**: https://github.com/generic-daw/generic-daw

**instructions to reproduce**:
1. clone above repo
2. disable nightly-only features present in `.rustfmt.toml` (not strictly necessary, this reproduces without)
3. navigate to https://github.com/generic-daw/generic-daw/blob/716331b78cec8fad10f8e09dbf92eb4ad76beae8/generic_daw_gui/src/arrangement_view.rs#L984
4. remove the `text::` prefix of `text::Wrapping::None`
5. auto-import `Wrapping` from `iced::widget::text::Wrapping`
6. observe that the import has been changed to `text::{self, Wrapping}`, and that all calls to the `text` function have broken

I'm sadly not sure how to minimize this, I've just encountered this specific case a bunch in this project and only just noticed that the aforementioned nightly-only rustfmt options aren't to blame.

ghost:
Fixes module/function collisions (rust-lang/rust-analyzer#19816￼, rust-lang/rust-analyzer#20790) and module/macro collisions (rust-lang/rust-analyzer#11278).

Fixes https://github.com/rust-lang/rust-analyzer/issues/19816
Fixes https://github.com/rust-lang/rust-analyzer/issues/20790
Fixes https://github.com/rust-lang/rust-analyzer/issues/11278

it was assisted by a LLM

### DIFF
--- a/crates/ide-db/src/imports/insert_use.rs
+++ b/crates/ide-db/src/imports/insert_use.rs
@@ -17,7 +17,7 @@ use crate::{
     RootDatabase,
     imports::merge_imports::{
         MergeBehavior, NormalizationStyle, common_prefix, eq_attrs, eq_visibility,
-        try_merge_imports, try_merge_imports_preserving_path, use_tree_cmp, wrap_in_tree_list,
+        try_merge_imports, use_tree_cmp, wrap_in_tree_list,
     },
 };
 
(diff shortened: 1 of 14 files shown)


crates/ide-db/src/imports/merge_imports.rs
```diff
@@ -64,6 +64,66 @@ pub fn try_merge_imports(
     make_use_with_tree(lhs, use_tree)
 }
 
+/// Merges imports without replacing the protected path with a nested `self` import.
+pub(super) fn try_merge_imports_preserving_path(
```
ChayimFriedman2: This should not be a separate method and absolutely not duplicate all code.

crates/ide-db/src/imports/insert_use.rs
```diff
@@ -157,7 +157,37 @@ pub fn insert_use_with_editor(
...
+        path,
+        cfg,
+        None,
+        syntax_editor,
+        try_merge_imports,
+    );
+}
+
+/// Inserts a use without turning a path that resolves in multiple namespaces into a nested `self`.
+pub fn insert_use_with_editor_preserving_namespaces(
```
ChayimFriedman2: Not a fan of this combinatorial explosion of helpers. When do we *not* need to preserve existing resolutions?

ChayimFriedman2 (changes requested):
We can't fix this properly because a proper fix will be to only do that when there's actual collisions (things will resolve differently), and that we can't check. That said this approach is good enough for me.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23219/5c156cdfb047ea171ba255dcdcd86236d98a389e..1cef731a8b35a140bedc648037b0870e7ebafca0)*

PR #23221, opened 2026-08-23
perf: avoid unnecessary conversions in with_impl_or_trait_item

hosseinmobarakian:
I optimized the iterator chain in with_impl_or_trait_item by replacing the filter_map -> find -> map sequence with a single find_map

This prevents unnecessary conversions to every item

### DIFF
--- a/crates/hir-ty/src/method_resolution/probe.rs
+++ b/crates/hir-ty/src/method_resolution/probe.rs
@@ -148,17 +148,18 @@ impl PickConstraintsForShadowed {
 
     fn candidate_may_shadow(&self, candidate: &Candidate<'_>) -> bool {
         // An item never shadows itself
-        candidate.item != self.def_id
+        candidate.item != self.def_id &&
             // and we're only concerned about inherent impls doing the shadowing.
             // Shadowing can only occur if the shadowed is further along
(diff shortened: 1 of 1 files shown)


ChayimFriedman2:
I have news for you: Rust iterators are zero-cost (except, well, in some cases. Not applicable here) and there is absolutely zero perf benefit for that.

Also, you've changed many unrelated things. Do you use an LLM?

hosseinmobarakian:
No my IDE formatter did that

hosseinmobarakian:
I have news for you too: Rust iterators' zero-cost is right, but the implementation logic makes a difference in execution.

You can check the compiled Assembly of the two logics here:

https://godbolt.org/z/a64Pdh9G1 //yours
https://godbolt.org/z/c376zvGTP  //my implementaion

You can't say there is no difference. @ChayimFriedman2

ChayimFriedman2:
You checked without optimizations.

With optimization there's still *a difference*, but to know if it matters you need to know to read assembly (or benchmark). I didn't bother to do that here, but it should not matter.

Also, we don't commit micro-optimizations like this unless there's a benchmark that proves they're worthy. **Not** a micro-benchmark: a benchmark that shows they speed up the total execution.

PR #23227, opened 2026-08-24
perf: Remove `UseId`s from `ItemScope`

ChayimFriedman2:
They're not used, and I want to shrink `ItemScope` because it consumes a significant amount of memory (this is only the first step).

It'd be nice if we could remove `UseId` completely (it's non-incremental and almost unused), but it's used in the import map, so I gave up for now.


ChayimFriedman2:
Closing since I want to do that in a more principled way.

PR #23233, opened 2026-08-25
internal: Require rustc citation for analysis changes when using AI

ChayimFriedman2:
[As I asked on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/LLM.20policy/near/619098018).

Do you think we should exempt maintainers from this? I have no problem with doing that (due to the higher trust), and I myself barely use AI at all, but treating maintainers differently might seem unfair to some.

### DIFF
--- a/AI_POLICY.md
+++ b/AI_POLICY.md
@@ -30,6 +30,14 @@ E-easy issues are usually easier for maintainers to just fix directly than write
 
 AI *may* be used to understand the codebase for E-easy+E-has-instructions contributions, but not to write any code.
 
+When using AI to author changes to *analysis* - the code responsible for analyzing Rust code and not for implementing IDE features, including
+but not limited to: type inference, MIR, name resolution, macro expansion - generally anything in the crates `parser`, `mbe`, `hir-expand`, `hir-def`, `hir-ty`,
+although there are exceptions; **including when using AI only to analyze bugs and not to write code**, you are required to include a citation
+of the rustc code responsible for the change you did, along with an explanation of how your change follows from it in case this is not immediately clear.
+
+The reason for that is that it is almost impossible to be fully correct in analysis if we implement things differently from rustc. We should not guess
+how to fix bugs in analysis without looking at the rustc code.
+
 This policy was adapted from [uv's AI policy].
 
 [uv's AI policy]: https://github.com/astral-sh/.github/blob/c5187e200db51bfe11d56e13053d29bd3793fdd8/AI_POLICY.md
--- a/AI_POLICY.md
+++ b/AI_POLICY.md
@@ -38,6 +38,6 @@ of the rustc code responsible for the change you did, along with an explanation 
 The reason for that is that it is almost impossible to be fully correct in analysis if we implement things differently from rustc. We should not guess
 how to fix bugs in analysis without looking at the rustc code.
 
-This policy was adapted from [uv's AI policy].
+This policy was inspired by [uv's AI policy].
 
 [uv's AI policy]: https://github.com/astral-sh/.github/blob/c5187e200db51bfe11d56e13053d29bd3793fdd8/AI_POLICY.md


BenjaminBrienen (approved):

ShoyuVanilla:
I think we should impose it to maintainers as well, not only because the fairness but also b/c when we scrutinize the implementation status quo later on, it would be helpful to look into the cited rustc code with the exact revision as they change from time to time.

BTW, should we separate the new paragraph from *This policy was adapted from uv's AI policy.* part?

AI_POLICY.md
```diff
@@ -30,6 +30,16 @@ E-easy issues are usually easier for maintainers to just fix directly than write
 
 AI *may* be used to understand the codebase for E-easy+E-has-instructions contributions, but not to write any code.
 
+When using AI to author changes to *analysis* - the code responsible for analyzing Rust code and not for implementing IDE features, including
+but not limited to: type inference, MIR, name resolution, macro expansion - generally anything in the crates `parser`, `mbe`, `hir-expand`, `hir-def`, `hir-ty`,
+although there are exceptions; **including when using AI only to analyze bugs and not to write code**, you are required to include a citation
+of the rustc code responsible for the change you did, along with an explanation of how your change follows from it in case this is not immediately clear.
```
Veykril, line 36: to be honest I fail to see how this is AI specific, if I make the change manually myself I'd assume I would need to back the change up with citations all the same?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23233/014d54b685ccc00f0774d755d1a4c889f7e8e512..d0c9b907010f185e76e403d8b3e7c9975f318a2e)*

AI_POLICY.md
```diff
@@ -30,6 +30,16 @@ E-easy issues are usually easier for maintainers to just fix directly than write
...
+
+The reason for that is that it is almost impossible to be fully correct in analysis if we implement things differently from rustc. We should not guess
+how to fix bugs in analysis without looking at the rustc code. When not using AI, the maintainers agree to take the responsibility to verify that
+(although of course providing this information is still appreciated). When using AI, partially due to trusting it less (especially for new contributors)
```
Veykril: I guess due to this, but that also confuses me. Authoring with AI does not free you from responsibility of the code change?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23233/014d54b685ccc00f0774d755d1a4c889f7e8e512..d0c9b907010f185e76e403d8b3e7c9975f318a2e)*

AI_POLICY.md
ChayimFriedman2: Like I said, it isn't AI specific; it's just that for human-authored changes I am more willing to do that myself (seeing this as part of my job as a reviewer), and that people are sending AI-generated PRs in too big volumes.

AI_POLICY.md
Veykril: That said, I follow now, but I think the wording of this paragraph should be touched up

AI_POLICY.md
Veykril: In fact, I think we should move or duplicate this into our agents.md file as a hard rule. That will make the LLM dig into rustc sources immediately.

AI_POLICY.md
ChayimFriedman2: I welcome suggestions.

AI_POLICY.md
ChayimFriedman2: I wanted to have a policy so I can require that from contributors; and this is the place from that. Putting this in AGENTS.md is helpful but not my goal.

AI_POLICY.md
Veykril: I think I would just drop the

> When not using AI, the maintainers agree to take the responsibility to verify that
(although of course providing this information is still appreciated). When using AI, partially due to trusting it less (especially for new contributors) 
and partially due to the easiness and amount of the process when AI is used, we require the contributor to do that instead.

part altogether. I don't see a reason to justify this in the text. We just require the contributor to backup their AI generated claims.

AI_POLICY.md
Veykril: Yea, moving it is the wrong call sorry, duplicating would be beneficial though. Or frankly we should probably just hard link the policy in the agents file with read instructions to also force-feed it to the llm.

AI_POLICY.md
ChayimFriedman2: Done.

ChayimFriedman2:
> BTW, should we separate the new paragraph from This policy was adapted from uv's AI policy. part?

We already made some changes from uv's policy, so maybe we should just drop it, or mention it only as inspiration?

Veykril (approved):

Veykril:
Yea I think we can reword it as inspiration or so

ChayimFriedman2:
I think I've got enough agreements so merging.

AI_POLICY.md
```diff
@@ -30,6 +30,14 @@ E-easy issues are usually easier for maintainers to just fix directly than write
 
 AI *may* be used to understand the codebase for E-easy+E-has-instructions contributions, but not to write any code.
 
-This policy was adapted from [uv's AI policy].
+When using AI to author changes to *analysis* - the code responsible for analyzing Rust code and not for implementing IDE features, including
+but not limited to: type inference, MIR, name resolution, macro expansion - generally anything in the crates `parser`, `mbe`, `hir-expand`, `hir-def`, `hir-ty`,
+although there are exceptions; **including when using AI only to analyze bugs and not to write code**, you are required to include a citation
```
Wilfred, line 35: I'm struggling slightly to parse this: does "exceptions" mean that sometimes those crates don't need references to rustc, or that analysing bugs with AI is an exception?

ChayimFriedman2: The former - some parts of those crates (e.g. import maps) actually implements IDE functionality.

PR #23235, opened 2026-08-26
fix: Fix some subtle bugs in docs rendering

ChayimFriedman2:
Match the behavior of rustdoc *precisely*:
 - Fix handling of mixed sugared and desugared docs.
 - Only trim spaces and tabs for indentation.
 - Do not use a separate indentation for outlined module.


crates/ide/src/syntax_highlighting/test_data/highlight_doctest.html
```diff
@@ -161,18 +161,18 @@
...
+<span class="comment documentation">/// </span><span class="keyword injected">let</span><span class="none injected"> </span><span class="punctuation injected">_</span><span class="none injected"> </span><span class="operator injected">=</span><span class="none injected"> </span><span class="function injected">example</span><span class="parenthesis injected">(</span><span class="operator injected">&</span><span class="bracket injected">[</span><span class="numeric_literal injected">1</span><span class="comma injected">,</span><span class="none injected"> </span><span class="numeric_literal injected">2</span><span class="comma injected">,</span><span class="none injected"> </span><span class="numeric_literal injected">3</span><span class="bracket injected">]</span><span class="parenthesis injected">)</span><span class="semicolon injected">;</span>
```
ChayimFriedman2, line 164: This test only removed the `none injected` classes for the leading space, since it's now trimmed as indentation thus unmapped. No real effect.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23235/014d54b685ccc00f0774d755d1a4c889f7e8e512..81f0eff3acf14ddf5ce641cbe5b96d58bb41c31d)*

crates/ide/src/hover/tests.rs
```diff
@@ -2626,7 +2626,7 @@ fn bar() { fo$0o(); }
 
             ---
 
-            \<- `　` here
+            　\<- `　` here
```
ChayimFriedman2, line 2629: I verified with the JSON version of rustdoc that we should indeed emit this non-ASCII space here (besides the fact that I copied that from rustdoc's code).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23235/014d54b685ccc00f0774d755d1a4c889f7e8e512..81f0eff3acf14ddf5ce641cbe5b96d58bb41c31d)*

Veykril (approved):

PR #23238, opened 2026-08-26
fix: Panic when computing extract_variable with macros

Wilfred:
extract_variable assumed that expanding macros and then mapping them back to original source code would produce a span for a single expression.

This isn't always true for macros with multiple arguments. If the macro expands and uses multiple arguments, we can end up trying to parse `x, y` from `foo!(x, y)`.

Instead, don't offer the assist if we can't map back to a valid Rust expression. This also removes an .unwrap() call (added in 38b37c1f38f94a02098da6d4694b30e544a4ae69).

AI disclosure: Code partly written by GPT-5.6, but commit message, comments and review by me.

### DIFF
--- a/crates/ide-assists/src/handlers/extract_variable.rs
+++ b/crates/ide-assists/src/handlers/extract_variable.rs
@@ -95,7 +95,7 @@ pub(crate) fn extract_variable(acc: &mut Assists, ctx: &AssistContext<'_, '_>) -
     let node = node.ancestors().take_while(|anc| anc.text_range() == node.text_range()).last()?;
     let range = node.text_range();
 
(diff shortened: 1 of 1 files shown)


ChayimFriedman2 (approved):
Ideally we'd support that as well, but for preventing the panic is more important.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23238/44fa5eaa45e3f2811d1035e6a970ccaaaadbbc69..477dc01a6e20b24a68c5613dd6032f2791ab32a8)*

ChayimFriedman2:
GitHub Actions incident...

PR #23240, opened 2026-08-26
feat: implement a simple hover auto traits in rpit

A4-Tacks:
Partial implement of rust-lang/rust-analyzer#23237

Example
---
```rust
//- minicore: send, unpin
fn foo() -> $0impl Trait {
    &raw const ()
}
```

->

> ```rust
> impl Trait + Unpin
> impl Trait = *const ()
> ```
>
> ---
>
> ```rust
> impl
> ```
>
> ---
>
> keyword docs

### DIFF
--- a/crates/hir-def/src/lang_item.rs
+++ b/crates/hir-def/src/lang_item.rs
@@ -322,6 +322,9 @@ pub fn crate_notable_traits(db: &dyn SourceDatabase, krate: Crate) -> Option<Box
     let mut traits = Vec::new();
 
     let crate_def_map = crate_def_map(db, krate);
+    if !crate_def_map.features().doc_notable_trait {
+        return None;
+    }
 
     for (_, module_data) in crate_def_map.modules() {
         for def in module_data.scope.declarations() {
@@ -341,6 +344,9 @@ pub fn crate_auto_traits(db: &dyn SourceDatabase, krate: Crate) -> Option<Box<[T
     let mut traits = Vec::new();
 
     let crate_def_map = crate_def_map(db, krate);
+    if !crate_def_map.features().auto_traits {
+        return None;
+    }
 
     for (_, module_data) in crate_def_map.modules() {
         for def in module_data.scope.declarations() {
--- a/crates/hir-def/src/lang_item.rs
+++ b/crates/hir-def/src/lang_item.rs
@@ -336,6 +336,25 @@ pub fn crate_notable_traits(db: &dyn SourceDatabase, krate: Crate) -> Option<Box
     if traits.is_empty() { None } else { Some(traits.into_iter().collect()) }
 }
 
+#[salsa::tracked(returns(as_deref))]
+pub fn crate_auto_traits(db: &dyn SourceDatabase, krate: Crate) -> Option<Box<[TraitId]>> {
+    let mut traits = Vec::new();
+
+    let crate_def_map = crate_def_map(db, krate);
+
+    for (_, module_data) in crate_def_map.modules() {
+        for def in module_data.scope.declarations() {
+            let ModuleDefId::TraitId(trait_) = def else { continue };
+            let sig = crate::signatures::TraitSignature::of(db, trait_);
+            if sig.flags.contains(crate::signatures::TraitFlags::AUTO) {
+                traits.push(trait_);
+            }
+        }
+    }
+
+    if traits.is_empty() { None } else { Some(traits.into_iter().collect()) }
+}
+
 macro_rules! language_item_table {
     (
         $LangItems:ident =>
--- a/crates/ide/src/hover/render.rs
+++ b/crates/ide/src/hover/render.rs
@@ -261,6 +261,45 @@ pub(super) fn keyword(
     Some(HoverResult { markup, actions })
 }
 
+pub(super) fn rpit(
+    sema: &Semantics<'_, RootDatabase>,
+    _config: &HoverConfig<'_>,
+    token: &SyntaxToken,
+    edition: Edition,
+    display_target: DisplayTarget,
+) -> Option<HoverResult> {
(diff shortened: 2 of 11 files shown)


crates/hir-def/src/lang_item.rs
```diff
@@ -336,6 +336,25 @@ pub fn crate_notable_traits(db: &dyn SourceDatabase, krate: Crate) -> Option<Box
     if traits.is_empty() { None } else { Some(traits.into_iter().collect()) }
 }
 
+#[salsa::tracked(returns(as_deref))]
```
ChayimFriedman2, line 342: Should IMO be the same query with `crate_notable_traits()`. Also both should check the presence of the unsafe feature like we do for `crate_lang_items()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23240/919d6c2b4cfe23e608854c927cff41e0f2b45c7d..061ec05182837fadc9f6e6d3d001abd128100529)*

crates/ide/src/hover/render.rs
```diff
@@ -261,6 +261,45 @@ pub(super) fn keyword(
     Some(HoverResult { markup, actions })
 }
 
+pub(super) fn rpit(
```
ChayimFriedman2, line 264: IMO this is the wrong way to fix. We should instead do that in display infra, for every printed opaque.

Granted, it'll need more work because we don't currently carry the `Option<GenericDefId>` required for trait solving (to get `trait_environment()`) in display.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23240/919d6c2b4cfe23e608854c927cff41e0f2b45c7d..061ec05182837fadc9f6e6d3d001abd128100529)*

ChayimFriedman2 (changes requested):

crates/ide/src/hover/render.rs
A4-Tacks: > IMO this is the wrong way to fix. We should instead do that in display infra, for every printed opaque.

I think these are two different features, regarding RPIT and any expressions

```rust
fn foo() -> Cell<impl Trait> { Cell::new(2) }
               //^^^^ impl Trait + Send + Sync + ...
               //     impl Trait = i32
```

```rust
fn foo() -> Cell<impl Trait> { Cell::new(2) }
                             //^^^^^^^^^^^^ Cell<i32>
                             //             implement auto traits: Send + ...
```

```rust
fn foo() -> Cell<impl Trait> { Cell::new(2) }
                                       //^ i32
                                       //  implement auto traits: Send + Sync + ...
```

ChayimFriedman2: No. I'm not saying to add a clause "implements auto traits" similar to "implements notable traits", only to add this information when displaying opaques.

A4-Tacks: I think this is a feature that is bound to the function return, and the opaque semantics in other places are different, such as in the `impl Trait` of the function parameter

ChayimFriedman2: This is a feature that is relevant to any opqaue - RPIT, ATPIT, ....

A4-Tacks: Hmmm…, On the other hand, the current implementation is simple and very practical (in the stable version)

Veykril:
I feel like we should also generally render these for function hovers? Given auto trait leakage is here to stay showing them in all relevant places where we render the rpit feels appropriate, or at least for Send and Sync (more unconditionally is probably too much)

A4-Tacks:
> I feel like we should also generally render these for function hovers?

I hope it's an expression. The last time the functions displayed 'notable traits', it caused me a lot of pain

> Given auto trait leakage is here to stay showing them in all relevant places where we render the rpit feels appropriate

These places should render 'auto traits' instead of 'rpit'

For expressions, they should be simple, but for types (especially generics), I don't think we're ready yet

```text
[2].iter();
         Iter<'_, i32>
         ─────────────────────────────────────────────────
         Implements notable traits: Iterator<Item = &i32>
         ─────────────────────────────────────────────────
         Implements auto traits:    Send + Sync + Unpin +
         RefUnwindSafe + UnwindSafe + Freeze + UnsafeUnpin
```

ChayimFriedman2:
I argue we should only show `Send` and `Sync`. The rest auto traits are just not interesting (`Send` and `Sync` could also be not interesting, but at least they sometimes are).

A4-Tacks:
I think we should show all, but we can do some heuristic sorting or hard coding to put more useful results at the forefront

ChayimFriedman2:
What other auto traits you consider interesting?

 - `Freeze` and `UnsafeUnpin` are unstable implementation details.
 - `UnwindSafe` and `RefUnwindSafe` might only be interesting when calling `catch_unwind()`, and easy to fix by adding `AssertUnwindSafe`.
 - `Unpin`, maybe.

A4-Tacks:
I sometimes notice `Freeze`, but the main reason is that I believe this feature should not hide the output, or at least use a blacklist instead of a whitelist

ChayimFriedman2:
Given auto traits are perma-unstable feature, I absolutely think this should be a whitelist.

A4-Tacks:
Considering unstable rust users, won't they feel confused about r-a when using unstable features?

ChayimFriedman2:
Auto traits are perma-unstable. They will not be stabilized. If someone is using them, that's on them.

A4-Tacks:
Maybe, when someone complains, consider whether to improve or not

>  - `Freeze` and `UnsafeUnpin` are unstable implementation details.

Although `Freeze` is unstable, it is a useful reference for me when using some libraries

PR #23242, opened 2026-08-26
fix: range pattern inside a parenthesis parsed as tuple pattern

### LINKED ISSUE #22985: FP `expected i32, found ({unknown},)`, match on const, pattern
```rust
fn main() {
    const PAT: u128 = 1;

    match 0 {
        (.. PAT) => {}
        _ => {}
    }
}
````

this compiles with rustc but `rust-analyzer diagnostics .` shows errors
```
at crate ra, file /tmp/ra/src/main.rs: Error RustcHardError("E0308") from LineCol { line: 4, col: 8 } to LineCol { line: 4, col: 16 }: expected i32, found ({unknown},)

diagnostic scan complete
```

### PR DESCRIPTION
**Pull request by dfireBird**
Fixes rust-lang/rust-analyzer#22985 

Let me know if we want to split up the expression.

### CODE REVIEW

File: crates/parser/src/tests/top_entries.rs [hunk 1]
```diff
@@ -335,6 +335,174 @@ fn expr() {
     );
 }
 
+#[test]
+fn range_pattern_with_paren() {
+    check(
+        TopEntryPoint::Expr,
+        r#"
+match 1 {
```
A4-Tacks: Why not create an inline test? Like `// test or_pattern`


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23242/4208616e5b268a94e742ebb836f666e9dcf22e4a..349fc354cac59eab8837ca6e61d7ff9b30244ad7)*

dfireBird: I used this test, to debug the issue and just kept it :sweat_smile: 
I could replace this.

A4-Tacks: Inline testing includes syntax tree, you need `cargo xtask codegen`

You see `crates/parser/test_data/`

dfireBird: Yeah I missed to check the last line, sorry.

dfireBird: Should I remove this test and only inline then?

A4-Tacks: Yes

dfireBird: I have changed it range_pat inline itself.

crates/parser/src/grammar/patterns.rs
```diff
@@ -97,6 +97,12 @@ fn pattern_single_r(p: &mut Parser<'_>, recovery_set: TokenSet) {
     //         (1.., _) => (),
     //         (..=2, _) => (),
     //     }
+    //
+    //     match () {
+    //         (..1) => (),
+    //         (..=3) => (),
+    //         (..2 | 4) => (),
+    //     }
```
A4-Tacks: You can get a clean patch by inserting this match before the original match instead of after it


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23242/70d74f4d134c45b073c82167fb7e7d61334bd8f5..23334ef1d723d2e40524c419c8d3055bdb0b9203)*

dfireBird: Thanks.

A4-Tacks (approved):

PR #23256, opened 2026-08-30
fix: issue 22826 

### LINKED ISSUE #22826: Generic shown on hover is incorrect when there is a `impl Trait` in the inputs and a const generic
**rust-analyzer version**: rust-analyzer version: 0.3.2971-standalone

**rustc version**: rustc 1.96.1 (31fca3adb 2026-06-26) (Arch Linux rust 1:1.96.1-1)

**relevant settings**: no relevant setting

```rust
pub fn main() {
    let lengths: usize = generic_tn([2, 3], |[v, w]| v.len() + w.len());
    //                   ^~~~~~~~~~ Hovering over this function shows:
    // example
    // pub fn generic_tn<T, const N: usize>(sizes: [usize; {const}], mut func: impl FnMut([Vec<u8>; {const}]) -> T) -> T
    // ---------------------------------
    // T = impl FnMut([Vec<u8>; 2]) -> …
    println!("{lengths}");
}

pub fn generic_tn<T, const N: usize>(
    sizes: [usize; N],
    mut func: impl FnMut([Vec<u8>; N]) -> T,
) -> T {
    let buffers: [Vec<u8>; N] = std::array::from_fn(|i| vec![0u8; sizes[i]]);
    func(buffers)
}
```

Hovering over a function with a generic output `T` and a generic input specified via the `impl Trait`-sugar mistakenly assigns to `T` the unnamed input generic rather than the output.

In the example above it displays
```rust
T = impl FnMut([Vec<u8>; 2]) -> usize
```
but really it should be
```rust
T = usize
```

The fact that `T` is an output is not relevant. This snippet has the same problem
```rust
pub fn main() {
    let same: bool = generic_tn([2, 3], |[v, w]| v.len() + w.len(), 5usize);
    println!("{same}");
}

pub fn generic_tn<T: PartialEq, const N: usize>(
    sizes: [usize; N],
    mut func: impl FnMut([Vec<u8>; N]) -> T,
    compare: T
) -> bool {
    let buffers: [Vec<u8>; N] = std::array::from_fn(|i| vec![0u8; sizes[i]]);
    func(buffers) == compare
}
```

The fact that there is a `const N` generic is important as in the snippet below the reported type is correct

```rust
pub fn main() {
    let length: usize = generic_tn(2, |v| v.len());
    //                   ^~~~~~~~~~ Hovering over this function shows:
    // example
    // pub fn generic_tn<T>(size: usize, mut func: impl FnMut(Vec<u8>) -> T) -> T
    // ---------
    // T = usize
    println!("{length}");
}

pub fn generic_tn<T>(
    size: usize,
    mut func: impl FnMut(Vec<u8>) -> T,
) -> T {
    let buffer: Vec<u8> = vec![0; size];
    func(buffer)
}
```

---

Image

Image

Image

j-ricardo-goncalves:
fixes issue by using a different approach on parent_len in crate/hir/src/lib.rs
added  coverage test

seems to solve the issue while doing the least possible changes

Fixes https://github.com/rust-lang/rust-analyzer/issues/22826.

### DIFF
--- a/crates/hir/src/lib.rs
+++ b/crates/hir/src/lib.rs
@@ -3956,21 +3956,9 @@ impl<'db> GenericSubstitution<'db> {
             TypeOrConstParamData::TypeParamData(param) => Some(param.name.clone()),
             TypeOrConstParamData::ConstParamData(_) => None,
         });
-        let parent_len = self.subst.len()
-            - generics
-                .iter_type_or_consts()
-                .filter(|g| matches!(g.1, TypeOrConstParamData::TypeParamData(..)))
-                .count();
-        let container_params = self.subst.as_slice()[..parent_len]
-            .iter()
-            .filter_map(|param| param.ty())
-            .zip(container_type_params.into_iter().flatten());
-        let self_params = self.subst.as_slice()[parent_len..]
-            .iter()
-            .filter_map(|param| param.ty())
-            .zip(type_params);
-        container_params
-            .chain(self_params)
+        self.subst
+            .types()
+            .zip(container_type_params.into_iter().flatten().chain(type_params))
             .filter_map(|(ty, name)| {
                 Some((
                     name?.symbol().clone(),
--- a/crates/ide/src/hover/tests.rs
+++ b/crates/ide/src/hover/tests.rs
@@ -12021,3 +12021,31 @@ fn bar() {
         "#]],
     );
 }
+
+#[test]
+fn subst_impl_trait_arg_with_const_generic() {
+    check(
+        r#"
+fn main() {
+    generic$0_tn([()], 1);
+}
+
+fn generic_tn<T, const N: usize>(_: [T; N], _: impl Sized) {}
+"#,
+        expect![[r#"
+            *generic_tn*
+
+            ```rust
+            ra_test_fixture
+            ```
+
+            ```rust
+            fn generic_tn<T, const N: usize>(_: [T; {const}], _: impl Sized)
+            ```
+
+            ---
+
+            `T` = `()`
+        "#]],
+    );
+}


j-ricardo-goncalves:
#22826

crates/ide/src/hover/tests.rs
```diff
@@ -12021,3 +12021,62 @@ fn bar() {
         "#]],
     );
 }
+
+#[test]
+fn subst_impl_trait_arg_with_const_generic() {
+    check(
```
ChayimFriedman2, line 12027: This test is more complicated than necessary and you also don't need more than one. This is enough:

```rust
//- minicore: sized
fn main() {
    generic$0_tn([()], 1);
}

fn generic_tn<T, const N: usize>(_: [T; N], _: impl Sized) {}
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23256/e96ea7a5805992ec868e8ab185bc371267048421..bb4d8c840d69f9910bcee681b593cc2ee8f995a0)*

crates/hir/src/lib.rs
```diff
@@ -3956,11 +3956,14 @@ impl<'db> GenericSubstitution<'db> {
...
-                .filter(|g| matches!(g.1, TypeOrConstParamData::TypeParamData(..)))
-                .count();
+        let parent_len = container
+            .and_then(|container| match container {
+                ItemContainerId::ImplId(container) => Some(container.into()),
+                ItemContainerId::TraitId(container) => Some(container.into()),
+                _ => None,
+            })
+            .map(|container| GenericParams::of(db, container).len())
+            .unwrap_or(0);
```
ChayimFriedman2: This is not good enough because of late-bound lifetimes.

I think the best solution is to avoid computing the parent length at all, instead directly chain
```
self.subst.types().zip(container_type_params.into_iter().flatten().chain(type_params)).filter_map(|(ty, name)| {
                Some((
                    name?.symbol().clone(),
                    Type { ty: EarlyBinder::bind(ty), owner: self.owner },
                ))
            })
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23256/e96ea7a5805992ec868e8ab185bc371267048421..bb4d8c840d69f9910bcee681b593cc2ee8f995a0)*

ChayimFriedman2 (changes requested):

crates/hir/src/lib.rs
j-ricardo-goncalves: did a similar approach in the new commit

crates/ide/src/hover/tests.rs
j-ricardo-goncalves: implemented your example

ChayimFriedman2 (approved):
Please squash.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23256/e96ea7a5805992ec868e8ab185bc371267048421..c5d06d87a14d35b581e6ad98ee4767a08f79d369)*

j-ricardo-goncalves:
squashed

j-ricardo-goncalves:
I think its done now, tried to clean up the git thing.

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23256/d1a301c8d4204f90aa3fe6877420aef19398d95b..4d5556291c1612ed59cb58c2a31b061b661d005a)*

PR #23257, opened 2026-08-30
fix: Fix `NamedTempFile`

ChayimFriedman2:
- Use `OpenOptions::write(true)`, otherwise it always fails.
 - If unable to copy the proc macro DLL in the proc macro server, use the original instead to at least succeed.
 - For Cargo.lock, use a temporary directory instead as Cargo only accepts literally-named Cargo.lock files, and also because `/proc/self/fd` does not work for other processes. Fortunately unlike the proc macro server we do run `Drop` for those.

Thanks to @onlycs for discovering those.

I'm merging this myself because this can make the proc macro server not working at all on Windows. @lnicola may I also invite a sync?

### DIFF
--- a/crates/proc-macro-srv/src/dylib.rs
+++ b/crates/proc-macro-srv/src/dylib.rs
@@ -27,7 +27,7 @@ impl Expander {
         let lib = lib.canonicalize_utf8()?;
         let modified_time = fs::metadata(&lib).and_then(|it| it.modified())?;
 
-        let file = ensure_file_with_lock_free_access(lib)?;
+        let file = ensure_file_with_lock_free_access(lib);
         let library = ProcMacroLibrary::open(file.path())?;
 
         Ok(Expander { inner: library, modified_time, _file: file })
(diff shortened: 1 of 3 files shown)


lnicola:
@ChayimFriedman2 would tomorrow be fine, so it matches the normal release syncs?

ChayimFriedman2:
Oh the previous PR wasn't synced yet (if it would, Windows nightly users wouldn't have proc macro support at all). Then yes this is fine.

lnicola:
@ChayimFriedman2 do you want to add a dependency on `tracing`? See https://triage.rust-lang.org/gha-logs/rust-lang/rust/99399134682#L2026-08-31T06:34:12.1438486Z

ChayimFriedman2:
Wait. Don't we check proc-macro-srv in CI? That's not good.

Anyway it doesn't really matter to me if we remove this logging or add tracing as a dependency.

lnicola:
I don't remember exactly, but if we do, we probably check on Linux only.

ChayimFriedman2:
That shouldn't matter for this because stdx is still included for Linux.

But we also should check on other platforms.

PR #23266, opened 2026-08-31
render const value in completions label details

BenjaminBrienen:
I think this would be neat. You may be brutal with the feedback on the code.

Screenshot_2026-08-31_21-06-16

### DIFF
--- a/crates/ide-completion/src/completions/format_string.rs
+++ b/crates/ide-completion/src/completions/format_string.rs
@@ -37,19 +37,24 @@ pub(crate) fn format_string(
     });
     ctx.scope.process_all_names(&mut |name, scope| {
         if let ScopeDef::ModuleDef(module_def) = scope {
+            let mut const_value = None;
             let symbol_kind = match module_def {
-                ModuleDef::Const(..) => SymbolKind::Const,
+                ModuleDef::Const(c) => {
+                    const_value = Some(c);
+                    SymbolKind::Const
+                }
                 ModuleDef::Static(..) => SymbolKind::Static,
                 _ => return,
             };
 
-            CompletionItem::new(
+            let mut builder = CompletionItem::new(
                 CompletionItemKind::SymbolKind(symbol_kind),
                 source_range,
                 name.display_no_db(ctx.edition).to_smolstr(),
                 ctx.edition,
-            )
-            .add_to(acc, ctx.db);
+            );
+            builder.const_value(const_value, ctx.db, ctx.display_target);
+            builder.add_to(acc, ctx.db);
         }
     });
 }
--- a/crates/ide-completion/src/item.rs
+++ b/crates/ide-completion/src/item.rs
@@ -2,7 +2,8 @@
 
 use std::{fmt, mem};
 
-use hir::Mutability;
+use hir::db::HirDatabase;
+use hir::{DisplayTarget, Mutability};
 use ide_db::text_edit::TextEdit;
 use ide_db::{
     RootDatabase, SnippetCap, SymbolKind, documentation::Documentation,
@@ -12,7 +13,7 @@ use itertools::Itertools;
 use macros::UpmapFromRaFixture;
 use smallvec::SmallVec;
 use stdx::{format_to, impl_from, never};
-use syntax::{Edition, SmolStr, TextRange, TextSize, format_smolstr};
+use syntax::{Edition, SmolStr, TextRange, TextSize, ToSmolStr, format_smolstr};
 
 use crate::{
     context::{CompletionContext, PathCompletionCtx},
@@ -497,6 +498,7 @@ impl CompletionItem {
             imports_to_add: Default::default(),
             doc_aliases: vec![],
             adds_text: None,
+            const_value: None,
             edition,
         }
     }
@@ -534,6 +536,7 @@ pub(crate) struct Builder {
     trait_name: Option<SmolStr>,
     doc_aliases: Vec<SmolStr>,
     adds_text: Option<SmolStr>,
+    const_value: Option<SmolStr>,
     label: SmolStr,
     insert_text: Option<String>,
     is_snippet: bool,
@@ -602,6 +605,9 @@ impl Builder {
             if !lookup_doc_aliases.is_empty() {
                 lookup = format_smolstr!("{lookup}{lookup_doc_aliases}");
             }
+        }
+        if let Some(const_value) = self.const_value {
+            to_detail_left(format_args!(" = {}", const_value.trim()));
         }
         if let Some(adds_text) = self.adds_text {
             to_detail_left(format_args!("(adds {})", adds_text.trim()));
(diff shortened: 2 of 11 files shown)


sk0g:
This would be very helpful! RustRover does something similar and it's quite helpful when looking over convenience shortcuts for types you aren't familiar with. 

Would it be possible to align the syntax of this a bit more with that of associated type defaults? https://rust-lang.github.io/rfcs/2532-associated-type-defaults.html

Eg 
```rs
            ct A  = 123  pub const A: i32
// Becomes
            ct A   pub const A: i32 = 123
```

Edit: actually not sure, looking at the examples in the screenshot, how it works right now would look a lot more reasonable than shoving everything to the right edge.

BenjaminBrienen:
CI failure is a regression in https://github.com/rust-lang/rust-analyzer/pull/23257

ChayimFriedman2:
You need to rebase on master.

BenjaminBrienen:
done

crates/ide-completion/src/completions/format_string.rs
```diff
@@ -37,19 +37,25 @@ pub(crate) fn format_string(
...
-            CompletionItem::new(
+            let mut builder = CompletionItem::new(
                 CompletionItemKind::SymbolKind(symbol_kind),
                 source_range,
                 name.display_no_db(ctx.edition).to_smolstr(),
                 ctx.edition,
-            )
-            .add_to(acc, ctx.db);
+            );
+            builder.const_value(const_value).clone().add_to(acc, ctx.db);
```
ChayimFriedman2: Why the `clone()` and why store in a variable `builder`?

ChayimFriedman2 (changes requested):
I think it will be better to display the evaluated const value, like we do in other places (e.g. hover).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23266/ce43a12d9fe4a24e5d6a86d29a57234efcc902b7..fc5a9efe7e12e6e060f22d238b2b77738c3595d1)*

crates/ide-completion/src/completions/format_string.rs
BenjaminBrienen: I got an error about needing to create a binding for a dropped temporary and then it said that I had to clone.

without clone:
```
error[E0507]: cannot move out of a mutable reference
  --> crates/ide-completion/src/completions/format_string.rs:58:13
   |
58 |             builder.const_value(const_value).add_to(acc, ctx.db);
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ------------------- value moved due to this method call
   |             |
   |             move occurs because value has type `item::Builder`, which does not implement the `Copy` trait
   |
note: `completions::<impl item::Builder>::add_to` takes ownership of the receiver `self`, which moves value
  --> crates/ide-completion/src/completions.rs:70:26
   |
70 |     pub(crate) fn add_to(self, acc: &mut Completions, db: &RootDatabase) {
   |                          ^^^^
help: consider cloning the value if the performance cost is acceptable
   |
58 |             builder.const_value(const_value).clone().add_to(acc, ctx.db);
   |                                             ++++++++
```

Apparently the binding is not necessary, so fixing that.

BenjaminBrienen:
Done. I tried adding a unit test for rendering u32::MAX, but the completion didn't show up at all. It does work, though:
Screenshot_2026-09-01_11-33-48

crates/ide-completion/src/completions/format_string.rs
ChayimFriedman2: Hmm, then you do need the variable but not the clone:
```rust
let mut builder = ...;
builder.const_value(...);
builder.add_to(...);
```

ChayimFriedman2:
You probably need to include something from minicore.

BenjaminBrienen:
Ok, I added that under `uint_consts`. Hopefully the name is ok.

crates/test-utils/src/minicore.rs
```diff
@@ -2433,6 +2434,32 @@ pub mod f64 {
 }
 // endregion:float_consts
 
+// region:uint_consts
```
ChayimFriedman2: Do not add this to minicore. Just define the consts in the test.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23266/4131c530f8b45771f376e32f230eb00d4600340c..26bf4c2350a52b641449a93450f75e086df76022)*

crates/ide-completion/src/completions/format_string.rs
```diff
@@ -49,6 +55,8 @@ pub(crate) fn format_string(
                 name.display_no_db(ctx.edition).to_smolstr(),
                 ctx.edition,
             )
+            .const_value(const_value)
+            .clone()
```
ChayimFriedman2: Like I said you don't need this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23266/4131c530f8b45771f376e32f230eb00d4600340c..26bf4c2350a52b641449a93450f75e086df76022)*

crates/ide-completion/src/completions/format_string.rs
```diff
@@ -37,8 +37,14 @@ pub(crate) fn format_string(
     });
     ctx.scope.process_all_names(&mut |name, scope| {
         if let ScopeDef::ModuleDef(module_def) = scope {
+            let mut const_value = None;
             let symbol_kind = match module_def {
-                ModuleDef::Const(..) => SymbolKind::Const,
+                ModuleDef::Const(c) => {
+                    if let Ok(value) = c.eval(ctx.db) {
```
ChayimFriedman2: Preferably, if we fail to evaluate we'd still show the written value. You can make `const_value()` take `Option<hir::Const>` so the code will be shared.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23266/4131c530f8b45771f376e32f230eb00d4600340c..26bf4c2350a52b641449a93450f75e086df76022)*

ChayimFriedman2 (changes requested):

BenjaminBrienen:
Is "feaures" a typo in rustc? 😂️

BenjaminBrienen:
https://github.com/rust-lang/rust/pull/162115

crates/ide-completion/src/item.rs
```diff
@@ -681,6 +687,21 @@ impl Builder {
...
         self
     }
+    pub(crate) fn const_value(
+        &mut self,
+        const_value: Option<hir::Const>,
+        db: &dyn HirDatabase,
+        display_target: DisplayTarget,
+    ) -> &mut Builder {
+        if let Some(const_value) = const_value {
+            if let Ok(evaluted_value) = const_value.eval(db) {
```
ChayimFriedman2:
```suggestion
            if let Ok(evaluated_value) = const_value.eval(db) {
```

ChayimFriedman2 (approved):
Just fix the typos and LGTM.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23266/4131c530f8b45771f376e32f230eb00d4600340c..7c6a3f8e62b7080318d44b0e1db241caf1cdc719)*

ChayimFriedman2:
You overrode the commit message...

ChayimFriedman2 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-analyzer/23266/4131c530f8b45771f376e32f230eb00d4600340c..a255dd53b8f3dc790fe4703bdeb7900f23d547ef)*

lnicola:
image

BenjaminBrienen:
@lnicola all good?

lnicola:
Yeah, I wanted a clean URL to the screenshot.

nurmohammed840:
Can there be a setting to limit a value to N (3, 4) digits after the decimal point?

```
PI = 3.1415...
```

ChayimFriedman2:
No, I don't think such setting carries its weight.

